PR #2, opened 2022-03-08
added brave to chrome-wheel-scroll-tabs

ArjixWasTaken:


m417z:
Thanks, added in https://github.com/ramensoftware/windhawk-mods/pull/4.
Note that this repository is the place from which Windhawk fetches the mods, you can't just update mods without also updating the version number. There are other restrictions too, refer to the readme. That's OK, Windhawk is new and I'll see how to make it clearer with time. Meanwhile, if you have any other improvements for my mods, you can submit them [here](https://github.com/m417z/my-windhawk-mods).

PR #8, opened 2022-04-19
Add lm-close-windbg-file-changed

learn-more:
This mod automatically dismisses the popup from WinDbg notifying about a file change.
When automating WinDbg, this dialog blocks progress.

### DIFF
--- a/mods/lm-close-windbg-file-changed.wh.cpp
+++ b/mods/lm-close-windbg-file-changed.wh.cpp
@@ -1,0 +1,79 @@
+// ==WindhawkMod==
+// @id           lm-close-windbg-file-changed
+// @name         Close WinDbg's file changed dialog
+// @description  Automatically rejects (or accepts) the dialog from WinDbg telling that a file is changed
+// @version      1.0
+// @author       Mark Jansen
+// @github       https://github.com/learn-more
+// @twitter      https://twitter.com/learn_more
+// @include      windbg.exe
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Close WinDbg's file changed dialog
+This mod automatically dismisses the popup from WinDbg notifying about a file change.
+When automating WinDbg, this dialog blocks progress.
+
+![demonstration](https://i.imgur.com/8CLg83M.png)
+*/
+// ==/WindhawkModReadme==
+
+// ==WindhawkModSettings==
+/*
+- Accept: false
+  $description: Enable this to accept the dialog instead of rejecting it (reloading the file instead of keeping the old file)
+*/
+// ==/WindhawkModSettings==
+
+
+#include <string>
+
+static int g_Accept = 0;
+
+
+typedef decltype(&MessageBoxW) MESSAGEBOXW;
+MESSAGEBOXW pMessageBoxW;
+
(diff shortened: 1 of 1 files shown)


mods/lm-close-windbg-file-changed.wh.cpp
```diff
@@ -0,0 +1,77 @@
+// ==WindhawkMod==
+// @id           lm-close-windbg-file-changed
+// @name         Close WinDbg's file changed dialog
+// @description  Automatically rejects (or accepts) the dialog from WinDbg telling that a file is changed
+// @version      1.0
+// @author       Mark Jansen
+// @github       https://github.com/learn-more
+// @twitter      https://twitter.com/learn_more
```
m417z: To verify your Twitter account, please send me (https://twitter.com/m417z) a direct message with the following content:

> I attest that I'm the sole owner of both this Twitter account (https://twitter.com/learn_more) and the following GitHub account: https://github.com/learn-more

P.S. It's not that something looks suspicious, it's just that I commit to Windhawk users that all profiles are verified.

mods/lm-close-windbg-file-changed.wh.cpp
```diff
@@ -0,0 +1,77 @@
...
+/*
+# Close WinDbg's file changed dialog
+This mod automatically dismisses the popup from WinDbg notifying about a file change.
+When automating WinDbg, this dialog blocks progress.
+*/
+// ==/WindhawkModReadme==
+
+// ==WindhawkModSettings==
+/*
+- Accept: 0
```
m417z: FYI, you can use `false` here to have a checkbox instead of a number field. As far as the mod is concerned, the value will be an integer of either `0` or `1`, so the code needs no changes.

mods/lm-close-windbg-file-changed.wh.cpp
```diff
@@ -0,0 +1,77 @@
...
+// @github       https://github.com/learn-more
+// @twitter      https://twitter.com/learn_more
+// @include      windbg.exe
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Close WinDbg's file changed dialog
+This mod automatically dismisses the popup from WinDbg notifying about a file change.
+When automating WinDbg, this dialog blocks progress.
```
m417z: How about adding a screenshot? It should be as easy as uploading it to https://imgur.com/ and adding it here like this:
https://github.com/ramensoftware/windhawk-mods/blob/6c2b4e0ee6af475c451c1ae866e54b312a328195/mods/notepad-dark-mode.wh.cpp#L24

m417z (approved):
Looks good. I left a couple of notes which are totally optional. Just verify your Twitter account and I'll merge it.

mods/lm-close-windbg-file-changed.wh.cpp
learn-more: done.

mods/lm-close-windbg-file-changed.wh.cpp
learn-more: Sure, I'll change this and add a screenshot later today or tomorrow.

mods/lm-close-windbg-file-changed.wh.cpp
learn-more: Screenshot added.

PR #10, opened 2022-04-23
Add lm-vs-solution-icon

learn-more:
This module adds an overlay icon for visual studio 2017 and 2019,
generated from the solution filename


m417z:
I had to fix the extension from `.cpp` to `.wh.cpp`. Not a big deal. Eventually I'll need to think of an automation pipeline that verifies these kinds of things.

learn-more:
Oops, sorry about that.

PR #80, opened 2023-01-04
Add initial author validation workflow

learn-more:
Here is the initial validator workflow.

Some things could be improved, but it serves as a starting point.

- [x] Add a license (I was thinking about the MIT license)
- [x] Clean up PR before merging (drop the test commits)

### DIFF
--- a/.github/pr_validation.py
+++ b/.github/pr_validation.py
@@ -1,0 +1,65 @@
+'''
+PROJECT:     PR Validation for WindHawk
+LICENSE:     MIT (https://spdx.org/licenses/MIT)
+PURPOSE:     Verifies the mod information in the modified mods.
+COPYRIGHT:   Copyright 2023 Mark Jansen <mark.jansen@reactos.org>
+'''
+from optparse import OptionParser
+from pathlib import Path
+import sys
+
+
+def add_warning(file: Path, line: int, message: str):
+    print(f'::warning file={file!s},line={line}::{message}')
+
+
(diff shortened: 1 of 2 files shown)


learn-more:
Here is the build after the 1st commit (no mods changed yet, so the check step will be skipped):
image

Here is the build after the 2nd commit (edited my own mod):
image

And here is the build after the 3rd commit (edited a mod that is not mine):
image

Full build history:
https://github.com/ramensoftware/windhawk-mods/actions/workflows/author_validation.yml

learn-more:
And the warning is now shown in the diff view:

image

learn-more:
@m417z This thing still displays green:
image

Do you want this to also show a warning?
Also, feel free to push to this PR to test some stuff out, if you want to play around with it.

m417z:
Looks great, and a good basis for other checks such as the file name matching the mod id (to also catch mistakes [like this](https://github.com/ramensoftware/windhawk-mods/pull/10#issuecomment-1107640496)).

> Do you want this to also show a warning?

Yes, I think that would be better, otherwise the warning is easier to miss.

learn-more:
Now the check fails on this.
There was no option to make it into a warning, but at least it's clearly visible

m417z:
Looks good, thanks. Is it ready to be merged?

learn-more:
It should be, unless you want me to push some test commits first?

m417z:
Thanks! That's fine, we'll test in production :)

PR #95, opened 2023-02-14
No Focus Rectangle 1.0

ItsProfessional:
This mod removes the windows focus rectangle. For more information, read the mod description.
It accomplishes this by hooking DrawFocusRect in user32.dll and returning TRUE.

I programmed it to return TRUE instead of FALSE because returning FALSE implies to the program that called the function that the function has failed (according to MSDN). Returning TRUE pretends that the focus rectangle has been drawn, while its actually not.

### DIFF
--- a/no-focus-rectangle.wh.cpp
+++ b/no-focus-rectangle.wh.cpp
@@ -1,0 +1,56 @@
+// ==WindhawkMod==
+// @id              no-focus-rectangle
+// @name            No Focus Rectangle
+// @description     Removes the focus rectangle
+// @version         1.0.0
+// @author          ItsProfessional
+// @github          https://github.com/ItsProfessional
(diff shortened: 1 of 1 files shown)


m417z:
The file should have been in the `mods` folder. I'll fix it manually.
@learn-more an opportunity to improve the gate file. I'm OK with it to be overly strict and fail on anything that is not a new mod submission or a mod update. That means, exactly one file in the mods folder, file name matches the id in the metadata, etc. according to the readme. In the rare case that it's not a mod submission/update PR, the gate can be ignored.

ItsProfessional:
Oh, lol. I usually make PRs by cloning and pushing my fork, but this time I got lazy and just added the file using github's interface and I added it to the root folder by mistake.

PR #99, opened 2023-02-15
Update PR validation to be more strict

learn-more:
As requested in #95

### DIFF
--- a/.github/pr_validation.py
+++ b/.github/pr_validation.py
@@ -11,6 +11,7 @@ import sys
 
 def add_warning(file: Path, line: int, message: str):
     print(f'::warning file={file!s},line={line}::{message}')
+    return 1
 
 
 def parse_file(file: Path, expected_author):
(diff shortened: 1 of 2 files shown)


learn-more:
@m417z This PR immediately shows how it works on other files.
I did only validate .cpp files on their contents, because parsing .ico files and the likes could give weird results.
Let me know if you want to drop that restraint and try to validate every file.

m417z:
No need to parse non-`.wh.cpp` files, but let's warn about them, since there should only be a single `mods/<mod-id>.wh.cpp` file per submission. And generally, it would be best to have a validation follow the readme instructions as closely as possible, e.g. only one added/modified source file, no change in the author name, version is increased (if not a new mod).

learn-more:
> No need to parse non-`.wh.cpp` files, but let's warn about them, since there should only be a single `mods/<mod-id>.wh.cpp` file per submission. And generally, it would be best to have a validation follow the readme instructions as closely as possible, e.g. only one added/modified source file, no change in the author name, version is increased (if not a new mod).

* Every file that is not in the mods folder will generate a warning
* Every file that ends with `.cpp` will be inspected 'as mod'
* Every mod that will be inspected needs to have the full extension `.wh.cpp`

I'll add a validation that warns if there is more than one file changed.

ItsProfessional:
Maybe add a check that compiles the mod and fails if error

m417z:
Thanks

PR #100, opened 2023-02-17
Classic Taskbar Context Menu 1.0

ItsProfessional:
This mod restores the classic taskbar context menu.

### DIFF
--- a/mods/classic-taskbar-context-menu.wh.cpp
+++ b/mods/classic-taskbar-context-menu.wh.cpp
@@ -6,7 +6,7 @@
 // @author          ItsProfessional
 // @github          https://github.com/ItsProfessional
 // @include         explorer.exe
-// @compilerOptions -lgdi32
+// @compilerOptions -lgdi32 -lcomctl32
 // ==/WindhawkMod==
 
 // ==WindhawkModReadme==
@@ -39,6 +39,9 @@ struct {
 // Windows Header Files:
 #include <windows.h>
 #include <shellapi.h>
+#include <commctrl.h>
+#include <string.h>
+#include <unordered_set>
 
 //msvcrt.lib;msvcrtd.lib
 #ifdef _DEBUG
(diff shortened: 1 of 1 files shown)


mods/classic-taskbar-context-menu.wh.cpp
```diff
@@ -0,0 +1,591 @@
...
+
+		if (fpMenu98Unload != NULL) {
+			fpMenu98Unload(NULL);
+		} else {
+			MessageBoxW(0, L"Failed to unload Menu98!", L"Fatal Error", MB_ICONERROR);
+		}
+	}
+
+
+	CloseHandle(CreateThread(NULL, 0, (LPTHREAD_START_ROUTINE)FreeLibraryAndExitThread, gLibModule, 0, NULL));
```
m417z: Looks good overall. Remove this, Windhawk will unload this mod, no need to do that manually and it can lead to problems.

mods/classic-taskbar-context-menu.wh.cpp
```diff
@@ -0,0 +1,591 @@
...
+
+	// The taskbar on the second screen
+	hWnd_TaskBar_SecondScreen = FindWindowA("Shell_SecondaryTrayWnd", NULL);
+	if (IsWindow(hWnd_TaskBar_SecondScreen)) {
+		OldWndProc_TaskBar_SecondScreen = GetWindowLongPtr(hWnd_TaskBar_SecondScreen, GWLP_WNDPROC);
+		
+        if (OldWndProc_TaskBar_SecondScreen != NULL) SetWindowLongPtr(hWnd_TaskBar_SecondScreen, GWLP_WNDPROC, (LONG_PTR)&WndProc_TaskBar_SecondScreen);
+	}
+
+	menu98Module = LoadLibraryA("menu98.dll");
```
m417z: These references can be seemingly removed as well.

ItsProfessional: I'm not sure what you mean. If you are talking about hWnd_TaskBar, it is used in the wndproc function

m417z: I'm talking about trying to load menu98.dll which will always fail.

ItsProfessional: Oh. why is that?

m417z: I assumed it was part of Taskbar-Context-Menu-Tweaker, but it's seems to be a different tool, so yeah, If the user has it installed it will probably work.

ItsProfessional: You're correct. it is not necessarily a part of taskbar context menu tweaker, but its from the same author https://github.com/rikka0w0/Menu98

PR #110, opened 2023-03-01
Make it clear that the non-immersive taskbar context menu mod does not work on Windows 11

ItsProfessional:


mods/non-immersive-taskbar-context-menu.wh.cpp
```diff
@@ -1,8 +1,8 @@
 // ==WindhawkMod==
-// @id              classic-taskbar-context-menu
-// @name            Classic Taskbar Context Menu
-// @description     Restores the classic taskbar context menu
-// @version         1.0.2
+// @id              non-immersive-taskbar-context-menu
```
m417z: I prefer to keep the ID so that existing users will be able to update. Otherwise it will be treated as a new mod.

ItsProfessional: Okay, I'll leave the ID and the file name intact and the name will be changed, is that fine

PR #115, opened 2023-03-03
Windows 7 Command Bar 1.0.0

ItsProfessional:
This mod restores the Windows 7 file command bar in the file explorer.
It accomplishes this by hooking CoCreateInstance in `ExplorerFrame.dll` (api set dll is `api-ms-win-core-com-l1-1-0.dll`) and returning `REGDB_E_CLASSNOTREG` if the clsid matches a specific value, otherwise it calls it original function and returns it.


yw4z:
@ItsProfessional is that also possible to add an option for hiding it?

Undisputed00x:
I think you're confussing the explorer's Command bar with the Menu bar https://www.elevenforum.com/t/add-or-remove-menu-bar-in-file-explorer-in-windows-11.7643/

PR #117, opened 2023-03-03
Shrink Address Bar Height 1.0.0

ItsProfessional:
This mod shrinks the address bar height in file explorer.
Note: It hooks into every process, because if I only hook explorer.exe, it won't work on save/open dialogs.

### DIFF
--- a/mods/shrink-address-bar-height.wh.cpp
+++ b/mods/shrink-address-bar-height.wh.cpp
@@ -13,6 +13,8 @@
 # Shrink Address Bar Height
 This mod shrinks the address bar height in file explorer.
 
+Note: This mod applies to explorer folders as well as save/open dialogs in all programs.
+
 The code is based on the implementation in [ExplorerPatcher](https://github.com/valinet/ExplorerPatcher).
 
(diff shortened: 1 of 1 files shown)


m417z:
> Note: It hooks into every process, because if I only hook explorer.exe, it won't work on save/open dialogs.

I think it's worth adding this to the readme.

ItsProfessional:
Wouldn't that cause useless confusion? Many people will think that the mod is dangerous or some random nonsense because of that, and I'm not just making this up, I've seen this happen. Also, hooking into all programs is (hopefully) not a bad thing, [atleast according to the ExplorerPatcher dev](https://github.com/valinet/ExplorerPatcher/wiki/Using-ExplorerPatcher-as-shell-extension):
> !!! IMPORTANT !!! Injecting in all process is no mean feat: that means most processes running on you system have this piece of code running into it. While I strive to provide clean and bug free releases, mistakes or problems may appear from time to time, and this has the potential of making more troublesome than usual to recover the system if something goes wrong, so beware before deciding towards using this.

ItsProfessional:
Wait I've just realized that the screenshot is wrong, I'll fix that up

ItsProfessional:
done

PR #118, opened 2023-03-03
Disable Navigation Bar 1.0.0

ItsProfessional:
This mod disables the navigation bar in file explorer.

### DIFF
--- a/mods/disable-navigation-bar.wh.cpp
+++ b/mods/disable-navigation-bar.wh.cpp
@@ -1,0 +1,62 @@
+// ==WindhawkMod==
+// @id              disable-navigation-bar
+// @name            Disable Navigation Bar
+// @description     Disables the navigation bar in file explorer
+// @version         1.0.0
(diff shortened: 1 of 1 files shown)


mods/disable-navigation-bar.wh.cpp
```diff
@@ -0,0 +1,62 @@
+// ==WindhawkMod==
+// @id              disable-navigation-bar
+// @name            Disable Navigation Bar
+// @description     Disables the navigation bar in file explorer
+// @version         1.0.0
+// @author          ItsProfessional
+// @github          https://github.com/ItsProfessional
+// @include         *
```
m417z, line 8: How about adding the same note as with the previous mod regarding targeting all apps?

ItsProfessional: Oh I forgot about that

ItsProfessional:
Done.

PR #126, opened 2023-03-05
Disable Immersive Context Menus 1.0.0

ItsProfessional:
This mod disables immersive context menus in file explorer.


ItsProfessional:
@m417z I've updated the mod can you try now?

PR #127, opened 2023-03-06
Taskbar Toolbar Separators 1.0.0

ItsProfessional:
This mod enables seperators between taskbar toolbars.

### DIFF
--- a/mods/taskbar-toolbar-separators.wh.cpp
+++ b/mods/taskbar-toolbar-separators.wh.cpp
@@ -1,0 +1,109 @@
+// ==WindhawkMod==
+// @id              taskbar-toolbar-separators
+// @name            Taskbar Toolbar Separators
+// @description     Enables seperators between taskbar toolbars
+// @version         1.0.0
+// @author          ItsProfessional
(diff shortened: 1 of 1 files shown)


mods/taskbar-toolbar-separators.wh.cpp
```diff
@@ -0,0 +1,109 @@
...
+    int       Y,
+    int       nWidth,
+    int       nHeight,
+    HWND      hWndParent,
+    HMENU     hMenu,
+    HINSTANCE hInstance,
+    LPVOID    lpParam
+)
+{
+    if ((*((WORD*)&(lpClassName)+1)) && !wcscmp(lpClassName, L"ReBarWindow32"))
```
m417z, line 45:
```
*((WORD*)&(lpClassName)+1)
```
That's not a correct check since it only checks 2 bytes out of 8. The pointer can be, for example, `0x0000_1234_0000_0000` and your check will skip it.

ItsProfessional:
For now, this is a draft. I might fix it later

Issue #45, opened 2022-09-16
Large taskbar icons - Width of taskbar icons


dr-consit:
When setting the size of the taskbar icons, it seems the width is constant.

I tried poking about and naively adding some width hooks, but this failed, probably because these functions aren't present in the relevant dll.

Anyway if you could point me towards a tool allowing me to list the hookable functions in the relevant dlls, I will have another poke about.

Otherwise, I would greatly appreciate a fix for this issue.

dr-consit:
Nvm about the tool request, I got the list by dumping in a loop in windhawk. Didn't really get me any nearer a solution though.
A cleverer man than I must take over.

dr-consit:
I was wondering if it might be feasible to "change" the DPI of the taskbar? 
I tried intercepting the different DPI calls that I thought might be relevant, but was unable to get anywhere. 
Any pointers would be much appreciated.

m417z:
The [Taskbar Labels for Windows 11](https://windhawk.net/mods/taskbar-labels) mod demonstrates how it's possible to change the width of the taskbar items. It's not very easy, and there's at least one unsolved issue with it: When there are too many items on the taskbar, Windows displays the ["taskbar overflow"](https://winaero.com/how-to-enable-the-new-taskbar-overflow-in-windows-11/). Increasing the items of taskbar buttons cause them to to be inaccessible before the taskbar overflow appears. So, for example, your taskbar might show only 20 items at a time, and if you open 21, the taskbar overflow appears, but if you make the items wider, only 15 can now fit. But still, the taskbar overflow appears only after 20 items.

dr-consit:
@m417z That's very cool, thank you.

Do you have any thoughts about the DPI approach? Any pointers to where to start looking, if I want to try that route, as I imagine it might not cause the mentioned issue?

m417z:
I'm not sure, it depends on the way explorer handles DPI. One thing you can try is hooking functions such as `GetDpiForSystem`, `GetDpiForMonitor`, `GetDpiForWindow`. Another thing you can try is hooking `MulDiv` and replacing the numerator when both the numerator and the denominator are 96 (or a numerator that matches your DPI).

imagicalFantasy:
I hope you will find a way to archive it properly one day !

m417z:
[...]

PR #145, opened 2023-03-21
Add taskbar item width setting

dr-consit:
Add functionality to allow the configuration of taskbar item width.

Most of the code is lifted from the taskbar label mod.

This may render this mod incompatible with the label mod?

### DIFF
--- a/mods/taskbar-icon-size.wh.cpp
+++ b/mods/taskbar-icon-size.wh.cpp
@@ -1,6 +1,6 @@
 // ==WindhawkMod==
 // @id              taskbar-icon-size
-// @name            Taskbar height and icon size
+// @name            Taskbar height, width and icon size
 // @description     Control the taskbar height and icon size, improve icon quality (Windows 11 only)
 // @version         1.1
 // @author          m417z
@@ -9,7 +9,7 @@
 // @homepage        https://m417z.com/
 // @include         explorer.exe
 // @architecture    x86-64
-// @compilerOptions -DWINVER=0x0605 -lole32
+// @compilerOptions -DWINVER=0x0605 -loleaut32 -lole32 -lruntimeobject
 // ==/WindhawkMod==
 
 // Source code is published under The GNU General Public License v3.0.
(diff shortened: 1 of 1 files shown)


imagicalFantasy:
I just tested it and it's work perfectly.

dr-consit:
@Gravarty Can you maybe say something about your setup (windows version)/settings/other mods?

Gravarty:
Windows 11 22H2 Moment 2 2261.1344

Dark Paint
Disable Immersive Context Menus
Middle Click to close on the taskbar
Non Immersive Taskbar Context Menu
Shrink Address Bar Height
Taskbar Clock Customization

Disabled:

Disable grouping on the taskbar
Taskbar height and icon size (Original)
Taskbar Labels for Windows 11

dr-consit:
OK, can you try disabling all other mods to see if that makes a difference, and also send the configuration settings from "Taskbar height and icon size"?

dr-consit:
@Gravarty Alright, I hadn't tested any kind of size reduction, only size increase.
I have added some settings, to allow you to adjust some settings on the start button specifically.
If you shrink the icon size, you can now set the start button alignment to "left" and reduce the padding. This should help with your issue.

Gravarty:
Oh nice, thanks! It's a bit better now, but changing the padding value, the button continues to move to the right instead of the left. Maybe that's just a hard coded restriction in the taskbar? I know it's unusual to have such a small aligment these days, I still have a 1080p monitor where modern elements like the taskbar are just too big :D

Screenshot_2

dr-consit:
Alrighty then, I have added a separate width setting for the start button.

Gravarty:
Fantastic, that's much better, thanks! :D There's still a tiny problem where the highlight frame is still overlapping and not centered with the button.. but still much better than the default gigantic taskbar and icon spacing.. Thanks!

m417z:
Thank you for the pull request @dr-consit. I'm not sure about having this functionality as part of the icon size mod, both because of the code duplication from the taskbar labels mod, and because of the potential incompatibility it introduces.

I made some basic improvements in the taskbar labels mod which allow to set a small item width and it will be applied to pinned items as well. I wrote several other thoughts about it here: https://github.com/ramensoftware/windhawk-mods/issues/45#issuecomment-1492958617

imagicalFantasy:
> Alrighty then, I have added a separate width setting for the start button.

Thank for start button settings ! It's awesome !

m417z:
I added an option to control the width of taskbar buttons in version 1.2 of the mod. The way I implemented it is different than what this pull request does - a small change which hooks the function that's used to retrieve the width. Since the change is small and simple, it's likely to be more robust and less likely to break with updates. A downside of this approach is that explorer has to be restarted to apply this option, which isn't ideal, but not too bad since it usually has to be done only during the first configuration.

I'm closing this pull request, since the feature is implemented, and it's not going to be merged in its current form in any case. If anything else is missing, let me know.

PR #237, opened 2023-06-07
Add Multi-instance regedit

learn-more:


m417z:
Nice. There's also the `-m` flag which creates a new regedit instance.

PR #261, opened 2023-07-23
DWM Ghost Mods 1.1

ephemeralViolette:


mods/dwm-ghost-mods.wh.cpp
```diff
@@ -0,0 +1,499 @@
...
+  sending inputs to it a few times.
+
+# ⚠ Important usage note ⚠
+
+In order to use this mod, you must enable Windhawk to inject into system processes in its advanced settings.
+If you do not do this, it will silently fail to inject. **Changing the Windhawk advanced settings will also
+affect any other mod you have installed, and may cause instability as any other mod that injects into all
+processes will now inject into system processes too.**
+
+This also means that the mod will not work with the portable version of Windhawk.
```
m417z: Have you tried that? It might work just by running it as administrator.

mods/dwm-ghost-mods.wh.cpp
```diff
@@ -0,0 +1,499 @@
...
+    #ifdef _WIN64
+    const int offset_same_teb_flags = 0x17EE;
+    #define NT_CURRENT_TEB_T __int64
+    #else
+    const int offset_same_teb_flags = 0x0FCA;
+    #define NT_CURRENT_TEB_T long
+    #endif
+
+    // 0x0400 = InitialThread
+    bool already_running = (!*(USHORT *)(((NT_CURRENT_TEB_T *)NtCurrentTeb()) + offset_same_teb_flags) & 0x0400);
```
m417z: The compiler shows a warning on this line:

> Logical not is only applied to the left hand side of this bitwise operator (fixes available)

As it's written, it's likely to always return false:

```cpp
(!*(USHORT *)(((NT_CURRENT_TEB_T *)NtCurrentTeb()) + offset_same_teb_flags) & 0x0400);
(!*(USHORT *)((x1) + offset_same_teb_flags) & 0x0400);
(!*(USHORT *)(x2) & 0x0400);
(!*(USHORT *)(x3) & 0x0400);
(!x4 & 0x0400);
(1 & 0x0400);
0;
```

Is it even necessary outside of testing?

ephemeralViolette: Removed. I added it as a safeguard when I started development and never removed it; didn't even notice the typo.

mods/dwm-ghost-mods.wh.cpp
ephemeralViolette: I haven't managed to get it working with the portable version at all. DWM is a protected process that you can only modify using a system account such as `NT AUTHORITY\SYSTEM`, which the standard version runs under since it uses a service.

Furthermore, while I don't suspect many people are manually running the portable version as SYSTEM, I tried that and still couldn't get it working at all.

I changed this paragraph to elaborate in the latest commit.

m417z:
Looks good. Congratulations on your first mod 🙂

mods/dwm-ghost-mods.wh.cpp
m417z: FYI it did just work for me with the portable version by running it as administrator. So your paragraph, which [was also copied to other mods as well](https://github.com/search?q=repo%3Aramensoftware%2Fwindhawk-mods+%22will+not+work+on+portable%22&type=code), doesn't seem to be correct. I only tested with Windhawk v1.4.1, the latest version at this point.

image

PR #283, opened 2023-09-02
Message Box Font Fix 1.1: Better loading of user32.dll

aubymori:


m417z:
I don't think it changes anything. Previously the mod was just loading user32.dll, now it tries to find an already loaded instance, and if it doesn't find it, it loads it. Both are pretty much the same except for the module's reference count which doesn't matter that much in this case.

aubymori:
Ah.

PR #285, opened 2023-09-02
Temporarily limit Message Box Font Fix to 64-bit applications

aubymori:
It relies on the Load Symbols Self Fix, and should be re-added once a Windhawk version with that fix integrated releases.


m417z:
Mods aren't designed to be removed, and I prefer not to introduce the option at this point. Users who already installed it may be surprised that they have a mod that isn't listed. For now, let's do one of the following, whatever you prefer:
* Limit it to 64-bit
* Remove all process inclusions altogether and update the readme
* Integrate the workaround

mods/msg-box-font-fix.wh.cpp
```diff
@@ -57,7 +61,12 @@ BOOL Wh_ModInit()
 {
     Wh_Log(L"Initializing Message Box Font Fix");
 
-    HMODULE hUser32 = LoadLibraryW(L"user32.dll");
+    HMODULE hUser32 = GetModuleHandleW(L"user32.dll");
+
+    if (!hUser32)
+    {
+        hUser32 = LoadLibraryW(L"user32.dll");
+    }
```
m417z: Somehow this code returned, probably by mistake.

PR #286, opened 2023-09-02
Center titlebar 1.0

rounk-ctrl:
wincentertitle's port

### DIFF
--- a/mods/center-titlebar.wh.cpp
+++ b/mods/center-titlebar.wh.cpp
@@ -12,7 +12,22 @@
 /*
 # Center Titlebar
 A port of valinet's [WinCenterTitle](https://github.com/valinet/WinCenterTitle).
+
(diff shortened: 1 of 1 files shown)


m417z:
How about adding a message similar to this:
https://github.com/ramensoftware/windhawk-mods/blob/f4942e9c7cba28ea5fa515307e4dd843969a7902/mods/dwm-ghost-mods.wh.cpp#L30-L40

Also, how about screenshots? You can quickly upload them to Imgur.

rounk-ctrl:
isnt the title self explanatory?

m417z:
As a user, I'd like to see a screenshot. It's self-explanatory regarding what it does, not regarding how the result actually looks.

But I don't mind, it's only a suggestion, let me know when it's ready and I'll merge it.

rounk-ctrl:
yeah done

PR #294, opened 2023-09-11
Classic Desktop Icons 1.0.0

aubymori:
Enables the classic selection style on desktop icons.

### DIFF
--- a/mods/classic-desktop-icons.wh.cpp
+++ b/mods/classic-desktop-icons.wh.cpp
@@ -1,0 +1,146 @@
+// ==WindhawkMod==
+// @id              classic-desktop-icons
+// @name            Classic Desktop Icons
+// @description     Enables the classic selection style on desktop icons.
+// @version         1.0.0
+// @author          aubymori
+// @github          https://github.com/aubymori
+// @include         explorer.exe
+// @compilerOptions -luser32 -luxtheme -lcomctl32
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Classic Desktop Icons
+
+This mod will enable classic theme on desktop icons when themes are enabled,
+and will optionally let you restore the Windows 2000 behavior of rendering
+the desktop color on the items' labels.
+
+## For label backgrounds to show, you must DISABLE desktop icon label shadows.
+
+## Screenshots
+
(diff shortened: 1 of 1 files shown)


mods/classic-desktop-icons.wh.cpp
```diff
@@ -0,0 +1,167 @@
...
+{
+    Wh_Log(L"Initializing Classic Desktop Icons");
+    LoadSettings();
+
+    // Initially update the desktop.
+    hDesktop = FindDesktopWindow();
+    UpdateDesktop();
+
+    OriginalWndProc = (WNDPROC)GetWindowLongPtr(hDesktop, GWLP_WNDPROC);
+    SetWindowLongPtr(hDesktop, GWLP_WNDPROC, (LONG_PTR)NewWndProc);
```
m417z: Using `SetWindowLongPtr` with `GWLP_WNDPROC` will make it incompatible with other mods that subclass the same window. You can include `<windhawk_utils.h>` and use:

* `WindhawkUtils::SetWindowSubclassFromAnyThread` for subclassing
* `WindhawkUtils::RemoveWindowSubclassFromAnyThread` for unsubclassing (in `Wh_ModUninit`)
* Change:\
`LRESULT CALLBACK NewWndProc(HWND hWnd, UINT uMsg, WPARAM wParam, LPARAM lParam)`\
to:\
`LRESULT CALLBACK NewWndProc(HWND hWnd, UINT uMsg, WPARAM wParam, LPARAM lParam, UINT_PTR uIdSubclass, DWORD_PTR dwRefData)`
* Change:\
`return OriginalWndProc(hWnd, uMsg, wParam, lParam);`\
to:\
`return DefSubclassProc(hWnd, uMsg, wParam, lParam);`
* Remove `OriginalWndProc`

m417z:
Other small comments:

* `YcSimpleFunctionHook` isn't used and can be removed
* In `Wh_ModSettingsChanged`, no need to call `LoadSettings` if you set `*bReload = TRUE`, the mod will be reloaded anyway

mods/classic-desktop-icons.wh.cpp
aubymori: Should be fixed now.

PR #297, opened 2023-09-13
Acrylic Effect Radius Changer 1.1.0

Dulappy:
Added an extra option for optimization in order to better mimic Windows 10's default behavior, as well as make the blur more usable at extremely high radii, or on lower-end devices.


Dulappy:
Please excuse the commit history, I haven't used GitHub very much.

mods/acrylic-effect-radius-changer.wh.cpp
```diff
@@ -51,6 +51,9 @@ You can scroll down to see some examples of different custom blur radii in actio
   - height: 30
   $name: Blur radius
   $description: Set the radius of the Acrylic blur effect to any number
+- optimization: true
+  $name: Optimizization
+  $description: Optmizes the blur, especially for higher radii, in exchange for graphical fidelity (virtually unnoticeable)
```
m417z: Typo: Optmizes :)

BTW it doesn't really matter, but I like to bump the version only for user-facing releases. In this case, the first merged version could be 1.0.0 and this version could be 1.0.1. A matter of preference, just bringing it up in case you thought it's mandatory to bump the version on each commit.

Dulappy: No worries, I know it's not mandatory to bump the version number, I just prefer doing it that way. This time I won't bump it as the changes are only a single character being added. I chose to bump the middle number instead of the last one, as this update adds a feature, instead of revising the backend, like previous updates did.

I think versioning is something people will never be able to fully agree on, lol.

PR #308, opened 2023-09-19
Classic Desktop Icons 1.1.0: Fix for initial Explorer launch

aubymori:


mods/classic-desktop-icons.wh.cpp
```diff
@@ -105,6 +119,61 @@ HWND FindDesktopWindow(void)
...
+    DWORD     dwStyle,
+    int       X,
+    int       Y,
+    int       nWidth,
+    int       nHeight,
+    HWND      hWndParent,
+    HMENU     hMenu,
+    HINSTANCE hInstance,
+    LPVOID    lpParam
+)
```
m417z, line 139: You're missing a calling convention (WINAPI), which will likely cause a crash in 32-bit. See existing mods that hook `CreateWindowExW`.

mods/classic-desktop-icons.wh.cpp
```diff
@@ -105,6 +119,61 @@ HWND FindDesktopWindow(void)
...
+    /**
+      * Check that the following criteria is met:
+      * - The window has a parent
+      * - lpClassName is non-null and not a bad pointer (MANY windows pass a bad pointer to lpClassName)
+      * - The window's class name is "SysListView32"
+      * - The window's parent's class name is "SHELLDLL_DefView"
+      */
+    if (hWndParent != NULL
+    && lpClassName != NULL
+    && !IsBadStringPtrW(lpClassName, 256))
```
m417z: Don't use `IsBadXxxPtr`.
https://devblogs.microsoft.com/oldnewthing/20060927-07/?p=29563
See how it's done in other mods that hook `CreateWindowExW` and compare the class name.

For example:
https://github.com/ramensoftware/windhawk-mods/blob/18081443a8c4f967d07c82997d3096f3b9faba6c/mods/taskbar-wheel-cycle.wh.cpp#L1088-L1090

mods/classic-desktop-icons.wh.cpp
```diff
@@ -105,6 +119,61 @@ HWND FindDesktopWindow(void)
...
+    {
+        if (0 == wcscmp(L"SysListView32", lpClassName))
+        {
+            LPWSTR lpPrntCls = (LPWSTR)malloc(sizeof(WCHAR) * 256);
```
m417z: This works, but why not just do
`WCHAR lpPrntCls[256];`

aubymori:
Should all be fixed now.

mods/classic-desktop-icons.wh.cpp
m417z: `TextualClassName` only makes sense for the argument (i.e. the first usage).

aubymori:
Okay, *now* it should be all good.

PR #316, opened 2023-09-24
Themed Regedit ListView 1.0.0

aubymori:
For a brief time during Windows 10's life, the Registry Editor had a themed ListView. This mod restores that.

**Before**:

Before

**After**:

After


m417z:
Nice!

PR #338, opened 2023-10-06
Fix Classic Theme Maximized Windows 1.2: Fix non-fully maximized windows

aubymori:
Namely open/save dialogs, which maximize but do not fill screen.


m417z:
As the automatic verification says, you cannot submit an update to a mod you didn't author. Only @ephemeralViolette can submit updates to it.

ephemeralViolette:
This update is okay to submit.

PR #344, opened 2023-10-12
Add Fix Darkmode ListViews mod

Reabstraction:
This is a quick mod I wrote for use with Rectify11 that fixes SysListView32's

### DIFF
--- a/mods/fix-darkmode-listviews.wh.cpp
+++ b/mods/fix-darkmode-listviews.wh.cpp
@@ -1,0 +1,157 @@
+// ==WindhawkMod==
+// @id              fix-darkmode-listviews
+// @name            Fix Darkmode ListViews
+// @description     Fixes ListViews in dark mode
+// @version         1.0
+// @author          Reabstraction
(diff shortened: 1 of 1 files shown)


m417z:
How about adding a short readme that explains in more detail what the mod does, preferably with screenshots?

Reabstraction:
I'll have to set up a Windows 11 VM to do that
Which I'll do later this day after work

Reabstraction:
@m417z I have added a basic description

Reabstraction:
It's basic but I'm no writer

WildByDesign:
@Reabstraction Great work.

If you don’t mind me asking, does this only affect Regedit (from your screenshot) or does this also affect other programs as well?

If other programs as well, could you let me know a few of them as examples?

Thank you. I am also a Rectify11 user so I’m just trying to get an idea of where all this may benefit my setup.

Reabstraction:
It affects all programs @WildByDesign

Not many apps seem to use SysListView32 though

WildByDesign:
Thank you

PR #356, opened 2023-11-05
Explorer 32px Icons 1.0.0

aubymori:
Beginning with Windows Vista, the "Medium icons" view (previous just "Icons") uses icons that are 48 pixels in size, instead of the previous 32 pixel size. This mod restores the 32 pixel size.

## YOU MUST BE USING AT LEAST WINDHAWK 1.4, OR ELSE EVERY 32-BIT APP WILL CRASH!

# IMPORTANT
You **must** be using Explorer with the SysListView32 view enabled for this mod to work. The default DirectUI view will not work.

**Before**:

Before

**After**:

After

### DIFF
--- a/mods/explorer-32px-icons.wh.cpp
+++ b/mods/explorer-32px-icons.wh.cpp
@@ -15,7 +15,7 @@ Beginning with Windows Vista, the "Medium icons" view (previous just "Icons") us
 icons that are 48 pixels in size, instead of the previous 32 pixel size. This mod
 restores the 32 pixel size.
 
-## YOU MUST BE USING AT LEAST WINDHAWK 1.4, OR ELSE EVERY 32-BIT APP WILL CRASH!
+**This mod will only work on Windhawk v1.4 and greater.**
 
 # IMPORTANT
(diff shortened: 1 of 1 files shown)


mods/explorer-32px-icons.wh.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+    WindhawkUtils::SYMBOL_HOOK hooks[] = {
+        {
+            {
+                L"private: void "
+                SCALCON
+                L" CDefView::_SwitchToViewModeAndIconSize(enum FOLDERVIEWMODE,int)"
+            },
+            (void **)&CDefView__SwitchToViewModeAndIconSize_orig,
+            (void *)CDefView__SwitchToViewModeAndIconSize_hook,
```
m417z: In version 1.4, you can remove these casts (both `(void **)` and `(void *)`), in which case the compiler will make sure that the hook and the original prototypes match. In addition to the added safety, not compiling in an older Windhawk version is a good side effect as the alternative is crashing all 32-bit processes (you could also change the warning to be less scary).

PR #358, opened 2023-11-05
Eradicate Immersive Menus 1.0.0

aubymori:
Gets rid of immersive menus system-wide. This mod should work on every Windows 10 version, from 1507 all the way up to 22H2. However, **this mod will only work on Windhawk v1.4 and greater.**

## Preview

*From left to right: Explorer, Taskbar, Explorer (classic), Taskbar (classic)*

**Before**:

Before

**After**:
After

### DIFF
--- a/mods/eradicate-immersive-menus.wh.cpp
+++ b/mods/eradicate-immersive-menus.wh.cpp
@@ -1,0 +1,353 @@
+// ==WindhawkMod==
+// @id              eradicate-immersive-menus
+// @name            Eradicate Immersive Menus
+// @description     Gets rid of immersive menus system-wide
+// @version         1.0.0
+// @author          aubymori
+// @github          https://github.com/aubymori
+// @include         *
+// @exclude         windhawk.exe
+// @exclude         dwm.exe
+// @exclude         ShellExperienceHost.exe
+// @exclude         wlanext.exe
+// @exclude         svchost.exe
+// @exclude         SearchIndexer.exe
+// @exclude         consent.exe
+// @compilerOptions -luser32
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Eradicate Immersive Menus
+Gets rid of immersive menus system-wide. This mod should work on every Windows 10 version, from 1507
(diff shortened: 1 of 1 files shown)


mods/eradicate-immersive-menus.wh.cpp
```diff
@@ -0,0 +1,353 @@
...
+    }
+#endif
+
+    return TRUE;
+}
+
+BOOL Wh_ModSettingsChanged(BOOL *bReload)
+{
+    *bReload = TRUE;
+    LoadSettings();
```
m417z: Doesn't matter all that much, but just FYI - this isn't necessary since the mod will be reloaded right away.

mods/eradicate-immersive-menus.wh.cpp
```diff
@@ -0,0 +1,353 @@
...
+restart Windhawk. If that does not work, simply add it to the exclusion list in the Advanced tab of this mod.
+
+## Failed to download symbols:
+
+Sometimes, Windhawk may fail to download the symbols for certain system files. This doesn't necessarily mean that your version
+of Windows is incompatible, but it means that whatever application had the mod hooked cannot connect to the internet (which is how Windhawk
+attempts to download symbols).
+
+You can work around this issue by manually downloading the symbols and dropping them in the Windhawk symbols
+folder. In order to do this, use a tool such as [PDB Downloader](https://github.com/rajkumar-rangaraj/PDB-Downloader/releases)
```
m417z: I'm not familiar with this tool, but symbols can also be downloaded with [Windhawk Symbol Helper](https://github.com/ramensoftware/windhawk-symbol-helper), which might or might not be simpler.
Windhawk Symbol Helper detects the Windhawk configuration such as the symbols directory automatically, so all there's to do is to input the dll path and click "Get symbols". For the portable version, the detection works if the tool is ran from Windhawk's folder.

m417z (commented):
Both comments are FYI, so LMW if you just want me to merge it as is.

aubymori:
Alright, you can merge now.

PR #360, opened 2023-11-05
Custom Shutdown Dialog 1.0.0

aubymori:
Override the classic shutdown dialog in Explorer which is invoked with `ALT`+`F4` with your own program.

### DIFF
--- a/mods/custom-shutdown-dialog.wh.cpp
+++ b/mods/custom-shutdown-dialog.wh.cpp
@@ -1,0 +1,107 @@
+// ==WindhawkMod==
+// @id              custom-shutdown-dialog
+// @name            Custom Shutdown Dialog
+// @description     Override the classic shutdown dialog in Explorer with your own
+// @version         1.0.0
+// @author          aubymori
+// @github          https://github.com/aubymori
+// @include         explorer.exe
+// @architecture    x86-64
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Custom Shutdown Dialog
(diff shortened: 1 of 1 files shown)


mods/custom-shutdown-dialog.wh.cpp
```diff
@@ -0,0 +1,107 @@
...
+        Wh_Log(L"Failed to hook _ShutdownDialogEx");
+        return FALSE;
+    }
+
+    return TRUE;
+}
+
+void Wh_ModSettingsChanged(void)
+{
+    LoadSettings();
```
m417z, line 100: Old string values aren't freed in this case. For easy management, you can use the new `StringSetting` from `windhawk_utils.h`. Check out the comments in the header file.

aubymori:
Should be good to go now.

m417z:
Why not `WindhawkUtils::StringSetting`? It was created just for these cases and I want to know if anything's wrong with it.
It makes the `Wh_FreeStringSetting` calls unnecessary, and prevents memory leaks in other cases, such as returning FALSE on line 78 or 92 after settings were loaded.

aubymori:
> Why not `WindhawkUtils::StringSetting`? It was created just for these cases and I want to know if anything's wrong with it. It makes the `Wh_FreeStringSetting` calls unnecessary, and prevents memory leaks in other cases, such as returning FALSE on line 78 or 92 after settings were loaded.

Ah, sorry, I was having a bit of confusion with it, I was attempting to call the constructor directly.

PR #379, opened 2023-11-20
UIRibbon Caption Icon Fix 1.0

YukisCoffee:


m417z:
* Calling conventions are missing, which is likely to cause a crash for 32-bit targets.
* How about adding a readme with a short explanation and screenshots?

YukisCoffee:
Fixed

PR #419, opened 2023-12-10
New mod Toggle Taskbar Autohide

m1lhaus:
This new mod adds missing feature from 7+ Taskbar Tweaker. Middle clicking on empty taskbar space toggles taskbar autohide feature.

The mod hooks into taskbar event proc callback using code ripped of from [taskbar-volume-control.wh.cpp](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/taskbar-volume-control.wh.cpp). As far as I can tell it is a magic for me, but it works. Whenever a mouse middle click event gets caught, method OnMouseClick() get called. There, using UIAutomation API a name of the element clicked is retrieved. If UIAutomation API returned a taskbar as an element, it means user clicked an empty space on a taskbar. If however other element is returned (start menu, search, etc.), the event is ignored and can be then processed by some other hook.

The reason why I used UIAutomation API is that it reliably returns an identifier of element clicked across Windows version. I am pretty sure this approach will also work for future Windows versions with just small adjustments (e.g. className). Although those are COM calls, since we are dealing here with a mouse click events, I can't see any drawback in the overhead (if there is any).

I tested the mod with Windows 10 22H2, Windows 11 22H1, 22H2, 23H2. I also tested Windows 11 23H2 with Explorer patcher. I tried both 100% DPI and 200% DPI (on WIndows 11 only). I tested clicking both primary and secondary taskbars.

This is my first mod and also the first time ever I did something with Win32 API. So if you see any way for improvement, please let me know.

### DIFF
--- a/mods/taskbar-empty-space-clicks.wh.cpp
+++ b/mods/taskbar-empty-space-clicks.wh.cpp
@@ -1,6 +1,6 @@
 // ==WindhawkMod==
 // @id              taskbar-empty-space-clicks
-// @name            Mouseclick empty taskbar space
+// @name            Click on empty taskbar space
 // @description     Trigger custom action when empty space on a taskbar is double/middle clicked.
 // @version         0.1
 // @author          m1lhaus
@@ -19,9 +19,23 @@
 
 // ==WindhawkModReadme==
 /*
-# Toggle taskbar autohide
-Toggle taskbar autohide feature with middle mouse click on the taskbar. Whenever you middle click on a free space on your taskbar (outside other UI
-elements on the taskbar), taskbar's autohide feature gets turned on or off.
+# Click on empty taskbar space
+
+This mod lets you assign an action to a mouse click on Windows taskbar. Double-click and middle-click actions are supported. This mod only modifies behaviour when empty space of the taskbar is clicked. Buttons, menus or other function of the taskbar are not affected. Both primary and secondary taskbars are supported.
+
+## Supported actions:
+
+1. **Show desktop** - Toggle show/hide desktop
+2. **Ctrl+Alt+Tab** - Opens Ctrl+Alt+Tab dialog
+3. **Task Manager** - Opens Windows default task manager
+4. **Mute system volume** - Toggle mute of system volume (all sound)
+5. **Taskbar auto-hide** - Toggle Windows taskbar auto-hide feature 
+6. **Win+Tab** - Opens Win+Tab dialog
+7. **Hide desktop icons** - Toggle show/hide of all desktop icons
+
+## Example
+
+Following animation shows **Taskbar auto-hide** feature. Feature gets toggled whenever user double-clicks the empty space on a taskbar.
 
 ![Demonstration of Toggle taskbar autohide mod for Windhawk](https://i.imgur.com/BRQrVnX.gif)
 
@@ -45,9 +59,10 @@ In case you are using old Windows taskbar on Windows 11 (**Explorer Patcher** or
   - ACTION_SHOW_DESKTOP: Show desktop
   - ACTION_ALT_TAB: Ctrl+Alt+Tab
   - ACTION_TASK_MANAGER: Task Manager
-  - ACTION_MUTE: Toggle mute system volume
-  - ACTION_TASKBAR_AUTOHIDE: Toggle taskbar auto-hide
-  - ACTION_WIN_TAB: Win+Tab / Ctrl+Win+Tab
+  - ACTION_MUTE: Mute system volume
+  - ACTION_TASKBAR_AUTOHIDE: Taskbar auto-hide
+  - ACTION_WIN_TAB: Win+Tab
+  - ACTION_HIDE_ICONS: Hide desktop icons 
 - middleClickAction: nothing
   $name: Middle click on empty space
   $options:
@@ -55,9 +70,10 @@ In case you are using old Windows taskbar on Windows 11 (**Explorer Patcher** or
   - ACTION_SHOW_DESKTOP: Show desktop
   - ACTION_ALT_TAB: Ctrl+Alt+Tab
   - ACTION_TASK_MANAGER: Task Manager
-  - ACTION_MUTE: Toggle mute system volume
-  - ACTION_TASKBAR_AUTOHIDE: Toggle taskbar auto-hide
-  - ACTION_WIN_TAB: Win+Tab / Ctrl+Win+Tab
+  - ACTION_MUTE: Mute system volume
+  - ACTION_TASKBAR_AUTOHIDE: Taskbar auto-hide
+  - ACTION_WIN_TAB: Win+Tab
+  - ACTION_HIDE_ICONS: Hide desktop icons 
 - oldTaskbarOnWin11: false
   $name: Use the old taskbar on Windows 11
   $description: >-
@@ -553,7 +569,8 @@ enum TaskBarAction {
     ACTION_TASK_MANAGER,
     ACTION_MUTE,
     ACTION_TASKBAR_AUTOHIDE,
-    ACTION_WIN_TAB
+    ACTION_WIN_TAB,
+    ACTION_HIDE_ICONS
 };
 
 static struct {
@@ -626,6 +643,7 @@ static IMMDeviceEnumerator *g_pDeviceEnumerator;
 static HWND g_hTaskbarWnd;
 static std::unordered_set<HWND> g_secondaryTaskbarWindows;
 static UIAutomationWrapper g_UIAutomation;
+static HWND g_hDesktopWnd;
 
 // TODO update me
 
@@ -1306,6 +1324,8 @@ TaskBarAction ParseMouseActionSetting(const wchar_t *option) {
         action = ACTION_TASKBAR_AUTOHIDE;
     } else if (equals(value, L"ACTION_WIN_TAB")) {
         action = ACTION_WIN_TAB;
+    } else if (equals(value, L"ACTION_HIDE_ICONS")) {
+        action = ACTION_HIDE_ICONS;
     } else {
         Wh_Log(L"Error: unknown action '%s' for option '%s'!", value, option);
         action = ACTION_NOTHING;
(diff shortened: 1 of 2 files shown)


m417z:
Nice. There's some dead code that you can remove, such as (from a short glance): `TaskBarVersionStr`, `SYMBOL_HOOK`, `HookSymbols`, `GetTaskbarViewDllPath`, `case WM_COPYDATA: [...]`.

m417z:
BTW looks like the relevant UI Automation definitions were recently added to Wine (and MinGW-w64). I'll look into bundling them with Windhawk's next version.
https://gitlab.winehq.org/wine/wine/-/merge_requests/569

m1lhaus:
I removed mentioned dead code. Yeah it would be great to have those UIAutomation bundled since now it makes to code unnecessarily long.

mods/toggle-taskbar-autohide.wh.cpp
```diff
@@ -0,0 +1,1108 @@
...
+    // From that we can't really tell reliably whether user clicked on the taskbar empty space or on some UI element on that taskbar, like 
+    // opened window, icon, start menu, etc. 
+
+    IUIAutomation *pUIAutomation = g_UIAutomation.getInstance();
+    if (pUIAutomation == NULL) {
+        Wh_Log(L"UIAutomationWrapper failed to get instance\n");
+        return false;
+    }
+
+    IUIAutomationElement *pWindowElement = NULL;
```
m417z: FYI you can use COM smart pointers to avoid having to call `Release`. Example:
https://github.com/ramensoftware/windhawk-mods/blob/90d6ccd363cb9764f081ea1981932b2b3dac68cf/mods/taskbar-grouping.wh.cpp#L726-L729

mods/toggle-taskbar-autohide.wh.cpp
```diff
@@ -0,0 +1,1108 @@
...
+    HRESULT hr = pUIAutomation->ElementFromPoint(pointerLocation, &pWindowElement);
+    if (FAILED(hr) || (pWindowElement == NULL)) {
+        Wh_Log(L"Failed to retrieve UI element from mouse click\n");
+        return false;
+    }
+
+    BSTR className = NULL;
```
m417z: You can use the `_bstr_t` smart pointer here.

mods/toggle-taskbar-autohide.wh.cpp
```diff
@@ -0,0 +1,1108 @@
...
+    Wh_Log(L"Middle mouse clicked at x=%ld, y=%ld", pointerLocation.x, pointerLocation.y);
+
+    // Note: The reason why UIAutomation interface is used is that it reliably returns a className of the element clicked.
+    // If standard Windows API is used, the className returned is always Shell_TrayWnd which is a parrent window wrapping the taskbar.
+    // From that we can't really tell reliably whether user clicked on the taskbar empty space or on some UI element on that taskbar, like 
+    // opened window, icon, start menu, etc. 
+
+    IUIAutomation *pUIAutomation = g_UIAutomation.getInstance();
+    if (pUIAutomation == NULL) {
+        Wh_Log(L"UIAutomationWrapper failed to get instance\n");
```
m417z: There's no need to end log lines with a newline.

mods/toggle-taskbar-autohide.wh.cpp
```diff
@@ -0,0 +1,1108 @@
...
+    WNDCLASS wndclass;
+    if (GetClassInfo(GetModuleHandle(NULL), L"Shell_TrayWnd", &wndclass)) {
+        HWND hWnd = FindCurrentProcessTaskbarWindows(&g_secondaryTaskbarWindows);
+        if (hWnd) {
+            HandleIdentifiedTaskbarWindow(hWnd);
+        }
+    }
+
+    if (!g_UIAutomation.init()) {
+        Wh_Log(L"UIAutomationWrapper failed to initialize\n");
```
m417z: It should be before the subclass code, otherwise, the subclass won't be unset and Explorer will crash.

m417z (commented):
Looks good overall, well done. I left a couple of non-critical comments, let me know if you want to address them, up to you. I can also merge it as is.

I had a similar mod on my list. I wanted to create a replacement to the 7+ Taskbar Tweaker options, which allowed to configure actions for various click events on the taskbar. The events could be e.g. middle click, double click, Ctrl+Alt+click, etc. The actions could be e.g. Show desktop, open start menu, trigger Alt+Tab, run a program, and more. You made the  core, difficult part of it, would you like to extend it and make it customizable? Adding options is fairly easy in Windhawk, I can help and show relevant mod examples.

m1lhaus:
I reworked all the findings. Thx for the hint with smart pointers. Man, Microsoft docu is a maze. 

Regarding you point with call order issue in init method. Isn't Wh_ModUninit called every time no matter the result of Wh_ModInit? That would be my expectations.

Regarding old 7+Taskbar functionality, I implemented just the toggle since hiding taskbar is the closest I can get to fullscreen experience on Windows. I didn't really think about the rest. Do you have a code I could re-use? I might take a look into it. If I decide to continue with other features as well, I guess the best would be to rename a mod and to start incrementally implementing those features, right?

m1lhaus:
OK, I will look into that and let you know. I'll take down the PR if I decide to give it a shot. Holidays are coming so there should  be some free time for that.

m417z:
Another action that you can add that was requested by a user is: Toggle hide desktop icons.
Here's an implementation you can use:
https://gist.github.com/m417z/2ef1273f7998b458c4d4c1002c8f3586

m1lhaus:
Thx, implemented. I will finish it soon.

m1lhaus:
I implemented most of the stuff from 7+Taskbar tweaker. I skipped Taskbar inspector and Start menu. I can't see any value in the inspector and I experienced crashes with the menu. I might take a look on Start menu feature in future. Now I would like to release what I got so far.

mods/taskbar-empty-space-clicks.wh.cpp
```diff
@@ -0,0 +1,1572 @@
+// ==WindhawkMod==
+// @id              taskbar-empty-space-clicks
+// @name            Click on empty taskbar space
+// @description     Trigger custom action when empty space on a taskbar is double/middle clicked.
```
m417z: Descriptions don't usually end with a period.

mods/taskbar-empty-space-clicks.wh.cpp
```diff
@@ -0,0 +1,1572 @@
...
+
+![Demonstration of Toggle taskbar autohide mod for Windhawk](https://i.imgur.com/BRQrVnX.gif)
+
+## Supported Windows versions are:
+- Windows 10 22H2 (prior versions are not tested, but should work as well)
+- Windows 11 21H2 - latest major
+
+I will not supporting Insider preview or other minor versions of Windows. However, feel free to [report any issues](https://github.com/m1lhaus/windhawk-mods/issues) related to those versions. I'll appreciate the heads-up in advance.  
+
+In case you are using old Windows taskbar on Windows 11 (**Explorer Patcher** or a similar tool), enable corresponding option on Settings menu. This options will be tested only with the latest major version of Windows 11 (e.g. 23H2).
```
m417z: I recently learned that the precise name is "ExplorerPatcher".

mods/taskbar-empty-space-clicks.wh.cpp
```diff
@@ -0,0 +1,1572 @@
...
+    LONG_PTR lpImmersiveWindowMessageService =
+        *(LONG_PTR *)(lpImmersiveShellHookService + DO10_3264(0, 0, , , , , , , 0x88, 0xE0, , , , , , , 0xA0, 0x108, 0x98, 0x100));
+    if (!lpImmersiveWindowMessageService)
+        return NULL;
+
+    return *(HWND *)(lpImmersiveWindowMessageService + DO10_3264(0, 0, , , , , , , 0x4C, 0x80, , , 0x50, 0x88, , , 0x58, 0x98, 0x54, 0x90));
+}
+
+TaskBarAction ParseMouseActionSetting(const wchar_t *option) {
+    const auto value = Wh_GetStringSetting(option);
```
m417z, line 1276: The value is never freed.

mods/taskbar-empty-space-clicks.wh.cpp
```diff
@@ -0,0 +1,1572 @@
...
+    } else {
+        Wh_Log(L"UIAutomation COM initilized");
+    }
+
+    // init COM interface for Volume control
+    const GUID XIID_IMMDeviceEnumerator = {0xA95664D2, 0x9614, 0x4F35, {0xA7, 0x46, 0xDE, 0x8D, 0xB6, 0x36, 0x17, 0xE6}};
+    const GUID XIID_MMDeviceEnumerator = {0xBCDE0395, 0xE52F, 0x467C, {0x8E, 0x3D, 0xC4, 0x57, 0x92, 0x91, 0x69, 0x2E}};
+    if (FAILED(CoCreateInstance(XIID_MMDeviceEnumerator, NULL, CLSCTX_INPROC_SERVER, XIID_IMMDeviceEnumerator, g_pDeviceEnumerator.put_void())) || !g_pDeviceEnumerator) {
+        Wh_Log(L"Failed to create DeviceEnumerator COM instance, ModInit failed");
+        return FALSE;
```
m417z, line 1555: `CoUninitialize` is missing in this case.

mods/taskbar-empty-space-clicks.wh.cpp
```diff
@@ -0,0 +1,1572 @@
...
+}
+
+bool FindDesktopWindow() {
+    HWND hParentWnd = FindWindow(L"Progman", NULL);     // Program Manager window
+	if (!hParentWnd) {
+        Wh_Log(L"Failed to find Progman window");
+        return false;
+    }
+
+	HWND hChildWnd = FindWindowEx(hParentWnd, NULL, L"SHELLDLL_DefView", NULL);     // parent window of the desktop
```
m417z: Some tabs are lurking here.

mods/taskbar-empty-space-clicks.wh.cpp
```diff
@@ -0,0 +1,1572 @@
...
+I will not supporting Insider preview or other minor versions of Windows. However, feel free to [report any issues](https://github.com/m1lhaus/windhawk-mods/issues) related to those versions. I'll appreciate the heads-up in advance.  
+
+In case you are using old Windows taskbar on Windows 11 (**Explorer Patcher** or a similar tool), enable corresponding option on Settings menu. This options will be tested only with the latest major version of Windows 11 (e.g. 23H2). 
+
+*/
+// ==/WindhawkModReadme==
+
+// ==WindhawkModSettings==
+/*
+- doubleClickAction: nothing
```
m417z: Replace `nothing` with `ACTION_NOTHING`, otherwise the default text just displays "nothing".

image

m417z:
Looks good overall. I left some small comments that you might want to address, nothing critical.

> I can't see any value in the inspector 

It can actually be handy for operations such as quickly closing multiple windows, and doing other operations such as changing AppIDs which affect grouping on the taskbar. But porting it isn't trivial, I didn't expect that you'd do that.

> I experienced crashes with the menu

Maybe it changed in Windows 11. I'll see if I can find a solution for it.

Also, I think it would be a good idea to add the following to have customization on par with 7+ Taskbar Tweaker:
* Allowing to run arbitrary commands, for example run Notepad/calculator on double-click.
* Adding more buttons (Button 4, button 5, and maybe right click) and modifiers (Ctrl, Alt, Shift). It can be implemented as an array of options, so that uses can add new hotkeys and define an action for each (since listing all permutations is too much). But maybe we can wait and see how many users actually want this.

mods/taskbar-empty-space-clicks.wh.cpp
m1lhaus: I implemented a wrapper that takes of Uninitialization even if Wh_ModUninit is not called.

mods/taskbar-empty-space-clicks.wh.cpp
m417z: Still there, but not that important, you can fix it in the next version if you want.

mods/taskbar-empty-space-clicks.wh.cpp
m417z: Still there, but not that important, you can fix it in the next version if you want.

m417z:
Looks good, well done! Merged.

m1lhaus:
OK, thx. I was still planning to take a look to StartMenu and to that whole keypress emulation thing as you suggested. I will take care of it in next version then.

PR #436, opened 2023-12-19
Aero Flyout Fix 1.0.0

aubymori:
Since Windows 10, many of the legacy tray flyouts either don't have thick borders or don't
float away from the taskbar. This mod adds back thick borders and makes flyouts float where
applicable.

**Before**:
Before

**After**:
After

### DIFF
--- a/mods/aero-flyout-fix.wh.cpp
+++ b/mods/aero-flyout-fix.wh.cpp
@@ -18,11 +18,6 @@ Since Windows 10, many of the legacy tray flyouts either don't have thick border
 float away from the taskbar. This mod adds back thick borders and makes flyouts float where
 applicable.
 
-# IMPORTANT: READ!
-Windhawk needs to hook into `winlogon.exe` to successfully capture Explorer starting. Please
-navigate to Windhawk's Settings, Advanced settings, More advanced settings, and make sure that
-`winlogon.exe` is in the Process inclusion list.
-
 **Before**:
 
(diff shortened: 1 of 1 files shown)


m417z:
How about dropping the requirement to be injected at startup? I didn't go over all the code, but overall it looks like you're not far away from achieving it. For example, instead of applying styles on `IsolationAwareCreateWindowExW`, use `CTrayClock_s_WndProc` which you're hooking already to apply the styles at `WM_CREATE` and on show (`WM_WINDOWPOSCHANGED` with `SWP_SHOWWINDOW`).

The benefits:
* No need to restart explorer after installing/enabling the mod.
* No need to configure Windhawk to hook into `winlogon.exe`, which isn't bulletproof either - the injection into winlogon can happen too late as well.

mods/aero-flyout-fix.wh.cpp
```diff
@@ -0,0 +1,453 @@
...
+            else
+            {
+                *set += offset - curOffset;
+            }
+        }
+    }
+    return { rc.left + dx, rc.top + dy };
+}
+
+/* == EXPLORER.EXE HOOKS == */
```
m417z: Tip: Are you familiar with `#pragma region`?
You can wrap code with something like:
```cpp
#pragma region explorer_hooks
// ...
#pragma endregion  // explorer_hooks
```
Which is a more conventional way of marking a block of code, and has the benefit of being collapsible in the editor.

aubymori:
@m417z Yeah, I forgot to remove that section. It's now removed. Everything is also in `#pragma region`s now too.

PR #446, opened 2023-12-24
Win32 Tray Clock Experience 1.0.0

aubymori:
This mod makes the clock button on the Windows 10 taskbar open the Win32 tray clock from Windows 8.1 and before instead of the XAML one from Windows 10 and up. This *should* work on Windows 11 with the legacy taskbar, but I have not tested it.

**Preview**:
Preview

### DIFF
--- a/mods/win32-tray-clock-experience.wh.cpp
+++ b/mods/win32-tray-clock-experience.wh.cpp
@@ -30,6 +30,8 @@ Based on the ExplorerPatcher implementation.
 #include <windhawk_utils.h>
 #include <winnt.h>
 
+#define STATUS_SUCCESS 0x00000000
+
 typedef NTSTATUS (NTAPI *RtlGetVersion_t)(PRTL_OSVERSIONINFOW);
 RtlGetVersion_t RtlGetVersion;
 
--- a/mods/win32-tray-clock-experience.wh.cpp
+++ b/mods/win32-tray-clock-experience.wh.cpp
@@ -1,0 +1,147 @@
+// ==WindhawkMod==
+// @id              win32-tray-clock-experience
+// @name            Win32 Tray Clock Experience
+// @description     Use the Win32 clock flyout instead of the XAML one
+// @version         1.0.0
+// @author          aubymori
+// @github          https://github.com/aubymori
+// @include         explorer.exe
+// @architecture    x86-64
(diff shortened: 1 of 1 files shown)


mods/win32-tray-clock-experience.wh.cpp
```diff
@@ -0,0 +1,147 @@
...
+        &ClockButton_v_OnClick_orig,
+        ClockButton_v_OnClick_hook,
+        false
+    }
+};
+
+BOOL Wh_ModInit(void)
+{
+    if (!WindhawkUtils::HookSymbols(
+        GetModuleHandleW(NULL),
```
m417z: For the old taskbar on Win11, you'll likely need to use `taskbar.dll`. Example:
https://github.com/ramensoftware/windhawk-mods/blob/f80c31b2a02a858d585c01b85fb3d5d5f1789555/mods/pinned-items-double-click.wh.cpp#L457-L470

Let me know if you want to add it or keep it as is, in which case it's probably better to update the description.

aubymori:
@m417z Should be good now. Tell me if there's any issues, this was all written and committed from my phone, lol.

m417z:
Impressive 🙂  but:
```
mod.wh.cpp:155:9: error: use of undeclared identifier 'STATUS_SUCCESS'
    if (STATUS_SUCCESS == RtlGetVersion(&osv))
        ^
1 error generated.
```
I'd also check the major version, one day MS might want to make it 11 or 12, who knows...

aubymori:
> Impressive 🙂 but:
> 
> ```
> mod.wh.cpp:155:9: error: use of undeclared identifier 'STATUS_SUCCESS'
>     if (STATUS_SUCCESS == RtlGetVersion(&osv))
>         ^
> 1 error generated.
> ```

The fact that `NTSTATUS` is defined but not `STATUS_SUCCESS` is really something...

> I'd also check the major version, one day MS might want to make it 11 or 12, who knows...

They didn't change it with 11 and it's not like they're going to bump the build number back to below 22000, so I don't see any problem with my checking.

m417z:
I agree that it's mainly esthetics in this case, it just makes me think about stories like this: https://devblogs.microsoft.com/oldnewthing/20040213-00/?p=40633

In any case, it compiles now, merging...

PR #447, opened 2023-12-25
classic-maximized-windows-fix 2.0

ephemeralViolette:
- Moved global window procedure hooking to leverage the availability of the theme engine (uxtheme), rather than using SetWindowsHookEx.
- Improved compatibility with many apps, such as pre-Windows 8 versions of Task Manager.

### DIFF
--- a/mods/classic-maximized-windows-fix.wh.cpp
+++ b/mods/classic-maximized-windows-fix.wh.cpp
@@ -262,6 +262,11 @@ struct CMWF_SYMBOL_HOOK {
 /*
  * CmwfHookSymbols: A custom hook wrapper which allows storing symbol hook results in the
  *                  module's shared memory for faster access.
+ *
+ * TODO: This is good for the current application, which hooks only one symbol, but it will
+ *       be less efficient to use for multiple symbol hooks at the moment because it requeries
+ *       each time. Please keep this in mind if you're looking to use this code for your own
+ *       purposes.
  */
 bool CmwfHookSymbols(
         HMODULE module,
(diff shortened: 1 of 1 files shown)


mods/classic-maximized-windows-fix.wh.cpp
```diff
@@ -242,90 +228,98 @@ bool InitCriticalSections()
...
+        {
+            address = nullptr;
+            WindhawkUtils::SYMBOL_HOOK proxyHook = {
+                symbolHooks[i].symbols,
+                &address,
+                NULL,
+                symbolHooks[i].optional
+            };
+
+            if (!WindhawkUtils::HookSymbols(module, &proxyHook, 1))
```
m417z, line 299: I don't see much value in this wrapper. Is it to have an in-memory cache instead of a registry cache? Did you measure the difference?

Also, this mod only hooks one symbol, but generally this is going to be very slow, since this wrapper handles each symbol separately, and so the symbols will be enumerated again for each symbol, and enumerating symbols is usually the slow part in symbol loading.

ephemeralViolette: Yes, that is the goal. I've noticed that application startup can still be quite slow, even with the registry cache, so I wanted to try something else out to make it nicer to use for hooking all running applications. It seems that the slowness of reading from the registry is dependent on the application in particular, but there are some for which it can be particularly noticeable (in my experience, Git and some old versions of Notepad).

I will improve the wrapper for future usage later, but I think it should be fine here since it's only used for the one symbol.

m417z: I don't mind merging it, but I doubt that the single registry read causes the slowness, Windhawk reads other values from the registry, the cache value really shouldn't matter. That's unlike the first run when the cache isn't available, for which symbols have to be downloaded and enumerated, but your wrapper doesn't improve this case.

ephemeralViolette: I elaborated on the possible inefficiency of the wrapper in the latest commit.

PR #449, opened 2023-12-26
classic-maximized-windows-fix v2.1

ephemeralViolette:
- Improved compatibility with Aero Snap. This was done by migrating to the undocumented `SetWindowRgnEx` function rather than the older public function.
- Rewrote the symbol hook wrapper CmwfHookSymbols for initial load with multiple symbol hooks.


ephemeralViolette:
There is a major bug I did not realise in testing, sorry.

PR #457, opened 2023-12-30
Classic Conhost 1.0.1: Add to README

aubymori:


m417z:
Only after merging I noticed that conhost.exe isn't included explicitly, which is necessary for Windhawk 1.4 and newer. Also, I don't quite understand why it targets all processes.

aubymori:
> Only after merging I noticed that conhost.exe isn't included explicitly, which is necessary for Windhawk 1.4 and newer. Also, I don't quite understand why it targets all processes.

All processes are include to ensure the console window actually gets themed since Windhawk can't intercept the creation of the `conhost.exe` process.

m417z:
How does it help? It's still asynchronous. Which process that the mod is injected to is faster than the asynchronous injection of the mod into conhost.exe?

PR #496, opened 2024-02-02
classic-uwp-fix v0.2

Ingan121:
This mod fixes UWP apps getting stuck at the splash screen when using the Classic theme.

### DIFF
--- a/mods/classic-uwp-fix.wh.cpp
+++ b/mods/classic-uwp-fix.wh.cpp
@@ -7,6 +7,7 @@
 // @github          https://github.com/Ingan121
 // @homepage        https://www.ingan121.com/
 // @include         ApplicationFrameHost.exe
+// @architecture    x86-64
 // ==/WindhawkMod==
 
 // ==WindhawkModReadme==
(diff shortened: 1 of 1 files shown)


mods/classic-uwp-fix.wh.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+No more need to run ApplicationFrameHost before enabling Classic!
+
+This works by hardcoding the splash fade animation values instead of fetching them from msstyles.
+*/
+// ==/WindhawkModReadme==
+
+#include <windhawk_utils.h>
+
+typedef INT64 (* AnimationHelpers__LoadThemeTransform_t)(void *, UINT, INT64 *, struct TA_TRANSFORM *, INT64 *, struct TA_TIMINGFUNCTION *);
```
m417z: Calling conventions are missing, which is likely to cause a crash for 32-bit targets. Please add calling conventions or, if the mod doesn't support 32-bit, limit it to 64-bit, which will also make compilation faster. See also:

Ingan121: I tried to fix it, but I had some problems trying to run it on a 32-bit VM so just limited it to 64-bit for now.

PR #507, opened 2024-02-06
Click on empty taskbar space v1.3

m1lhaus:
**Changelog:**

Features:
- added option to control "Combine Taskbar buttons" feature on secondary taskbars
- added feature to send custom key press
- added feature to run arbitrary command/process

**Dev:**
- implemented file logger in order to capture log after fresh Windows start (and debug hook) or to capture crashes
- implemented trace logger to log method entry/exit to debug execution order  
- implemented logging levels to make stuff really verbose if needed

### DIFF
--- a/mods/taskbar-empty-space-clicks.wh.cpp
+++ b/mods/taskbar-empty-space-clicks.wh.cpp
@@ -1617,44 +1617,47 @@ void OpenTaskManager(HWND taskbarhWnd)
     SendMessage(taskbarhWnd, WM_COMMAND, MAKELONG(420, 0), 0);
 }
 
-BOOL ToggleVolMuted()
+/**
+ * Toggles the volume mute state via Windows Core Audio API through COM interface.
+ */
+void ToggleVolMuted()
 {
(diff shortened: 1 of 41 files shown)


mods/taskbar-empty-space-clicks.wh.cpp
```diff
@@ -1697,40 +2083,69 @@ BOOL Wh_ModInit()
...
+    bool desktopWindowFound = false;
+    for (int i = 0; i < 10; i++)
     {
-        Wh_Log(L"Failed to find Desktop window. Hide icons feature will not be available!");
+        if (FindDesktopWindow())
+        {
+            desktopWindowFound = true;
+            break;
+        }
+        std::this_thread::sleep_for(std::chrono::milliseconds(50));
```
m417z: `Wh_ModInit` blocks the process execution if the mod is loaded before the process (explorer.exe) started executing. In this case, the desktop window will never be found, and this will just delay the initialization. IMO it's best to find it lazily when the relevant action is performed.

PR #508, opened 2024-02-07
Create classic-theme-explorer.wh.cpp

Anixx:
I took source code from here: https://gitdab.com/Cynosphere/WindhawkMods/src/branch/main The license is specified as MIT, so I think, we have a right to include it (anyway, it is based on EP code). Besides this, I fixed this bug that comes from Explorer Patcher and which EP team refused to fix: https://github.com/valinet/ExplorerPatcher/issues/2316


m417z:
As the failed check indicates, you can't submit a mod that specifies another author. You can mention the original author in the comments or the mod description, but a mod you submit is a mod you take ownership of and only you can update, so it must specify you as the author.

PR #515, opened 2024-02-08
Create w11-dwm-fix.wh.cpp

teknixstuff:


mods/w11-dwm-fix.wh.cpp
```diff
@@ -0,0 +1,106 @@
+// ==WindhawkMod==
+// @id              w11-dwm-fix
+// @name            Bring Back the Borders!
+// @description     Restores borders and corners
+// @version         1.0.0
+// @author          teknixstuff
+// @github          https://github.com/teknixstuff
+// @twitter         https://twitter.com/teknixstuff
```
m417z, line 8: To verify your Twitter account, please send me (https://twitter.com/m417z) a direct message with the following content:

> I attest that I'm the sole owner of both this Twitter account (https://twitter.com/teknixstuff) and the following GitHub account: https://github.com/teknixstuff

It's not that something looks suspicious, it's just that I commit to Windhawk users that all profiles are verified.

teknixstuff: Done. Also note that this mod is only confirmed to work on 23H2. On 22H2 it may be broken.

m417z: That's something that would be great to have in the readme.

m417z: Also, while at it, how about adding screenshots demonstrating what the mod does?

teknixstuff:
I've got a v1.1.0 that now is confirmed to work on 22H2 as well, and also plays nicer with DWMBlurGlass.

PR #528, opened 2024-02-17
Create classic-taskbar-buttons-lite.wh.cpp

Anixx:
3D taskbar buttons for Classic theme, lightweight

### DIFF
--- a/mods/classic-taskbar-buttons-lite.wh.cpp
+++ b/mods/classic-taskbar-buttons-lite.wh.cpp
@@ -11,7 +11,7 @@
 
 // ==WindhawkModReadme==
 /*
-# Classic Taskbar 3D buttons
+# Classic Taskbar Fix
 Lightweight mod which restores 3D buttons on taskbar when using Windows Classic theme. 
 The idea is based on the mod by Aubymori (https://github.com/aubymori).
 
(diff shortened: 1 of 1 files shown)


m417z:
Please include `@github`.

Anixx:
Done

mods/classic-taskbar-buttons-lite.wh.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+#include <windhawk_utils.h>
+
+
+typedef struct tagBUTTONRENDERINFOSTATES {
+    char data[12];
+} BUTTONRENDERINFOSTATES, *PBUTTONRENDERINFOSTATES;
+
+/* Draw taskbar item */
+typedef void (* CTaskBtnGroup__DrawBar_t)(void *, HDC, void *, void *);
```
m417z, line 44: Calling conventions are missing, which is likely to cause a crash for 32-bit targets. Please add calling conventions or, if the mod doesn't support 32-bit, limit it to 64-bit, which will also make compilation faster. See also:

Anixx:
Done. Not sure if I did it right though. Do you mind including these declarations into Windhawk includes or a mod template?

Anixx:
Wait, it seems, it now does not work... I have no idea...

Anixx:
Fixed now.

PR #533, opened 2024-02-19
Add Classic Explorer Treeview

CyprinusCarpio:
This mod changes the appearance of the default explorer treeview (also known as navigation pane and/or folder pane), to make it look more like the one used before Vista.

It hooks 3 member functions of CNscTree in explorerframe.dll:

1. CNscTree::_CreateTreeview
2. CNscTree::ApplyTopMargin
3. CNscTree::SetStateImageList

Mod has been tested by a few people for over a week and it seems that it's stable and ready to be pushed to this repository.

### DIFF
--- a/mods/classic-explorer-treeview.wh.cpp
+++ b/mods/classic-explorer-treeview.wh.cpp
@@ -1,0 +1,1040 @@
+// ==WindhawkMod==
+// @id              classic-explorer-treeview
+// @name            Classic Explorer Treeview
+// @description     Modifies Folder Treeview in file explorer so as to make it look more classic.
+// @version         1.0
+// @author          Waldemar
+// @github          https://github.com/CyprinusCarpio
+// @include         explorer.exe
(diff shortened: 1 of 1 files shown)


mods/classic-explorer-treeview.wh.cpp
```diff
@@ -0,0 +1,1040 @@
...
+        SetCursor(LoadCursor(NULL, IDC_ARROW));
+        return TRUE;
+    }
+
+    return DefSubclassProc(hWnd, uMsg, wParam, lParam);
+}
+
+typedef HWND (*NSCCreateTreeview_t)(void*, HWND);
+NSCCreateTreeview_t NSCCreateTreeviewOriginal;
+HWND NSCCreateTreeviewHook(void* pThis, HWND hWnd)
```
m417z: Calling conventions are missing, which is likely to cause a crash for 32-bit targets. Please add calling conventions or, if the mod doesn't support 32-bit, limit it to 64-bit, which will also make compilation faster. See also:

PR #536, opened 2024-02-22
Classic Explorer Treeview: Update 1.0.1

CyprinusCarpio:
Minor update.
Among other changes, the Treeview control is no longer subclassed using `SetWindowSubclassFromAnyThread`, instead a member function of the `CNscTree` is hooked. Correct symbols are provided for both 32 and 64 bit versions, but the hook function signature remains the same on both. I'm not sure if that's correct, I copied the signature from another mod.


m417z:
Next time, please include the changelog changes in the commit message.

PR #541, opened 2024-02-26
Patch 17

Anixx:
Adds WS_EX_CLIENTEDGE (3D border) to some apps to look better in Classic theme. Includes SysListView32 control in 
File Explorer, legacy (XP-like) file picker, Classic Notepad and Wolfram Mathematica.

### DIFF
--- a/mods/clientedge-in-apps.wh.cpp
+++ b/mods/clientedge-in-apps.wh.cpp
@@ -1,0 +1,69 @@
+// ==WindhawkMod==
+// @id              clientedge-in-apps
+// @name            Clientedge (3D border) in choosen apps
+// @description     Adds WS_EX_CLIENTEDGE style to some apps to look better in Classic theme.
+// @version         1.0.0
+// @author          anixx
+// @github          https://github.com/Anixx
+// @include         *
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+Adds WS_EX_CLIENTEDGE (3D border) to some apps to look better in Classic theme. Includes SysListView32 control in 
+File Explorer, legacy (XP-like) file picker, Classic Notepad and Wolfram Mathematica.
+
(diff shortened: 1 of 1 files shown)


mods/clientedge-in-apps.wh.cpp
```diff
@@ -0,0 +1,72 @@
...
+CreateWindowExW_t CreateWindowExW_Orig;
+HWND WINAPI CreateWindowExW_Hook(DWORD dwExStyle,LPCWSTR lpClassName,LPCWSTR lpWindowName,
+DWORD dwStyle,int X,int Y,int nWidth,int nHeight,HWND hWndParent,HMENU hMenu,HINSTANCE hInstance,LPVOID lpParam) {
+
+    WCHAR wszClassName[256];
+    ZeroMemory(wszClassName, 256);
+
+    // Explorer and XP-like File Picker
+
+    if ((*((WORD*)&(lpClassName)+1)) && !wcscmp(lpClassName, L"SysListView32")) {
```
m417z: `(*((WORD*)&(lpClassName)+1))` is not a good check, it checks 2 bytes out of 8 in 64-bit, there can be a pointer that looks like `0xaaaabbbb0000cccc` which you'll ignore.

Anixx: What should I use instead?

m417z: Here's an example:
https://github.com/ramensoftware/windhawk-mods/blob/1cd6db586b784a7f60cd854d8c18e726b6e42626/mods/taskbar-thumbnail-reorder.wh.cpp#L586-L589

Anixx: Some hint, at which line?

Anixx:
Cjanged this.

PR #549, opened 2024-03-04
Create classic-file-picker-dialog.wh.cpp

Anixx:
Classic file picker

### DIFF
--- a/mods/classic-file-picker-dialog.wh.cpp
+++ b/mods/classic-file-picker-dialog.wh.cpp
@@ -1,0 +1,67 @@
+// ==WindhawkMod==
+// @id              classic-file-picker-dialog
+// @name            Classic File Picker dialog
+// @description     Redirect the Windows Vista+ file picker to the Windows XP one
+// @version         1.0
+// @author 	        Anixx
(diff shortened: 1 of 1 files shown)


mods/classic-file-picker-dialog.wh.cpp
```diff
@@ -0,0 +1,67 @@
...
+    param_1 = (param_1 > 1) ? param_1 -= 2 : param_1;
+    return CFileOpenSave__s_CreateInstance_orig(param_1, param_2, param_3);
+}
+
+BOOL Wh_ModInit() {
+    HMODULE module = LoadLibrary(L"comdlg32.dll");
+
+    WindhawkUtils::SYMBOL_HOOK symbolHook[] = {
+        {
+            {L"int" SCALCON L"IsCommonItemDialogAvailable(void)"},
```
m417z, line 53: Are you sure it should be `__thiscall` for 32-bit? I checked the copies of `comdlg32.dll`, and the relevant symbol is:
```
int __stdcall IsCommonItemDialogAvailable(void)
```

Same for the second symbol below.

Anixx:
Fixed this

PR #566, opened 2024-03-16
Create hide-action-center-icon.wh.cpp

Anixx:
Hide Action Center icon

### DIFF
--- a/mods/hide-action-center-icon.wh.cpp
+++ b/mods/hide-action-center-icon.wh.cpp
@@ -1,0 +1,37 @@
+// ==WindhawkMod==
+// @id              hide-action-center-icon
+// @name            Hide Action Center icon
+// @description     Hides Action Center icon from Win10 taskbar
+// @version         1.0.0
+// @author          anixx
+// @github          https://github.com/Anixx
+// @include         explorer.exe
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/* Removes the Action Center icon (cogwheel) from Win10 taskbar */
+// ==/WindhawkModReadme==
+
+using CreateWindowExW_t = decltype(&CreateWindowExW);
+CreateWindowExW_t CreateWindowExW_Orig;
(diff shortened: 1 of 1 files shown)


mods/hide-action-center-icon.wh.cpp
```diff
@@ -0,0 +1,37 @@
...
+/* Removes the Action Center icon (cogwheel) from Win10 taskbar */
+// ==/WindhawkModReadme==
+
+using CreateWindowExW_t = decltype(&CreateWindowExW);
+CreateWindowExW_t CreateWindowExW_Orig;
+HWND WINAPI CreateWindowExW_Hook(DWORD dwExStyle,LPCWSTR lpClassName,LPCWSTR lpWindowName,
+DWORD dwStyle,int X,int Y,int nWidth,int nHeight,HWND hWndParent,HMENU hMenu,HINSTANCE hInstance,LPVOID lpParam) {
+
+    WCHAR wszClassName[200];
+    ZeroMemory(wszClassName, 200);
```
m417z: These two lines can be removed.

mods/hide-action-center-icon.wh.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+	HWND hWnd = CreateWindowExW_Orig(dwExStyle,lpClassName,lpWindowName,dwStyle,X,Y,nWidth,nHeight,hWndParent,hMenu,hInstance,lpParam);
+	if ((((ULONG_PTR)lpClassName & ~(ULONG_PTR)0xffff) != 0) && !wcscmp(lpClassName, L"ControlCenterButton"))
+    {   
+        BYTE* lpDisplayCCButton = (BYTE*)(GetWindowLongPtrW(hWnd, 0) + 120);
```
m417z, line 27: These offsets change with time, and it can break in the future. How about adding the following to the readme:
* The Windows 10 version that the mod was designed for. I don't think this mod will work on all Win10 versions.
* That explorer has to be restarted for the mod to take effect.

Anixx:
Fixed. The code is based on EP code so, it should work everywhere where EP works.

PR #567, opened 2024-03-17
Create classic-list-group-fix.wh.cpp

travy-patty:
This mod makes group headers on Classic Theme appear how they do on Windows XP.

Mod is being uploaded with permission of @aubymori, the original creator of the mod.


aubymori:
Confirming that I gave Travis permission to submit this

PR #583, opened 2024-03-31
Create classic-taskbar-buttons-lite-vs-without-spacing.wh.cpp

OrthodoxWindows:
Added "Classic Taskbar 3D buttons with extended compatibility" mod. This mod is close to the equivalent of anixx, but it supports visual styles, and does not contain spacing between icons (which, under the classic theme, fixes the problem when displaying icons without the labels).

### DIFF
--- a/mods/classic-taskbar-buttons-lite-vs-without-spacing.wh.cpp
+++ b/mods/classic-taskbar-buttons-lite-vs-without-spacing.wh.cpp
@@ -1,0 +1,140 @@
+// ==WindhawkMod==
+// @id              classic-taskbar-buttons-lite-vs-without-spacing
+// @name            Classic Taskbar 3D buttons with extended compatibility
+// @description     Restoring the 3D taskbar buttons with support for visual styles and the taskbar without labels
+// @version         1.0
+// @author          OrthodoxWin32
+// @github          https://github.com/OrthodoxWindows
+// @include         explorer.exe
+// @compilerOptions -lgdi32 -lUxTheme
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Classic Taskbar Fix
+Restores 3D buttons on taskbar when using Windows Classic theme or Windows Visual Style. Based on mod by Anixx (https://github.com/anixx), itself founded from a mod by Aubymori (https://github.com/aubymori).
+Known issue : Under the classic theme, icons have no space between them. It is possible to add space, but this prevents the taskbar from functioning correctly if the labels are hidden.
+*/
+// ==/WindhawkModReadme==
+
+#include <windhawk_utils.h>
+#include <windows.h>
+#include <uxtheme.h>
+#include <tmschema.h>
+
+
+typedef struct tagBUTTONRENDERINFOSTATES {
+    char data[12];
+} BUTTONRENDERINFOSTATES, *PBUTTONRENDERINFOSTATES;
+
+/* Remove clock hover effect */
+typedef void (* GDIHelpers_FillRectARGB_t)(HDC, LPCRECT, BYTE, DWORD, bool);
+GDIHelpers_FillRectARGB_t GDIHelpers_FillRectARGB_orig;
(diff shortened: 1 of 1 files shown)


mods/classic-taskbar-buttons-lite-vs-without-spacing.wh.cpp
```diff
@@ -0,0 +1,140 @@
...
+#include <uxtheme.h>
+#include <tmschema.h>
+
+
+typedef struct tagBUTTONRENDERINFOSTATES {
+    char data[12];
+} BUTTONRENDERINFOSTATES, *PBUTTONRENDERINFOSTATES;
+
+/* Remove clock hover effect */
+typedef void (* GDIHelpers_FillRectARGB_t)(HDC, LPCRECT, BYTE, DWORD, bool);
```
m417z: Calling conventions are missing in the typedefs, which is likely to cause a crash for 32-bit targets. Please add calling conventions or, if the mod doesn't support 32-bit, limit it to 64-bit, which will also make compilation faster. See also:

OrthodoxWindows: I'll watch it when I have time.

OrthodoxWindows:
It should work (on my side it works).

mods/classic-taskbar-buttons-lite-vs-without-spacing.wh.cpp
m417z: What I meant is that the typedefs should have calling conventions like the function signatures, for example:
```cpp
typedef void (*__cdecl GDIHelpers_FillRectARGB_t)(HDC, LPCRECT, BYTE, DWORD, bool);
```
Also, in recent Windhawk versions you can remove the casts, and the compiler will verify that the hook signature matches the original function pointer. For example:
```cpp
            &GDIHelpers_FillRectARGB_orig,
            GDIHelpers_FillRectARGB_hook,
```
Instead of:
```cpp
            (void **)&GDIHelpers_FillRectARGB_orig,
            (void *)GDIHelpers_FillRectARGB_hook,
```

If you do that, you'll also notice that your `CTaskBtnGroup__DrawBar_t` signature doesn't match the hook (the last argument, `PBUTTONRENDERINFOSTATES`, is missing), but it doesn't affect anything since you never call `CTaskBtnGroup__DrawBar_orig`.

OrthodoxWindows: It's okay, I modified it.

mods/classic-taskbar-buttons-lite-vs-without-spacing.wh.cpp
```diff
@@ -0,0 +1,139 @@
...
+            &GDIHelpers_FillRectARGB_orig,
+            GDIHelpers_FillRectARGB_hook,
+            FALSE
+        },
+        {
+            {
+                L"private: void __cdecl CTaskBtnGroup::_DrawBar(struct HDC__ *,struct BUTTONRENDERINFO const &,struct BUTTONRENDERINFOSTATES const &)"
+            },
+            (void **)&CTaskBtnGroup__DrawBar_orig,
+            (void *)CTaskBtnGroup__DrawBar_hook,
```
m417z: Any reason you left the casts here?

OrthodoxWindows: If I remove them the compilation fails.

m417z: That's because the signature doesn't match, it should be:
```cpp
typedef void (*__cdecl CTaskBtnGroup__DrawBar_t)(void *, HDC, void *, PBUTTONRENDERINFOSTATES);
```
instead of:
```cpp
typedef void (*__cdecl CTaskBtnGroup__DrawBar_t)(void *, HDC, void *, void *);
```

OrthodoxWindows: It works, thank you.

PR #596, opened 2024-04-15
Aerexplorer

aubymori:
This mod includes various tweaks for the Windows Explorer to look and behave
more like its Windows Vista, 7, 8, or older Windows 10 counterpart.

# Showcase

**Configured like Windows Vista**

Windows Vista

**Configured like Windows 7**

Windows 7

**Configured like Windows 8.0**

Windows 8.0

**Configured like Windows 10 RTM**

Windows 10 RTM

### DIFF
--- a/mods/aerexplorer.wh.cpp
+++ b/mods/aerexplorer.wh.cpp
@@ -1,0 +1,3068 @@
+// ==WindhawkMod==
+// @id              aerexplorer
+// @name            Aerexplorer
+// @description     Various tweaks for Windows Explorer to make it more like older versions.
+// @version         1.5.7
+// @author          aubymori
+// @github          https://github.com/aubymori
+// @include         *
+// @compilerOptions -lgdi32 -lcomctl32 -lole32 -loleaut32 -luxtheme -ldwmapi
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Aerexplorer
+This mod includes various tweaks for the Windows Explorer to look and behave
+more like its Windows Vista, 7, 8, or older Windows 10 counterpart.
+
+# IMPORTANT: READ!
+Windhawk needs to hook into `svchost.exe` to successfully capture the creation of Explorer windows.
+Please navigate to Windhawk's Settings, Advanced settings, More advanced settings, and make sure that
+`svchost.exe` is in the Process inclusion list.
+
+# Showcase
+
+**Configured like Windows Vista**
+
+![Windows Vista](https://raw.githubusercontent.com/aubymori/images/main/aerexplorer-vista.png)
+
+**Configured like Windows 7**
+
+![Windows 7](https://raw.githubusercontent.com/aubymori/images/main/aerexplorer-7.png)
+
+**Configured like Windows 8.0**
+
+![Windows 8.0](https://raw.githubusercontent.com/aubymori/images/main/aerexplorer-8.png)
+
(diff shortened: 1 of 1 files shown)


mods/aerexplorer.wh.cpp
```diff
@@ -0,0 +1,3068 @@
...
+        pSharedMemoryCache((void **)pSharedMemoryCache) {}
+
+    CMWF_SYMBOL_HOOK() = default;
+};
+
+/*
+ * CmwfHookSymbols: A custom hook wrapper which allows storing symbol hook results in the
+ *                  module's shared memory for faster access.
+ */
+bool CmwfHookSymbols(
```
m417z, line 2430: I don't mind, but I hardly believe that it has any effect on performance. Did you measure it?

aubymori: > I don't mind, but I hardly believe that it has any effect on performance. Did you measure it?

It's been a while since this was added to Aerexplorer, but if I remember correctly, initialization with this is noticeably faster than without, considering the large amount of symbol hooks this mod uses.

m417z: Perhaps it was compared to loading symbols each time, using `Wh_FindFirstSymbol`/`Wh_FindNextSymbol` and not `WindhawkUtils::HookSymbols` which has a caching mechanism.

aubymori: Aerexplorer has used `WindhawkUtils` from the start.

m417z: Versions before 1.4 (1.3, 1.3.1) also had a bug with caching of 32-bit and 64-bit processes overriding each other. Maybe that's why you got slowness. If not, I'm not sure what it was, but if you can figure it out, let me know, I'd rather improve it in Windhawk for all mods.

BTW there's the issue of all processes loading the symbols simultaneously which is not ideal as well, and I have it on my list to improve it, but it happens only at the first time before symbols are loaded, and this shared memory trick doesn't improve it.

m417z: I was looking at symbol usage of various mods, and stumbled upon the following in Aerexplorer:
```cpp
    /* This needs to be hooked EARLY. */
    const CMWF_SYMBOL_HOOK isCplHooks[] = {/*...*/};
        
    if (!CmwfHookSymbols(
        hExplorerFrame,
        isCplHooks,
        ARRAYSIZE(isCplHooks)
    ))
    {/*...*/}

    const CMWF_SYMBOL_HOOK efHooks[] = {/*...*/};

    if (!CmwfHookSymbols(
        hExplorerFrame,
        efHooks,
        ARRAYSIZE(efHooks)
    ))
    {/*...*/}
```

Running `WindhawkUtils::HookSymbols` more than once for the same module isn't a good idea, since the cache will be overridden every time, and so each call will enumerate all symbols, which can be slow. Perhaps that's why you saw that using the `CmwfHookSymbols` wrapper helps in your case. I suggest to merge the two hook arrays for ExplorerFrame.dll and I think it's also better to avoid the `CmwfHookSymbols` wrapper since I don't believe it has any positive effect on performance.

Regarding your "This needs to be hooked EARLY" comment, your separation has no effect, as the hooks are only applied once `Wh_ModInit` returns. See [the Mod lifetime wiki](https://github.com/ramensoftware/windhawk/wiki/Mod-lifetime) for details.

PR #599, opened 2024-04-15
Eradicate Immersive Menus 1.1.0 - Use shared memory to boost hooking performance

aubymori:


m417z:
Have you seen [my other comment](https://github.com/ramensoftware/windhawk-mods/pull/596#discussion_r1565569972)?

> I don't mind, but I hardly believe that it has any effect on performance. Did you measure it?

aubymori:
> Have you seen [my other comment](https://github.com/ramensoftware/windhawk-mods/pull/596#discussion_r1565569972)?
> 
> > I don't mind, but I hardly believe that it has any effect on performance. Did you measure it?

I thought it would be good to implement considering it's a global mod, but I see now that it doesn't really have any significant effect. I'll remove it while keeping the other improvements.

PR #601, opened 2024-04-17
Add Remove Command Bar

CyprinusCarpio:
This mod removes the Vista or 7 Command Bar from file explorer windows without having to patch shellstyle.dll using external tools.

### DIFF
--- a/mods/remove-command-bar.wh.cpp
+++ b/mods/remove-command-bar.wh.cpp
@@ -1,0 +1,102 @@
+// ==WindhawkMod==
+// @id              remove-command-bar
+// @name            Remove Command Bar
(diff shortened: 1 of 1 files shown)


Undisputed00x:
@CyprinusCarpio can you also make a setting for user to manually adjust the bottom margin, your code has -40rp when in my case for CommandBar to be completely hidden, bottom margin needed to be -52rp.

CyprinusCarpio:
Ok, will do.

> @CyprinusCarpio can you also make a setting for user to manually adjust the bottom margin, your code has -40rp when in my case for CommandBar to be completely hidden, bottom margin needed to be -52rp.

Ok, will do

PR #604, opened 2024-04-19
Classic Explorer Treeview: Update 1.1

CyprinusCarpio:
- Enabling the folders pane will now hide real InfoBands and vice versa
- Fixed a hidden bug where clicking the X button many times would hang Explorer
- Added a working entry in the Explorer Bars submenu
- Removed 32bit "support"


CyprinusCarpio:
Mod didn't work on 32 bit installs anyways, I tested. I think 32bit support doesn't make sense for a explorer.exe mod, if your explorer.exe is 32 bit then you've got worse problems to worry about than how it looks.

PR #608, opened 2024-04-21
UWP clean

arukateru:
The Universal Windows Platform (UWP), in the spirit of being unoptimized, keep all of their many processes operating all of the time: at the logon of the user, RuntimeBroker automatically restarting, and even when UWP apps are closed. This mod makes ApplicationFrameHost and RuntimeBroker automatically terminate unless Settings or Windows Security are open. You can add to the list of apps to keep the programs open if you so choose.


m417z:
@arukateru not sure if you noticed, but as the failed check says, your file name doesn't match the mod id.

arukateru:
Changed

PR #645, opened 2024-05-04
Chrome UI Tweaks 1.0.0

VasherMC:
Tweak some parts of the Chrome UI. Uses binary-based hooking instead of symbol-based hooking, and as such currently only applies to x86-64. Currently has two tweaks: Disable Simplified Bookmark Flow and MenuConfig (with configurable values).

Currently I use '-1' as a default value instead of '0' (because I want users to be able to explicitly set a value to zero); not sure if this is the best solution. (Also several MenuConfig fields are omitted from the settings because either I believe they don't apply to Windows, or they can be changed via the OS).

### DIFF
--- a/mods/chrome-ui-tweaks.wh.cpp
+++ b/mods/chrome-ui-tweaks.wh.cpp
@@ -25,7 +25,7 @@ meaning fewer menu items can fit on screen.
 
 This tweak lets you configure many menu UI values.
 This will affect several kinds of menus:
-  - main options/settings menu (three dots at the top right, "Customize and control Google Chrome")
+  - Main options/settings menu (three dots at the top right, "Customize and control Google Chrome")
   - Bookmark menus (not including the bookmark bar itself)
   - Context menu (when you right-click)
 
@@ -37,6 +37,7 @@ and you will probably want to keep them relatively small (below 50), but feel fr
 All other MenuConfig mod settings are negative one (-1) by default.
 If any MenuConfig mod setting is -1, it will be ignored.
 You can use this to reset a value back to default.
+Other negative values are increased by one (eg, setting of -3 will be interpreted as -2).
 Actual default values are described for each option if they exist/are known;
 the four potential defaults are:
 * Old, for default values usually originating from before Chrome 75 (before 2019)
@@ -70,13 +71,15 @@ Latest tested Chrome version: 124.0.6367.119 (stable), 126.0.6457.0 (canary)
 - menuconfig:
   - item:
     - item_vertical_margin: 3
-      $description: Margins between the item top/bottom and its contents. Old = 4, CR2023 = 6, Win10 = 3
+      $description: Margins between the item top/bottom and its contents.
+                    Creates spacing between items that is still clickable. Old = 4, CR2023 = 6, Win10 = 3
     - between_item_vertical_padding: -1
-      $description: Padding, if any, between successive menu items. Default = 0, Win11 = 2
+      $description: Padding, if any, between successive menu items.
+                    Creates spacing between items that is NOT clickable. Default = 0, Win11 = 2
     - item_horizontal_padding: -1
       $description: Horizontal padding between components in a menu item. Default = 8, Win11 = 12
     - item_horizontal_border_padding: -1
-      $description: Additional padding between the item left/right and its contents. Old = 0, CR2023 = 12, Win10 = -2
+      $description: Additional padding between the item left/right and its contents. Old = 0, CR2023 = 12, Win10 = -2 (setting -3)
     - item_corner_radius: -1
       $description: Radius of selection background on menu items. Default = 0, Win11 = 4
     $description: Config related to individual Menu items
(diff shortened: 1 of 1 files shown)


mods/chrome-ui-tweaks.wh.cpp
```diff
@@ -0,0 +1,847 @@
...
+When creating a new bookmark, you will immediately see the "Edit Bookmark" window.
+
+## Menu UI customization
+"Chrome Refresh 2023" brought significant changes to the user interface, including rounding the corners
+of menus (as well as many other UI elements) and significantly increasing the spacing between items,
+meaning fewer menu items can fit on screen.
+
+This tweak lets you configure many menu UI values.
+This will affect several kinds of menus:
+  - main options/settings menu (three dots at the top right, "Customize and control Google Chrome")
```
m417z: OCD remark: `main` not capitalized :)

mods/chrome-ui-tweaks.wh.cpp
```diff
@@ -0,0 +1,847 @@
...
+const std::string_view BookmarkBubble_instructions =
+    "\x88\x5c\x24\x30"sv        // mov byte ptr [rsp+0x30], bl  (already_bookmarked = bl)
+    "\x48\x89\x74\x24\x28"sv    // mov qword ptr [rsp+0x28], rsi
+    "\x48\x89\x6c\x24\x20"sv    // mov qword ptr [rsp+0x20], rbp
+    "\x48\x89\xf9"sv            // mov rcx, rdi
+    "\x48\x89\xc2"sv            // mov rdx, rax
+    "\x4d\x89\xf8"sv            // mov r8, r15
+    "\x4d\x89\xe1"sv            // mov r9, r12
+    "\xe8"sv;                   // call
+const std::string_view BookmarkBubble_prologue = "AWAVAUATVWUS"sv;
```
m417z, line 634: [Nice one](https://stackoverflow.com/questions/39322552/meaning-of-a-common-string-in-executables) :)

mods/chrome-ui-tweaks.wh.cpp
```diff
@@ -0,0 +1,847 @@
...
+
+1. To search the symbols, we need to download chrome.dll.pdb from google's symbol server
+   for our chrome.dll version.
+2. chrome.dll.pdb is upwards of **4 GB**, which is all mapped into the process memory.
+3. over 50% of the time parsing the symbol table is spent in undecorating symbol names,
+   even with the noUndecoratedSymbols option set - maybe I just didn't figure out the
+   right way, or maybe undecorating is an unavoidable part of traversing symbols with
+   the DIA library Windhawk uses (msdia140_windhawk.dll).
+   The last time I tried it, it was busy parsing, undecorating, and traversing symbols
+   for over 4 hours (pegging one core at 100%) before I killed the process.
```
m417z, line 820: I thought it took about 8 minutes with `noUndecoratedSymbols`? Still a lot, but not as much.

mods/chrome-ui-tweaks.wh.cpp
```diff
@@ -0,0 +1,847 @@
...
+4. While the symbol->address mapping can be cached, it needs to be updated every time
+   the program updates. Due to relatively frequent minor version and security updates,
```
m417z, line 822: I thought about another possible approach: generating the mapping on the server. It will require some effort, but what I had in mind is having a repository with such mappings, and a scheduled GitHub Actions workflow that will run from time to time and check for Chrome updates. If there's an update, it will download it, get the symbols, and will created the mapping for the required functions. The mod will look for the mappings in that repository.

Pros:
* Advanced mods can be created more easily without having to create and maintain signatures.
* More resilient to Chrome updates.

Cons:
* Such a GitHub Actions workflow has to be developed.
* There can be a time gap until the mappings are updated to a newly released version (less of a problem IMO as Chrome updates gradually).
* Versions without a mapping (beta etc.) won't be supported.

I think what you did is perfectly fine, just some food for thought for future development.

In fact, I'll probably develop something like this, manual and temporary for now, due to [the Microsoft symbols incident](https://ramensoftware.com/windhawk-and-symbol-download-errors).

m417z:
> Currently I use '-1' as a default value instead of '0' (because I want users to be able to explicitly set a value to zero); not sure if this is the best solution.

I think it's fine. I [encountered it too](https://github.com/ramensoftware/windhawk-mods/blob/a36ce2c5a985dcc2822d344ccf8aacd4650fb59a/mods/taskbar-clock-customization.wh.cpp#L1685-L1689), and I chose to use `0` as the default value and adjust the other values such that negative values get a `+1`, and so all values can be selected. While fine for padding that can be positive or negative, it's more confusing for positive-only values such as corner radius, so I think what you did is fine.

BTW I see that in your case, some values can be negative too, so maybe you'll want to `+1` negative values too.

Ideally, Windhawk would provide some "optional int" for this case. Maybe I'll add something like this in the future.

mods/chrome-ui-tweaks.wh.cpp
VasherMC: I agree that sounds like a reasonable idea. I'm not familiar with Github Actions and probably won't try that myself, since I'm more focused on the mod itself (especially since it's already working). I'd probably want to switch over to that if you end up making it, maybe keeping the existing method as a fallback.

mods/chrome-ui-tweaks.wh.cpp
VasherMC: IIRC, it had been running for about 8 minutes when I first posted about it; after that it was still going for a few hours after I woke up my machine the next day before I pulled the plug. I didn't really try to test it again after that, and only really see myself looking into it again when the instruction patterns break.

PR #653, opened 2024-05-04
Taskbar Thumbnail Reorder v1.0.7

m417z:
* Added an online symbol cache mechanism as a temporary workaround for the unavailable Microsoft symbols. Currently, this makes the mod work on Windows 11 versions 22631.3447 and 22631.3527. For more details, refer to [the relevant blog post](https://ramensoftware.com/windhawk-and-symbol-download-errors).

### DIFF
--- a/mods/taskbar-thumbnail-reorder.wh.cpp
+++ b/mods/taskbar-thumbnail-reorder.wh.cpp
@@ -2,14 +2,14 @@
 // @id              taskbar-thumbnail-reorder
 // @name            Taskbar Thumbnail Reorder
 // @description     Reorder taskbar thumbnails with the left mouse button
-// @version         1.0.6
+// @version         1.0.7
 // @author          m417z
 // @github          https://github.com/m417z
 // @twitter         https://twitter.com/m417z
 // @homepage        https://m417z.com/
 // @include         explorer.exe
 // @architecture    x86-64
-// @compilerOptions -lcomctl32 -lversion
+// @compilerOptions -lcomctl32 -lversion -lwininet
 // ==/WindhawkMod==
 
 // Source code is published under The GNU General Public License v3.0.
@@ -48,6 +48,7 @@ or a similar tool), enable the relevant option in the mod's settings.
 // ==/WindhawkModSettings==
 
 #include <commctrl.h>
+#include <wininet.h>
 
 #include <algorithm>
 #include <string>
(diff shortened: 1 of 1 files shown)


AndroidDeveloperLB:
Say, is this only for the thumbnails, or also for the tasks in case it's not grouped?
For example if there are multiple tasks of the same app, and I want to reorder them, would it let me?

m417z:
@AndroidDeveloperLB Yes, the order of the grouped taskbar items is synchronized with the order of the thumbnails.

AndroidDeveloperLB:
@m417z Doesn't seem to work. It moves the entire group.

Watch:
[2024-07-23_03-46-51.zip](https://github.com/user-attachments/files/16341396/2024-07-23_03-46-51.zip)

m417z:
Moving items on the taskbar moves the entire group. Reordering thumbnails changes the order of the group on the taskbar, too, which is what I thought you were asking.

AndroidDeveloperLB:
@m417z Sorry but I don't understand. I asked about dragging of tasks.
Are you saying that I can drag the thumbnails, which will also change order of tasks?
If so, for some reason it doesn't work for me, perhaps because of a combination of the mods I've chosen?
Only when "Customize the old taskbar on Windows 11" is turned off, it works, but even then with various glitches. 

Please check it out:

[Carnac_2024-07-23_10-51-56.zip](https://github.com/user-attachments/files/16345452/Carnac_2024-07-23_10-51-56.zip)

Can you please create a mod to change the order of the tasks by dragging the tasks themselves? It was possible using "7+ Taskbar Tweaker" in the past...

m417z:
> Are you saying that I can drag the thumbnails, which will also change order of tasks?

Yes.

> Only when "Customize the old taskbar on Windows 11" is turned off, it works, but even then with various glitches.

Since you aren't using the old taskbar, you don't need to turn it on.

You can see on 00:50 that it works, you can re-order the items.

> Can you please create a mod to change the order of the tasks by dragging the tasks themselves? It was possible using "7+ Taskbar Tweaker" in the past...

I think that what you're looking for is the [Disable grouping on the taskbar](https://windhawk.net/mods/taskbar-grouping) mod.

AndroidDeveloperLB:
> you don't need to turn it on
I don't think I enabled it. Was turned on when installed.

> You can see on 00:50 that it works, you can re-order the items.
I know. That's why I wrote "...it works, but even then with various glitches.".

About the "Disable grouping on the taskbar ", it doesn't seem to do anything at all, even if I try to change its settings and disable other relevant mods

m417z:
From the description:

> Note: After enabling the mod, the relevant windows must be reopened to apply the grouping settings.

m417z:
> What I have is already ungrouped, no?

What you have is what I call "grouped, not combined". Not a single button, but still stuck together - move together, have thumbnails together. The taskbar-grouping mod makes all items completely separated.

AndroidDeveloperLB:
@m417z I see. In the past I had just "7+ Taskbar Tweaker" that had it all. Now I have multiple mods here.
Hopefully they will all work nicely together.

Thank you very much.

m417z:
Having them as separate mods have advantages. For example, if one of the mods stops working, you can still use the other mods.

AndroidDeveloperLB:
@m417z Most choices in life have advantages and disadvantages
:)

PR #669, opened 2024-05-08
feat(mods/dot-hide): add setting to not hide some entries

SunsetTechuila:
I was annoyed that the `.github` folder was hidden in the repositories, so made this

I have never written c++ code before, so check carefully


m417z:
@realgam3 FYU, you might want to incorporate these changes and submit them as part of a new version.

PR #676, opened 2024-05-11
Remove arukateru mods

Typical0:
This pull request removes all mods made by arukateru (also known as Alcatel). Today, a discovery has been made that the creator of those 2 mods has spread modified version of DWMBlurGlass (impersonating the creator of the original project, Maplepse) with a token logger ([this one](https://github.com/can-kat/cstealer) to be exact). We believe that he might bring malicious changes to both of these mods.


OrthodoxWindows:
From what I have read of two mods published by Alcatel, there does not seem to be any malicious code. This proposed deletion therefore makes no sense.

TheInsane101:
> From what I have read of the two mods published by Alcatel, there does not seem to be any malicious code. This proposed deletion therefore makes no sense. I specify that I have no particular interest in this story, I condemn like everyone else what Alcatel has committed. I just ask that healthy code be preserved. Especially since each update of a mod must be approved, so I don't see where the risk is.

We can store the current versions of the mods as archives; like how many WH mods are distributed in forums instead of the WH store

m417z:
Mods aren't normally being removed from the repository. This rule can be bypassed in extraordinary circumstances, such as a mod that contains malicious code, or in another case of law violation. This isn't the case, both mods are short and clearly have no malicious code.

Still, I understand the concern. Therefore, I added a rule that prevents this author from making submissions, including new mods and updates to existing mods:
https://github.com/ramensoftware/windhawk-mods/blob/c90686ffc164d30bd1567b18aae6539541f6fad7/.github/pr_validation.py#L14-L17

PR #689, opened 2024-05-23
Modernize Folder Picker Dialog 1.0.0

aubymori:
This mod replaces the outdated, hard-to-use "Browse For Folder" dialog with amodern styled folder picker dialog, which is easier to navigate.

**Before**:
Before

**After**:
After

### DIFF
--- a/mods/modernize-folder-picker-dialog.wh.cpp
+++ b/mods/modernize-folder-picker-dialog.wh.cpp
@@ -6,7 +6,7 @@
 // @author          aubymori
 // @github          https://github.com/aubymori
 // @include         *
-// @compilerOptions -lole32
+// @compilerOptions -lole32 -loleaut32
 // ==/WindhawkMod==
 
 // ==WindhawkModReadme==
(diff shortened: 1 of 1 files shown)


mods/modernize-folder-picker-dialog.wh.cpp
```diff
@@ -0,0 +1,115 @@
...
+using SHBrowseForFolderW_t = decltype(&SHBrowseForFolderW);
+SHBrowseForFolderW_t SHBrowseForFolderW_orig = nullptr;
+PIDLIST_ABSOLUTE WINAPI SHBrowseForFolderW_hook(
+    LPBROWSEINFOW lpbi
+)
+{
+    if (!lpbi) return NULL;
+
+    /* Make the dialog */
+    IFileOpenDialog *pDialog = nullptr;
```
m417z: This object is never freed. Same for `pRoot`, `pResult` and `pszPath` below.
You can use `winrt::com_ptr` to have all but `pszPath` automatically freed, for example:
https://github.com/ramensoftware/windhawk-mods/blob/16c1c93b12f2e2fb707c419efb19a7d3c3b80a11/mods/explorer-frame-classic.wh.cpp#L444

For `pszPath`, you need to call `CoTaskMemFree`, I don't think there's a smart object for it to free it automatically.

PR #693, opened 2024-05-27
Add TranslucentStartMenu theme

Undisputed00x:


m417z:
As stated in the readme, only the original author (in this case, me) can update their mods.
I added your theme and released a new version.
Next time, you can submit pull requests to my mods here:

PR #706, opened 2024-05-30
RegEdit Disable Beep 1.0.0

ItsProfessional:
# RegEdit Auto Trim Whitespace on Navigation Bar
When an invalid key is entered into the navigation bar, RegEdit plays a beep sound. This mod disables this sound.

### DIFF
--- a/mods/regedit-disable-beep.wh.cpp
+++ b/mods/regedit-disable-beep.wh.cpp
@@ -6,7 +6,6 @@
 // @author          ItsProfessional
 // @github          https://github.com/ItsProfessional
 // @include         regedit.exe
-// @compilerOptions -lshlwapi
 // ==/WindhawkMod==
 
 // Source code is published under The GNU General Public License v3.0.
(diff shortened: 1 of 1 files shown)


mods/regedit-disable-beep.wh.cpp
```diff
@@ -0,0 +1,91 @@
...
+// @author          ItsProfessional
+// @github          https://github.com/ItsProfessional
+// @include         regedit.exe
+// ==/WindhawkMod==
+
+// Source code is published under The GNU General Public License v3.0.
+
+// ==WindhawkModReadme==
+/*
+# RegEdit Auto Trim Whitespace on Navigation Bar
```
m417z: You left the title from the previous mod.

ItsProfessional: Fixed.

ItsProfessional:
By the way, do you have an explanation for [this](https://discord.com/channels/923944342991818753/923954372981960754/1244916546732691477)? OnCreate_hook` doesn't seem to get called most of the time. Is it because Windhawk is unable to hook earlier than this function was called or something?

PR #721, opened 2024-06-04
CE Disable Process Button Flashing v1.0.1

ItsProfessional:
Fix terminated offset according to Cheat Engine 7.5. I somehow ended up using an older portable copy of Cheat Engine 7.4 (released 2 years ago) lying around on my disk whilst REing Cheat Engine for this mod, where the `terminated` offset was `0x10`. On the latest version, Cheat Engine 7.5, it is `0xC`. 32-bit offset is still same.


m417z:
Update the version number.

ItsProfessional:
Done.

PR #728, opened 2024-06-12
Vertical Taskbar for Windows 11 v1.0

m417z:


bagarwa:
Discovered this mod via XDA today. I have some feedback, but not finding a proper place to submit it. So leaving it here. We can delete it if that's not appropriate. 

Feedback: 
Though the taskbar moves to the left side, certain taskbar related actions remain in their original place. 
   - Clicking Start button / Search button opens those panels at bottom center of screen, instead of left center. 
   - Clicking OneDrive icon in system tray now opens a floating OneDrive window at bottom center of screen.
   - Clicking [WiFi-Volume-Battery] part of the taskbar opens that panel at bottom right corner, instead of bottom left.
   - Clicking [date-time-notification] part of the taskbar opens that panel at bottom right corner, instead of bottom left.
The disconnect between the click point [action], and result is a little jarring. Typically, users expect results to happen at the same place where the action was taken. 

I know it's the very first version. So, please accept the above as a feature request. 
Cheers.

m417z:
Thanks for the feedback @bagarwa. The proper way to submit the feedback is to create an issue for each bug report or enhancement idea.

pilanwu:
Don's compatible with  mod Windows 11 Taskbar Styler

m417z:
@pilanwu the existing themes are probably not compatible, but generally styling should work. What didn't work for you?

PR #750, opened 2024-06-24
timer-resolution-control: Force limits immediately on mod load and setting change, restore original resolution on mod unload

levitation:
timer-resolution-control: During mod start, force timer resolution limit if the resolution was already changed by the program. During settings change, enforce the new settings immediately by enforcing new limits or restoring the resolution desired earlier by the program. On mod unload restore the resolution desired by the program.

### DIFF
--- a/mods/timer-resolution-control.wh.cpp
+++ b/mods/timer-resolution-control.wh.cpp
@@ -2,8 +2,8 @@
 // @id           timer-resolution-control
 // @name         Timer Resolution Control
 // @description  Prevent programs from changing the Windows timer resolution and increasing power consumption
-// @version      1.0
-// @author       m417z
+// @version      1.1
+// @author       m417z, levitation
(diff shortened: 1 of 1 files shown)


levitation:
I am not sure what the validation check means here. As I understand the readme, since I am modifying an existing mod, I cannot be the author.

m417z:

> The mod's `github` metadata value must match the pull request author's GitHub profile. This means that you can only submit an update for a mod that you originally submitted. If you'd like to update a mod that you didn't submit, you can either submit the changes to the mod author, or submit a new mod instead.

In this case, I'm the mod author, and I develop my mods at https://github.com/m417z/my-windhawk-mods. You can submit a PR there.

Generally, Windhawk mods can have only one author, similarly to a GitHub repository which can have only a single owner. Additional contributors, credits for borrowed code, etc. can be specified in the readme.

PR #757, opened 2024-06-26
Windows 7 UAC Dim 1.0.0

aubymori:
This mod restores the full desktop screenshot on the background of the secure desktop for UAC. Since Windows 8, it has only displayed the wallpaper.

**Before**:
Before

**After**:
After


mods/win7-uac-dim.wh.cpp
```diff
@@ -0,0 +1,121 @@
...
+        &OnPaint_orig,
+        OnPaint_hook,
+        false
+    }
+};
+
+BOOL Wh_ModInit(void)
+{
+    /* Wait for windows and menus to go away */
+    Sleep(250);
```
aubymori: I want to preemptively say that this is probably the best you're going to get when it comes to this. Another application that captures the screen (Snipping Tool) just waits a set time before capturing the screen, and you can see the Snipping Tool window on the screenshot it takes if you use the regkey to slow down DWM animations with the shift key. Likewise, you can see the window or menu UAC was launched from if it launches fast enough, even in 7. It just seems that Windows 7's UAC is slow enough that these fade away before it opens most of the time, even on real hardware. Obviously, having a half faded window or menu item on the UAC background isn't very pleasant, so I just had to do this.

mods/win7-uac-dim.wh.cpp
```diff
@@ -0,0 +1,121 @@
+// ==WindhawkMod==
+// @id              win7-uac-dim
+// @name            Windows 7 UAC Dim
```
m417z: IMO the name makes the impression that the mod is for Win7. How about something like "Win7-style UAC Dim"?

mods/win7-uac-dim.wh.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+int x = GetSystemMetrics(SM_XVIRTUALSCREEN);
+int y = GetSystemMetrics(SM_YVIRTUALSCREEN);
+int cx = GetSystemMetrics(SM_CXVIRTUALSCREEN);
+int cy = GetSystemMetrics(SM_CYVIRTUALSCREEN);
+
+HBITMAP g_hbDesktop = NULL;
+
+void (*OnPaint_orig)(HWND) = nullptr;
+void OnPaint_hook(HWND hWnd)
```
m417z: Calling conventions are missing in the typedefs, which is likely to cause a crash for 32-bit targets. Please add calling conventions or, if the mod doesn't support 32-bit, limit it to 64-bit, which will also make compilation faster. See also:

PR #763, opened 2024-06-28
StartIsBack++ Tweaker (2.9.20) - v0.4

Erizur:
# StartIsBack++ Tweaker (2.9.20)
Modifies StartIsBack++'s internal functions for more customizable options.
Might not work if the StartIsBack++ DLL installed has been patched previously.
**Works only on StartIsBack++ 2.9.20. Functionality on older/newer versions is not guaranteed.**

## Features
As of now, the mod can do interesting things as restoring Windows 7's links, disable Hottracking or disable Start Menu animations.
Read the options or the [WinClassic post](https://winclassic.net/thread/2377/startisback-tweaker-2-9-20) for more info.

### DIFF
--- a/mods/sib-plusplus-tweaker.wh.cpp
+++ b/mods/sib-plusplus-tweaker.wh.cpp
@@ -1,0 +1,453 @@
+// ==WindhawkMod==
+// @id              sib-plusplus-tweaker
+// @name            StartIsBack++ Tweaker
+// @description     Modify StartIsBack++'s features (2.9.20)
+// @version         0.3
+// @author          Erizur
+// @github          https://github.com/Erizur
+// @include         explorer.exe
+// @architecture    x86-64
+// @compilerOptions -lcomdlg32 -luser32 -lole32 -lgdi32
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# StartIsBack++ Tweaker (2.9.20)
+Modifies StartIsBack++'s internal functions for more customizable options.
+Might not work on your end if you patched your StartIsBack++ DLL previously.
+**Works only on StartIsBack++ 2.9.20. Functionality on older/newer versions is not guaranteed.**
+
+## Features
+As of now, the mod can do interesting things as restoring Windows 7's links, disable Hottracking or disable Start Menu animations.
+Read the options or the [WinClassic post](https://winclassic.net/thread/2377/startisback-tweaker-2-9-20) for more info.
+
+## Special Thanks
+- Wiktorwiktor12: FindPattern function & Custom Folder help.
+- Aubymori: Additional help.
+*/
+// ==/WindhawkModReadme==
+
+// ==WindhawkModSettings==
+/*
+- SIBPath: "%PROGRAMFILES(X86)%\\StartIsBack\\StartIsBack64.dll"
+  $name: StartIsBack64.dll Path Location
+  $description: Place the path where your SIB++ dll (64-bit) is located. (Case Sensitive!)
+- DHTracking: false
+  $name: Remove Hottracking
+  $description: Remove SIB's Hottracking (Windows 7 Style) method.
+- DisableSMAnim: false
+  $name: Disable Start Menu animations
+  $description: Disable Start Menu's fade for accuracy to older Windows.
+- ForceDWM: false
+  $name: Force DWM Composition
+  $description: Force Taskbar to use DWM to render the taskbar blur. Won't work with Custom Color enabled.
+- DisableCDSB: false
+  $name: Disable Custom Drawn Scrollbar
+  $description: Enable this to draw the Start Menu scrollbar natively.
+- DisableCustomOrb: FALSE
+  $name: Disable Custom Orb
+  $description: Enable this to make SIB++ not hook into the Start Button.
+- RestoreAPPadding: FALSE
+  $name: Fix "All Programs" Menu Padding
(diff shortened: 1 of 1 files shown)


mods/sib-plusplus-tweaker.wh.cpp
```diff
@@ -0,0 +1,453 @@
...
+    DWORD old;
+    if(mod_settings.DHTracking == TRUE)
+    {
+        Wh_Log(L"- Disable Hottracking...");
+        uint8_t* p_trackingBG = (uint8_t*)(FindPattern("48 8B C4 48 89 50 10 48 89 48 08",(uintptr_t)g_hStartIsBackModule));
+        uint8_t* p_trackingBorder = (uint8_t*)(FindPattern("4C 8B DC 49 89 6B 18 49 89 73 20",(uintptr_t)g_hStartIsBackModule));
+
+        VirtualProtect(p_trackingBG, 1, PAGE_EXECUTE_READWRITE, &old);
+        p_trackingBG[0] = 0xC3;
+        VirtualProtect(p_trackingBG, 1, old, 0);
```
m417z: Does this work?
[From the documentation](https://learn.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-virtualprotect) for the last argument:
> If this parameter is NULL or does not point to a valid variable, the function fails.

mods/sib-plusplus-tweaker.wh.cpp
```diff
@@ -0,0 +1,453 @@
...
+        uint8_t* p_dwmComposition = (uint8_t*)(FindPattern("48 8B CB E8 43 1D 03 00 84 C0 75 4E",(uintptr_t)g_hStartIsBackModule)) - 9;
+
+        VirtualProtect(p_dwmComposition, 9, PAGE_EXECUTE_READWRITE, &old);
+        for(int i=0; i < 9; i++) p_dwmComposition[i] = 0x90;
+        VirtualProtect(p_dwmComposition, 9, old, 0);
+    }
+    if(mod_settings.MatchSevenFolders == TRUE)
+    {
+        Wh_Log(L"- Match 7 Start Menu Links...");
+        SpecialFoldersList = decltype(SpecialFoldersList)(REL((uintptr_t)(FindPattern("48 8D 35 ?? ?? ?? ?? 4C 8B F1 4C 8D 2D ?? ?? ?? ?? 33 FF",(uintptr_t)g_hStartIsBackModule)),3) - 8);
```
m417z: You're passing it to `REL(` which evaluates it twice, so a pattern will be searched twice. For reasons such as this, I'd avoid using macros like these unless absolutely necessary.

mods/sib-plusplus-tweaker.wh.cpp
```diff
@@ -0,0 +1,453 @@
...
+        }
+    }
+
+    Wh_Log(L"All done!");
+    return TRUE;
+}
+
+// The mod is being unloaded, free all allocated resources.
+void Wh_ModUninit() {
+    Wh_Log(L"Exiting SIB++ Tweaker.");
```
m417z, line 467: You might want to restart explorer here too, otherwise it will remain with the patches.

mods/sib-plusplus-tweaker.wh.cpp
```diff
@@ -0,0 +1,453 @@
...
+    MODULEINFO miSib = { 0 };
+    if (!GetModuleInformation(
+        GetCurrentProcess(),
+        g_hStartIsBackModule,
+        &miSib,
+        sizeof(MODULEINFO)
+    ))
+    {
+        Wh_Log(L"Failed to get size of StartIsBack module! Exiting.");
+        return;
```
m417z: I'd expect Wh_ModInit to return FALSE in this case.

mods/sib-plusplus-tweaker.wh.cpp
Erizur: This works fine. I made sure to not run it if it isn't properly setup. Might have to add a fallback incase the pattern isn't found by whatever reason.

m417z: It might work in practice, but if the second `VirtualProtect` call fails, the memory will always stay with `PAGE_EXECUTE_READWRITE`.

According to the documentation it doesn't work. I'd replace `VirtualProtect(p_trackingBG, 1, old, 0);` with `VirtualProtect(p_trackingBG, 1, old, &old);` here and in the other places.

Erizur: Noted. Will do.

mods/sib-plusplus-tweaker.wh.cpp
Erizur: Done.

mods/sib-plusplus-tweaker.wh.cpp
Erizur: Converted void into a BOOL to handle it. Done.

mods/sib-plusplus-tweaker.wh.cpp
Erizur: It kind of is necessary. Improved it a little.

m417z: I mean, it's unnecessary for it to be a macro. It can be a function, can't it?

Erizur: It sure can. However, I find macros easy to use if done properly. I should probably note that this specific snippet was made by someone else who I credited and got permission from.

PR #770, opened 2024-07-01
Notepad Remove Launch New App Banner 1.0.0

ItsProfessional:
# Notepad Remove Launch New App Banner
This mod removes the banner in classic notepad (which stills exists in the latest version of Windows 11 in `System32`, but can only be opened if the App Execution Alias is disabled) that tells you to launch the new app.

Screenshot

### DIFF
--- a/mods/notepad-remove-launch-new-app-banner.wh.cpp
+++ b/mods/notepad-remove-launch-new-app-banner.wh.cpp
@@ -1,0 +1,70 @@
+// ==WindhawkMod==
+// @id              notepad-remove-launch-new-app-banner
+// @name            Notepad Remove Launch New App Banner
+// @description     Removes the banner in classic notepad that tells you to launch the new app.
+// @version         1.0.0
+// @author          ItsProfessional
+// @github          https://github.com/ItsProfessional
+// @include         C:\Windows\System32\notepad.exe
+// ==/WindhawkMod==
+
+// Source code is published under The GNU General Public License v3.0.
+
+// ==WindhawkModReadme==
+/*
+# Notepad Remove Launch New App Banner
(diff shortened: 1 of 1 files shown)


mods/notepad-remove-launch-new-app-banner.wh.cpp
```diff
@@ -0,0 +1,70 @@
+// ==WindhawkMod==
+// @id              notepad-remove-launch-new-app-banner
+// @name            Notepad Remove Launch New App Banner
+// @description     Removes the banner in classic notepad that tells you to launch the new app.
+// @version         1.0.0
+// @author          ItsProfessional
+// @github          https://github.com/ItsProfessional
+// @include         C:\Windows\System32\notepad.exe
```
m417z: How about avoiding the hardcoded Windows path with:
```suggestion
// @include         %SystemRoot%\System32\notepad.exe
```

m417z: Also, what about `C:\Windows\notepad.exe`? `C:\Windows\SysWOW64\notepad.exe`?

mods/notepad-remove-launch-new-app-banner.wh.cpp
```diff
@@ -0,0 +1,71 @@
+// ==WindhawkMod==
+// @id              notepad-remove-launch-new-app-banner
+// @name            Notepad Remove Launch New App Banner
+// @description     Removes the banner in classic notepad that tells you to launch the new app.
+// @version         1.0.0
+// @author          ItsProfessional
+// @github          https://github.com/ItsProfessional
+// @include         %SystemRoot%\System32\notepad.exe
+// @include         %SystemRoot%\notepad.exe
```
m417z, line 10: Maybe you missed my edit... `C:\Windows\SysWOW64\notepad.exe` as well.

PR #775, opened 2024-07-06
Create mod "Maintain colour temperature of f.lux"

levitation:
Maintain colour temperature of f.lux. 

Keeps your preferred f.lux colour temperature settings, eliminating automatic changes. You can choose the desired colour temperature from f.lux dropdown menu, and f.lux will not change it later automatically.

### DIFF
--- a/mods/maintain-flux-colour-temperature.wh.cpp
+++ b/mods/maintain-flux-colour-temperature.wh.cpp
@@ -10,6 +10,14 @@
 // @include         flux.exe
 // ==/WindhawkMod==
 
+// Source code is published under The GNU General Public License v3.0.
+//
+// For bug reports and feature requests, please open an issue here:
+// https://github.com/levitation-opensource/my-windhawk-mods/issues
+//
+// For pull requests, development takes place here:
+// https://github.com/levitation-opensource/my-windhawk-mods/
+
 // ==WindhawkModReadme==
 /*
 # Maintain colour temperature of f.lux
--- a/mods/maintain-flux-colour-temperature.wh.cpp
+++ b/mods/maintain-flux-colour-temperature.wh.cpp
@@ -53,12 +53,6 @@ There are many posts in f.lux forum asking for this functionality since 2016. Un
 #include <windowsx.h>
 
 
-#ifndef WH_MOD
-#define WH_MOD
-#include <mods_api.h>
-#endif
-
-
 template <typename T>
 BOOL Wh_SetFunctionHookT(
     FARPROC targetFunction,
--- a/mods/maintain-flux-colour-temperature.wh.cpp
+++ b/mods/maintain-flux-colour-temperature.wh.cpp
@@ -4,7 +4,7 @@
 // @description     Keeps your preferred f.lux colour temperature settings, eliminating automatic changes
 // @version         1.0
 // @author          Roland Pihlakas
-// @github          https://github.com/levitation-opensource/
+// @github          https://github.com/levitation/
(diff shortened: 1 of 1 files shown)


mods/maintain-flux-colour-temperature.wh.cpp
```diff
@@ -0,0 +1,349 @@
...
+
+
+template <typename T>
+BOOL Wh_SetFunctionHookT(
+    FARPROC targetFunction,
+    T hookFunction,
+    T* originalFunction
+) {
+    return Wh_SetFunctionHook((void*)targetFunction, (void*)hookFunction, (void**)originalFunction);
+}
```
m417z, line 63: `Wh_SetFunctionHookT` is already part of Windhawk, and can be used by including `windhawk_utils.h`.

mods/maintain-flux-colour-temperature.wh.cpp
```diff
@@ -0,0 +1,349 @@
...
+// ==/WindhawkModSettings==
+
+
+#include <windowsx.h>
+
+
+#ifndef WH_MOD
+#define WH_MOD
+#include <mods_api.h>
+#endif
```
m417z: Why was this added?

mods/maintain-flux-colour-temperature.wh.cpp
levitation: Thank you for including in the Windhawk, this is very useful feature and I believe it will save lots of time for future developers! 

Right now this seems to be available only in most recent Windhawk versions. Since the functionality appears to be very new, some users might still not have it (for example, I do not have it in my Windhawk). I do not want to force users to upgrade Windhawk in order to use my mod. Many people might even not understand why compilation errors occur and that these errors are related to the need to upgrade Windhawk. They might just give up and think the mod does not work for them.

mods/maintain-flux-colour-temperature.wh.cpp
levitation: I am using Visual Studio to write my code. This enables intellisense checks and other conveniences. Also it enables me saving the code while it is still incomplete and not ready to compile. 

This header is included for intellisense code analysis purposes. Else I get lots of noisy error messages from intellisense. And of course it also means that I am unable to see real errors while I am developing since they are drowned in the noise. 

Including these lines does no harm since they activate only during development in editors external to Windhawk.

mods/maintain-flux-colour-temperature.wh.cpp
m417z: OK, makes sense.

mods/maintain-flux-colour-temperature.wh.cpp
m417z: You're free to use it during development, I just think that adding it can be confusing for readers of the mod code. `WH_MOD` is an internal definition of Windhawk, and its meaning may change in the future, or it can be removed. I see little risk here since even if removed, this block won't have any effect, I just think that it shouldn't be part of the submitted mod.

levitation: Ok. I removed the WH_MOD usage and mods_api.h inclusion. Maybe there is some other way to achieve the compatibility with intellisense?

m417z: Depending on your environment, perhaps you can configure "Forced Include File":
https://learn.microsoft.com/en-us/cpp/build/reference/fi-name-forced-include-file?view=msvc-170

PR #779, opened 2024-07-06
UIFILE Override v1.0

tetawaves:
Mod that adds the ability to override DirectUI XML files same way as OldNewExplorer, loading them from current shellstyle.dll.

### DIFF
--- a/mods/uifile-override.wh.cpp
+++ b/mods/uifile-override.wh.cpp
@@ -1,7 +1,7 @@
 // ==WindhawkMod==
 // @id              uifile-override
 // @name            UIFILE Override
-// @description     Override DirectUI XML files
+// @description     Override UIFILE resources
 // @version         1.0
 // @author          xalejandro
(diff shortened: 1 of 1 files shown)


mods/uifile-override.wh.cpp
```diff
@@ -0,0 +1,202 @@
...
+// @author          xalejandro
+// @github          https://github.com/tetawaves
+// @include         *
+// @compilerOptions -luxtheme -lshlwapi
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# UIFILE Override
+Mod that overrides DirectUI XML files same way as OldNewExplorer, loading them
```
m417z: Can this mod be useful for somebody who's not familiar with OldNewExplorer? I suggest to improve the description and make it clear what the mod does even for those who aren't familiar with OldNewExplorer. Adding screenshots, if relevant, can help.

PR #780, opened 2024-07-07
UXTheme hook 1.0

rounk-ctrl:
uxtheme hooking mod

### DIFF
--- a/mods/uxtheme-hook.wh.cpp
+++ b/mods/uxtheme-hook.wh.cpp
@@ -1,0 +1,169 @@
+// ==WindhawkMod==
+// @id              uxtheme-hook
+// @name            UXTheme hook
+// @description     Allows you to apply custom themes
+// @version         1.0
+// @author          rounk-ctrl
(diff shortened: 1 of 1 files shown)


mods/uxtheme-hook.wh.cpp
```diff
@@ -0,0 +1,169 @@
...
+        Element_PaintBg(This, hdc, value, pRect, pClipRect, pExcludeRect, pTargetRect);
+    }
+}
+#pragma endregion
+
+
+using SetSysColors_t = decltype(&SetSysColors);
+SetSysColors_t SetSysColors_orig;
+
+int SetSysColors_hook(int cElements, const INT *lpaElements, const COLORREF *lpaRgbValues)
```
m417z, line 99: @rounk-ctrl missed this before merging:

Calling conventions are missing, which is likely to cause a crash for 32-bit targets. Please add calling conventions or, if the mod doesn't support 32-bit, limit it to 64-bit, which will also make compilation faster. See also:

PR #781, opened 2024-07-07
Update uxtheme-hook.wh.cpp

rounk-ctrl:
fix calling conventions
directly applying .theme works

### DIFF
--- a/mods/uxtheme-hook.wh.cpp
+++ b/mods/uxtheme-hook.wh.cpp
@@ -2,7 +2,7 @@
 // @id              uxtheme-hook
 // @name            UXTheme hook
 // @description     Allows you to apply custom themes
-// @version         1.0
+// @version         1.1
 // @author          rounk-ctrl
 // @github          https://github.com/rounk-ctrl
 // @include         winlogon.exe
@@ -38,6 +38,13 @@ the colors in control panel might be weird sometimes. works fine if you leave co
 #include <vector>
 #include <uxtheme.h>
 
+#ifdef _WIN64
+#define STDCALL  __cdecl
+#define SSTDCALL L"__cdecl"
+#else
+#define STDCALL  __stdcall
+#define SSTDCALL L"__stdcall"
+#endif
 typedef unsigned __int64 QWORD;
 
 struct {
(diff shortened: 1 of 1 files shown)


mods/uxtheme-hook.wh.cpp
```diff
@@ -96,7 +104,7 @@ VOID WINAPI Element_PaintBgHook(class Element* This, HDC hdc, class Value* value
 using SetSysColors_t = decltype(&SetSysColors);
 SetSysColors_t SetSysColors_orig;
 
-int SetSysColors_hook(int cElements, const INT *lpaElements, const COLORREF *lpaRgbValues)
+int STDCALL SetSysColors_hook(int cElements, const INT *lpaElements, const COLORREF *lpaRgbValues)
```
m417z: It should be WINAPI. BTW it would be best if you could test it on 32-bit and make sure that it works and doesn't crash.

mods/uxtheme-hook.wh.cpp
```diff
@@ -132,13 +140,13 @@ BOOL Wh_ModInit() {
     }
     
     // for logonui
-    Wh_SetFunctionHook((void*)SetSysColors, (void*)SetSysColors_hook,
+    WindhawkUtils::Wh_SetFunctionHookT((void*)SetSysColors, (void*)SetSysColors_hook,
```
m417z: The point of `Wh_SetFunctionHookT` is that it verifies the types if you don't use casts. If you use casts, there's no point in using it.

BTW same for `WindhawkUtils::SYMBOL_HOOK`. You can remove the casts and if it doesn't compile, you have a problem with the types.

mods/uxtheme-hook.wh.cpp
rounk-ctrl: but most of the mods, currently on windhawk, use this convention
and WINAPI does the same thing

mods/uxtheme-hook.wh.cpp
rounk-ctrl: ohh, ill remove the casts

mods/uxtheme-hook.wh.cpp
m417z: Yes, `WINAPI` is defined as `__stdcall` so they're the same, but it's more correct to use `WINAPI` for WinAPI functions such as `SetSysColors`. Which mod uses something other than `WINAPI` for WinAPI hooks?

mods/uxtheme-hook.wh.cpp
m417z: Just remove the casts.

mods/uxtheme-hook.wh.cpp
rounk-ctrl: Oh okay, ill change it to WINAPI.

rounk-ctrl:
not yet, ill add which all version of windows this mod works on

PR #786, opened 2024-07-09
Create disable-edge-window-tab-manager

ilyfairy:


m417z:
It seems to be similar to [Prevent Edge from lagging your Taskbar process](https://windhawk.net/mods/edge-window-tab-manager-block) which was just merged yesterday.

ilyfairy:
Okay, didn't notice.

PR #789, opened 2024-07-11
Create mod "Classic taskbar background fix"

levitation:
Fixes Taskbar **background** in classic theme by replacing **black background** with a classic button face colour. 

NB! This mod does not fix the buttons' colour, for that there are other mods available.

Before:

Before

After:

After

### DIFF
--- a/mods/classic-taskbar-background-fix.wh.cpp
+++ b/mods/classic-taskbar-background-fix.wh.cpp
@@ -26,7 +26,7 @@ Fixes Taskbar background in classic theme by replacing black background with a c
 
 Install this mod if you have issues with black background around taskbar buttons and tray icons, like illustrated in the picture below. This mod does not fix the buttons' colour, for that purpose there are other mods available.
 
-This mod can be considered as an alternative to installing the @valinet's modified OpenShell StartMenuDLL.dll.
+This mod can be considered as an alternative to installing the @valinet's modified version of OpenShell StartMenuDLL.dll.
 
 Before:
 
@@ -35,6 +35,15 @@ Before:
 After:
 
 ![After](https://raw.githubusercontent.com/levitation-opensource/my-windhawk-mods/main/screenshots/after-classic-taskbar-background-fix.png)
+
+
+## Mod configuration
+
+If you use some other mod for rendering buttons than 'Classic Taskbar 3D buttons Lite' and your Taskbar is sometimes extremely full (with small buttons) and you want to preserve the full contrast of the button borders, then go to the settings of the current mod and under "Compatibility with classic Taskbar buttons mods" choose "No compatibility adjustments". Otherwise you can keep this setting at its default value - "Enhance compatibility with 'Classic Taskbar 3D buttons Lite'".
+
+
+## Acknowledgements
+I would like to thank @Anixx for testing the mod during its development and illustrating the issues found.
 */
 // ==/WindhawkModReadme==
 
@@ -47,6 +56,11 @@ After:
   - highlightOnHover: Yes, and use highlight colour on mouse over
   - blackOnHover: Yes, and use black colour on mouse over
   - no: No
+- CompatWithTaskbarButtonsMods: classic-taskbar-buttons-lite
+  $name: Compatibility with classic Taskbar buttons mods
+  $options:
+  - classic-taskbar-buttons-lite: Enhance compatibility with 'Classic Taskbar 3D buttons Lite'
+  - no: No compatibility adjustments
 */
 // ==/WindhawkModSettings==
 
@@ -57,10 +71,24 @@ After:
 #include <uxtheme.h>
 
 
+//clang compiler does not have these macros defined. Definitions taken from <minwindef.h>
+#ifndef max
+#define max(a,b)            (((a) > (b)) ? (a) : (b))
+#endif
+#ifndef min
+#define min(a,b)            (((a) < (b)) ? (a) : (b))
+#endif
+
+
 enum class RepaintDesktopButtonConfig {
     yes,
     highlightOnHover,
     blackOnHover,
+    no
+};
+
+enum class CompatWithTaskbarButtonsModsConfig {
+    classicTaskbarButtonsLite,
     no
 };
 
@@ -98,12 +126,15 @@ std::mutex g_hdcMapMutex;
 std::map<HDC, MemDCInfo> g_hdcMap;      //using map not unordered_map since the latter becomes slow when doing many insertions and removals. Also it is expected that the current map will contain only a few elements at a time
 
 RepaintDesktopButtonConfig g_repaintDesktopButtonConfig;
+CompatWithTaskbarButtonsModsConfig g_compatWithTaskbarButtonsModsConfig;
 
 
 using BeginPaint_t = decltype(&BeginPaint);
 BeginPaint_t pOriginalBeginPaint;
 using EndPaint_t = decltype(&EndPaint);
 EndPaint_t pOriginalEndPaint;
+using DrawFrameControl_t = decltype(&DrawFrameControl);
+DrawFrameControl_t pOriginalDrawFrameControl;
 //using DrawThemeParentBackground_t = decltype(&DrawThemeParentBackground);     //this declaration is a bit different in Windhawk headers than in Visual Studio, so using manual declaration instead to avoid both intellisense and compilation errors
 typedef HRESULT(WINAPI* DrawThemeParentBackground_t)(HWND, HDC, const RECT*);
 DrawThemeParentBackground_t pOriginalDrawThemeParentBackground;
@@ -123,6 +154,15 @@ RepaintDesktopButtonConfig DesktopButtonConfigFromString(PCWSTR string) {
     }
     else {
         return RepaintDesktopButtonConfig::yes;
+    }
+}
+
+CompatWithTaskbarButtonsModsConfig CompatWithTaskbarButtonsModsConfigFromString(PCWSTR string) {
+    if (wcscmp(string, L"no") == 0) {
+        return CompatWithTaskbarButtonsModsConfig::no;
+    }
+    else {
+        return CompatWithTaskbarButtonsModsConfig::classicTaskbarButtonsLite;
     }
 }
 
@@ -215,7 +255,7 @@ void ConditionalFillRect(HDC hdc, const RECT& rect, COLORREF oldColor, COLORREF 
         return;
     }
 
-    //Create a compatible DC and bitmap. Even though hdc is already a memDC, we need to create one more memDC, since we need to get access to pixels using GetDIBits and SetDIBits, which does not support providing x coordinates.
+    //Create a compatible DC and bitmap. Even though hdc is already a memDC, we need to create one more memDC, since we need to get access to pixels using ExtFloodFill which does not support providing top-left coordinates, or alternatively, using GetDIBits and SetDIBits, which does not support providing left-right coordinates.
     HDC memDC = CreateCompatibleDC(hdc);
     if (!memDC) {
         Wh_Log(L"CreateCompatibleDC failed");
@@ -261,7 +301,7 @@ void ConditionalFillRect(HDC hdc, const RECT& rect, COLORREF oldColor, COLORREF 
                             }
                             else {
                                 //blit the modified content back to hdc
-                                //TODO: try to blit directly to original hdc, non to memDC from BeginPaint?
+                                //TODO: try to blit directly to original hdc, not to memDC from BeginPaint?
                                 if (!BitBlt(hdc, rect.left, rect.top, rect.right - rect.left, rect.bottom - rect.top, memDC, 0, 0, SRCCOPY))
                                     Wh_Log(L"BitBlt to hdc failed");
                             }
@@ -304,7 +344,7 @@ void ConditionalFillRect(HDC hdc, const RECT& rect, COLORREF oldColor, COLORREF 
                         }
                         else {
                             //blit the modified content back to hdc
-                            //TODO: try to blit directly to original hdc, non to memDC from BeginPaint?
+                            //TODO: try to blit directly to original hdc, not to memDC from BeginPaint?
                             if (!BitBlt(hdc, rect.left, rect.top, rect.right - rect.left, rect.bottom - rect.top, memDC, 0, 0, SRCCOPY))
                                 Wh_Log(L"BitBlt to hdc failed");
                         }
@@ -321,6 +361,7 @@ void ConditionalFillRect(HDC hdc, const RECT& rect, COLORREF oldColor, COLORREF 
     }
 }
 
+//BeginPaint/EndPaint hook currently fixes the background around taskbar buttons and Start button
 HDC WINAPI BeginPaintHook(
     IN  HWND          hWnd,
     OUT LPPAINTSTRUCT lpPaint
@@ -395,6 +436,7 @@ HDC WINAPI BeginPaintHook(
     return hdc;
 }
 
+//BeginPaint/EndPaint hook currently fixes the background around taskbar buttons and Start button
 BOOL WINAPI EndPaintHook(
     IN HWND                 hWnd,
     IN const PAINTSTRUCT*   lpPaint
@@ -454,6 +496,57 @@ BOOL WINAPI EndPaintHook(
     return pOriginalEndPaint(hWnd, lpPaint);
 }
 
+//this hook is needed in case the buttons are offset in their hdc so that there are spaces in between buttons and some background appears between buttons
+BOOL WINAPI DrawFrameControlHook(
+    IN HDC    hdc,
+    IN LPRECT lprc,
+    IN UINT   uType,
+    IN UINT   uState
+) {
+    int colorIndex = COLOR_3DFACE;
+    if (
+        g_hwndTaskbar   //is the current process the taskbar process?
+        && uType == DFC_BUTTON
+        && ((uState & DFCS_BUTTONPUSH) != 0)
+        && lprc
+        && g_compatWithTaskbarButtonsModsConfig == CompatWithTaskbarButtonsModsConfig::classicTaskbarButtonsLite
+        //cannot use WindowFromDC and WindowNeedsBackgroundRepaint check here since WindowFromDC will fail for some reason
+    ) {
+        HBRUSH brush = GetSysColorBrush(colorIndex);
+        if (!brush) {            //verify that the brush is supported by the current system
+            Wh_Log(L"GetSysColorBrush failed - is the colour supported by current OS?");
+        }
+        else {                
+            COLORREF buttonFace = GetSysColor(colorIndex);
+
+            RECT rect;
+            if (!GetClipBox(hdc, &rect)) {
+                SetRectEmpty(&rect);
+            }
+
+            //Mitigation for case @Anixx classic-taskbar-buttons-lite mod is being used, which changes the left and right offsets of taskbar buttons before they are rendered. Without the mitigation there would appear black OR dark lines on left and right side of the button.
+            //
+            //This mitigation will not usually interfere with the operation of mod classic-taskbar-buttons-lite-vs-without-spacing by @OrthodoxWin32. Normally the buttons will have still one pixel of space between them even with that mod, and we are here extending the rect size also only by one pixel. If the taskbar is extremely full, then the one pixel space between the buttons vanishes, but the result of repainting does not seem too visible in this case either. Unfortunately selective repaint would not be trivial in this case.
+            // 
+            //Postprocessing the result of DrawFrameControl by conditional colour replacement fill method would not work reliably here since in certain conditions the sides of the offset buttons will not appear exactly black, but dark instead (for example 25-25-25). Even though the offset sides are outside of the lprc, the DrawFrameControl can still fill these offset sides with that dark colour if the original button background (before offsetting by classic-taskbar-buttons-lite) is left here unpainted before the call to DrawFrameControl.
+
+            RECT fillRect = *lprc;
+            fillRect.left = max(lprc->left - 1, rect.left);
+            fillRect.right = min(lprc->right + 1, rect.right);
+
+            FillRect(hdc, &fillRect, brush);
+        }
+    }
+
+    return pOriginalDrawFrameControl(
+        hdc,
+        lprc,
+        uType,
+        uState
+    );
+}
+
+//this hook currently fixes regions near tray area
 bool DrawThemeParentBackgroundInternal(HWND hwnd, HDC hdc) {
 
     RECT rect;
@@ -469,28 +562,22 @@ bool DrawThemeParentBackgroundInternal(HWND hwnd, HDC hdc) {
         if (colorIndex == blackColorIndex) {  //Repaint the area as black again. Used for "show desktop" button if the mod settings say so.
 
             HBRUSH brush = CreateSolidBrush(black);
-            if (brush) {            //verify that the brush is supported by the current system
-                HGDIOBJ holdbrush = SelectObject(hdc, brush);
-                if (holdbrush && holdbrush != HGDI_ERROR) {
-
-                    FillRect(hdc, &rect, brush);
-                    SelectObject(hdc, holdbrush);
-
-                    return true;
-                }
+            if (!brush) {
+                Wh_Log(L"CreateSolidBrush failed");
+            }
+            else {
+                FillRect(hdc, &rect, brush);
+                return true;
             }
         }
         else {
             HBRUSH brush = GetSysColorBrush(colorIndex);
-            if (brush) {            //verify that the brush is supported by the current system
-                HGDIOBJ holdbrush = SelectObject(hdc, brush);
-                if (holdbrush && holdbrush != HGDI_ERROR) {
-
-                    FillRect(hdc, &rect, brush);
-                    SelectObject(hdc, holdbrush);  // select old brush
-
-                    return true;
-                }
+            if (!brush) {            //verify that the brush is supported by the current system
+                Wh_Log(L"GetSysColorBrush failed - is the colour supported by current OS?");
+            }
+            else {
+                FillRect(hdc, &rect, brush);
+                return true;
             }
         }
     }
@@ -630,11 +717,16 @@ void Wh_ModAfterInit(void) {
 
 void LoadSettings() {
 
-    //config option to keep "show desktop" button black
-    PCWSTR configString = Wh_GetStringSetting(L"RepaintDesktopButton");
-    g_repaintDesktopButtonConfig = DesktopButtonConfigFromString(configString);
+    PCWSTR configString;
 
+    //config option to keep "show desktop" button black
+    configString = Wh_GetStringSetting(L"RepaintDesktopButton");
+    g_repaintDesktopButtonConfig = DesktopButtonConfigFromString(configString);
     Wh_FreeStringSetting(configString);     
+
+    configString = Wh_GetStringSetting(L"CompatWithTaskbarButtonsMods");
+    g_compatWithTaskbarButtonsModsConfig = CompatWithTaskbarButtonsModsConfigFromString(configString);
+    Wh_FreeStringSetting(configString);
 }
 
 BOOL Wh_ModInit() {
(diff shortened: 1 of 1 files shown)


mods/classic-taskbar-background-fix.wh.cpp
```diff
@@ -0,0 +1,368 @@
...
+
+After:
+
+![After](https://raw.githubusercontent.com/levitation-opensource/my-windhawk-mods/main/screenshots/after-classic-taskbar-background-fix.png)
+*/
+// ==/WindhawkModReadme==
+
+// ==WindhawkModSettings==
+/*
+- RepaintDesktopButton: yes
```
m417z, line 57: How about using a boolean here?

mods/classic-taskbar-background-fix.wh.cpp
```diff
@@ -0,0 +1,368 @@
...
+        Wh_Log(L"Not a taskbar process, not hooking");
+        *abort = true;
+        return false;
+    }
+
+
+    Wh_Log(L"Initialising hooks...");
+
+
+    HMODULE hUser32 = GetModuleHandle(L"user32.dll"); 	//when we reach here then we can be sure that user32.dll has been already loaded
```
m417z: What you did works, but I'm pretty sure user32.dll is available from the very beginning, so you can just set the hooks in `Wh_ModInit`. Avoiding `Wh_ApplyHookOperations` is also better for performance.

mods/classic-taskbar-background-fix.wh.cpp
```diff
@@ -0,0 +1,368 @@
...
+            /*bInitialState = */FALSE,              // initial state is nonsignaled
+            /*lpName = */NULL						// object name
+        );
+
+        if (!g_initThreadStopSignal) {
+            Wh_Log(L"CreateEvent failed");
+            return FALSE;
+        }
+
+        g_initThread = CreateThread(
```
m417z, line 995: That's fine, but in my personal opinion this approach is tricky to maintain and easy to get wrong, e.g. look at the complex `Wh_ModUninit` implementation. I'd prefer to just hook `CreateWindowEx` and catch the taskbar creation there. In any case, it's just an FYI/tip/opinion, you can keep it as is.

mods/classic-taskbar-background-fix.wh.cpp
```diff
@@ -0,0 +1,368 @@
...
+// https://github.com/levitation-opensource/my-windhawk-mods/issues
+//
+// For pull requests, development takes place here:
+// https://github.com/levitation-opensource/my-windhawk-mods/
+
+// ==WindhawkModReadme==
+/*
+# Classic Taskbar background fix
+
+Fixes Taskbar background in classic theme by replacing black background with a classic button face colour.
```
m417z: Can you add more details, e.g. which WIndows versions are supported? Which method of classic themes? A classic theme is not a native Windows feature anymore, right?

mods/classic-taskbar-background-fix.wh.cpp
levitation: Tried the CreateWindowEx version now. 

It works generally nicely, but there is one bug which is maybe some interaction between Windhawk and ExplorerPatcher. 

The issue manifests if I execute the following sequence: 
1. kill explorer.exe (or just restart it in step 3)
2. load the mod
3. start explorer.exe (or just restart it here)
4. unload the mod
Now explorer.exe either hangs or crashes, tried number of times, happens always for me. 

When I do not kill/restart explorer.exe, in other words the mod starts after explorer.exe, then the issue does not manifest during mod unload. Tried number of times as well.

I tested with another mod that uses CreateWindowEx hooking on explorer.exe and same pattern described above happens with that mod as well. This other mod is named classic-taskbar-context-menu and is created by another author. When creating my mod, I did not copy their code, so I do not think that my or their code have same bug. More likely it is something related to hooked function handling either in Windhawk or ExplorerPatcher.

mods/classic-taskbar-background-fix.wh.cpp
levitation: There is some chance I will add more options in the future. If I add more options in the future, would it be compatible with users who have already a boolean setting in registry?

mods/classic-taskbar-background-fix.wh.cpp
levitation: I tested it both on Win10 and Win11. For classic theme I use Windhawk mod named classic-theme-enable. 

Classic theme still seems to be native to Windows. As I understand it, essentially this "classic-theme-enable" mod disables themes, and this reactivates the default behaviour of Windows, which is classic theme. 

There are a few registry parameters that need to be adjusted for nicer UI appearance. Also one might want to change colours. I use a program named "Desktop Themes" v1.87 for that. It is available here https://archive.org/details/desktop_themes_v187_ZIP

There are few programs that have minor visual glitches. Various Windhawk mods deal with this and so these are solved as well. 

Most problematic program is Taskbar. Fortunately there are a number of Windhawk mods, each fixing a different problem of classic theme in Taskbar. Additionally I used ExplorerPatcher and Open Shell Menu. The only problem that still manifested for me was this background issue. 

The only totally problematic program uncompatible with classic theme is Task Manager. I use Process Hacker instead. Process Hacker can be configured to handle Ctrl-Alt-Del as well.

All other programs have been running fine, have been using classic theme for about a month. Right now my systems look entirely classic (except for programs that have their own built in themes). I am quite intensive user using many different programs.

I guess I could later write some document about my setup so that others could conveniently reproduce it.

mods/classic-taskbar-background-fix.wh.cpp
levitation: I guess this question depends mostly on how to solve the next question. Right now there is a small issue with hooking CreateWindowEx, described below.

mods/classic-taskbar-background-fix.wh.cpp
levitation: Maybe until this gets resolved, lets use the thread method. And later when no issues with CreateWindowEx manifest anymore, then I can push a new version with CreateWindowEx, which I have already written as well.

mods/classic-taskbar-background-fix.wh.cpp
levitation: Alternatively, yes, I could set the hook in Wh_ModInit. And then use the thread for setting a flag indicating whether the hook function should actually do anything besides passing on to the original function.

mods/classic-taskbar-background-fix.wh.cpp
m417z: A boolean is saved as a number, 0 or 1, so yes, you should be able to switch to a number or a string and use the 0 and 1 values. I didn't test this specific case, so I suggest to try it and make sure that it works. But I see that you added more options, so never mind.

mods/classic-taskbar-background-fix.wh.cpp
m417z: I assume Win11 is only supported with the Win10 taskbar, e.g. with ExplorerPatcher? Why not add it, and all the explanation that you've written here, even as-is, to the readme?

Some of the feedback I rightfully got for Windhawk is that it can be difficult to know which mods are relevant for which Windows versions. While I can add some tag to specify the relevant Windows versions, it's sometimes more complicated than that, e.g. with additional requirements such as having ExplorerPatcher installed. Windhawk is less helpful if this kind of information has to be deduced with trial and error or by asking in forums and chats.

mods/classic-taskbar-background-fix.wh.cpp
m417z: Can you post a crash dump, along with that mod version? You can either post a Drive (or similar service) link, or [email it to me](https://ramensoftware.com/contact). I didn't encounter such an issue with my mods, so I suspect there's something you got wrong on your end.

You should be able to get a crash dump with the following steps:
* Open regedit
* Go to: `HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\Windows Error Reporting`
* Create a key named `LocalDumps`
* Create a DWORD value named `DumpType` with value `2`
* Trigger the crash
* Go to the `%LocalAppData%\CrashDumps` folder, you should see a dump file in there

> Maybe until this gets resolved, lets use the thread method. And later when no issues with CreateWindowEx manifest anymore, then I can push a new version with CreateWindowEx, which I have already written as well.

Sure, we can do that if we don't find a quick solution or if you don't have the time/desire to work on it, up to you, let me know.

mods/classic-taskbar-background-fix.wh.cpp
Anixx: We definitely need tags, but more refined than just windows versions. Some suggestions: windws shell, windows styles, DWM, cassic theme, gaming, taskbar, file manager...

m417z: Tags are a different feature. Even if you have tags that say "taskbar" or "classic theme", it doesn't tell you what are the requirements, like Win10, Win11, ExplorerPatcher, etc.

Anixx:
Your mod creates font corruption for me: 
изображение

Anixx:
Your mod fails to fix the Explorer Patrcher's built-in Win10-sty taskbar:
изображение

Anixx:
Even worse: the background is even blue with default OpenShell install and your patch:
изображение

Anixx:
Also, all the tray items are flickering on hover.

levitation:
> 
> Everything flicker on hover, including the taskbar buttons and tray items. Taskbar buttons have unstable on hover black frames.

For flickering I have an update which I will send soon. It seems the flickering manifests mostly in console sessions and rarely or not at all in RDP sessions.

Regarding other issues I guess it depends on the combination of other software and mods you use. To get started, I would ask, do you need this mod to begin with? Did you have black background around the taskbar buttons and tray icons, as in the first picture of my two screenshots in the pull request?

levitation:
> Your mod fails to fix the Explorer Patrcher's built-in Win10-style taskbar: изображение

You picture shows black buttons. Note that my mod is for fixing the background, not buttons. For buttons there are other mods available in Windhawk marketplace. For example, I use "Classic Taskbar 3D buttons with extended compatibility" ("classic-taskbar-buttons-lite-vs-without-spacing"). I do not intend to compete with existing functionalities, so I fixed only what was missing.

Anixx:
The black background issue manifests for everyone who is using the Classic Theme. Our current mitigation for this issue is using a patched version or StartMenuDLL.dll file for OpenShell https://github.com/valinet/ExplorerPatcher/files/13359466/StartMenuDLL.zip So, having Windhawk mod to fix this could be handier. The blue background I experience with original (unmodified) OpnShell DLL. The corrupted fonts, corrupted tray icons and flicker effects after enabling your mod I experience both with modified DLL and without.

levitation:
> The black background issue manifests for everyone who is using the Classic Theme. Our current mitigation for this issue is using a patched version or StartMenuDLL.dll file for OpenShell https://github.com/valinet/ExplorerPatcher/files/13359466/StartMenuDLL.zip So, having Windhawk mod to fix this could be handier. The blue background I experience with original (unmodified) OpnShell DLL. The corrupted fonts, corrupted tray icons and flicker effects after enabling your mod I experience both with modified DLL and without.

Thank you for the details! These look very useful. Glad to know that there could be use to this mod.

The flickering should be fixed now. 

The fonts rendering and tray icons problem probably is still present, but if it somehow disappeared then please let me know.

Regarding blue background, I will try to investigate that. I have not encountered it, but perhaps my OpenShell settings are different, so I will play with my OpenShell settings and see if I can reproduce the blue colour. 

I am not using a patched StartMenuDLL.dll, did not know it exists. But will take a look at that too. I hope it's code is also available. Maybe the code for that dll is better than the one I have developed here, then I will try to adapt necessary bits from that dll's source into a Windhawk mod.

Anixx:
Here is the tutorial on using Classic theme:
https://github.com/valinet/ExplorerPatcher/discussions/167

The comments below by Valinet include the description on how to patch the OpenShell DLL. You may have a look into it and re-implement your mod in a less glitchy way.

Anixx:
I have tried your last version. The flickering effect is indeed gone, but the font corruption remains, not only on taskbar but also in the flyer popups:

изображение

levitation:
Ok, I think I solved most of the issues.

* The blue background issue Anixx encountered is probably still unresolved.

* With the dark theme, my mod alone works ok for fixing the taskbar background and fonts are not corrupted as long as the fonts are not black. For adjusting the taskbar buttons, not background, one might still need the other Windhawk mods specialised on that purpose.
* With the bright theme, it is necessary to work together with OpenShell. To avoid using patched OpenShell.dll, I created a few additional hooks into Windows API, which trigger somewhat similar behaviour as the patched OpenShell dll code would do (not exactly same, but sufficiently). This enables getting desired look on the taskbar IF certain OpenShell settings are set to certain values.

With the bright theme, and the new hooks in the mod, and correct OpenShell settings, the taskbar still had a black bottom area in case the taskbar was multi-row. To handle that without corrupting the fonts, I re-enabled the repaint procedure in my mod and modified the repaint procedure a bit. At the same time, that modified repaint procedure logic would not work alone without OpenShell on the entire bright-theme taskbar though. So for bright theme all the components are needed:
* * OpenShell installation
* * Correct OpenShell settings
* * Mod hooking a few Windows API functions that OpenShell uses - this fixes the background around start button and in the tray area
* * Mod hooks for repainting the bottom part of the taskbar buttons area background, using a modified procedure to exclude buttons themselves - this fixes the background around taskbar buttons without corrupting the fonts
** To ensure correct colours on the buttons, other Windhawk mods are available.

If I could somehow hook into OpenShell dll, then the user would not need to make their OpenShell settings correct since I could force the correct settings from the mod side.

I will push my updated code later.

levitation:
Found a way to make the mod work independently from OpenShell.

The "Customize Taskbar" setting in OpenShell needs to be turned off in case unpatched OpenShell is being used. It would be unnecessary and would interfere with this mod.

I disabled the "show desktop" button rendering settings under mod settings temporarily. If everything else works as intended then I will restore this functionality.

For me there appeared a small issue during testing which manifests both with my mod and also with patched OpenShell dll - when I am using bright Windows theme, various tray icons are still rendered with white colour and therefore become almost invisible. Since this issue manifests with OpenShell dll as well, then I guess this is related to my general Windows settings and not a problem with this mod.

If anybody else has also problems with tray icons having white colour then I can probably implement a repaint method for these icons that makes the white icons black. But if that icon colour problem happens only for me then I will not bother since I am actually intending to use dark theme.

Anixx:
Hello! I have tested the last version.
* With the patched OpenShell DLL - your mod mostly have no effect, the corrupted fonts are gone (they appear for a fraction of second when the taskbar gets shown, then the fonts become normal)
* With original OpenShell DLL - the blue background is gone but the Start button and task buttons have corrupted borders:

изображение

levitation:
> Hello! I have tested the last version.

@Anixx Thank you a lot for testing! 

Happy to hear that the text and background were okay now.

Hopefully with the latest commit I have fixed the taskbar buttons border issue now as well.

Anixx:
There are still white lines to the left and to the right of the Start button and black lines between the taskbar buttons:
изображение

levitation:
@Anixx Thank you for the picture! 

Can you please check the mod settings tab: Is "Compatibility with classic Taskbar buttons mods" set to "Enhance compatibility with 'Classic Taskbar 3D buttons Lite'"? 

The phenomena that your picture illustrates is the one which this setting is supposed to fix.

"Enhance compatibility with 'Classic Taskbar 3D buttons Lite'" should be normally the default setting, but maybe the setting value got mixed somehow when I added the setting only after the mod was installed, and you need to adjust the setting manually.

Anixx:
I have just tested your mod with Explorer Patcher's own taskbar. It is not black anymore, but white. The fonts and tray icons are corrupted like it was prevuously with standard taskbar. Initially it appears in desired, button color but then becomes white, including the quick launch area.
изображение

Anixx:
> Can you please check the mod settings tab: Is "Compatibility with classic Taskbar buttons mods" set to "Enhance compatibility with 'Classic Taskbar 3D buttons Lite'"?

Yes, it is enabled.

Anixx:
I tested your mod with the mod "3D buttons with extended compatibility" and the result is the same: black vertical lines between taskbuttons and white vertical lines around the Start button.

If no 3D mod is active, the lines also present but there are other glitches as well: text is not visible on the task buttons except the current task and some of the buttons are black:

изображение

Anixx:
I also noticed that the vertical black lines do not appear in the "Lite" mod if the taskbar has two rows of buttons:

изображение

levitation:
@Thank you so much @Anixx for your patience and initiative in testing! 

This two row Taskbar versus one row Taskbar finding you made was very helpful. I was able to reproduce the lines.

It took me some time, but now couple of additional small tweaks came to my mind, which hopefully solved both the white lines around Start Button, and black lines around Taskbar buttons.

Note, an enabled "classic taskbar buttons" mod installation is still needed alongside my "classic taskbar background" mod, and that is how I prefer to keep it.

Anixx:
Indeed, now it works with the unmodified StartMenuDLL.dll as expected!

But only with a 3D taskbar buttons mod. Without this mod, the issues are as before.
You said that you do not plan to fix it but take into account the following:

* Flat taskbar buttons are the default mode. I think, you should better empathize in the mod description that a 3D button mod is required.
* When one enables the 3D taskbar buttons, some functionality gets lost (the progressbar indication on taskbar buttons), so there are many legitimate users of Classic theme who would need the flat taskbar mode.
* In Windows 11 24H2 (the upcoming Windows release) the Win10-style taskbar is broken and in later builds it is removed completely. So, the only way to go would be to use the Win10-like taskbar shipped with EP. But the 3D taskbar buttons mod does not work with EP's taskbar. As such, as of now, the EP taskbar with your mod looks like Win10 taskbar without the 3D buttons mod and has the same issues. I think, you should clarify that your mod does not support the EP's own taskbar.

Anixx:
In the latest version the taskbar buttons flicker and have black frames when the taskbar is oriented vertically.

levitation:
> In the latest version the taskbar buttons flicker and have black frames on hover when the taskbar is oriented vertically.

Would you please be able to share a screenshot of the black frames? I am unable to reproduce these issues at the moment. I guess it is not possible to screenshot the flicker part, so with regards to that I will try to figure something out on my own.

Anixx:
изображение
As you can see, some have black frame from the left, some from the right, some from both sides. The frames appear on hover.

levitation:
@Anixx Thank you for the picture! I think I was able to solve the black lines on the sides of buttons in case of vertical Taskbar. Regarding flickering, it is probably still present, but maybe it is gone as a side effect to the other changes. Please let me know how it looks now!

Anixx:
Now there are lines on the left on all buttons, they disappear on changing of number of buttons but re-appear on hover:
изображение

Anixx:
There is 1-time flickering when changing the taskbar position. Also, the verticlal lines around buttons change state on hover, which looks like flickering. Also, if you move the buttons, this appears:

изображение

Anixx:
There is 1-time flickering when changing the taskbar position. Also, the verticlal lines around buttons change state on hover, which looks like flickering. Also, if you move the buttons, this appears:

изображение

and:

levitation:
@Anixx Thank you for the feedback! Pushed one more adjustment to the vertical lines that appeared in your machine on hover. I hope that adjustments fixes it, but I was unable to replicate it myself, so I am not sure.

Regarding the black background on button drag, I did not fix that yet and it is possible I will not be able to fix it. As I understand, the large black boxes appear only when dragging the buttons into the empty area. I assume normally people do not do that, since it has no benefit. 

I see that there may occasionally appear also smaller black boxes while dragging the buttons into the middle of other buttons. But this seems relatively minor issue and right now anything I tried did not help.

Anixx:
Yes, the vertical lines are fixed now. The black boxes do not appear when the taskbar has two rows.

levitation:
> Yes, the vertical lines are fixed now. The black boxes do not appear when the taskbar has two rows.

Thank you once again! 

Indeed, with one row I can replicate the black boxes while dragging buttons on horizontal taskbar. But unfortunately it appears that any additional tricks I tested with do not fix that. Fortunately, it does not seem very important issue, so I will leave it like that for time being.

I will work on the Readme text based on the recommendations you made in your earlier message https://github.com/ramensoftware/windhawk-mods/pull/789#issuecomment-2267650899

Anixx:
Interestingly, the black boxes do not appear when moving the buttons up un vertical taskbar:
изображение

OrthodoxWindows:
I just tested your mod. Combine with one or the other mod to obtain 3D buttons, the black area bug when moving is not very annoying, because as soon as the icons are replaced the problem disappears. But with the original flat icons (without additional mod) it is really unusable, whereas the Open-Shell version encounters no problem : 
Jy7HeFcd6m

Additionally, I noticed a problem with the "Classic Taskbar 3D buttons with extended compatibility" mod, the icons present in the same group are close to each other : 
3cB1ob5Lj5
This is not the case with Open-Shell:
uPuqTjGNkY

This does not happen with the "Classic Taskbar 3D buttons Lite" mod, except that I need to use "Classic Taskbar 3D buttons with extended compatibility", because only unlike "Classic Taskbar 3D buttons Lite" it is compatible with icons without labels.

levitation:
@OrthodoxWindows Yes, this background mod is intended to be used together with buttons mod. That way people can choose which buttons mod to use. 

Regarding buttons appearing too close - Under mod Settings, there is option "Enhance compatibility with 'Classic Taskbar 3D buttons Lite'" - if you turn it off (set to "No compatibility adjustments"), then the spaces between nearby buttons should become better in case you use "Classic Taskbar 3D buttons with extended compatibility" buttons mod. Does that alternate option change the small buttons appearance better for you?

I myself am also using "Classic Taskbar 3D buttons with extended compatibility" mod. The default option is currently set to "Enhance compatibility with 'Classic Taskbar 3D buttons Lite'" since without that setting the buttons with that buttons mod will look even worse (there would be black lines in certain cases). 

Would you prefer me to rename the alternate option from "No compatibility adjustments" to "Enhance compatibility with 'Classic Taskbar 3D buttons with extended compatibility'" or do you have other suggestions for the alternate setting title?

In the future I could maybe detect the presence of "Classic Taskbar 3D buttons Lite" mod automatically, but right now it seemed too complex, so I used user setting instead.

OrthodoxWindows:
> Regarding buttons appearing too close - Under mod Settings, there is option "Enhance compatibility with 'Classic Taskbar 3D buttons Lite'" - if you turn it off (set to "No compatibility adjustments"), then the spaces between nearby buttons should become better in case you use "Classic Taskbar 3D buttons with extended compatibility" buttons mod. Does that alternate option change the small buttons appearance to look better for you?
> 
> I myself am also using "Classic Taskbar 3D buttons with extended compatibility" mod. Yet the default option is currently set to "Enhance compatibility with 'Classic Taskbar 3D buttons Lite'" since without that setting the buttons with that buttons mod will look even worse (there would be black lines in certain cases).
> 
> Would you prefer me to rename the alternate option from "No compatibility adjustments" to "Enhance compatibility with 'Classic Taskbar 3D buttons with extended compatibility'" or do you have other suggestions for the alternate setting title? In general I expect the "No compatibility adjustments" setting to work for most buttons mods, while "Classic Taskbar 3D buttons Lite" is the special case that needs its own compatibility setting.
> 
> In the future I could maybe detect the presence of "Classic Taskbar 3D buttons Lite" mod automatically, but right now it seemed too complex, so I created a mod setting instead.

Thanks you, I didn't see the parameter exist. Now the problem no longer arises.

OrthodoxWindows:
> @OrthodoxWindows Happy to hear that the black areas during button move do not disturb you much.
> 
> Yes, this background mod is intended to be used together with a buttons mod. It does not render buttons on its own. That way people can choose which buttons mod to use.
> 
> If there would exist some flat buttons mod (?), then they could even use that. But yes, they need to use some buttons mod.
> 
> If the expectation for the user to use a separate buttons mod is too confusing, I could merge your and my mods code into one and create one more option in the mod settings which would then enable this merged code and would render the buttons as well, without the user having to install a separate buttons mod. Curious, what do you think?
>
My point was that with Open-Shell there is no need to use a button mod.

levitation:
> * When one enables the 3D taskbar buttons, some functionality gets lost (the progressbar indication on taskbar buttons), so there are many legitimate users of Classic theme who would need the flat taskbar mode.
> * In Windows 11 24H2 (the upcoming Windows release) the Win10-style taskbar is broken and in later builds it is removed completely. So, the only way to go would be to use the Win10-like taskbar shipped with EP. But the 3D taskbar buttons mod does not work with EP's taskbar. As such, as of now, the EP taskbar with your mod looks like Win10 taskbar without the 3D buttons mod and has the same issues. I think, you should clarify that your mod does not support the EP's own taskbar.

Both of these issues are the responsibility of buttons mods. Therefore I do not think it is appropriate to mention these issues in my mod, which deals with background: 
* When there is no flat buttons mod present for Windhawk, then I hope one will be created.
* When the 3D buttons mods currently do not work in Windows 11 24H2 then I hope the buttons mods will be updated someday accordingly.

*(Please allow me to clarify my choices: I other words, I am afraid it would not be prudent for me to take responsibility for other people's work. Or for me to try to bear the burden of entire world - considering my limitations as a human being, that would be unsustainable - the only result would be that I would not be able to contribute even the little amount I otherwise would contribute. With this mod I wanted to share the work I did for my own use, in the hope that it helps somebody else as well. And then added a bit of extra compatibility with different button mods and for vertical Taskbar layout. Unfortunately, presently I do not have capacity and motivation to create a mod that would incorporate the code of currently implemented and non-implemented button mods so that my mod would support rendering all possible button appearances, as well as then to maintain the compatibility of these button rendering functionalities with future Windows releases. That being said, if there manifests an urgent need, maybe I could contribute to the continued Windows compatibility of the button mods in the future. I hope these practical human constraint based considerations sound understandable.)*

I am afraid that people would have trouble reading even the amount of text that is already present in my readme, so adding more would just add more mental overload to the prospective readers.

I made it more prominent that a buttons mod is needed. Hopefully people do not glance over it. But there seems to be some reasonable limit to how prominent I can make that warning. I am afraid if I paint the warning red and write it in uppercase then that will just look too loud and ugly.

With these latest commits I hope the mod is mostly ready for publication. I am curious how the updated readme text looks to you - ? If there is anything you would like me to adjust in the readme text, please let me know!

m417z:
> I am curious how the updated readme text looks to you - ?

Looks better, thanks. I was mainly concerned about users who don't know much about classic themes, and that it might not be obvious whether the mod can be useful for them. This helps: "Install this mod if you have issues with black background [...]", as it makes the purpose of the mod clearer.

A user still might wonder how they can configure a classic theme, and IMO it can be nice to link to a resource with information on the topic, but I'm not sure there's such a resource. In any case, it looks good as it is, merging.

levitation:
> A user still might wonder how they can configure a classic theme, and IMO it can be nice to link to a resource with information on the topic, but I'm not sure there's such a resource. In any case, it looks good as it is, merging.

Oh, indeed. I forgot that you asked for this instruction, my apologies. I will then think about adding a link to existing instruction / discussion on that topic or maybe briefly documenting my setup, since the presently existing instruction has become somewhat long.

Thank you for merging :)

Anixx:
This is the tutorial on Classic theme: https://github.com/valinet/ExplorerPatcher/discussions/167

OrthodoxWindows:
@levitation Thank you for all these explanations. 
However, I have a technical question: why does your mod require the use of a button mod to work properly, while the modified version of StartMenuDLL.dll applies the background color without the need for a button mod?

levitation:
> @levitation Thank you for all these explanations. However, I have a technical question: why does your mod require the use of a button mod to work properly, while the modified version of StartMenuDLL.dll applies the background color without the need for a button mod?

My mod uses probably quite different approach to fix the background. Modified version of StartMenuDLL uses mostly the functionalities present in the original StartMenuDLL, by enabling them properly. The latter in turn contains a lot of logic and tricks, I have been unable to trace down where exactly and how it achieves the classic Taskbar appearance. I have investigated its code but have not been successful in replicating its behaviour. I have also tried may other tricks not present in my current mod code, which also were not producing desired results.

Regardless, considering how many different tastes people have for button appearance, I tend to think it might be good to keep background and buttons as separate mods.

PR #794, opened 2024-07-19
Add "Accent Color Sync"

CatmanFan:
Mod to automatically synchronise OpenGlass & StartIsBack++ colors with DWM/Windows accent color if either of them are installed.

### DIFF
--- a/mods/accent-color-sync.wh.cpp
+++ b/mods/accent-color-sync.wh.cpp
@@ -1,0 +1,392 @@
+// ==WindhawkMod==
+// @id                  accent-color-sync
+// @name                Accent color sync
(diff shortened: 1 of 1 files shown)


mods/accent-color-sync.wh.cpp
```diff
@@ -0,0 +1,392 @@
+// ==WindhawkMod==
+// @id                  accent-color-sync
+// @name                Accent color sync
+// @description         Syncs OpenGlass & StartIsBack++ colors with DWM
+// @version             1.0
+// @author              CatmanFan
+// @github              https://github.com/CatmanFan
+// @exclude             C:\Windows\system32\*
```
m417z: Since only explorer.exe is included, this line has no effect and can be removed.

PR #799, opened 2024-07-21
Update "Accent Color Sync"

CatmanFan:
Rewritten almost completely so that it now requires the OpenGlass Legacy edition ([source code](https://github.com/ALTaleX531/OpenGlass/tree/legacy), [fork](https://github.com/ALTaleX531/OpenGlass/pull/14)), among other updates

### DIFF
--- a/mods/accent-color-sync.wh.cpp
+++ b/mods/accent-color-sync.wh.cpp
@@ -312,7 +312,7 @@ long __cdecl CCTrackBar_SetThumbPosition_hook(class CCTrackBar* This, int value)
 }
 #pragma endregion
 
-WindhawkUtils::SYMBOL_HOOK hooks[] = {
+WindhawkUtils::SYMBOL_HOOK dui70dll_hooks[] = {
     {
         {L"StrToID"},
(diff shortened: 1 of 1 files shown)


mods/accent-color-sync.wh.cpp
```diff
@@ -172,225 +165,264 @@ BOOL set_DWORD(std::wstring sk, std::wstring v, DWORD data)
...
 
-    return FALSE;
+    return CCTrackBar_SetThumbPosition(This, value);
 }
+#pragma endregion
 
-void ColorFunction(BOOL force = false)
-{
-    if (!isWorking) 
+WindhawkUtils::SYMBOL_HOOK hooks[] = {
```
m417z: Have you seen the warning for this line? I added a new rule for naming symbol hook variables, to be able to optimize loading symbols. Let me know if the message is not clear enough.

CatmanFan: I saw it, but actually didn't understand what it was supposed to mean at first until after scanning through it this morning. It should be fixed now

PR #836, opened 2024-08-03
Fix Qt scrollbar grippers 1.0

kawapure:


kawapure:
The check failed because the Twitter handle "requires manual verification" - I can verify my ownership of the account by any means necessary if they would be provided, or the field can removed.

image

m417z:
Please send me (https://twitter.com/m417z) a direct message with the following content:

> I attest that I'm the sole owner of both this Twitter account (https://twitter.com/kawaipure) and the following GitHub account: https://github.com/kawapure

kawapure:
Done!

mods/fix-qt-scrollbar-grippers.wh.cpp
```diff
@@ -0,0 +1,87 @@
+// ==WindhawkMod==
+// @id              fix-qt-scrollbar-grippers
+// @name            Fix Qt scrollbar grippers
+// @description     Fixes scrollbar grippers in Qt 6.1+
+// @version         1.0
+// @author          kawapure
+// @github          https://github.com/kawapure
+// @twitter         https://twitter.com/kawaipure
+// @homepage        https://kawapure.github.io
+// @include         ida64.exe
```
m417z, line 10: Was it meant to only target IDA?

kawapure: IDA was the application that I tested against, but it should work for any Qt 6.1+ application using native theme scrollbar controls. I thought that it would be obvious to any end user to specify custom targets. Should this be clarified in the README?

m417z: Yes, better clarify it, something like:
https://github.com/ramensoftware/windhawk-mods/blob/a480102c52e3cd16e1cea214913e1395155c82c0/mods/ignore-focus-loss.wh.cpp#L20-L23

kawapure: I copied that clarification.

PR #837, opened 2024-08-04
Update classic-file-picker-dialog.wh.cpp

Anixx:
Fix reference to another mod


mods/classic-file-picker-dialog.wh.cpp
m417z, line 1: While at it, can you rename the `symbolHook` variable to `comdlg32DllHooks` according to the instructions?

Anixx:
Done

mods/classic-file-picker-dialog.wh.cpp
Anixx: Done

PR #848, opened 2024-08-10
Create legacy-alt-tab.wh.cpp

Anixx:


mods/legacy-alt-tab.wh.cpp
```diff
@@ -0,0 +1,49 @@
...
+// @description  Enables legacy Alt+Tab dialog from Windows XP and Windows 2000
+// @version      1.0
+// @author       Anixx
+// @github       https://github.com/Anixx
+// @include      explorer.exe
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+Enables legacy Alt+Tab dialog from Windows XP and Windows 2000
```
m417z, line 13: How about adding a screenshot?

Anixx: Done

PR #868, opened 2024-08-17
Aerexplorer 1.6.3: Make compilable on Windhawk 1.5

aubymori:


m417z:
You should be able to keep compatibility with earlier Windhawk versions by:
* Adding `-DWINVER=0x0A00` to compilerOptions.
* Adding a copy of the relevant function declarations, for example:
https://github.com/ramensoftware/windhawk-mods/blob/069694253210b43492a520720b96abc9ebd0e48b/mods/taskbar-icon-size.wh.cpp#L124-L134

I think it's worth to at least keep compatibility with 1.4.

aubymori:
> You should be able to keep compatibility with earlier Windhawk versions by:
> 
>     * Adding `-DWINVER=0x0A00` to compilerOptions.
> 
>     * Adding a copy of the relevant function declarations, for example:
>       https://github.com/ramensoftware/windhawk-mods/blob/069694253210b43492a520720b96abc9ebd0e48b/mods/taskbar-icon-size.wh.cpp#L124-L134
> 
> 
> I think it's worth to at least keep compatibility with 1.4.

Alright, I went ahead and did that.

m417z:
Thanks. Could you also rename the hook variables according to the instructions, e.g. [here](https://github.com/ramensoftware/windhawk-mods/pull/868/files#diff-31354d2fd537ce4f2064606ad330a089de4928117463bcee6f611d0b8f806e77R2558)? Let me know if you need a clarification.

aubymori:
Done.

m417z:
Looks good, the script still parses `windowsStorageDllHooks` even though there's a comment, let me fix that.

m417z:
Closing and re-opening to trigger a new check.

PR #869, opened 2024-08-17
Transparent Screensaver Fix 1.0.0

aubymori:
This mod allows transparent screensavers, such as the Bubbles screensaver, to see your desktop again, rather than displaying a solid color.

**Before**:
Before

**After**:
After

### DIFF
--- a/mods/transparent-screensaver-fix.wh.cpp
+++ b/mods/transparent-screensaver-fix.wh.cpp
@@ -1,0 +1,129 @@
+// ==WindhawkMod==
+// @id              transparent-screensaver-fix
+// @name            Transparent Screensaver Fix
+// @description     Allows transparent screensavers to see your desktop again
+// @version         1.0.0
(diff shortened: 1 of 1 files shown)


mods/transparent-screensaver-fix.wh.cpp
```diff
@@ -0,0 +1,129 @@
...
+    && 0 == wcsicmp(psi->lpDesktop, L"Winsta0\\Screen-saver"))
+    {
+        psi->lpDesktop = nullptr;
+    }
+    return CUser_CreateProcessW_orig(
+        pThis, a2, a3, a4, psi, a6, a7
+    );
+}
+
+const WindhawkUtils::SYMBOL_HOOK winlogonHooks[] = {
```
m417z:
```suggestion
const WindhawkUtils::SYMBOL_HOOK winlogonExeHooks[] = {
```

PR #870, opened 2024-08-17
Update classic-taskbar-buttons-lite.wh.cpp

Anixx:
Fix for artefacts

### DIFF
--- a/mods/classic-taskbar-buttons-lite.wh.cpp
+++ b/mods/classic-taskbar-buttons-lite.wh.cpp
@@ -111,7 +111,7 @@ BOOL Wh_ModInit(void)
 {
     HMODULE hExplorer = GetModuleHandleW(NULL);
 
-    WindhawkUtils::SYMBOL_HOOK hooks[] = {
+    WindhawkUtils::SYMBOL_HOOK explorerExeHooks[] = {
         {
             {
                 L"private: void " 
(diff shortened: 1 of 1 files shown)


m417z:
Please rename the `hooks` variable according to the instructions while at it. For example, you can rename it to `explorerExeHooks`.

Anixx:
Done

levitation:
@Anixx Now that you made the adjustments to your mod, my workaround to your mod's previous version causes reverse artefacts - The buttons appear slightly flatter than they should. Fortunately it is not glaringly visible, so users may not notice it.

There is also no way for my mod to know with reasonable effort whether the user is using older or newer version of your mod. For time being, I will keep my workaround to your mod's old version active.

levitation:
@Anixx Do I get it right that you actually liked the bigger spacing between Taskbar buttons you had in the earlier version of the mod? And now some other incompatibility than my mod caused you to remove the spacing? 

If so, then maybe you could actually keep the spacing, if you first paint the button area with taskbar background colour using FillRect(), then change the offsets as you did in the earlier version of your mod, and then finally render the button with DrawFrameControl() ? 

Curious, how does that sound to you?

Anixx:
@levitation the spacing is the same, not changed.

levitation:
@Anixx Indeed, I need to restart Taskbar process to see the spaces, otherwise they disappear upon the activation of your mode. 

There is one more nuance, the spacing is visible only between buttons of different programs. If there is a group of buttons from multiple instances of same program, then these buttons now do not have spacing. For example, if you launch Notepad multiple times.

Is that how you intended it to be? Then I know that if I change my mod accordingly, then it is a purposeful work.

Anixx:
It is how Windows intends it to be. If one dislikes different spaces, he can disable groupping (for instance, in 7+TT) or installing the "Disable groupping" mod.

levitation:
@Anixx Understood, I will then consider that you will keep this grouping option active (unless user disables it via 7+TT or by other means). Thank you again for responding and letting me know!

PR #874, opened 2024-08-18
Adjustment for painting multi-row horisontal taskbar that is fully populated in all rows

levitation:
Without this adjustment, small horisontal black lines appear at the end of taskbar buttons area in case of multi-row horisontal taskbar that is fully populated in all rows


levitation:
Not sure why Github says there are conflicts. In any case, you can just use version of code I pushed.

m417z:
Please resolve the conflicts so that it'll be possible to merge the pull request. See:
https://stackoverflow.com/questions/16306012/github-pull-request-showing-commits-that-are-already-in-target-branch

PR #885, opened 2024-08-22
Classic theme enable with extended compatibility

levitation:
This mod version adds support for RDP sessions and compatibility with early / system start of Windhawk.

Added also personally validated and updated instructions for setting up rest of classic theme configuration in case of a new operating system installation.

Initially I proposed to merge the updates to Anixx's classic theme enable mod, but they suggested that I create a separate mod in order to be able to provide better support. So after a bit of further polishing and adding the instructions I am now creating a pull request for publishing it as a separate mod.


levitation:
Initially I added accidentally two screenshot files into the PR. After I removed them it seems the PR mod validation check step still chocked somehow. Please let me know if I need to create a new PR.

m417z:
LMK when it's ready to be merged.

levitation:
> LMK when it's ready to be merged.

Thank you! Yes, it is ready for merge.

mods/classic-theme-enable-with-extended-compatibility.wh.cpp
```diff
@@ -0,0 +1,656 @@
...
+2. Open "Task Scheduler Library" section 
+3. Find the row titled "WindhawkRunUITask", open it by double clicking
+4. Go to Triggers
+5. Click "New..." button
+6. Select Begin the task: "At startup"
+7. OK
+8. Click "New..." button
+9. Select Begin the task: "At log on"
+10. OK
+11. OK
```
m417z, line 195: Does it actually help? It just launches the UI, unless I'm missing something.

levitation: @m417z Maybe you are correct, I am not sure about this part, and likely you know it better. It seemed that creating a system start trigger improved the chance that Windhawk is started early enough. I see that Windhawk is also run as a service. Could there be any chance that the service does not hook into other processes until UI process is launched?

I have seen a few times that even with these Task Scheduler triggers, in one of my machines, the Windhawk might hook winlogon.exe too late. This manifests in some programs starting without classic theme.

m417z: The two are mostly independent, and the service doesn't wait for the UI. By default, the UI doesn't run on startup, the service does. In fact, the UI folder can be fully removed after installing mods and Windhawk will keep on working.

levitation: Ok, I need to do more experiments with this then to verify whether triggering the UI somehow helps to speed up the start of the service part.

m417z: I doubt that, likely some placebo effect. You can do some experiments, but eventually I prefer to remove this misleading information in order to avoid confusion, unless you indeed find something surprising.

Now that I think about it, what might work (but I didn't try it) is picking a system service which starts early, and adding the Windhawk service to its dependency list.

levitation:
@OrthodoxWindows Thank you so much for your elaborate suggestions!

I created a Google Doc where we could work together on the Readme text. Would you be willing to add your modification points there in Suggesting mode so I can see better where you want to add the recommendations? How does that sound to you? Here it is:
https://docs.google.com/document/d/1Ko9WhRFB7XHqsAug3KKSo4PsuUxSL9_SToHiouRU1k8/edit?usp=sharing

mods/classic-theme-enable-with-extended-compatibility.wh.cpp
levitation: Thank you for the idea, sounds good! I will try it.

levitation: I imagine it should be a system service which is non-essential. Else there would be a risk that if Windhawk is unable to start for some reason, then the system service does not start either.

m417z: It'd be better, but perhaps you must choose an essential service to ensure the load time that's needed for this mod. Hooking into `winlogon.exe` already has risks and might require booting in safe mode to disable Windhawk if something goes wrong.

levitation: Alternative would be perhaps creating a separate Task Scheduler task that triggers Windhawk service start using net start command.

m417z: You can try that, but it feels more fragile to me. Also, I don't know how early the Task Scheduler service runs, perhaps not earlier than Windhawk does now.

OrthodoxWindows:
Thanks, I'll look into that.

OrthodoxWindows:
I will try to edit the document in Suggesting mode

OrthodoxWindows:
@levitation Have you been able to read my suggested edits?

levitation:
@OrthodoxWindows Thank you so much! I apologise for the confusion, I was not sure whether you finished and at the same time had to deal with various priority stuff in the meanwhile. I hope to be able to take a look soon.

oep42:
Thanks for creating this mod. I tried it by following the instructions for setting up the classic theme [here](https://docs.google.com/document/d/1Ko9WhRFB7XHqsAug3KKSo4PsuUxSL9_SToHiouRU1k8/edit) (as edited by OrthodoxWin32). I am using Windows 11 Version 22H2 (Build 22621.2506).

**About 2. Additional needed software**

As a replacement for Task Manager I choose [System Explorer](https://systemexplorer.net/), because System Informer is too complicated for me. System Explorer can be launched with Ctrl+Alt+Insert, after adding this as shortcut key to its startup properties. (System Explorer's CPU usage info was found to be incorrect with high CPU usage. To monitor performance, Windows' own "Resource Monitor" can be used.)

I also installed the Classic Task Manager, in case I use Ctrl+Alt+Delete.
(**Suggestion:** Add to the text of the link to the Classic Task Manager, that the "Windows 7-like Task Manager for Windows 11/10/8.1/8" should be installed, not "Classic Task Manager for Windows 11 (Windows 10 version)".)

I had a problem with ExplorerPatcher. ExplorerPatcher did not work properly because I use a firewall that blocks everything by default and does not notify me when a program tries to communicate with the Internet. I just had no idea that additional Internet access was needed. Because of this, the necessary symbols were not downloaded. This is my solution for this problem:
--- Make sure your firewall allows "C:\Windows\explorer.exe" to communicate with the Internet.
--- Go to the tab Advanced of ExplorerPatcher's Properties.
--- Make sure "Enable symbols download" is enabled.
--- Enable "Enable Console". A Windows console window will appear. After a while, messages appear in the console window (symbols are downloaded).
--- Disable "Enable Console" when no new messages appear in the console window.
--- Restart the PC.
(**Suggestion:** Add info to the instructions that for ExplorerPatcher to work, "C:\Windows\explorer.exe" should be allowed to communicate with the Internet, and that a temporary "Enable Console" is required to download symbols.)

**About 3. Needed changes in Windhawk settings**

For the six mods I decided to use, only these two items have to be added to the process inclusion list: "dllhost.exe" and "winlogon.exe".
(**Suggestion:** Specify here which items are needed for which mods.)

**About 5. Additional needed classic theme related mods**

I decided to use these six mods: "Classic Task Dialog Fix", "Classic Taskbar 3D buttons Lite", "Classic Taskbar background fix", "Classic theme transparency fix", "Classic UWP Fix", and "Clientedge Everywhere". The other mods had no added value for me. 
--- In all programs without hard-coded rounded corners that I tried, rounded corners were disabled, so the mod "Disable rounded corners in Windows 11" has no added value for me. (**Suggestion:** Add a note that this mod may not be needed to disable rounded corners.) [**Note:** Toast notifications still have rounded corners, which is perfectly fine for me. I am not looking for nostalgia where everything is the same as in the past, but for a user interface that helps me be as productive as possible.]
--- For me, the mod "Fake High Contrast" caused some very ugly dialog boxes, and may not be needed at all, because in the Effects tab of "WinClassicThemeConfig", the option "High Contrast mode" seems to have the same effect.

**About 6. How to configure system start of Windhawk**

I skipped this. For me, this mod works without setting anything in Task Scheduler.
(**Suggestion:** Add a note that this may not be needed at all.)

**About 7. Adjusting the colours, window metric  and fonts**

I used "WinClassicThemeConfig" (its dialog box is named "Appearance Properties") with the reg file to get the scheme present in Windows 2000 and XP, with "Color scheme = Windows Standard" and "Size = 4-Windows 10". After completing the instructions for setting up the classic theme, this results in the following user interface:

Windows user interface with Classic Theme

In the "Appearance Properties" dialog box, items of this user interface can can be enlarged for those who find things too small. (For me, "WinClassicThemeConfig" is clearly superior to "Desktop Themes" and "Desktop Architect".)
(**Suggestion:** The current version of WinClassicThemeConfig has a higher version number than 1.5.5.2, so I would omit that number from the link text.)

**Other remarks**

With version 1.2.1 of this mod, I had a few of these crashes: "Your session was logged off because DWM crashed". With the current version 1.3.1, these crashes seem to be a thing of the past.

A thing that could be improved is the user interface of File Explorer, especially the title bar. This is what an active and inactive File Explorer window looks like now:

Active and inactive window of Windows File Explorer

The buttons in the upper right corner of the File Explorer window do not work now.

**Conclusion:** Nice mod that works well.

levitation:
@oep42 Thank you for elaborate feedback! I plan to return to updating the readme file of the mod, once I finally get over the current busy spell and can give this topic the attention it deserves.

oep42:
**Some additional remarks**

The problem with the title bar of the user interface of File Explorer is fixed by selecting "Control Interface = Windows 7 Command Bar" or "Control Interface = Windows 10 Ribbon" in "Properties/File Explorer" of ExplorerPatcher.
(**Suggestion:** Add to "1. Needed registry changes", to the changes in the registry for ExplorerPatcher, that in "Properties/File Explorer" "Control Interface = Windows 7 Command Bar" is selected. Also add that in "Properties/File Explorer", the option "Shrink address bar height" is activated.)
(**Suggestion:** Add a note to the Troubleshooting section for what to do after restoring the default settings of ExplorerPatcher and the title bar of the user interface of File Explorer looks weird.)

Having to change things in the Windows registry, may create a barrier for those who want to use the Classic Theme but are not well versed in the registry. Suggestion to make using the Classic Theme more accessible:  

Create a single reg file that can be downloaded (with instructions how to use) that contains all the changes in the registry that are necessary for using the Classic Theme (except for section "7. Adjusting the colours, window metric and fonts"):
--- In this, do not include "If you want the 3D borders in menus, import the following reg file", because there is no need for it.
--- In this, rename this registry key from "HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Themes\DefaultColors\Standard" into "HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Themes\DefaultColors\Standard-Backup".
--- In this, include the necessary registry changes for ExplorerPatcher, Explorer, OpenShell, and 7 Taskbar Tweaker. (Perhaps also include the registry change for InprocServer32.)

It would be cool if there would be an easy way to undo all these changes in the registry. This may not be possible with a reg file, because the initial state of the registry before making the changes needed for the Classic Theme, may be different in each computer. So perhaps something like an installer file could be created to modify the registry for the Classic Theme (which remembers the initial state), and an uninstaller file to undo those changes and restore the registry to its initial state.

oep42:
I have had three DWM crashes today. See [here](https://github.com/ramensoftware/windhawk-mods/issues/1344).

levitation:
@oep42 Please note that this mod injects into winlogon.exe not into dwm.exe. Additionally this mod does nothing after enabling classic theme. It does not hook any functions. So if dwm.exe crashes then I suspect there is general problem with Classic Theme in your machine, not with this mod in particular.

oep42:
> I have had three DWM crashes today. See [here](https://github.com/ramensoftware/windhawk-mods/issues/1344).

These crashes were not caused by the "Classic theme enable with extended compatibility" mod. Sorry for bothering you about these crashes.

PR #900, opened 2024-08-26
 Windows 7 Alt+Tab Loader 2.0.0 

aubymori:
- Complete rewrite
- Support for UWP app icons
- Aero Peek now works

### DIFF
--- a/mods/win7-alttab-loader.wh.cpp
+++ b/mods/win7-alttab-loader.wh.cpp
@@ -2,11 +2,11 @@
 // @id              win7-alttab-loader
 // @name            Windows 7 Alt+Tab Loader
 // @description     Loads Windows 7 Alt+Tab on Windows 10.
-// @version         1.0.3
+// @version         2.0.0
 // @author          aubymori
 // @github          https://github.com/aubymori
 // @include         explorer.exe
-// @compilerOptions -lole32 -luuid -ldwmapi -luxtheme -ldbghelp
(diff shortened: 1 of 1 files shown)


mods/win7-alttab-loader.wh.cpp
```diff
@@ -16,568 +16,332 @@
...
     }
-
-    return PostMessageW(hWnd, uMsg, wParam, lParam);
+    CWindowInfo__GetIconProc_orig(hWnd, uMsg, upParam, lResult);
 }
 
-HTHEME OpenThemeData_Hook(HWND hWnd, LPCWSTR pszClassList)
-{
-    if (wcscmp(pszClassList, L"AltTab") == 0)
+const WindhawkUtils::SYMBOL_HOOK altTabDllHooks[] = {
```
m417z, line 233: You're challenging my symbol cache system :)

> "altTab.Dll" is not recognized as a Windows file name

That's because it's not available in Win10 and newer. Interesting that Microsoft still provides symbols for Win7 binaries. I wonder how long they keep it and what about e.g. WinXP.

From https://en.wikipedia.org/wiki/Windows_7:
> Mainstream support ended
> [...]
> Exceptions exist until at most October 8, 2024, see [§ Support lifecycle](https://en.wikipedia.org/wiki/Windows_7#Support_lifecycle) for details.

They might to decide removing these symbols at some point.

PR #925, opened 2024-09-05
Accent Color Sync v1.4

CatmanFan:
**Changelog:**
* Replaced DWM Colorization Parameters calculation formula with [ALTaleX's method](https://github.com/ALTaleX531/dwm_colorization_calculator).
* Compatibility check now returns the actual Windows version instead of the version returned from Explorer. This fixes the issue of the mod not being compatible with [explorer7](https://github.com/Erizur/explorer7-releases/).
* Updated readme.


m417z:
> * if you are using ports of the Windows 7 explorer to newer versions

Alternatively, instead of checking the OS version, you can check the version of the running instance of Explorer. You can find an example here:

https://github.com/ramensoftware/windhawk-mods/blob/fc6e85e377c3948bcf182bd71fe74f5f0afea3b5/mods/taskbar-thumbnail-reorder.wh.cpp#L832-L854

Let me know if you'd like to adapt it, or have it merged as is.

CatmanFan:
> Let me know if you'd like to adapt it, or have it merged as is.

I have just implemented your suggestion now, it should be safe to merge under this condition.

m417z:
Do you still need the "ex7" option? Maybe I misunderstood what you're trying to check. What's your compatibility check trying to check? The Windows version or the version of Explorer? Like, Windows 8.1 isn't supported, but Win7 explorer on Win10 is? Then why didn't it work correctly before this PR?

CatmanFan:
> Do you still need the "ex7" option? Maybe I misunderstood what you're trying to check. What's your compatibility check trying to check? The Windows version or the version of Explorer? Like, Windows 8.1 isn't supported, but Win7 explorer on Win10 is? Then why didn't it work correctly before this PR?

Apologies if I had not clarified this earlier: basically, the compatibility feature checks to see if the current Windows/Explorer version is Windows 10 or higher, and causes the mod to stop running if this check is not met. The Windows 7 explorer on 10 would be detected as running on a Win7 OS by the mod, and as a result the "ex7" option was added as a stopgap to disable the compatibility check completely so as to add support for it.

m417z:
> The Windows 7 explorer on 10 would be detected as running on a Win7 OS by the mod

Are you saying that `IsWindows10OrGreater` returned false while running in a Win7 explorer on Win10? That might be because of its old manifest.

In this case you can use a function which always returns the real Windows version, such as RtlGetVersion and RtlGetNtVersionNumbers, both have example code in mods (you can search in this repo).

m417z:
Let me know if you need help with this, I didn't mean to delay your new version.

Here, you can use this function:
```cpp
void GetNtVersionNumbers(ULONG* pNtMajorVersion,
                         ULONG* pNtMinorVersion,
                         ULONG* pNtBuildNumber) {
    using RtlGetNtVersionNumbers_t =
        void(WINAPI*)(ULONG * pNtMajorVersion, ULONG * pNtMinorVersion,
                      ULONG * pNtBuildNumber);
    static RtlGetNtVersionNumbers_t pRtlGetNtVersionNumbers = []() {
        HMODULE hNtdll = GetModuleHandle(L"ntdll.dll");
        if (hNtdll) {
            return (RtlGetNtVersionNumbers_t)GetProcAddress(
                hNtdll, "RtlGetNtVersionNumbers");
        }

        return (RtlGetNtVersionNumbers_t) nullptr;
    }();

    if (pRtlGetNtVersionNumbers) {
        pRtlGetNtVersionNumbers(pNtMajorVersion, pNtMinorVersion,
                                pNtBuildNumber);
        // The upper 4 bits are reserved for the type of the OS build.
        // https://dennisbabkin.com/blog/?t=how-to-tell-the-real-version-of-windows-your-app-is-running-on
        *pNtBuildNumber &= ~0xF0000000;
        return;
    }

    // Use GetVersionEx as a fallback.
    OSVERSIONINFO versionInfo = {sizeof(OSVERSIONINFO)};
    if (GetVersionEx(&versionInfo)) {
        *pNtMajorVersion = versionInfo.dwMajorVersion;
        *pNtMinorVersion = versionInfo.dwMinorVersion;
        *pNtBuildNumber = versionInfo.dwBuildNumber;
        return;
    }

    *pNtMajorVersion = 0;
    *pNtMinorVersion = 0;
    *pNtBuildNumber = 0;
}
```

Usage example:
```cpp
    ULONG majorVersion;
    ULONG minorVersion;
    ULONG buildNumber;
    GetNtVersionNumbers(&majorVersion, &minorVersion, &buildNumber);
    if (majorVersion >= 10) {
        Wh_Log(L"Win10 or newer: %u.%u.%u", majorVersion, minorVersion,
               buildNumber);
    }
```

CatmanFan:
> Let me know if you need help with this, I didn't mean to delay your new version.

It's all good, nonetheless. :) I already implemented this fix yesterday, but I had been busy working on the update throughout today, which was the main reason for the delay.

The original version has been usurped by v1.4, the changelog at the original post has been updated to reflect this.

PR #936, opened 2024-09-07
Classic Clock Button Behavior 1.0.0

aubymori:
In Windows XP and before, the tray clock did nothing on single click, and opened the Date and Time CPL on double click. This mod makes the Windows 10 tray clock button do that.

### DIFF
--- a/mods/classic-clock-button-behavior.wh.cpp
+++ b/mods/classic-clock-button-behavior.wh.cpp
@@ -1,0 +1,90 @@
+// ==WindhawkMod==
+// @id              classic-clock-button-behavior
+// @name            Classic Clock Button Behavior
+// @description     Makes the tray clock open the Date and Time CPL on double click
+// @version         1.0.0
+// @author          aubymori
+// @github          https://github.com/aubymori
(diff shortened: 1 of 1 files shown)


mods/classic-clock-button-behavior.wh.cpp
```diff
@@ -0,0 +1,90 @@
...
+            L"timedate.cpl",
+            NULL,
+            NULL,
+            SW_SHOWNORMAL
+        );
+    }
+    else
+    {
+        g_bDoubleClicked = true;
+        SetTimer(
```
m417z, line 57: You could just save the first click time and avoid the timer. Here's an example:
https://github.com/ramensoftware/windhawk-mods/blob/bf5b0746e2cb9d22e109b664ca3a63c430199f18/mods/pinned-items-double-click.wh.cpp#L113-L131

Just a suggestion, let me know if you want to adapt it or have it merged as is.

aubymori: It's fine as is.

PR #937, opened 2024-09-07
Logon, Logoff & Shutdown Sounds Restored - v1.0.0

TorutheRedFox:
Restores the playback of events once played by Explorer during the start, end, locking and unlocking of a logon session.

### DIFF
--- a/mods/logon-logoff-shutdown-sounds.wh.cpp
+++ b/mods/logon-logoff-shutdown-sounds.wh.cpp
@@ -95,6 +95,10 @@ LRESULT _OnSessionChange_hook(void* _this, WPARAM wParam, LPARAM lParam)
     return _OnSessionChange_orig(_this, wParam, lParam);
 }
 
+#ifndef SND_SYSTEM // because this is sometimes missing???
+#define SND_SYSTEM 0x00200000
+#endif
+
 DWORD WINAPI PlaySoundFileThreadProc( LPVOID lpParam ) 
 {
     PlaySoundW((WCHAR*)lpParam, 0, SND_NODEFAULT | SND_MEMORY | SND_SYSTEM);
@@ -236,7 +240,6 @@ STDAPI_(BOOL) SHRegisterClassW(const WNDCLASSW* pwc)
     }
     return TRUE;
 }
-
 
 HWND SHCreateWorkerWindowW(WNDPROC pfnWndProc, HWND hwndParent, DWORD dwExStyle, DWORD dwFlags, HMENU hmenu, void * p)
 {
--- a/mods/logon-logoff-shutdown-sounds.wh.cpp
+++ b/mods/logon-logoff-shutdown-sounds.wh.cpp
@@ -76,9 +76,9 @@ STDAPI_(void) ExplorerPlaySound(LPCTSTR pszSound)
     }
 }
 
-typedef LRESULT (* _OnSessionChange_t)(void* _this, WPARAM wParam, LPARAM lParam);
+typedef LRESULT WINAPI (* _OnSessionChange_t)(void* _this, WPARAM wParam, LPARAM lParam);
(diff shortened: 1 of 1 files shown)


mods/logon-logoff-shutdown-sounds.wh.cpp
```diff
@@ -0,0 +1,582 @@
+// ==WindhawkMod==
+// @id              logon-logoff-shutdown-sounds
+// @name            Logon, Logoff & Shutdown Sounds Restored
+// @description     Restores the logon, logoff and shutdown sounds from earlier versions of Windows
+// @version         1.0.0
+// @author          Toru the Red Fox
+// @github          https://github.com/TorutheRedFox
+// @twitter         https://twitter.com/TorutheRedFox
```
m417z, line 8: Please send me (https://x.com/m417z) a direct message with the following content:

> I attest that I'm the sole owner of both this X account (https://x.com/TorutheRedFox) and the following GitHub account: https://github.com/TorutheRedFox

mods/logon-logoff-shutdown-sounds.wh.cpp
```diff
@@ -0,0 +1,582 @@
...
+        && szFileName[0])
+    {
+        // flags are relevant, we try to not stomp currently playing sounds
+        PlaySound(szFileName, NULL, SND_FILENAME | SND_ASYNC | SND_NODEFAULT | SND_NOSTOP);
+    }
+}
+
+typedef LRESULT (* _OnSessionChange_t)(void* _this, WPARAM wParam, LPARAM lParam);
+_OnSessionChange_t _OnSessionChange_orig;
+LRESULT _OnSessionChange_hook(void* _this, WPARAM wParam, LPARAM lParam)
```
m417z: Calling conventions are missing, which is likely to cause a crash for 32-bit targets. Please add calling conventions or, if the mod doesn't support 32-bit, limit it to 64-bit, which will also make compilation faster. See:

mods/logon-logoff-shutdown-sounds.wh.cpp
```diff
@@ -0,0 +1,582 @@
...
+};
+
+CSoundWnd::CSoundWnd()
+    : m_refCount(1)
+    , m_hwndSound(nullptr)
+    , m_thread(nullptr)
+{
+}
+
+BOOL CSoundWnd::Init()
```
m417z: I see that some of the code is based on code from ExplorerPatcher. If significant parts are copied or heavily inspired by other projects, consider crediting them and specifying a compatible license for your mod.

mods/logon-logoff-shutdown-sounds.wh.cpp
```diff
@@ -0,0 +1,582 @@
...
+
+    LoadSettings();
+
+    g_hExplorer = GetModuleHandleW(NULL);
+    if (!g_hExplorer)
+        Wh_Log(L"Failed to get Explorer's handle");
+
+    g_hShlwapi = LoadLibrary(TEXT("shlwapi.dll"));
+
+    WindhawkUtils::SYMBOL_HOOK aHooks[] = {
```
m417z: As the message says, rename the hooks variable. In this case, rename it to `explorerExeHooks` or one of the other accepted variants.

TorutheRedFox:
The only thing based on what ExplorerPatcher did iirc is moving the code to wait for the sound to finish to WM_QUERYSESSIONEND, otherwise it's just reverse engineered code from vista's x86 explorer and 7's x64 explorer, and yeah I forgot about x86 I should fix that lol

mods/logon-logoff-shutdown-sounds.wh.cpp
m417z: You only had to add it for hooks, in your case just `_OnSessionChange_hook`, but the additional calling conventions don't harm.

PR #941, opened 2024-09-08
Old This PC Commands 1.0.1: Codebase cleanup

aubymori:
- Remove unnecessary `#include` and `DEFINE_GUID` statements
- Rely on public `IOpenControlPanel` interface instead of finding `COpenControlPanel::Open` from symbols
- Move symbol hook definitions and update the name to match modern conventions

### DIFF
--- a/mods/old-this-pc-commands.wh.cpp
+++ b/mods/old-this-pc-commands.wh.cpp
@@ -50,8 +50,11 @@ HRESULT OpenCplPage(IUnknown *punk, LPCWSTR lpPage, LPCWSTR lpSubPage)
             if (SUCCEEDED(hr))
             {
                 hr = pocp->Open(lpPage, lpSubPage, punk2);
+                pocp->Release();
             }
+            punk2->Release();
         }
+        pows->Release();
     }
     return hr;
 }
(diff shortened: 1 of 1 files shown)


mods/old-this-pc-commands.wh.cpp
```diff
@@ -28,40 +28,33 @@ to `ExplorerFrame.dll` takes place to remove the redirect.
...
                 CLSCTX_INPROC_SERVER | CLSCTX_INPROC_HANDLER | CLSCTX_LOCAL_SERVER | CLSCTX_REMOTE_SERVER,
-                IID_OpenControlPanel,
-                &pocp
+                IID_PPV_ARGS(&pocp)
             );
             if (SUCCEEDED(hr))
             {
-                hr = COpenControlPanel_Open(pocp, lpPage, lpSubPage, pu2);
+                hr = pocp->Open(lpPage, lpSubPage, punk2);
+                pocp->Release();
```
m417z, line 53: Perhaps you know that, but you can use [`winrt::com_ptr`](https://github.com/search?q=repo%3Aramensoftware%2Fwindhawk-mods+winrt%3A%3Acom_ptr&type=code) to have COM pointers released automatically for you.

PR #954, opened 2024-09-15
Message Box Fix 2.0.0

aubymori:
- Restructure code to match modern conventions
- Remove option to fix the font and make it always fix it; there's not much use to having it disabled.
- Revamp "Remove background" option
  - Renamed to "Classic style"
  - Now shows message boxes with Windows XP positioning and scaling
  - May break with per-monitor DPI, have not tested. Regardless, I don't think anyone using a Windows XP/2000 theme would want per-monitor DPI.

Here's what classic style looked like before:
message-box-fix-after-classic

And now:
image

### DIFF
--- a/mods/msg-box-font-fix.wh.cpp
+++ b/mods/msg-box-font-fix.wh.cpp
@@ -6,7 +6,7 @@
 // @author          aubymori
 // @github          https://github.com/aubymori
 // @include         *
-// @compilerOptions -luser32 -lgdi32 -lcomctl32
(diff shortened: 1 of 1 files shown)


m417z:
You can make it compatible with v1.4 by simply adding `-DWINVER=0x0A00` to `@compilerOptions`.

aubymori:
> You can make it compatible with v1.4 by simply adding `-DWINVER=0x0A00` to `@compilerOptions`.

Ah, I tried to do that with those `WINUSERAPI` definitions but I forgot about that. Did that just now.

PR #961, opened 2024-09-17
Custom Desktop Watermark 1.0.0

aubymori:
This mod allows you to completely customize the desktop watermark, putting in your own lines of text.

## Examples

**Windows 7 (not genuine)**:

Windows 7 (not genuine)

**Windows XP Professional x64 Edition**:

Windows XP Professional x64 Edition

### DIFF
--- a/mods/custom-desktop-watermark.wh.cpp
+++ b/mods/custom-desktop-watermark.wh.cpp
@@ -162,12 +162,15 @@ void LoadSettings(void)
     for (int i = 0;; i++)
     {
         LPCWSTR lpszText = Wh_GetStringSetting(L"lines[%i].text", i);
-        if (!lpszText || !*lpszText)
+        if (!*lpszText)
+        {
+            Wh_FreeStringSetting(lpszText);
             break;
+        }
         bool title = Wh_GetIntSetting(L"lines[%i].title", i);
         WatermarkLine line = { lpszText, title };
         g_lines.push_back(line);
-        Wh_FreeStrintSetting(lpszText);
+        Wh_FreeStringSetting(lpszText);
     }
 }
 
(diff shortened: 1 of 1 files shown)


mods/custom-desktop-watermark.wh.cpp
```diff
@@ -0,0 +1,199 @@
...
+{
+    std::wstring text;
+    bool title;
+};
+
+std::vector<WatermarkLine> g_lines;
+bool g_bClassic = false;
+
+bool (*CDesktopWatermark_s_WantWatermark_orig)(void);
+bool CDesktopWatermark_s_WantWatermark_hook(void)
```
m417z, line 63: Calling conventions are missing, which is likely to cause a crash for 32-bit targets. Please add calling conventions or, if the mod doesn't support 32-bit, limit it to 64-bit, which will also make compilation faster. See:

mods/custom-desktop-watermark.wh.cpp
```diff
@@ -0,0 +1,199 @@
...
+};
+
+void LoadSettings(void)
+{
+    g_lines.clear();
+    g_bClassic = Wh_GetIntSetting(L"classic");
+
+    for (int i = 0;; i++)
+    {
+        LPCWSTR lpszText = Wh_GetStringSetting(L"lines[%i].text", i);
```
m417z, line 164: The text is never freed, causing a memory leak. You can use `WindhawkUtils::StringSetting`.

aubymori:
Fixed both comments.

mods/custom-desktop-watermark.wh.cpp
```diff
@@ -0,0 +1,201 @@
...
+        if (!lpszText || !*lpszText)
+            break;
+        bool title = Wh_GetIntSetting(L"lines[%i].title", i);
+        WatermarkLine line = { lpszText, title };
+        g_lines.push_back(line);
+        Wh_FreeStrintSetting(lpszText);
```
m417z: Typo. Also:
* Not freeing when breaking at `!*lpszText` is still a leak.
* `Wh_GetStringSetting` never returns NULL. The check doesn't hurt, just FYI.

aubymori:
Fixed again.

PR #1014, opened 2024-09-29
classic right-click context menu (instead of win 11)

ttodua:
switches from win11 style to win10 style

screenshot

I understand that per se this mod is not a "hook" but i view Windhawk as a broader framework to apply customizations with any viable ways. (whatever possible for now. if you or someone have a dedicated time to do this with other ways, ok, but now this is the only mod that will do the job)

### DIFF
--- a/mods/classic-right-click-context-menu.cpp
+++ b/mods/classic-right-click-context-menu.cpp
@@ -1,0 +1,76 @@
+// ==WindhawkMod==
+// @id              use-windows-10-right-click-context-menu
+// @name            Switch Windows 11 right-click context menu to Windows 10 style
+// @description     This simplest mod will switch Win11 style right-click context menu in explorer into classic (Win 10) style
+// @version         0.1
+// @author          Puvox Software
+// @github          https://github.com/puvox
+// @homepage        https://puvox.software/
+// @include         explorer.exe
+// @compilerOptions -lcomdlg32
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Disable Win11 new context menu in explorer
+You need to enter the "Settings" page of this mod to enable it.
+
+# Getting started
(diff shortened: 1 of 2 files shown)


m417z:
As you noted, ideally Windhawk mods shouldn't make permanent changes to the system. You can try hooking registry access instead, here's a mod that does that which can serve as an example: [Windows 10 taskbar on Windows 11](https://windhawk.net/mods/win10-taskbar-on-win11).

ttodua:
@m417z I tried but can't make it work (not proficient in c++), would you have one min to give a more specific code on the fly, to achieve what this PR tries?

m417z:
I have it on my list to create such a mod, but currently I have more high-priority tasks.

ttodua:
is it optimal to have this current mod merged until you release better mod (then remove this mode), does that sound acceptable? if not, i will surely understand.

m417z:
I'd rather not merge this mod as is. If you need help improving it, you can try asking for help on the [Windhawk Discord server](https://discord.com/servers/windhawk-923944342991818753).

ttodua:
> I'd rather not merge this mod as is. If you need help improving it, you can try asking for help on the [Windhawk Discord server](https://discord.com/servers/windhawk-923944342991818753).

i've tried but no-one replied. 
it's pity that while PR is not damaging as is, and has some clear precautionary step (like "explicit need to enable/disable in settings) still does its job for users (and most windhawk users probably are not very newbies). windhawk could be more scalable if such PRs are approved (which actually work or uses any viable approach to achieve its goal), otherwise development could be pedalized .
anyway, decision yours, thanks for everything, keep it on 👍

m417z:
I just published [Classic context menu on Windows 11](https://windhawk.net/mods/explorer-context-menu-classic).

ttodua:
@m417z thanks!

PR #1016, opened 2024-09-29
StartIsBack++ Tweaker 0.7.1

Erizur:
Fixed a restart loop problem.


m417z:
It seems to me that the problem in this part of the code:
```cpp
        if (!isInitialThread) {
            system("taskkill /F /IM explorer.exe & start explorer");
            return FALSE;
        }
```

I also [addressed it on Discord](https://discord.com/channels/923944342991818753/923954164583759873/1267575259947597917).

> If Windhawk didn't inject code into the parent process (e.g. it's excluded, protected, or running as admin while Windhawk portable is used), then the injection happens with a slight delay and `isInitialThread` is false. In general, I don't think restarting explorer without prompting the user is a good user experience. The "Taskbar height and icon size" mod has code that prompts the user to restart explorer to apply the changes, you can borrow it.
> https://github.com/ramensoftware/windhawk-mods/blob/7e0bd27ae1d12ae639497fbc9b48bb791f98b078/mods/taskbar-icon-size.wh.cpp#L723

Erizur:
Fixed again, sorry!

PR #1059, opened 2024-10-11
Fixed issue of video drivers not starting properly in some computers when Windhawk with Classic theme mod is started very early during system start.

levitation:


mods/classic-theme-enable-with-extended-compatibility.wh.cpp
```diff
@@ -535,7 +536,12 @@ DWORD WINAPI InitThreadFunc(LPVOID param) {
     bool isRetry = false;
 retry:  //If Windhawk loads the mod too early then the classic theme initialisation will fail. Therefore we need to loop until the initialisation succeeds. Also if the mod is loaded into a RDP session too early then for some reason that would block the RDP session from successfully connecting. So we need to wait for session "active" state in case of RDP sessions. This is another reason for having a loop here.
     if (isRetry) {
-        if (WaitForSingleObject(g_initThreadStopSignal, 1) != WAIT_TIMEOUT) {
+#ifdef _DEBUG
+        DWORD waitInterval = 1000;
+#else 
+        DWORD waitInterval = 1;
+#endif
+        if (WaitForSingleObject(g_initThreadStopSignal, waitInterval) != WAIT_TIMEOUT) {
```
m417z, line 544: Is this the actual change? I assume it will still be 1?

levitation: No, actual change is on line 398. Before it applied the session connected status check only to RDP sessions. Now it applies it to console sessions as well. 

The code snippet you cited is 
```
#ifdef _DEBUG
DWORD waitInterval = 1000;
#else 
DWORD waitInterval = 1;
#endif
```
So the interval is still 1 in a live mod.

m417z: Totally missed it, saw only the comment :)

PR #1081, opened 2024-10-13
Create mod "Taskbar language indicator layout control"

levitation:


m417z:
Nice, I don't use RDP all that much, but users were asking for it on the 7+ Taskbar Tweaker feedback website:
https://tweaker.userecho.com/topics/163-dont-hide-language-bar-in-rdp

You might want to leave a comment there with your solution. LMK if you don't feel like creating an account etc., in which case I'll post it on your behalf.

levitation:
@m417z Thank you for pointing to the feedback website! I left an announcement under the post you referred to.

PR #1092, opened 2024-10-16
Accent Color Sync v1.52

CatmanFan:
**Changelog:**
* OpenGlass: Removed old registry name option as it's redundant for the latest versions.
* OpenGlass: Fixed incompatibility with OpenGlass DComp (a registry value had been fixed at zero, which caused a visual bug on that version).
* Fixed compatibility for legacy versions of Windows 10 (1709 and below).
* **Omitted changelog from the previous version (1.51):**
  * Fixed a minor bug in which the "Fixed opacity" option in the mod settings was accidentally left enabled. Please disable the option manually if you are not planning on using the feature.
  * Fixed a bug in which the OpenGlass options were incorrectly swapped.
  * Updated readme to correct some things.

### DIFF
--- a/mods/accent-color-sync.wh.cpp
+++ b/mods/accent-color-sync.wh.cpp
@@ -671,7 +671,6 @@ void setColorizationBalance(bool bruteforce = FALSE)
     writeColorizationBalance(dwmSettings.color_balance, dwmSettings.afterglow_balance, dwmSettings.blur_balance);
     
     // Other registry values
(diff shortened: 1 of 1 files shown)


m417z:
> Omitted changelog from the previous version (1.51):

In fact I adjusted the changelog generation to allow fixups for cases such as 1.51, and the changelog looks fine now:
https://windhawk.net/mods/accent-color-sync

As I wrote on Discord, it's best for the changelog to be part of the commit message, otherwise I have to copy it manually, and it's easy to miss.

PR #1125, opened 2024-10-22
Taskbar classic context menu v1.0.1

m417z:
* Fixed the mod not working in some cases. Specifically, if the Windows TaskbarShiftRightClickCrash feature flag was enabled, right click acted as a left click instead of showing the classic context menu.
* The "Customize the old taskbar on Windows 11" option no longer needs to be enabled for the mod to work with the ExplorerPatcher taskbar.


fusionsynth:
Hi! After the latest W11 Update KB5052093, this mod unfortunately doesn't work anymore.

PR #1140, opened 2024-10-25
Fix Qt scrollbar grippers 1.1: Fix control detection

kawapure:
Previously, we attempted to draw the gripper overlay for any control that existed, not just scrollbars. This is because the control theme opened by the HTHEME was not checked. Now, we only try drawing scrollbar grippers if we know the control itself is a scrollbar.

### DIFF
--- a/mods/fix-qt-scrollbar-grippers.wh.cpp
+++ b/mods/fix-qt-scrollbar-grippers.wh.cpp
@@ -32,7 +32,8 @@ for injection.
 
 In Windhawk, go to the "Advanced" tab and scroll down to 
 "Custom process inclusion list". In that box, put the filename of the `.exe`. 
-The mod will immediately apply to those programs after you click "Save". 
+The mod will immediately apply to those programs after you click "Save", although the best results usually require a restart
+of that program because Qt may cache the bitmap, causing the element to sometimes appear unmodified.
 */
 // ==/WindhawkModReadme==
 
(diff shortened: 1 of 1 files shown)


mods/fix-qt-scrollbar-grippers.wh.cpp
```diff
@@ -40,20 +40,56 @@ The mod will immediately apply to those programs after you click "Save".
...
 #include <uxtheme.h>
 #include <vsstyle.h>
+#include <vector>
+#include <algorithm>
+
+std::vector<HTHEME> g_hScrollbarThemes;
+
+using OpenThemeData_t = decltype(&OpenThemeData);
+OpenThemeData_t OpenThemeData_orig;
+HTHEME OpenThemeData_hook(HWND hwnd, LPCWSTR pszClassList)
```
m417z: Calling conventions are missing, which is likely to cause a crash for 32-bit targets. Please add calling conventions or, if the mod doesn't support 32-bit, limit it to 64-bit, which will also make compilation faster. See:

ALTaleX531:
```cpp
// uxtheme!GetThemeClass exported entry 74
HRESULT STDMETHODCALLTYPE GetThemeClass(HTHEME hTheme, LPCWSTR pszClassIdList, int cchClass);
```
You can actually use GetThemeClass to get the class name of HTHEME without the need to hook additional functions.

kawapure:
> ```c++
> // uxtheme!GetThemeClass exported entry 74
> HRESULT STDMETHODCALLTYPE GetThemeClass(HTHEME hTheme, LPCWSTR pszClassIdList, int cchClass);
> ```
> 
> You can actually use GetThemeClass to get the class name of HTHEME without the need to hook additional functions.

Thank you for the information. This is very useful. I tried to find similar functionality but I failed to find such a function looking in IDA myself.

mods/fix-qt-scrollbar-grippers.wh.cpp
kawapure: This part has been rewritten.

PR #1186, opened 2024-11-01
Start button always on the left v1.2

m417z:
* Fixed the jumpy animation that occurred when the taskbar is animated.
* Fixed taskbar items overlapping with the start button when the taskbar is full.
* Fixed the mod when the "Start menu on the left" option is disabled.


ewanrf:
Would it be possible to have File Explorer on the left as well? next to Start

m417z:
@ewanrf currently that's not possible, it's not something that can be easily added, and I don't plan working on it for now.

PR #1232, opened 2024-11-10
Resource Redirect v1.1.9

m417z:
* Improved compatibility with some resources in some cases in the experimental mode.
* Fixed string redirection in some cases (non-experimental mode).
* Added support for theme.ini files larger than 32 KB (yes, somebody had a file larger than that).
* Added a workaround to the error message that says "Runtime Error - Not Enough space for thread Data". The error could occur when a process is shutting down due to a somewhat specific incompatibility between the Microsoft C++ runtime library and Windhawk's C++ library. It's not really harmful, as the process is shutting down anyway, but it can be annoying.


manaphychu:
https://github.com/ramensoftware/windhawk-mods/issues/1301

PR #1235, opened 2024-11-12
New mod per-app-language-preferences 1.0

yezhiyi9670:


yezhiyi9670:
I have been copying my previous code, so apologies for the careless mistakes.

For the last point, it is intended for changing the UI language, not the input langauge.

image

m417z:
That's what I understood from the linked Microsoft post, which says "Let me use a different input method for each app window". Perhaps this link is not quite about the same, then. Also, how about posting this screenshot to the description?

Also, using a static string variable makes the hook non-thread safe. 

Also, "Some compiler options and includes can be removed, e.g. wininet" - includes too.

yezhiyi9670:
What would be the best practice if I need a string buffer in the hook? Just make it non-static?

m417z:
Yes, in this case, why not? Also another small notice, the size that's specified includes the null terminator, so it should be 2049, not 2048.

PR #1267, opened 2024-11-22
Improved support for touch screens

m1lhaus:
Double tap on touch screens is now much more benevolent in terms of click position on the screen. It takes into account screen DPI since in hi-res screens (like my 14'' 4k laptop) the position diffs between clicks when performing double tap can be significant. Also, since middle mouse button click is not possible with touch and detecting touch gesture is quite cumbersome (to work reliably across both Windows 10 and Windows 11 taskbars), I implemented support for triple tap which activates middle mouse click action. Fixes https://github.com/m1lhaus/windhawk-mods/issues/11

### DIFF
--- a/mods/taskbar-empty-space-clicks.wh.cpp
+++ b/mods/taskbar-empty-space-clicks.wh.cpp
@@ -2,7 +2,7 @@
 // @id              taskbar-empty-space-clicks
 // @name            Click on empty taskbar space
 // @description     Trigger custom action when empty space on a taskbar is double/middle clicked
-// @version         1.5
+// @version         1.6
 // @author          m1lhaus
 // @github          https://github.com/m1lhaus
 // @include         explorer.exe
@@ -888,7 +888,7 @@ static com_ptr<IMMDeviceEnumerator> g_pDeviceEnumerator;
 
 // since the mod can't be split to multiple files, the definition order becomes somehow complicated
 bool IsTaskbarWindow(HWND hWnd);
-bool isMouseDoubleClick(LPARAM lParam);
+bool isMouseDoubleClick(LPARAM lParam, HWND hWnd);
 bool OnMouseClick(HWND hWnd, WPARAM wParam, LPARAM lParam, TaskBarAction taskbarAction);
 
 // =====================================================================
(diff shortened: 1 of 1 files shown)


mods/taskbar-empty-space-clicks.wh.cpp
```diff
@@ -1936,55 +2064,112 @@ void StartProcess(const std::wstring &command)
...
-    if (!taskbarClicked)
+
+    HWND hWnd = currentClick.hWnd;
+    UINT dpi = GetDpiForWindow(hWnd);
+    float dpiScale = static_cast<float>(dpi) / 96.0f; // 96 DPI is the standard scaling factor
+
+    // GetSystemMetrics(SM_CXDOUBLECLK) is suitable just for mouse, not really for touch
+    const int MAX_POS_OFFSET_PX = 15;
+    // if user has hi-res screen, every slight movement result in big pixel offset
+    const int MAX_POS_OFFSET_PX_SCALED = dpiScale * MAX_POS_OFFSET_PX;
```
m417z: Instead of using floating points, `MulDiv` is usually used for scaling, something like:
```cpp
    const int MAX_POS_OFFSET_PX_SCALED = MulDiv(MAX_POS_OFFSET_PX, dpi, 96);
```
It also takes care of rounding.

mods/taskbar-empty-space-clicks.wh.cpp
```diff
@@ -1936,55 +2064,112 @@ void StartProcess(const std::wstring &command)
...
     }
 
-    // Note: The reason why UIAutomation interface is used is that it reliably returns a className of the element clicked.
-    // If standard Windows API is used, the className returned is always Shell_TrayWnd which is a parrent window wrapping the taskbar.
-    // From that we can't really tell reliably whether user clicked on the taskbar empty space or on some UI element on that taskbar, like
-    // opened window, icon, start menu, etc.
+    // Check if the current event is within the double-click time and distance
+    bool result = abs(previousClick.position.x - currentClick.position.x) <= GetSystemMetrics(SM_CXDOUBLECLK) &&
+                  abs(previousClick.position.y - currentClick.position.y) <= GetSystemMetrics(SM_CYDOUBLECLK) &&
+                  ((currentClick.timestamp - previousClick.timestamp) <= GetDoubleClickTime());
```
m417z, line 2107: I assume DPI is relevant here too. Perhaps you can use `GetSystemMetricsForDpi` here.

PR #1269, opened 2024-11-22
Balloon Sound Fix 1.0.0

aubymori:


mods/balloon-sound-fix.wh.cpp
```diff
@@ -0,0 +1,101 @@
...
+// @architecture    x86-64
+// @compilerOptions -lwinmm
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Balloon Sound Fix
+Windows XP played a sound when showing balloon notifications. However, this
+functionality has been broken since Windows Vista. This mod fixes that issue in
+Windows 10.
```
m417z: What about Win11?

aubymori: > What about Win11?

I've just realized a better hook I can do to fix this problem that should be less reliant on struct offsets. I'll push that, which should hopefully work on Win11. As for later versions of Win11, I'll see what I can do to integrate this fix into ep_taskbar.

aubymori: Pushed.

m417z:
Did you test it on Win11? I see that taskbar.dll has a `SHPlaySound` function too, I'd assume that's what Win11 uses with the new taskbar?

aubymori:
> Did you test it on Win11? I see that taskbar.dll has a `SHPlaySound` function too, I'd assume that's what Win11 uses with the new taskbar?

Could you test for me? You need to set `HKEY_CURRENT_USER\SOFTWARE\Policies\Microsoft\Windows\Explorer\EnableLegacyBalloonNotifications` to `dword:00000000` and then you can use this PowerShell script to create a balloon:
```ps1
Add-Type -AssemblyName System.Windows.Forms
$global:balmsg = New-Object System.Windows.Forms.NotifyIcon
$path = (Get-Process -id $pid).Path
$balmsg.Icon = [System.Drawing.Icon]::ExtractAssociatedIcon($path)
$balmsg.BalloonTipIcon = [System.Windows.Forms.ToolTipIcon]::Warning
$balmsg.BalloonTipText = ‘This is a pop-up message text for the Windows user'
$balmsg.BalloonTipTitle = "Attention $Env:USERNAME"
$balmsg.Visible = $true
$balmsg.ShowBalloonTip(20000)
```
If Windows 11 does still show legacy balloons and this mod doesn't work, I'll update it to also hook Taskbar.dll

m417z:
I'll test soon. BTW the shortest mod which can be used as a reference for supporting both the Win10 taskbar, the Win11 taskbar, the Win10 taskbar on Win11 (with an option), and the EP taskbar, is `taskbar-primary-on-secondary-monitor`. It requires a couple of hundreds of boilerplate lines of code, but it should be fairly straightforward to just adjust the hook symbols and the hook function.

m417z:
I couldn't make it work on a Win11 23H2 VM:

image

This guide implies that it should work:
https://www.makeuseof.com/windows-11-10-balloon-notifications/
But perhaps it was removed at some point in newer Win11 builds. I didn't find definitive info regarding supported builds for this registry option.

BTW I think it's worth mentioning this registry tweak in the mod's readme, or alternatively linking to a guide like that one.

aubymori:
> I couldn't make it work on a Win11 23H2 VM:
> 
> image
> 
> This guide implies that it should work: https://www.makeuseof.com/windows-11-10-balloon-notifications/ But perhaps it was removed at some point in newer Win11 builds. I didn't find definitive info regarding supported builds for this registry option.
> 
> BTW I think it's worth mentioning this registry tweak in the mod's readme, or alternatively linking to a guide like that one.

Did you restart Explorer after setting that key?

m417z:
Yes, I rebooted the VM too.

aubymori:
So I guess Win11's taskbar just doesn't support legacy balloon notifications anymore. I'll just update the README to include the regkey to get balloon notifications and go bake a patch into ep_taskbar then, I guess.

aubymori:
@m417z Could you please merge if there isn't any other issues?

PR #1293, opened 2024-12-03
Disable Virtual Desktop Transition Animation v1.0

kawapure:


kawapure:
That's great...

image

mods/disable-virtual-desktop-transition.wh.cpp
```diff
@@ -0,0 +1,205 @@
...
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Disable Virtual Desktop Transition Animation
+
+This mod disables the animation seen when switching between virtual desktops in Windows 10.
+
+In order to use this mod, you have to hook into `dwm.exe`. In order to do this, you must go into Windhawk advanced settings
+and remove the exclusion entry for `<critical_system_processes>`.
```
m417z: Allowing injection into all critical processes isn't necessary, it should be enough to allow `dwm.exe`. You can copy the message from here:

https://github.com/ramensoftware/windhawk-mods/blob/10dd219bec0369013eeeb9a3c0bee7efc3987f41/mods/disable-rounded-corners.wh.cpp#L34-L40

mods/disable-virtual-desktop-transition.wh.cpp
```diff
@@ -0,0 +1,205 @@
...
+	}
+
+    inline static bool IsWindows11OrGreater()
+    {
+        return GetVersionInfo()->dwMajorVersion >= 10 &&
+               GetVersionInfo()->dwMinorVersion >= 22000;
+    }
+};
+
+__declspec(noinline) HRESULT CVirtualDesktop__Initialize(void *pThis, void *a2)
```
m417z, line 114: The calling convention is missing here, which is likely to cause a crash for 32-bit targets. Please add calling conventions or, if the mod doesn't support 32-bit, limit it to 64-bit, which will also make compilation faster. See:

mods/disable-virtual-desktop-transition.wh.cpp
```diff
@@ -0,0 +1,205 @@
...
+    *(void **)pAnim = (void *)g_curVftableVirtualDesktopSwitch;
+
+    return pAnim;
+}
+
+// The mod is being initialized, load settings, hook functions, and do other
+// initialization stuff if required.
+BOOL Wh_ModInit()
+{
+    bool fAlreadyRunning = !(*(USHORT*)(((long long)NtCurrentTeb()) + 0x17EE) & 0x0400);
```
m417z: Can't `fAlreadyRunning` be false on system start? I think it's better to make the check stricter, e.g. by checking that it happens more than once in a short interval for the same session.

kawapure: It's never failed to work on system startup for me.

kawapure: ^ Or for other users who have tested it. It only cancels initialisation if DWM is restarted due to a crash, although I don't understand how it works, just that it works. If you disable the mod and re-enable it, then Windhawk will also succeed to inject it just fine.

m417z: Looks like on boot, dwm.exe is launched by wininit.exe which is a protected process. That's why Windhawk isn't able to intercept the process creation, and the injection happens asynchronously. But that doesn't mean that `fAlreadyRunning` will be always true in this case. If Windhawk is quick enough with the asynchronously injection, it will inject its code before dwm.exe actually starts running, and `fAlreadyRunning` will be false. Also, Windhawk might improve in the future to inject synchronously. So while this check is fine for testing, I'd not include it in this form. I suggest to either remove it or add a more robust check.

m417z: I looked for alternatives, and came up with this function:
https://gist.github.com/m417z/58d2fffd846c358b961f8c70d4f46c0d

Looks like a warning is logged when DWM restarts, it should be a more reliable check.

image

kawapure: Latest commit resolves this issue.

PR #1309, opened 2024-12-06
Add dark-menus.wh.cpp

MGGSK:


mods/dark-menus.wh.cpp
```diff
@@ -0,0 +1,63 @@
...
+// @author          Mgg Sk
+// @github          https://github.com/MGGSK
+// @include         *
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Dark mode context menus
+Forces dark mode for all win32 context menus to create a more consistent UI.
+*/
```
m417z, line 21: Nice. How about adding screenshots to the readme?

mods/dark-menus.wh.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+    SetPreferredAppMode = reinterpret_cast<SetPreferredAppMode_T>(pSetPreferredAppMode);
+    FlushMenuThemes = reinterpret_cast<FlushMenuThemes_T>(pFlushMenuThemes);
+
+    HRESULT hResult = ApplyTheme();
+    return SUCCEEDED(hResult);
+}
+
+//Restores the default theme.
+BOOL Wh_ModUninit()
```
m417z: `Wh_ModUninit()` returns `void`.

mods/dark-menus.wh.cpp
```diff
@@ -0,0 +1,63 @@
...
+using FlushMenuThemes_T = void (WINAPI *)();
+using SetPreferredAppMode_T = HRESULT (WINAPI *)(AppMode appMode);
+
+FlushMenuThemes_T FlushMenuThemes;
+SetPreferredAppMode_T SetPreferredAppMode;
+
+//Applies the theme to all menus.
+HRESULT ApplyTheme()
+{
+    FlushMenuThemes();
```
m417z, line 45: Shouldn't `FlushMenuThemes()` be called after `SetPreferredAppMode`?

MGGSK: This is the correct way to apply a menu theme because the SetPreferredAppMode function only works when the menu has no theme applied and the FlushMenuThemes function removes all themes from the menu.

m417z:
Congratulations on your first mod :-)

Issue #1157, opened 2024-10-27
[Taskbar on top for Windows 11 1.03] Start menu flies out from the bottom


watchmakerful:
@m417z A visual bug similar to #1053, but not identical to it. With this mod applied, the start menu animation still begins from the bottom, not from the top. The menu flies out from the bottom and hides to the bottom as well. Windows 11 23H2.
PS, the tray overflow panel and the keyboard layout selector are also animated from the bottom (but displayed correctly on the top).

m417z:
Taskbar on top for Windows 11 v1.1.5 adds an option to change the animation direction. The option isn't compatible with all configurations, so I'm leaving this issue open until there's a better way to do it, but for now it helps in some cases.

ferrazpedro:
When I enable the animation coming from the top, the phone link part is upside down:

<img width="845" height="742" alt="Image" src="https://github.com/user-attachments/assets/ce3b5a86-ba82-41c1-b6a5-2ca36478392b" />

atombun:
is there a way for this to work with the redesigned start menu...? as now its the default and cant be changed its like the one part of my system that really annoys me (as well as the task tray and when you hover over apps in the taskbar, they all have the same issue)

coinman-dev:
Regarding the redesigned Start menu: it turns out StartMenu.dll already has animations for all four taskbar edges. They're just disabled by a feature flag, and it's possible to turn them on.

What I found in `StartMenu.dll` (MicrosoftWindows.Client.Core, Windows 11 24H2 build 26100.9457):

- `AnimationProvider::CreateRootTranslationAnimation(CompositionPropertySet const &, TaskbarEdge)` chooses the translation expression by edge:
  - Left (0): `-AnimationProps.Width * (1 - AnimationProps.VisibilityProgress)`
  - Top (1): `-AnimationProps.Height * (1 - AnimationProps.VisibilityProgress)`
  - Right (2): `AnimationProps.Width * (1 - AnimationProps.VisibilityProgress)`
  - Bottom (3): `AnimationProps.Height * (1 - AnimationProps.VisibilityProgress)`

  `GetConditionExpressionForDismiss` uses the same enum values.
- That per-edge choice is gated by `wil::details::FeatureImpl<__WilExternalFeatureTraits_Feature_59213768>::__private_IsEnabled`. When the feature is off, which is the default, the Bottom expression is always used, whatever edge is passed.
[...]

PR #1324, opened 2024-12-11
fix: adjust startup menu position for multi screen

### LINKED ISSUE #1323: [Taskbar-on-top] Start menu always on right of screen for multi screen user (win11)
I use 2 screens and after download this mod when I open Startup menu it always on right of screen, cutting menu in middle.

image

After analyzing the source code, I realized that the code was having problems identifying the middle of the screen to do the calculation. I made some changes to the code and managed to make it work.

How do I make the code available for validation?

robertoruas:
Fixes #1323.

### DIFF
--- a/mods/taskbar-on-top.wh.cpp
+++ b/mods/taskbar-on-top.wh.cpp
@@ -1,6 +1,6 @@
 // ==WindhawkMod==
-// @id              taskbar-on-top
-// @name            Taskbar on top for Windows 11
+// @id              taskbar-on-top-fork
+// @name            Taskbar on top for Windows 11 - Fork
 // @description     Moves the Windows 11 taskbar to the top of the screen
 // @version         1.0.5
 // @author          m417z
@@ -1052,14 +1052,24 @@ BOOL WINAPI SetWindowPos_Hook(HWND hWnd,
         };
 
         HMONITOR monitor = MonitorFromPoint(pt, MONITOR_DEFAULTTONEAREST);
+
+        MONITORINFO monitorInfo = {sizeof(MONITORINFO)};
+        if (GetMonitorInfo(monitor, &monitorInfo)) {
+            // Corrigir o posicionamento horizontal
+            if (X < monitorInfo.rcWork.left) {
+                X = monitorInfo.rcWork.left;
+            } else if (X > monitorInfo.rcWork.right - cx) {
+                X = monitorInfo.rcWork.right - cx;
+            }
+
+            // Corrigir o posicionamento vertical
+            if (Y < monitorInfo.rcWork.top) {
+                Y = monitorInfo.rcWork.top;
+            }
+        }
         if (GetTaskbarLocationForMonitor(monitor) == TaskbarLocation::bottom) {
             return original();
         }
-
-        MONITORINFO monitorInfo{
-            .cbSize = sizeof(MONITORINFO),
-        };
-        GetMonitorInfo(monitor, &monitorInfo);
 
         if (g_inCTaskListThumbnailWnd_DisplayUI) {
             Y = std::max(monitorInfo.rcWork.top, pt.y);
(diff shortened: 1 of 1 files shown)


mods/taskbar-on-top.wh.cpp
```diff
@@ -1052,15 +1052,25 @@ BOOL WINAPI SetWindowPos_Hook(HWND hWnd,
...
+                X = monitorInfo.rcWork.left;
+            } else if (X > monitorInfo.rcWork.right - cx) {
+                X = monitorInfo.rcWork.right - cx;
+            }
+
+            // Corrigir o posicionamento vertical
+            if (Y < monitorInfo.rcWork.top) {
+                Y = monitorInfo.rcWork.top;
+            }
+        }
```
m417z, line 1069: This change is related to thumbnail previews, not the start menu. Is it necessary?

mods/taskbar-on-top.wh.cpp
```diff
@@ -1084,15 +1094,25 @@ BOOL WINAPI SetWindowPos_Hook(HWND hWnd,
...
```
m417z, line 1111: This change is related to popups, not the start menu. Is it necessary?

mods/taskbar-on-top.wh.cpp
```diff
@@ -1135,15 +1155,25 @@ BOOL WINAPI SetWindowPos_Hook(HWND hWnd,
...
```
m417z, line 1172: Same comment - this change is related to jump lists, not the start menu. Is it necessary?

mods/taskbar-on-top.wh.cpp
```diff
@@ -1245,7 +1275,19 @@ HRESULT WINAPI DwmSetWindowAttribute_Hook(HWND hwnd,
...
+        HMONITOR monitor = MonitorFromWindow(hwnd, MONITOR_DEFAULTTONEAREST);
+
+        MONITORINFO monitorInfo = {sizeof(MONITORINFO)};
+        if (!GetMonitorInfo(monitor, &monitorInfo)) {
+            return original(); // Fallback caso falhe.
+        }
+
+        // Centralizar no monitor ativo
+        int xNew = (monitorInfo.rcWork.left + monitorInfo.rcWork.right - cx) / 2;
+        int yNew = monitorInfo.rcWork.top;
```
m417z, line 1288: This seems to be the real fix. It's not clear to me why x and y have to be adjusted. I'd expect them to be handled by the taskbar and be correct already. Perhaps the issue is caused by something else, like another mod. How can I reproduce it? Does it occur if all other mods are disabled?

robertoruas: Yes. This behavior occur if I enable just this mod. 
My windows 11 version is 24h2 build 26100.2605. 

I noticed anothers strange behaviors:
- Even though adjust is applied when you click in Power or in profile image the startup menu is displaced to right side of screen.
- The startup menu slides from mid of screen to top. 
- The notifications slides from bottom right side not from top right side following taskbar position.

m417z: > Yes. This behavior occur if I enable just this mod.
> My windows 11 version is 24h2 build 26100.2605.

I couldn't reproduce the problem on my Win11 23H2. I don't have a 24H2 computer with multiple monitors right now, I'll try to reproduce it later. I prefer to understand what's going on there and why this check is needed.

Meanwhile, can you try the original code with this line removed: `*wParam = ABE_TOP;`? Does it fix the misplaced start menu?

Also, can you post a video recording of how it looks, as well as debug logs? Perhaps it will help reproduce the problem.

Please attach debug logs by following the instructions on the Troubleshooting page:
https://github.com/ramensoftware/windhawk/wiki/Troubleshooting#some-or-all-windhawk-mods-dont-work

> * Disable the mod.
> * Go to the mod's Advanced tab.
> * Set **Debug logging** to **Detailed debug logs**.
> * Click on **Show log output**.
> * Enable the mod.
> * Verify that the mod doesn't work, copy the log and save it to a file.
> * Set **Debug logging** back to **None**.

<hr>

> * Even though adjust is applied when you click in Power or in profile image the startup menu is displaced to right side of screen.

That's another reason to prefer getting to the root cause.

> * The startup menu slides from mid of screen to top.
> * The notifications slides from bottom right side not from top right side following taskbar position.

These are known limitations, see https://github.com/ramensoftware/windhawk-mods/issues/1157.

m417z:
Thanks for the contribution attempt. I don't plan on merging the PR in this form, closing, see my latest comment for the rational.

PR #1326, opened 2024-12-11
Create cef-titlebar-enabler.wh.cpp

Ingan121:


mods/cef-titlebar-enabler-universal.wh.cpp
```diff
@@ -0,0 +1,259 @@
+// ==WindhawkMod==
+// @id              cef-titlebar-enabler-universal
+// @name            CEF/Spotify Titlebar Enabler
+// @description     Force native frames and title bars for CEF apps
+// @version         0.1
+// @author          Ingan121
+// @github          https://github.com/Ingan121
+// @twitter         https://twitter.com/Ingan121
```
m417z, line 8: Please send me (https://twitter.com/m417z) a direct message with the following content:

> I attest that I'm the sole owner of both this Twitter account (https://twitter.com/Ingan121) and the following GitHub account: https://github.com/Ingan121

mods/cef-titlebar-enabler-universal.wh.cpp
```diff
@@ -0,0 +1,259 @@
...
+
+typedef int CEF_CALLBACK (*is_frameless_t)(struct _cef_window_delegate_t* self, struct _cef_window_t* window);
+int CEF_CALLBACK is_frameless_hook(struct _cef_window_delegate_t* self, struct _cef_window_t* window) {
+    Wh_Log(L"is_frameless_hook");
+    return 0;
+}
+
+typedef _cef_window_t* (*cef_window_create_top_level_t)(void* delegate);
+cef_window_create_top_level_t cef_window_create_top_level_original;
+_cef_window_t* cef_window_create_top_level_hook(void* delegate) {
```
m417z: The calling convention is missing here, which is likely to cause a crash for 32-bit targets. Please add calling conventions or, if the mod doesn't support 32-bit, limit it to 64-bit, which will also make compilation faster. See:

mods/cef-titlebar-enabler-universal.wh.cpp
```diff
@@ -0,0 +1,259 @@
...
+    BOOL isAuxProcess = FALSE;
+    for (int i = 1; i < argc; i++) {
+        if (wcsstr(argv[i], L"--type=") != NULL) {
+            isAuxProcess = TRUE;
+            break;
+        }
+    }
+    if (isAuxProcess) {
+        Wh_Log(L"Auxilliary process detected, skipping");
+        return TRUE;
```
m417z: Why not `return FALSE;` here to have the mod unloaded?

mods/cef-titlebar-enabler-universal.wh.cpp
Ingan121: Done

mods/cef-titlebar-enabler-universal.wh.cpp
Ingan121: Done

Anixx:
Can you please detailize what should be done with Electron? In all my attempts, the window frame is created buggy and gets glitched when the window is partially moved outside of the screen.

mods/cef-titlebar-enabler-universal.wh.cpp
Anixx: Should I use return FALSE always when I want the mod to exit regardless of success? Here, for instance: https://github.com/ramensoftware/windhawk-mods/blob/main/mods/classic-theme-enable.wh.cpp

mods/cef-titlebar-enabler-universal.wh.cpp
Ingan121: This mod already worked fine without adding a calling convention there on Spotify 1.1.x (which only had 32-bit releases). As adding a calling convention did not cause any harm, I added it anyway.

However, adding it to `cef_panel_create` caused Spotify 1.1.x to crash, so I did not add it there.

Also, as stated in the readme, 32-bit releases of 1.2.x are not supported. I did not figure out how to get it to work on those versions. (This mod simply does nothing on them.)

mods/cef-titlebar-enabler-universal.wh.cpp
Ingan121: Changed that to false.

m417z: @Anixx I think that it's best to return `TRUE` as long as the mod has some effect, and return `FALSE` only if the mod unloads without doing any changes, either due to an error or because the target isn't relevant/supported.

mods/cef-titlebar-enabler-universal.wh.cpp
m417z: > As adding a calling convention did not cause any harm

That's not enough to make sure that the calling convention is correct. If you don't have headers or documentation, you can check the calling convention via assembly. Usually it's either cdecl or stdcall, but there are other calling conventions too, such as fastcall. If you need help with that, you can upload the binary and I'll take a look.

> However, adding it to `cef_panel_create` caused Spotify 1.1.x to crash

By not adding a calling convention, you're leaving it up to the compiler to pick one. I'm not sure what's the default that Windhawk's compiler uses, maybe cdecl, but I'd rather not rely on it, as it might change in the future. It's best to specify the calling convention explicitly.

> Also, as stated in the readme, 32-bit releases of 1.2.x are not supported.

In this case, how about returning `FALSE` for these versions to have an indication the the mod wasn't loaded successfully? It can be something like:
```cpp
#ifndef _WIN64
  if (...1.2.x version...) {
    return FALSE;
  }
#endif
```

Ingan121: Inspected the binary with IDA and found out those two functions are cdecl. Btw there are no calling conventions defined for those functions in the CEF source code. It seems it's just using the compiler default.

Also found out that I simply missed that struct sizes are different per architecture; now it supports both x86 and x86-64 on all supported versions.

I also added support for more versions and tested all Spotify releases since 1.1.60. Now, there are no unsupported versions between CEF 90.4-132 and Spotify 1.1.60-1.2.52.

PR #1369, opened 2024-12-21
Remove Windows Spotlight Desktop Shortcut 1.0

Erizur:
This mod removes the new Spotlight shortcut that appears on the desktop when you use Spotlight wallpaper mode.
Works on latest Windows 10 and LTSC 2021.
I haven't tested it on Windows 11 myself as I don't have a Windows 11 machine. Should work just fine though.


Erizur:
Seems like it merged all my changes. Woops.

PR #1375, opened 2024-12-22
Create remove-activate-windows-watermark.wh.cpp

MRGamer69420:


m417z:
As I also asked in Discord: isn't this functionality already available via the [Desktop Watermark Tweaks](https://windhawk.net/mods/desktop-watermark-tweaks) mod?

aubymori:
> As I also asked in Discord: isn't this functionality already available via the [Desktop Watermark Tweaks](https://windhawk.net/mods/desktop-watermark-tweaks) mod?

Sorry for the late response but that mod pertains to the watermark with the Windows edition and build number, not the "Activate Windows" watermark.

aubymori:
I just looked at the code and it looks AI generated and as if it wouldn't even do anything reliable (why are we removing `WS_EX_LAYERED` from every window? why is the only include "windows"?). I think it'd be best to just close this PR.

PR #1381, opened 2024-12-25
CEF/Spotify Titlebar Enabler 0.5

Ingan121:
* Support making Spotify's custom window controls transparent
* Try to avoid crashes on untested newer releases of CEF/Spotify
* Support using DWM backgrounds with MicaForEveryone (requires hex patching `Spotify.exe`; see the mod details for instructions)


m417z:
> * requires hex patching `Spotify.exe`

Why not have an option for applying the patch in memory?
Here's an example for a simple mod that patches some code when it's loaded:
https://windhawk.net/mods/visual-studio-anti-rich-header

PR #1406, opened 2025-01-05
Dark mode context menus 1.1

MGGSK:
Added a option to only use context menus in dark mode if Windows is in dark mode.
Also cleaned up the code.

### DIFF
--- a/mods/dark-menus.wh.cpp
+++ b/mods/dark-menus.wh.cpp
@@ -65,6 +65,8 @@ HRESULT ApplyTheme(AppMode inputTheme = Max)
             theme = AllowDark;
         else
             theme = ForceDark;
+
+        Wh_FreeStringSetting(savedTheme);
     }
 
     //Apply the theme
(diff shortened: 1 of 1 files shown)


mods/dark-menus.wh.cpp
```diff
@@ -40,10 +52,30 @@ FlushMenuThemes_T FlushMenuThemes;
...
 //Applies the theme to all menus.
-HRESULT ApplyTheme()
+HRESULT ApplyTheme(AppMode inputTheme = Max)
 {
+    AppMode theme = inputTheme;
+
+    //Get the saved theme from the settings.
+    if(theme == Max)
+    {
+        PCWSTR savedTheme = Wh_GetStringSetting(L"AppMode");
```
m417z, line 62: The string is never freed, use `Wh_FreeStringSetting` to free it.

MGGSK: Fixed it

PR #1456, opened 2025-01-22
Update clientedge-in-apps.wh.cpp

Anixx:
Add support for the Control Panel


Anixx:
Needs fixes, adds double clienteddge

PR #1458, opened 2025-01-23
Per-app Language Preferences 1.1

yezhiyi9670:
The `GetUserDefaultUILanguage` is actually called very often in some apps such as `explorer.exe`. As I have observed, allocating the image path buffer and determining the language for each call can add much overhead. Now I choose to determine the language ID only once for each process at startup, and then return the pre-determined language ID for subsequent calls.

This approach seems to work fine, but may not be entirely correct. Please let me know if there is some better way of implementing this.

### DIFF
--- a/mods/per-app-language-preferences.wh.cpp
+++ b/mods/per-app-language-preferences.wh.cpp
@@ -154,7 +154,7 @@ BOOL Wh_ModInit() {
 void Wh_ModSettingsChanged() {
     Wh_Log(L">");
 
-    // determine_my_langid();
+    determine_my_langid();
 }
 
 // The mod is being unloaded, free all allocated resources.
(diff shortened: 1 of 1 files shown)


mods/per-app-language-preferences.wh.cpp
```diff
@@ -65,30 +65,42 @@ Note that it changes the language only and does not affect the encoding used in
...
+
+        int lang_id = -1;
+        for(int index = 0; ; index++) {
+            PCWSTR glob = Wh_GetStringSetting(L"programList[%d].glob", index);
+            if(!*glob) {
+                Wh_FreeStringSetting(glob);
+                break;
+            }
+            if(S_OK == PathMatchSpecExW(filename_buf, glob, PMSF_MULTIPLE)) {
+                lang_id = Wh_GetIntSetting(L"programList[%d].langId", index);
```
m417z: The string is not freed with `Wh_FreeStringSetting` in this case. Note that you can also use `WindhawkUtils::StringSetting` to have it freed automatically.

PR #1470, opened 2025-01-29
Classic Start Menu Folders Icon v1.0

tetawaves:
Changes the default icon of Start Menu folders that was used in Windows XP and before.


mods/classic-start-menu-folders-icon.wh.cpp
```diff
@@ -0,0 +1,212 @@
...
+        }
+    }
+
+    return result;
+}
+
+BOOL Wh_ModInit() {
+    Wh_Log(L"Init");
+
+    const WindhawkUtils::SYMBOL_HOOK windowsStorageDllHooks[] = {
```
m417z: `windows.storage.dll` can't be represented by a variable name. Please use a comment instead, for example:
```suggestion
    // windows.storage.dll
    const WindhawkUtils::SYMBOL_HOOK windowsStorageHooks[] = {
```

PR #1473, opened 2025-01-30
CEF/Spotify Tweaks 0.6

Ingan121:
* Add support for Spotify 1.2.55 and 1.2.56
* Support directly enabling transparent web content rendering without having to patch `Spotify.exe` manually
* Add options to disable forced dark mode or force-enable Chrome extension support
* Add a JavaScript API for interacting with this mod or the main window that can be used within Spicetify extensions and themes

### DIFF
--- a/mods/cef-titlebar-enabler-universal.wh.cpp
+++ b/mods/cef-titlebar-enabler-universal.wh.cpp
@@ -9,7 +9,7 @@
 // @homepage        https://www.ingan121.com/
 // @include         spotify.exe
 // @include         cefclient.exe
-// @compilerOptions -lcomctl32 -luxtheme -ldwmapi
+// @compilerOptions -lcomctl32 -luxtheme -ldwmapi -DWINVER=0x0A00
 // ==/WindhawkMod==
 
 // ==WindhawkModReadme==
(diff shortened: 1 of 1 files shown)


m417z:
Are you OK with not supporting Windhawk  1.4?
It might be enough to make it compatible by simply adding `-DWINVER=0x0A00` to `@compilerOptions`.

mods/cef-titlebar-enabler-universal.wh.cpp
```diff
@@ -471,25 +2008,30 @@ BOOL Wh_ModInit() {
...
-        return FALSE;
-    }
+    char* pbExecutable = (char*)GetModuleHandle(NULL);
+
+    #ifdef _WIN64
+        const size_t OFFSET_SAME_TEB_FLAGS = 0x17EE;
+    #else
+        const size_t OFFSET_SAME_TEB_FLAGS = 0x0FCA;
+    #endif
+    BOOL isInitialThread = *(USHORT*)((BYTE*)NtCurrentTeb() + OFFSET_SAME_TEB_FLAGS) & 0x0400;
```
m417z, line 2026: I wouldn't suggest to use this check outside of testing. Even on process start, `Wh_ModInit` isn't guaranteed to run on the initial thread, depending on the Windhawk injection settings and the parent process.

mods/cef-titlebar-enabler-universal.wh.cpp
```diff
@@ -569,6 +2171,11 @@ void Wh_ModUninit() {
 
 // The mod setting were changed, reload them.
 void Wh_ModSettingsChanged() {
+    if (g_isSpotifyRenderer) {
+        // Won't work in a sandboxed renderer process
```
m417z, line 2183: That's interesting, does `Wh_GetIntSetting` always return zero in this case? Since the mod is able to load, it must work at some point. Perhaps the process is hardened during initialization, after the mod is already loaded.

mods/cef-titlebar-enabler-universal.wh.cpp
Ingan121: The problem is that sandboxed renderer hooks must begin in the initial thread, as once Chrome code runs the process gets hardened (integrity drops from Low to Untrusted, etc.) and all pipe connections fail. I already experienced early injection failing when using `spicetify watch -l` but didn't find better alternatives other than neutering the sandbox security. (Fortunately didn't find any other failing scenarios yet other than that specific command for debugging CSS and JS.)

m417z: OK, that's justified in this case. That means that the mod won't work if loaded asynchronously, but that's probably the best we can do in this case.

mods/cef-titlebar-enabler-universal.wh.cpp
Ingan121: [Yes, it always returns zero.](https://github.com/user-attachments/assets/127ba34c-2862-485b-8060-05b0f5f8cfaf)
Seems it only works in the initial thread (before sandbox hardening)

PR #1493, opened 2025-02-06
Add support for the Win10 alt+tab on Win11

L3r0yThingz:


JFMugen:
@L3r0yThingz This might be a wrong place to ask this but hopefully its not.. May I ask for a small addition to your mod if its not too much trouble? I don't even know if it is possible to make to be honest. 

You know how Windows switches between windows on light speed when you hold down alt-tab right? So it makes holding down alt-tab pretty much pointless. I know you can just hold down Alt, but personally I use alt-tab with a macro on my mouse, and when there are more than 2 windows to switch it forces me to either use keyboard or manually select windows. So this change would be very helpful for that macro. 

I was wondering whether its possible to replace holding down alt-tab with ctrl-alt-tab. Your script put a 150ms delay by default right? So if you hold that long, it will open Ctrl-Alt-Tab window instead and let you choose whatever you want with mouse.

PR #1505, opened 2025-02-10
CEF/Spotify Tweaks 0.7

Ingan121:
* Add support for Spotify 1.2.57
* Fix JS API being unavailable on Spotify 1.2.4-1.2.32: use `window._getSpotifyModule('ctewh')` instead on these versions
* The transparent rendering, anti-forced dark mode, and Chrome extension enabler mods can now be applied even if the mod is loaded a bit late


Ingan121:
Unfortunately, injection after CEF initialization/main window creation is still not really supported for some reasons like `cef_window_create_top_level` hooks not being applied, reverting `SetWindowThemeAttribute` to get DWM window controls back seemingly impossible, and getting the main window handle becoming complicated. (excluding obvious factors like memory patches after CEF init being pointless, and the sandbox issue)

Issue #1510, opened 2025-02-10
Per-app language preferences mod does not work


Anixx:
@yezhiyi9670

For me, the mod does not work at all. I tried it with winver.exe, notepad.exe and other programs. It does not work regardless if the full path specified or not.

The system locale is ru-RU and I am trying to run programs in English locale (1033). The Windows version is Win 11 23H2.

![Image](https://github.com/user-attachments/assets/d51f4a51-0eb5-4499-beca-3af33a3590c0)

PR #1517, opened 2025-02-11
Per-app Language Preferences 1.2

yezhiyi9670:
Tries to improve compatibility, though still not working for a large number of programs, including UWP apps (notepad, mspaint, etc.) and winver.

Related: #1510


Anixx:
The version 1.2 still does not work in the Classic notepad:
{B6E354CA-D1F1-4ABC-9286-A157C4EE0D96}

As well as in notepad++, classic (Win7) calculator and basically I have no program where it works.

PR #1528, opened 2025-02-14
Restore Button Pulse Animation 1.0.0

aubymori:
In Windows Vista and 7, buttons played a "pulse" animation when they were focused. This mod restores that functionality.

Preview


Amrsatrio:
Does not work on 26100.3194 ARM64.

PR #1543, opened 2025-02-15
Create valinet-systray-sound-volume.wh.cpp

Anixx:


m417z:
In this and your other submissions of other's mods, have you tried contacting the authors and get them to submit the mods themselves? The readme is explicit about the author and the license, so I see no fundamental problem here, but I just think that it'd be better for the original authors to submit and maintain their own mods. If they explicitly say that they don't want to be bothered, then sure, you can maintain it on their behalf.

@valinet That's your mod, any thoughts about it?

valinet:
@m417z Do you plan on introducing 'third party repositories' support in Windhawk (in one form or another, that's another discussion) or you'd rather keep everything centralized, as it is now? What's the philosophy of Windhawk on this matter? Thanks.

m417z:
That's the first time third party repositories are brought up. While I didn't give it much thought, I don't see much value in it for users or developers. For users, I think that a single, centralized place for mods with verified authors makes using Windhawk much easier. For developers, it requires creating a pull request for each mod release, which I don't think is too demanding. For drafts and experiments, the code can always be shared to bypass the official store.

Why would you prefer third party repositories in Windhawk? Is there a reason for you to prefer not submitting your mods here instead?

If Windhawk ever becomes so popular that managing the mods in a single repository becomes a problem, third party repositories can be considered, perhaps similarly to Linux apt.

Anixx:
Fron a contrubuting user's perspective, third-party repositories can be convenient because in this case the user can make pull requests directly to the repository. In current situation only the mod author can contribute. Also, separate repositories can serve as a substitute for categories, uniting mods dedicated to a specific topic or modesets.

On the other hand, trirdparty repos may not necessary have all the needed checks set up, and any glitches that would arise from this would be blamed on Windhawk as a whole.

Anixx:
How is this possible if the build checks require the commit to originate from the mod's author? The checks would fail.

m417z:
The original author is still the one who submits new versions to the official repository. But everybody can submit pull requests to the author's repository.
```
         Contributors
          │        │
          ▼        ▼
    ──┬───────────────────┬──► Mod author's repo
      │                   │
    v1.0                v2.0
      │                   │
      ▼                   ▼
ramensoftware/      ramensoftware/
 windhawk-mods       windhawk-mods
```

That addresses your statement that "In current situation only the mod author can contribute". I see it as a good thing that each author takes full responsibility for their mods. What's the alternative?

m417z:
> 1. The author's repository usually gets deleted after their mod is submitted to the repo

That's up to the author, who might be uninterested in contributions, or might decide to stop maintaining the mod completely. If the author is interested in contributions, they can keep or re-create the repo.

> 2. Even if it is not deleted, it goes unsync with the Windhawk repo and SRs from it become unmergeable. Even if the changes in the main repo affect other files, unrelated to the mods folder.

That's a matter of knowing how to use git. Unfortunately, it can be cumbersome to do it directly from GitHub's web interface sometimes, but it's the most popular git service, and is probably the best platform overall. The upside is that because GitHub is so popular, there are guides for just about everything.

m417z:
@RibShark What I was trying to say in my previous messages is that you can do that. You can have a repo that has only your mod(s), your own files/scripts/readme/etc. I, myself, have a separate repo for developing my mods, which I could also host on GitLab, Dropbox, or any other place:

The interaction with this repository is only needed when releasing a new version, which usually doesn't happen that often. Instead of this repository, I could also have some form on, say, https://windhawk.net/submit-new-mod, with a "Browse..." button to upload the new version source code. But then I wouldn't have a way to leave comments on the code and discuss the submission, which is something GitHub provides nicely, and many developers are already familiar with it.

Maybe I can have a simplified submission form that integrates with GitHub, something similar to this:
https://new-issue.ant.design/
It isn't top priority among the things in my TODO list regarding Windhawk, but I'll keep it in mind, perhaps it can help some people who are less familiar with git or GitHub.

PR #1544, opened 2025-02-15
Explorerframe fixes for win11: more workarounds for unpredictable menubar appearance on later 23H2 and 24H2 builds

CyprinusCarpio:
- More workarounds for unpredictable menu bar appearance on later 23H2 and 24H2 versions

### DIFF
--- a/mods/explorerframe-fixes-for-win11-22h2plus.wh.cpp
+++ b/mods/explorerframe-fixes-for-win11-22h2plus.wh.cpp
@@ -2,7 +2,7 @@
 // @id              explorerframe-fixes-for-win11-22h2plus
 // @name            Explorerframe fixes for Win11 22H2+
 // @description     Fixes three problems with file explorer
-// @version         1.0
+// @version         1.0.1
 // @author          Waldemar
 // @github          https://github.com/CyprinusCarpio
 // @include         explorer.exe
@@ -25,9 +25,6 @@ Black artifacts briefly visible in new file explorer windows under Classic theme
 Changes will be visible in new file explorer windows.
 */
 // ==/WindhawkModReadme==
-#include <windhawk_utils.h>
-#include <Windows.h>
-#include <vector>
 
 // ==WindhawkModSettings==
 /*
@@ -39,6 +36,11 @@ Changes will be visible in new file explorer windows.
   $description: Use classic background color for explorer windows.
 */
 // ==/WindhawkModSettings==
+
+#include <windhawk_utils.h>
+#include <Windows.h>
+#include <shdeprecated.h>
+#include <vector>
 
 bool g_settingDisplayMenuBar;
 bool g_settingFixCWCBackground;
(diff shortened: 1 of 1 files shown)


CyprinusCarpio:
Again, less than surgical approach used. In the future, I may find what it is exactly about the `WM_WININICHANGE` message that fixes the problem, and do that more precisely, but for now I think it's sufficient to send this message and mitigate the one negative effect I identified during my testing.

Needs testing by @Anixx before merging.

Anixx:
I am on build 22631.4751 now and I do not experience any unpredictable menubar appearances so far. Even with the older mod. At which build number did you first exeprience troubles on 23H2?

Anixx:
As to 24H2, I deleted that virtual machine, will need to install it anew...

CyprinusCarpio:
> I am on build 22631.4751 now and I do not experience any unpredictable menubar appearances with the older mod, only without it. At which build number did you first exeprience troubles on 23H2?

Strange indeed. I'm testing on build 22631.4391 and the problem is the same as on 24H2: the menu bar doesn't want to stay put if it's the only toolbar, and it appears when the color theme changes or the resolution is changed (`WM_SETTINGSCHANGE` handler). Only after this message has been handled, does the menu bar stay if it's the only toolbar. Sending this message immediately is the first part of the solution. The second part is mitigating the problem the first part causes: that is the inability to save the layout of toolbars, if the menu bar was the only toolbar present when the window was opened.

Anixx:
Okay, I tried it and it does not work worse than the prevuous version on 23H2.

There is one glitch though, which, I believe existed in the prevuous version as well: upon change of the color scheme, the status bar in the folder windows of OpenShell (Classic Explorer) becomes white, so the windows should be re-opened. This is not the case when this mod is off:

изображение

CyprinusCarpio:
Alright, thanks for testing. So this can be merged. As for your report, I'll diagnose, and fix it in another PR one of these days.

Amrsatrio:
Hi @CyprinusCarpio (also pinged you on Discord, just in case)
https://github.com/ramensoftware/windhawk-mods/blob/81c1adc075f9038e4ac13ae77646e6de6c9e32d7/mods/explorerframe-fixes-for-win11-22h2plus.wh.cpp#L50
FYI, regarding the menu bar issue on windows 11 builds with tabs/TIFE (37634385) feature flag enabled,
I made a definitive real fix here which also fixes the last window size and position saving of explorerframe windows with 10 ribbon, though it uses FindPattern.
https://github.com/Amrsatrio/ExplorerPatcher/commit/75178ec35c41dbacc5092844a920188b41bdb52e
https://github.com/Amrsatrio/ExplorerPatcher/commit/df7d604169085c375c23bfc82184149ce4ad2e46
Also bonus: Alt+D fix on builds with FEMNB (40950262) enabled
https://github.com/Amrsatrio/ExplorerPatcher/commit/a80d9dc8442f58073a82a59dbada65cdc07bc9a6

PR #1575, opened 2025-02-22
w11-dwm-fix v3

teknixstuff:
Full rewrite of the mod.

### DIFF
--- a/mods/w11-dwm-fix.wh.cpp
+++ b/mods/w11-dwm-fix.wh.cpp
@@ -156,7 +156,7 @@ BOOL Wh_ModInit() {
         },
     };
 
-    WindhawkUtils::SYMBOL_HOOK uDWM22H2dllHooks[] = {
+    WindhawkUtils::SYMBOL_HOOK uDWMdllHooks22H2[] = {
         {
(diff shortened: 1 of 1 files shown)


m417z:
I don't think there's an easy way to get rid of the duplication. At this point, you can either adjust the variable names, which means all three should have the same name. You can either move them to functions or different scopes. Alternatively, you can use a comment as the validation comment suggests.

mods/w11-dwm-fix.wh.cpp
```diff
@@ -11,64 +11,113 @@
...
-project by m417z.
+char* s_pDesktopManagerInstance;
 
-Based on the
-[Win11DisableRoundedCorners](https://github.com/valinet/Win11DisableRoundedCorners)
-project by Valentin Radu.
+bool g_inCTopLevelWindow_UpdateWindowVisuals = false;
+bool g_inCLegacyNonClientBackground_SetBorderColor = false;
 
-## ⚠ Important usage note ⚠
```
m417z, line 31: Why not keep this usage note?

teknixstuff: oops, i forgot lol

m417z:
Is it ready to be merged?

teknixstuff:
> Is it ready to be merged?

It is now.

Issue #136, opened 2023-03-16
"Shrink Address Bar Height" mod crashes some 32-bit processes whenever Open/Save dialogs are shown


SteffanDonal:
Pinging @ItsProfessional as this is your mod 😉

I'm on Windows 11, 22H2 Build 22621.1413.

Below is a detailed log I captured of it happening in a 32-bit process that I'm using. Unfortunately, this one is "private", but you should be able to reproduce this reliably when attempting to open/save any file in `ResourceHacker`.

```
10:22:44.867 19320 devtool.exe  [WH] [`anonymous-namespace'::ModDebugLoggingScope::ModDebugLoggingScope]: >>> Entering LoadedMod::LoadedMod
10:22:44.867 19320 devtool.exe  [WH] [LoadedMod::LoadedMod]: Windhawk v1.2
10:22:44.867 19320 devtool.exe  [WH] [LoadedMod::LoadedMod]: Mod id: shrink-address-bar-height
10:22:44.867 19320 devtool.exe  [WH] [LoadedMod::LoadedMod]: Mod version: 1.0.0
10:22:44.867 19320 devtool.exe  [WH] [LoadedMod::LoadedMod]: Mod base address: 743A0000
10:22:44.867 19320 devtool.exe  [WH] [`anonymous-namespace'::ModDebugLoggingScope::~ModDebugLoggingScope]: <<< Exiting LoadedMod::LoadedMod
10:22:44.867 19320 devtool.exe  [WH] [`anonymous-namespace'::ModDebugLoggingScope::ModDebugLoggingScope]: >>> Entering LoadedMod::Initialize
10:22:44.867 19320 devtool.exe  [WH] [`anonymous-namespace'::ModDebugLoggingScope::ModDebugLoggingScope]: >>> Entering LoadedMod::SetFunctionHook
10:22:44.867 19320 devtool.exe  [WH] [LoadedMod::SetFunctionHook]: Target: 758B8CA0
10:22:44.867 19320 devtool.exe  [WH] [LoadedMod::SetFunctionHook]: Hook: 743A13B0
10:22:44.867 19320 devtool.exe  [WH] [`anonymous-namespace'::ModDebugLoggingScope::~ModDebugLoggingScope]: <<< Exiting LoadedMod::SetFunctionHook
[...]

PR #1581, opened 2025-02-24
Create disable-ms-teams-hotkey.wh.cpp

ygoe:


mods/disable-ms-teams-hotkey.wh.cpp
```diff
@@ -0,0 +1,48 @@
...
+Microsoft Teams occupies the keyboard shortcut Ctrl+Shift+Space to do whatever nonsense with it while sharing the screen.
+This hotkey is also assigned in Visual Studio by default to show the parameters info of the method under the cursor.
+That information is often very useful while coding, but it's not available in a team coding session when sharing the screen.
+So this mod prevents Teams from ever occupying that hotkey so that it remains available to Visual Studio like it's supposed to.
+*/
+// ==/WindhawkModReadme==
+
+BOOL(*pOriginalRegisterHotKey)(HWND hWnd, int id, UINT fsModifiers, UINT vk);
+
+BOOL RegisterHotKeyHook(HWND hWnd, int id, UINT fsModifiers, UINT vk)
```
m417z: The calling convention is missing here, which is likely to cause a crash for 32-bit targets. Please add calling conventions or, if the mod doesn't support 32-bit, limit it to 64-bit, which will also make compilation faster. See:

ygoe: This would also apply to the source I used: https://windhawk.net/mods/disable-feedback-hub-hotkey

mods/disable-ms-teams-hotkey.wh.cpp
```diff
@@ -0,0 +1,48 @@
...
+# Disable Microsoft Teams hotkeys used by Visual Studio
+
+Microsoft Teams occupies the keyboard shortcut Ctrl+Shift+Space to do whatever nonsense with it while sharing the screen.
+This hotkey is also assigned in Visual Studio by default to show the parameters info of the method under the cursor.
+That information is often very useful while coding, but it's not available in a team coding session when sharing the screen.
+So this mod prevents Teams from ever occupying that hotkey so that it remains available to Visual Studio like it's supposed to.
+*/
+// ==/WindhawkModReadme==
+
+BOOL(*pOriginalRegisterHotKey)(HWND hWnd, int id, UINT fsModifiers, UINT vk);
```
m417z: Here too:
```suggestion
BOOL(WINAPI *pOriginalRegisterHotKey)(HWND hWnd, int id, UINT fsModifiers, UINT vk);
```

mods/disable-ms-teams-hotkey.wh.cpp
m417z: Right, it's a bug there too. I didn't notice it back then, now it's up to @ItsProfessional to fix it.

This issue [caused crashes in the past](https://github.com/ramensoftware/windhawk-mods/issues/136). For that mod, which only targets explorer.exe, it's less severe, since the 32-bit explorer.exe is only used in 32-bit systems which aren't common nowadays.

mods/disable-ms-teams-hotkey.wh.cpp
ygoe: Alright, I don't know enough of the details of C. But I can't accept your suggestion because the GitHub UI doesn't show me the button in a too small container. Please accept it without my formal consent.

m417z: Yes, I have this problem too, looks like a GitHub bug. Just update the source code by adding the missing keyword, thanks.

ygoe: Your reply comment gave the box enough space to fit the accept button. 🙂

m417z:
Congrats on your first mod 🙂

PR #1653, opened 2025-03-15
Dark mode context menus 1.3

MGGSK:
-Added feedback links to the readme
-Added menubar screenshots to the readme
-Added support for menubars
-Added a German translation
-Code cleanup

### DIFF
--- a/mods/dark-menus.wh.cpp
+++ b/mods/dark-menus.wh.cpp
@@ -64,11 +64,6 @@ const HBRUSH brBackground = CreateSolidBrush(0x262626);
 const HBRUSH brItemBackgroundHot = CreateSolidBrush(0x353535);
 const HBRUSH brItemBackgroundSelected = CreateSolidBrush(0x454545);
 
-//Code taken from https://github.com/adzm/win32-darkmode/blob/darkmenubar/win32-darkmode/UAHMenuBar.h and https://github.com/adzm/win32-darkmode/blob/darkmenubar/win32-darkmode/win32-darkmode.cpp
-//MIT license, see LICENSE
-//Copyright(c) 2021 adzm / Adam D. Walling
-#pragma region UAHMenuBar
-
 #define WM_UAHDRAWMENU         0x91
 #define WM_UAHDRAWMENUITEM     0x92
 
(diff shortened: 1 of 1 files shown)


mods/dark-menus.wh.cpp
```diff
@@ -112,20 +328,27 @@ BOOL Wh_ModInit() {
...
+    ApplyTheme();
+    return TRUE;
 }
 
 //Restores the default theme.
 void Wh_ModUninit()
 {
     Wh_Log(L"Restoring the default theme.");
+
+    Wh_RemoveFunctionHook((void*) DefWindowProcW);
```
m417z: Per documentation, this function can't be called after `Wh_ModBeforeUninit` returns. It just fails here, and is generally unnecessary in this case.

mods/dark-menus.wh.cpp
```diff
@@ -31,15 +40,196 @@ Forces dark mode for all win32 context menus to create a more consistent UI. Req
...
+#include <uxtheme.h>
+#include <vsstyle.h>
 #include <windows.h>
 
+#include <windhawk_api.h>
+#include <winnt.h>
+
+const COLORREF crItemForeground = 0xFFFFFF;
+const COLORREF crItemDisabled = 0xAAAAAA;
+const HBRUSH brBackground = CreateSolidBrush(0x262626);
```
m417z, line 61: There's a memory leak here, brushes created with `CreateSolidBrush` should be freed with `DeleteObject`.

mods/dark-menus.wh.cpp
```diff
@@ -31,15 +40,196 @@ Forces dark mode for all win32 context menus to create a more consistent UI. Req
...
+	FillRect(hdc, &rcAnnoyingLine, brBackground);
+	ReleaseDC(hWnd, hdc);
+}
+
+#pragma endregion CodeBasedOnNotepad++
+
+//Code based on https://github.com/adzm/win32-darkmode/blob/darkmenubar/win32-darkmode/win32-darkmode.cpp
+#pragma region CodeBasedOnWin32DarkMode
+
+HTHEME g_menuTheme = nullptr;
```
m417z: To have better support for multiple threads, I think it's better to have this as `thread_local`.

mods/dark-menus.wh.cpp
```diff
@@ -75,23 +290,24 @@ HRESULT ApplyTheme(const AppMode inputTheme = Max)
...
-        FARPROC pRtlGetNtVersionNumbers = GetProcAddress(hNtDll, "RtlGetNtVersionNumbers");
+        const HMODULE hNtDll = LoadLibraryExW(L"ntdll.dll", nullptr, LOAD_LIBRARY_SEARCH_SYSTEM32);
+        const FARPROC pRtlGetNtVersionNumbers = GetProcAddress(hNtDll, "RtlGetNtVersionNumbers");
 	    RtlGetNtVersionNumbers = reinterpret_cast<RtlGetNtVersionNumbers_T>(pRtlGetNtVersionNumbers);
     }
 
     DWORD build;
     RtlGetNtVersionNumbers(nullptr, nullptr, &build);
 
-    build &= ~0xF0000000;
```
m417z: Why was it removed? As far as I understand, [it's necessary](https://dennisbabkin.com/blog/?t=how-to-tell-the-real-version-of-windows-your-app-is-running-on#:~:text=BuildNumber%20%26%20~0xF0000000), and without it the condition below will always be true.

PR #1662, opened 2025-03-16
Added SafeDiscShim

RibShark:
This adds SafeDiscShim as a Windhawk mod. This allows for SafeDisc protected games to run without their driver, making them more compatible with modern versions of Windows. Note that this does not bypass any security mechanisms; the games still require original discs to run and the logic in ProcessSecDrvIoctl() is 1:1 with the logic in the latest version of the secdrv.sys driver (which also hardcodes the values to return).


m417z:
Looks great! Congratulations on your first mod :)

PR #1669, opened 2025-03-19
Taskbar Clock Customization: Add user-customizable weekday abbreviations

eugenesvk:
Allows sweet short 1-letter abbreviations for the weekdays like `M` for Monday not to waste space in the previous taskbar.
Microsoft's default `ddd` goes to a min of 3 letters

(by the way, how would you create default values at the settings parsing stage? No matter what I've tried (creating `StringSetting WeekdayName = L"M"` from literals and what not), I just got undefined behavior, so only the verbose approach of assigning these literals post parsing as in this PR worked)

### DIFF
--- a/mods/taskbar-clock-customization.wh.cpp
+++ b/mods/taskbar-clock-customization.wh.cpp
@@ -105,9 +105,15 @@ styles, such as the font color and size.
 - WeekdayFormat: dddd
   $name: Week day format
   $description: >-
-    Leave empty for the default format. For syntax refer to the following page:
+    Leave empty for the default format.
+    'custom' use 'WeekdayFormatCustom', otherwise for syntax refer to the following page:
 
     https://docs.microsoft.com/en-us/windows/win32/intl/day--month--year--and-era-format-pictures
+- WeekdayFormatCustom: [M, T, W, R, F, S, U]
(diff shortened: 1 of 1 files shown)


m417z:

> The mod's `github` metadata value must match the pull request author's GitHub profile. This means that you can only submit an update for a mod that you originally submitted. If you'd like to update a mod that you didn't submit, you can either submit the changes to the mod author, or submit a new mod instead.

In this case, I'm the mod author, and I develop my mods at https://github.com/m417z/my-windhawk-mods. You can submit a PR there.

eugenesvk:
Yes, and since you're the original author owning this repo, didn't think it makes any difference. Anyway, moved the PR there, but then could you add the correct link https://github.com/m417z/my-windhawk-mods, not just https://github.com/m417z, to your mods so people know where to submit edits to?

m417z:
It's mentioned here:
https://github.com/ramensoftware/windhawk-mods/blob/8c2f0e04d5f2f6892fa24eb7a11b1efbcc51504c/mods/taskbar-clock-customization.wh.cpp#L20-L21

eugenesvk:
Oh, I'm sorry, got confused

PR #1684, opened 2025-03-21
Windows 10 Colored Borders 1.0.0

aubymori:
Starting from Windows 10 version 1809, the borders of active windows aren't colored unless the user enables the option to color titlebars. This mod reverts that behavior and makes active borders always use the user's accent color regardless.

**Preview**:
Preview

### DIFF
--- a/mods/win10-colored-borders.wh.cpp
+++ b/mods/win10-colored-borders.wh.cpp
@@ -1,0 +1,73 @@
+// ==WindhawkMod==
+// @id              win10-colored-borders
+// @name            Windows 10 Colored Borders
+// @description     Makes active borders always use the accent color in Windows 10
+// @version         1.0.0
+// @author          aubymori
+// @github          https://github.com/aubymori
+// @include         dwm.exe
+// @architecture    x86-64
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Windows 10 Colored Borders
+Starting from Windows 10 version 1809, the borders of active windows aren't colored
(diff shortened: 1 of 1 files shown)


mods/win10-colored-borders.wh.cpp
```diff
@@ -0,0 +1,73 @@
...
+# Windows 10 Colored Borders
+Starting from Windows 10 version 1809, the borders of active windows aren't colored
+unless the user enables the option to color titlebars. This mod reverts that behavior
+and makes active borders always use the user's accent color regardless.
+
+**Preview**:
+
+![Preview](https://raw.githubusercontent.com/aubymori/images/refs/heads/main/win10-colored-borders-preview.png)
+*/
+// ==/WindhawkModReadme==
```
m417z, line 31: How about adding:

https://github.com/ramensoftware/windhawk-mods/blob/10dd219bec0369013eeeb9a3c0bee7efc3987f41/mods/disable-rounded-corners.wh.cpp#L34-L40

aubymori: > How about adding:
> 
> https://github.com/ramensoftware/windhawk-mods/blob/10dd219bec0369013eeeb9a3c0bee7efc3987f41/mods/disable-rounded-corners.wh.cpp#L34-L40

Shouldn't it inject already with the default configuration? (Only if targeted explicitly)

aubymori: Okay, I just checked and you're correct. I'll add this to the README.

PR #1708, opened 2025-03-23
Taskbar keyboard-only auto-hide v1.1.1

m417z:
* Fixed Win+num hotkeys not working when the "Fully hide" option is enabled.

### DIFF
--- a/mods/taskbar-auto-hide-keyboard-only.wh.cpp
+++ b/mods/taskbar-auto-hide-keyboard-only.wh.cpp
@@ -2,7 +2,7 @@
 // @id              taskbar-auto-hide-keyboard-only
 // @name            Taskbar keyboard-only auto-hide
 // @description     When taskbar auto-hide is enabled, the taskbar will only be unhidden with the keyboard, hovering the mouse over the taskbar will not unhide it
-// @version         1.1
+// @version         1.1.1
 // @author          m417z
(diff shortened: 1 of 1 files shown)


mods/taskbar-auto-hide-keyboard-only.wh.cpp
```diff
@@ -151,6 +151,25 @@ BOOL WINAPI ShowWindow_Hook(HWND hWnd, int nCmdShow) {
...
+    if (g_settings.fullyHide) {
+        WCHAR szClassName[32];
+        if (GetClassName(hWnd, szClassName, ARRAYSIZE(szClassName)) &&
+            _wcsicmp(szClassName, L"MSTaskSwWClass") == 0) {
+            Wh_Log(L">");
+            return TRUE;
+        }
+    }
+
+    BOOL ret = IsWindowVisible_Original(hWnd);
```
whindsaks, line 168: Why not do IsWindowVisible_Original first and only if invisible, do the extra class name work?

m417z: No reason, just didn't think about it. [Changed](https://github.com/m417z/my-windhawk-mods/commit/76d66215112405ae6360762d0f33fdccd0a9c257).

PR #1727, opened 2025-03-28
Translucent windows v1.0

Undisputed00x:


mods/translucent-windows.wh.cpp
```diff
@@ -0,0 +1,494 @@
...
+    if (!pNtUserCreateWindowEx)
+        return FALSE;
+
+    Wh_SetFunctionHook((void*)pNtUserCreateWindowEx,
+                       (void*)HookedNtUserCreateWindowEx,
+                       (void**)&NtUserCreateWindowEx_Original);
+    return TRUE;
+}
+
+BOOL Wh_ModAfterInit()
```
m417z: `Wh_ModAfterInit` should return `void`.
```suggestion
void Wh_ModAfterInit() 
```

mods/translucent-windows.wh.cpp
```diff
@@ -0,0 +1,494 @@
...
+
+void Wh_ModUninit(void) 
+{
+
+}
+
+BOOL Wh_ModSettingsChanged(BOOL* bReload) 
+{
+    Wh_Log(L"SettingsChanged");
+    LoadSettings();
```
m417z: No need to load settings at this point since the mod is about to be reloaded anyway.
```suggestion
```

mods/translucent-windows.wh.cpp
```diff
@@ -0,0 +1,494 @@
...
+    #endif
+    bool isInitialThread = *(USHORT*)((BYTE*)NtCurrentTeb() + OFFSET_SAME_TEB_FLAGS) & 0x0400;
+    if (!isInitialThread)
+        ApplyForExistingWindows();
+    return TRUE;
+}
+
+void Wh_ModUninit(void) 
+{
+
```
m417z: How difficult would it be to restore the customizations when the mod unloads?
One way to do it is to have a global boolean variable called `unloading` or similar, then in `Wh_ModUninit` you can do something like:
```cpp
unloading = true;
ApplyForExistingWindows();
```

And in `EnumWindowsProc`:
```cpp
if (unloading)
    RestoreWindowCustomizations(hWnd);
else
    NewWindowShown(hWnd);
```

mods/translucent-windows.wh.cpp
```diff
@@ -0,0 +1,494 @@
...
+  $name: Effects
+  $description: Windows 11 version >= 22621.xxx (22H2) is required
+  $options:
+  - none: Default
+  - acrylicblur: Acrylic (BlurBehind)
+  - acrylicsystem: Acrylic (SystemBackdrop)
+  - mica: Mica
+  - mica_tabbed: MicaAlt
+- ExtendFrame: TRUE
+  $name: Extend effects into entire window
```
m417z, line 61: Here's how the mod affects the stock Windows theme with the default settings:

image

Perhaps the default settings can be tweaked for the common setup. For example, maybe it's better to have the `ExtendFrame` option disabled by default. What do you think?

Especially given that now disabling the mod doesn't revert the changes, it can result in a bad experience of having to reopen all windows to make them usable again.

m417z: These settings worked for me with a stock theme:

image

mods/translucent-windows.wh.cpp
Undisputed00x: I don't think restoring the affected windows to their original state is easy, there should be some logic in DWM to apply different kinds of title bars to different processes and BlurBehind is also a special case that has different rendering process. Even restarting dwm.exe doesn't have an effect, reverse engineering inside DWM is probably needed to see if that's feasible

m417z: I see that you added a note for `ThemeBackground` and `ExtendFrame`, what about `type`? If setting it to Default works, IMO it should be done on unload.

Undisputed00x: Resetting the effect type to default also does not work

m417z:
Congrats on your first mod!

PR #1764, opened 2025-04-05
File Explorer Toolbar Links Item v1.0

kawapure:
This mod restores the ability to display the "Links" toolbar, hidden since Windows Vista, in Windows 10 and 11.

The links toolbar can be used to provide a custom set of quick links that the user can access across the top of any File Explorer window.

Preview image

### DIFF
--- a/mods/explorer-toolbar-links-item.wh.cpp
+++ b/mods/explorer-toolbar-links-item.wh.cpp
@@ -8,6 +8,7 @@
 // @twitter         https://twitter.com/kawaipure
 // @homepage        https://kawapure.github.io
 // @include         explorer.exe
+// @architecture    x86-64
 // @compilerOptions -luser32 -lole32 -luuid
 // ==/WindhawkMod==
 
@@ -338,8 +339,8 @@ void Wh_ModUninit()
     Wh_Log(L"Uninit");
 }
 
-BOOL Wh_ModSettingsChanged(BOOL *pfReload)
+BOOL Wh_ModSettingsChanged()
 {
-    *pfReload = TRUE;
+    LoadSettings();
     return TRUE;
 }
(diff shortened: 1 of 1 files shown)


kawapure:
Is it okay for the 1.4.1 compatibility check to fail? I wasn't targeting for it anyway.

mods/explorer-toolbar-links-item.wh.cpp
```diff
@@ -0,0 +1,345 @@
...
+        }
+    }
+
+    return hMenu;
+}
+
+void (*CInternetToolbar___ShowVisible)(CInternetToolbar *pThis, DWORD dwNewVisibleItemFlags, BOOL fNotifyShellBrowser);
+
+void (*CInternetToolbar___ShowContextMenu_orig)(CInternetToolbar *pThis, HWND hWnd, LPARAM lParam, RECT *prc);
+void CInternetToolbar___ShowContextMenu_hook(CInternetToolbar *pThis, HWND hWnd, LPARAM lParam, RECT *prc)
```
m417z, line 195: The calling convention is missing here, which is likely to cause a crash for 32-bit targets. Please add calling conventions or, if the mod doesn't support 32-bit, limit it to 64-bit, which will also make compilation faster. See:

mods/explorer-toolbar-links-item.wh.cpp
```diff
@@ -0,0 +1,345 @@
...
+
+// The mod is being unloaded, free all allocated resources.
+void Wh_ModUninit()
+{
+    Wh_Log(L"Uninit");
+}
+
+BOOL Wh_ModSettingsChanged(BOOL *pfReload)
+{
+    *pfReload = TRUE;
```
m417z: Why reload instead of just calling `LoadSettings` here?

m417z:
> Is it okay for the 1.4.1 compatibility check to fail?

Yes, it's mostly informative at this point. Unfortunately GitHub checks don't have the concept of warnings, they either pass or fail. I'll remove the check for 1.4.1 once a new version is released, which will hopefully happen in the near future.

mods/explorer-toolbar-links-item.wh.cpp
kawapure: x86-32 offsets have been preserved in case they become useful again in the future. I will need to set up a 32-bit virtual machine to test, though.

PR #1765, opened 2025-04-06
File Explorer Toolbar Links Item 1.0.1

kawapure:
Fixed a minor injection issue that some users may experience.


m417z:
Looks like the changes accidentally revert some things from the previous submission (architecture, `Wh_ModSettingsChanged`).

kawapure:
Oh, good catch. I'll correct that.

PR #1767, opened 2025-04-07
Translucent windows v1.3

Undisputed00x:
* Add rules for processes
* Fix Control Panel title bar text color not being applied when using Acrylic Blur Behind
* Fix Snipping Tool not being affected by mod

### DIFF
--- a/mods/translucent-windows.wh.cpp
+++ b/mods/translucent-windows.wh.cpp
@@ -1,8 +1,8 @@
 // ==WindhawkMod==
 // @id              translucent-windows
 // @name            Translucent Windows
-// @description     Enables native translucent effects on windows
-// @version         1.2
+// @description     Enables native translucent effects in Windows 11
+// @version         1.3
 // @author          Undisputed00x
 // @github          https://github.com/Undisputed00x
 // @include         *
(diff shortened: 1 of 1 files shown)


mods/translucent-windows.wh.cpp
```diff
@@ -567,98 +651,190 @@ BOOL GetColorSetting(LPCWSTR hexColor, COLORREF& outColor)
...
+                if(RuledSettings->BorderFlag)
+                {
+                    LPCWSTR pszBorderStyle = Wh_GetStringSetting(L"RuledPrograms[%d].BorderColor.borderstyles", i);
+                    RuledSettings->BorderFlag = GetColorSetting(pszBorderStyle, RuledSettings->BorderColor);
+                    Wh_FreeStringSetting(pszBorderStyle);
+                }
+
+                RuledSettings->MenuBorderFlag = RuledSettings->BorderFlag;
+
+                SettingsMap.emplace(currProcID, std::move(RuledSettings));
```
m417z: Why have a map if the only key that's used for it is the current process id?
Why not just have a single global `Settings` struct for the current process?
The current approach works too, I can merge it as is, it's just a suggestion. Let me know if you'd like to change it.

Undisputed00x: Done it, ty.

PR #1769, opened 2025-04-07
File Explorer Toolbar Links Item 1.1

kawapure:
This update adds support for WoW64 (unfortunately, 32-bit Windows 10 seems to ignore the symbol hooks), fixes a crash in Windows 11, and improves the user experience.

The Links toolbar can now be enabled through "Options" dropdown in the View tab when using the Ribbon UI, like other toolbars can:

Ribbon UI enablement preview image

Additionally, the toolbars menu now always appears in the views menu of the menu bar when the classic Explorer interface is enabled:

Sole views menu item example image

All in all, these changes make the necessarily-hacky menu item show up in all the right places and behave just the same as any other toolbar option does.

### DIFF
--- a/mods/explorer-toolbar-links-item.wh.cpp
+++ b/mods/explorer-toolbar-links-item.wh.cpp
@@ -2,13 +2,12 @@
 // @id              explorer-toolbar-links-item
 // @name            File Explorer Toolbar Links Item
 // @description     Restores the ability to display the hidden "Links" toolbar in Windows 10 and 11.
-// @version         1.0.1
+// @version         1.1
 // @author          Isabella Lulamoon (kawapure)
 // @github          https://github.com/kawapure
 // @twitter         https://twitter.com/kawaipure
 // @homepage        https://kawapure.github.io
 // @include         explorer.exe
-// @architecture    x86-64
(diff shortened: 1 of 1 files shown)


m417z:
> unfortunately, 32-bit Windows 10 seems to ignore the symbol hooks

What do you mean by that? Does `HookSymbols` return false? Did you try enabling detailed logging?

kawapure:
It does not return false. I deleted my VM but I'll check again tomorrow. Right now, I'm consumed with setting up an ARM64 emulator so I can prepare for ARM64 support in Windhawk.

kawapure:
If I can figure out what the issue is with x86-32 tomorrow, then I'll drop a patch. Otherwise it's no big deal; WoW64 is what matters more, and the inclusion of this allows the changes to work in WoW64 contexts (not that there's any reason to right now, but if a mod were to come along to allow toolbars to display in any process's ExplorerFrame, then it would be a simple change to support that).

m417z:
I mean, if currently it doesn't work with the 32-bit explorer.exe, there's no need to compile it for 32-bit and the manifest can keep specifying x64-only, right?

kawapure:
Fair point.

PR #1770, opened 2025-04-07
Favorites in Navigation Pane v1.0.1

tetawaves:
* Fix memory leak
* Make sure to close registry key

### DIFF
--- a/mods/favorites-in-navpane.wh.cpp
+++ b/mods/favorites-in-navpane.wh.cpp
@@ -116,18 +116,17 @@ HRESULT WriteOrderRegistry(IStream *pStream)
 
 BOOL IsIShellItemFavorites(IShellItem *pIShellItem)
 {
+    BOOL isFavorites = FALSE;
+    
     if (pIShellItem)
     {
         LPWSTR pszName;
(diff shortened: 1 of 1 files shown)


mods/favorites-in-navpane.wh.cpp
```diff
@@ -104,7 +104,8 @@ HRESULT WriteOrderRegistry(IStream *pStream)
                 if (SUCCEEDED(hr))
                 {
                     hr = RegSetValueExW(hKey, L"Order", 0, REG_BINARY, lpData, streamSize.QuadPart);
-                }   
+                }
+                RegCloseKey(hKey);
```
m417z: This should be inside the `if`.

mods/favorites-in-navpane.wh.cpp
```diff
@@ -104,7 +104,8 @@ HRESULT WriteOrderRegistry(IStream *pStream)
                 if (SUCCEEDED(hr))
```
m417z, line 104: This should be:
```suggestion
                if (hr == ERROR_SUCCESS)
```
And `hr` is a misleading name here as it's not an `HRESULT`, it's a classic `GetLastError`-like error.

PR #1776, opened 2025-04-09
Translucent windows v1.4

Undisputed00x:
* Added color options for active and inactive window states

### DIFF
--- a/mods/translucent-windows.wh.cpp
+++ b/mods/translucent-windows.wh.cpp
@@ -2,11 +2,11 @@
 // @id              translucent-windows
 // @name            Translucent Windows
 // @description     Enables native translucent effects in Windows 11
-// @version         1.3
+// @version         1.4
 // @author          Undisputed00x
 // @github          https://github.com/Undisputed00x
 // @include         *
-// @compilerOptions -ldwmapi -luxtheme
+// @compilerOptions -ldwmapi -luxtheme -lcomctl32
 // ==/WindhawkMod==
 
 // ==WindhawkModReadme==
@@ -66,16 +66,22 @@ maximized or snapped to the edge of the screen, this is caused by default by the
 - TitlebarColor:
     - ColorTitlebar: FALSE
       $name: Enable
-    - titlerbarstyles: "FF0000"
-      $name: Color
+    - titlerbarstyles_active: "FF0000"
+      $name: Focused window color
+      $description: Color in hexadecimal RGB format e.g. Red = FF0000 or SystemAccentColor = 1
+    - titlerbarstyles_inactive: "00FFFF"
+      $name: Unfocused window color
       $description: Color in hexadecimal RGB format e.g. Red = FF0000 or SystemAccentColor = 1
   $name: Titlebar color
   $description: Windows 11 version >= 22000.xxx (21H2) is required. Overrides effects settings
 - TitlebarTextColor:
     - ColorTitlebarText: FALSE
       $name: Enable
-    - titlerbarcolorstyles: "FF0000"
-      $name: Color
+    - titlerbarcolorstyles_active: "FF0000"
+      $name: Focused window color
+      $description: Color in hexadecimal RGB format e.g. Red = FF0000 or SystemAccentColor = 1
+    - titlerbarcolorstyles_inactive: "00FFFF"
+      $name: Unfocused window color
       $description: Color in hexadecimal RGB format e.g. Red = FF0000 or SystemAccentColor = 1
   $name: Titlebar text color
   $description: >-
@@ -85,8 +91,16 @@ maximized or snapped to the edge of the screen, this is caused by default by the
 - BorderColor:
     - ColorBorder: FALSE
       $name: Enable
-    - borderstyles: "FF0000"
-      $name: Color
+    - borderstyles_active: "FF0000"
+      $name: Focused window color
+      $description: >-
+        Color in hexadecimal RGB format e.g. Red = FF0000
+
+        Transparent = 0
+        
+        SystemAccentColor = 1
+    - borderstyles_inactive: "00FFFF"
+      $name: Unfocused window color
       $description: >-
         Color in hexadecimal RGB format e.g. Red = FF0000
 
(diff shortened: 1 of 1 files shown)


mods/translucent-windows.wh.cpp
```diff
@@ -542,6 +666,10 @@ void NewWindowShown(HWND hWnd)
     }
     else
         EnableColoredTitlebar(hWnd);
+    
+    if(g_settings.BorderFlag || g_settings.CaptionTextFlag || g_settings.TitlebarFlag)
+        if(SetWindowSubclass(hWnd, SubclassProc, 0, 0))
```
m417z: `SetWindowSubclass` only works for windows in the same thread it's called from. It means that it will work for windows created in the hook, but not for windows which already exist when the mod is enabled. As an alternative that works in all cases, you can use `WindhawkUtils::SetWindowSubclassFromAnyThread`. 

Example:
https://github.com/ramensoftware/windhawk-mods/blob/10f1463d8bd2609b6278c5c548c58a1ed8ffcd98/mods/explorer-name-windows.wh.cpp#L285

Undisputed00x: Thanks for your suggestions as always

Undisputed00x: @m417z seems there is an issue with control panel where inactive state colors are applied on launch

mods/translucent-windows.wh.cpp
```diff
@@ -849,6 +991,16 @@ void Wh_ModUninit(void)
...
+
+    std::unordered_set<HWND> subclassedWindows;
+    {
+        std::lock_guard<std::mutex> guard(g_subclassedWindowsMutex);
+        subclassedWindows = std::move(g_subclassedWindows);
+        g_subclassedWindows.clear();
+    }
+
+    for (HWND hWnd : subclassedWindows)
+        SendMessage(hWnd, g_unsubclassRegisteredMessage, 0, 0);
```
m417z: Use `WindhawkUtils::RemoveWindowSubclassFromAnyThread`. `g_unsubclassRegisteredMessage` should no longer be needed.

mods/translucent-windows.wh.cpp
m417z: Maybe related to asynchronous injection? Relevant discussion: https://github.com/ramensoftware/windhawk/issues/103.

mods/translucent-windows.wh.cpp
Undisputed00x: Done, the problem stays with the border colors /w blur behind when launching the control panel. It seems that something is wrong with the logic in HookedDwmSetWindowAttribute

mods/translucent-windows.wh.cpp
Undisputed00x: Wow interesting, but thankfully its not related to such thing, only a logic bug.

mods/translucent-windows.wh.cpp
```diff
@@ -515,6 +562,77 @@ BOOL IsWindowClass(HWND hWnd, LPCWSTR ClassName)
...
+                        g_settings.g_BorderColor = g_settings.BorderInactiveColor;
+                        DwmSetWindowAttribute(hWnd, BORDER_COLOR, &g_settings.BorderInactiveColor, sizeof(g_settings.BorderInactiveColor));
+                    }
+                }
+            }
+            break;
+        }
+        case WM_NCDESTROY:
+        {
+            UnsubclassWindow(hWnd);
```
m417z: No need to unsubclass here, it's enough to do:
```suggestion
            std::lock_guard<std::mutex> guard(g_subclassedWindowsMutex);
            g_subclassedWindows.erase(hWnd);
```

mods/translucent-windows.wh.cpp
```diff
@@ -542,6 +660,10 @@ void NewWindowShown(HWND hWnd)
     }
     else
         EnableColoredTitlebar(hWnd);
+    
+    if(g_settings.BorderFlag || g_settings.CaptionTextFlag || g_settings.TitlebarFlag)
+        if(WindhawkUtils::SetWindowSubclassFromAnyThread(hWnd, SubclassProc, 0))
+            g_subclassedWindows.insert(hWnd);
```
m417z, line 656: Since you're using a lock:
```suggestion
        {
            std::lock_guard<std::mutex> guard(g_subclassedWindowsMutex);
            g_subclassedWindows.insert(hWnd);
        }
```

mods/translucent-windows.wh.cpp
```diff
@@ -849,6 +984,16 @@ void Wh_ModUninit(void)
...
+
+    std::unordered_set<HWND> subclassedWindows;
+    {
+        std::lock_guard<std::mutex> guard(g_subclassedWindowsMutex);
+        subclassedWindows = std::move(g_subclassedWindows);
+        g_subclassedWindows.clear();
+    }
+
+    for (HWND hWnd : subclassedWindows)
+        UnsubclassWindow(hWnd);
```
m417z: Since `g_subclassedWindows` is empty at this point, here it's enough to do:
```suggestion
        WindhawkUtils::RemoveWindowSubclassFromAnyThread(hWnd, SubclassProc);
```

mods/translucent-windows.wh.cpp
```diff
@@ -127,46 +141,81 @@ maximized or snapped to the edge of the screen, this is caused by default by the
...
+static UINT ENABLE = 1;
+static constexpr UINT SYSTEMBACKDROP_TYPE = 38; // DWMWA_SYSTEMBACKDROP_TYPE
+static constexpr UINT NONE = 1; // DWMSBT_NONE
+static constexpr UINT MAINWINDOW = 2; // DWMSBT_MAINWINDOW
+static constexpr UINT TRANSIENTWINDOW = 3; // DWMSBT_TRANSIENTWINDOW
+static constexpr UINT TABBEDWINDOW = 4; // DWMSBT_TABBEDWINDOW
+static constexpr UINT COLOR_DEFAULT = 0xFFFFFFFF; // DWMWA_COLOR_DEFAULT
+static constexpr UINT COLOR_NONE = 0xFFFFFFFE; // DWMWA_COLOR_NONE
+
+UINT g_unsubclassRegisteredMessage = RegisterWindowMessage(L"Windhawk_Unsubclass");
```
m417z:
```suggestion
```

mods/translucent-windows.wh.cpp
```diff
@@ -515,6 +562,78 @@ BOOL IsWindowClass(HWND hWnd, LPCWSTR ClassName)
...
+    WindhawkUtils::RemoveWindowSubclassFromAnyThread(hWnd, SubclassProc);
+
+    std::lock_guard<std::mutex> guard(g_subclassedWindowsMutex);
+
+    auto it = g_subclassedWindows.find(hWnd);
+    if (it != g_subclassedWindows.end()) {
+        g_subclassedWindows.erase(it);
+    }
+}
+
```
m417z:
```suggestion
```

PR #1798, opened 2025-04-15
Windows 7/8.x Alt+Tab Loader 2.1.0

aubymori:
- Remove the `LoadStringW` hook hack. Users must now supply an MUI file.
- Add error message boxes for AltTab.dll or its string resources failing to load.
- Support Windows 8.x's AltTab.dll
  - Change mod name to reflect this
- Slightly revise README to be more clear and accurate

### DIFF
--- a/mods/win7-alttab-loader.wh.cpp
+++ b/mods/win7-alttab-loader.wh.cpp
@@ -1,8 +1,8 @@
 // ==WindhawkMod==
 // @id              win7-alttab-loader
-// @name            Windows 7 Alt+Tab Loader
-// @description     Loads Windows 7 Alt+Tab on Windows 10.
-// @version         2.0.0
+// @name            Windows 7/8.x Alt+Tab Loader
+// @description     Loads Windows 7/8.x Alt+Tab on Windows 10.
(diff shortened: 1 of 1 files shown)


aubymori:
Believe that PR mod validation failing is due to the fact that AltTab.dll was removed in Windows 10. Should be safe to ignore.

kawapure:
> Believe that PR mod validation failing is due to the fact that AltTab.dll was removed in Windows 10. Should be safe to ignore.

Yes, the list is validated against Winbindex (https://winbindex.m417z.com/data/filenames.json) which does not acknowledge this file name. It is valid for Windows versions prior to 10 (which Winbindex does not acknowledge), and the symbol server still hosts symbols for old versions of Windows.

kawapure:
By the way, in my opinion, the extent of the changes justifies a version bump to 3.0.0.

m417z:
Let me know when it's ready to be merged.

kawapure:
Aubrey went to bed, but if my comment had interrupted the merge process, then it's not a problem. Aubrey explained to me why she didn't bump the version number that high to me on IM.

PR #1807, opened 2025-04-19
add force-chinese-ime-alt1

barrypp:
Use another method (set_IME_mode trigger by MSUIM.Msg.Private) to trigger the IME configuration process. The previous method does not work on my computer


barrypp:
@u3l6

u3l6:
I know the previous method I used is not working now. And I'm not capable of maintaining force-chinese-ime🥲. I'm glad that you can update it

barrypp:
thanks. 

I feel like my code needs performance improvements. It works fine, but I feel like hooking DispatchMessage consumes too much CPU
Also, I'm not sure if the MSUIM.Msg.Private message ID is the same on different computers.

u3l6:
At least your code really works.

I released force-chinese-ime without sufficient testing, and then I found that it seems not to be working as I expected, and I'm not able to remove the mod. I'm still hoping to remove the old version if @m417z allows

barrypp:
Have you tried the code? I want to know if the ID of MSUIM.Msg.Private is consistent on different computers. On my computer, this value does not change after restart.

u3l6:
ok I’ll try it later

m417z:
You can ignore the 1.4.1 version failure. Let me know if you want to have it merged or if you want to do more testing.

mods/force-chinese-ime-alt1.wh.cpp
```diff
@@ -0,0 +1,147 @@
...
+using RegisterWindowMessageW_t = decltype(&RegisterWindowMessageW);
+RegisterWindowMessageW_t RegisterWindowMessageW_Original;
+UINT WINAPI RegisterWindowMessageW_Hook(LPCWSTR lpString) 
+{
+    UINT r = RegisterWindowMessageW_Original(lpString);
+    if (0 == lstrcmpW(lpString,L"MSUIM.Msg.Private"))
+    {
+        Wh_Log(L"r, %x, msg, %s",r,lpString);
+        WM_MSUIM_Msg_Private = r;
+        Wh_RemoveFunctionHook((void*)RegisterWindowMessageW);
```
m417z: This function doesn't do anything without also calling `Wh_ApplyHookOperations`.

mods/force-chinese-ime-alt1.wh.cpp
```diff
@@ -0,0 +1,147 @@
...
+#include <windhawk_api.h>
+#include <windhawk_utils.h>
+#include <windows.h>
+
+// Define IME control constants if not defined
+#ifndef IMC_SETCONVERSIONMODE
+#define IMC_SETCONVERSIONMODE 0x0002
+#endif
+
+UINT WM_MSUIM_Msg_Private = 0xc05a;
```
m417z: Why not simply do `UINT WM_MSUIM_Msg_Private = RegisterWindowMessageW(L"MSUIM.Msg.Private");`? Then the hook also won't be necessary.

Barrypp-zzx: Nice, didn't know it could be used this way

barrypp:
@m417z  code updated.

m417z:
Great. How about having a more descriptive name and readme?

For the name, instead of "Force Chinese IME Mode - alt1" it can be something like "Force Chinese IME Mode - Improved".

And I assume you can base your readme on the one from the original mod. But the screenshots were removed there unfortunately.

barrypp:
how about "Chinese IME fixed mode" as the name. @m417z

barrypp:
@m417z code updated.

PR #1820, opened 2025-04-24
Mark force-chinese-ime as deprecated

u3l6:


m417z:
Please update the version number too.

PR #1832, opened 2025-04-28
New MOD "edge-hot-corner-desktop-switch"

sanskarprasad:
Switch the desktop by moving the cursor to the edge of the screen.

### DIFF
--- a/mods/edge-hot-corner-desktop-switch.wh.cpp
+++ b/mods/edge-hot-corner-desktop-switch.wh.cpp
@@ -162,8 +162,13 @@ DWORD WINAPI MonitorThread(LPVOID) {
 }
 
 BOOL Wh_ModInit() {
-    if (!FindCurrentProcessTaskbarWnd())
-        return FALSE;
+    if (!FindCurrentProcessTaskbarWnd()) {
+        HWND hTaskbarWnd = FindWindowW(L"Shell_TrayWnd", nullptr);
+        if (hTaskbarWnd) {
+            // The taskbar exists, but it's not owned by the current process.
+            return FALSE;
+        }
+    }
 
     Wh_Log(L"EdgeHotCorner: Initializing mod");
     LoadSettings();
--- a/mods/edge-hot-corner-desktop-switch.wh.cpp
+++ b/mods/edge-hot-corner-desktop-switch.wh.cpp
@@ -4,7 +4,7 @@
 // @description     Switches virtual desktops when the mouse hovers at the left or right screen edge for a moment.
 // @version         0.1
 // @author          You
-// @github          https://github.com/yourusername/edge-hot-corner-desktop-switch
+// @github          https://github.com/sanskarprasad
 // @homepage        https://your-personal-homepage.example.com/edge-hot-corner-desktop-switch
 // @include         explorer.exe
 // ==/WindhawkMod==
--- a/mods/edge-hot-corner-desktop-switch.wh.cpp
+++ b/mods/edge-hot-corner-desktop-switch.wh.cpp
@@ -70,7 +70,6 @@ void SimulateDesktopSwitch(int direction) {
     HWND hTaskbarWnd = FindWindowW(L"Shell_TrayWnd", nullptr);
     if (hTaskbarWnd) {
         SetForegroundWindow(hTaskbarWnd);
-        Sleep(50);
     }
 
     WORD vkDir = (direction < 0 ? VK_LEFT : VK_RIGHT);
--- a/mods/edge-hot-corner-desktop-switch.wh.cpp
+++ b/mods/edge-hot-corner-desktop-switch.wh.cpp
@@ -1,0 +1,173 @@
+// ==WindhawkMod==
+// @id              edge-hot-corner-desktop-switch
+// @name            Edge Hot-Corners Desktop Switch
+// @description     Switches virtual desktops when the mouse hovers at the left or right screen edge for a moment.
+// @version         0.1
+// @author          You
+// @github          https://github.com/yourusername/edge-hot-corner-desktop-switch
+// @homepage        https://your-personal-homepage.example.com/edge-hot-corner-desktop-switch
+// @include         explorer.exe
+// ==/WindhawkMod==
+
+// ==WindhawkModSettings==
+/*
+- thresholdMs: 1000
+  $name: Time threshold at edge (ms)
+  $description: >
+    Milliseconds the cursor must stay in the hot zone before
+    switching desktops.
+
+- edgeWidth: 5
+  $name: Detection edge width (px)
+  $description: >
+    Width in pixels of the left/right screen hot zone.
+*/
+// ==/WindhawkModSettings==
+
+#include <Windows.h>
+
+// #include <windhawk-sdk.h>
+
+struct {
+    int thresholdMs;
+    int edgeWidth;
+} settings;
+
+volatile bool g_Running = false;
+HANDLE g_hThread = nullptr;
+
+void LoadSettings() {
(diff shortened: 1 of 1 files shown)


mods/edge-hot-corner-desktop-switch.wh.cpp
```diff
@@ -0,0 +1,173 @@
+// ==WindhawkMod==
+// @id              edge-hot-corner-desktop-switch
+// @name            Edge Hot-Corners Desktop Switch
+// @description     Switches virtual desktops when the mouse hovers at the left or right screen edge for a moment.
+// @version         0.1
+// @author          Sanskar Prasad
+// @github          https://github.com/sanskarprasad
+// @homepage        https://your-personal-homepage.example.com/edge-hot-corner-desktop-switch
```
m417z: This is not a valid website, update it or remove it.

mods/edge-hot-corner-desktop-switch.wh.cpp
```diff
@@ -0,0 +1,173 @@
...
+    inputs[4].ki.wVk = VK_CONTROL;
+    inputs[4].ki.dwFlags = KEYEVENTF_KEYUP;
+
+    // Win up   (extended + key-up)
+    inputs[5].type = INPUT_KEYBOARD;
+    inputs[5].ki.wVk = VK_LWIN;
+    inputs[5].ki.dwFlags = KEYEVENTF_EXTENDEDKEY | KEYEVENTF_KEYUP;
+
+    // Send them all in one go
+    UINT sent = SendInput(_countof(inputs), inputs, sizeof(INPUT));
```
m417z, line 101: This won't work if an elevated program is focused, such as Windhawk or Task Manager.
As a workaround, you can set the taskbar as the foreground window before calling `SendInput`:
https://github.com/ramensoftware/windhawk-mods/blob/8c63a08f1a5c540f9061b4a896dfdf18a0bcd843/mods/taskbar-scroll-actions.wh.cpp#L674-L678

mods/edge-hot-corner-desktop-switch.wh.cpp
```diff
@@ -0,0 +1,173 @@
...
+    Wh_Log(L"EdgeHotCorner",
+           L"SimulateDesktopSwitch(%s): SendInput injected %u events",
+           (direction < 0 ? L"Left" : L"Right"), sent);
```
m417z: Your usage of `Wh_Log` is incorrect, only `EdgeHotCorner` is printed.

mods/edge-hot-corner-desktop-switch.wh.cpp
```diff
@@ -0,0 +1,173 @@
...
+    return 0;
+}
+
+BOOL Wh_ModInit() {
+    Wh_Log(L"EdgeHotCorner", L"Initializing mod");
+    LoadSettings();
+
+    // Start the monitoring thread
+    g_Running = true;
+    g_hThread = CreateThread(nullptr, 0, MonitorThread, nullptr, 0, nullptr);
```
m417z, line 176: There can be more than one `explorer.exe` process, e.g. if the "Launch Folder Windows in a Separate Process" option is enabled.
You can make sure the mod is only loaded in the process that owns the taskbar. You can run something like this at the beginning of `Wh_ModInit`:
```cpp
if (!FindCurrentProcessTaskbarWnd()) {
  return FALSE;
}
```
The function code:
https://github.com/m417z/my-windhawk-mods/blob/04f0edd8a7b12d0875c1861a294336d968508d73/mods/icon-resource-redirect.wh.cpp#L2175-L2194

m417z:
Nice mod. I left several comments.
I'm also not sure about the performance and energy consumption implications of running a constant loop each 20 milliseconds. I'd increase the delay (at least when `enterTime` is zero), and also would skip all the checks if the mouse pos doesn't change, which is the common case.

mods/edge-hot-corner-desktop-switch.wh.cpp
```diff
@@ -0,0 +1,193 @@
...
+    settings.thresholdMs = Wh_GetIntSetting(L"thresholdMs");
+    settings.edgeWidth   = Wh_GetIntSetting(L"edgeWidth");
+}
+
+void SimulateDesktopSwitch(int direction) {
+    // Force-focus the taskbar so SendInput will be accepted
+    HWND hTaskbarWnd = FindWindowW(L"Shell_TrayWnd", nullptr);
+    if (hTaskbarWnd) {
+        SetForegroundWindow(hTaskbarWnd);
+        Sleep(50);
```
m417z: The sleep wasn't necessary in my tests. Did you need to add it, or did you add it just in case?

mods/edge-hot-corner-desktop-switch.wh.cpp
```diff
@@ -0,0 +1,193 @@
...
+            zone = 0;
+        }
+    }
+
+    return 0;
+}
+
+BOOL Wh_ModInit() {
+    if (!FindCurrentProcessTaskbarWnd())
+        return FALSE;
```
m417z: Actually there's a small problem with this: if the first explorer process is starting, the taskbar won't be created yet, and so the mod won't be loaded. How about:

```suggestion
    if (!FindCurrentProcessTaskbarWnd()) {
        HWND hTaskbarWnd = FindWindowW(L"Shell_TrayWnd", nullptr);
        if (hTaskbarWnd) {
            // The taskbar exists, but it's not owned by the current process.
            return FALSE;
        }
    }
```

mods/edge-hot-corner-desktop-switch.wh.cpp
Sanskaaar1: Thought that the
SetForegroundWindow(hTaskbarWnd) requests a focus change, but the actual OS processing can be asynchronous.

Immediately calling SendInput() after SetForegroundWindow() can fail if the foreground window hasn't actually changed yet.

m417z: Interesting, according to this blog post it should work:
https://devblogs.microsoft.com/oldnewthing/20161118-00/?p=94745

> It immediately sets the input queue associated with the window as being the foreground input queue. Among other thing, it means that keyboard input will be directed to that input queue.

sanskarprasad:
hey! i didnt quiet understand the check errors above!
could you please guide me

m417z:
I think it's a bug in the script which should already be fixed, let me close and re-open the PR to re-run it.

mods/edge-hot-corner-desktop-switch.wh.cpp
Sanskaaar1: Thank you for the reference — that blog post was very insightful.

sanskarprasad:
oh great!

mods/edge-hot-corner-desktop-switch.wh.cpp
m417z: I think it's ready to be merged. Would you like to leave the Sleep?

sanskarprasad:
lets remove the sleep(50).

mods/edge-hot-corner-desktop-switch.wh.cpp
Sanskaaar1: Your call!

m417z: I'd remove it if it was my mod, as it adds an unnecessary delay, but it's your mod. Let me know when the mod is ready to be merged.

Sanskaaar1: I've removed it
lets merge

sanskarprasad:
> Congratulations on your first mod! :)

Thank you for guiding me.
would love to contribute more.

PR #1845, opened 2025-05-01
fix issue for fullscreen apps

sanskarprasad:
Now the mod will not work when in full screen.


m417z:
Here's a function you can use to detect fullscreen:

```cpp
bool IsInFullScreenMode() {
    QUERY_USER_NOTIFICATION_STATE pquns;
    if (FAILED(SHQueryUserNotificationState(&pquns))) {
        return false;
    }

    switch (pquns) {
        case QUNS_NOT_PRESENT:
        case QUNS_BUSY:
        case QUNS_RUNNING_D3D_FULL_SCREEN:
            return true;
    }

    return false;
}
```

PR #1854, opened 2025-05-02
now detects the entire screen real-estate instead of a single screen

sanskarprasad:


m417z:
This means it will stop working on setups like this:

image

Making it work perfectly can be tricky, but perhaps it can at least be an option for either per-monitor edges or global edges. What do you think?

sanskarprasad:
Is there a way to detect this orientation(setup) like this?
Then simple if-else would work
, or we have to consider giving the user an option (per-monitor or global)

m417z:
I don't think there's an easy way to detect it. But even if it was, I can imagine that some users might prefer to have these areas only:

image

And some might prefer these (e.g. if they mostly use monitor 1):

image

The ideal is probably to have this, but it can be tricky to support any monitor arrangement this way:

image

sanskarprasad:
Yes and there can be a function for detecting if there's a change in the orientation of the monitors.

m417z:
I see that you pushed an update, let me know if it's ready for review/merge.

PR #1856, opened 2025-05-02
Add winmark-disabler.wh.cpp v1.0

nukedrust-pixel:


m417z:
As we discussed on Discord, this is not a good fit for a Windhawk mod.

PR #1859, opened 2025-05-04
Make Task Manager Tabs Taller

JoeYe-233:
Original Windows 10 Task Manager tab labels are too small on high DPI screens, making them difficult to click, this mod makes them taller.

taskmgr-taller-tab-before-after

### DIFF
--- a/mods/taskmgr-taller-tab.wh.cpp
+++ b/mods/taskmgr-taller-tab.wh.cpp
@@ -1,0 +1,75 @@
+// ==WindhawkMod==
+// @id              taskmgr-taller-tab
+// @name            Make Task Manager Tabs Taller
+// @description     Windows 10 Task Manager tab labels are too small on high DPI screens, making them difficult to click, this mod makes them taller.
+// @name:zh-CN            增大任务管理器选项卡尺寸
+// @description:zh-CN   Windows 10 任务管理器选项卡标签在高分屏上实在太小, 根本不好点击, 安装后会增大选项卡标签的高度.让其更易于使用.
+// @version         1.0
+// @author          Joe Ye
+// @github          https://github.com/JoeYe-233
+// @include         Taskmgr.exe
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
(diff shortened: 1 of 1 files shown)


mods/taskmgr-taller-tab.wh.cpp
```diff
@@ -0,0 +1,75 @@
...
+    HMENU hMenu,
+    HINSTANCE hInstance,
+    LPVOID lpParam
+)
+{
+    HWND hWnd = CreateWindowExW_Original(
+        dwExStyle, lpClassName, lpWindowName, dwStyle,
+        x, y, nWidth, nHeight, hWndParent, hMenu, hInstance, lpParam
+    );
+    TabCtrl_SetItemSize(hWnd, 100, 37); // 宽度为 100，高度为 37
```
m417z, line 58: Wouldn't it be better to check some properties of the window here, such as the class name, just to be safe?

JoeYe-233: Yes, I actually tried to do so, however the "lpClassName" variable seems not readable by WindHawk. Any operation invoving reading the variable (even like trying to print it out with Wh_Log) resulted in an "Access violation: unreadable" exception and TaskManager just crashed😥

I'm not quite sure if there is any better way to do this 🤔, but _technically_ since there is only one "SysTabControl32" instance in TaskManager, it shouldn't cause a lot trouble (TabCtrl_SetItemSize only has effect on SysTabControl32😋)

Access Violation

m417z: You need to check that is a textual class name before accessing it. Example:
https://github.com/ramensoftware/windhawk-mods/blob/f0a0be743d492524840adb9bbd51b3b56ae901f2/mods/explorer-name-windows.wh.cpp#L346

JoeYe-233: That's cool. I'll look into it, thanks!

JoeYe-233: It finally works! Thanks so much. I'll close this PR and open a new one!

PR #1860, opened 2025-05-04
Make Task Manager Tabs Taller

JoeYe-233:
Windows 10 Task Manager tab labels are too small on high DPI screens, making them difficult to click. This mod makes them taller and easier to click.

taskmgr-taller-tab-before-after


m417z:
Congratulations on your first mod :)

PR #1898, opened 2025-05-13
Pill Taskbar

EMs1-Dev:
First release


m417z:
Thanks for the submission. Why is it a separate mod and not just a theme for Windows 11 Taskbar Styler?

Also, it looks very similar to the existing TaskbarXII theme.

PR #1913, opened 2025-05-17
added mod auto-hide-desktop-after-inactivity.wh.cpp

Alekseev2014s:
Automatically hides all windows and shows the desktop after a period of user inactivity

### DIFF
--- a/mods/auto-hide-desktop-after-inactivity.wh.cpp
+++ b/mods/auto-hide-desktop-after-inactivity.wh.cpp
@@ -1,0 +1,327 @@
+// ==WindhawkMod==
+// @id              auto-hide-desktop-after-inactivity
+// @name            Auto Hide Desktop After Inactivity
+// @description     Automatically hides all windows and shows the desktop after a period of user inactivity
+// @version         0.5
+// @author          alekseev2014s
+// @github          https://github.com/Alekseev2014s
+// @include         explorer.exe
+// @compilerOptions -lcomdlg32
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Auto Hide Desktop After Inactivity
+
+This mod automatically minimizes all open windows and hides the desktop
+after a period of user inactivity. It's useful for shared environments,
+kiosks, or simply to reduce visual clutter after you're away from your PC.
+
+## Features
+- **Inactivity timeout**: Automatically hide the desktop after no input for a configurable number of seconds.
+- **Lock screen timeout**: Optionally lock the workstation after a longer period of inactivity.
+- **Fullscreen app detection**: Prevents the desktop from being hidden if a fullscreen application is active.
+- **Process exclusions**: You can specify a list of executable names (like `vlc.exe`) to ignore during monitoring.
+
+## How it works
+- When the system is idle for the specified timeout, the mod simulates a `Win+D` keystroke to hide all windows and show the desktop.
+- If configured, after further inactivity, it automatically locks the workstation.
+- User input such as mouse movement or keyboard activity restores the previous window state.
+- Foreground fullscreen applications and excluded processes will prevent the desktop from being hidden.
+
+## Getting started
+1. Install the mod in Windhawk.
+2. Adjust the settings for:
+   - Timeout (in seconds)
+   - Lock timeout (optional)
+   - Excluded processes
+   - Fullscreen app handling
+3. Let the system idle and watch the desktop auto-hide behavior in action.
+
+## Example use cases
+- **Privacy**: Automatically hides sensitive windows when away.
+- **Kiosk mode**: Resets to desktop after inactivity.
+- **Clutter reduction**: Keeps your desktop clean when you're not using the PC.
+
+## Troubleshooting
+- If the desktop is not hidden as expected, check that no fullscreen app or excluded process is active.
+- To ensure it only runs once, the mod uses a global mutex to avoid interfering with other `explorer.exe` instances.
+*/
+// ==/WindhawkModReadme==
+
+// ==WindhawkModSettings==
+/*
+- timeoutSeconds: 300
+  $name: Inactivity timeout (seconds)
+  $description: The number of seconds of inactivity before hiding the desktop.
+
+- lockTimeoutSeconds: 600
+  $name: Lock screen timeout (seconds)
+  $description: The number of seconds of inactivity before locking the workstation.
+
+- excludeProcesses: ""
+  $name: Excluded processes
+  $description: Semicolon-separated list of executable names to exclude from triggering desktop show (e.g. "vlc.exe;notepad.exe")
+
+- ignoreFullscreenApps: true
+  $name: Ignore fullscreen applications
+  $description: Prevent desktop from being shown if a fullscreen application is active.
+*/
+// ==/WindhawkModSettings==
+
+#include <windows.h>
+#include <vector>
+
+std::vector<HWND> g_minimizedWindows;
+std::vector<std::wstring> excludedProcessList;
+
+DWORD setting_timeoutSeconds = 300;
+DWORD setting_lockTimeoutSeconds = 600;
+bool setting_ignoreFullscreenApps = true;
+
+constexpr DWORD kMouseMoveThreshold = 10;
+
+HANDLE hThread = nullptr;
+bool stopThread = false;
+bool desktopShownByMod = false;
+POINT lastMousePos = {};
+
+HANDLE g_singletonMutex = nullptr;
+
+bool lockPending = false;
+DWORD timeDesktopShown = 0;
+
+enum class DesktopState {
+    Normal,
+    HiddenByMod
+};
+
+DesktopState desktopState = DesktopState::Normal;
+
+bool IsDesktopVisible() {
+    HWND shellWindow = GetShellWindow();
+    HWND foreground = GetForegroundWindow();
+    return foreground == shellWindow;
+}
+
+void ToggleDesktopWithWinD() {
+    keybd_event(VK_LWIN, 0, 0, 0);           // Win down
+    keybd_event('D', 0, 0, 0);               // D down
+    keybd_event('D', 0, KEYEVENTF_KEYUP, 0); // D up
+    keybd_event(VK_LWIN, 0, KEYEVENTF_KEYUP, 0); // Win up
+}
+
+void HideDesktop() {
+    if (!IsDesktopVisible()) {
+        ToggleDesktopWithWinD();
+    }
+    GetCursorPos(&lastMousePos);
+    desktopShownByMod = true;
+    timeDesktopShown = GetTickCount();
+    lockPending = (setting_lockTimeoutSeconds > 0);
+}
+
+void RestoreWindows() {
+    if (IsDesktopVisible()) {
+        ToggleDesktopWithWinD();
+    }
+
+    desktopShownByMod = false;
+}
+
+void LoadSettings() {
+    int value;
+    value = Wh_GetIntSetting(L"timeoutSeconds");
+    setting_timeoutSeconds = (value > 0) ? value : 300;
+
+    value = Wh_GetIntSetting(L"lockTimeoutSeconds");
+    setting_lockTimeoutSeconds = (value >= 0) ? value : 0;
+
+    setting_ignoreFullscreenApps = Wh_GetIntSetting(L"ignoreFullscreenApps");
+
+    const wchar_t* excludeList = Wh_GetStringSetting(L"excludeProcesses");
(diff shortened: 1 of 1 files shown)


mods/auto-hide-desktop-after-inactivity.wh.cpp
```diff
@@ -0,0 +1,327 @@
...
+
+        Sleep(200);
+    }
+
+    Wh_Log(L"Monitor thread exiting");
+    return 0;
+}
+
+bool IsMainExplorerProcess() {
+    g_singletonMutex = CreateMutexW(nullptr, TRUE, L"Global\\WH_AutoShowDesktop_Mutex");
```
m417z: This will just pick the first Explorer process. If it turns out to be a folder process, once it's closed, the mod will stop working.

How about the following solution instead?

https://github.com/ramensoftware/windhawk-mods/blob/9d7df8f53b5d834fbed6034c7c7a2dd03033f66a/mods/edge-hot-corner-desktop-switch.wh.cpp#L184-L191

mods/auto-hide-desktop-after-inactivity.wh.cpp
```diff
@@ -0,0 +1,327 @@
...
+                if (dx >= kMouseMoveThreshold || dy >= kMouseMoveThreshold) {
+                    Wh_Log(L"Restoring windows unlock input");
+                    RestoreWindows();
+                    desktopState = DesktopState::Normal;
+                    effectiveLastInputTime = GetTickCount();
+                }
+            }
+        }
+
+        Sleep(200);
```
m417z: This can be improved. If the timeout is 600 seconds, it'll be looping 3000 times after the last activity, although it can be easily calculated from the code that you need to sleep at least 600 seconds.

The code can be something like (pseudo-code):
```cpp
int msToSleep = min(lockTimeoutSeconds, timeoutSeconds) * 1000 - idleTime;
if (msToSleep > 0)
    Sleep(msToSleep);
```

mods/auto-hide-desktop-after-inactivity.wh.cpp
```diff
@@ -0,0 +1,327 @@
...
+            }
+        }
+    }
+
+    CloseHandle(hProcess);
+    return result;
+}
+
+bool IsForegroundWindowFullscreen() {
+    HWND hwnd = GetForegroundWindow();
```
m417z: There's a WinAPI function which allows to query the fullscreen state:
https://github.com/ramensoftware/windhawk-mods/blob/9d7df8f53b5d834fbed6034c7c7a2dd03033f66a/mods/edge-hot-corner-desktop-switch.wh.cpp#L69-L83

mods/auto-hide-desktop-after-inactivity.wh.cpp
```diff
@@ -0,0 +1,327 @@
...
+  $name: Inactivity timeout (seconds)
+  $description: The number of seconds of inactivity before hiding the desktop.
+
+- lockTimeoutSeconds: 600
+  $name: Lock screen timeout (seconds)
+  $description: The number of seconds of inactivity before locking the workstation.
+
+- excludeProcesses: ""
+  $name: Excluded processes
+  $description: Semicolon-separated list of executable names to exclude from triggering desktop show (e.g. "vlc.exe;notepad.exe")
```
m417z, line 68: Why not use an array instead? For example:

https://github.com/ramensoftware/windhawk-mods/blob/9d7df8f53b5d834fbed6034c7c7a2dd03033f66a/mods/taskbar-grouping.wh.cpp#L93-L94

mods/auto-hide-desktop-after-inactivity.wh.cpp
Alekseev2014s: an ambiguous moment, because after minimizing windows during an event, you need to quickly display the windows

mods/auto-hide-desktop-after-inactivity.wh.cpp
m417z: Update the description, it's no longer a semicolon-separated list.

mods/auto-hide-desktop-after-inactivity.wh.cpp
```diff
@@ -0,0 +1,408 @@
...
+
+    setting_ignoreFullscreenApps = Wh_GetIntSetting(L"ignoreFullscreenApps");
+
+    excludedProcessList.clear();
+
+    for (int i = 0;; i++) {
+        wchar_t settingName[64];
+        swprintf(settingName, ARRAYSIZE(settingName), L"excludedPrograms[%d]", i);
+
+        const wchar_t* item = Wh_GetStringSetting(settingName);
```
m417z, line 182: You never free it with `Wh_FreeStringSetting`.

mods/auto-hide-desktop-after-inactivity.wh.cpp
m417z: In the most recent update, you changed some other description, but not this one (search for "Semicolon" in the code).

mods/auto-hide-desktop-after-inactivity.wh.cpp
m417z: You need to call it even when an empty string is returned.

m417z:
This mod has no good reason to run as part of explorer.exe, except that it's convenient. We discussed it here: https://github.com/ramensoftware/windhawk-mods/pull/1916#issuecomment-2889149966

> This mod is injected into explorer and runs as part of it. This is suboptimal for the following reasons:
> * There can be more than one explorer.exe process in case the "launch folder windows in separate process" option is enabled, and for other reasons.
> * This mod doesn't have a good reason to be injected into explorer.exe, it's just convenient. But by doing so, a mistake or instability in the mod affects the whole shell.
> 
> This is not the first mod to take this approach. Here are other recent mods:
> * https://github.com/ramensoftware/windhawk-mods/pull/1832
> * https://github.com/ramensoftware/windhawk-mods/pull/1913
> 
> And maybe also:
> * https://windhawk.net/mods/classic-theme-enable
> * https://windhawk.net/mods/magnifier-mod
> 
> All three mods could be separate tools, they don't use and don't need Windhawk's injection and hooking capabilities. On the other hand, I can understand that the ease of mod installation is appealing. I wonder if Windhawk needs to come up with a new type of mods which run as part of a dedicated process.

Currently, Windhawk doesn't have a built-in functionality for something like that, but for now, I implemented some code that can be added to a mod to make it run as part of a separate windhawk.exe process:

<details><summary>Details</summary>
<p>

```cpp
////////////////////////////////////////////////////////////////////////////////
// Windhawk tool mod implementation for mods which don't need to inject to other
// processes or hook other functions. Context:
// https://github.com/ramensoftware/windhawk-mods/pull/1916
//
// The mod will load and run in a dedicated windhawk.exe process.
//
// Paste the code below as part of the mod code, and use these callbacks:
// * WhTool_ModInit
// * WhTool_ModSettingsChanged
// * WhTool_ModUninit
//
// Currently, other callbacks are not supported.

bool g_isToolModProcessLauncher;
HANDLE g_toolModProcessMutex;

void WINAPI EntryPoint_Hook() {
    Wh_Log(L">");
    ExitThread(0);
}

BOOL Wh_ModInit() {
    bool isService = false;
    bool isToolModProcess = false;
    bool isCurrentToolModProcess = false;
    int argc;
    LPWSTR* argv = CommandLineToArgvW(GetCommandLine(), &argc);
    if (!argv) {
        Wh_Log(L"CommandLineToArgvW failed");
        return FALSE;
    }

    for (int i = 1; i < argc; i++) {
        if (wcscmp(argv[i], L"-service") == 0) {
            isService = true;
            break;
        }
    }

    for (int i = 1; i < argc - 1; i++) {
        if (wcscmp(argv[i], L"-tool-mod") == 0) {
            isToolModProcess = true;
            if (wcscmp(argv[i + 1], WH_MOD_ID) == 0) {
                isCurrentToolModProcess = true;
            }
            break;
        }
    }

    LocalFree(argv);

    if (isService) {
        return FALSE;
    }

    if (isCurrentToolModProcess) {
        g_toolModProcessMutex =
            CreateMutex(nullptr, TRUE, L"windhawk-tool-mod_" WH_MOD_ID);
        if (!g_toolModProcessMutex) {
            Wh_Log(L"CreateMutex failed");
            ExitProcess(1);
        }

        if (GetLastError() == ERROR_ALREADY_EXISTS) {
            Wh_Log(L"Tool mod already running (%s)", WH_MOD_ID);
            ExitProcess(1);
        }

        if (!WhTool_ModInit()) {
            ExitProcess(1);
        }

        IMAGE_DOS_HEADER* dosHeader =
            (IMAGE_DOS_HEADER*)GetModuleHandle(nullptr);
        IMAGE_NT_HEADERS* ntHeaders =
            (IMAGE_NT_HEADERS*)((BYTE*)dosHeader + dosHeader->e_lfanew);

        DWORD entryPointRVA = ntHeaders->OptionalHeader.AddressOfEntryPoint;
        void* entryPoint = (BYTE*)dosHeader + entryPointRVA;

        Wh_SetFunctionHook(entryPoint, (void*)EntryPoint_Hook, nullptr);
        return TRUE;
    }

    if (isToolModProcess) {
        return FALSE;
    }

    g_isToolModProcessLauncher = true;
    return TRUE;
}

void Wh_ModAfterInit() {
    if (!g_isToolModProcessLauncher) {
        return;
    }

    WCHAR currentProcessPath[MAX_PATH];
    switch (GetModuleFileName(nullptr, currentProcessPath,
                              ARRAYSIZE(currentProcessPath))) {
        case 0:
        case ARRAYSIZE(currentProcessPath):
            Wh_Log(L"GetModuleFileName failed");
            return;
    }

    WCHAR
    commandLine[MAX_PATH + 2 +
                (sizeof(L" -tool-mod \"" WH_MOD_ID "\"") / sizeof(WCHAR)) - 1];
    swprintf_s(commandLine, L"\"%s\" -tool-mod \"%s\"", currentProcessPath,
               WH_MOD_ID);

    HMODULE kernelModule = GetModuleHandle(L"kernelbase.dll");
    if (!kernelModule) {
        kernelModule = GetModuleHandle(L"kernel32.dll");
        if (!kernelModule) {
            Wh_Log(L"No kernelbase.dll/kernel32.dll");
            return;
        }
    }

    using CreateProcessInternalW_t = BOOL(WINAPI*)(
        HANDLE hUserToken, LPCWSTR lpApplicationName, LPWSTR lpCommandLine,
        LPSECURITY_ATTRIBUTES lpProcessAttributes,
        LPSECURITY_ATTRIBUTES lpThreadAttributes, WINBOOL bInheritHandles,
        DWORD dwCreationFlags, LPVOID lpEnvironment, LPCWSTR lpCurrentDirectory,
        LPSTARTUPINFOW lpStartupInfo,
        LPPROCESS_INFORMATION lpProcessInformation,
        PHANDLE hRestrictedUserToken);
    CreateProcessInternalW_t pCreateProcessInternalW =
        (CreateProcessInternalW_t)GetProcAddress(kernelModule,
                                                 "CreateProcessInternalW");
    if (!pCreateProcessInternalW) {
        Wh_Log(L"No CreateProcessInternalW");
        return;
    }

    STARTUPINFO si{
        .cb = sizeof(STARTUPINFO),
    };
    PROCESS_INFORMATION pi;
    if (!pCreateProcessInternalW(nullptr, currentProcessPath, commandLine,
                                 nullptr, nullptr, FALSE, NORMAL_PRIORITY_CLASS,
                                 nullptr, nullptr, &si, &pi, nullptr)) {
        Wh_Log(L"CreateProcess failed");
        return;
    }

    CloseHandle(pi.hProcess);
    CloseHandle(pi.hThread);
}

    if (g_isToolModProcessLauncher) {
        return;
    }

    WhTool_ModSettingsChanged();
}

void Wh_ModUninit() {
    if (g_isToolModProcessLauncher) {
        return;
    }

    WhTool_ModUninit();
    ExitProcess(0);
}
```

</p>
</details> 

Please do the following:
* Add the code above at the end of your mod code
* Replace `Wh_ModInit`, `Wh_ModSettingsChanged`, `Wh_ModUninit` in your code to `WhTool_*`, e.g. `WhTool_ModInit`
* Change the target process from `explorer.exe` to `windhawk.exe`
* Remove the `FindCurrentProcessTaskbarWnd` check, as it's now guaranteed that only one instance will be running

Please try it and let me know if it behaves as expected in all cases. If it is, let's adapt it. In the future, such functionality might become part of Windhawk.

Alekseev2014s:
I stopped working because it turned out that IsDesktopVisible() does not work correctly on Windows 10. On 11 everything is fine. I was unable to quickly implement support for 10. I will continue working as soon as I have time.

PR #1916, opened 2025-05-18
Auto Theme Switcher

tinodin:
Automatically changes between light and dark mode based on a set schedule.

### DIFF
--- a/mods/apple-music-miniplayer-always-on-top.wh.cpp
+++ b/mods/apple-music-miniplayer-always-on-top.wh.cpp
@@ -1,0 +1,76 @@
+// ==WindhawkMod==
+// @id              apple-music-miniplayer-always-on-top
+// @name            Apple Music Miniplayer Always On Top
+// @description     Keeps the Apple Music MiniPlayer on top
+// @version         1.0
+// @author          tinodin
+// @github          https://github.com/tinodin
+// @include         AppleMusic.exe
+// ==/WindhawkMod==
+
+#include <Windows.h>
+#include <string>
+
+volatile bool g_Running = false;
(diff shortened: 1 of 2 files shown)


m417z:
This mod is injected into explorer and runs as part of it. This is suboptimal for the following reasons:
* There can be more than one explorer.exe process in case the "launch folder windows in separate process" option is enabled, and for other reasons.
* This mod doesn't have a good reason to be injected into explorer.exe, it's just convenient. But by doing so, a mistake or instability in the mod affects the whole shell.

This is not the first mod to take this approach. Here are other recent mods:
* https://github.com/ramensoftware/windhawk-mods/pull/1832
* https://github.com/ramensoftware/windhawk-mods/pull/1913

All three mods could be separate tools, they don't use and don't need Windhawk's injection and hooking capabilities. On the other hand, I can understand that the ease of mod installation is appealing. I wonder if Windhawk needs to come up with a new type of mods which run as part of a dedicated process.

m417z:
Just to be clear, I didn't mean to say that the mod won't be accepted. I was just thinking out loud about the problems with the current approach. Still, I think that such mods can be valuable. Since there's no mechanism in Windhawk right now for such tool-like mods, perhaps some temporary solution can be implemented in the mod itself. I'll give it some thought. Meanwhile, feel free to reopen the PR and we'll keep discussing it.

tinodin:
Yes sorry, I know I just deleted the fork as I accidentally added another mod to the fork. If you want I can create a new pull request as I can not reopen this one.

m417z:
Go ahead and create a new one.

PR #1919, opened 2025-05-19
Auto Theme Switcher

tinodin:
Automatically changes between light and dark mode based on a set schedule.


tinodin:
For your info, I embedded ThemeTool.exe in the mod which is an old executable from Windows 7 (Blog with download link: https://rsn.home.blog/2018/10/26/change-themes-on-the-command-line).

m417z:
Embedding an executable that way isn't acceptable for mods in the repository. One core principle of Windhawk mods is that they're open source. You have the following options:
* Add an option to specify the path of `ThemeTool.exe`. The mod users will have to download it manually and provide its path.
* Reimplement it in the mod, which will result in a better user experience but will require more work.

Also, the executable requires .NET Framework 3.5 which isn't installed by default.

I look at the way that ThemeTool is implemented. It uses the `ITheme`, `IThemeManager` interfaces, similar to the ones that can be found [here](https://github.com/AutoDarkMode/Windows-Auto-Night-Mode/blob/8daffc4006f3d2ed2eb000207adaacc5d165dd35/AutoDarkModeSvc/Handlers/IThemeManager/Interfaces.cs#L43). Here's [another example for AHK](https://github.com/qwerty12/AutoHotkeyScripts/blob/a9423f59c945a3a031cb38b25cf461a34de9a6d3/SetThemeFromdotThemeFile.ahk#L12). Porting it to C++ shouldn't be very difficult.

tinodin:
Ok, I will look into implementing it.

mods/auto-theme-switcher.wh.cpp
```diff
@@ -0,0 +1,377 @@
...
+        using ApplyThemeFunc = HRESULT(WINAPI*)(IUnknown*, BSTR);
+        ApplyThemeFunc ApplyThemeMethod = (ApplyThemeFunc)vtable[4];
+        hr = ApplyThemeMethod(pThemeManager.get(), bstrTheme);
+        if (FAILED(hr)) {
+            Wh_Log(L"[Theme] Failed to apply theme.");
+        }
+
+        Wh_Log(L"[Theme] Successfully applied theme");
+
+        CoUninitialize();
```
m417z: Calling `CoUninitialize` before `pThemeManager` and `bstrTheme` are freed can cause problems, that's why I called it outside of the function in my example.

mods/auto-theme-switcher.wh.cpp
```diff
@@ -0,0 +1,377 @@
...
+    // apply wallpaper to lock screen
+    if (Wh_GetIntSetting(L"LockScreen"))
+        ApplyLockScreen();
+}
+
+// Based on:
+// https://github.com/qwerty12/AutoHotkeyScripts/blob/a9423f59c945a3a031cb38b25cf461a34de9a6d3/SetThemeFromdotThemeFile.ahk
+void ApplyTheme(PCWSTR themePath) {
+    std::thread([=]() {
+        Wh_Log(L"[Theme] Waiting for explorer to load...");
```
m417z, line 205: Is this waiting necessary? In any case, I think it's better to create a thread only if waiting is necessary,

m417z:
I've been thinking about the "mods as tools" situation (https://github.com/ramensoftware/windhawk-mods/pull/1916#issuecomment-2889149966). For the reasons I stated in that comment, I think it's best to have each such mod to run in a separate process. Windhawk doesn't have a built-in functionality for something like that, but for now, I implemented some code to handle that.

Here's your mod with that code added at the bottom, and with the (now unnecessary) mutex removed:
https://gist.github.com/m417z/3b2dafc3fc2f7e88dffe04883f72c8cf

Please try it and let me know if it behaves as expected in all cases. If it is, let's adapt it. In the future, such functionality might become part of Windhawk.

mods/auto-theme-switcher.wh.cpp
tinodin: This waiting is there to prevent it trying to change the theme while on the lockscreen/before the shell is fully initialized because it wouldn't work.

m417z: OK. I think it's better to get rid of that thread, as it can cause various problems. Here are some problems I can think of:
* You're passing `themePath`, but after the thread is done waiting, this string pointer might no longer be valid
* Multiple `ApplyTheme` calls can have multiple threads accumulated and waiting to be called
* The mod can be unloaded while the thread is still running

Instead, you can do the following:
* Only call `ApplyTheme` from the `ThemeScheduler` thread and get rid of the duplicate code in `LoadSettings`
* Call `CoInitialize`/`CoUninitialize` once in the `ThemeScheduler` thread
* Remove the thread logic and `CoInitialize`/`CoUninitialize` in `ApplyTheme`, like in the original code
* Call `ApplyTheme` only if explorer is initialized. If not, you can have a shorter retry delay in `ThemeScheduler`

I think it will result in a more robust implementation. What do you think?

mods/auto-theme-switcher.wh.cpp
```diff
@@ -0,0 +1,377 @@
...
+}
+
+void LoadSettings() {
+    g_lightTime = ParseScheduleTime(Wh_GetStringSetting(L"Light"));
+    g_darkTime = ParseScheduleTime(Wh_GetStringSetting(L"Dark"));
+
+    auto rawLightPath = Wh_GetStringSetting(L"LightThemePath");
+    g_lightThemePath = rawLightPath ? TrimQuotes(rawLightPath) : L"";
+
+    auto rawDarkPath = Wh_GetStringSetting(L"DarkThemePath");
```
m417z, line 351: `Wh_FreeStringSetting` needs to be called to free the strings that are returned by `Wh_GetStringSetting`.

tinodin:
> I've been thinking about the "mods as tools" situation ([#1916 (comment)](https://github.com/ramensoftware/windhawk-mods/pull/1916#issuecomment-2889149966)). For the reasons I stated in that comment, I think it's best to have each such mod to run in a separate process. Windhawk doesn't have a built-in functionality for something like that, but for now, I implemented some code to handle that.
> 
> Here's your mod with that code added at the bottom, and with the (now unnecessary) mutex removed: https://gist.github.com/m417z/3b2dafc3fc2f7e88dffe04883f72c8cf
> 
> Please try it and let me know if it behaves as expected in all cases. If it is, let's adapt it. In the future, such functionality might become part of Windhawk.

The only thing that is a bit annoying is that when the mod is loading you get the mouse loading indicator for 3-5 seconds

mods/auto-theme-switcher.wh.cpp
```diff
@@ -0,0 +1,518 @@
...
+    if (Wh_GetIntSetting(L"LockScreen"))
+        ApplyLockScreen();
+}
+
+void ApplyThemeOrAppearance(bool useLightTheme) {
+    PCWSTR themePath = useLightTheme ? g_lightThemePath.c_str() : g_darkThemePath.c_str();
+
+    if (*themePath) {
+        Wh_Log(L"[Theme] Waiting for explorer to load...");
+        for (;;) {
```
m417z, line 206: How about checking `g_exitFlag` here to avoid being stuck on unload?

mods/auto-theme-switcher.wh.cpp
```diff
@@ -0,0 +1,518 @@
...
+        0xC1B7,
+        0x4045,
+        {0x8F, 0xD0, 0xFF, 0xD6, 0x5D, 0x3F, 0xC7, 0x92}};
+
+    winrt::com_ptr<IUnknown> pThemeManager;
+    HRESULT hr =
+        CoCreateInstance(CLSID_IThemeManager, nullptr, CLSCTX_INPROC_SERVER,
+                        IID_IThemeManager, pThemeManager.put_void());
+    if (FAILED(hr) || !pThemeManager) {
+        Wh_Log(L"[Theme] Failed to apply theme.");
```
m417z, line 185: Not returning here will cause a crash. I'd keep `ApplyTheme` as it originally is and call `ApplyLockScreen` outside of it.

mods/auto-theme-switcher.wh.cpp
```diff
@@ -0,0 +1,518 @@
...
+
+BOOL WhTool_ModInit() {
+    LoadSettings();
+    StartScheduler();
+    return TRUE;
+}
+
+void WhTool_ModSettingsChanged() {
+    LoadSettings();
+    StartScheduler();
```
m417z, line 364: What about the thread that already runs in this case?

tinodin: StartScheduler() checks if the thread exists and just signals it to wake up, so it won’t create multiple threads

m417z:
Now the only thing that's missing is a readme. Please add it with a description of the mod, what it does, and how to use it.

m417z:
Congratulations on your first mod! 🎉

PR #1932, opened 2025-05-23
Auto Theme Switcher v1.0.1

tinodin:
- Fixed changing theme not instantly updating the taskbar background


m417z:
Interesting, maybe Windows only allows signed processes to write there. What if you change the target from `windhawk.exe` to `explorer.exe`, but keep the "tool mod" code? This should still run the mod in a separate process, but in explorer.exe this time.

tinodin:
Yes that is what I did. That is working.

m417z:
In this case, we can submit this change if that's the only viable solution. The "tool mod" code still provides the advantage that the code is fully isolated from other processes, unlike originally where it was injected into a running instance of explorer.

PR #1940, opened 2025-05-25
Auto Theme Switcher v1.0.3

tinodin:
- Add functionality to run custom scripts on switch with -light/-dark argument


mods/auto-theme-switcher.wh.cpp
```diff
@@ -58,12 +63,53 @@ HANDLE g_wakeEvent = nullptr;
...
+    std::wstring ext = g_scriptPath.substr(g_scriptPath.find_last_of(L'.'));
+    std::wstring command;
+
+    if (_wcsicmp(ext.c_str(), L".ps1") == 0) {
+        command = L"powershell.exe -NoProfile -ExecutionPolicy Bypass -File \"" + g_scriptPath + L"\" " + (useLightTheme ? L"-light" : L"-dark");
+    } else if (_wcsicmp(ext.c_str(), L".bat") == 0 || _wcsicmp(ext.c_str(), L".cmd") == 0) {
+        command = L"cmd.exe /c \"" + g_scriptPath + L"\" " + (useLightTheme ? L"-light" : L"-dark");
+    } else {
+        Wh_Log(L"Unsupported script extension: %s", ext.c_str());
+        return;
```
m417z: How about using `ShellExecute` as a fallback? Then you'll be able to support custom programs or scripts (e.g. Python).

tinodin: Should I use CreateProcessW like the other filetypes or ShellExecute?

m417z: `CreateProcessW` will limit it to programs. `ShellExecute` will allow specifying any file type, so I think `ShellExecute` is preferable.

PR #1945, opened 2025-05-26
Windows 95/NT 4.0 Menubars 1.0.0

aubymori:
This mod restores the appearance of menubars from Windows 95/NT 4.0, in whichthey had no hover effect, and changed to the selection color while selected.

This mod will only work with the classic theme.

**Before**:
Before

**After**:
After


aubymori:
Closed while fixing up.

PR #1950, opened 2025-05-26
Add "Disable Windows Ink modifier tooltips"

wily-coyote:
This PR adds a mod that hooks into DWM to disable the following functions:

- `CTrackingTooltip::Initialize`
- `CTrackingTooltip::Update`
- `CTrackingTooltip::UpdateTooltipLocation`

These functions are related to Windows Ink. When holding down a modifier key, DWM renders a cursor-tracking tooltip that displays which modifiers are held.

### DIFF
--- a/mods/no-windows-ink-tooltips.wh.cpp
+++ b/mods/no-windows-ink-tooltips.wh.cpp
@@ -72,6 +72,7 @@ const WindhawkUtils::SYMBOL_HOOK udwmDllHooks[] = {
 BOOL Wh_ModInit() {
 	if(!WindhawkUtils::HookSymbols(GetModuleHandleW(L"udwm.dll"), udwmDllHooks, ARRAYSIZE(udwmDllHooks))){
 		Wh_Log(L"Cannot hook uDWM.dll");
+		return FALSE;
 	}
 	return TRUE;
 }


mods/no-windows-ink-tooltips.wh.cpp
```diff
@@ -0,0 +1,73 @@
+// ==WindhawkMod==
+// @id				no-windows-ink-tooltips
+// @name			Disable Windows Ink modifier tooltips
+// @description		Stops DWM from rendering Windows Ink key modifier tooltips
+// @version			1.0.0
+// @author			wily-coyote
+// @include			dwm.exe
```
m417z, line 7: dwm.exe is excluded by default. How about adding information about it to the readme? Similar to:
https://github.com/ramensoftware/windhawk-mods/blob/d24cd9074413004acff7f87f7e9d6cab89210938/mods/disable-rounded-corners.wh.cpp#L34-L40

mods/no-windows-ink-tooltips.wh.cpp
```diff
@@ -0,0 +1,73 @@
...
+// @github			https://github.com/wily-coyote
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+This mod removes the tooltip that appears next to the cursor when any of the
+modifier keys are pressed while using a Windows Ink tablet by preventing the
+tooltip window from being created and updated.
```
m417z, line 15: How about adding a screenshot to clarify which tooltip windows the mod handles?

mods/no-windows-ink-tooltips.wh.cpp
```diff
@@ -0,0 +1,73 @@
...
+	return 0;
+}
+
+void (*CTrackingTooltip_Update_orig)(void *pThis, const tagPOINT* point, unsigned short* unk);
+void CTrackingTooltip_Update_hook(void *pThis, const tagPOINT* point, unsigned short* unk){
+	return;
+}
+
+void (*CTrackingTooltip_UpdateTooltipLocation_orig)(void *pThis);
+void CTrackingTooltip_UpdateTooltipLocation_hook(void *pThis){
```
m417z: The calling convention is missing here, which is likely to cause a crash for 32-bit targets. Please add calling conventions or, if the mod doesn't support 32-bit, limit it to 64-bit, which will also make compilation faster. See:

mods/no-windows-ink-tooltips.wh.cpp
```diff
@@ -0,0 +1,73 @@
...
+		},
+		(void**)&CTrackingTooltip_UpdateTooltipLocation_orig,
+		(void*)CTrackingTooltip_UpdateTooltipLocation_hook,
+		false
+	}
+};
+
+BOOL Wh_ModInit() {
+	if(!WindhawkUtils::HookSymbols(GetModuleHandleW(L"udwm.dll"), udwmDllHooks, ARRAYSIZE(udwmDllHooks))){
+		Wh_Log(L"Cannot hook uDWM.dll");
```
m417z, line 74: It makes sense to `return FALSE` here.

PR #1954, opened 2025-05-27
Translucent windows v1.5

Undisputed00x:
* Added Rainbow effect
* Additional windows theme parts are modified
* Added text alpha blending
* Changed BlurBehind to AccentBlurBehind
* Added DWMAPI immersive darkmode setting option

### DIFF
--- a/mods/translucent-windows.wh.cpp
+++ b/mods/translucent-windows.wh.cpp
@@ -2,23 +2,23 @@
 // @id              translucent-windows
 // @name            Translucent Windows
 // @description     Enables native translucent effects in Windows 11
-// @version         1.4.2
+// @version         1.5.0
 // @author          Undisputed00x
 // @github          https://github.com/Undisputed00x
 // @include         *
-// @compilerOptions -ldwmapi -luxtheme -lcomctl32
+// @compilerOptions -ldwmapi -luxtheme -lcomctl32 -lgdi32
 // ==/WindhawkMod==
 
 // ==WindhawkModReadme==
 /*
 
-### Acrylic (BlurBehind)
-![BlurBehind](https://i.imgur.com/S8OGRvc.png)
+### Blur (AccentBlurBehind)
+![AccentBlurBehind](https://i.imgur.com/S8OGRvc.png)
 ### Acrylic (SystemBackdrop)
 ![Acrylic SystemBackdrop](https://i.imgur.com/66cAiEC.png)
-### Mica
+### Mica (SystemBackdrop)
 ![Mica](https://i.imgur.com/7BGDhJa.png)
-### MicaAlt
+### MicaAlt (SystemBackdrop)
 ![MicaTabbed](https://i.imgur.com/EeviFqO.png)
 
 # FAQ
@@ -29,13 +29,13 @@
 #
 * The background effects do not affect modern windows (UWP/WinUI)
 
-* It is highly recommended to use the mod with black/dark window themes like the perfect black Rectify11 theme.
+* It is highly recommended to use the mod with black/dark window themes like the Rectify11 "Dark theme with Mica".
 
 * Extending effects to the entire window can result in text being unreadable or even invisible in some cases. 
-Light mode theme, HDR enabled or white background behind the window can cause this. 
-This is because some GDI rendering operations do not preserve alpha channel values.
+Light mode theme, HDR enabled, black textcolor or white background behind the window can cause this. 
+This is because most GDI rendering operations do not preserve alpha values.
 
-* Acrylic blur behind effect may show a bleeding effect at the edges of a window when 
+* Blur effect may show a bleeding effect at the edges of a window when 
 maximized or snapped to the edge of the screen, this is caused by default by the windows DWMAPI.
 
 */
@@ -46,26 +46,44 @@ maximized or snapped to the edge of the screen, this is caused by default by the
 - ThemeBackground: FALSE
   $name: Optimize windows theme
   $description: >-
-     Fill with black color the file explorer background in order to render a clear translucent effect.
-      ⚠ Be aware when changing this setting, the affected windows will need to be reopened to see the change.
+    Fills with black color windows theme parts in order to nullify the alpha of those elements to render a clear translucent effect.
+     ⚠ Be aware when changing this setting, the affected windows will need to be reopened to see the change.
+- TextAlphaBlend: FALSE
+  $name: Text alpha blending
+  $description: >-
+    Alpha blends Windows GDI text rendering. It may not affect all rendered text, use with caution as it may cause problems with some applications (ExplorerBlurMica implementation)
 - type: none
   $name: Effects
   $description: >-
      Windows 11 version >= 22621.xxx (22H2) is required.
-      ⚠ Be aware when changing the effect type from Acrylic BlurBehind or back to Default, the affected windows will need to be reopened to see the change.
   $options:
   - none: Default
-  - acrylicblur: Acrylic (BlurBehind)
+  - acrylicblur: Blur (AccentBlurBehind)
   - acrylicsystem: Acrylic (SystemBackdrop)
-  - mica: Mica
-  - mica_tabbed: MicaAlt
+  - mica: Mica (SystemBackdrop)
+  - mica_tabbed: MicaAlt (SystemBackdrop)
+- AccentBlurBehind: "0F000000"
+  $name: AccentBlurBehind color blend
+  $description: >-
+    Blending color with blur background.
+     Color in hexadecimal ARGB format e.g. 0F000000
+- ImmersiveDarkTitle: TRUE
+  $name: Immersive darkmode titlebar
+  $description: >-
+    Enables the flag for immersive darkmode titlebars, affects the tintcolor of SystemBackdrop effects and caption buttons color state
 - ExtendFrame: FALSE
   $name: Extend effects into entire window
   $description: >-
-    Extends the effects into the entire window background using DwmExtendFrameIntoClientArea. (Required for BlurBehind)
+    Extends the effects into the entire window background using DwmExtendFrameIntoClientArea. (Required for AccentBlurBehind)
+- RainbowSpeed: 1
+  $name: Rainbow Speed
+  $description: Adjust the refresh speed of the rainbow colors (Value between 1 and 100).
 - TitlebarColor:
     - ColorTitlebar: FALSE
       $name: Enable
+    - RainbowTitlebar: FALSE
+      $name: Rainbow Titlebar
+      $description: Enables rainbow effect on windows titlebar.
     - titlerbarstyles_active: "FF0000"
       $name: Focused window color
       $description: >-
@@ -81,6 +99,9 @@ maximized or snapped to the edge of the screen, this is caused by default by the
 - TitlebarTextColor:
     - ColorTitlebarText: FALSE
       $name: Enable
+    - RainbowTextColor: FALSE
+      $name: Rainbow Titlebar text
+      $description: Enables rainbow effect on windows titlebar text.
     - titlerbarcolorstyles_active: "FF0000"
       $name: Focused window color
       $description: >-
@@ -96,10 +117,13 @@ maximized or snapped to the edge of the screen, this is caused by default by the
   $name: Titlebar text color
   $description: >-
      Windows 11 version >= 22000.xxx (21H2) is required.
-      NOTE: This settings affects only Win32 windows. Since Win11 24H2 File Explorer changed titlebar text rendering to DirectWrite API. To modify the text color of File Explorer title bar, use Windhawk's File Explorer Styler mod.
+      NOTE: This settings affects only Win32 windows. Since Win11 22H2 File Explorer changed titlebar text rendering to DirectWrite API. To modify the text color of File Explorer title bar, use Windhawk's File Explorer Styler mod.
 - BorderColor:
     - ColorBorder: FALSE
       $name: Enable
+    - RainbowBorder: FALSE
+      $name: Rainbow Border
+      $description: Enables rainbow effect on windows borders.
     - borderstyles_active: "FF0000"
       $name: Focused window color
       $description: >-
@@ -115,8 +139,8 @@ maximized or snapped to the edge of the screen, this is caused by default by the
         Default = 1
         SystemAccentColor = 2
     - MenuBorderColor: FALSE
-      $name: Extend colored borders to classic context menus and taskbar thumbnails
-      $description: Enable this option to get colored borders on windows classic context menus and taskbar thumbnails.
+      $name: Extend colored borders to classic context menus
+      $description: Enable this option to get colored borders on windows classic context menus.
   $name: Border color
   $description: >-
       Windows 11 version >= 22000.xxx (21H2) is required.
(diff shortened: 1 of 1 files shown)


mods/translucent-windows.wh.cpp
```diff
@@ -1024,25 +1945,39 @@ void Wh_ModAfterInit()
         ApplyForExistingWindows();
 }
 
-void Wh_ModUninit(void) 
+void Wh_ModBeforeUninit()
 {
+    {
+        std::lock_guard<std::mutex> guard(rainbowWindowsMutex);
+        for (auto& [hWnd, pair] : g_rainbowWindows)
+            PostMessageW(hWnd, msgRainbowTimer, RAINBOW_UNLOAD, 0);
```
m417z: `PostMessageW` is asynchronous, which is dangerous in this case - by the time it's processed, the mod might be unloaded and the timer might fire up and cause a crash.

mods/translucent-windows.wh.cpp
```diff
@@ -461,6 +499,20 @@ HWND WINAPI HookedNtUserCreateWindowEx(DWORD dwExStyle,
     return hWnd;
 }
 
+BOOL WINAPI HookedNtUserDestroyWindow(HWND hWnd)
+{
+    {
+        // Unlike WM_DESTROY it captures all hwnd created from NtUserCreatWindowExW
```
m417z: `WM_DESTROY` is reliable from my experience, and I'd rather use it than an undocumented function which isn't guaranteed to be called at all cases (maybe it does, I haven't looked at it). Did you encounter cases in which `WM_DESTROY` isn't fired?

mods/translucent-windows.wh.cpp
```diff
@@ -1024,25 +1945,39 @@ void Wh_ModAfterInit()
...
 
     std::unordered_set<HWND> subclassedWindows;
     {
         std::lock_guard<std::mutex> guard(g_subclassedWindowsMutex);
+
         subclassedWindows = std::move(g_subclassedWindows);
         g_subclassedWindows.clear();
+    
+        for (HWND hWnd : subclassedWindows)
+            WindhawkUtils::RemoveWindowSubclassFromAnyThread(hWnd, SubclassProc);
```
m417z: Why did you move the loop inside the locked region? It was outside to avoid deadlocks. `RemoveWindowSubclassFromAnyThread` calls `SendMessage` which waits for the window to respond. Consider the following execution order:
* `g_subclassedWindowsMutex` is locked in this thread
* The UI thread fires `WM_DESTROY` for a window
  * The `WM_DESTROY` subclass handler tries to lock `g_subclassedWindowsMutex`, and waits for it to become available
* While `g_subclassedWindowsMutex` is still locked here, `RemoveWindowSubclassFromAnyThread` is called
  * `RemoveWindowSubclassFromAnyThread` waits for the window to respond
* Deadlock

mods/translucent-windows.wh.cpp
Undisputed00x: Yes with a simple log on Set/Kill Timers when starting file explorer sets 2 timers in window proc, but WM_DESTROY kills 1 timer

mods/translucent-windows.wh.cpp
```diff
@@ -509,131 +1247,232 @@ void EnableBlurBehind(HWND hWnd)
...
 
-BOOL IsWindowClass(HWND hWnd, LPCWSTR ClassName)
+void CALLBACK MyRainbowTimerProc(HWND, UINT, UINT_PTR t_id, DWORD)
 {
-    WCHAR ClassNameBuffer[256]; 
-    GetClassNameW(hWnd, ClassNameBuffer, sizeof(ClassNameBuffer));
-    if(!wcscmp(ClassName, ClassNameBuffer))
-        return TRUE;
-    return FALSE;
+    for(auto& hWnd : g_rainbowWindows)
```
m417z: Using the `g_rainbowWindowsMutex` lock is needed here, since you're accessing `g_rainbowWindows` from multiple threads, but not below where you don't. To be more precise, reading simultaneously from multiple threads is allowed, but you might also be modifying `g_rainbowWindows` in another thread.

When I suggested using `GetProp`, I hoped that the mutex will be unnecessary, but it's tricky since the only thing we get here is the timer id, not the window handle.

Still, the props help minimize the time that's necessary to hold the lock. So the code can be something like:
```cpp
    RainbowData data = nullptr;
    {
        std::lock_guard<std::mutex> guard(g_rainbowWindowsMutex);
        for(auto& hWnd : g_rainbowWindows)
        {
            HANDLE value = GetPropW(hWnd, g_RainbowPropStr.c_str());
            if (value && (((RainbowData*)value)->WndTimer == t_id))
            {
                data = (RainbowData*)value;
                break;
            }
        }
    }

    if (data)
    {
        data->WndHue = fmod(data->WndHue + g_settings.RainbowSpeed, 360.0f);
        if (g_settings.TitlebarRainbowFlag)
        {
            g_settings.g_TitlebarColor = HSLToRGB(fmod(data->WndHue + 240.0f, 360.0f), 1.0f, 0.5f);
            DwmSetWindowAttribute(hWnd, DWMWA_CAPTION_COLOR, &g_settings.g_TitlebarColor, sizeof(COLORREF));
        }
        if (g_settings.CaptionRainbowFlag)
        {
            if (g_settings.TitlebarRainbowFlag)
                g_settings.g_CaptionColor = HSLToRGB(fmod(data->WndHue + 120.0f, 360.0f), 1.0f, 0.5f);
            else
                g_settings.g_CaptionColor = HSLToRGB(data->WndHue, 1.0f, 0.5f);
            DwmSetWindowAttribute(hWnd, DWMWA_TEXT_COLOR, &g_settings.g_CaptionColor, sizeof(COLORREF));
        }
        if (g_settings.BorderRainbowFlag)
        {
            g_settings.g_BorderColor = HSLToRGB(data->WndHue, 1.0f, 0.5f);  
            DwmSetWindowAttribute(hWnd, DWMWA_BORDER_COLOR, &g_settings.g_BorderColor, sizeof(COLORREF));
        }
    }
```

mods/translucent-windows.wh.cpp
```diff
@@ -509,131 +1247,232 @@ void EnableBlurBehind(HWND hWnd)
...
             break;
+        }   
+        case WM_DESTROY:
+        {       
+            HANDLE value = RemovePropW(cwp->hwnd, g_RainbowPropStr.c_str());
+            if (value)
+            {
+                RainbowData* data = (RainbowData*)value;
+                KillTimer(NULL, data->WndTimer);
+                g_rainbowWindows.erase(cwp->hwnd);
```
m417z: Mutex?
```suggestion
                {
                    std::lock_guard<std::mutex> guard(g_rainbowWindowsMutex);
                    g_rainbowWindows.erase(cwp->hwnd);
                }
```

mods/translucent-windows.wh.cpp
```diff
@@ -509,131 +1247,232 @@ void EnableBlurBehind(HWND hWnd)
...
+            if (cwp->message == g_msgRainbowTimer)
+            {
+                switch (cwp->wParam)
+                {
+                    case RAINBOW_LOAD:
+                    {
+                        HANDLE value = GetPropW(cwp->hwnd, g_RainbowPropStr.c_str());
+                        if (!value)
+                        {
+                            g_rainbowWindows.insert(cwp->hwnd);
```
m417z: Mutex? Also makes sense to insert it only if `SetTimer` succeeds.
```suggestion
                            {
                                std::lock_guard<std::mutex> guard(g_rainbowWindowsMutex);
                                g_rainbowWindows.insert(cwp->hwnd);
                            }
```

mods/translucent-windows.wh.cpp
```diff
@@ -206,51 +259,85 @@ maximized or snapped to the edge of the screen, this is caused by default by the
...
+std::mutex g_rainbowWindowsMutex;
+std::unordered_set<HWND> g_rainbowWindows;
+
+struct RainbowData {
+float WndHue;
+UINT_PTR WndTimer;
+HWND WndHwnd;
+};
+
+std::mutex g_drawTextMutex;
```
m417z: Unused.
```suggestion
```

mods/translucent-windows.wh.cpp
```diff
@@ -665,61 +1507,53 @@ void NewWindowShown(HWND hWnd)
...
-            std::lock_guard<std::mutex> guard(g_subclassedWindowsMutex);
-            g_subclassedWindows.insert(hWnd);
+    {
+        std::lock_guard<std::mutex> guard(g_allCallWndProcHooksMutex);   
+        DWORD dwThreadId = GetWindowThreadProcessId(hWnd, NULL);
+        HHOOK callWndProcHook = SetWindowsHookEx(WH_CALLWNDPROC, CallWndProc, nullptr, dwThreadId);
+        if (callWndProcHook) {
+            g_callWndProcHook = callWndProcHook;
+            g_allCallWndProcHooks.insert(callWndProcHook);
         }
```
m417z: Using a blocking operation such as `SendMessage` while holding a lock is dangerous.

```suggestion
        {
            std::lock_guard<std::mutex> guard(g_allCallWndProcHooksMutex);   
            DWORD dwThreadId = GetWindowThreadProcessId(hWnd, NULL);
            HHOOK callWndProcHook = SetWindowsHookEx(WH_CALLWNDPROC, CallWndProc, nullptr, dwThreadId);
            if (callWndProcHook) {
                g_callWndProcHook = callWndProcHook;
                g_allCallWndProcHooks.insert(callWndProcHook);
            }
        }
```

mods/translucent-windows.wh.cpp
```diff
@@ -509,131 +1247,232 @@ void EnableBlurBehind(HWND hWnd)
...
+        }
+    }
 }
 
-LRESULT CALLBACK SubclassProc(HWND hWnd, UINT uMsg, WPARAM wParam, LPARAM lParam, DWORD_PTR dwRefData)
+LRESULT CALLBACK CallWndProc(int nCode, WPARAM wParam, LPARAM lParam)
 {
-    switch (uMsg)
+    const CWPSTRUCT* cwp = (const CWPSTRUCT*)lParam;
+    if (nCode != HC_ACTION || !IsWindowEligible(cwp->hwnd)) {
```
m417z: I'd rather not call `IsWindowEligible` here as this callback is called a lot, most of the time with messages you don't care about. Also, a window might become non-eligible, but you'd still want to clean it up in `RAINBOW_UNLOAD` or `WM_DESTROY`. Perhaps the `IsWindowEligible` check only makes sense for `WM_ACTIVATE`?

Alternatively, if you meant to check whether it's a window you handled when it was created, you can attach a property to it with `SetProp` in `NewWindowShown`, and check it here.

mods/translucent-windows.wh.cpp
Undisputed00x: Locks are also necessary here
```cpp
 g_settings.g_TitlebarColor = HSLToRGB(fmod(data->WndHue + 240.0f, 360.0f), 1.0f, 0.5f);
 DwmSetWindowAttribute(hWnd, DWMWA_CAPTION_COLOR, &g_settings.g_TitlebarColor, sizeof(COLORREF));
```
As g_TitlebarColor and the other global color values are used in HookedDwmSetWindowAttribute to override the changes of the colors made internally

Maybe instead of locking these parts too, changes those global color variables to thread_local is an option , or maybe also setting these colors to window properties in the NewWindowShown func.

m417z: I see. That's less of a problem, I think they can be used without a lock.

First, `DwmSetWindowAttribute` for that window is likely to be run from the window's thread as well, in which case there's no concurrent access.

But even if another thread changes it, all these values (of types `COLORREF`, `BOOL`, etc.) are handled by most CPUs (x32, x64, ARM64) to be read and written atomically, so no locks are needed. You could use `std::atomic` to be sure about that, but `DwmSetWindowAttribute` receives a regular pointer, so it'll just make the code less readable with more casts, I don't think it's worth it.

And even if somehow the values here become incorrect, which shouldn't be the case given the above, the worst thing that'll happen is setting an incorrect color. That's not as severe as accessing an incorrect std::map which will likely cause a crash.

Undisputed00x: There is a race condition as these global values are used inside the HookedDwmSetWindowAttribute. The color inconsistency is visible if rainbow setting is enabled.

m417z: I don't think that it's caused by a race condition due to the reasons I listed, but perhaps I'm missing something. What's the easiest way to reproduce it? Also probably unrelated, but note that when you're calling `DwmSetWindowAttribute`, you're calling your hook. You might want to call `DwmSetWindowAttribute_orig` directly.

m417z: Perhaps the issue is caused by the following: `WndHue` is a per-window variable, but `TitlebarRainbowFlag`, `CaptionRainbowFlag`, `BorderRainbowFlag` are global, even though they're changed according to `WndHue`. It doesn't look right.

These three values are initialized from settings, which is fine, but then each window does it's own modifications to them. What seems to make sense here is to have these values for each window, which start with values from the global settings, and then change independently for each window.

In my mods, I try to manage a settings struct in such a way that the data is never changed unless settings are changed. I think it makes the code and the variables easier to reason about.

Undisputed00x: Calling the original function DwmSetWindowAttribute_orig doesn't create any issue it seems

m417z: I think that it reduces the chance for it because it runs faster (`IsWindowEligible` isn't called all the time), which is a good thing, but I still think it's better to not have the global variables stomped by different threads. Especially since you already have the fundamentals for implementing it properly.

m417z:
Looks good except for the shared state that I suggested to change, do you want to merge it as is?

Undisputed00x:
> Looks good except for the shared state that I suggested to change, do you want to merge it as is?

The shared color values are needed for HookedDwmSetWindowAttribute and used when windows are created or changed active state. In the rainbow function actually i can just declare local variable and pass them to the DwmSetWindowAttribute_orig as the continuous call of DwmSetWindowAttribute overrides the initial values of the window

```cpp
data->WndHue = fmod(data->WndHue + g_settings.RainbowSpeed, 360.0f);
        if (g_settings.TitlebarRainbowFlag)
        {
            COLORREF TitlebarClr = HSLToRGB(data->WndHue, 1.0f, 0.5f); 
            DwmSetWindowAttribute_orig(data->WndHwnd, DWMWA_CAPTION_COLOR, &TitlebarClr, sizeof(COLORREF));
        }
        if (g_settings.CaptionRainbowFlag)
        {
            COLORREF CaptionTextClr;
            if (g_settings.TitlebarRainbowFlag)
                CaptionTextClr = HSLToRGB(fmod(data->WndHue + 120.0f, 360.0f), 1.0f, 0.5f);
            else
                CaptionTextClr = HSLToRGB(data->WndHue, 1.0f, 0.5f);
            DwmSetWindowAttribute_orig(data->WndHwnd, DWMWA_TEXT_COLOR, &CaptionTextClr, sizeof(COLORREF));
        }
        if (g_settings.BorderRainbowFlag)
        {
            COLORREF BorderClr = HSLToRGB(data->WndHue, 1.0f, 0.5f);  
            DwmSetWindowAttribute_orig(data->WndHwnd, DWMWA_BORDER_COLOR, &BorderClr, sizeof(COLORREF));
        }
```

Besides that putting those global colors as properties in every window i think will create the unnecessary small overhead of inserting and accessing them.

m417z:
Which overhead are you referring to? In the timer, you're getting the props pointer anyway. There will be additional overhead in storing these per-window, and accessing them in other places such as on activation, but these events are not as common and I think it's insignificant. I think that correctness is preferable in this case because just like you've seen weird results before switching to `DwmSetWindowAttribute_orig`, perhaps it can also happen now on slower computers or when the CPU is busy.

PR #1969, opened 2025-06-01
Winver Custom Text v1.0.0

ilikecoding-197:
my first mod :)

### DIFF
--- a/mods/winver-custom-text.wh.cpp
+++ b/mods/winver-custom-text.wh.cpp
@@ -1,0 +1,113 @@
+// ==WindhawkMod==
+// @id              winver-custom-text
+// @name            Winver Custom Text
+// @description     Allows you to set the text displayed in winver.exe.
+// @version         1.0
+// @author          Thomas Shrader
+// @github          https://github.com/ilikecoding-197
+// @include         winver.exe
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+Allows you to set the text displayed in winver.exe.
+*/
+// ==/WindhawkModReadme==
+
+// ==WindhawkModSettings==
+/*
+- title: About Windows
+  $name: Title
+  $description: What displays in the title.
+- firstline: Microsoft Windows
+  $name: First line
+  $description: What displays on the first line, after the Windows logo.
+- otherstuff: ""
+  $name: Other stuff
+  $description: Info appearing after copy right and stuff.
+*/
+// ==/WindhawkModSettings==
+
+// The source code of the mod starts here. This sample was inspired by the great
+// article of Kyle Halladay, X64 Function Hooking by Example:
+// http://kylehalladay.com/blog/2020/11/13/Hooking-By-Example.html
+// If you're new to terms such as code injection and function hooking, the
+// article is great to get started.
+
+#include <shellapi.h>
+#include <stdio.h>
+
+struct {
+    PCWSTR title;
+    PCWSTR firstLine;
+    PCWSTR otherstuff;
+} Settings;
+
+struct {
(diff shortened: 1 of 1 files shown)


mods/winver-custom-text.wh.cpp
```diff
@@ -0,0 +1,113 @@
...
+    return ShellAboutW_Orginal(hWnd, ShellAboutArgs.szApp, Settings.otherstuff, hIcon);
+}
+
+void LoadSettings() {
+    Wh_Log(L"LoadSettings();");
+    Settings.title = Wh_GetStringSetting(L"title");
+    Settings.firstLine = Wh_GetStringSetting(L"firstline");
+    Settings.otherstuff = Wh_GetStringSetting(L"otherstuff");
+
+    ShellAboutArgs.szApp = new wchar_t[1024];
```
m417z, line 72: This memory is never freed. You can free it in `FreeSettings`, or, more conveniently, you can just use `std::wstring` which is freed automatically.

For mod settings, you can use `WindhawkUtils::StringSetting`, in which case you won't need `FreeSettings` at all.

mods/winver-custom-text.wh.cpp
```diff
@@ -0,0 +1,113 @@
...
+// @description     Allows you to set the text displayed in winver.exe.
+// @version         1.0
+// @author          Thomas Shrader
+// @github          https://github.com/ilikecoding-197
+// @include         winver.exe
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+Allows you to set the text displayed in winver.exe.
```
m417z, line 13: How about adding a screenshot for those unfamiliar with `winver.exe`?

m417z (commented):
Well done, looks great! I left minor comments.

mods/winver-custom-text.wh.cpp
codingisfun2831t: Updated to be freed!

mods/winver-custom-text.wh.cpp
codingisfun2831t: Done.

ilikecoding-197:
Fixed up!

m417z:
Congratulations on your first mod! :)

mods/winver-custom-text.wh.cpp
```diff
@@ -0,0 +1,119 @@
...
+) {
+    Wh_Log(L"ShellAboutW(NULL, L\"%s\", L\"%s\", NULL);", szApp, szOtherStuff);
+    return ShellAboutW_Orginal(hWnd, ShellAboutArgs.szApp, Settings.otherstuff, hIcon);
+}
+
+void LoadSettings() {
+    Wh_Log(L"LoadSettings();");
+    Settings.title = Wh_GetStringSetting(L"title");
+    Settings.firstLine = Wh_GetStringSetting(L"firstline");
+    Settings.otherstuff = Wh_GetStringSetting(L"otherstuff");
```
m417z, line 70: Just noticed that `Settings.otherstuff` isn't freed as well. You can fix it in the next update. Or as I said in the other comment, just use objects which take care of it for you.

mods/winver-custom-text.wh.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+    ShellAboutArgs.szApp = new wchar_t[1024];
```
aubymori, line 72: Constant length; this can be allocated on the stack rather than the heap. Change the definition in the `ShellAboutArgs` struct to `WCHAR szApp[1024];`, and remove this line as well as the `delete ShellAboutArgs.szApp;` line in `FreeSettings`.

PR #1975, opened 2025-06-02
Windows 10 Colorization Tweaks 1.1.0

aubymori:
- Add the ability to force windows to use either light or dark theme always
- Rename the mod from "Windows 10 Colored Borders" to "Windows 10 Colorization Tweaks"

### DIFF
--- a/mods/win10-colored-borders.wh.cpp
+++ b/mods/win10-colored-borders.wh.cpp
@@ -62,29 +62,29 @@ enum WINDOWCOLOR
     WC_DARK
 } g_eWindowColor = WC_NORMAL;
 
-void (*CGlassColorizationParameters_AdjustWindowColorization_orig)(void *, void *, float, BYTE);
+void (*CGlassColorizationParameters_AdjustWindowColorization_orig)(void *, void *, float, DWORD);
 void CGlassColorizationParameters_AdjustWindowColorization_hook(
     void *pThis,
     void *pgpcc,
     float flUnknown,
-    BYTE  bFlags
+    DWORD dwFlags
 )
 {
-    if (g_fColorBorders && !(bFlags & 0x8))
-        bFlags &= ~0x4;
+    if (g_fColorBorders && !(dwFlags & 0x8))
+        dwFlags &= ~0x4;
     switch (g_eWindowColor)
     {
         case WC_NORMAL:
             break;
         case WC_LIGHT:
-            bFlags &= ~0x80;
+            dwFlags &= ~0x80;
             break;
         case WC_DARK:
-            bFlags |= 0x80;
+            dwFlags |= 0x80;
             break;
     }
     CGlassColorizationParameters_AdjustWindowColorization_orig(
-        pThis, pgpcc, flUnknown, bFlags
+        pThis, pgpcc, flUnknown, dwFlags
     );
 }
 


mods/win10-colored-borders.wh.cpp
```diff
@@ -11,39 +11,72 @@
...
-# Windows 10 Colored Borders
-Starting from Windows 10 version 1809, the borders of active windows aren't colored
-unless the user enables the option to color titlebars. This mod reverts that behavior
-and makes active borders always use the user's accent color regardless.
+# Windows 10 Colorization Tweaks
+This mod allows you to tweak the colorization of windows in Windows 10, including
+always showing colored borders on active windows and always using dark or light mode
+windows.
 
-## ⚠ Important usage note ⚠
```
m417z, line 19: Why was it removed?

mods/win10-colored-borders.wh.cpp
```diff
@@ -11,39 +11,72 @@
...
+    WC_DARK
+} g_eWindowColor = WC_NORMAL;
+
+void (*CGlassColorizationParameters_AdjustWindowColorization_orig)(void *, void *, float, BYTE);
 void CGlassColorizationParameters_AdjustWindowColorization_hook(
     void *pThis,
     void *pgpcc,
     float flUnknown,
-    DWORD dwFlags
+    BYTE  bFlags
```
m417z: Is this change correct? Enums are usually `int` (`DWORD` in x86/64 as it was previously). That's also what I see in my Win11 in assembly, didn't check in Win10.

mods/win10-colored-borders.wh.cpp
```diff
@@ -11,39 +11,72 @@
...
+![Dark windows](https://raw.githubusercontent.com/aubymori/images/refs/heads/main/win10-colorization-tweaks-dark-window.png)
 */
 // ==/WindhawkModReadme==
 
+// ==WindhawkModSettings==
+/*
+- colorborders: true
+  $name: Always color borders
+  $description: Always color active borders, regardless if titlebars are colored or not.
+- windowcolor: normal
```
m417z, line 44: The value here, `normal`, is different than the option below, `Normal`. I don't think that the options here are case-insensitive. You don't use that string in code, but it'll be shown incorrectly for users after installing the mod.

mods/win10-colored-borders.wh.cpp
aubymori: Fixed.

mods/win10-colored-borders.wh.cpp
aubymori: IDA shows `char` on my end:
image

m417z: Can you show the first usage in both decompiled view and assembly?

mods/win10-colored-borders.wh.cpp
aubymori: I think you're confused; in `$options`, the key describes the internal value, and the value describes the UI name. The current definitions are correct.

m417z: Right, my bad.

mods/win10-colored-borders.wh.cpp
m417z: Assembly view:

image

aubymori: image
image

It does seem to use a DWORD, but it never compares it to any value above 0xFF, and the mod doesn't seeem to have any issue with treating it as a BYTE.

m417z: Currently, no flags beyond the low 8 bits are used, so you're right that it'll work either way. Having the mod treat is as `BYTE` means that the other 24 bits are unspecified, and so if more flags are added in the future, things might break.

On the other hand, using `DWORD` will pass along all 32 bits, and so even if the internal type is meant to be `BYTE`, it will work correctly.

Up to you, let me know if you want to have it merged as is.

aubymori: Eh, I went ahead and reverted it. I doubt they have an enum that's a single byte internally anyway.

PR #1978, opened 2025-06-03
Explorer Double Click Up 1.0.0

wrldspawn:


m417z:
Nice! I did a quick test on Windows 11, and it seems to work there too.

How about handling windows which are already open when the mod is enabled? Can be perhaps implemented by enumerating the open windows in `Wh_ModAfterInit`.

PR #1996, opened 2025-06-05
Restore Button Pulse Animation 1.0.2

aubymori:
- Update `LoadComCtlModule` to reject comctl32.dll version 5.82

### DIFF
--- a/mods/restore-button-pulse-anim.wh.cpp
+++ b/mods/restore-button-pulse-anim.wh.cpp
@@ -2,11 +2,11 @@
 // @id              restore-button-pulse-anim
 // @name            Restore Button Pulse Animation
 // @description     Restores the pulse animation that focused buttons played in Windows Vista/7
-// @version         1.0.1
+// @version         1.0.2
 // @author          aubymori
 // @github          https://github.com/aubymori
 // @include         *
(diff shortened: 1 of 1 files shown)


mods/restore-button-pulse-anim.wh.cpp
```diff
@@ -293,6 +326,16 @@ HMODULE LoadComCtlModule(void)
     ULONG_PTR ulCookie;
     ActivateActCtx(hActCtx, &ulCookie);
     HMODULE hComCtl = LoadLibraryW(L"comctl32.dll");
+    /**
+      * Certain processes will ignore the activation context and load
+      * comctl32.dll 5.82 anyway. If that occurs, just reject it.
+      */
+    VS_FIXEDFILEINFO *pVerInfo = GetModuleVersionInfo(hComCtl, nullptr);
+    if (HIWORD(pVerInfo->dwFileVersionMS) < 6)
```
m417z: I didn't see comctl32.dll without version info, but IMO it wouldn't hurt to be defensive here just in case.

```suggestion
    if (pVerInfo && HIWORD(pVerInfo->dwFileVersionMS) < 6)
```

PR #1997, opened 2025-06-06
Windows 11 Accent Window Border 1.0.0

Guerra24:
When you enable *Show accent color on title bars and window borders* in build 22000 the accent color appeared behind the Mica effect but in 22621 and newer it appears on top and it doesn't look that great, this mod restores the former behavior.

Before:
image

After:
image

I know enough C++ to pull this off but not enough to go the patch-dwm-route which should be better, but I think it's fine as-is. Any suggestions codewise are welcomed.


m417z:
I think it'd be a good idea to include your detailed description and screenshots from here in the mod readme.

Issue #1936, opened 2025-05-24
[Restore Button Pulse Animation] LoadComCtlModule returns comctl32.dll version 5


m417z:
## Background

_If short on time, you can skip to **The problem** below._

@aubymori, I was looking at the numerous reports of the Windhawk progress dialog being displayed too often, for example:
* https://github.com/ramensoftware/windhawk-mods/issues/1618
* https://github.com/ramensoftware/windhawk/issues/427
* https://github.com/ramensoftware/windhawk/issues/153
* https://github.com/ramensoftware/windhawk/issues/102
* https://github.com/ramensoftware/windhawk/discussions/421

It's challenging to troubleshoot given that it's not easy to reproduce and that users aren't very cooperative, but so far, I figured the following:
* Most reports are related to mods which inject into all processes.
* Most/many reports mention mods which depend on comctl32.dll symbols.
  * Aerexplorer: [1](https://github.com/ramensoftware/windhawk-mods/issues/1618), [2](https://github.com/ramensoftware/windhawk/issues/102#issuecomment-2169134999)
  * [Classic Task Dialog Fix](https://github.com/ramensoftware/windhawk/issues/427#issuecomment-2765007004)
  * [Classic File Picker dialog](https://github.com/ramensoftware/windhawk/issues/153) (that one, by @Anixx,  just uses `LoadLibrary(L"comdlg32.dll")`)
  * [Button Pulse Animation](https://github.com/ramensoftware/windhawk/discussions/421#discussioncomment-13190051)

Currently, when `WindhawkUtils::HookSymbols` is called, the mod tries to download symbols from the online cache server, then from Microsoft servers. Upon success, the offsets are cached and the next `WindhawkUtils::HookSymbols` calls are very fast. That's the happy flow, no problem here.

If `WindhawkUtils::HookSymbols` doesn't succeed - no internet, no symbols were published, or the target binary doesn't have the requested symbols - then the process is retried on every call of `WindhawkUtils::HookSymbols`, i.e. for every new process. That's not good as is means that every process that's launched looks for symbols in the online cache server and then in Microsoft's servers. It causes a delay, and may cause the progress dialog to pop up if the initialization takes more than 3 seconds (by default).

Windhawk can certainly be improved in this area, especially for cases in which mods don't have much to do on their own, e.g. if there's no internet. In the next Windhawk update, I plan on remembering such failures and retrying at most once every 4 hours.

[...]

PR #2043, opened 2025-06-23
Aerexplorer 1.7.0

aubymori:
- Introduce option for Windows 8 beta navigation bar background
  - The logic to draw the background this way is not present in modern Windows. Previously, a hack in themes allowed this navbar background, but also interfered with Internet Explorer.
- Fix a bug where glass would apply on ribbon visibility change with the option disabled
- Change the names of most hook function arguments and class members (macros) to match the names from private debug symbols
- Completely rewrite init function to use macros, it is now less repetitive
- Modify some of the option descriptions to be clearer

### DIFF
--- a/mods/aerexplorer.wh.cpp
+++ b/mods/aerexplorer.wh.cpp
@@ -2,7 +2,7 @@
 // @id              aerexplorer
 // @name            Aerexplorer
 // @description     Various tweaks for Windows Explorer to make it more like older versions.
-// @version         1.6.5
+// @version         1.7.0
 // @author          aubymori
 // @github          https://github.com/aubymori
 // @include         *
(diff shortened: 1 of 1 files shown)


mods/aerexplorer.wh.cpp
```diff
@@ -2529,511 +2547,463 @@ HMODULE LoadComCtlModule(void)
...
-        DwmExtendFrameIntoClientArea_hook,
-        &DwmExtendFrameIntoClientArea_orig
-    ))
+    LOAD_MODULE(shell32)
+    HOOK_SYMBOLS(shell32, shell32DllHooks)
+
+    LOAD_MODULE_(windows.storage, windows_storage)
+    HOOK_SYMBOLS_(windows.storage, windows_storage, windowsStorageHooks)
+
+    HMODULE comctl32 = LoadComCtlModule();
```
m417z, line 3030: What about https://github.com/ramensoftware/windhawk-mods/issues/1936#issuecomment-2947070043?

aubymori: Fixed.

PR #2073, opened 2025-06-29
Delete mods/chromium-native-titlebar.wh.cpp

Anixx:
This mod is superseded by "Titlebar For Everyone" mod. It is recommended to use that mod.


m417z:
Mods aren't normally being removed from the repository. In this case, it's better to push an update which clarifies that the mod is superseded by another mod in the description.

Anixx:
This mod is absolutely useless now and in the description it was said that it is meant as a temporary solution. Should I submit something that breaks build so that it not to appear in the repo?

m417z:
There's no need to break it, just point out that it's useless in the description. You can have it as a BIG AND SCARY WARNING, for example:
https://github.com/ramensoftware/windhawk-mods/blob/3fda39eb73bcfe77a571b34ce29334100cfdf974/mods/classic-conhost.wh.cpp#L19-L25

Some existing users might prefer to keep using it for their specific reasons, there's no reason to break it for them.

Anixx:
If someone uses it, deletion will not break it for them. It only removes from the repo, not from their systems. As to the reasons, there are none, actually.

PR #2081, opened 2025-06-30
Added New Mod Volume Hid Sender

rom4ster:
This mod allows you to send HID input for the MOUNTAIN EVEREST MAX keyboard everytime the volume is changed. This solution is not perfect. The volume will frequently be off by one. It is also possible to desync the volume temporarily (until next volume update) by taking a very long time to decide which volume you want.

Custom timeouts may be supported in the future

Arbritary HID output will be supported in the future

Requires HID HIDApiTester.

Special thanks to Borked on discord for coming up with the idea and helping me when I encountered issues.

### DIFF
--- a/mods/volume-hid-sender.wh.cpp
+++ b/mods/volume-hid-sender.wh.cpp
@@ -264,11 +264,11 @@ void Set_Hooks() {
     if (user32Module) {
         // SendMessageW_t * pSendMessageW =
         //     GetProcAddress(user32Module, "SendMessageW");
-        if (true) {
+
             WindhawkUtils::SetFunctionHook(SendMessageW,
                                SendMessage_Hook,
                                &SendMessageW_Original);
-        }
+        
     }
 
 }


mods/volume-hid-sender.wh.cpp
```diff
@@ -0,0 +1,323 @@
+// ==WindhawkMod==
+// @id              volume-hid-sender
+// @name            Volume HID Sendeer
+// @description     HID Mod For Volume (Mountain Everest Max Only)
+// @version         0.1
+// @author          rom4ster
+// @github          https://github.com/rom4ster
+// @include         explorer.exe
+// @include         ShellHost.exe
+// @compilerOptions -lcomctl32 -lole32 -loleaut32 -lurlmon -lruntimeobject -Wl,--export-all-symbols -v
```
m417z: Please remove `-v`. And do you really need `-Wl,--export-all-symbols`?

mods/volume-hid-sender.wh.cpp
```diff
@@ -0,0 +1,323 @@
...
+    return (LPWSTR ) completestr->c_str();
+
+}
+
+
+DWORD WINAPI ThreadFunc(void * data) {
+        Sleep(2 SECONDS);
+        float vol = GetSystemVolume(VolumeUnit::Scalar);
+        Wh_Log(L"Volume Info: %f", vol);
+        startup(L"I:\\hidapitest\\hidapitester.exe", FormCommand(vol));
```
m417z: I don't think this is going to work for users who don't have this file there.

mods/volume-hid-sender.wh.cpp
```diff
@@ -0,0 +1,323 @@
...
+    std::wstringstream ss;
+    ss << voll;
+    std::wstring volstring = ss.str();
+
+    std::wstring commandp1(L"64,0x11,0x83,00,00,");
+    std::wstring commandp2(L",00,00,00,00,00,00,00,00,00,00,00,00,00,00,00,00,00,00,00,00,00,00,00,00,00,00,00,00,00,00,00,00,00,00,00,00,00,00,00,00,00,00,00,00,00,00,00,00,00,00,00,00,00,00,00,00,00,00,00"); //Setting
+    std::wstring command = commandp1 + volstring + commandp2;
+    std::wstring fullstr(HID_PATH);
+
+    std::wstring * completestr = new std::wstring(fullstr SPACE L"--vidpid" SPACE vid + L":" + pid SPACE L"--usagePage" SPACE usage_page SPACE L"--usage" SPACE usage SPACE L"--open" SPACE L"--send-output" SPACE command);
```
m417z: This won't work correctly if the path contains spaces. It also leaks memory because you never free it. You can just return `std::wstring` instead of using `new` like this.

mods/volume-hid-sender.wh.cpp
```diff
@@ -0,0 +1,323 @@
...
+BOOL FileExists(LPCTSTR szPath)
+{
+  DWORD dwAttrib = GetFileAttributes(szPath);
+
+  return (dwAttrib != INVALID_FILE_ATTRIBUTES && 
+         !(dwAttrib & FILE_ATTRIBUTE_DIRECTORY));
+}
+
+void Init_Settings() {
+    HID_PATH = Wh_GetStringSetting(HIDTesterSetting);
```
m417z, line 302: You never free it, and you don't reload it when settings are changed.

mods/volume-hid-sender.wh.cpp
rom4ster: I just copy pasted them from another mod, i can probably do without those,

rom4ster: re3moved

mods/volume-hid-sender.wh.cpp
rom4ster: Taken from settings now.

mods/volume-hid-sender.wh.cpp
rom4ster: If the path has spaces then quotes can be used, I now provide this automatically to save user effort. I now free the things.

mods/volume-hid-sender.wh.cpp
rom4ster: Added reload implementation. 
It is FREEE :)

mods/volume-hid-sender.wh.cpp
rom4ster: I now no longer free the things, I just use the almighty stack.

mods/volume-hid-sender.wh.cpp
m417z: Please remove `-v`.

mods/volume-hid-sender.wh.cpp
```diff
@@ -0,0 +1,353 @@
...
+    if (!FileExists(HID_PATH)) {
+        Download_EXE();
+        return FALSE;
+    }
+    return TRUE;
+}
+
+
+void Wh_ModSettingsChanged() {
+    Init_Settings();
```
m417z, line 317: Here, the old settings are not freed.

mods/volume-hid-sender.wh.cpp
rom4ster: removed, that got back in through a regression.

mods/volume-hid-sender.wh.cpp
rom4ster: Now I un init  settings before init new ones (includes guard for nullptr)

rom4ster: Resolved crash for function order. I did not save a file before copy pasting whoops

mods/volume-hid-sender.wh.cpp
```diff
@@ -0,0 +1,357 @@
...
+        Sleep(2 SECONDS);
+        //Sleep(15 SECONDS);
+        //vol = GetSystemVolume(VolumeUnit::Scalar);
+        //Wh_Log(L"Volume Info: %f", vol);
+        //startup(L"hidapitester.exe", FormCommand(vol));
+        return 0;
+}
+
+
+LRESULT SendMessage_Hook(HWND hWnd, UINT Msg, WPARAM wParam, LPARAM lParam) {
```
m417z: I only noticed this now:

The calling convention is missing here, which is likely to cause a crash for 32-bit targets. Please add calling conventions or, if the mod doesn't support 32-bit, limit it to 64-bit, which will also make compilation faster. See:

mods/volume-hid-sender.wh.cpp
```diff
@@ -0,0 +1,357 @@
...
+        endpointVolume->Release();
+        CoUninitialize();
+
+        return currentVolume;
+    }
+
+
+
+
+using SendMessageW_t = LRESULT (HWND hWnd, UINT Msg, WPARAM wParam, LPARAM lParam);
```
m417z: `WINAPI` is missing here too. Or, even better, just use:
```cpp
using SendMessageW_t = decltype(&SendMessageW);

File: mods/volume-hid-sender.wh.cpp (hunk 6 above)
Comment (rom4ster) [reply mods/volume-hid-sender.wh.cpp]: So I added WINAPI to the hook function, and use winhawkutils to hook it like the tutorial was saying.

File: mods/volume-hid-sender.wh.cpp [hunk 8]
```diff
@@ -0,0 +1,357 @@
...
+
+void Set_Hooks() {
+        HMODULE user32Module = LoadLibrary(L"user32.dll");
+    if (user32Module) {
+        void* pSendMessageW =
+            (void*)GetProcAddress(user32Module, "SendMessageW");
+        if (pSendMessageW) {
+            WindhawkUtils::SetFunctionHook(pSendMessageW,
+                               (void*) SendMessage_Hook,
+                               (void**)&SendMessageW_Original);
```
m417z: The idea of using `WindhawkUtils::SetFunctionHook` is that you can remove the casts (`(void*)` and `(void**)`). If you remove them and it doesn't compile, that means that there's a problem, which is what happens with your code.

mods/volume-hid-sender.wh.cpp
rom4ster: Oh thats far better thanks.

mods/volume-hid-sender.wh.cpp
rom4ster: Ok, turns out i had a pointer on the original instead of just the original. Removing if true soon.

rom4ster: I think its ok now. Lots of new learnings for me on this one sorry :(

Issue #2063, opened 2025-06-28
[Multiple] Insecure LoadLibrary causing potential code execution


namazso:
The following mods inject into everything and use `LoadLibrary` to load a module that is not a KnownDll, resulting in malicious hijacking if the executable is not in a trusted folder.

The most trivial example is that installers in your Downloads folder will start loading random dlls from your Downloads directory, however it's a clear security issue if you consider running executables where a local attacker (such as another local user) can write files in the same folder.

The following mods appear to be vulnerable:

- [ ] https://github.com/ramensoftware/windhawk-mods/blob/main/mods/aerexplorer.wh.cpp#L3018  Aerexplorer by @aubymori 
- [ ] https://github.com/ramensoftware/windhawk-mods/blob/main/mods/classic-maximized-windows-fix.wh.cpp#L415 Fix Classic Theme Maximized Windows by @ephemeralViolette
- [ ] https://github.com/ramensoftware/windhawk-mods/blob/main/mods/classic-menus.wh.cpp#L56 Classic Menus by @aubymori 
- [ ] https://github.com/ramensoftware/windhawk-mods/blob/main/mods/classic-start-menu-folders-icon.wh.cpp#L200 Classic Start Menu Folders Icon by @tetawaves
- [ ] https://github.com/ramensoftware/windhawk-mods/blob/main/mods/classic-theme-transparency-fix.wh.cpp#L42 Fixes transparency glitches in Classic theme by @Anixx
- [ ] https://github.com/ramensoftware/windhawk-mods/blob/main/mods/disk-pie-chart.wh.cpp#L543 Disk Pie Chart by @aubymori 
- [ ] https://github.com/ramensoftware/windhawk-mods/blob/main/mods/dwm-unextend-frames.wh.cpp#L259 DWM Unextend Frames by @aubymori
- [ ] https://github.com/ramensoftware/windhawk-mods/blob/main/mods/explorer-details-better-file-sizes.wh.cpp#L2113 Better file sizes in Explorer details by @m417z
- [ ] https://github.com/ramensoftware/windhawk-mods/blob/main/mods/favorites-in-navpane.wh.cpp#L294 Favorites in Navigation Pane by @tetawaves 
- [ ] https://github.com/ramensoftware/windhawk-mods/blob/main/mods/fix-basic-caption-text.wh.cpp#L145 Fix Basic Caption Text by @aubymori 
[...]

PR #2085, opened 2025-06-30
Translucent Windows v1.5.1

Undisputed00x:
* Libraries are now loaded with the LOAD_LIBRARY_SEARCH_SYSTEM32 flag to prevent malicious or accidental dll hijacking. https://github.com/ramensoftware/windhawk-mods/issues/2063
* Fixed issue with the accent color not being the correct color in some programs, e.g. snippingtool.exe.
* Fixed a bug with system colors not being applied and unloaded correctly.
* Added logs.

### DIFF
--- a/mods/translucent-windows.wh.cpp
+++ b/mods/translucent-windows.wh.cpp
@@ -2,7 +2,7 @@
 // @id              translucent-windows
 // @name            Translucent Windows
 // @description     Enables native translucent effects in Windows 11
-// @version         1.5.0
+// @version         1.5.1
 // @author          Undisputed00x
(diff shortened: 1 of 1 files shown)


bropines:
Is it possible to fix the transparency of the fonts? When the mod is applied, all the text in 'control' becomes semi-transparent and looks terrible on a white background image

Undisputed00x:
> Is it possible to fix the transparency of the fonts? When the mod is applied, all the text in 'control' becomes semi-transparent and looks terrible on a white background image

The mod already offers an option to make most of the GDI text opaque. I'll look into finding a better solution to this in the future if possible.

PR #2087, opened 2025-07-01
Legacy Shell Message Boxes 1.0.0

aubymori:
This mod makes message boxes opened by the `ShellMessageBox` API use regular message boxes like in Windows XP and before instead of task dialogs, like they do in Windows Vista and later.

**Before**:
Before

**After**:
After

### DIFF
--- a/mods/legacy-shell-message-boxes.wh.cpp
+++ b/mods/legacy-shell-message-boxes.wh.cpp
@@ -5,8 +5,7 @@
 // @version         1.0.0
 // @author          aubymori
 // @github          https://github.com/aubymori
-// @include         explorer.exe
-// @include         notepad.exe
+// @include         *
 // @compilerOptions -lshlwapi -lversion
 // ==/WindhawkMod==
 
@@ -71,7 +70,7 @@ int WINAPI ShellMessageBoxW_hook(
     if (pszMsg)
     {
         g_fInSMB = true;
-        nRet = ShellMessageBoxW_orig(hAppInst, hWnd, pszMsg, lpcTitle, fuStyle);
+        nRet = ShellMessageBoxW_orig(hAppInst, hWnd, L"%s", lpcTitle, fuStyle, pszMsg);
         LocalFree(pszMsg);
     }
     else
@@ -101,7 +100,7 @@ int WINAPI ShellMessageBoxA_hook(
     if (pszMsg)
     {
         g_fInSMB = true;
-        nRet = ShellMessageBoxA_orig(hAppInst, hWnd, pszMsg, lpcTitle, fuStyle);
+        nRet = ShellMessageBoxA_orig(hAppInst, hWnd, "%s", lpcTitle, fuStyle, pszMsg);
         LocalFree(pszMsg);
     }
     else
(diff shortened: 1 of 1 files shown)


mods/legacy-shell-message-boxes.wh.cpp
```diff
@@ -0,0 +1,256 @@
...
+{
+    HMODULE hShell32 = LoadLibraryW(L"shell32.dll");
+    ACTCTXW actCtx = { sizeof(actCtx) };
+    actCtx.dwFlags = ACTCTX_FLAG_RESOURCE_NAME_VALID | ACTCTX_FLAG_HMODULE_VALID;
+    actCtx.lpResourceName = MAKEINTRESOURCEW(124);
+    actCtx.hModule = hShell32;
+    HANDLE hActCtx = CreateActCtxW(&actCtx);
+    ULONG_PTR ulCookie;
+    ActivateActCtx(hActCtx, &ulCookie);
+    HMODULE hComCtl = LoadLibraryW(L"comctl32.dll");
```
m417z: Can you please use `LoadLibrarExW` as discussed recently in https://github.com/ramensoftware/windhawk-mods/issues/2063?

mods/legacy-shell-message-boxes.wh.cpp
```diff
@@ -0,0 +1,256 @@
...
+**After**:
+
+![After](https://raw.githubusercontent.com/aubymori/images/refs/heads/main/legacy-shell-message-boxes-after.png)
+*/
+// ==/WindhawkModReadme==
+
+#include <shlwapi.h>
+#include <windhawk_utils.h>
+
+DWORD g_dwSMBThreadID = (DWORD)-1;
```
m417z: Optional suggestion: Instead of using shared value which can be overridden by different threads, use something like:
```cpp
thread_local bool g_inSMB = false;
```

mods/legacy-shell-message-boxes.wh.cpp
```diff
@@ -0,0 +1,257 @@
...
+)
+{
+    va_list args;
+    va_start(args, fuStyle);
+    LPWSTR pszMsg = ConstructMessageStringW(hAppInst, lpcText, &args);
+    int nRet = 0;
+    if (pszMsg)
+    {
+        g_fInSMB = true;
+        nRet = ShellMessageBoxW_orig(hAppInst, hWnd, pszMsg, lpcTitle, fuStyle);
```
m417z: If I understand correctly from a quick glance, the correct way to call this is:
```suggestion
        nRet = ShellMessageBoxW_orig(hAppInst, hWnd, L"%s", lpcTitle, fuStyle, pszMsg);
```
As you don't want `pszMsg` to be formatted, right?

mods/legacy-shell-message-boxes.wh.cpp
```diff
@@ -0,0 +1,257 @@
+// ==WindhawkMod==
+// @id              legacy-shell-message-boxes
+// @name            Legacy Shell Message Boxes
+// @description     Makes ShellMessageBox use message boxes rather than task dialogs
+// @version         1.0.0
+// @author          aubymori
+// @github          https://github.com/aubymori
+// @include         explorer.exe
+// @include         notepad.exe
```
m417z: Was this an intentional change in the last commit?

PR #2127, opened 2025-07-12
Explorer Double Click Up v1.0.1

wrldspawn:
- Use LoadLibraryEx with LOAD_LIBRARY_SEARCH_SYSTEM32 #2063


OsamaHJT:
@wrldspawn sorry for the tag, I have a suggestion for this mod, I hope I'm doing this in the right place.

first, thank you for this mod, it's very handy. I want to suggest adding a similar shortcut to toggle hidden items, maybe using a modifier key with double clicking, maybe on the same mod or as a separate one.

I don't know programming so suggesting this was my last hope.

I hope to hear back from you.
thank you again!

PR #2156, opened 2025-07-16
Added prevent-focus-stealing.wh.cpp v1.0

bawNg:


mods/prevent-focus-stealing.wh.cpp
```diff
@@ -0,0 +1,371 @@
+// ==WindhawkMod==
+// @id              prevent-focus-stealing
+// @name            Prevent Focus Stealing
+// @description     Prevents specific apps from stealing focus when a window is opened.
+// @version         1.0.0
+// @author          bawNg
+// @github          https://github.com/bawNg
+// @include         *
+// @architecture    x86-64
```
m417z: If all processes are targeted, why limit the mod to x86-64?

mods/prevent-focus-stealing.wh.cpp
```diff
@@ -0,0 +1,371 @@
...
+// ==/WindhawkModSettings==
+
+#include <windhawk_utils.h>
+#include <windows.h>
+#include <shlwapi.h>
+#include <vector>
+#include <set>
+#include <map>
+
+#define WM_USER_REMOVE_NOACTIVATE (WM_USER + 1)
```
m417z: It's not safe to assume that this message isn't used by windows for their own needs. Consider using `RegisterWindowMessage` instead.

But do you even need it? Why not just call `SetWindowLongPtrW(hWnd, GWL_EXSTYLE, ...)` from another thread?

mods/prevent-focus-stealing.wh.cpp
```diff
@@ -0,0 +1,371 @@
...
+        g_hookedHwnds.erase(hWnd);
+        g_blockedHwnds.erase(hWnd);
+    }
+    if (uMsg == WM_USER_REMOVE_NOACTIVATE) {
+        Wh_Log(L"Removing WS_EX_NOACTIVATE from window 0x%p", hWnd);
+        SetWindowLongPtrW(hWnd, GWL_EXSTYLE, GetWindowLongPtrW(hWnd, GWL_EXSTYLE) & ~WS_EX_NOACTIVATE);
+    }
+    if (uMsg == WM_ACTIVATE && (wParam == WA_ACTIVE || wParam == WA_CLICKACTIVE)) {
+        auto it = g_blockedHwnds.find(hWnd);
+        if (it != g_blockedHwnds.end() && GetTickCount64() - it->second < INITIAL_UI_FOCUS_BLOCK_MS) {
```
m417z: I'm not sure about the chances for real issues, but `g_blockedHwnds` may be accessed from different thread, which is unsafe. Since it's used to store window-specific data, you can use `SetProp`/`GetProp`, which should also result in simpler code.

mods/prevent-focus-stealing.wh.cpp
```diff
@@ -0,0 +1,371 @@
...
+        }
+    }
+Done:
+    if (g_includedWindows.empty()) {
+        return FALSE;
+    }
+
+    if (!Wh_SetFunctionHook((void *)CreateWindowExW, (void *)CreateWindowExW_Hook, (void **)&CreateWindowExW_Orig)) {
+        Wh_Log(L"Failed to hook CreateWindowExW");
+        return FALSE;
```
m417z: Note that in this case, `Wh_FreeStringSetting` isn't called for allocated data.

You might want to consider using [`WindhawkUtils::StringSetting`](https://github.com/ramensoftware/windhawk/wiki/Development-tips#notable-utilities) which calls `Wh_FreeStringSetting` on destruction.

mods/prevent-focus-stealing.wh.cpp
bawNg: That was just left over from the mod I forked to create this one - will remove it.

mods/prevent-focus-stealing.wh.cpp
bawNg: This is a good point, will have to address this. Delaying until the next message loop tick was necessary to make it work reliably.

mods/prevent-focus-stealing.wh.cpp
bawNg: g_blockedHwnds is only accessed from `CreateWindowExW_Hook`, `SetWindowTextW_Hook`, and the subclass proc, which I assumed all had to be called from the message loop thread?

m417z: Any thread can call `SetWindowTextW`, and several threads can have a message loop. The simplest example is: create two threads, and call MessageBox in each,

BTW `SetWindowTextW` can be called across processes too, and there are other way to set a window text (e.g. `WM_SETTEXT`).

m417z:
Is it ready to be merged?

bawNg:
Yeah, it should be good. You can go ahead and merge it.

PR #2162, opened 2025-07-17
Basic Themer 1.1.1

aubymori:
Fix for Windows 11 24H2

### DIFF
--- a/mods/basic-themer.wh.cpp
+++ b/mods/basic-themer.wh.cpp
@@ -2,7 +2,7 @@
 // @id              basic-themer
 // @name            Basic Themer
 // @description     Applies the Windows Basic theme to desktop windows
-// @version         1.1.0
+// @version         1.1.1
 // @author          aubymori
 // @github          https://github.com/aubymori
 // @include         dwm.exe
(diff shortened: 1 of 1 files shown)


mods/basic-themer.wh.cpp
```diff
@@ -286,7 +287,8 @@ void LoadSettings(void)
     }
 }
 
-const WindhawkUtils::SYMBOL_HOOK hooks[] = {
+// Windows 11 23H2 and before
+const WindhawkUtils::SYMBOL_HOOK uDWMDllHooks[] = {
     {
         {
             L"public: long __cdecl CWindowList::GetSyncedWindowDataByHwnd(struct HWND__ *,class CWindowData * *)"
```
m417z: Instead of adding `uDWMDllHooks_win11`, it'd be simpler to just add the alternative symbol to the array here, like:

```suggestion
            L"public: long __cdecl CWindowList::GetSyncedWindowDataByHwnd(struct HWND__ *,class CWindowData * *)",
            L"public: void __cdecl CWindowList::GetSyncedWindowDataByHwnd(struct HWND__ *,class CWindowData * *)"
```

More importantly, having two `HookSymbols` calls for the same module will screw up the caching. It'd be fine if you checked the Windows/module version in runtime and called only right one, but you're always calling both.

aubymori:
I kinda forgot about this and didn't get notified of your comments, I'll re-open and fix once my other PR is merged.

PR #2167, opened 2025-07-19
Add Internet Status Indicator Mod

almas-cp:


mods/internet-status-indicator.wh.cpp
```diff
@@ -0,0 +1,960 @@
+// ==WindhawkMod==
+// @id              internet-status-indicator
+// @name            Internet Status Indicator
+// @description     Real-time network connectivity monitoring with visual indicators as a Tray Icon
+// @version         0.7
+// @author          ALMAS CP
+// @github          https://github.com/almas-cp
+// @homepage        https://github.com/almas-cp
+// @include         explorer.exe
+// @architecture    amd64
```
m417z: Any reason to limit it to amd64?

mods/internet-status-indicator.wh.cpp
```diff
@@ -0,0 +1,960 @@
...
+
+void LoadSettings() {
+    NetworkSettings settings;
+    
+    settings.checkInterval = Wh_GetIntSetting(L"checkInterval");
+    
+    // Convert wide string settings to narrow strings
+    PCWSTR targetHostW = Wh_GetStringSetting(L"targetHost");
+    PCWSTR secondaryHostW = Wh_GetStringSetting(L"secondaryHost");
+    PCWSTR httpTestUrlW = Wh_GetStringSetting(L"httpTestUrl");
```
m417z: These strings are never freed, a call to `Wh_FreeStringSetting` is missing.

almas-cp:
You're right
I have made the changes as specified, is there anything else?

m417z:
Several users reported that the tray icon isn't being created on startup. The reason is that the mod doesn't handle [the taskbar creation notification](https://learn.microsoft.com/en-us/windows/win32/shell/taskbar?redirectedfrom=MSDN#taskbar-creation-notification).

almas-cp:
Changing the target process from windhawk.exe to explorer.exe  fixes the issue, do you want to proceed with that?

m417z:
Maybe it makes it less likely because the mod starts working at a later stage this way, but it can still run too early, before the tray is ready. Also, it won't restore the tray icon if explorer crashes or restarts.

Have you tried adding handling for the "TaskbarCreated" message as I suggested?

Issue #1210, opened 2024-11-07
Feature Request - Do not show selected systray icons


Marshall-Brooks:
@m417z 
I started discussing this here: https://github.com/ramensoftware/windhawk-mods/issues/1170 but I noticed some of the UWP Spy Window that I included doesn't show up.

PS Tray Factory does work and works nicely for this. Tray Control and the AHK script I posted do not work any longer.

I made a little bit of progress:
![image](https://github.com/user-attachments/assets/d664bd3b-2042-4dd0-872f-bb3d0f1a2c74)
Above is the entry for the Windows Security Center Icon.  If I go to the entry above it - Windows.UI.Xaml.Controls.ContentPresenter - and set Visiblity=Collapsed. The icon goes away, which is part of what I want to do. However, it still leaves a blank Space where the icon "Used" to be:
![image](https://github.com/user-attachments/assets/7f571699-e493-45d6-944b-aaa8d1ef4e5b)
Remaining issues:

- I don't know how to get rid of the grid space used by the icon.
- There are a lot of controls with the same name. I don't know how to hide the .ContentPresenter for the icon I want to remove as opposed to the ones I don't want to remove.
- Ideally, I would want the mod to never show this icon - i.e. if I closed and re-opened the program that I want hidden, the icon should never show up. However, I would be okay with the program simply removing the icon if it existed and having to wait a bit after startup  before the mod ran - but that makes it awkward for windhawk b/c I don't want to restart WH after 5 minutes nor wait 5 minutes before starting WH.

m417z:
It's not a good fit for an enhancement issue for Windows 11 Taskbar Styler, since you're basically looking for a specific style. You might also want to check/ask [here](https://github.com/ramensoftware/windows-11-taskbar-styling-guide/issues).

It's also not a good for an enhancement issue for Taskbar Tray Icon Spacing, since it's not about spacing.

HenryTheVacuum:
![ApplicationFrameHost_Y6oL4hAZxr](https://github.com/user-attachments/assets/ced49025-b87c-4dab-983d-399c2c9da814)

You can turn them on and off without mods in windows settings.

Marshall-Brooks:
@m417 - Agreed, it would be best as a new separate modification, however, it does look like it could be controlled from Taskbar Styler. I use that for a lot of things not specifically styling - for example, hiding the "bell" notification icon.

I'll ask in the thread you linked to.

[...]

PR #2195, opened 2025-07-23
Disable Tray Icons 1.0.0

aubymori:
This mod will disable all tray icons for a specific application.

### DIFF
--- a/mods/disable-tray-icons.wh.cpp
+++ b/mods/disable-tray-icons.wh.cpp
@@ -1,0 +1,66 @@
+// ==WindhawkMod==
+// @id              disable-tray-icons
+// @name            Disable Tray Icons
+// @description     Disable all tray icons for a specific application
+// @version         1.0.0
+// @author          aubymori
+// @github          https://github.com/aubymori
+// @compilerOptions -lshell32
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Disable Tray Icons
+This mod will disable all tray icons for a specific application.
+
+## How to use
+1. Go to the "Advanced" tab of this mod
+2. Find the "Custom process inclusion list"
+3. Add the name of the EXE you want to disable tray icons for.
+   - If you don't know the name of the EXE, you can find it using 
+     Task Manager.
+4. Click Save, and if the program is open already, restart it.
+*/
+// ==/WindhawkModReadme==
+
+#include <shellapi.h>
+
+using Shell_NotifyIconW_t = decltype(&Shell_NotifyIconW);
+Shell_NotifyIconW_t Shell_NotifyIconW_orig;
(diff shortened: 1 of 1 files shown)


m417z:
Neat! This mostly addresses https://github.com/ramensoftware/windhawk-mods/issues/1210.

I thought about a similar mod too, and my idea was to target Explorer and do something like (pseudo-code):
```cpp
TaskbarWndProc(...) {
  switch (signature) {
  case WM_COPYDATA:
    if (magic == 0x34753423) {
      return TRUE;
    }
    break;
  }
  return default(...);
}
```

But this simple approach works as well.

Marshall-Brooks:
Not sure if this is the correct place to post this, but ...

First off - thanks to @aubymori and @m417z for this mod. I was able to disable a tray icon with it that I wasn't able to disable with #1210.

I have two comments that weren't obvious to me, and a question:

- First, if you don't allow Windhawk access to all processes, you also have to go into the Advance Settings for Windhawk and include the process that you are including in the mod.

- Second, it might work otherwise, but it worked for me with the .exe of the file rather than the name of the file, i.e. if you aren't sure where the icon comes from, you need to go to Task Manager and find the process, then right-click on that and go to details to figure out the .exe of the process.

Question - what would the process be for "Safely Remove Hardware and Eject Media" I don't see it all the time. I was able to hide it using #1210, but that still leaves the space it occupies on the grid.  #1210 works, but I would prefer to disable it totally.

Thank you again!

m417z:
> Question - what would the process be for "Safely Remove Hardware and Eject Media"

It's explorer.exe.

Marshall-Brooks:
For the second comment, I'm going to try starting WH earlier in my startup routine - still a manual process, but hopefully ahead of the system process.

What I may end up doing is using both mods - i.e. ideally Disable Tray Icons will disable the tray icon, if the tray program starts first, then #1210 will hide the icon, but the space it occupies will still be used.

(That or look into closing and re-opening the program I am trying to block in my second startup routine ...)

`It's explorer.exe.`

@m417z Thank you!

@aubymori - That brings up another hypothetical issue for the mod. If a certain program launches two or more systray icons and I only want to disable one of them, I don't see a way to do so with this mod. Hypothetical examples: 

- Program B has a systray icon that normally just tells me the program is running and I don't care. However, the icon turns yellow when an update is available and I would like to know that. I can't hide the normal icon and show the yellow one. (I possibly could with #1210 if the hover text was different for the different states.

- I use a program called KeyboardNotifier which uses a single icon for status, but there are similar programs with three icons for Caps Lock, Num Lock, and Scroll Lock. My keyboard doesn't even have Scroll Lock and it really isn't used outside of the console window. Unless the program supports it or uses different .exe's for each tray icon, I couldn't hide the scroll lock indicator.

But I don't see an easy solution for these options, and they are hypothetical anyway.

PR #2202, opened 2025-07-25
Add Hide Network UNC Paths in Explorer 1.0

danalec:
This mod removes UNC paths from network drives in Explorer and dialogs.
It declutters lengthy network drive names in Explorer.

Tested on Windows 11 Pro N 24H2.


mods/hide-unc.wh.cpp
```diff
@@ -0,0 +1,356 @@
...
+    WCHAR* exeName = wcsrchr(exePath, L'\\');
+    if (exeName) exeName++; else exeName = exePath;
+    
+    if (wcsicmp(exeName, L"csrss.exe") == 0 || 
+        wcsicmp(exeName, L"dwm.exe") == 0 ||
+        wcsicmp(exeName, L"winlogon.exe") == 0 ||
+        wcsicmp(exeName, L"services.exe") == 0 ||
+        wcsicmp(exeName, L"lsass.exe") == 0) {
+        return TRUE;
+    }
```
aubymori: This isn't needed; you can simply add `@exclude` statements to your mod heaer to exclude these applications
Example:
https://github.com/ramensoftware/windhawk-mods/blob/dfdd41d6b0c84e75c65896510c3ae1a38252647e/mods/eradicate-immersive-menus.wh.cpp#L9-L15

Also, FYI, if this code were needed, it should have returned `FALSE` instead of `TRUE` to remove the mod from those applications instead of keeping it loaded.

mods/hide-unc.wh.cpp
```diff
@@ -0,0 +1,348 @@
...
+    WCHAR* unc = wcsstr(text, L" (\\\\");
+    if (unc) {
+        WCHAR* p = unc + 2;
+        int depth = 1;
+        while (*p && depth > 0) {
+            if (*p == L'(') depth++;
+            else if (*p == L')') depth--;
+            p++;
+        }
+        if (*p) wcscpy_s(unc, maxLen - (unc - text), p);
```
m417z: Is this correct?

> The wcscpy function, used for copying wide-character strings, has undefined behavior when the source and destination strings overlap.

If the strings can overlap, you might need to use `memmove`.

mods/hide-unc.wh.cpp
```diff
@@ -0,0 +1,348 @@
...
+    HRESULT hr = SHCreateItemFromParsingName_Original(pszPath, pbc, riid, ppv);
+    
+    if (SUCCEEDED(hr) && ppv && *ppv) {
+        void*** vtbl = (void***)*ppv;
+        if (vtbl && *vtbl) {
+            if (!GetDisplayName_Original) {
+                GetDisplayName_Original = (GetDisplayName_t)(*vtbl)[5];
+                DWORD oldProtect;
+                if (VirtualProtect(&(*vtbl)[5], sizeof(void*), PAGE_READWRITE, &oldProtect)) {
+                    (*vtbl)[5] = (void*)GetDisplayName_Hook;
```
m417z: Are these vtbl hooks removed when the mod is unloaded?

mods/hide-unc.wh.cpp
```diff
@@ -0,0 +1,348 @@
...
+CreateWindowExW_t CreateWindowExW_Original;
+
+HWND WINAPI CreateWindowExW_Hook(DWORD dwExStyle, LPCWSTR lpClassName, LPCWSTR lpWindowName, 
+    DWORD dwStyle, int X, int Y, int nWidth, int nHeight, HWND hWndParent, HMENU hMenu, HINSTANCE hInstance, LPVOID lpParam) {
+    
+    HWND hWnd = CreateWindowExW_Original(dwExStyle, lpClassName, lpWindowName, dwStyle, X, Y, nWidth, nHeight, hWndParent, hMenu, hInstance, lpParam);
+    
+    if (hWnd && lpClassName) {
+        WCHAR className[256] = {0};
+        if (HIWORD(lpClassName)) {
```
m417z: That's not a correct check, it can be a valid address with a zero HIWORD such as `12341234_00000000`.

mods/hide-unc.wh.cpp
m417z: You might want to use `Wh_SetFunctionHook` and `Wh_ApplyHookOperations` which doesn't have this problem, and which is also safer in case multiple mods hook the same function.

PR #2209, opened 2025-07-25
Hide Dotfiles (Explorer Only) v1.0.1

danalec:
Implements COM interface hooking to filter dot-prefixed files/folders from Explorer views while maintaining full system accessibility. Designed for cross-platform developers who want clean GUI browsing without breaking toolchain workflows.

Coexists with Global Solutions: Designed to work alongside system-wide dotfile hiding mods like [dot-hide](https://windhawk.net/mods/dot-hide)

In other words: this Hide Dotfiles functionality is purely cosmetic - it only hides files from view **only** in explorer.exe

### DIFF
--- a/mods/hide-dotfiles-explorer.wh.cpp
+++ b/mods/hide-dotfiles-explorer.wh.cpp
@@ -170,8 +170,8 @@ bool ShouldHideFile(std::wstring_view fileName) noexcept {
     }
     
     auto matchesPattern = [](std::wstring_view name, const std::vector<std::wstring>& patterns) noexcept -> bool {
-        return std::ranges::any_of(patterns, [name](const std::wstring& pattern) {
-            std::wstring temp(name);
+        std::wstring temp(name);
+        return std::ranges::any_of(patterns, [&temp](const std::wstring& pattern) {
             return PathMatchSpecW(temp.c_str(), pattern.c_str()) != FALSE;
         });
     };


mods/hide-dotfiles-explorer.wh.cpp
m417z, line 1: Looks good!

Optional suggestion: How about adding an option to exclude certain file/folder names, like it was suggested here: https://github.com/ramensoftware/windhawk-mods/pull/669

If you prefer to keep it as is or add it later, that's OK too. Let me know what you prefer.

danalec: I'll look onto that. Thanks for pointing me out.

danalec:
@m417z Can we have a feature to restart the process included?

I mean by example: this mod changes explorer.exe behavior but user needs to trigger `Stop-Process -Name explorer -Force; Remove-Item "$env:LOCALAPPDATA\IconCache.db" -Force -EA 0; Start-Process explorer`

Restart Process Button: Add a button in the UI that, when clicked, restarts the current process (e.g., the application or service).
On-demand via the button, or Automatically after mod activation (hook into the mod activation event).
Permissions: Ensure these tasks run with the required elevated privileges (if needed, prompt UAC).

mods/hide-dotfiles-explorer.wh.cpp
```diff
@@ -1,24 +1,105 @@
...
+Run this PowerShell command to restart Explorer and clear the icon cache:
+```powershell
+Stop-Process -Name explorer -Force; Remove-Item "$env:LOCALAPPDATA\IconCache.db" -Force -EA 0; Start-Process explorer
+```
+*/
+// ==/WindhawkModReadme==
+
+// ==WindhawkModSettings==
+/*
+- excludePaths: ""
```
m417z: * I think it makes more sense to make it an array. Example:
  https://github.com/ramensoftware/windhawk-mods/blob/4f480eff9a5cecad737798bfafa461e1fb23c250/mods/icon-resource-redirect.wh.cpp#L283-L287
* The request I linked wasn't about excluding paths (which can be helpful too), but about always showing a file/folder with a specific name, regardless of where it's located. For example, always show `.github` folders.

mods/hide-dotfiles-explorer.wh.cpp
```diff
@@ -86,73 +198,142 @@ class DotfileFilterEnum : public IEnumIDList {
...
-    HMODULE hShell32 = LoadLibrary(L"shell32.dll");
-    if (!hShell32) return FALSE;
+    LoadSettings();
     
     IShellFolder* pDesktop = nullptr;
     HRESULT hr = SHGetDesktopFolder(&pDesktop);
-    if (FAILED(hr) || !pDesktop) return FALSE;
+    
+    if (FAILED(hr) || !pDesktop) {
+        return FALSE;
```
m417z, line 366: Note that when FALSE is returned, you have a memory leak, `Wh_FreeStringSetting` isn't called in this case. In your case, you don't need to keep it, you can free it right in `LoadSettings`.

mods/hide-dotfiles-explorer.wh.cpp
danalec: I'll take a look in how to make it an array but it already works with path or file or even folder.
 
When it is excluded, it always shows. Or did I get wrong?

m417z: IMO it's not clear from the readme and the option description.

m417z:
> Can we have a feature to restart the process included?

Currently that's not planned, but it can be implemented by the mod, for example:

https://github.com/ramensoftware/windhawk-mods/blob/3326875811c957ae1eb55fe8c57bd218c464aa90/mods/icon-resource-redirect.wh.cpp#L2093

danalec:
> > Can we have a feature to restart the process included?
> 
> Currently that's not planned, but it can be implemented by the mod, for example:
> 
> https://github.com/ramensoftware/windhawk-mods/blob/3326875811c957ae1eb55fe8c57bd218c464aa90/mods/icon-resource-redirect.wh.cpp#L2093

Sorry, I've got confused. Does clearing the cache restarts the only included process like explorer.exe?

danalec:
> The example that I linked runs this command which only restarts explorer.exe:
> 
> https://github.com/ramensoftware/windhawk-mods/blob/3326875811c957ae1eb55fe8c57bd218c464aa90/mods/icon-resource-redirect.wh.cpp#L369-L379

I managed to update without restarting the process but I learned a lot, thank you. I might use it in another mod in the near future

mods/hide-dotfiles-explorer.wh.cpp
danalec: Have I now managed to get around the memory leak problem? There have been so many changes that I am uncertain...

mods/hide-dotfiles-explorer.wh.cpp
```diff
@@ -0,0 +1,311 @@
...
+- Configurable exclusion paths where dotfiles will remain visible
+
+## Configuration
+You can exclude specific paths, folders, or files from having their dotfiles hidden. In the mod settings, add paths separated by commas:
+- Single path: C:\Projects
+- Multiple paths: C:\Program Files", "D:\My Documents"
+
+You can also exclude specific folders or files directly, for example:
+- Folder: C:\Projects\.git
+- File: C:\Projects\.env
```
m417z: So, can you just specify ".git" or ".env" to exclude all such folders/files? If yes, I think it's worth documenting it.

mods/hide-dotfiles-explorer.wh.cpp
m417z: You marked it as resolved but you didn't change the option to be an array.

mods/hide-dotfiles-explorer.wh.cpp
m417z: Yes, looks good now.

mods/hide-dotfiles-explorer.wh.cpp
```diff
@@ -0,0 +1,311 @@
...
+
+## Restarting Explorer after enabling
+Run this PowerShell command to restart Explorer and clear the icon cache:
+```powershell
+Stop-Process -Name explorer -Force; Remove-Item "$env:LOCALAPPDATA\IconCache.db" -Force -EA 0; Start-Process explorer
+```
```
m417z: > I managed to update without restarting the process

So if the PowerShell command isn't needed anymore, you might want to remove it.

m417z:
Is it ready? What about the settings which could be an array? You marked it as resolved (twice) but it's still a string.

danalec:
> Is it ready? What about the settings which could be an array? You marked it as resolved (twice) but it's still a string. 

Almost, maybe tomorrow I'll finished it.
I think it can ignore it even as a string multiple targets...

mods/hide-dotfiles-explorer.wh.cpp
```diff
@@ -0,0 +1,290 @@
...
+    
+    HMODULE hShell32 = GetModuleHandle(L"shell32.dll");
+    if (!hShell32) return FALSE;
+    
+    void* pEnumObjects = nullptr;
+    for (int ordinal = 1; ordinal < 100; ordinal++) {
+        FARPROC pFunc = GetProcAddress(hShell32, (LPCSTR)ordinal);
+        if (pFunc) {
+            BYTE* pBytes = (BYTE*)pFunc;
+            if (pBytes[0] == 0x48 && pBytes[1] == 0x89 && pBytes[2] == 0x5C) {
```
m417z: This doesn't look very reliable, why was it added? Why not always use `SHGetDesktopFolder` like it was done previously?

danalec: Sure, I will take a look.

danalec:
if it is ok, I'll be adding screenshots and we're ready.

m417z:
Looks like you changed most of the code, now using hooks similar to what [Dot Hide](https://windhawk.net/mods/dot-hide) uses instead of the previous `IShellFolder_EnumObjects_Hook`. Any reason for that?

danalec:
> Looks like you changed most of the code, now using hooks similar to what [Dot Hide](https://windhawk.net/mods/dot-hide) uses instead of the previous `IShellFolder_EnumObjects_Hook`. Any reason for that?

My initial thought was to approach targeting NtQueryDirectoryFile and FindFirstFile to Explorer's folder views.
- intercept at the kernel boundary and filter dotfiles before they even reach user-mode applications.
This would ensure universal coverage across all processes... then I've faced NtQueryDirectoryFile and it was too complex for me.

After investigating alternative methods, I discovered the IShellFolder::EnumObjects COM interface.
Hooking into the Shell's documented extensibility model rather than intercepting low-level kernel operations would be much better and fit in Windhawk's.

The IShellFolder interface has remained stable since the Windows 9x era. Unlike many NT APIs, which are historically notorious for frequent changes between builds, COM offered a simpler and more reliable choice. I chose it for that reason, though I did struggle with intercepting directory enumeration. Fortunately, the documentation is solid, which gave me more confidence.

danalec:
> Looks like you changed most of the code, now using hooks similar to what [Dot Hide](https://windhawk.net/mods/dot-hide) uses instead of the previous `IShellFolder_EnumObjects_Hook`. Any reason for that?

Simplified the hooking approach because my initial approach of using NtQueryDirectoryFile and FindFirstFile to Explorer's folder views was too complex. IShellFolder::EnumObjects COM interface is way better documented.

I've found a problem in how it is dealing with the names (name with spaces, by example: "New Folder"). I'll probably going to need to use the previous approach too. I'm just having difficult time with the crashes...

mods/hide-dotfiles-explorer.wh.cpp
```diff
@@ -0,0 +1,690 @@
+// ==WindhawkMod==
+// @id              hide-dotfiles-explorer
+// @name            Hide Dotfiles (Explorer only) v1.0.1
```
m417z: There's no need to add the version to the name, the `@version` field is used for that.

mods/hide-dotfiles-explorer.wh.cpp
```diff
@@ -0,0 +1,690 @@
...
+            wcscpy_s(g_currentPath, MAX_PATH, path);
+        }
+        return true;
+    }
+    
+    g_currentPath[0] = L'\0';
+    return false;
+}
+
+class FileVisibilityFilterEnum : public IEnumIDList {
```
m417z: This class doesn't seem to be used.

mods/hide-dotfiles-explorer.wh.cpp
```diff
@@ -0,0 +1,690 @@
...
+    
+    if (!Wh_SetFunctionHook((void*)NtQueryDirectoryFileEx_Original, (void*)NtQueryDirectoryFileEx_Hook, (void**)&NtQueryDirectoryFileEx_Original)) {
+        return FALSE;
+    }
+    
+    if (!Wh_SetFunctionHook((void*)FindFirstFileW_Original, (void*)FindFirstFileW_Hook, (void**)&FindFirstFileW_Original)) {
+        return FALSE;
+    }
+    
+    if (!Wh_SetFunctionHook((void*)FindNextFileW_Original, (void*)FindNextFileW_Hook, (void**)&FindNextFileW_Original)) {
```
m417z: Is it necessary to hook `FindFirstFileW` and `FindNextFileW` now that you hook `NtQueryDirectoryFile` and `NtQueryDirectoryFileEx` like the Dot Hide does?

And generally, it now looks like your mod is very similar to Dot Hide, just with extra customization settings.

mods/hide-dotfiles-explorer.wh.cpp
```diff
@@ -0,0 +1,690 @@
...
+        if (str.empty()) return;
+        
+        std::wstring current;
+        current.reserve(str.size());
+        
+        for (wchar_t c : str) {
+            if (c == L',') {
+                current = TrimWhitespace(current);
+                if (!current.empty()) {
+                    current = ParseQuotedString(current);
```
m417z: I assumed that the support for quotes is for folder names with a comma, such as:
```
C:\folder1, C:\folder2, "C:\folder,with,comma,3", C:\folder4
```
But it doesn't look like it's being handled.

mods/hide-dotfiles-explorer.wh.cpp
```diff
@@ -0,0 +1,690 @@
...
+}
+
+thread_local WCHAR g_currentPath[MAX_PATH] = {0};
+
+bool GetCurrentDirectoryFromHandle(HANDLE fileHandle) {
+    WCHAR path[MAX_PATH] = {0};
+    DWORD pathLen = GetFinalPathNameByHandleW(fileHandle, path, MAX_PATH, FILE_NAME_NORMALIZED);
+    
+    if (pathLen > 0 && pathLen < MAX_PATH) {
+        if (wcsncmp(path, L"\\\\?\\\\?", 4) == 0) {
```
m417z: Did you perhaps mean:
```
        if (wcsncmp(path, L"\\\\?\\", 4) == 0) {
```

mods/hide-dotfiles-explorer.wh.cpp
```diff
@@ -0,0 +1,690 @@
...
+            ProcessDirectoryListing(FileInformation, FileInformationClass, &bytesReturned);
+            IoStatusBlock->Information = bytesReturned;
+        }
+    }
+    
+    return status;
+}
+
+void LoadSettings() {
+    std::lock_guard<std::mutex> lock(g_hookMutex);
```
m417z: The mutex is only locked here, it doesn't seem to be necessary.

mods/hide-dotfiles-explorer.wh.cpp
```diff
@@ -0,0 +1,690 @@
...
+    ULONG Length,
+    FILE_INFORMATION_CLASS FileInformationClass,
+    ULONG QueryFlags,
+    PUNICODE_STRING FileName
+);
+
+NtQueryDirectoryFileEx_t NtQueryDirectoryFileEx_Original;
+
+typedef HRESULT (WINAPI* SHCreateShellFolderView_t)(LPCITEMIDLIST pidl, IShellFolderView** ppv);
+SHCreateShellFolderView_t SHCreateShellFolderView_Original;
```
m417z: Unused.

mods/hide-dotfiles-explorer.wh.cpp
```diff
@@ -0,0 +1,690 @@
...
+}
+
+BOOL WINAPI FindNextFileW_Hook(HANDLE hFindFile, LPWIN32_FIND_DATAW lpFindFileData) {
+    BOOL result = FindNextFileW_Original(hFindFile, lpFindFileData);
+    
+    while (result && lpFindFileData) {
+        WCHAR fileName[MAX_PATH];
+        wcscpy_s(fileName, MAX_PATH, lpFindFileData->cFileName);
+        
+        if (!ShouldHideFile(nullptr, fileName)) {
```
m417z: Doesn't `FindNextFileW` also return `.` and `..`? These aren't being checked in `ShouldHideFile`.

mods/hide-dotfiles-explorer.wh.cpp
```diff
@@ -0,0 +1,690 @@
...
+        if (nextEntryOffset == 0) {
+            currentEntrySize = *bytesReturned - totalBytesRead;
+        }
+        
+        bool shouldHide = false;
+        if (fileName && fileNameLength > 0) {
+            std::wstring fileNameStr(fileName, fileNameLength);
+            
+            if (fileNameStr != L"." && fileNameStr != L"..") {
+                shouldHide = ShouldHideFile(g_currentPath, fileNameStr.c_str());
```
m417z: Looks like `g_currentPath` is never used in the function.

mods/hide-dotfiles-explorer.wh.cpp
danalec: yes but it only hides the icons. the NtQueryDirectoryFile is necessary for the whitelist

mods/hide-dotfiles-explorer.wh.cpp
danalec: paths aren't supported anymore, only folder names and file names because it is going to handle only the icons. there is no need to handle paths. 

this Hide Dotfiles functionality is purely cosmetic - it only hides files from view in explorer.exe.

that's why complete directory paths aren't necessary, since it's just for visually hiding dotfiles.

mods/hide-dotfiles-explorer.wh.cpp
danalec: the actual string has four backslashes followed by a question mark ( \\?\\? ), not three backslashes ( \\?\\ ). The comparison checks if the path starts with the Windows long path prefix \\?\ which is represented as L"\\?\\?" in the source code due to C++ string escaping.

mods/hide-dotfiles-explorer.wh.cpp
danalec: mutex was indeed unnecessary and has been removed

mods/hide-dotfiles-explorer.wh.cpp
danalec: as I mentioned, I simplified the "cosmetic" hiding of icons only operation so I no longer saw the need for g_currentPath and forgot to remove it...

mods/hide-dotfiles-explorer.wh.cpp
m417z: It's not about paths or names. The same can apply for just file names:
```
name, name2, "name,with,comma,3", name4
```
I just don't understand why you have handling for quotes if commas between quotes aren't handled.

mods/hide-dotfiles-explorer.wh.cpp
m417z: What do you mean? To make my question more specific: If you remove the `FindFirstFileW` and `FindNextFileW` hooks, and leave `NtQueryDirectoryFile` and `NtQueryDirectoryFileEx`, what will change?

danalec: I see now, I can remove that because it is not longer needed system-wide hiding..

mods/hide-dotfiles-explorer.wh.cpp
danalec: the current parsing logic splits on every comma without considering whether it's inside quotes.

the ParseQuotedString function exists to handle quoted strings, but it's only applied after splitting, which means commas within quotes are treated as separators 😓

I need to implement proper CSV parsing that respects the quoted strings...

thanks for pointing it out. sorry for not understand it earlier

m417z: I think it will be easier to just have the settings as an array instead of a comma-separated list, as I noted earlier. Then no parsing will be needed. Here's an example I posted earlier:
https://github.com/ramensoftware/windhawk-mods/blob/4f480eff9a5cecad737798bfafa461e1fb23c250/mods/icon-resource-redirect.wh.cpp#L283-L287

mods/hide-dotfiles-explorer.wh.cpp
```diff
@@ -0,0 +1,368 @@
...
+- **No file attribute modification**: Files remain unchanged on disk
+- **System-wide filtering**: Affects all Explorer views and file dialogs
+- **Performance optimized**: Uses low-level NT API hooks for minimal overhead
+
+## Configuration
+You can exclude specific files from being hidden using the dotfile whitelist, or specify 
+additional files to always hide regardless of whether they start with a dot.
+
+### Filename Matching
+- **Case-sensitive**: Filenames must match exactly including case (e.g., `.Gitignore` will not match `.gitignore`)
```
m417z: That doesn't seem to be the case, the code uses `towlower`.

mods/hide-dotfiles-explorer.wh.cpp
```diff
@@ -0,0 +1,368 @@
...
+NtQueryDirectoryFileEx_t NtQueryDirectoryFileEx_Original;
+
+std::wstring TrimWhitespace(const std::wstring& str) {
+    size_t start = str.find_first_not_of(L" \t\r\n");
+    if (start == std::wstring::npos) return L"";
+    size_t end = str.find_last_not_of(L" \t\r\n");
+    return str.substr(start, end - start + 1);
+}
+
+std::wstring ToLowerCase(const std::wstring& str) {
```
m417z: Unused.

mods/hide-dotfiles-explorer.wh.cpp
```diff
@@ -0,0 +1,368 @@
...
+  $description: Comma-separated list of filenames to always hide regardless of dotfile status
+*/
+// ==/WindhawkModSettings==
+
+#include <windows.h>
+#include <shlobj.h>
+#include <shlwapi.h>
+#include <shobjidl.h>
+#include <winternl.h>
+#include <windhawk_utils.h>
```
m417z: `windhawk_utils.h` seems to be unused.

mods/hide-dotfiles-explorer.wh.cpp
```diff
@@ -0,0 +1,368 @@
...
+        return;
+    }
+    
+    FileInfoType* currentEntry = static_cast<FileInfoType*>(FileInformation);
+    FileInfoType* writeEntry = currentEntry;
+    ULONG_PTR totalBytesRead = 0;
+    ULONG_PTR totalBytesWritten = 0;
+    
+    while (totalBytesRead < *bytesReturned) {
+        ULONG nextEntryOffset = currentEntry->NextEntryOffset;
```
m417z: What about long paths? I assume it will crash in this case, which is not good.

mods/hide-dotfiles-explorer.wh.cpp
danalec: ah I see now. much simpler!

mods/hide-dotfiles-explorer.wh.cpp
danalec: dynamic allocation with std::wstring
replaced the fixed MAX_PATH buffer

better?

mods/hide-dotfiles-explorer.wh.cpp
danalec: Wh_ functions are being used. Doesn't it require the windhawk_utils.h ?

mods/hide-dotfiles-explorer.wh.cpp
m417z: Yes. You could use `std::string_view` and avoid the allocation altogether, but at least it shouldn't crash now.

mods/hide-dotfiles-explorer.wh.cpp
m417z: No, Wh_ functions come from `windhawk_api.h` which is included automatically. See https://github.com/ramensoftware/windhawk/wiki/Development-tips#windhawk-utils for info about `windhawk_utils.h`.

mods/hide-dotfiles-explorer.wh.cpp
m417z: Your current settings are invalid, replace `[]` with `[""]` to specify that it's an array of strings.

mods/hide-dotfiles-explorer.wh.cpp
m417z: What I meant is that you can use `std::wstring_view` here:
```cpp
std::wstring fileName(currentEntry->FileName, fileNameLength);
```
And then pass it to `ShouldHideFile` as `std::wstring_view`.

danalec: ShouldHideFile to accept std::wstring_view instead of raw pointers maybe?

m417z: Yes, that's what I mean.

mods/hide-dotfiles-explorer.wh.cpp
```diff
@@ -0,0 +1,320 @@
...
+- **Exact matching**: Full filename must match exactly (no partial matches)
+
+### Settings
+- **Dotfile Whitelist**: List of dotfiles to show (e.g., `.gitignore`, `.env`)
+- **Always Hide**: List of filenames to always hide, even if they don't start with a dot (e.g., `desktop.ini`, `Thumbs.db`)
+
+Examples:
+- To show `.gitignore`: add `.gitignore` to dotfile whitelist
+- To hide `desktop.ini`: add `desktop.ini` to always-hide list
+- `.env.local` will not be shown if only `.env` is whitelisted
```
m417z: Optional suggestion: How about allowing patterns such as `.env.*`? You can use `strmatch`:
https://github.com/ramensoftware/windhawk-mods/blob/a1e5471c16c55ce3827e8f85367aafa200d0e69a/mods/icon-resource-redirect.wh.cpp#L381-L422

It should also be easy to make it case-insensitive, just add `tolower` to `str[0] != pat[0]`.

mods/hide-dotfiles-explorer.wh.cpp
```diff
@@ -0,0 +1,320 @@
...
+
+## Important Notes
+- **No file attributes are modified** - hidden files retain their original attributes
```
m417z, line 58: Optional suggestion: How about adding an option to choose one of the three modes:
* Never show files (current behavior)
* Show files as hidden (add `FILE_ATTRIBUTE_HIDDEN` attribute, Dot Hide behavior)
* Show files as system (add `FILE_ATTRIBUTE_HIDDEN`, `FILE_ATTRIBUTE_SYSTEM` attributes)

I think that it shouldn't be difficult to add, and I believe some users wouldn't want to have the files removed from the listing completely.

For example, with a folder being listed with the system attribute, it won't be shown in most configurations, but it can still be opened from the address bar or Win+R. With the current behavior, there's no way to access dot folders.

What do you think?

m417z (commented):
Looks good. Here are two optional suggestions.

mods/hide-dotfiles-explorer.wh.cpp
danalec: modifying the file attributes?

danalec: ah, do you mean like 'ls -a' ? I see, understood

m417z: Returning different attributes in the listing, not modifying the actual attributes on disk. That's what Dot Hide does:
https://github.com/ramensoftware/windhawk-mods/blob/3d38cbfa6b1095206cca01f4c2aa10fe54348205/mods/dot-hide.wh.cpp#L40-L46

mods/hide-dotfiles-explorer.wh.cpp
```diff
@@ -172,28 +169,19 @@ bool ShouldHideFile(std::wstring_view fileName) noexcept {
...
     }
     
-    auto caseInsensitiveCompare = [](std::wstring_view lhs, std::wstring_view rhs) noexcept -> bool {
-        return lhs.size() == rhs.size() && 
-               std::ranges::equal(lhs, rhs, [](wchar_t a, wchar_t b) noexcept {
-                   return ::towlower(a) == ::towlower(b);
-               });
+    auto matchesPattern = [](std::wstring_view name, const std::vector<std::wstring>& patterns) noexcept -> bool {
+        return std::ranges::any_of(patterns, [name](const std::wstring& pattern) {
+            return PathMatchSpecW(name.data(), pattern.c_str()) != FALSE;
```
m417z: Using `name.data()` is incorrect since `name` might not be null-terminated.

mods/hide-dotfiles-explorer.wh.cpp
```diff
@@ -171,7 +171,8 @@ bool ShouldHideFile(std::wstring_view fileName) noexcept {
     
     auto matchesPattern = [](std::wstring_view name, const std::vector<std::wstring>& patterns) noexcept -> bool {
         return std::ranges::any_of(patterns, [name](const std::wstring& pattern) {
-            return PathMatchSpecW(name.data(), pattern.c_str()) != FALSE;
+            std::wstring temp(name);
```
m417z: That works, but it'd be more efficient to create a `std::wstring` only once, and not for each pattern.

danalec:
> Thank you for the patience and the cooperation!

I am the one who thanks you
I learned a lot!

PR #2210, opened 2025-07-25
Hide Network UNC Paths in Explorer

danalec:
This mod removes UNC paths from network drives in Explorer and dialogs.
It declutters lengthy network drive names in Explorer.

-added [aubymori](https://github.com/aubymori) and [m417z](https://github.com/m417z) suggestions
-reorganized branches for easier multiple mods submission (beg pardon for closing #2202)


mods/hide-unc.wh.cpp
```diff
@@ -0,0 +1,346 @@
...
+            GetClassNameW(hWnd, className, 256);
+        } else {
+            wcscpy_s(className, 256, lpClassName);
+        }
+        
+        if (wcscmp(className, L"ComboBox") == 0 && hWndParent) {
+            WCHAR parentClass[256];
+            GetClassNameW(hWndParent, parentClass, 256);
+            if (wcsstr(parentClass, L"#32770")) {
+                WNDPROC oldProc = (WNDPROC)SetWindowLongPtr(hWnd, GWLP_WNDPROC, (LONG_PTR)ComboBoxSubclassProc);
```
m417z: This also isn't being restored, right? I suggest using `WindhawkUtils::SetWindowSubclassFromAnyThread`, remember which windows are subclassed, and remove the subclass on uninit. You can look at existing mods for examples.

mods/hide-unc.wh.cpp
```diff
@@ -0,0 +1,398 @@
...
+            WindhawkUtils::RemoveWindowSubclassFromAnyThread(hWnd, ComboBoxSubclassProc);
+        }
+        g_subclassedWindows.clear();
+    }
+    
+    if (g_GetDisplayNameHooked && GetDisplayName_Original) {
+        if (g_pShellItem) {
+            void*** vtbl = (void***)g_pShellItem;
+            if (vtbl && *vtbl) {
+                Wh_SetFunctionHook((*vtbl)[5], (void*)GetDisplayName_Original, NULL);
```
m417z: This is not correct and not necessary. It was fine without this code, hooks are automatically removed before `Wh_ModUninit` is called. See:
https://github.com/ramensoftware/windhawk/wiki/Mod-lifetime

mods/hide-unc.wh.cpp
```diff
@@ -0,0 +1,398 @@
...
+    Wh_SetFunctionHook((void*)SetWindowTextW, (void*)SetWindowTextW_Hook, (void**)&SetWindowTextW_Original);
+    
+    return TRUE;
+}
+
+void Wh_ModUninit() {
+    {
+        std::lock_guard<std::mutex> lock(g_subclassedWindowsMutex);
+        for (HWND hWnd : g_subclassedWindows) {
+            WindhawkUtils::RemoveWindowSubclassFromAnyThread(hWnd, ComboBoxSubclassProc);
```
m417z: Calling this while holding the lock is dangerous and can cause a deadlock. A possible solution is to grab the data, release the lock, and then unsubclass, similar to:
https://github.com/ramensoftware/windhawk-mods/blob/dfdd41d6b0c84e75c65896510c3ae1a38252647e/mods/icon-resource-redirect.wh.cpp#L571

mods/hide-unc.wh.cpp
```diff
@@ -0,0 +1,398 @@
...
+#include <commctrl.h>
+#include <commdlg.h>
+#include <shlobj.h>
+#include <shobjidl.h>
+#include <shlwapi.h>
+#include <vector>
+#include <algorithm>
+#include <mutex>
+
+#include "windhawk_utils.h"
```
m417z:
```suggestion
#include <windhawk_utils.h>
```

mods/hide-unc.wh.cpp
```diff
@@ -0,0 +1,378 @@
...
+// @include         explorer.exe
+// @include         *
+// @exclude         csrss.exe
+// @exclude         dwm.exe
+// @exclude         winlogon.exe
+// @exclude         services.exe
+// @exclude         lsass.exe
+// @exclude         nvcplui.exe
+// @exclude         nvcontainer.exe
+// @architecture    x86-64
```
m417z: Since you're targeting all processes, I assume it shouldn't be limited to x64, right?

danalec: > Since you're targeting all processes, I assume it shouldn't be limited to x64, right?

I wasn't sure if I would run into any problems using the COM vtable on a different architecture.
Also I wasn't sure if void*** vtbl = (void***)*ppv;) would work either.
After researching, I think there is no such problem. Indeed, there is no need of line 17.

mods/hide-unc.wh.cpp
```diff
@@ -0,0 +1,382 @@
...
+// @github          https://github.com/danalec
+// @include         explorer.exe
+// @include         *
+// @exclude         csrss.exe
+// @exclude         dwm.exe
+// @exclude         winlogon.exe
+// @exclude         services.exe
+// @exclude         lsass.exe
+// @exclude         nvcplui.exe
+// @exclude         nvcontainer.ex
```
m417z: Looks like you lost the last letter here.

PR #2212, opened 2025-07-26
Create quicklaunch-3dbuttons.wh.cpp

Anixx:
3D buttons in QuickLaunch

### DIFF
--- a/mods/quicklaunch-3dbuttons.wh.cpp
+++ b/mods/quicklaunch-3dbuttons.wh.cpp
@@ -1,0 +1,56 @@
+// ==WindhawkMod==
+// @id              quicklaunch-3dbuttons
+// @name            3D buttons on Quick Launch bar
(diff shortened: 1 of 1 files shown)


mods/quicklaunch-3dbuttons.wh.cpp
```diff
@@ -0,0 +1,56 @@
+// ==WindhawkMod==
+// @id              quicklaunch-3dbuttons
+// @name            3D buttons on Quick Launch bar
+// @description     Makes the buttons on the Quick Launch bar 3-dimensional
+// @version         1.0.0
+// @author          anixx
+// @github          https://github.com/Anixx
+// @include         *
```
m417z: If the mod only customizes the taskbar, why target all processes and not only explorer.exe?

Anixx:
Oh, yes.

PR #2216, opened 2025-07-27
Hide pin v1.0

### LINKED ISSUE #1215: Disable pin icon in folders
Hi, is it possible to introduce the ability to hide pin icons on windhawk?

Senza titolo

Thanks

danalec:
Removes pin icons from File Explorer's navigation pane by intercepting the tree view's item state queries.

Resolves #1215


mods/hide-pin.wh.cpp
```diff
@@ -0,0 +1,125 @@
...
+    return CallWindowProc(g_originalTreeWndProc, hWnd, uMsg, wParam, lParam);
+}
+
+typedef HWND (WINAPI *CreateWindowExW_t)(DWORD dwExStyle, LPCWSTR lpClassName, LPCWSTR lpWindowName, DWORD dwStyle, int X, int Y, int nWidth, int nHeight, HWND hWndParent, HMENU hMenu, HINSTANCE hInstance, LPVOID lpParam);
+CreateWindowExW_t CreateWindowExW_Original;
+
+HWND WINAPI CreateWindowExW_Hook(DWORD dwExStyle, LPCWSTR lpClassName, LPCWSTR lpWindowName, DWORD dwStyle, int X, int Y, int nWidth, int nHeight, HWND hWndParent, HMENU hMenu, HINSTANCE hInstance, LPVOID lpParam) {
+    HWND hWnd = CreateWindowExW_Original(dwExStyle, lpClassName, lpWindowName, dwStyle, X, Y, nWidth, nHeight, hWndParent, hMenu, hInstance, lpParam);
+    
+    if (hWnd && lpClassName && !IsBadReadPtr(lpClassName, sizeof(WCHAR))) {
```
m417z: Don't use `IsBadReadPtr`: [IsBadXxxPtr should really be called CrashProgramRandomly](https://devblogs.microsoft.com/oldnewthing/20060927-07/?p=29563).

mods/hide-pin.wh.cpp
```diff
@@ -0,0 +1,125 @@
...
+        if (wcsicmp(lpClassName, WC_TREEVIEW) == 0 || 
+            wcsicmp(lpClassName, L"SysTreeView32") == 0) {
+            wchar_t parentClass[256];
+            HWND hParent = GetParent(hWnd);
+            if (hParent && GetClassNameW(hParent, parentClass, 256)) {
+                if (wcsstr(parentClass, L"SHELLDLL_DefView") || 
+                    wcsstr(parentClass, L"NamespaceTreeControl")) {
+                    g_originalTreeWndProc = (WNDPROC)SetWindowLongPtr(hWnd, GWLP_WNDPROC, (LONG_PTR)TreeViewSubclassProc);
```
m417z: This isn't being reverted when the mod is unloaded. Like with the other mod, you can use `WindhawkUtils::SetWindowSubclassFromAnyThread`.

mods/hide-pin.wh.cpp
```diff
@@ -0,0 +1,125 @@
...
+// @version         1.0
+// @author          @danalec
+// @github          https://github.com/danalec
+// @include         explorer.exe
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+Removes pin icons from File Explorer's navigation pane by intercepting the tree view's item state queries.
+*/
```
m417z, line 16: How about adding a screenshot to show what exactly the mod does?

danalec: I'm trying to figure out how to do that. Should I upload it somewhere?

m417z: You can use imgur.com or GitHub, see the documentation:
https://github.com/ramensoftware/windhawk/wiki/Creating-a-new-mod#readme

mods/hide-pin.wh.cpp
```diff
@@ -82,7 +81,49 @@ LRESULT WINAPI DefWindowProcW_Hook(HWND hWnd, UINT Msg, WPARAM wParam, LPARAM lP
...
+    return hWnd;
+}
+
 BOOL Wh_ModInit() {
+    HMODULE hComctl32 = LoadLibrary(L"comctl32.dll");
+    if (hComctl32) {
+        pSetWindowSubclass = (SetWindowSubclass_t)GetProcAddress(hComctl32, "SetWindowSubclass");
+        pRemoveWindowSubclass = (RemoveWindowSubclass_t)GetProcAddress(hComctl32, "RemoveWindowSubclass");
+        pDefSubclassProc = (DefSubclassProc_t)GetProcAddress(hComctl32, "DefSubclassProc");
+    }
```
m417z, line 127: Why is this needed?

mods/hide-pin.wh.cpp
```diff
@@ -82,7 +81,49 @@ LRESULT WINAPI DefWindowProcW_Hook(HWND hWnd, UINT Msg, WPARAM wParam, LPARAM lP
...
+    
+    if (hWnd && lpClassName && HIWORD(lpClassName)) {
+        if (wcsicmp(lpClassName, WC_TREEVIEW) == 0 || 
+            wcsicmp(lpClassName, L"SysTreeView32") == 0) {
+            wchar_t parentClass[256];
+            HWND hParent = GetParent(hWnd);
+            if (hParent && GetClassNameW(hParent, parentClass, 256)) {
+                if (wcsstr(parentClass, L"SHELLDLL_DefView") || 
+                    wcsstr(parentClass, L"NamespaceTreeControl")) {
+                    WindhawkUtils::SetWindowSubclassFromAnyThread(hWnd, TreeViewSubclassProc, 0);
```
m417z: The subclass is not being removed on mod unload.

danalec: the subclasses are being removed on mod unload the Wh_ModUninit() function now.

for each window that still exists (IsWindow(window.hWnd)), it restores the original window procedure using SetWindowLongPtr(window.hWnd, GWLP_WNDPROC, (LONG_PTR)window.originalProc)
then clears the tracking vector and deletes the critical section.

this effectively removes all the subclasses by restoring the original window procedures that were saved when the windows were first subclassed in CreateWindowExW_Hook

it undoes the subclassing that was done with SetWindowLongPtr during creation.

mods/hide-pin.wh.cpp
```diff
@@ -0,0 +1,160 @@
...
+    
+    if (hWnd && lpClassName && HIWORD(lpClassName)) {
+        if (wcsicmp(lpClassName, WC_TREEVIEW) == 0 || 
+            wcsicmp(lpClassName, L"SysTreeView32") == 0) {
+            wchar_t parentClass[256];
+            HWND hParent = GetParent(hWnd);
+            if (hParent && GetClassNameW(hParent, parentClass, 256)) {
+                if (wcsstr(parentClass, L"SHELLDLL_DefView") || 
+                    wcsstr(parentClass, L"NamespaceTreeControl")) {
+                    WNDPROC originalProc = (WNDPROC)SetWindowLongPtr(hWnd, GWLP_WNDPROC, (LONG_PTR)TreeViewSubclassProc);
```
m417z: Is there a reason you decided keep `SetWindowLongPtr` instead of using `WindhawkUtils::SetWindowSubclassFromAnyThread` like in your other mod which was merged?

The problem with `SetWindowLongPtr` is that if more than one mod/tool subclasses the same window, removing the subclass can cause problems or crashes.

danalec: I could not be able to make SetWindowSubclassFromAnyThread work properly.

I'll try again and show the actual code even broken.

m417z: Yes, please post the code that doesn't work. Perhaps there's a mistake in the code, or perhaps there's a problem with the subclass function, in which case I'd like to understand and fix it.

mods/hide-pin.wh.cpp
m417z: Looks like you restored that code, is it the final code to review? I don't understand the purpose of this. 

Also, you don't seem to be calling `SetWindowSubclassFromAnyThread` in `Wh_ModUninit`.

danalec:
I've updated by mistake a new version that I'm working to remove the pins over the "quick access".

I went back because the previous working version is ready to go and I'll take time to investigate the "quick access" pins. And they don't look removable...

CyprinusCarpio:
I suggest you take a look at the 'Classic Explorer Treeview' mod. It removes pin icons in a simpler way.

danalec:
I've finally found the pin, it is Microsoft.UI.Xaml.Controls.FontIcon - PinnedFillGlyphFontIcon

I've used [UWPSpy](https://github.com/m417z/UWPSpy/releases/) from @m417z m417z

PR #2231, opened 2025-07-31
Update UXTheme hook

rounk-ctrl:
- fix arm64 support
- remove buggy DUI control panel header fix

### DIFF
--- a/mods/uxtheme-hook.wh.cpp
+++ b/mods/uxtheme-hook.wh.cpp
@@ -2,7 +2,7 @@
 // @id              uxtheme-hook
 // @name            UXTheme hook
 // @description     Allows you to apply custom themes
-// @version         1.3
+// @version         1.4
 // @author          rounk-ctrl
 // @github          https://github.com/rounk-ctrl
 // @include         winlogon.exe
@@ -10,7 +10,7 @@
 // @include         systemsettings.exe
 // @include         logonui.exe
 // @include         rundll32.exe
-// @compilerOptions -luxtheme -lgdi32 -lole32
+// @compilerOptions -lole32
 // ==/WindhawkMod==
 
 // ==WindhawkModReadme==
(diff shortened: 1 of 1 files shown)


mods/uxtheme-hook.wh.cpp
```diff
@@ -42,35 +43,26 @@ the colors in control panel might be weird sometimes. works fine if you leave co
...
 #define STDCALL  __cdecl
 #define SSTDCALL L"__cdecl"
 #else
-#define STDCALL  __stdcall
-#define SSTDCALL L"__stdcall"
+#define STDCALL  __thiscall
+#define SSTDCALL L"__thiscall"
 #endif
-typedef unsigned __int64 QWORD;
+typedef unsigned long QWORD;
```
m417z: That doesn't look right, but it also seems that `QWORD` is no longer used, so I guess you can just remove it.

mods/uxtheme-hook.wh.cpp
```diff
@@ -179,43 +132,28 @@ BOOL Wh_ModInit() {
...
     }};
 
-    std::vector<LPCWSTR> files{L"uxtheme.dll", L"uxinit.dll", L"themeui.dll"};
+    LPCWSTR files[] = {L"uxtheme.dll", L"uxinit.dll", L"themeui.dll"};
 
-    for (LPCWSTR dll : files)
+    for (int i = 0; i < int(ARRAYSIZE(files)); ++i)
     {
-        HMODULE hMod = LoadLibrary(dll);
+        HMODULE hMod = LoadLibrary(files[i]);
```
m417z: It'd be better to use `LoadLibraryEx` with `LOAD_LIBRARY_SEARCH_SYSTEM32` here, see https://github.com/ramensoftware/windhawk-mods/issues/2063.

Issue #1618, opened 2025-03-03
[Aerexplorer] Can this pop-up possibly be done away with? :-(


TheMouseAvenger:
This infernal, constant pop-up is always getting in the way of what I'm trying to do, & it's driving me absolutely mad!  Is there a way to turn it off, or can something be done to make that possible? :-(

![Image](https://github.com/user-attachments/assets/4f946bdc-1f38-46ba-8111-b462e9c0806a)

m417z:
It was discussed recently [here](https://github.com/ramensoftware/windhawk/discussions/421#discussioncomment-12362603):

> The only way to keep it from showing is to increase the delay after which it's shown during the long operation. The delay can be set to be very long, but I wouldn't recommend it. The window popping up should be rare unless something is wrong, in which case I'd rather fix the problem.

In your case, it seems to be related to aerexplorer, which is busy initializing. For how long is the popup open?

@aubymori any idea why would it take so long for the mod to initialize? By default, the timeout for the popup to appear is 3 seconds.

m417z:
Can you also check if there are any crash reports during that time?
See [How to find crash report from Windows Event Viewer](https://forums.flightsimulator.com/t/how-to-find-crash-report-from-windows-event-viewer/591676).

aubymori:
> [@aubymori](https://github.com/aubymori) any idea why would it take so long for the mod to initialize? By default, the timeout for the popup to appear is 3 seconds.

My only guess is the large amount of hooks it installs.

m417z:
Once hooks are cached, setting them should be very fast, unless something is wrong with the caching.

@TheMouseAvenger another question - do you have a way to trigger the popup? e.g. by running some program? Maybe cmd, since I see conhost.exe on the screenshot. This can help with the troubleshooting.

TheMouseAvenger:
> Once hooks are cached, setting them should be very fast, unless something is wrong with the caching.
> 
> [@TheMouseAvenger](https://github.com/TheMouseAvenger) another question - do you have a way to trigger the popup? e.g. by running some program? Maybe cmd, since I see conhost.exe on the screenshot. This can help with the troubleshooting.

I'll see what I can do!

TheMouseAvenger:
> Can you also check if there are any crash reports during that time? See [How to find crash report from Windows Event Viewer](https://forums.flightsimulator.com/t/how-to-find-crash-report-from-windows-event-viewer/591676).

Is this what you had in mind?

[...]

PR #2233, opened 2025-07-31
Aerexplorer 1.8.0

aubymori:
- Fixes for classic theme
  - NOTE: Arrows not showing on the navigation pane is a bug in comctl32.dll that will probably be fixed in a separate mod at another time.
- Use `LoadLibraryExW` with `LOAD_LIBRARY_SEARCH_SYSTEM32` (#2063)
- Option for "Views" dropdown on left like Windows Vista
- Option for icons on all command bar items like Windows Vista
- Reject ExplorerFrame.dll with a build number higher than 21332 (#1618)

### DIFF
--- a/mods/aerexplorer.wh.cpp
+++ b/mods/aerexplorer.wh.cpp
@@ -92,7 +92,7 @@ Feel free to try it on other versions, but it may not work.
   $description: Dropdown icon width in address bar. Address toolbar sizing must be set to "Custom".
 - refreshwidth: 25
   $name: Refresh icon width
-  $description: Refresh icon width in address bar. Address toolbar sizing must be set to "Custom".
+  $description: Refresh icon widFth in address bar. Address toolbar sizing must be set to "Custom".
 - oldsearch: true
   $name: Old search box
   $description: Disable the EdgeHTML-based search box and use the old one instead. You must enable this option to fix the bug where the placeholder disappears.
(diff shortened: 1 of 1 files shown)


mods/aerexplorer.wh.cpp
```diff
@@ -3021,6 +3144,13 @@ BOOL Wh_ModInit(void)
     LOAD_MODULE(ExplorerFrame)
     HOOK_SYMBOLS(ExplorerFrame, explorerframeDllHooks)
 
+    VS_FIXEDFILEINFO *pVerInfo = GetModuleVersionInfo(ExplorerFrame, nullptr);
+    if (HIWORD(pVerInfo->dwFileVersionLS) > 21332)
```
m417z: How about checking `pVerInfo` to be safe?

mods/aerexplorer.wh.cpp
```diff
@@ -92,7 +92,7 @@ Feel free to try it on other versions, but it may not work.
   $description: Dropdown icon width in address bar. Address toolbar sizing must be set to "Custom".
 - refreshwidth: 25
   $name: Refresh icon width
-  $description: Refresh icon width in address bar. Address toolbar sizing must be set to "Custom".
+  $description: Refresh icon widFth in address bar. Address toolbar sizing must be set to "Custom".
```
m417z: Accidental typo

mods/aerexplorer.wh.cpp
```diff
@@ -3021,6 +3144,13 @@ BOOL Wh_ModInit(void)
     LOAD_MODULE(ExplorerFrame)
     HOOK_SYMBOLS(ExplorerFrame, explorerframeDllHooks)
 
+    VS_FIXEDFILEINFO *pVerInfo = GetModuleVersionInfo(ExplorerFrame, nullptr);
```
m417z, line 3147: Just after merging, I noticed that the check should have been done before `HOOK_SYMBOLS`. Currently, it doesn't help as `Wh_SetFunctionHook` will still be called and fail.

aubymori: Went ahead and opened #2237 which fixes this along with another minor issue.

PR #2247, opened 2025-08-03
Translucent Windows v1.6

Undisputed00x:
* Introduce custom windows theme parts using Direct2D API
* Fix window flickering

### DIFF
--- a/mods/translucent-windows.wh.cpp
+++ b/mods/translucent-windows.wh.cpp
@@ -2,26 +2,29 @@
 // @id              translucent-windows
 // @name            Translucent Windows
 // @description     Enables native translucent effects in Windows 11
-// @version         1.5.1
+// @version         1.6.0
 // @author          Undisputed00x
 // @github          https://github.com/Undisputed00x
 // @include         *
-// @compilerOptions -ldwmapi -luxtheme -lcomctl32 -lgdi32
+// @compilerOptions -ldwmapi -luxtheme -lcomctl32 -lgdi32 -ld2d1 -lmsimg32
 // ==/WindhawkMod==
 
 // ==WindhawkModReadme==
 /*
 
 ### Blur (AccentBlurBehind)
-![AccentBlurBehind](https://i.imgur.com/S8OGRvc.png)
+![AccentBlurBehind](https://i.imgur.com/tSf5ztk.png)
 ### Acrylic (SystemBackdrop)
-![Acrylic SystemBackdrop](https://i.imgur.com/66cAiEC.png)
+![Acrylic SystemBackdrop](https://i.imgur.com/YNktLTu.png)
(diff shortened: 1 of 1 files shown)


mods/translucent-windows.wh.cpp
```diff
@@ -859,12 +1053,12 @@ VOID STDCALL Element_PaintBgHook(class Element* This, HDC hdc, class Value* valu
         // 4-> cpanel top bar and side bar (white image)
         // 1-> some new cp page style (cp_hub_frame)
         if (v45==4)
-        { 
+        {
             HWND wnd = WindowFromDC(hdc);
             HTHEME hTh = OpenThemeData(wnd, L"ControlPanel");
             COLORREF clrBg;
             GetThemeColor(hTh, 2, 0, TMT_FILLCOLOR, &clrBg);
```
m417z: So now this call is unnecessary, right?

mods/translucent-windows.wh.cpp
```diff
@@ -1629,43 +4668,40 @@ void NewWindowShown(HWND hWnd)
...
-    
 }
 
-void DwmExpandFrameIntoClientAreaHook()
+VOID DwmExpandFrameIntoClientAreaHook()
 {
-    Wh_SetFunctionHook((void*)GetProcAddress(GetModuleHandle(L"dwmapi.dll"), "DwmExtendFrameIntoClientArea"),
-                       (void*)HookedDwmExtendFrameIntoClientArea,
-                       (void**)&DwmExtendFrameIntoClientArea_orig);
+    WindhawkUtils::SetFunctionHook((VOID*)DwmExtendFrameIntoClientArea, (VOID*)HookedDwmExtendFrameIntoClientArea, (VOID**)&DwmExtendFrameIntoClientArea_orig);
```
m417z: The idea of `WindhawkUtils::SetFunctionHook` is that you can remove the `void*` hooks, and the compiler makes sure that the signature of the hook is compatible with the signature of the hooked function. See https://github.com/ramensoftware/windhawk/wiki/Development-tips#hooks-and-calling-conventions.

Undisputed00x: Thanks, it actually found a bug in the HookedDrawThemeTextEx args definitions type

Undisputed00x:
@Acercandr0 Are you using a higher screen scale than the default? A user from the Windhawk Discord channel reported that this is happening because of this.

Undisputed00x:
I'll upload an fix pr shortly, theme parts are cached so you'll see the difference once you unload/load the mod.

Undisputed00x:
@olufemisong Can you close all programs running in the background and check if the white window appears again? I assume there is an overlay running in the background. Also, can you post the mod logs to see which programs are affected?

PR #2260, opened 2025-08-05
Create taskbar-icon-numberer.wh.cpp

jsfdez:
A mod that adds keyboard shortcut numbers (1-9, 0) to taskbar icons like 7+ Taskbar Numberer:

image

### DIFF
--- a/mods/taskbar-icon-numberer.wh.cpp
+++ b/mods/taskbar-icon-numberer.wh.cpp
@@ -8,17 +8,14 @@
 // @homepage        https://tightcorner.substack.com/p/reverse-engineering-windows-11s-taskbar
 // @include         explorer.exe
 // @architecture    x86-64
-// @compilerOptions -DWINVER=0x0A00 -lole32 -loleaut32 -lruntimeobject -Wextra -lcomctl32
+// @compilerOptions -DWINVER=0x0A00 -lole32 -loleaut32 -lruntimeobject
 // ==/WindhawkMod==
 
 // ==WindhawkModReadme==
 /*
 # Taskbar Icon Numberer for Windows 11
-
 Displays keyboard shortcut numbers (1-9, 0) as overlays on Windows 11 taskbar icons, replicating the functionality of the classic 7+ Taskbar Numberer tool.
-
 ![](https://raw.githubusercontent.com/jsfdez/images/main/taskbar-icon-numberer.wh.cpp.png)
-
 ## Features
 - Numbers 1-9 for first nine apps, 0 for tenth app
 - Fully customizable appearance (position, size, colors)
@@ -26,14 +23,11 @@ Displays keyboard shortcut numbers (1-9, 0) as overlays on Windows 11 taskbar ic
 - Supports light and dark themes
 - Hidden overlays for apps beyond position 10
 - Efficient overlay management with minimal performance impact
-
 ## Usage
 Numbers correspond to Windows + [number] keyboard shortcuts for quick app launching. Simply press Win+1 to open the first app, Win+2 for second, etc.
-
 ## Known Issues
 - Numbers don't appear until first taskbar interaction (hover over any button)
 - Occasional wrong numbers during initial display (self-corrects after reordering)
-
 ## Customization
 Configure number position (corners), font size (8-16px), text color, and stroke color through mod settings.
 */
@@ -78,7 +72,6 @@ Configure number position (corners), font size (8-16px), text color, and stroke 
 #include <atomic>
 #include <string>
 #include <unordered_set>
-#include <set>
 #include <mutex>
 #include <vector>
 #include <functional>
@@ -97,17 +90,11 @@ using RunFromWindowThreadProc_t = void(WINAPI*)(void* parameter);
 using TaskListButton_UpdateVisualStates_t = void(WINAPI*)(void* pThis);
 TaskListButton_UpdateVisualStates_t TaskListButton_UpdateVisualStates_Original;
 
-using TaskListButton_UpdateButtonPadding_t = void(WINAPI*)(void* pThis);
-TaskListButton_UpdateButtonPadding_t TaskListButton_UpdateButtonPadding_Original;
-
-using TaskListButton_UpdateBadgeSize_t = void(WINAPI*)(void* pThis);
-TaskListButton_UpdateBadgeSize_t TaskListButton_UpdateBadgeSize_Original;
-
 enum class NumberPosition {
     topLeft,
     topRight,
     bottomLeft,
-    bottomRight
+    bottomRight,
 };
 
 struct {
@@ -123,14 +110,30 @@ struct ButtonInformation {
     FrameworkElement overlay = nullptr;
     int currentNumber = -1;
     bool isVisible = false;
-    bool operator<(const ButtonInformation& other) const { return winrt::get_abi(button) < winrt::get_abi(other.button); }
+    
+    // Constructor for easier creation
+    explicit ButtonInformation(const FrameworkElement& button = nullptr) : button(button) {}
+    
+    bool isSameButton(const FrameworkElement& otherButton) const {
+        auto thisButton = button.get();
+        if (!thisButton) return false;
+        return winrt::get_abi(thisButton.as<winrt::Windows::Foundation::IUnknown>()) == winrt::get_abi(otherButton.as<winrt::Windows::Foundation::IUnknown>());
+    }
 };
 
-std::atomic<bool> g_taskbarViewDllLoaded;
-std::atomic<bool> g_unloading;
-std::atomic<DWORD> g_initialTime;
+// Simplified globals
+std::atomic<bool> g_taskbarViewDllLoaded{false};
+std::atomic<bool> g_unloading{false};
 std::mutex g_overlayMutex;
-std::set<ButtonInformation> g_trackedButtons;
+std::vector<ButtonInformation> g_trackedButtons;
+
+// Helper function to find button in vector
+auto FindButtonInVector = [](const std::vector<ButtonInformation>& buttons, const FrameworkElement& targetButton) {
+    return std::find_if(buttons.begin(), buttons.end(), 
+        [&targetButton](const ButtonInformation& info) {
+            return info.isSameButton(targetButton);
+        });
+};
 
 // Utility functions
 void LogException(const char* functionName) {
@@ -287,24 +290,30 @@ bool RunFromWindowThread(HWND hWnd, RunFromWindowThreadProc_t proc, void* procPa
     return true;
 }
 
+// Simplified text creation helper
+TextBlock CreateTextBlock(const std::wstring& text, double fontSize, const winrt::Windows::UI::Color& color, const Thickness& margin = {}) {
+    TextBlock textBlock;
+    textBlock.Text(text);
+    textBlock.FontSize(fontSize);
+    textBlock.Margin(margin);
+    
+    auto brush = SolidColorBrush();
+    brush.Color(color);
+    textBlock.Foreground(brush);
+    
+    return textBlock;
+}
+
 FrameworkElement CreateNumberOverlay(int number) {
+    if (number < 1) {
+        Wh_Log(L"CreateNumberOverlay: Invalid number %d, skipping creation", number);
+        return nullptr;
+    }
+    
     try {
-        // Validate number range to prevent invalid numbers like 11
-        if (number < 1) {
-            Wh_Log(L"CreateNumberOverlay: Invalid number %d, skipping creation", number);
-            return nullptr;
-        }
-        
-        TextBlock numberText;
-        numberText.Text(number == 10 ? L"0" : std::to_wstring(number));
-        numberText.FontSize(g_settings.numberSize);
-        
-        Wh_Log(L"CreateNumberOverlay: Creating overlay with number %d", number);
-        
+        std::wstring text = (number == 10) ? L"0" : std::to_wstring(number);
         auto textColor = ParseHexColor(g_settings.numberColor);
-        auto brush = SolidColorBrush();
-        brush.Color(textColor);
-        numberText.Foreground(brush);
+        auto strokeColor = ParseHexColor(g_settings.backgroundColor);
         
         Grid textContainer;
         textContainer.Name(L"WindhawkNumberOverlay");
@@ -314,23 +323,17 @@ FrameworkElement CreateNumberOverlay(int number) {
             for (int dy = -1; dy <= 1; dy++) {
                 if (dx == 0 && dy == 0) continue;
                 
-                TextBlock strokeText;
-                strokeText.Text(number == 10 ? L"0" : std::to_wstring(number));
-                strokeText.FontSize(g_settings.numberSize);
-                strokeText.Margin(Thickness{static_cast<double>(dx), static_cast<double>(dy), 0, 0});
-                
-                auto strokeColor = ParseHexColor(g_settings.backgroundColor);
-                auto strokeBrush = SolidColorBrush();
-                strokeBrush.Color(strokeColor);
-                strokeText.Foreground(strokeBrush);
-                
+                auto strokeText = CreateTextBlock(text, g_settings.numberSize, strokeColor, 
+                    Thickness{static_cast<double>(dx), static_cast<double>(dy), 0, 0});
                 textContainer.Children().Append(strokeText);
             }
         }
         
+        // Main text
+        auto numberText = CreateTextBlock(text, g_settings.numberSize, textColor);
         textContainer.Children().Append(numberText);
         
-        // Position based on settings
+        // Set position and properties
         switch (g_settings.numberPosition) {
             case NumberPosition::topLeft:
                 textContainer.HorizontalAlignment(HorizontalAlignment::Left);
@@ -349,18 +352,43 @@ FrameworkElement CreateNumberOverlay(int number) {
                 textContainer.VerticalAlignment(VerticalAlignment::Bottom);
                 break;
         }
-        textContainer.Margin(Thickness{0, 0, 2, 2});
         
+        textContainer.Margin(Thickness{0, 0, 2, 2});
         Canvas::SetZIndex(textContainer, 1000);
+        
         if (number > 10) {
             textContainer.Visibility(Visibility::Collapsed);
         }
         
+        Wh_Log(L"CreateNumberOverlay: Created overlay with number %d", number);
         return textContainer;
         
     } catch (...) {
         LogException(__FUNCTION__);
         return nullptr;
+    }
+}
+
+// Simplified button cleanup
+void CleanupButtonOverlays(const ButtonInformation& buttonInfo) {
+    try {
+        auto button = buttonInfo.button.get();
+        if (!button) return;
+        
+        auto iconPanel = FindChildByName(button, L"IconPanel");
+        if (!iconPanel) return;
+        
+        auto panelChildren = iconPanel.as<Panel>().Children();
+        for (uint32_t j = 0; j < panelChildren.Size();) {
+            auto child = panelChildren.GetAt(j).try_as<FrameworkElement>();
+            if (child && child.Name() == L"WindhawkNumberOverlay") {
+                panelChildren.RemoveAt(j);
+            } else {
+                j++;
+            }
+        }
+    } catch (...) {
+        LogException(__FUNCTION__);
     }
 }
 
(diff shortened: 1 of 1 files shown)


mods/taskbar-icon-numberer.wh.cpp
```diff
@@ -0,0 +1,468 @@
...
+  - topLeft: Top left
+  - topRight: Top right
+  - bottomLeft: Bottom left
+  - bottomRight: Bottom right
+- numberSize: 10
+  $name: Number font size
+  $description: Size of the overlay numbers (8-16)
+- numberColor: "#FFFFFF"
+  $name: Number color
+  $description: Text color for the numbers (hex format #RRGGBB or #AARRGGBB)
```
m417z: The `#` symbol is used for a comment in YAML. Use quotes.

mods/taskbar-icon-numberer.wh.cpp
```diff
@@ -0,0 +1,468 @@
...
+- backgroundColor: "#80000000"
+  $name: Stroke/outline color
+  $description: Outline color around the numbers for better visibility (hex format #RRGGBB or #AARRGGBB)
```
m417z: The `#` symbol is used for a comment in YAML. Use quotes.

mods/taskbar-icon-numberer.wh.cpp
```diff
@@ -0,0 +1,468 @@
...
+- showOnlyRunning: true
+  $name: Show only on running apps
+  $description: Only show numbers on apps that are currently running
```
m417z: What's the purpose of this being `true`? Win+# keyboard shortcuts affect pinned items too.

mods/taskbar-icon-numberer.wh.cpp
```diff
@@ -0,0 +1,468 @@
...
+                color.G = static_cast<uint8_t>(std::wcstoul(hex.substr(5, 2).c_str(), nullptr, 16));
+                color.B = static_cast<uint8_t>(std::wcstoul(hex.substr(7, 2).c_str(), nullptr, 16));
+            } else { // #RRGGBB
+                color.A = 255;
+                color.R = static_cast<uint8_t>(std::wcstoul(hex.substr(1, 2).c_str(), nullptr, 16));
+                color.G = static_cast<uint8_t>(std::wcstoul(hex.substr(3, 2).c_str(), nullptr, 16));
+                color.B = static_cast<uint8_t>(std::wcstoul(hex.substr(5, 2).c_str(), nullptr, 16));
+            }
+        } catch (...) {
+            color.A = 255; color.R = 255; color.G = 255; color.B = 255;
```
m417z, line 167: I'm not a fan of using `catch (...)` in any function like you do. At least, have you considered logging the error? Something like:
https://github.com/ramensoftware/windhawk-mods/blob/4f32b2163d3b8938660000bb6cf62634e022635c/mods/windows-11-taskbar-styler.wh.cpp#L3646-L3648

mods/taskbar-icon-numberer.wh.cpp
```diff
@@ -0,0 +1,468 @@
...
+                textContainer.VerticalAlignment(VerticalAlignment::Bottom);
+                break;
+        }
+        textContainer.Margin(Thickness{0, 0, 2, 2});
+        
+        Canvas::SetZIndex(textContainer, 1000);
+        
+        std::lock_guard<std::mutex> lock(g_overlayMutex);
+        iconPanelElement.as<Panel>().Children().Append(textContainer);
+        g_numberedButtons.insert(buttonPtr);
```
m417z: You don't seem to be using `g_numberedButtons` anywhere. It looks like you planned on clearing the labels on mod unload, but didn't implement it.

mods/taskbar-icon-numberer.wh.cpp
```diff
@@ -0,0 +1,468 @@
...
+    try {
+        int childrenCount = Media::VisualTreeHelper::GetChildrenCount(element);
+        for (int i = 0; i < childrenCount; i++) {
+            auto child = Media::VisualTreeHelper::GetChild(element, i).try_as<FrameworkElement>();
+            if (child && child.Name() == name) {
+                return child;
+            }
+        }
+    } catch (...) {
+        // Ignore errors
```
m417z: Same as above for `catch (...)`. In this case, you can use `winrt::to_hresult` for logging:
https://github.com/ramensoftware/windhawk-mods/blob/4f32b2163d3b8938660000bb6cf62634e022635c/mods/taskbar-thumbnail-size.wh.cpp#L112-L115

mods/taskbar-icon-numberer.wh.cpp
```diff
@@ -0,0 +1,468 @@
...
+  $description: Text color for the numbers (hex format #RRGGBB or #AARRGGBB)
+- backgroundColor: "#80000000"
+  $name: Stroke/outline color
+  $description: Outline color around the numbers for better visibility (hex format #RRGGBB or #AARRGGBB)
+- showOnlyRunning: true
+  $name: Show only on running apps
+  $description: Only show numbers on apps that are currently running
+- maxNumbers: 10
+  $name: Maximum numbers to show
+  $description: Maximum number of taskbar items to number (1-10)
```
m417z: I'm still getting numbers after 10:

Image

mods/taskbar-icon-numberer.wh.cpp
```diff
@@ -0,0 +1,468 @@
...
+    try {
+        auto panel = taskbarRepeater.as<Panel>();
+        if (!panel) return;
+        
+        auto children = panel.Children();
+        
+        // Collect all TaskListButtons with their X positions
+        std::vector<std::pair<double, FrameworkElement>> buttons;
+        
+        for (uint32_t i = 0; i < children.Size(); i++) {
```
m417z, line 578: From my experience, enumerating `taskbarRepeater` don't work that well with `Children()` which returns some off-screen elements. It works better with `ItemsRepeater::TryGetElement`:
https://github.com/ramensoftware/windhawk-mods/blob/4f32b2163d3b8938660000bb6cf62634e022635c/mods/taskbar-labels.wh.cpp#L1429-L1439

mods/taskbar-icon-numberer.wh.cpp
```diff
@@ -0,0 +1,468 @@
...
+        TaskListButton_get_IsRunning_Original(
+            winrt::get_abi(taskListButtonElement.as<winrt::Windows::Foundation::IUnknown>()),
+            &isRunning);
+        return isRunning;
+    } catch (...) {
+        return false;
+    }
+}
+
+void CreateNumberOverlay(FrameworkElement taskListButtonElement, int number) {
```
m417z: I noticed that this function is called for each taskbar item on each update operation (e.g. mouse hover), which seems inefficient. What do you think about adding an optimization such that the overlay is updated only when necessary? You can get the current text and create a new element only if there's no label yet or if the number changed.

Even if the number changes, if the label already exists, you don't need to recreate it each time.

mods/taskbar-icon-numberer.wh.cpp
```diff
@@ -0,0 +1,468 @@
...
+        // Update all buttons at once to ensure consistent numbering
+        UpdateAllTaskbarNumbers(parent);
+        
+    } catch (...) {
+        // Ignore errors
+    }
+}
+
+bool HookTaskbarViewDllSymbols(HMODULE module) {
+    WindhawkUtils::SYMBOL_HOOK user32DllHooks[] = {
```
m417z: The variable name is incorrect. A correct example for this case:
https://github.com/ramensoftware/windhawk-mods/blob/4f32b2163d3b8938660000bb6cf62634e022635c/mods/taskbar-on-top.wh.cpp#L1789-L1790

mods/taskbar-icon-numberer.wh.cpp
```diff
@@ -0,0 +1,468 @@
...
+                }
+            }
+        }
+        
+    } catch (...) {
+        // Ignore errors
+    }
+}
+
+void WINAPI TaskListButton_UpdateVisualStates_Hook(void* pThis) {
```
m417z, line 650: Note that this function is called for secondary taskbars too, so numbers are added on all monitors. It makes more sense for the mod to only show numbers on the primary taskbar since that's what Win+# hotkeys affect.

m417z (commented):
Well done! I wanted to create a mod like this, but didn't have the time to work on it.

mods/taskbar-icon-numberer.wh.cpp
```diff
@@ -0,0 +1,468 @@
...
+Adds keyboard shortcut numbers (1-9, 0) to taskbar icons, similar to the classic 
+7+ Taskbar Numberer tool. Numbers appear as small overlays on the taskbar button icons.
+
+Features:
+- Shows numbers 1-9 for the first 9 taskbar items, then 0 for the 10th
+- Customizable number position, size, and colors
+- Works with Windows 11 taskbar
+- Supports both light and dark themes
+
+The numbers correspond to the Windows + [number] keyboard shortcuts for quick app launching.
```
m417z: How about adding a screenshot to the readme?

mods/taskbar-icon-numberer.wh.cpp
jsfdez: Yes, you are right; this is a leftover from an attempt to exclude pinned applications, but I was unable to make it work. Removing this

mods/taskbar-icon-numberer.wh.cpp
jsfdez: added!

mods/taskbar-icon-numberer.wh.cpp
jsfdez: Now I'm logging any exception in `LogException`

mods/taskbar-icon-numberer.wh.cpp
jsfdez: done in `LogException`

mods/taskbar-icon-numberer.wh.cpp
jsfdez: I'll try to optimize this

mods/taskbar-icon-numberer.wh.cpp
jsfdez: Extracted the `RemoveOverlay` function, and now it's called for each button during the mod unload.

mods/taskbar-icon-numberer.wh.cpp
jsfdez: I didn't have any issue with `Children` in this code, but took your function anyway :)

mods/taskbar-icon-numberer.wh.cpp
jsfdez: True, is there a way to check if this is the primary taskbar?

m417z: In one of the mods, I'm using this method, it isn't very elegant but it's been working correctly so far. 

https://github.com/ramensoftware/windhawk-mods/blob/54dde4e2305ee64046cb84ffb2520f83dfe26f4f/mods/taskbar-vertical.wh.cpp#L1363-L1381

mods/taskbar-icon-numberer.wh.cpp
m417z: Add the comment too.

mods/taskbar-icon-numberer.wh.cpp
```diff
@@ -0,0 +1,840 @@
...
+// @id              taskbar-icon-numberer
+// @name            Taskbar Icon Numberer for Windows 11
+// @description     Add keyboard shortcut numbers (1-9, 0) to taskbar icons like 7+ Taskbar Numberer
+// @version         0.0.6
+// @author          js
+// @github          https://github.com/jsfdez
+// @homepage        https://tightcorner.substack.com/p/reverse-engineering-windows-11s-taskbar
+// @include         explorer.exe
+// @architecture    x86-64
+// @compilerOptions -DWINVER=0x0A00 -lole32 -loleaut32 -lruntimeobject -Wextra -lcomctl32
```
m417z: Remove `-Wextra`.

mods/taskbar-icon-numberer.wh.cpp
```diff
@@ -0,0 +1,840 @@
...
+struct {
+    NumberPosition numberPosition;
+    int numberSize;
+    std::wstring numberColor;
+    std::wstring backgroundColor;
+} g_settings;
+
+// Track only primary taskbar buttons with their overlay state
+struct ButtonInformation {
+    void* pointer = nullptr;
```
m417z: If the object is freed, using the pointer might not be safe. Consider using `winrt::weak_ref`.

Because an object can be freed and then a new object can reuse the same address in memory, I think it's better to use `std::vector` here rather than `std::set`.

mods/taskbar-icon-numberer.wh.cpp
```diff
@@ -0,0 +1,840 @@
...
+        Wh_Log(L"UpdateButtonOverlay: Exception occurred for button %p", buttonPtr);
+        LogException(__FUNCTION__);
+    }
+}
+
+void UpdateAllTaskbarNumbers(FrameworkElement taskbarRepeater) {
+    if (g_unloading) return;
+    
+    // Skip processing for first 2 seconds after init to avoid startup issues
+    if (GetTickCount() - g_initialTime < 2000) return;
```
m417z: This doesn't look reliable. What kinds of issues?

mods/taskbar-icon-numberer.wh.cpp
```diff
@@ -0,0 +1,840 @@
...
+                                    auto child = panelChildren.GetAt(j).try_as<FrameworkElement>();
+                                    if (child && child.Name() == L"WindhawkNumberOverlay") {
+                                        panelChildren.RemoveAt(j);
+                                    } else {
+                                        j++;
+                                    }
+                                }
+                            }
+                        }
+                    } catch (...) {}
```
m417z: Log?

mods/taskbar-icon-numberer.wh.cpp
```diff
@@ -0,0 +1,840 @@
...
+    } catch (...) {
+        LogException(__FUNCTION__);
+    }
+}
+
+void WINAPI TaskListButton_UpdateButtonPadding_Hook(void* pThis) {
+    if (TaskListButton_UpdateButtonPadding_Original) {
+        TaskListButton_UpdateButtonPadding_Original(pThis);
+    }
+    TaskListButton_UpdateVisualStates_Hook(pThis);
```
m417z: This will call `TaskListButton_UpdateVisualStates_Original`, I'm not sure it's desirable here.

mods/taskbar-icon-numberer.wh.cpp
```diff
@@ -0,0 +1,840 @@
...
+}
+
+void WINAPI TaskListButton_UpdateBadgeSize_Hook(void* pThis) {
+    if (TaskListButton_UpdateBadgeSize_Original) {
+        TaskListButton_UpdateBadgeSize_Original(pThis);
+    }
+    TaskListButton_UpdateVisualStates_Hook(pThis);
```
m417z: This will call `TaskListButton_UpdateVisualStates_Original`, I'm not sure it's desirable here.

mods/taskbar-icon-numberer.wh.cpp
```diff
@@ -0,0 +1,840 @@
...
+}
+
+
+bool HookTaskbarViewDllSymbols(HMODULE module) {
+    WindhawkUtils::SYMBOL_HOOK symbolHooks[] = {
```
m417z, line 679: Please add back the comment with the target modules.

mods/taskbar-icon-numberer.wh.cpp
```diff
@@ -0,0 +1,840 @@
...
+        if (auto xamlRoot = taskListButtonElement.XamlRoot()) {
+            if (IsSecondaryTaskbar(xamlRoot)) return;
+        }
+
+        auto parent = Media::VisualTreeHelper::GetParent(taskListButtonElement).as<FrameworkElement>();
+        if (!parent || parent.Name() != L"TaskbarFrameRepeater") return;
+
+        if (g_unloading) {
+            std::lock_guard<std::mutex> lock(g_overlayMutex);
+            g_trackedButtons.clear();
```
m417z: I don't think you want to clear it here, it's already cleared in `RemoveAllNumberOverlays` (`Wh_ModBeforeUninit`), and if it's cleared here first, then the numbers won't be removed.

mods/taskbar-icon-numberer.wh.cpp
jsfdez: Changed to `winrt::weak_ref<FrameworkElement>`.

I'm using a `std::set` to avoid adding logic to find potential duplicates and to find the buttons more easily.

mods/taskbar-icon-numberer.wh.cpp
jsfdez: Leftover, removed

mods/taskbar-icon-numberer.wh.cpp
jsfdez: You're right, removed!

mods/taskbar-icon-numberer.wh.cpp
m417z: I understand, but consider the following:

* You add button1 to the set.
* button1 is freed, but it's still in the set.
* button2 is created, and it reuses the same pointer.
* You try to add button2 to the set, but this pointer is already in the set.

It might work if you're careful to check for freed items each time before adding new ones, but it can be error prone.

mods/taskbar-icon-numberer.wh.cpp
```diff
@@ -119,11 +119,11 @@ struct {
 
 // Track only primary taskbar buttons with their overlay state
 struct ButtonInformation {
-    void* pointer = nullptr;
+    winrt::weak_ref<FrameworkElement> button;
     FrameworkElement overlay = nullptr;
     int currentNumber = -1;
     bool isVisible = false;
-    bool operator<(const ButtonInformation& other) const { return pointer < other.pointer; }
+    bool operator<(const ButtonInformation& other) const { return winrt::get_abi(button) < winrt::get_abi(other.button); }
```
m417z: As far as I understand, `winrt::get_abi` here returns the pointer of `weak_ref`, not the pointer of the button itself, so each item will be unique and you can get duplicates.

What I'd do is use `std::vector` and check manually when inserting that the button isn't there yet.

mods/taskbar-icon-numberer.wh.cpp
```diff
@@ -370,11 +374,9 @@ void RemoveAllNumberOverlays() {
                 std::lock_guard<std::mutex> lock(g_overlayMutex);
                 for (const auto& buttonInfo : g_trackedButtons) {
                     try {
-                        winrt::Windows::Foundation::IUnknown buttonIUnknown;
-                        winrt::copy_from_abi(buttonIUnknown, buttonInfo.pointer);
-                        auto button = buttonIUnknown.try_as<FrameworkElement>();
+                        auto button = buttonInfo.button;
                         if (button) {
-                            auto iconPanel = FindChildByName(button, L"IconPanel");
+                            auto iconPanel = FindChildByName(button.get(), L"IconPanel");
```
m417z: `button.get()` can return null here. The way to use `weak_ref` is something like:
```cpp
if (auto button = buttonWeak.get()) {
  auto iconPanel = FindChildByName(button, L"IconPanel");
  // ...
}
```

mods/taskbar-icon-numberer.wh.cpp
```diff
@@ -404,11 +408,11 @@ void RemoveAllNumberOverlays() {
 
 void UpdateOverlayWithNewSettings(const ButtonInformation& buttonInformation) {
     if (!buttonInformation.overlay) {
-        Wh_Log(L"UpdateOverlayWithNewSettings: No overlay to update for button %p", buttonInformation.pointer);
+        Wh_Log(L"UpdateOverlayWithNewSettings: No overlay to update for button %p", &buttonInformation.button);
```
m417z: That's only a log, but I think the correct way to get the pointer here is `winrt::get_abi(buttonInformation.button.get())`.

mods/taskbar-icon-numberer.wh.cpp
```diff
@@ -471,69 +475,73 @@ void UpdateOverlayWithNewSettings(const ButtonInformation& buttonInformation) {
...
+            Wh_Log(L"UpdateButtonOverlay: No IconPanel found for button %p", buttonPointer);
             return;
         }
 
         std::lock_guard<std::mutex> lock(g_overlayMutex);
         
-        auto it = g_trackedButtons.find(ButtonInformation(buttonPtr));
+        winrt::Windows::Foundation::IUnknown buttonIUnknown;
+        winrt::copy_from_abi(buttonIUnknown, buttonPointer);
+        auto button = buttonIUnknown.try_as<FrameworkElement>();
```
m417z: You assume it'll work, so it's better to use `as` instead of `try_as` here.

m417z:
Is it ready to be merged?

rottakore:
It is already working somewhat fine, thanks! I'm already using it! :)

m417z:
Merged in https://github.com/ramensoftware/windhawk-mods/pull/3610.

isrbaral:
wow, I legit today decided to check if this existed yet for win11, thanks for merging, when can we expect it to be available in the mod explorer?

m417z:
@isrbaral it's here: https://windhawk.net/mods/taskbar-numberer

isrbaral:
wow, you're fast! ty! there appears to be another one just like it too...

m417z:
The other mod (https://windhawk.net/mods/numbered-taskbar) uses a different approach - an overlay managed by a separate process. This approach has several downsides due to the more loose integration, but it's less likely to break in future updates.

PR #2301, opened 2025-08-14
Dark mode message boxes 1.0

MGGSK:
Enables dark mode for message boxes.

### DIFF
--- a/mods/dark-message-boxes.wh.cpp
+++ b/mods/dark-message-boxes.wh.cpp
@@ -375,5 +375,6 @@ void Wh_ModUninit()
     DeleteObject(g_background);
     DeleteObject(g_btnBackground);
 
-    DeleteObject(g_hTextFont);
+    if(g_hTextFont)
+        DeleteObject(g_hTextFont);
 }


mods/dark-message-boxes.wh.cpp
```diff
@@ -0,0 +1,310 @@
...
+        return (LRESULT)g_barBackground;
+
+    case WM_PAINT:
+    {
+        PAINTSTRUCT ps;
+        HDC hdc = BeginPaint(hWnd, &ps);
+    
+        RECT rcBottomBar;
+        GetClientRect(hWnd, &rcBottomBar);
+        rcBottomBar.top = rcBottomBar.bottom - 45;
```
m417z: Should 45 be adjusted according to the window DPI? Also the font size.

mods/dark-message-boxes.wh.cpp
```diff
@@ -0,0 +1,310 @@
...
+    UpdateWindow(hWnd);
+}
+
+using SoftModalMessageBox_T = int(WINAPI*)(MSGBOXDATA*);
+SoftModalMessageBox_T SoftModalMessageBox;
+SoftModalMessageBox_T SoftModalMessageBox_Original;
+
+int WINAPI SoftModalMessageBox_hook(MSGBOXDATA* lpData)
+{
+    std::thread darkModeThread{SetDarkMode, lpData->lpszCaption, lpData->dwStyle};
```
m417z, line 259: Using a thread here has several potential problems:
- The thread might run too late, which will cause the message box to become dark with a delay. 
- If message box creation fails, or if it's closed too fast, the thread might become stuck in an infinite loop.
- What if there are several message boxes with the same title (e.g. "Error")?
- The thread loop may cause 100% CPU usage for some time, which isn't so good, especially for laptops.

Have you considered an alternative solution, such as hooking CreateWindow or a similar API?

marius-sucan:
the value of 45 and the font size 16, should be adjusted based on dpi.

MGGSK:
> the value of 45 and the font size 16, should be adjusted based on dpi.... it does not look correct on my system

I updated the mod to be dpi aware.

mods/dark-message-boxes.wh.cpp
```diff
@@ -0,0 +1,379 @@
...
+void Wh_ModUninit()
+{
+    FreeLibrary(g_hUser32);
+    FreeLibrary(g_hShlwApi);
+    FreeLibrary(g_hImageres);
+
+    DeleteObject(g_background);
+    DeleteObject(g_btnBackground);
+
+    DeleteObject(g_hTextFont);
```
m417z: Looks like `g_hTextFont` can be null.

m417z:
Is the mod ready for submission? Are you planning to look at the thread usage (https://github.com/ramensoftware/windhawk-mods/pull/2301#discussion_r2277606179)?

MGGSK:
> Is the mod ready for submission? Are you planning to look at the thread usage ([#2301 (comment)](https://github.com/ramensoftware/windhawk-mods/pull/2301#discussion_r2277606179))?

I am currently working on removing the thread usage.

Issue #2279, opened 2025-08-09
[Taskbar Clock Customization] New performance field formattings


OmiCron07:
After updating the mod, I noticed the new **System Performance Metrics** fields. While I appreciate the added functionality, I find the current display a bit distracting due to excessive movement in the text.

###  Suggestions for Improvement

- **CPU Usage**
  - Always display two digits, even for values below 10%  
    _Example: `05%` instead of `5%`_

- **Download/Upload Speeds**
  - Allow locking the unit to **Mb/s** for consistency (configurable)
  - Always show two or three digits  
    _Example: `120 Mb/s`, `008 Mb/s`_
  - Consider removing decimal places to reduce visual jitter (configurable)

These changes would help minimize shifting text and improve readability.

<img width="293" height="58" alt="Image" src="https://github.com/user-attachments/assets/98be62cf-2a46-4829-8a50-354d98d00b73" />

<img width="290" height="49" alt="Image" src="https://github.com/user-attachments/assets/aa2bbbec-4fd5-4de5-a869-471d6d423c99" />

Jerrk:
its a bit jarring 😂 

![Image](https://github.com/user-attachments/assets/e6efef17-8946-444e-abde-e2bab7a87997)

m417z:
Thanks for the suggestion, I agree that it can be improved. Note that I also got the feedback that the proper metrics are MB/s and KB/s (with capital B).

> * _Example: `120 Mb/s`, `008 Mb/s`_

120 MB/s is a huge amount for network activity. Perhaps it'd be better to have `01.2 MB/s`.

For the leading zeros, I noticed that two spaces have the exact same width as a digit with the default font, I think it'd be better for padding. In that example, `  1.2 MB/s`.

Unfortunately, `M` and `K` have different widths. I thought about using another symbol, for example monospace Unicode characters: 𝙼, 𝙺, but it doesn't look so good.

> * Always display two digits, even for values below 10%

In this case, 100% will still cause some movement.

Also, I noticed that a tab can be used, which aligns the text. It's quite limited, but perhaps can still be useful. For example:
`%upload_speed%<tab>%download_speed%<tab>...more_stuff...`
The tab character can be copied from Notepad, for example.

m417z:
[...]

Issue #2278, opened 2025-08-09
[Taskbar Clock Customization] Metrics stay at 0% or 0 B/s


gilnett:
The patterns for real-time system performance metrics are not working — they all stay at 0% or 0 B/s.

<img width="283" height="118" alt="Image" src="https://github.com/user-attachments/assets/a8e9fbec-6df9-4c45-bbef-60a5fdcf42cd" />

m417z:
Please attach debug logs by following the instructions on the Troubleshooting page:
https://github.com/ramensoftware/windhawk/wiki/Troubleshooting#some-or-all-windhawk-mods-dont-work

> * Disable the mod.
> * Go to the mod's Advanced tab.
> * Set **Debug logging** to **Detailed debug logs**.
> * Click on **Show log output**.
> * Enable the mod.
> * Verify that the mod doesn't work, copy the log and save it to a file.
> * Set **Debug logging** back to **None**.

m417z:
Thanks for the analysis @berrnd!

> probably [`PdhAddEnglishCounterA`](https://learn.microsoft.com/en-us/windows/win32/api/pdh/nf-pdh-pdhaddenglishcountera) might be the solution

That indeed seems to be the case:
http://www.clutteredlab.com/2024/10/dealing-with-localized-performance.html

Can you please try replacing `PdhAddCounter` with `PdhAddEnglishCounter` in the code? Does it fix it for you?

To try it, disable the existing mod, then click on "Fork", make the replacements in the code, click "Compile Mod", then "Exit Editing Mode".

gilnett:
here is my logs if you want:
```

22:41:44.011 7356 explorer.exe  [WH] [taskbar-clock-customization] [2648:GetLocalTime_Hook_Win11]: >
22:41:44.011 7356 explorer.exe  [WH] [taskbar-clock-customization] [2599:ThreadPoolTimer_CreateTimer_Hook]: > 10000000
22:41:44.011 7356 explorer.exe  [WH] [taskbar-clock-customization] [2648:GetLocalTime_Hook_Win11]: >
22:41:44.011 7356 explorer.exe  [WH] [taskbar-clock-customization] [1682:SampleData]: PdhCollectQueryData error 800007D5
22:41:44.011 7356 explorer.exe  [WH] [taskbar-clock-customization] [2231:ClockSystemTrayIconDataModel_GetTimeToolTipString2_Hook]: >
[...]

PR #2314, opened 2025-08-16
[Taskbar Clock Customization] Make network metrics unit, amount of decimals and unit display configurable

berrnd:
In reference to https://github.com/ramensoftware/windhawk-mods/commit/7068e6957690dc8bb1c1d1449041593fc34ff51e, #2279 and https://github.com/ramensoftware/windhawk-mods/issues/2278#issuecomment-3172680480:

The network metrics (upload/download transfer speed) are currently fixed to MB/s and a "constant width"-amount of decimal places.

For me personally MBit/s and always 2 decimal places and not showing the unit itself (since I know I want to see MBit/s) still makes more sense (probably due to that here in Germany that's the common unit in which internet speed is advertised). Of course this is up to very personal preference, so I added 3 new settings to make all that configurable:

grafik

I think the setting descriptions should be self-explanatory.

The setting defaults are the current state, so no changed behavior on update or for new users of this mod.

### DIFF
--- a/mods/taskbar-clock-customization.wh.cpp
+++ b/mods/taskbar-clock-customization.wh.cpp
@@ -156,6 +156,21 @@ styles, such as the font color and size.
   $description: >-
     The update interval, in seconds, of the system performance metrics such as
     CPU and RAM usage.
+- NetworkMetricsUnit: mbs
+  $name: Network metrics unit
+  $description: >-
+    The unit to use for displaying the upload/download transfer rate.
+  $options:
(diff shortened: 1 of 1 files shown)


m417z:
Thanks. As the failed check says, you can't submit updates to mods of other developers. 

Please post the pull request here:

berrnd:
Thanks, done there: https://github.com/m417z/my-windhawk-mods/pull/35

PR #2327, opened 2025-08-21
Create office-ui-reverter.wh.cpp

Amrsatrio:
Reverts Office 2022+/365 UI to Office 2016/2019. Tested on x64 and ARM64 versions of Microsoft 365.

image

### DIFF
--- a/mods/office-ui-reverter.wh.cpp
+++ b/mods/office-ui-reverter.wh.cpp
@@ -1,0 +1,586 @@
+// ==WindhawkMod==
+// @id              office-ui-reverter
+// @name            Office UI Reverter
+// @description     Reverts Office 2022+/365 UI to Office 2016/2019
+// @version         1.0.0
+// @author          Amrsatrio
+// @github          https://github.com/Amrsatrio
+// @twitter         https://twitter.com/amrsatrio
+// @include         WINWORD.EXE
+// @include         EXCEL.EXE
+// @include         POWERPNT.EXE
+// @include         OUTLOOK.EXE
+// @include         ONENOTE.EXE
+// @include         MSPUB.EXE
+// @include         MSACCESS.EXE
+// @architecture    x86-64
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Office UI Reverter
+
+Reverts the **Office 2022+/Microsoft 365 UI** back to the look of **Office 2019** or **Office 2016**.
+
+Some users prefer the older, less rounded, less visually distracting, and more compact Ribbon UI. This mod lets you force Office to render with those styles.
+
+## Features
+- Choose between:
+  - **Default** (no modifications)
+  - **Office 2019**
+  - **Office 2016**
+- Works with multiple Office apps (Word, Excel, PowerPoint, Outlook, OneNote, Publisher, Access).
+- UI style patching is permanent throughout the lifetime of the process; mod loading/unloading won't result in unexpected behavior and crashes.
+
+## ⚠️ Note
+- Older styles may lack icons for new features such as *Copilot*. Those are not covered by this mod.
+- Be advised that Microsoft can remove the older styles anytime in the future. When that happens, I do not plan on resurrecting the older styles.
+  - As long as Access still uses the Office 2019 style, there's no need to worry about it.
+
+## Technical Details
+- Hooks `LoadLibraryExW` to detect when `mso40uiWin32Client.dll` is loaded.
+- Locates and patches the internal `VisualVersion::GetVisualStyleForSurface()` function.
+- Replaces the return value with your selected style.
+
+## Compatibility
+- Works with **Office 2022+ perpetual** and **Microsoft 365 desktop apps**. Office 2019 and older are **not** supported.
+- Supports x64 and ARM64 builds. 32-bit x86 is **not** supported.
+- Tested on **Windows 11 23H2 and 24H2**, should also work on **Windows 10** with latest Office builds.
+*/
+// ==/WindhawkModReadme==
+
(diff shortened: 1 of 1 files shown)


mods/office-ui-reverter.wh.cpp
```diff
@@ -0,0 +1,592 @@
+// ==WindhawkMod==
+// @id              office-ui-reverter
+// @name            Office UI Reverter
+// @description     Reverts Office 2022+/365 UI to Office 2016/2019
+// @version         1.0.0
+// @author          Amrsatrio
+// @github          https://github.com/Amrsatrio
+// @twitter         https://twitter.com/amrsatrio
```
m417z, line 8: To verify your Twitter account, please send me (https://twitter.com/m417z) a direct message with the following content:

> I attest that I'm the sole owner of both this Twitter account (https://twitter.com/amrsatrio) and the following GitHub account: https://github.com/Amrsatrio

P.S. It's not that something looks suspicious, it's just that I commit to Windhawk users that all profiles are verified.

mods/office-ui-reverter.wh.cpp
```diff
@@ -0,0 +1,592 @@
...
+    if (dosHeader->e_magic == IMAGE_DOS_SIGNATURE)
+    {
+        PIMAGE_NT_HEADERS64 ntHeader = (PIMAGE_NT_HEADERS64)((BYTE*)dosHeader + dosHeader->e_lfanew);
+        if (ntHeader->Signature == IMAGE_NT_SIGNATURE)
+        {
+            PIMAGE_SECTION_HEADER firstSection = IMAGE_FIRST_SECTION(ntHeader);
+            for (unsigned int i = 0; i < ntHeader->FileHeader.NumberOfSections; ++i)
+            {
+                PIMAGE_SECTION_HEADER section = firstSection + i;
+                if (!strcmp((const char*)section->Name, pszSectionName))
```
m417z: It's not required for a section name to be null terminated. Consider using either `strncmp` or `memcmp`.

mods/office-ui-reverter.wh.cpp
```diff
@@ -0,0 +1,592 @@
...
+        for (const IMAGE_CHPE_RANGE_ENTRY& range : *chpeRanges)
+        {
+            ULONG typeMask = magic == IMAGE_NT_OPTIONAL_HDR32_MAGIC ? 1 : 3;
+            ULONG start = range.StartOffset & ~typeMask; // RVA
+            ULONG type = range.StartOffset & typeMask; // 0 = Arm64, 1 = Arm64EC, 2 = Amd64
+            if (type == 1) // Arm64EC
+            {
+                bIsARM64 = true;
+                pbSearchBegin = (PBYTE)hModule + RVAToFileOffset((PBYTE)hModule, start);
+                cbSearch = range.Length;
```
m417z, line 357: AFAIK there can be several ARM64EC ranges. Up to you if you want to change it to try all such ranges.

mods/office-ui-reverter.wh.cpp
```diff
@@ -0,0 +1,592 @@
...
+    Wh_Log(L">");
+
+    // Don't patch if we're not loaded during startup
+    // https://github.com/ramensoftware/windhawk/wiki/Development-tips#avoid-mod-loading-on-process-startup
+#ifdef _WIN64
+    const size_t OFFSET_SAME_TEB_FLAGS = 0x17EE;
+#else
+    const size_t OFFSET_SAME_TEB_FLAGS = 0x0FCA;
+#endif
+    bool bIsInitialThread = *(USHORT*)((BYTE*)NtCurrentTeb() + OFFSET_SAME_TEB_FLAGS) & 0x0400;
```
m417z: Is it really necessary? It means that if the injection is async (e.g. the creator process is excluded in Windhawk), the mod won't apply, although perhaps it had the chance to apply otherwise.

mods/office-ui-reverter.wh.cpp
```diff
@@ -0,0 +1,592 @@
...
+void Wh_ModUninit()
+{
+    Wh_Log(L">");
+}
+
+void Wh_ModSettingsChanged()
+{
+    Wh_Log(L">");
+
+    // LoadSettings();
```
m417z, line 576: How about adding a note in the readme that relaunching Office is required for new settings to apply?

mods/office-ui-reverter.wh.cpp
Amrsatrio: DM'd.

mods/office-ui-reverter.wh.cpp
Amrsatrio: Sure, changing this. (This was copied from EP, maybe I need to change occurrences of this over there as well....)

mods/office-ui-reverter.wh.cpp
Amrsatrio: This is good enough. In the Office DLL there are only range types 0 1 and 2.

mods/office-ui-reverter.wh.cpp
Amrsatrio: Sure.

mods/office-ui-reverter.wh.cpp
Amrsatrio: My intention is to make sure that late injection should not cause unexpected behavior. I cannot tolerate crashes at all in a program suite where data losses can be infuriating.

m417z: Then how about checking for the loaded module (`GetModuleHandle`) only if loaded synchronously? And if it's always loaded dynamically, is the check even needed?

PR #2331, opened 2025-08-22
Disable grouping on the taskbar v1.3.10

m417z:
* The "Place ungrouped items together" option can now be applied only for non-pinned items.
* Fixed the mod not applying on boot when used with ExplorerPatcher.


CoLaZa32:
G

PR #2353, opened 2025-08-24
Per-Application UI Language 1.0.0

aubymori:
This mod allows you to set the list of preferred UI languages for each application. This is mainly useful if you wish to use or test another language for an application that does not have a built-in language option (such as applications that use the built-in MUI system).

**Example**: Japanese [OpenWithEx](https://github.com/aubymori/OpenWithEx) on an English system
Example

### DIFF
--- a/mods/per-app-ui-language.wh.cpp
+++ b/mods/per-app-ui-language.wh.cpp
@@ -1,0 +1,149 @@
+// ==WindhawkMod==
+// @id              per-app-ui-language
+// @name            Per-Application UI Language
+// @description     Set the Windows UI language per application.
+// @version         1.0.0
+// @author          aubymori
+// @github          https://github.com/aubymori
+// @twitter         https://twitter.com/jack
+// @homepage        https://your-personal-homepage.example.com/
+// @include         *
+// @compilerOptions -lcomdlg32
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Per-Application UI Language
+This mod allows you to set the list of preferred UI languages for each application.
+This is mainly useful if you wish to use or test another language for an application
+that does not have a built-in language option (such as applications that use the built-in
+MUI system).
+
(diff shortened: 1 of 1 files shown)


m417z:
How does it compare to https://windhawk.net/mods/per-app-language-preferences?

aubymori:
> How does it compare to https://windhawk.net/mods/per-app-language-preferences?

Trying that mod, it doesn't seem to affect MUI applications at all.

mods/per-app-ui-language.wh.cpp
```diff
@@ -0,0 +1,147 @@
...
+                
+                Wh_FreeStringSetting(pszLang);
+                break;
+            }
+
+            Wh_FreeStringSetting(pszLang);
+        }
+    }
+
+    if (!SetProcessPreferredUILanguages(MUI_LANGUAGE_NAME, pszLanguages, nullptr))
```
m417z: IMO it's best to only call this for configured processes, here and on uninit, to avoid side effects such as clearing process-configured languages.

aubymori: Done

mods/per-app-ui-language.wh.cpp
```diff
@@ -0,0 +1,152 @@
...
+                
+                Wh_FreeStringSetting(pszLang);
+                break;
+            }
+
+            Wh_FreeStringSetting(pszLang);
+        }
+    }
+
+    if ((pszLanguages || g_fLangSet) || !SetProcessPreferredUILanguages(MUI_LANGUAGE_NAME, pszLanguages, nullptr))
```
m417z: I think you meant:

```suggestion
    if ((pszLanguages || g_fLangSet) && !SetProcessPreferredUILanguages(MUI_LANGUAGE_NAME, pszLanguages, nullptr))
```

aubymori: Fixed

PR #2381, opened 2025-08-31
Add quick hide desktop icons.wh.cpp

youlanan:
Add quick-hide-desktop-icons mod
- Toggles desktop icon visibility with customizable mouse clicks (left/right/middle, single/double) or keyboard hotkeys
- Supports English, Simplified Chinese, Traditional Chinese, and Japanese
- Tested on Windows 11 24H2

### DIFF
--- a/mods/quick-hide-desktop-icons.wh.cpp
+++ b/mods/quick-hide-desktop-icons.wh.cpp
@@ -8,7 +8,7 @@
 // @description:zh-CN   通过可自定义的热键或鼠标点击快速隐藏/显示桌面图标
 // @description:zh-TW   透過可自訂的快捷鍵或滑鼠點擊快速隱藏/顯示桌面圖標
 // @description:ja-JP   カスタマイズ可能なホットキーまたはマウスクリックでデスクトップアイコンを素早く非表示/表示
-// @version             0.2.4
+// @version             0.2.6
 // @author              youlanan
 // @github              https://github.com/youlanan
 // @include             explorer.exe
@@ -72,19 +72,15 @@
   $description:ja-JP: 使用するマウスボタン
   $options:
   - left: Left button
-  - right: Right button
   - middle: Middle button
   $options:zh-CN:
   - left: 左键
-  - right: 右键
   - middle: 中键
   $options:zh-TW:
   - left: 左鍵
-  - right: 右鍵
   - middle: 中鍵
   $options:ja-JP:
   - left: 左ボタン
-  - right: 右ボタン
   - middle: 中ボタン
 
 - clickType: double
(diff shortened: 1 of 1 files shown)


mods/quick-hide-desktop-icons.wh.cpp
```diff
@@ -0,0 +1,455 @@
...
+} g_settings;
+
+// State variables
+static bool g_iconsHidden = false;
+static HWND g_desktopListView = nullptr;
+static DWORD g_lastClickTime = 0;
+static POINT g_lastClickPos = {0, 0};
+
+// Find desktop icons window
+HWND FindDesktopIconsWindow() {
```
m417z: Have you considered that there can be several explorer.exe instances if the "launch folder windows in a separate process" option is enabled? In this case, non-desktop processes can find the windows of another process, which is probably not what you want.

m417z (commented):
Thanks for the submission!

mods/quick-hide-desktop-icons.wh.cpp
youlanan: Thank you for pointing out the problem, and I will try to fix and improve it.

mods/quick-hide-desktop-icons.wh.cpp
```diff
@@ -145,311 +141,342 @@
...
+    if (g_defView && IsWindow(g_defView)) return g_defView;
+
+    DWORD currentPid = GetCurrentProcessId();
+    DWORD shellPid = GetShellProcessIdCached();
+    HWND shellWnd = GetShellWindow();
+    HWND defView = shellWnd ? FindWindowExW(shellWnd, nullptr, L"SHELLDLL_DefView", nullptr) : nullptr;
+
+    if (defView) {
+        DWORD pid;
+        GetWindowThreadProcessId(defView, &pid);
```
m417z, line 241: It should be enough to check for the process id of one of the windows, e.g. `shellWnd`. All child windows are part of the same process.

mods/quick-hide-desktop-icons.wh.cpp
```diff
@@ -145,311 +141,342 @@
...
+            return g_defView;
         }
     }
 
-    HWND workerW = nullptr;
-    while ((workerW = FindWindowExW(nullptr, workerW, L"WorkerW", nullptr))) {
-        HWND defView = FindWindowExW(workerW, nullptr, L"SHELLDLL_DefView", nullptr);
+    HWND worker = g_lastWorkerW && IsWindow(g_lastWorkerW) ? g_lastWorkerW : nullptr;
+    DWORD start = GetTickCount();
+    while ((worker = FindWindowExW(nullptr, worker, L"WorkerW", nullptr))) {
```
m417z, line 250: It'd be better to add a delay in each iteration (e.g. use `Sleep(20);`) to avoid stressing the CPU unnecessarily.

mods/quick-hide-desktop-icons.wh.cpp
```diff
@@ -145,311 +141,342 @@
...
+        if (!DispatchMessageW_Original) {
+            Wh_SetFunctionHook((void*)DispatchMessageW, (void*)DispatchMessageW_Hook, (void**)&DispatchMessageW_Original);
+            QH_LOG(L"Hook installed");
+        }
+    } else if (!isShellNow && g_isShellProcess) {
+        g_isShellProcess = false;
+        if (DispatchMessageW_Original) {
+            #ifdef Wh_RemoveFunctionHook
+            Wh_RemoveFunctionHook((void*)DispatchMessageW, (void*)DispatchMessageW_Hook);
+            #endif
```
m417z, line 442: `Wh_RemoveFunctionHook` is not a macro, so this `#ifdef` block will never be compiled. In addition, `Wh_SetFunctionHook` and `Wh_RemoveFunctionHook` don't have any effect until `Wh_ApplyHookOperation` is called, unless hooks are set in `Wh_ModInit`. Generally, it's recommended to set the hooks once in `Wh_ModInit` and then use them when needed, or just call the original function.

Refer to the documentation for details:
https://github.com/ramensoftware/windhawk/wiki/Creating-a-new-mod#wh_setfunctionhook

mods/quick-hide-desktop-icons.wh.cpp
```diff
@@ -145,311 +141,342 @@
...
-        Wh_Log(L"Restored desktop icons");
+        QH_LOG(L"Icons restored");
     }
-
-    g_desktopListView = nullptr;
-    Wh_Log(L"Uninitialized");
+    if (DispatchMessageW_Original) {
+        #ifdef Wh_RemoveFunctionHook
+        Wh_RemoveFunctionHook((void*)DispatchMessageW, (void*)DispatchMessageW_Hook);
+        #endif
```
m417z, line 468: Hooks are removed automatically before `Wh_ModUninit` is called.

PR #2387, opened 2025-09-02
Hide pin (xaml)

danalec:
https://github.com/ramensoftware/windhawk-mods/pull/2216#issuecomment-3243598890


m417z:
I'm posting here my comment from Discord:

The PR check fails only for Windhawk 1.5.1 because it doesn't have the right headers, you can ignore this failure.

Why are you creating a new mod and not using file explorer styler instead? I believe it should work.
Also, a problem with having more than one mod using XAML diagnostics is that these mods won't work together, as the API only supports a single consumer.

m417z:
> yea, because the skinning stuff is more important

I believe that you can configure the styler mod to hide the element. Have you tried it?

> Is it possible to add a conditional loading for the api request?

What do you mean by "conditional loading"?

PR #2427, opened 2025-09-15
Old Regedit Tree Icons 1.0.0

aubymori:
In Windows XP and before, the Registry Editor's tree view had its own icons instead of using system icons. This mod reverts that behavior and makes it use its own icons again.

**Before**:
Before

**After**:
After

### DIFF
--- a/mods/old-regedit-tree-icons.wh.cpp
+++ b/mods/old-regedit-tree-icons.wh.cpp
@@ -97,23 +97,16 @@ DECLARE_HOOK_FUNCTION(void, WINAPI, AddSystemImageIcon,
 
 #define STDCALL_STR FOR_64_32(L"__cdecl", L"__stdcall")
 
-#define BEGIN_SYMBOL_HOOKS(name) const WindhawkUtils::SYMBOL_HOOK name[] = {
-
-#define SIMPLE_SYMBOL_HOOK(name, ...) \
-{                                     \
-    {                                 \
-        __VA_ARGS__                   \
-    },                                \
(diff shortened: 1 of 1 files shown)


mods/old-regedit-tree-icons.wh.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+#ifdef _WIN64
+#   define FOR_64_32(for64, for32) for64
+#else
+#   define FOR_64_32(for64, for32) for32
+#endif
+
+#define STDCALL_STR FOR_64_32(L"__cdecl", L"__stdcall")
+
+#define BEGIN_SYMBOL_HOOKS(name) const WindhawkUtils::SYMBOL_HOOK name[] = {
```
m417z: These macros will prevent the symbol cache scripts from extracting and caching the mod's symbols. It should be possible to adjust the scripts to handle these macros, but it'd be easier to avoid using the macros. It should be a simple change. Thanks.

aubymori: Fixed

aubymori:
@m417z FYI, I don't know where the "symbol cache scripts" are if they are publicly accessible, but probably the easiest way to allow for macros like the ones I used is to run the file through Clang's preprocessor. Obviously, this will only work if you can get the Windhawk compiler wherever the symbol cache scripts are run.

m417z:
Well, it still ended up being unsupported due to `#define STDCALL_STR FOR_64_32(L"__cdecl", L"__stdcall")` 😐

PR #2445, opened 2025-09-21
new mod: Explorer double F2 to rename file extension

leaumar:
As the comment in the mod says, I only have basic C++ experience and none using windows shell apis. I did test this in every way I could think and found no problems after fixing initial issues like crashes on unloading (not unhooking my function in uninit).

Usecase: I often rename zip files to cbz (image sets), misnamed videos/images from websites to the right extension, js to ts, that kind of stuff. Qdir/Qttabbar and some other tools implement this same feature and I miss having it in vanilla explorer. There's already the mod to suppress the rename warning, now let's make the rename itself easier to get to. Ctrl-backspace in explorer's rename field works weirdly so changing just the extension has always been a PITA.

Edit: reference to [extension-change-no-warning](https://windhawk.net/mods/extension-change-no-warning)

### DIFF
--- a/mods/explorer-double-f2-rename-extension.wh.cpp
+++ b/mods/explorer-double-f2-rename-extension.wh.cpp
@@ -16,7 +16,8 @@ second press selects just the extension. That's handy to e.g. rename a zip file
 to a cbz file. This mod implements that same feature: double-press F2
 in Explorer to rename a file's extension.
 
-This works great together with [extension-change-no-warning](https://windhawk.net/mods/extension-change-no-warning).
+This works great together with
+[extension-change-no-warning](https://windhawk.net/mods/extension-change-no-warning).
 */
 // ==/WindhawkModReadme==
 
@@ -31,24 +32,106 @@ This works great together with [extension-change-no-warning](https://windhawk.ne
 // ChatGPT and the Windhawk Discord server.
 
 #include <string>
+#include <vector>
 
-DWORD explorerThreadId;
-HHOOK keyboardHook;
+static void SelectExtension(HWND editControl) {
+    // typical max filename length
+    std::wstring buffer(260, L'\0');
+    int copied = GetWindowTextW(editControl, buffer.data(), (int)buffer.size());
+    if (copied <= 0) {
+        return;
+    }
+    // trim to actual length
+    buffer.resize(copied);
+
+    size_t dotIndex = buffer.find_last_of(L'.');
+    if (dotIndex > 0 && dotIndex != std::wstring::npos) {
+        int start = (int)dotIndex + 1;
+        SendMessageW(editControl, EM_SETSEL, start, (WPARAM)buffer.size());
+        std::wstring extension = buffer.substr(dotIndex + 1);
+        Wh_Log(L"Selected extension \"%s\".", extension.c_str());
+    }
+}
 
 static ULONGLONG lastF2Time = 0;
 static bool lastWasF2 = false;
+static bool CALLBACK DoHandleKeyboard(int nCode, WPARAM wParam, LPARAM lParam) {
+    bool shouldProcess = nCode >= 0;
+    bool isKeyUp = lParam & 0x80000000;
+    if (!shouldProcess || !isKeyUp) {
+        return false;
+    }
 
-static void SelectExtension(HWND edit) {
-    int length = GetWindowTextLengthW(edit);
-    if (length > 0) {
-        std::wstring buffer(length, L'\0');
-        GetWindowTextW(edit, &buffer[0], length + 1);
+    bool isF2 = wParam == VK_F2;
+    if (!isF2) {
+        lastWasF2 = false;
+        return false;
+    }
 
-        size_t dotIndex = buffer.find_last_of(L'.');
-        int start = (dotIndex != std::wstring::npos && dotIndex != 0)
-                        ? (int)dotIndex + 1
-                        : 0;
-        SendMessageW(edit, EM_SETSEL, start, length);
+    auto doubleF2Time = (DWORD)Wh_GetIntSetting(L"DoubleF2MilliSeconds");
+    ULONGLONG now = GetTickCount64();
+
+    auto delta = now - lastF2Time;
+    lastF2Time = now;
+    bool isDouble = lastWasF2 && (delta <= doubleF2Time);
+    lastWasF2 = true;
+
+    Wh_Log(L"F2 pressed: isDouble=%d, delta=%llu.", isDouble, delta);
+
+    if (isDouble) {
+        HWND focus = GetFocus();
+        if (focus != nullptr) {
+            wchar_t cls[32];
+            GetClassNameW(focus, cls, _countof(cls));
+            if (_wcsicmp(cls, L"Edit") == 0) {
+                Wh_Log(L"Double F2 in edit control, selecting extension.");
+                SelectExtension(focus);
+                return true;
+            }
+        }
+    }
+
+    return false;
+}
+static LRESULT CALLBACK HandleKeyboard(int nCode,
+                                       WPARAM wParam,
+                                       LPARAM lParam) {
+    bool handled = DoHandleKeyboard(nCode, wParam, lParam);
+    return handled ? 0 : CallNextHookEx(nullptr, nCode, wParam, lParam);
+}
+
+static std::vector<HHOOK> keyboardHooks;
+static void AddKeyboardHook(DWORD threadId) {
+    HHOOK hook =
+        SetWindowsHookExW(WH_KEYBOARD, HandleKeyboard, nullptr, threadId);
+    if (hook != nullptr) {
+        keyboardHooks.push_back(hook);
+        Wh_Log(L"Installed keyboard hook %p on thread %u", hook, threadId);
+    }
+}
+static void RemoveKeyboardHooks() {
+    for (HHOOK hook : keyboardHooks) {
+        if (hook != nullptr) {
+            bool ok = UnhookWindowsHookEx(hook);
+            Wh_Log(L"Unhook %p -> %d", hook, ok);
+        }
+    }
+    keyboardHooks.clear();
+}
+
+static void HookIfExplorer(HWND windowHandle) {
+    if (windowHandle != nullptr && IsWindow(windowHandle)) {
+        wchar_t clazz[64];
+        if (GetClassNameW(windowHandle, clazz, _countof(clazz))) {
+            if (_wcsicmp(clazz, L"CabinetWClass") == 0 ||
+                _wcsicmp(clazz, L"ExploreWClass") == 0) {
+                DWORD threadId =
+                    GetWindowThreadProcessId(windowHandle, nullptr);
+                AddKeyboardHook(threadId);
+                Wh_Log(L"Hooked Explorer window: hwnd=0x%p class=%s.",
+                       windowHandle, clazz);
+            }
+        }
     }
 }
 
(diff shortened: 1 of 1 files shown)


mods/explorer-double-f2-rename-extension.wh.cpp
```diff
@@ -0,0 +1,171 @@
...
+                        return 0;  // swallow
+                    }
+                }
+            }
+        } else {
+            lastWasF2 = false;
+        }
+    }
+
+    return CallNextHookEx(keyboardHook, nCode, wParam, lParam);
```
m417z: The first parameter is ignored, it can be:
```suggestion
    return CallNextHookEx(nullptr, nCode, wParam, lParam);
```

mods/explorer-double-f2-rename-extension.wh.cpp
```diff
@@ -0,0 +1,171 @@
...
+    Wh_Log(L"Double F2 uninitializing.");
+
+    if (keyboardHook != nullptr) {
+        BOOL ok = UnhookWindowsHookEx(keyboardHook);
+        Wh_Log(L"UnhookWindowsHookEx(keyboardHook) -> %d", ok ? 1 : 0);
+        keyboardHook = nullptr;
+    }
+
+    // If you ever installed additional per-window hooks when new explorer
+    // windows were created, make sure you unhook them when you replace them.
```
m417z: This comment doesn't look correct, and the unhooking logic isn't correct either, since only the last `keyboardHook` is unhooked. A solution is to keep all hook handles in an array.

mods/explorer-double-f2-rename-extension.wh.cpp
```diff
@@ -0,0 +1,171 @@
...
+*/
+// ==/WindhawkModSettings==
+
+// I am not an experienced C++ programmer and do not know the Windows APIs.
+// This code was cobbled together with a lot trial & error, using input from
+// ChatGPT and the Windhawk Discord server.
+
+#include <string>
+
+DWORD explorerThreadId;
```
m417z: This doesn't need to be a global variable.

m417z (commented):
Looks nice overall, well done!

mods/explorer-double-f2-rename-extension.wh.cpp
leaumar: Is it ok to just keep them all until uninit and bulk unhook them then?

mods/explorer-double-f2-rename-extension.wh.cpp
```diff
@@ -121,51 +163,34 @@ static HWND WINAPI HookedCreateWindowExW(DWORD dwExStyle,
...
-    explorerThreadId = GetWindowThreadProcessId(shellWnd, nullptr);
-    keyboardHook = SetWindowsHookExW(WH_KEYBOARD, HandleKeyboard, nullptr,
-                                     explorerThreadId);
 
+    Wh_Log(L"Hooking the desktop (shell) thread.");
+    DWORD shellThreadId = GetWindowThreadProcessId(GetShellWindow(), nullptr);
+    AddKeyboardHook(shellThreadId);
+
+    Wh_Log(L"Hooking already open Explorer windows.");
+    EnumWindows(EnumWindowsProc, 0);
```
m417z, line 208: `EnumWindows` enumerates windows from all processes. Before setting a hook on a window's thread, check that it's actually part of the current process. You can use `GetWindowThreadProcessId` which you're already calling, and `GetCurrentProcessId()`.

mods/explorer-double-f2-rename-extension.wh.cpp
```diff
@@ -31,24 +32,106 @@ This works great together with [extension-change-no-warning](https://windhawk.ne
...
+    std::wstring buffer(260, L'\0');
+    int copied = GetWindowTextW(editControl, buffer.data(), (int)buffer.size());
+    if (copied <= 0) {
+        return;
+    }
+    // trim to actual length
+    buffer.resize(copied);
+
+    size_t dotIndex = buffer.find_last_of(L'.');
+    if (dotIndex > 0 && dotIndex != std::wstring::npos) {
```
m417z: I didn't try the mod yet, but from the code I'd assume that the `dotIndex > 0` condition will make it so that an extension isn't selected for file names starting with periods, e.g. `.git`. If my understanding is correct, I think it's better to remove that check.

leaumar: I was on the fence about it but in the end I decided to keep it this way because the tail part of dotfiles isn't really a filetype extension, it's just its meaningful name and the dot is only there because someone on unix felt it shouldn't show up in `ls`. Files change extension all the time because they're mis-mimetyped, but renaming a dotfile turns it into something totally different in purpose.

On the other hand, purely lexicographically it would be more consistent, and it's not like there's anything else of use that double-f2 could do on a dotfile... I'll change it tomorrow

mods/explorer-double-f2-rename-extension.wh.cpp
```diff
@@ -0,0 +1,200 @@
...
+        HookIfExplorerFileView(hwnd, threadId);
+    }
+
+    // continue enumeration
+    return true;
+}
+
+void Wh_ModInit() {
+    Wh_Log(L"Hooking the desktop (shell) thread.");
+    DWORD shellThreadId = GetWindowThreadProcessId(GetShellWindow(), nullptr);
```
m417z, line 204: `GetShellWindow()` can return `nullptr` if the mod is loaded before the desktop is created. There are two issues with it:
* `shellThreadId` will be 0 and `AddKeyboardHook` will behave in an unexpected way.
* When the desktop is created at a later time, its thread won't be hooked. Maybe checking for `hwnd == GetShellWindow()` in the `CreateWindowExW` hook will work. In either way, I suggest to test the mod when it's loaded early too, e.g. after restarting explorer or after rebooting.

mods/explorer-double-f2-rename-extension.wh.cpp
```diff
@@ -0,0 +1,200 @@
...
+                                         int nHeight,
+                                         HWND hWndParent,
+                                         HMENU hMenu,
+                                         HINSTANCE hInstance,
+                                         LPVOID lpParam) {
+    // always call the original first
+    HWND hwnd = originalCreateWindowExW(dwExStyle, lpClassName, lpWindowName,
+                                        dwStyle, X, Y, nWidth, nHeight,
+                                        hWndParent, hMenu, hInstance, lpParam);
+    HookIfExplorerFileView(hwnd, 0);
```
m417z: You can pass `GetCurrentThreadId()` instead of 0 here.

mods/explorer-double-f2-rename-extension.wh.cpp
```diff
@@ -0,0 +1,200 @@
...
+                                       WPARAM wParam,
+                                       LPARAM lParam) {
+    bool handled = DoHandleKeyboard(nCode, wParam, lParam);
+    return handled ? 0 : CallNextHookEx(nullptr, nCode, wParam, lParam);
+}
+
+static std::vector<HHOOK> keyboardHooks;
+static void AddKeyboardHook(DWORD threadId) {
+    HHOOK hook =
+        SetWindowsHookExW(WH_KEYBOARD, HandleKeyboard, nullptr, threadId);
```
m417z: I'd add an extra check that `threadId` isn't zero here too. Calling `SetWindowsHookExW` with a zero thread id hooks into all threads, and that's not what you're trying to do here.

mods/explorer-double-f2-rename-extension.wh.cpp
leaumar: Does it still need the 0 test and GetWindowThreadProcessId in HookIfExplorerFileView then or will it always be a good value?

mods/explorer-double-f2-rename-extension.wh.cpp
leaumar: 0 is covered upstream in one caller, the other caller is the desktop hook your other comment is about. I feel like there's some redundancy here but I can't put my finger on it, getting lost a little in all the magic numbers. If it's 0 here, just skip the entire hook?

mods/explorer-double-f2-rename-extension.wh.cpp
m417z: It's needed for the other place where you pass 0. Here you know the thread id, so it's a minor optimization to save the additional query.

mods/explorer-double-f2-rename-extension.wh.cpp
leaumar: I'm not sure what you'd have me do tbh. And why check in the hook proc if the hwnd == shellwindow if we're calling the keyboard hook next anyway, I don't think you mean to skip it in that case?

mods/explorer-double-f2-rename-extension.wh.cpp
leaumar: I don't pass 0 anywhere else though, only here (it's GetCurrentThreadId() now)

mods/explorer-double-f2-rename-extension.wh.cpp
m417z: Yes, it's covered in the first place, and will be coverer in the second place. But it is bullet-proof? For example, `GetWindowThreadProcessId` might return zero because the window is destroyed right after the call to `IsWindow`.

It's a bit pedantic, it's just that `SetWindowsHookExW` has special and potentially unexpected behavior for zero, which I think is best avoided.

> If it's 0 here, just skip the entire hook?

Yes, that's what I meant.

mods/explorer-double-f2-rename-extension.wh.cpp
m417z: Right, for some reason I thought you did. In this case sure, the test can be removed.

mods/explorer-double-f2-rename-extension.wh.cpp
m417z: I'll try to clarify, let me know if it's still unclear. I wrote similar logic many times, so perhaps I'm missing something that seems obvious to me, but isn't really obvious, in my explanation.

Let's say I install the mod, then reboot. explorer.exe will start with the mod's `Wh_ModInit` being called early in the process launch, before any windows are created. In this case, `GetShellWindow` will return null (it's not created yet) and `EnumWindows` below will also find no windows which belong to this process, because there are none yet.

After `Wh_ModInit` returns, the process will begin initializing. The `CreateWindowExW` hook will capture and handle any newly created explorer window, but it won't handle the desktop, because `GetShellWindow` was null in `Wh_ModInit`, and the `CreateWindowExW` hook has no special treatment for it when it's created.

Does it make sense?

leaumar: Ok, but then how can we detect the desktop being created and hook into it after init if CreateWindowExW won't trigger for it?

m417z: Currently, in `HookedCreateWindowExW` you only call `HookIfExplorerFileView` which handles explorer windows. My suggestion was to add `HookIfShellWindow` which will check if the created window is the same as `GetShellWindow()`, and apply the hook if that's the case.

I also suggested to test it because I'm not sure `GetShellWindow` will return the right window handle just after the window is created. If not, you might need to resort to another way to check it (e.g. using the class name).

leaumar: Oh ok, so CreateWindowExW should fire for the desktop (I thought you said it wouldn't), and then yeah I can do that check.

But I think just adding Progman or WorkerW next to CabinetWClass and ExploreWClass in the existing hook on window creation will also do the trick though? I just tested it by restarting explorer, before my change the mod no longer worked on the restarted desktop, and after adding Progman to the check it works.

PR #2468, opened 2025-09-28
Winver Custom Text v1.0.1

ilikecoding-197:
Fixed README, and fixed up some other things

Also, theres seems to be another mod, [Winver Text Customizer](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/winver-text-customizer.wh.cpp), that seems to do the exact same thing my mod does, so what's that about?

### DIFF
--- a/mods/winver-custom-text.wh.cpp
+++ b/mods/winver-custom-text.wh.cpp
@@ -2,7 +2,7 @@
 // @id              winver-custom-text
 // @name            Winver Custom Text
 // @description     Allows you to set the text displayed in winver.exe.
-// @version         1.0
+// @version         1.0.1
 // @author          Thomas Shrader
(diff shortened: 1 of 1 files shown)


mods/winver-custom-text.wh.cpp
```diff
@@ -69,7 +70,6 @@ void LoadSettings() {
     Settings.firstLine = Wh_GetStringSetting(L"firstline");
     Settings.otherstuff = Wh_GetStringSetting(L"otherstuff");
 
-    ShellAboutArgs.szApp = new wchar_t[1024];
     swprintf((wchar_t *)ShellAboutArgs.szApp, L"%s#%s", Settings.title, Settings.firstLine);
```
m417z, line 73: I don't think you need this cast anymore (`(wchar_t *)`).

PR #2472, opened 2025-09-30
Close explorer file ext change dialog v1.0

CandyTek:


Anixx:
Why would one need this if we already have a mod that disables this dialog altogether? https://windhawk.net/mods/extension-change-no-warning

PR #2494, opened 2025-10-04
Add mod to redirect Modern Notepad to Classic Editor

CherryDT:
This mod redirects Windows 11's Notepad to a configurable classic editor, including an elevate-on-Ctrl+Shift feature.


m417z:
I tried the mod, and it doesn't work for me. It causes the modern Notepad to be launched again and again, resulting in a huge amount of Notepad windows after a single launch.

I tried to implement a similar mod at some point, but came to the conclusion that the redirection functionality is deeply integrated into Windows, and it makes more sense to just disable the modern Notepad via the settings, under "App execution aliases".

CherryDT:
This mod goes beyond what disabling the alias can do, but I forgot that the app execution alias needs to be disabled too for this to work. I now added a check for it and if the user agrees I automatically disable the alias.

CherryDT:
Btw, in case you wonder why I did force-push all the time - sorry, it was a misunderstanding, I just realized. When I read the readme, somehow I read the PR has to consist of a single _commit_, when in fact it just said to contain a single _file_.

CherryDT:
I just realized (through testing) that `%programfiles%` _does_ work in `@include`, I didn't find this in the docs. Would be nice to add it because I think in general it should be preferred not to hardcode such paths!

mods/modern-notepad-to-classic-editor.wh.cpp
```diff
@@ -0,0 +1,429 @@
...
+  return 0;
+}
+
+// ---------------------------------------------------------------------
+// Windhawk entry points
+// ---------------------------------------------------------------------
+BOOL Wh_ModInit() {
+  // Load settings
+  {
+    std::wstring pathSetting = Wh_GetStringSetting(L"classicPath");
```
m417z: There's a memory leak here. You can use `WindhawkUtils::StringSetting` for a C++ wrapper that frees it automatically.

mods/modern-notepad-to-classic-editor.wh.cpp
```diff
@@ -0,0 +1,429 @@
...
+    if (pathSetting.empty()) pathSetting = L"%SystemRoot%\\System32\\notepad.exe";
+    g_classicPathSetting = pathSetting;
+
+    g_ctrlShiftElevate = Wh_GetIntSetting(L"ctrlShiftElevate", 1) != 0;
+    g_waitForClassic   = Wh_GetIntSetting(L"waitForClassic", 1) != 0;
+  }
+
+  // Only act inside the WindowsApps Notepad process
+  std::wstring cur = GetCurrentExePath();
+  if (!IsModernWindowsAppsNotepadPath(cur)) {
```
m417z: Isn't this check redundant since the mod only targets the modern Notepad already?

mods/modern-notepad-to-classic-editor.wh.cpp
```diff
@@ -0,0 +1,429 @@
...
+            return FALSE; // Don't patch
+          }
+        }
+      }
+      RegCloseKey(hKey);
+    }
+  }
+
+  // If this is already an "inner" launch with /SESSION, exit quietly
+  if (ArgsContainSession(argv)) {
```
m417z: You can use `isSessionLaunch` here.

mods/modern-notepad-to-classic-editor.wh.cpp
```diff
@@ -0,0 +1,429 @@
...
+    if (RegOpenKeyExW(HKEY_CURRENT_USER, L"Software\\Microsoft\\Windows\\CurrentVersion\\App Paths\\notepad.exe", 0, KEY_READ | KEY_WRITE, &hKey) == ERROR_SUCCESS) {
+      WCHAR value[1024] = {};
+      DWORD size = sizeof(value);
+      DWORD type;
+      if (RegQueryValueExW(hKey, NULL, NULL, &type, (BYTE*)value, &size) == ERROR_SUCCESS && type == REG_SZ) {
+        std::wstring val(value);
+        if (val.find(L"WindowsApps") != std::wstring::npos) {
+          int result = isSessionLaunch ? IDNO : MessageBoxW(NULL, L"It seems like an app execution alias for notepad.exe to the modern Notepad is set in Windows settings and this mod cannot work while the alias is enabled. Do you want to disable this alias?", L"Redirect Modern Notepad to Classic Editor (Windhawk Mod)", MB_YESNO | MB_ICONQUESTION);
+          if (result == IDYES) {
+            RegCloseKey(hKey);
```
m417z: Here, `RegCloseKey(hKey)` will be called twice.

mods/modern-notepad-to-classic-editor.wh.cpp
CherryDT: Fixed

mods/modern-notepad-to-classic-editor.wh.cpp
CherryDT: Fixed

mods/modern-notepad-to-classic-editor.wh.cpp
CherryDT: Yes, at some point I had just `notepad.exe` because I didn't want to hardcode the program files path and didn't realize environment variables work there.

mods/modern-notepad-to-classic-editor.wh.cpp
CherryDT: Fixed.

mods/modern-notepad-to-classic-editor.wh.cpp
```diff
@@ -0,0 +1,465 @@
...
+# Redirect Modern Notepad to Classic Editor
+
+Windows 11 replaces the classic Notepad with a Microsoft Store app.  
+This mod redirects all launches of the **modern Notepad** (including context menu items like “Edit in Notepad”) to your chosen **classic editor** instead.
+
+There is already a way to do this "half-way" natively by disabling the `notepad.exe` app execution alias in Windows' settings, but this doesnt work for the default "Edit in Notepad" context menu item, default file associations, `ms-notepad:` links and other entry points. This mod handles all of these cases seamlessly.
+
+By default, it launches the old `notepad.exe` from `%SystemRoot%\System32`, but you can configure it to run **any editor of your choice** (e.g. Notepad2, Notepad++, VS Code).
+
+## ⚠️ IMPORTANT: You need to also disable the app execution alias for Notepad in Windows Settings > Apps > App execution aliases, otherwise this mod won't work!
```
m417z:
```suggestion
## ⚠️ IMPORTANT: You need to also disable the app execution alias for Notepad in Windows Settings > Apps > Advanced app settings > App execution aliases, otherwise this mod won't work!
```

mods/modern-notepad-to-classic-editor.wh.cpp
```diff
@@ -0,0 +1,465 @@
...
+      DWORD size = sizeof(value);
+      DWORD type;
+      LONG queryRet = RegQueryValueExW(hKey, NULL, NULL, &type, (BYTE*)value, &size);
+      RegCloseKey(hKey);
+      if (queryRet == ERROR_SUCCESS && type == REG_SZ) {
+        std::wstring val(value);
+        if (val.find(L"WindowsApps") != std::wstring::npos) {
+          if (!isSessionLaunch) {
+            // Ask the user to disable app alias for notepad
+            MessageBoxW(NULL, L"The app execution alias for Notepad is enabled, which interferes with this mod. Please disable it in Windows Settings > Apps > App execution aliases.", L"Modern Notepad to Classic Editor", MB_ICONWARNING);
```
m417z:
```suggestion
            MessageBoxW(NULL, L"The app execution alias for Notepad is enabled, which interferes with this mod. Please disable it in Windows Settings > Apps > Advanced app settings > App execution aliases.", L"Modern Notepad to Classic Editor", MB_ICONWARNING);
```

mods/modern-notepad-to-classic-editor.wh.cpp
```diff
@@ -0,0 +1,465 @@
...
+}
+
+// Check Ctrl+Shift pressed
+static bool IsCtrlShiftPressed() {
+  return (GetAsyncKeyState(VK_CONTROL) & 0x8000) &&
+         (GetAsyncKeyState(VK_SHIFT) & 0x8000);
+}
+
+// Identify WindowsApps Notepad executable
+static bool IsModernWindowsAppsNotepadPath(const std::wstring& exePath) {
```
m417z: The function is unused,

mods/modern-notepad-to-classic-editor.wh.cpp
```diff
@@ -0,0 +1,465 @@
...
+  int X,
+  int Y,
+  int nWidth,
+  int nHeight,
+  HWND hWndParent,
+  HMENU hMenu,
+  HINSTANCE hInstance,
+  LPVOID lpParam)
+{
+  Sleep(INFINITE); // stall forever
```
m417z:
```suggestion
  Sleep(INFINITE); // stall forever
  return NULL;
```

mods/modern-notepad-to-classic-editor.wh.cpp
```diff
@@ -0,0 +1,465 @@
...
+    SHELLEXECUTEINFOW sei = { sizeof(sei) };
+    sei.lpVerb = L"runas";
+    sei.lpFile = classicPath.c_str();
+    std::wstring argTail = BuildArgsOnly();
+    sei.lpParameters = argTail.c_str();
+    sei.nShow = SW_SHOWNORMAL;
+
+    if (ShellExecuteExW(&sei)) {
+      if (g_waitForClassic && sei.hProcess) {
+        WaitForSingleObject(sei.hProcess, INFINITE);
```
m417z, line 344: This case, and the `WaitForSingleObject` below, cause the hourglass/loading cursor to be shown for some time. It can be avoided with [something like this](https://stackoverflow.com/a/3865519).

CherryDT:
Thanks, implemented all your suggestions.

Also removed the `ms-notepad:` URI handling because regular immersive notepad doesn't handle it either (and I was not handling encoded spaces anyway).

mods/modern-notepad-to-classic-editor.wh.cpp
```diff
@@ -0,0 +1,505 @@
...
+// Windhawk entry points
+// ---------------------------------------------------------------------
+BOOL Wh_ModInit() {
+  // Load settings
+  {
+    g_classicPathSetting = WindhawkUtils::StringSetting::make(L"classicPath");
+    if (g_classicPathSetting.empty()) g_classicPathSetting = L"%SystemRoot%\\System32\\notepad.exe";
+
+    g_ctrlShiftElevate = Wh_GetIntSetting(L"ctrlShiftElevate", 1) != 0;
+    g_waitForClassic   = Wh_GetIntSetting(L"waitForClassic", 1) != 0;
```
m417z:
```suggestion
    g_ctrlShiftElevate = Wh_GetIntSetting(L"ctrlShiftElevate") != 0;
    g_waitForClassic   = Wh_GetIntSetting(L"waitForClassic") != 0;
```

mods/modern-notepad-to-classic-editor.wh.cpp
```diff
@@ -0,0 +1,505 @@
...
+          tdc.hwndParent = NULL;
+          tdc.hInstance = NULL;
+          tdc.dwFlags = TDF_USE_COMMAND_LINKS | TDF_ALLOW_DIALOG_CANCELLATION;
+          tdc.dwCommonButtons = TDCBF_CANCEL_BUTTON;
+          tdc.pszWindowTitle = L"Modern Notepad to Classic Editor";
+          tdc.pszMainIcon = TD_WARNING_ICON;
+          tdc.pszMainInstruction = L"App execution alias for Notepad is enabled";
+          tdc.pszContent = L"The app execution alias for Notepad is enabled, which interferes with this mod. To make this mod work, you need to either disable the app execution alias globally or install a local bypass.\n\nDisabling the alias globally globally should be fine in most cases, but it doesn't allow overriding notepad with a custom editor when invoked directly. Installing a local bypass can be useful in advanced scenarios and creates a registry entry that overrides the alias for this app only, but this modification persists even after disabling the mod, even though it should not have any effect on normal operations. To undo it, reset the Notepad app in Settings > Apps > Installed apps > Notepad (...) > Advanced > Reset.\n\nChoose an option below:";
```
m417z: "globally globally" - appears twice.
```suggestion
          tdc.pszContent = L"The app execution alias for Notepad is enabled, which interferes with this mod. To make this mod work, you need to either disable the app execution alias globally or install a local bypass.\n\nDisabling the alias globally should be fine in most cases, but it doesn't allow overriding notepad with a custom editor when invoked directly. Installing a local bypass can be useful in advanced scenarios and creates a registry entry that overrides the alias for this app only, but this modification persists even after disabling the mod, even though it should not have any effect on normal operations. To undo it, reset the Notepad app in Settings > Apps > Installed apps > Notepad (...) > Advanced > Reset.\n\nChoose an option below:";
```

PR #2495, opened 2025-10-04
Add mod to reduce Dropbox Explorer CPU usage

CherryDT:
This mod disables SHChangeNotify calls in Dropbox to reduce CPU usage and prevent Explorer slowdowns during large syncs.

### DIFF
--- a/mods/dropbox-reduce-explorer-cpu-usage.wh.cpp
+++ b/mods/dropbox-reduce-explorer-cpu-usage.wh.cpp
@@ -1,0 +1,64 @@
+// ==WindhawkMod==
+// @id              dropbox-reduce-explorer-cpu-usage
+// @name            Dropbox: Reduce Explorer CPU usage on sync
+// @description     In Dropbox, disable SHChangeNotify calls to avoid Explorer slowdowns and high CPU usage from per-file overlay updates during large syncs.
+// @version         1.1
+// @author          David Trapp (CherryDT)
+// @github          https://github.com/CherryDT
+// @include         dropbox.exe
+// @architecture    x86-64
+// @compilerOptions -lshell32
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Dropbox: Reduce Explorer CPU usage on sync
+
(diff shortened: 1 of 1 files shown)


mods/dropbox-reduce-explorer-cpu-usage.wh.cpp
```diff
@@ -0,0 +1,89 @@
...
+// ---------------------------------------------------------------------
+// Windhawk entry points
+// ---------------------------------------------------------------------
+BOOL Wh_ModInit() {
+  // Ensure shell32 is loaded.
+  HMODULE hShell32 = GetModuleHandleW(L"shell32.dll");
+  if (!hShell32) {
+    hShell32 = LoadLibraryW(L"shell32.dll");
+    if (!hShell32)
+      return TRUE; // soft-fail
```
m417z: Why "soft-fail"? Why not return `FALSE` in this case?

mods/dropbox-reduce-explorer-cpu-usage.wh.cpp
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  auto real = reinterpret_cast<FnSHChangeNotify>(
+    GetProcAddress(hShell32, "SHChangeNotify"));
+  if (!real) {
+    // Some toolchains/platforms might use stdcall-decorated name on 32-bit.
```
m417z: This is an exported function, so I believe `SHChangeNotify` should always work. Did you encounter a Windows version where it's exported as `_SHChangeNotify@16`?

Also, the mod only targets 64-bit, so this code is never compiled.

mods/dropbox-reduce-explorer-cpu-usage.wh.cpp
```diff
@@ -0,0 +1,89 @@
...
+{
+  // No-op
+  return;
+}
+
+#if __WIN64
+  #define WINAPI_STR L"__cdecl"
+#else
+  #define WINAPI_STR L"__stdcall"
+#endif
```
m417z: This is unused.

PR #2517, opened 2025-10-07
Add mod to hide ghost windows and inactive Groupy 2 tabs in Alt+Tab

CherryDT:
This mod filters out small or titleless windows from the Alt+Tab window switcher, preventing "ghost" windows from appearing in the list (for example, the ThinkPad External Keyboard with Trackpoint Driver is notorious for having two 0x0 windows that show up in Alt+Tab all the time, and this mod will get rid of them).

It can also hide inactive Groupy 2 tabs (which unlike in Groupy 1 do show up in Alt+Tab by default) and/or all minimized windows, depending on settings.

### DIFF
--- a/mods/alt-tab-hide-ghost-windows.wh.cpp
+++ b/mods/alt-tab-hide-ghost-windows.wh.cpp
@@ -1,6 +1,6 @@
 // ==WindhawkMod==
 // @id              alt-tab-hide-ghost-windows
-// @name            Alt+Tab: Hide ghost windows, inactive Groupy 2 tabs and/or minimized windows
+// @name            Alt+Tab: Hide ghost windows
 // @description     Hides windows in Alt+Tab that are too small or have no title, preventing "ghost" windows from cluttering the switcher. Can also hide inactive Groupy 2 tabs and/or minimized windows.
 // @version         1.0
 // @author          David Trapp (CherryDT)


mods/alt-tab-hide-ghost-windows.wh.cpp
```diff
@@ -0,0 +1,607 @@
+// ==WindhawkMod==
+// @id              alt-tab-hide-ghost-windows
+// @name            Alt+Tab: Hide ghost windows, inactive Groupy 2 tabs and/or minimized windows
```
m417z: How about shortening the name to make it fit the preview card? Maybe just "Alt+Tab: Hide ghost windows"?

mods/alt-tab-hide-ghost-windows.wh.cpp
```diff
@@ -0,0 +1,607 @@
...
+            {
+                {LR"(public: virtual long __cdecl XamlAltTabViewHost::Show(struct IImmersiveMonitor *,enum ALT_TAB_VIEW_FLAGS,struct IApplicationView *))"},
+                &XamlAltTabViewHost_Show_Original,
+                XamlAltTabViewHost_Show_Hook,
+                true,
+            },
+            {
+                {LR"(public: virtual long __cdecl winrt::impl::consume_Windows_Internal_Shell_TaskGroups_ITaskGroupWindowInformation<struct winrt::Windows::Internal::Shell::TaskGroups::ITaskGroupWindowInformation>::Position(struct winrt::Windows::Foundation::Rect const &)const )"},
+                &ITaskGroupWindowInformation_Position_Original,
+                ITaskGroupWindowInformation_Position_Hook,
```
m417z, line 508: I compared it to `alt-tab-per-monitor`, and there are some symbol changes I wonder about - some symbols became optional and some symbols were changed. Why is that?

```diff
@@ -413,16 +503,19 @@ BOOL Wh_ModInit() {
                 {LR"(public: virtual long __cdecl XamlAltTabViewHost::Show(struct IImmersiveMonitor *,enum ALT_TAB_VIEW_FLAGS,struct IApplicationView *))"},
                 &XamlAltTabViewHost_Show_Original,
                 XamlAltTabViewHost_Show_Hook,
+                true,
             },
             {
-                {LR"(public: __cdecl winrt::impl::consume_Windows_Internal_Shell_TaskGroups_ITaskGroupWindowInformation<struct winrt::Windows::Internal::Shell::TaskGroups::ITaskGroupWindowInformation>::Position(struct winrt::Windows::Foundation::Rect const &)const )"},
+                {LR"(public: virtual long __cdecl winrt::impl::consume_Windows_Internal_Shell_TaskGroups_ITaskGroupWindowInformation<struct winrt::Windows::Internal::Shell::TaskGroups::ITaskGroupWindowInformation>::Position(struct winrt::Windows::Foundation::Rect const &)const )"},
                 &ITaskGroupWindowInformation_Position_Original,
                 ITaskGroupWindowInformation_Position_Hook,
+                true,
             },
             {
                 {LR"(long __cdecl XamlAltTabViewHost_CreateInstance(struct XamlViewHostInitializeArgs const &,struct _GUID const &,void * *))"},
                 &XamlAltTabViewHost_CreateInstance_Original,
                 XamlAltTabViewHost_CreateInstance_Hook,
+                true,
             },
             // For the old Win10 (non-XAML) Alt+Tab (can be enabled with
             // ExplorerPatcher):
@@ -439,7 +532,7 @@ BOOL Wh_ModInit() {
                 true,
             },
             {
-                {LR"(long __cdecl CAltTabViewHost_CreateInstance(struct IMultitaskingData *,struct IMultitaskingViewManagerInternal *,struct IApplicationViewSwitcher *,struct IImmersiveAppCrusher *,struct IMultitaskingViewVisibilityServiceInternal *,struct IMultitaskingViewGestureState *,struct IApplicationViewCollection *,struct ITouchGestureSettings *,struct _GUID const &,void * *))"},
+                {LR"(long __cdecl CAltTabViewHost_CreateInstance(struct IMultitaskingData *,struct IMultitaskingViewManagerInternal *,struct IApplicationViewSwitcher *,struct IImmersiveAppCrusher *,struct IMultitaskingViewVisibilityServiceInternal *,struct IMultitaskingViewGestureState *,struct IApplicationViewCollection *,struct Windows::Internal::ComposableShell::Tabs::ITabController *,struct ITabViewManager *,struct _GUID const &,void * *))"},
                 &CAltTabViewHost_CreateInstance_Win11_Original,
                 CAltTabViewHost_CreateInstance_Win11_Hook,
                 true,
```

mods/alt-tab-hide-ghost-windows.wh.cpp
```diff
@@ -0,0 +1,607 @@
...
+
+enum class WinVersion {
+    Unsupported,
+    Win10,
+    Win11,
+};
+
+WinVersion g_winVersion;
+
+std::atomic<DWORD> g_threadIdForAltTabShowWindow;
```
m417z, line 85: Looks like `g_threadIdForAltTabShowWindow` and the relevant code and hooks can be removed. They can also stay if you prefer to be more in sync with `alt-tab-per-monitor` to make further sync easier.

mods/alt-tab-hide-ghost-windows.wh.cpp
CherryDT: That's what I initially had, I just thought mentioning Groupy would be helpful. However I just realized search also works in description, so it will be fine.

mods/alt-tab-hide-ghost-windows.wh.cpp
CherryDT: You are right, these are from unrelated troubleshooting which turned out to be another thing entirely, but it worked either way... I reverted those changes now.

mods/alt-tab-hide-ghost-windows.wh.cpp
CherryDT: There are places where these are read, so I'd rather leave them... I have to say I do not 100% understand the `alt-tab-per-monitor` code, I just adjusted it to my needs.

mods/alt-tab-hide-ghost-windows.wh.cpp
m417z: You updated the name in the readme, but not the actual name in the metadata.

CherryDT: Oops. Fixed.

CherryDT:
> Doesn't seem to work for me, even when set to 100s of pixels.

What happens if you enable "Hide minimized windows"? Does that work? (I want to understand whether it works at all.)

What is your Windows build version?

CherryDT:
This mod is based on the [Alt+Tab per monitor mod by L3r0y](https://windhawk.net/mods/alt-tab-per-monitor), can you try if that mod also doesn't work with your setup?

CherryDT:
I don't have an external monitor here right now either, @L3r0yThingz could you maybe chime in on this, is your mod supposed to work with that virtual desktop option too?

PR #2520, opened 2025-10-07
Classic Minimize/Maximize Animations 1.0.0

aubymori:
From Windows 95 to XP, and in Vista and above without DWM running, windows had a minimize/maximize animation that involved the titlebar moving across the screen. This mod makes that animation play with DWM enabled by reimplementing it.

## Previews

**Top-level window**:
Top-level window

**MDI child window**:
MDI child window

### DIFF
--- a/mods/classic-min-max-animations.wh.cpp
+++ b/mods/classic-min-max-animations.wh.cpp
@@ -1,0 +1,887 @@
+// ==WindhawkMod==
+// @id              classic-min-max-animations
+// @name            Classic Minimize/Maximize Animations
+// @description     Restores the classic minimize/maximize animations used without DWM
+// @version         1.0.0
+// @author          aubymori
+// @github          https://github.com/aubymori
+// @include         *
+// @compilerOptions -lgdi32
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Classic Minimize/Maximize Animations
+From Windows 95 to XP, and in Vista and above without DWM running, windows had
+a minimize/maximize animation that involved the titlebar moving across the screen.
+This mod makes that animation play with DWM enabled by reimplementing it.
+
+## Usage note
+You must turn off the regular window animations in order for this mod to look correct.
+To do this:
+1. Run `sysdm.cpl` either from Start search or the Run dialog (WinKey + R).
+2. Go to the "Advanced" tab.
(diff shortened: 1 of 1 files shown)


aubymori:
Preferably before this gets merged, I need help on a weird esoteric bug. For some reason, some window will erroneously gain the `WS_EX_TOPMOST` extended style after being minimized/restored. This seems to happen outside of the animation code, so I can't easily detect it. I fixed it mostly by calling SetWindowPos on the animation window with `HWND_BOTTOM` as `hwndInsertAfter`, but it still happens in some programs, the only one I found so far being conhost. I haven't found a way to effectively get the animation window to display at the top without using the `WS_EX_TOPMOST` style either... I don't know why having this style is seemingly infectious to other windows. It's really strange.

mods/classic-min-max-animations.wh.cpp
```diff
@@ -0,0 +1,887 @@
...
+void Wh_ModUninit(void)
+{
+    if (g_hwndAnim && IsWindow(g_hwndAnim))
+    {
+        DestroyWindow(g_hwndAnim);
+        UnregisterClassW(c_szAnimClassName, g_hinst);
+    }
+    if (g_hAnimWndThread)
+    {
+        TerminateThread(g_hAnimWndThread, 0);
```
m417z: It's preferable to avoid calling `TerminateThread` if possible. It's not generally safe as the thread can be in a sensitive state, such as holding the process heap or some lock. Rather, signal the thread that it needs to exit and wait for its completion. Let me know if it's you feel it's not worth the effort and you'd rather keep it as is.

aubymori: It's probably not worth the effort. It'd be super hard for a user to disable the mod during the 0.25 second animation. However, I'd like to see if I can make a workaround for the topmost bug I mentioned in my other comment, so don't merge this just yet.

m417z: If it's a short-lived thread, why not just wait for it to complete instead of terminating it?

aubymori: You're probably right, actually. I imagine it exits shortly after I destroy the animation window anyway. I'll remove the termination code alongside adding the topmost hack.

aubymori: Both done.

PR #2521, opened 2025-10-07
Command Prompt Folder Larp 1.0

kawapure:
Make the current directory in Command Prompt appear as though it's a different folder.

This is an eyecandy tweak to make it seem like an older version of Windows, where, for example, the "Users" folder was known as "Documents and Settings".

This mod does not currently change the behaviour of `cd` or `dir` commands.

### DIFF
--- a/mods/cmd-folder-larp.wh.cpp
+++ b/mods/cmd-folder-larp.wh.cpp
@@ -207,7 +207,11 @@ void WINAPI SetConsoleTitleW_hook(LPCWSTR szTitle)
 const WindhawkUtils::SYMBOL_HOOK g_rghookCmd[] = {
     {
         {
+#ifdef _WIN64
             L"void __cdecl PrintPrompt(void)",
+#else
+            L"void __stdcall PrintPrompt(void)",
+#endif
         },
         &PrintPrompt_orig,
         PrintPrompt_hook,


mods/cmd-folder-larp.wh.cpp
```diff
@@ -0,0 +1,414 @@
...
+                // Append the rest of the source, including the leading "\"
+                wcscat_s(szBuffer, szRestOfSrc);
+            }
+
+            ZeroMemory(szTarget, nBufferLength);
+            wcscpy_s(szTarget, nBufferLength, szBuffer);
+            return true;
+        }
+        else if (pRedir->szFrom[0] == pRedir->szTo[0] == L'\0'
+            && pRedir->szFrom[1] == pRedir->szTo[1] == L'\0')
```
m417z: I assume you meant:
```cpp
else if (pRedir->szFrom[0] == L'\0' &&
         pRedir->szTo[0] == L'\0' &&
         pRedir->szFrom[1] == L'\0' &&
         pRedir->szTo[1] == L'\0') {
```

mods/cmd-folder-larp.wh.cpp
```diff
@@ -0,0 +1,414 @@
...
+            // End of array.
+            break;
+        }
+    }
+
+    return false;
+}
+
+DWORD ( *GetCurrentDirectoryW_orig)(DWORD nBufferLength, LPWSTR lpBuffer);
+DWORD GetCurrentDirectoryW_hook(DWORD nBufferLength, LPWSTR lpBuffer)
```
m417z: The calling convention is missing here, which is likely to cause a crash for 32-bit targets. Please add calling conventions or, if the mod doesn't support 32-bit, limit it to 64-bit, which will also make compilation faster. See:

mods/cmd-folder-larp.wh.cpp
```diff
@@ -0,0 +1,414 @@
...
+    Wh_Log(L"Entered");
+    if (g_fEnableGetCurrentDirectoryHooks)
+    {
+        WCHAR szBuffer[MAX_PATH] = { 0 };
+        GetCurrentDirectoryW_orig(ARRAYSIZE(szBuffer), szBuffer);
+
+        ApplyPathReplacementToString(szBuffer, ARRAYSIZE(szBuffer));
+
+        ZeroMemory(lpBuffer, nBufferLength);
+        wcscpy_s(lpBuffer, nBufferLength, szBuffer);
```
m417z, line 169: The process will crash if the buffer isn't large enough. You might want to have a more graceful handling, such as setting `ERROR_INSUFFICIENT_BUFFER` or giving up on the replacement.

mods/cmd-folder-larp.wh.cpp
```diff
@@ -0,0 +1,414 @@
...
+        rgMap[i].index = i;
+        rgMap[i].length = wcslen(pRedirectionsTemp[i].szFrom);
+    }
+
+    std::sort(rgMap, &rgMap[g_cRedirections], [](const SortMap &a, const SortMap &b)
+    {
+        return a.length > b.length;
+    });
+
+    g_rgRedirections = (PathRedirection *)malloc(sizeof(PathRedirection) * g_cRedirections);
```
m417z: Looks like it's never freed. Generally, using `std::vector` here will make the implementation simpler, and will also take care of the memory leaks. Note that even if you free `g_rgRedirections` in `Wh_ModUninit`, you'll still have a memory leak in the places where you `return FALSE` in `Wh_ModInit`.

mods/cmd-folder-larp.wh.cpp
kawapure: Are these not equivalent operations? I'll reformat it anyway, but I thought they were equivalent.

m417z: No.

<details><summary>AI explanation</summary>
<p>

> No. That expression is **not correct** in C or C++.
> 
> `==` does not chain. The statement
> 
> ```c
> pRedir->szFrom[0] == pRedir->szTo[0] == L'\0'
> ```
> 
> is parsed as
> 
> ```c
> ((pRedir->szFrom[0] == pRedir->szTo[0]) == L'\0')
> ```
> 
> which compares a Boolean (`0` or `1`) to a wide character constant — meaningless.
> 
> If you intend to test that both characters are `L'\0'`, write it explicitly:
> 
> ```c
> else if (pRedir->szFrom[0] == L'\0' &&
>          pRedir->szTo[0] == L'\0' &&
>          pRedir->szFrom[1] == L'\0' &&
>          pRedir->szTo[1] == L'\0') {
>     ...
> }
> ```
> 
> That version behaves as expected.
> 

</p>
</details>

mods/cmd-folder-larp.wh.cpp
kawapure: Oh, right. This would be

```cpp
        ZeroMemory(lpBuffer, nBufferLength);
        if (wcscpy_s(lpBuffer, nBufferLength, szBuffer) != 0)
        {
            return GetCurrentDirectoryW_orig(nBufferLength, lpBuffer);
        }
        return (DWORD)wcslen(lpBuffer);
```

correct?

m417z: No, as `wcscpy_s` will crash if `nBufferLength` isn't large enough. Also, `ZeroMemory` isn't really necessary here. You can do something like (untested):

```cpp
DWORD nResultLength = wcslen(szBuffer);
if (nResultLength + 1 > nBufferLength)
{
    return GetCurrentDirectoryW_orig(nBufferLength, lpBuffer);
}
wcscpy_s(lpBuffer, nBufferLength, szBuffer);
return nResultLength;
```

mods/cmd-folder-larp.wh.cpp
kawapure: Fixed in 6f00d16 (wait sorry the GitHub UI didn't refresh when this was marked as resolved so I tried it myself)

mods/cmd-folder-larp.wh.cpp
m417z: Also, how can this condition ever be true? Aren't `szFrom` and `szTo` always null-terminated strings? In which case if `szFrom[0]` is zero, it'd be incorrect to go on and look at `szFrom[1]`.

mods/cmd-folder-larp.wh.cpp
```diff
@@ -0,0 +1,421 @@
...
+    ApplyPathReplacementToString(szBuffer, ARRAYSIZE(szBuffer));
+
+    return SetConsoleTitleW_orig(szBuffer);
+}
+
+// cmd.exe
+const WindhawkUtils::SYMBOL_HOOK g_rghookCmd[] = {
+    {
+        {
+            L"void __cdecl PrintPrompt(void)",
```
m417z, line 211: 32-bit seems to be using `void __stdcall PrintPrompt(void)`. Either adapt the mod to 32-bit, or limit it to 64-bit in the metadata.

mods/cmd-folder-larp.wh.cpp
kawapure: My rationale was that they're multibyte strings; two 00 bytes for a terminator. I missed the part where I'm messing with 16-bit numbers instead of 8-bit numbers somehow.

PR #2522, opened 2025-10-08
Custom Command Prompt Startup Text 1.0.0

aubymori:
This mod allows you to change the version and copyright text displayed at the startup of Command Prompt without messing up scripts that use the `ver` command to read OS version, and without modifying the `cmd.exe` binary on disk.

### DIFF
--- a/mods/custom-cmd-startup-text.wh.cpp
+++ b/mods/custom-cmd-startup-text.wh.cpp
@@ -1,0 +1,222 @@
+// ==WindhawkMod==
+// @id              custom-cmd-startup-text
+// @name            Custom Command Prompt Startup Text
+// @description     Replace the version and copyright text that shows 
+// @version         1.0.0
+// @author          aubymori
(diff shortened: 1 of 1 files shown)


mods/custom-cmd-startup-text.wh.cpp
```diff
@@ -0,0 +1,222 @@
...
+    }
+}
+
+void PrintModError(LPCWSTR pszFormat, ...)
+{
+    LPCWSTR c_szErrorHeader = L"Error in Windhawk mod Custom Command Prompt Startup Text:\r\n";
+    CmdPutChars_orig(c_szErrorHeader, wcslen(c_szErrorHeader));
+
+    va_list args;
+    va_start(args, pszFormat);
```
m417z, line 85: `va_end` is missing.

aubymori: Fixed.

PR #2528, opened 2025-10-09
Create disable-voice-access-hotkey.wh.cpp

clemmyn23:
This mod disables the Voice Access (Win+Ctrl+S) hotkey in Windows.

It's weird I can't disable the hotkey in Windows settings. 
https://support.microsoft.com/en-us/topic/use-voice-access-to-control-your-pc-author-text-with-your-voice-4dcd23ee-f1b9-4fd1-bacc-862ab611f55d
image

### DIFF
--- a/mods/disable-voice-access-hotkey.wh.cpp
+++ b/mods/disable-voice-access-hotkey.wh.cpp
@@ -17,8 +17,10 @@ This mod disables the Voice Access (Win+Ctrl+S) hotkey in Windows.
 
 #include <windhawk_utils.h>
 
-BOOL WINAPI (*pOriginalRegisterHotKey)(HWND hWnd, int id, UINT fsModifiers, UINT vk);
-BOOL WINAPI RegisterHotKeyHook(HWND  hWnd, int id, UINT fsModifiers, UINT vk)
+using RegisterHotKey_t = decltype(&RegisterHotKey);
+RegisterHotKey_t RegisterHotKey_Original;
+
+BOOL WINAPI RegisterHotKey_Hook(HWND  hWnd, int id, UINT fsModifiers, UINT vk)
 {
     if (fsModifiers == (MOD_CONTROL | MOD_WIN | MOD_NOREPEAT) && vk == 'S')
     {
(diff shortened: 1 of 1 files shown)


mods/disable-voice-access-hotkey.wh.cpp
```diff
@@ -0,0 +1,38 @@
...
+
+// ==WindhawkModReadme==
+/*
+# Disable Voice Access Hotkey
+This mod disables the Voice Access (Win+Ctrl+S) hotkey in Windows.
+*/
+// ==/WindhawkModReadme==
+
+BOOL(*pOriginalRegisterHotKey)(HWND hWnd, int id, UINT fsModifiers, UINT vk);
+BOOL RegisterHotKeyHook(HWND hWnd, int id, UINT fsModifiers, UINT vk)
```
m417z: The calling convention is missing here, which is likely to cause a crash for 32-bit targets. Please add calling conventions. See:

mods/disable-voice-access-hotkey.wh.cpp
```diff
@@ -0,0 +1,40 @@
...
+
+    return pOriginalRegisterHotKey(hWnd, id, fsModifiers, vk);
+}
+
+
+BOOL Wh_ModInit() {
+    HMODULE hUser32 = GetModuleHandle(L"user32.dll");
+
+    void* origFunc = (void*)GetProcAddress(hUser32, "RegisterHotKey");
+    WindhawkUtils::SetFunctionHook(origFunc, (void*)RegisterHotKeyHook, (void**)&pOriginalRegisterHotKey);
```
m417z: The idea of `WindhawkUtils::SetFunctionHook` is that you can remove the `void*` hooks, and the compiler makes sure that the signature of the hook is compatible with the signature of the hooked function. See https://github.com/ramensoftware/windhawk/wiki/Development-tips#hooks-and-calling-conventions.

You can do something like:

```suggestion
    auto origFunc = (decltype(&RegisterHotKey))GetProcAddress(hUser32, "RegisterHotKey");
    WindhawkUtils::SetFunctionHook(origFunc, RegisterHotKeyHook, &pOriginalRegisterHotKey);
```

clemmyn23: 🙇 thank you. 

Been years since I've touched cpp code my bad about this sloppiness 😅

m417z:
Thanks! Congratulations on your first mod :)

PR #2548, opened 2025-10-14
Add Explorer Font Changer mod

cristeigabriela:
This mod allows you to intercept all GDI font handle creation, and we can change the font face name to anything we desire at that time. [Came from a stupid idea that I did quickly in WinDbg a while ago](https://bsky.app/profile/cristei.bsky.social/post/3lzmqhdlhc22q), but found myself wanting this in practice.

image
Using Berkeley Mono Variable as an example font.

### DIFF
--- a/mods/explorer-font-changer.wh.cpp
+++ b/mods/explorer-font-changer.wh.cpp
@@ -6,7 +6,6 @@
 // @author          Gabriela Cristei
 // @github          https://github.com/cristeigabriela
 // @include         explorer.exe
-// @compilerOptions -std=c++23
 // ==/WindhawkMod==
 
 // ==WindhawkModReadme==
@@ -49,15 +48,11 @@ https://bsky.app/profile/cristei.bsky.social/post/3lzmqhdlhc22q
 // If you're new to terms such as code injection and function hooking, the
 // article is great to get started.
 
-#include <memory>
 #include <string>
+#include <windhawk_utils.h>
 
 using namespace std::string_view_literals;
-
-// Manage setting resource using `std::unique_ptr`.
-using string_setting_raii_t = std::unique_ptr<
-    const WCHAR, decltype(&Wh_FreeStringSetting)
->;
+using namespace WindhawkUtils;
 
 using create_font_indirectw_hook_t = decltype(&CreateFontIndirectW);
 create_font_indirectw_hook_t create_font_indirectw_original = nullptr;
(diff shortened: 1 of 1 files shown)


mods/explorer-font-changer.wh.cpp
```diff
@@ -0,0 +1,95 @@
...
+// article of Kyle Halladay, X64 Function Hooking by Example:
+// http://kylehalladay.com/blog/2020/11/13/Hooking-By-Example.html
+// If you're new to terms such as code injection and function hooking, the
+// article is great to get started.
+
+using CreateFontIndirectWHook_t = decltype(&CreateFontIndirectW);
+CreateFontIndirectWHook_t pOriginal = nullptr;
+HFONT __stdcall CreateFontIndirectWHook(const LOGFONTW* font) {
+    HWND currWindow = GetActiveWindow();
+    // TODO(gabriela): localize strings or at least check by icon instead.
```
m417z: If that limitation means that it will crash on non-English OSes, then perhaps it's better to find a solution to it before publishing it.

mods/explorer-font-changer.wh.cpp
```diff
@@ -0,0 +1,95 @@
...
+    // this will only work for english!
+    char title[300] = {0};
+    if (GetWindowTextA(currWindow, &title[0], sizeof(title)) != 0
+            && strstr(title, "Run") == title) {
```
m417z: Why use `strstr` and not `wcscmp` here and below?

mods/explorer-font-changer.wh.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+Basically, restart your Explorer first for this to take effect.
+
+# Disabling Explorer Font Changer
+Simply change the font name to "None", or turn off the plugin.
+
+### Formerly did this with WinDbg for a joke, but liked it
+https://bsky.app/profile/cristei.bsky.social/post/3lzmqhdlhc22q
+*/
+// ==/WindhawkModReadme==
```
m417z, line 30: How about adding a screenshot to the readme?

mods/explorer-font-changer.wh.cpp
```diff
@@ -46,50 +49,65 @@ Simply change the font name to "None", or turn off the plugin.
...
-    }
+#include <memory>
+#include <string>
+
+using namespace std::string_view_literals;
+
+// Manage setting resource using `std::unique_ptr`.
+using string_setting_raii_t = std::unique_ptr<
+    const WCHAR, decltype(&Wh_FreeStringSetting)
+>;
```
m417z: This is fine, but note that Windhawk comes with a `WindhawkUtils::StringSetting` util in `windhawk_utils.h`. See:

mods/explorer-font-changer.wh.cpp
```diff
@@ -6,12 +6,15 @@
 // @author          Gabriela Cristei
 // @github          https://github.com/cristeigabriela
 // @include         explorer.exe
+// @compilerOptions -std=c++23
```
m417z: Why is it needed? C++23 is already the default in recent Windhawk versions.

cristeigabriela: oops! thank you!

mods/explorer-font-changer.wh.cpp
cristeigabriela: oh cool! thanks!

PR #2561, opened 2025-10-19
Implement GPU usage metric support in taskbar clock

neptotech:
Added GPU usage metrics to taskbar clock customization.

image

I would always love to monitor metrics from tooltip instead of me having to open task manager. Thus added GPU metric along with existing CPU, RAM. 

Just thought to share it(and merge if approved).

@m417z Can you consider this(Didn't know how to submit it to you, I have read only original author can edit their mods..)


m417z:
Thanks for the contribution. You can find the repo for pull requests in the mod comments:

https://github.com/ramensoftware/windhawk-mods/blob/a1e5471c16c55ce3827e8f85367aafa200d0e69a/mods/taskbar-clock-customization.wh.cpp#L17-L21

PR #2567, opened 2025-10-20
Add Close Explorer on Esc mod

lieyanbang:
### Mod Name
Close Explorer on Esc

### Author
[@lieyanbang](https://github.com/lieyanbang)

### Description
Press Esc in File Explorer to close the current window safely.  
Ignores rename, address bar, and text inputs. Optional IME compatibility.

### Tags
explorer, esc, keyboard, usability, window

### Version
1.1

### DIFF
--- a/mods/close-explorer-on-esc.wh.cpp
+++ b/mods/close-explorer-on-esc.wh.cpp
@@ -19,7 +19,6 @@
 Press **Esc** in File Explorer to close the current window.
 The mod avoids accidental closure when:
 - You are renaming a file/folder
-- Focus is in the address bar, search box, or any text field
 - The foreground window is not the current Explorer
 
 Optional:
@@ -36,7 +35,6 @@ If the focused control is safe (not text editing, not renaming), it sends `WM_CL
 在资源管理器中按下 **Esc** 可关闭当前窗口。
 模块会避免误关以下场景：
 - 正在**重命名**
-- 焦点在**地址栏**、**搜索框**或**任何文本输入框**
 - 当前前台窗口不是该资源管理器实例
 
 可选项：
@@ -164,6 +162,11 @@ static DWORD WINAPI HookThreadProc(LPVOID) {
     g_hook = SetWindowsHookExW(WH_KEYBOARD_LL, LowLevelKeyboardProc, hMod, 0);
     if (g_hookReadyEvent) SetEvent(g_hookReadyEvent);
 
+    if (!g_hook) {
+        g_hookThreadId = 0;
+        return 0;
+    }
+
     // Message loop required by WH_KEYBOARD_LL delivery.
     MSG msg;
     while (GetMessageW(&msg, nullptr, 0, 0) > 0) {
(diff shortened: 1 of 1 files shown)


mods/close-explorer-on-esc.wh.cpp
```diff
@@ -0,0 +1,170 @@
...
+
+// --- Windhawk entry points ---
+
+static void LoadSettings() {
+    g_allowImeCandidate = Wh_GetIntSetting(L"allowWhenImeCandidate") != 0;
+}
+
+BOOL Wh_ModInit() {
+    LoadSettings();
+    g_hook = SetWindowsHookExW(WH_KEYBOARD_LL, LowLevelKeyboardProc, GetModuleHandleW(nullptr), 0);
```
m417z: From [the documentation](https://learn.microsoft.com/en-us/windows/win32/winmsg/lowlevelkeyboardproc):

m417z (commented):
Thanks for the submission.

mods/close-explorer-on-esc.wh.cpp
lieyanbang: Thanks for the review and the pointer! You’re right — per the docs, `WH_KEYBOARD_LL` is delivered to the installing thread via its message queue, so the installing thread must have a message loop. `Wh_ModInit` isn’t guaranteed to run on a thread with a message loop.

I’ve updated the mod to install the low-level keyboard hook from a dedicated thread that runs a message loop (similar to the example you linked). On unload, the mod posts `WM_QUIT` to that thread and waits for it to exit cleanly. I also obtain the module handle for the hook proc with `GetModuleHandleEx` to pass the correct `HMODULE` to `SetWindowsHookEx`.

Changes:

- Create a dedicated hook thread with `GetMessage` loop.

- Use `GetModuleHandleEx(..., &LowLevelKeyboardProc, ...)` for the `hMod` parameter.

- Graceful teardown: `PostThreadMessage(WM_QUIT)` + `WaitForSingleObject`.

- Bumped version to 1.2.

Let me know if you’d like me to split this into smaller commits or adjust any style details.

mods/close-explorer-on-esc.wh.cpp
```diff
@@ -145,26 +152,69 @@ static LRESULT CALLBACK LowLevelKeyboardProc(int code, WPARAM wParam, LPARAM lPa
...
+        g_hookReadyEvent = nullptr;
+        return FALSE;
+    }
+
+    // Wait for the hook attempt to complete (install success or fail).
+    WaitForSingleObject(g_hookReadyEvent, 5000);
+    CloseHandle(g_hookReadyEvent);
+    g_hookReadyEvent = nullptr;
+
     return g_hook != nullptr;
```
m417z: If `g_hook` is `nullptr`, the mod will be unloaded while the thread is still running. In this case, you can exit the thread right away, and wait for its completion in `Wh_ModInit` before returning `FALSE`.

lieyanbang: Thanks for the follow-up!

Good point — if `SetWindowsHookEx` fails and `g_hook` is null, the hook thread would still
be running when the mod is unloaded. I've updated the code to handle this case properly.

**Changes made:**
- If `g_hook` is null after attempting to install the hook, the thread now exits immediately.
- `Wh_ModInit` waits for the hook thread to finish before returning `FALSE`.
- This ensures no race condition or dangling thread during mod unload.

Appreciate the detailed feedback and review!

PR #2570, opened 2025-10-20
Add Fully Customizable WinVer mod

ItsTauTvyDas:
This mod lets you fully customize your winver.exe, there are similar mods but they did not meet my needs as they weren't customizable enough.

It might be too much for such application like winver.exe but I want my Windows customized and so decided to make a mod like this.

## Features
* Modify any text
* Custom logo (it just creates a custom child window in the logo area, transparency is supported!)
* Change the title
* Change window parameters (make it always top, custom size, open location)
* Change background colors

Custom logo also supports animated gif files, with transparency too

## Screenshots
Screenshot 2025-10-20 17-37-24

ezgif-848f4f2e0b4006

ezgif-883355eb4a8815

### DIFF
--- a/mods/fully-customizable-winver.wh.cpp
+++ b/mods/fully-customizable-winver.wh.cpp
@@ -1,5 +1,5 @@
 // ==WindhawkMod==
-// @id              fully-custom-winver
+// @id              fully-customizable-winver
 // @name            Fully Customizable WinVer
 // @description     Fully customize your winver.exe by even "replacing" the logo!
 // @version         1.0.0


mods/fully-customizable-winver.wh.cpp
```diff
@@ -0,0 +1,1260 @@
+// ==WindhawkMod==
+// @id              fully-customizable-winver
+// @name            Fully Customizable WinVer
+// @description     Fully customize your winver.exe by even "replacing" the logo!
+// @version         1.0.0
+// @author          ItsTauTvyDas
+// @github          https://github.com/ItsTauTvyDas
+// @twitter         https://x.com/ItsTauTvyDas
```
m417z, line 8: To verify your X (Twitter) account, please send me (https://x.com/m417z) a direct message with the following content:

> I attest that I'm the sole owner of both this Twitter account (https://x.com/ItsTauTvyDas) and the following GitHub account: https://github.com/ItsTauTvyDas

mods/fully-customizable-winver.wh.cpp
```diff
@@ -0,0 +1,1260 @@
...
+#define LOGO_BACKGROUND_ID 100
+#define LOGO_IMAGE_ID 101
+#define MAX_LOGO_HEIGHT 72
+#define DEFAULT_BAR_BACKGROUND RGB(241, 243, 249)
+#define DEFAULT_BACKGROUND RGB(240, 240, 240)
+#define DEFAULT_TEXT_COLOR RGB(0, 0, 0)
+
+#define WM_CAPTURE_LOGO_BACKGROUND WM_USER + 1
+
+#ifndef WH_EDITING
```
m417z: Why is `WH_EDITING` needed here and below?

mods/fully-customizable-winver.wh.cpp
```diff
@@ -0,0 +1,1260 @@
...
+    if (g_logoImage) {
+        DeleteLogoImage();
+        ShowWindow(g_logoImage, SW_HIDE);
+    }
+
+    g_canAnimatedDrawLogo = FALSE;
+    PostMessageW(hWnd, WM_CAPTURE_LOGO_BACKGROUND, 0, 0);
+
+    g_clearLogoBackground = Wh_GetIntSetting(L"logo.clearBackground");
+    Wh_Log(L"Clear background on every gif frame -> %d", g_clearLogoBackground);
```
m417z: This setting seems to be unused.

mods/fully-customizable-winver.wh.cpp
```diff
@@ -0,0 +1,1260 @@
...
+            }
+            MoveWindow(g_logoImage, rect.left, rect.top, rect.right, rect.bottom, TRUE);
+            PrintRect(L"Logo position", rect);
+        } else {
+            Wh_Log(L"Failed to load logo: %ls", GetLastError());
+        }
+    }
+    g_logoImageRect = rect;
+    g_logoImageRect.right += g_logoImageRect.left;
+    g_logoImageRect.bottom += g_logoImageRect.top;
```
m417z: Seems to be unused.

mods/fully-customizable-winver.wh.cpp
```diff
@@ -0,0 +1,1260 @@
...
+}
+
+void Wh_ModUninit() {
+    Wh_Log(L"Uninit");
+    if (g_hHook) {
+        UnhookWinEvent(g_hHook);
+    }
+    // Partially reset some stuff
+    if (IsWindow(g_logoImage)) {
+        RemoveWindowSubclass(g_logoImage, WndSubclassProc, 1);
```
m417z: `RemoveWindowSubclass` can't be called from `Wh_ModUninit` as it doesn't work across threads. I didn't try the mod, but I assume that it will cause problems with mod unload.

You can use `WindhawkUtils::SetWindowSubclassFromAnyThread` instead, see:

mods/fully-customizable-winver.wh.cpp
```diff
@@ -0,0 +1,1260 @@
...
+    if (!fullBmp) {
+        ReleaseDC(g_mainWindow, hdc);
+        return nullptr;
+    }
+
+    HDC fullDC = CreateCompatibleDC(hdc);
+    HGDIOBJ oldFull = SelectObject(fullDC, fullBmp);
+
+    BOOL ok = FALSE;
+#if defined(PW_RENDERFULLCONTENT)
```
m417z: Why is this necessary? If it means that different code will be generated for different Windhawk versions, then just define `PW_RENDERFULLCONTENT` if it's not defined, like:
```cpp
#ifndef PW_RENDERFULLCONTENT
#define PW_RENDERFULLCONTENT 0x00000002
#endif
```

mods/fully-customizable-winver.wh.cpp
ItsTauTvyDas: I remember the winver.exe crashing (freezing) and then I tried removing the subclass and it worked, it doesn't crash or anything

m417z: Maybe `RemoveWindowSubclass` works in practice (`SetWindowSubclass` surely doesn't work across threads), but it's documented to not be supported, so I'd still change it. Switching it to use the `WindhawkUtils` variant shouldn't be difficult.

ItsTauTvyDas: I now used the one from utils class but I still remove the subclass on uninit (or should I do before uninit?) because the window still freezes if I disable the mod

m417z: If you switched to `WindhawkUtils::SetWindowSubclassFromAnyThread`, you should call `WindhawkUtils::RemoveWindowSubclassFromAnyThread` before the mod is unloaded. Calling it in `Wh_ModUninit` as you do now is fine.

mods/fully-customizable-winver.wh.cpp
ItsTauTvyDas: I read a little more and PW_RENDERFULLCONTENT was first introduced in win 8.1 so i will just delete that if check

mods/fully-customizable-winver.wh.cpp
ItsTauTvyDas: relic from the past 😁

ItsTauTvyDas:
I think i fixed now everything, made some improvements too

m417z:
Thanks! Congratulations on your first mod :)

PR #2587, opened 2025-10-23
Explorer Font Changer 0.2: Live font changes, glowing text option.

cristeigabriela:
These changes implement the following functionality:

- Live font changes:
Through hooking into `DrawText[Ex]W`, we can now change fonts at the time of rendering, instead of depending on intercepting font creation!

- Glow text:
This toggles on the "Glow text" functionality, proxying calls to `user32!DrawTextW` and `user32!DrawTextExW` to `uxtheme.dll`'s "DrawGlowText" function. Thanks to the mod "Translucent Windows" for bringing this function to my attention!

- Glow: Red (0-255):
The red value of the glow text color (0-255).

- Glow: Green (0-255):
The green value of the glow text color (0-255).

- Glow: Blue (0-255):
The blue value of the glow text color (0-255).

- Glow: Alpha (0-255):
The alpha value of the glow text color (0-255).

- Glow: Intensity (1-..):
The intensity value of the glow that surrounds the text.

### DIFF
--- a/mods/explorer-font-changer.wh.cpp
+++ b/mods/explorer-font-changer.wh.cpp
@@ -74,6 +74,7 @@ This can be checked by others by going to System Informer, looking for `explorer
 */
 // ==/WindhawkModSettings==
 
+
 #include <string>
 #include <windhawk_utils.h>
 #include <uxtheme.h>
@@ -82,6 +83,14 @@ using namespace std::string_view_literals;
 using namespace WindhawkUtils;
 
 namespace util {
+    StringSetting s_font_name;
+    bool s_glow = false;
+    int s_glow_r = 255;
+    int s_glow_g = 255;
+    int s_glow_b = 255;
+    int s_glow_a = 255;
+    int s_glow_intensity = 0;
+
     // Class to declare transparent RAII types.
     template <class T, auto Fn>
     class raii {
@@ -111,10 +120,10 @@ namespace util {
 
     // Alters search path to look in `System32` folder.
     HMODULE load_os_library(
-        const std::wstring_view library
+        LPCWSTR library
     ) {
         return LoadLibraryExW(
-            library.data(),
+            library,
             nullptr,
             LOAD_LIBRARY_SEARCH_SYSTEM32
         );
@@ -194,8 +203,7 @@ namespace util {
     void change_font_in_struct(
         LOGFONTW* font
     ) {
-        auto set_font_name = StringSetting::make(L"font.name");
-        auto font_name = std::wstring_view(set_font_name.get());
+        auto font_name = std::wstring_view(s_font_name.get());
 
         // Check font configuration.
         if (font_name != L"None"sv) {
@@ -214,9 +222,7 @@ namespace util {
         }
     }
 
-    // NOTE: upon destruction of `unique_hfont_t`, the handle will be sent to
-    // `DestroyObject`.
-    unique_hfont_t hdc_update_font(
+    std::pair<unique_hfont_t, LOGFONTW> hdc_update_font(
         HDC hdc
     ) {
         // Get current selected font.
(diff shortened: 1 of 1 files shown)


mods/explorer-font-changer.wh.cpp
```diff
@@ -37,30 +52,104 @@ Simply change the font name to "None", or turn off the plugin.
...
+    // as letting it fall back on, for example, `DrawTextW`, while
+    // using this inside said function, would cause recursion.
+    DTT_CALLBACK_PROC fallback_draw_text,
+    LPARAM lparam
+);
+
+// Function is not exported by name, but it is found in the resource table.
+static auto draw_text_with_glow_fn = reinterpret_cast<draw_text_with_glow_t>(
+    GetProcAddress(
+        GetModuleHandleW(L"uxtheme.dll"),
```
m417z: Explorer has uxtheme.dll in its import table, but for other apps, perhaps it's worth checking for its presence or just using `LoadLibrary` (actually, [`LoadLibraryEx`](https://github.com/ramensoftware/windhawk-mods/issues/2063)). WDYT?

cristeigabriela: good call! pushed some changes that implements this :-)

mods/explorer-font-changer.wh.cpp
```diff
@@ -37,64 +56,317 @@ Simply change the font name to "None", or turn off the plugin.
...
-    if (font_name != L"None"sv) {
-        if (font_name.size() <= 31) {
-            // `face_name` points to a total of 32 WORDs.
-            auto face_name = static_cast<WCHAR*>(font->lfFaceName);
+    // Alters search path to look in `System32` folder.
+    HMODULE load_os_library(
+        const std::wstring_view library
+    ) {
+        return LoadLibraryExW(
+            library.data(),
```
m417z: This is correct for all calls of `load_os_library`, but as a general note, `std::wstring_view` is not guaranteed to be null-terminated. I'd have the function receive a `const WCHAR*` instead.

mods/explorer-font-changer.wh.cpp
```diff
@@ -37,64 +56,317 @@ Simply change the font name to "None", or turn off the plugin.
...
+        auto get() const {
+            return m_handle;
+        }
+
+    private:
+        T m_handle;
+    };
+
+    // RAII `HFONT`.
+    using unique_hfont_t = raii<HFONT, DeleteObject>;
```
m417z, line 119: FYI you can also use [WIL](https://github.com/microsoft/wil) which is included with Windhawk. It has types such as [`unique_hfont`](https://github.com/search?q=repo%3Amicrosoft%2Fwil+unique_hfont&type=code). Currently there's [one known issue](https://github.com/ramensoftware/windhawk/issues/553) with it but there's also a workaround.

mods/explorer-font-changer.wh.cpp
```diff
@@ -37,64 +56,317 @@ Simply change the font name to "None", or turn off the plugin.
...
+                );
+            } else {
+                Wh_Log(L"Trying to change font to \"%s\": size too long (%d)", font_name.data(), font_name.size());
+            }
         }
     }
 
-    return create_font_indirectw_original(_font);
+    // NOTE: upon destruction of `unique_hfont_t`, the handle will be sent to
+    // `DestroyObject`.
```
m417z: :)
```suggestion
    // `DeleteObject`.
```

mods/explorer-font-changer.wh.cpp
```diff
@@ -37,64 +56,317 @@ Simply change the font name to "None", or turn off the plugin.
...
+        auto h_new_font = CreateFontIndirectW(&font);
+
+        // Select font.
+        SelectObject(hdc, h_new_font);
+
+        return h_new_font;
+    }
+
+    bool is_glow_enabled() {
+        return Wh_GetIntSetting(L"font.glow") == 1;
```
m417z: In case these functions are called a lot, you might want to cache the settings values, and update them when `Wh_ModSettingsChanged` is called.

mods/explorer-font-changer.wh.cpp
```diff
@@ -37,64 +56,317 @@ Simply change the font name to "None", or turn off the plugin.
...
+        auto text = std::wstring_view{ lpchText, static_cast<size_t>(cchText) };
+        auto hr = util::draw_text_with_glow(
+            hdc, lprc, text, format, GetTextColor(hdc), util::get_glow_abgr(), util::get_glow_intensity(),
+            [](HDC hdc, LPWSTR lpchText, INT cchText, LPRECT lprc, UINT format, LPARAM lParam) WINAPI {
+                return draw_textw_original(hdc, lpchText, cchText, lprc, format);
+            }
+        );
+
+        if (SUCCEEDED(hr)) {
+            return TRUE;
```
m417z: The function is supposed to return the height of the text. Same for `draw_text_exw_hook` below.

m417z (commented):
Looks good. Most comments are minor/FYI but the return value in the hooks might be worth fixing.

mods/explorer-font-changer.wh.cpp
cristeigabriela: Hi! I haven't seen any examples of it being used, which is why I went along with Translucent Windows' decision to just return `TRUE` in this scenario.

Unfortunately, DrawTextWithGlow from uxtheme.dll doesn't return the text height, but rather a HRESULT. Frankly, there may be something worth looking for in `rect` re. this, but I haven't spent time looking at this.

Before making my choice, I considered multple things:
- I checked DrawTextExWorker in user32.dll, the logic to replicate the height calculation by the end was cumbersome, and it would require unexported functions.
- I considered GetTextMetrics.
- I considered GetTextExtentPoint32W.
- I considered returning the `LOGFONTW` and using lfHeight`.

None of the above replicated the expected result, so instead, I decided to just return `draw_text[_ex]w_original` with `format | DT_CALCRECT` so that it would return the appropriate height.

mods/explorer-font-changer.wh.cpp
cristeigabriela: Of course! Sorry, I've been working with Rust at my job, and we lift all C-strs to Rust's String, so I've had a tendency to try and use string_views. I know that they intentionally don't provide `c_str()` for the reasons you also mention above, but I've yet in my life to see a string_view that doesn't point to a non-nul terminated array (except through library interop)

mods/explorer-font-changer.wh.cpp
cristeigabriela: Seems like a huge headache for now, honestly

mods/explorer-font-changer.wh.cpp
cristeigabriela: oops :)

mods/explorer-font-changer.wh.cpp
cristeigabriela: Well, they're not really called a lot as a lot of the rendering is retained, but it's a better idea to only get them when need be.

cristeigabriela:
implemented most of your requested changes! i'll get to the changelog once i'm back home, let me know if you need any more changes!

PR #2623, opened 2025-10-29
Color Picker Dialog Hex Input 1.0.0

aubymori:
Windows' native color picker dialog does not come with a way to input hex codes. This mod adds an input that allows you to type hex colors in both the standard 6-digit format and the CSS 3-digit format (for example, `ABC` would result in `AABBCC`).

**Preview**:
Preview

### DIFF
--- a/mods/color-picker-dialog-hex-input.wh.cpp
+++ b/mods/color-picker-dialog-hex-input.wh.cpp
@@ -180,8 +180,8 @@ LRESULT CALLBACK HexInputSubclassProc(
                             if (!IsGoodCharacter(*psz))
                             {
                                 MessageBeep(MB_OK);
-                                CloseClipboard();
                                 GlobalUnlock(hData);
+                                CloseClipboard();
                                 return 0;
                             }
                         }


mods/color-picker-dialog-hex-input.wh.cpp
```diff
@@ -0,0 +1,563 @@
...
+    ))
+    {
+        Wh_Log(L"Failed to hook one or more symbol functions in comdlg32.dll");
+        fSucceeded = FALSE;
+        goto Exit;
+    }
+
+Exit:
+    if (hGdi32)
+        FreeLibrary(hGdi32);
```
m417z, line 560: It's being freed even on a successful init, which seems like a mistake.

mods/color-picker-dialog-hex-input.wh.cpp
```diff
@@ -0,0 +1,563 @@
...
+            }
+            break;
+        }
+        case WM_PASTE:
+            if (OpenClipboard(hwnd))
+            {
+                HANDLE hData = GetClipboardData(CF_UNICODETEXT);
+                if (hData)
+                {
+                    LPCWSTR pszText = (LPCWSTR)GlobalLock(hData);
```
m417z, line 175: `GlobalUnlock` is missing.

mods/color-picker-dialog-hex-input.wh.cpp
```diff
@@ -0,0 +1,562 @@
...
+                    if (pszText)
+                    {
+                        for (LPCWSTR psz = pszText; *psz; psz++)
+                        {
+                            if (!IsGoodCharacter(*psz))
+                            {
+                                MessageBeep(MB_OK);
+                                CloseClipboard();
+                                GlobalUnlock(hData);
```
m417z, line 183: Not sure if that's strictly incorrect, but all samples call GlobalUnlock first, then CloseClipboard.

PR #2633, opened 2025-10-30
Add Windows 11 Custom Title Bar Colours mod

Th3Fanbus:
This is a mod I made for myself to adjust the colours of title bars. My first Windhawk mod. It is also the first time I use Windows APIs, so I probably made mistakes somewhere. But the mod does work.

### DIFF
--- a/mods/win11-custom-title-bar-colours.wh.cpp
+++ b/mods/win11-custom-title-bar-colours.wh.cpp
@@ -1,0 +1,218 @@
+// ==WindhawkMod==
+// @id              win11-custom-title-bar-colours
+// @name            Windows 11 Custom Title Bar Colours
+// @description     Allows changing the colour of active/inactive title bars, as well as enabling/disabling Immersive Dark Mode.
+// @version         1.0.0
+// @author          Th3Fanbus
+// @github          https://github.com/Th3Fanbus
+// @include         *
+// @exclude         devenv.exe
+// @compilerOptions -ldwmapi -luser32
+// ==/WindhawkMod==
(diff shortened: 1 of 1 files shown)


Th3Fanbus:
> thanks @Th3Fanbus ! do you think it would be possible to bring to Windhawk a mod that resumes the work of the abandoned [translucentflyouts](https://github.com/ALTaleX531/TranslucentFlyouts)?

If you're asking _me_ to revive that project, definitely no - I have next to no knowledge about Windows APIs.

maxoku:
@Th3Fanbus 
It's a great mod and something I was looking for. Unfortunately it doesn't work properly in every situation.
1. "Use Immersive Dark Mode" changes the mode for some apps. If the option is chosen then apps that don't support dark mode will have forced dark background but text will still stay dark so it become unreadable. If option is disabled then the opposite happens to the apps with dark theme. They have background forced to white, while the text stays bright as well.
Couldn't it be make neutral to not affect the theme? If not then maybe an ability exclude for that option, so chosen apps would have opposite setting?
2. It doesn't fill the whole title bar of some apps which have title bars with more height. E.g. Task Manager, Windows 11 File Explorer and Notepad with tabs.
3. File Explorer windows changes it's color from inactive to active only when clicked on a tab. It doesn't when clicked on title bar or anywhere inside the window.
4. Doesn't work on all windows, especially Windows settings windows, web browsers and PWAs.
5. It doesn't change the text color of title bar, so it might not fit to the changed color. An option to affect text color would be nice too.

PR #2636, opened 2025-10-30
Add Toggle Hidden Files mod

Asteski:
# Toggle Hidden Files

This mod allows you to toggle the visibility of hidden files in Windows Explorer using the Ctrl+H keyboard shortcut.

## Features
- Ctrl+H hotkey that works only when Explorer windows are focused
- Toggles the "Show hidden files" setting
- Optional: Also toggle protected OS files
- Automatically refreshes Explorer windows
- Works with all Windows Explorer windows

## Usage
1. **Focus an Explorer window** - Click on or open any File Explorer window
2. **Press Ctrl+H** - Use the keyboard shortcut to toggle hidden files visibility
3. **The setting will be applied immediately** to all Explorer windows

## Settings
- **Also toggle protected OS files**: When enabled, Ctrl+H will also show/hide protected operating system files

## Technical Details
- Only activates when Windows Explorer windows are in focus
- Modifies the standard registry settings for showing hidden files
- Sends refresh messages to all Explorer windows
- Handles proper cleanup when the mod is unloaded
- Explorer process must be restarted for changes to take effect

### DIFF
--- a/mods/toggle-hidden-files.wh.cpp
+++ b/mods/toggle-hidden-files.wh.cpp
@@ -1,0 +1,304 @@
+// ==WindhawkMod==
+// @id           toggle-hidden-files
+// @name         Toggle Hidden Files
+// @description  Toggle the visibility of hidden files in Windows Explorer using Ctrl+H
+// @version      1.0.0
+// @author       Asteski
+// @github       https://github.com/Asteski
+// @include      explorer.exe
+// @compilerOptions -std=c++20
+// ==/WindhawkMod==
+
+// ==WindhawkModSettings==
+/*
+- toggleProtectedFiles: true
+  $name: Also toggle protected OS files
+  $description: When enabled, Ctrl+H will also toggle the visibility of protected operating system files
+*/
+// ==/WindhawkModSettings==
+
+// ==WindhawkModReadme==
+/*
+# Toggle Hidden Files
+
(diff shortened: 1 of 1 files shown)


mods/toggle-hidden-files.wh.cpp
```diff
@@ -0,0 +1,304 @@
+// ==WindhawkMod==
+// @id           toggle-hidden-files
+// @name         Toggle Hidden Files
+// @description  Toggle the visibility of hidden files in Windows Explorer using Ctrl+H
+// @version      1.0.0
+// @author       Asteski
+// @github       https://github.com/Asteski
+// @include      explorer.exe
+// @compilerOptions -std=c++20
```
m417z, line 9: Why is this compiler option needed?

mods/toggle-hidden-files.wh.cpp
```diff
@@ -0,0 +1,304 @@
...
+    return CallNextHookEx(g_hKeyboardHook, nCode, wParam, lParam);
+}
+
+// Mod initialization
+BOOL Wh_ModInit() {
+    // Load settings
+    LoadSettings();
+    
+    // Install keyboard hook
+    g_hKeyboardHook = SetWindowsHookExW(WH_KEYBOARD_LL, KeyboardHookProc, GetModuleHandle(nullptr), 0);
```
m417z, line 280: From [the documentation](https://learn.microsoft.com/en-us/windows/win32/winmsg/lowlevelkeyboardproc):

PR #2661, opened 2025-11-02
Add new mod: Taskbar Dock Animation

Ph0en1x-dev:
### Taskbar Dock Animation (Ultimate)
This mod adds a macOS-like taskbar animation on hover.  
The version is experimental — a few issues are still being worked on.

**Author:** Ph0en1x-dev  
**GitHub:** https://github.com/Ph0en1x-dev

### DIFF
--- a/mods/taskbar-dock-animation.wh.cpp
+++ b/mods/taskbar-dock-animation.wh.cpp
@@ -111,9 +111,8 @@ struct DockAnimationContext {
     bool isInitialized = false;
     winrt::weak_ref<FrameworkElement> taskbarFrame;
     std::vector<TaskbarIconInfo> icons;
-    winrt::event_token pointerMovedToken;
-    winrt::event_token pointerExitedToken;
 };
+
 
 // One context per monitor (keyed by XAML instance pointer)
 std::map<void*, DockAnimationContext> g_contexts;
@@ -449,20 +448,13 @@ void InitializeAnimationHooks(void* pThis, FrameworkElement const& taskbarFrame)
         ctx.taskbarFrame = taskbarFrame;
         ctx.isInitialized = false;
 
-        ctx.pointerMovedToken = taskbarFrame.PointerMoved(
-            [pThis](auto const&, Input::PointerRoutedEventArgs const& args) {
-                OnTaskbarPointerMoved(pThis, args);
-            });
-        ctx.pointerExitedToken = taskbarFrame.PointerExited(
-            [pThis](auto const&, Input::PointerRoutedEventArgs const&) {
-                OnTaskbarPointerExited(pThis);
-            });
-
         g_contexts[pThis] = std::move(ctx);
 
-        Wh_Log(L"DockAnimation: Monitor %p registered.", pThis);
-    } catch (winrt::hresult_error const& e) {
-        Wh_Log(L"DockAnimation: Failed to attach mouse events for %p: %s", pThis, e.message().c_str());
+        Wh_Log(L"DockAnimation: Monitor %p registered (hook-based).", pThis);
+    }
+    catch (winrt::hresult_error const& e) {
+        Wh_Log(L"DockAnimation: Failed to initialize context for %p: %s",
+               pThis, e.message().c_str());
     }
 }
 
@@ -729,8 +721,10 @@ void Wh_ModBeforeUninit() {
             Media::CompositionTarget::Rendering(g_renderingToken);
             Wh_Log(L"DockAnimation: Unhooked CompositionTarget::Rendering.");
         }
-    } catch (winrt::hresult_error const& e) {
-        Wh_Log(L"DockAnimation: HRESULT error unhooking rendering: %s", e.message().c_str());
+    }
+    catch (winrt::hresult_error const& e) {
+        Wh_Log(L"DockAnimation: HRESULT error unhooking rendering: %s",
+               e.message().c_str());
     }
 
     HWND hTaskbar = FindWindow(L"Shell_TrayWnd", NULL);
@@ -741,29 +735,23 @@ void Wh_ModBeforeUninit() {
                 try {
                     for (auto& pair : g_contexts) {
                         auto& ctx = pair.second;
-                        auto frame = ctx.taskbarFrame.get();
-                        if (frame) {
-                            frame.PointerMoved(ctx.pointerMovedToken);
-                            frame.PointerExited(ctx.pointerExitedToken);
-                            ResetAllIconScales(ctx.icons);
-                        }
+                        ResetAllIconScales(ctx.icons);
                     }
                     g_contexts.clear();
                     Wh_Log(L"DockAnimation: UI contexts cleaned up.");
-                } catch (winrt::hresult_error const& e) {
-                     Wh_Log(L"DockAnimation: HRESULT error during UI thread cleanup: %s", e.message().c_str());
                 }
+                catch (...) {}
             },
             nullptr);
-    } else {
-         try {
-            g_contexts.clear();
-         } catch (...) {}
+    }
+    else {
+        g_contexts.clear();
     }
 
     g_taskbarViewDllLoaded = false;
     g_hooksApplied = false;
 }
+
 
 void Wh_ModSettingsChanged() {
     Wh_Log(L"DockAnimation: Settings changed.");


mods/taskbar-dock-animation.wh.cpp
```diff
@@ -0,0 +1,436 @@
...
+    int effectRadius;
+} g_settings;
+
+// Global flags
+std::atomic<bool> g_taskbarViewDllLoaded = false;
+std::atomic<bool> g_fastPathHooksApplied = false;
+
+// Per-monitor context
+struct TaskbarIconInfo {
+    FrameworkElement element{ nullptr };
```
m417z: It's better to use `winrt::weak_ref` to avoid keeping destroyed icons in memory.

mods/taskbar-dock-animation.wh.cpp
```diff
@@ -0,0 +1,436 @@
...
+std::map<void*, DockAnimationContext> g_contexts;
+
+// Unclip only the taskbar icon container (TaskListButtonPanel) without changing taskbar size or window behavior.
+void DisableTaskbarIconClipping() {
+    try {
+        HWND hTaskbar = FindWindow(L"Shell_TrayWnd", NULL);
+        if (!hTaskbar) return;
+
+        FrameworkElement root{ nullptr };
+        winrt::copy_from_abi(root, hTaskbar);
```
m417z: That is wrong. `winrt::copy_from_abi` receives a WinRT object, not a Windows handle. It will most likely cause a crash.

Unfortunately, there's no easy way to get the taskbar element. Here's one way, but it's a bit complicated:
https://github.com/ramensoftware/windhawk-mods/blob/a44c656ddf26e242a365576bb63f8a0e5b66d36a/mods/taskbar-start-button-position.wh.cpp#L276

Another approach is to use one of the events. For example, here:
```cpp
        if (!ctx.isInitialized) {
            DisableTaskbarIconClipping(); // <-- right here
            RefreshIconPositions(ctx);
            ctx.isInitialized = true;
        }
```

But since currently you don't use `element.Clip()` anywhere else, and this function doesn't work, is it really needed?

mods/taskbar-dock-animation.wh.cpp
```diff
@@ -0,0 +1,436 @@
...
+            info.element = element;
+            info.originalCenterX = point.X + (element.ActualWidth() / 2.0);
+            ctx.icons.push_back(info);
+        } catch (...) {}
+    }
+
+    int childrenCount = Media::VisualTreeHelper::GetChildrenCount(element);
+    for (int i = 0; i < childrenCount; i++) {
+        auto child = Media::VisualTreeHelper::GetChild(element, i).try_as<FrameworkElement>();
+        if (child) FindAllTaskbarButtonsRecursive(child, ctx);
```
m417z: Instead of going over all elements recursively, wouldn't it be faster and easier to just go over all the child elements of `TaskbarFrameRepeater`?

mods/taskbar-dock-animation.wh.cpp
```diff
@@ -0,0 +1,436 @@
...
+
+// Per-monitor context
+struct TaskbarIconInfo {
+    FrameworkElement element{ nullptr };
+    double originalCenterX = 0.0;
+};
+
+struct DockAnimationContext {
+    bool isInitialized = false;
+    FrameworkElement taskbarFrame{ nullptr };
```
m417z: It's better to use `winrt::weak_ref` here too.

mods/taskbar-dock-animation.wh.cpp
```diff
@@ -0,0 +1,436 @@
...
+            LoadLibraryExW_Hook,
+            &LoadLibraryExW_Original
+        );
+    }
+    return TRUE;
+}
+
+void Wh_ModAfterInit() {
+    Wh_Log(L"DockAnimation: Wh_ModAfterInit");
+    if (g_taskbarViewDllLoaded && !g_fastPathHooksApplied) {
```
m417z: What is `g_fastPathHooksApplied` useful for? It's always `false` here, and it's never accessed afterwards.

mods/taskbar-dock-animation.wh.cpp
```diff
@@ -0,0 +1,436 @@
...
+        Wh_ApplyHookOperations();
```
m417z: This will call `Wh_ApplyHookOperations` in more cases than necessary. I'm not sure what you tried to do, but I suggest copying the logic from one of the existing mods.

mods/taskbar-dock-animation.wh.cpp
```diff
@@ -0,0 +1,436 @@
...
+}
+
+void Wh_ModSettingsChanged() {
+    Wh_Log(L"DockAnimation: Settings changed.");
+    LoadSettings();
+    try {
+        for (auto& pair : g_contexts) {
+            auto& ctx = pair.second;
+            ctx.isInitialized = false;
+            ResetAllIconScales(ctx.icons);
```
m417z: This won't work - `Wh_ModSettingsChanged` runs in a different thread than the UI thread. I suggest just not calling it here, the sizes will be updated on the next mouse hover anyway.

mods/taskbar-dock-animation.wh.cpp
```diff
@@ -0,0 +1,436 @@
...
+
+// Cleanup
+void Wh_ModBeforeUninit() {
+    Wh_Log(L"DockAnimation: Wh_ModBeforeUninit");
+    try {
+        for (auto& pair : g_contexts) {
+            auto& ctx = pair.second;
+            if (ctx.taskbarFrame) {
+                ctx.taskbarFrame.PointerMoved(ctx.pointerMovedToken);
+                ctx.taskbarFrame.PointerExited(ctx.pointerExitedToken);
```
m417z: This won't work - `Wh_ModBeforeUninit` runs in a different thread than the UI thread. That's why Explorer crashes on unload. You can find the taskbar thread with `FindWindow(L"Shell_TrayWnd", NULL)` and `GetWindowThreadProcessId`, and then use `RunFromWindowThread` which can be copied from one of the mods, for example:
https://github.com/ramensoftware/windhawk-mods/blob/a44c656ddf26e242a365576bb63f8a0e5b66d36a/mods/taskbar-start-button-position.wh.cpp#L329

Ph0en1x-dev: Thanks a lot for the detailed review and helpful notes!
I’ve reviewed all your comments and fully understand what needs to be improved — really appreciate the clarification and the technical insight.

I’ve also updated the mod itself: added smoother, more aesthetic animations and introduced a new idle “breathing” effect when the cursor stays over an icon — it gives the interaction a more natural, polished feel.

I’ll apply the remaining technical fixes (like using winrt::weak_ref, safe cleanup, and simplified traversal) in the next update to make the mod fully compliant and stable.

Thanks again for your time and guidance — your feedback really helps me polish this project!

mods/taskbar-dock-animation.wh.cpp
```diff
@@ -0,0 +1,719 @@
...
+    };
+
+    try {
+        // Try to use TaskbarFrameRepeater children (faster path)
+        FrameworkElement repeater = EnumChildElements(
+            taskbarFrame,
+            [](FrameworkElement child) {
+                auto className = winrt::get_class_name(child);
+                return std::wstring_view(className).find(L"TaskbarFrameRepeater") != std::wstring_view::npos;
+            });
```
m417z: You could use `FindChildByClassName` instead. Matching a substring isn't needed here.

https://github.com/ramensoftware/windhawk-mods/blob/813f2444023ba0ae224d1469fac92f3ea688b71e/mods/taskbar-start-button-position.wh.cpp#L147-L152

mods/taskbar-dock-animation.wh.cpp
```diff
@@ -0,0 +1,719 @@
...
+
+        if (repeater) {
+            int count = VisualTreeHelper::GetChildrenCount(repeater);
+            for (int i = 0; i < count; i++) {
+                auto child = VisualTreeHelper::GetChild(repeater, i).try_as<FrameworkElement>();
+                if (!child) continue;
+                processButtonElement(child);
+            }
+        } else {
+            // Fallback: recursive search for TaskListButton
```
m417z: Why is the fallback needed?

mods/taskbar-dock-animation.wh.cpp
```diff
@@ -275,53 +423,122 @@ void InitializeAnimationHooks(void* pThis, FrameworkElement const& taskbarFrame)
...
 }
 
+// Main animation loop (pulse)
+void OnCompositionTargetRendering(winrt::Windows::Foundation::IInspectable const&,
+                                  winrt::Windows::Foundation::IInspectable const&) {
+    try {
+        void* pThis_key = g_activeContextKey.load();
+        if (pThis_key == nullptr) {
+            g_isBouncing = false;
+            return;
```
m417z, line 468: Since this function is called once per frame, perhaps it's better to unregister it once bounce is over.

m417z (commented):
In addition to the notes I left, I tried the mod, and I saw two problems:
* The mod doesn't work right away, likely because `TaskbarFrame_MeasureOverride_Hook` is only called on events such as resolution change. I have an idea for fixing it, I'll post it later when I have more time.
* The mod crashes when unloaded.

mods/taskbar-dock-animation.wh.cpp
```diff
@@ -608,17 +635,19 @@ BOOL Wh_ModInit() {
 
 void Wh_ModAfterInit() {
     Wh_Log(L"DockAnimation: Wh_ModAfterInit");
-    if (g_taskbarViewDllLoaded && !g_fastPathHooksApplied) {
-        Wh_ApplyHookOperations();
-        g_fastPathHooksApplied = true;
+    if (g_taskbarViewDllLoaded) {
+        if (!g_hooksApplied.exchange(true)) {
+            Wh_ApplyHookOperations();
+            Wh_Log(L"DockAnimation: Hooks applied (fast path).");
```
m417z: You don't need to call it if `HookTaskbarViewDllSymbols` was called in `Wh_ModInit`, because after `Wh_ModInit` returns, hooks are applied automatically.

mods/taskbar-dock-animation.wh.cpp
```diff
@@ -0,0 +1,757 @@
...
+
+bool HookTaskbarViewDllSymbols(HMODULE module) {
+    // Taskbar.View.dll
+    WindhawkUtils::SYMBOL_HOOK taskbarViewHooks[] = {
+        {
+            {
+                LR"(public: virtual int __cdecl winrt::impl::produce<struct winrt::Taskbar::implementation::TaskbarFrame,struct winrt::Windows::UI::Xaml::IFrameworkElementOverrides>::MeasureOverride(struct winrt::Windows::Foundation::Size,struct winrt::Windows::Foundation::Size *))"
+            },
+            &TaskbarFrame_MeasureOverride_Original,
+            TaskbarFrame_MeasureOverride_Hook,
```
m417z: Regarding this problem:

> The mod doesn't work right away, likely because `TaskbarFrame_MeasureOverride_Hook` is only called on events such as resolution change. I have an idea for fixing it, I'll post it later when I have more time.

Here's an idea for a fix:

Instead of hooking `MeasureOverride`, you can hook these functions:

`OnPointerMoved`:
```
public: virtual int __cdecl winrt::impl::produce<struct winrt::Taskbar::implementation::TaskbarFrame,struct winrt::Windows::UI::Xaml::Controls::IControlOverrides>::OnPointerMoved(void *)
```

`OnPointerExited`:
```
public: virtual int __cdecl winrt::impl::produce<struct winrt::Taskbar::implementation::TaskbarFrame,struct winrt::Windows::UI::Xaml::Controls::IControlOverrides>::OnPointerExited(void *)
```

Note that you'll also need to verify the class name, because these functions may also be called for other elements. You can look at existing mods to see how it's done, example:
https://github.com/ramensoftware/windhawk-mods/blob/056cead7a22636a34b5a531af18f11613d9468cb/mods/taskbar-show-desktop-button-aero-peek.wh.cpp#L160-L207

Let me know if you need help.

Ph0en1x-dev: Thank you for the detailed idea — I’ve now implemented your suggested approach using OnPointerMoved and OnPointerExited hooks for the TaskbarFrame.
The mod initializes instantly without requiring any user interaction, and the behavior remains stable across Windows 11 25H2 build.

The legacy MeasureOverride hook is still present as a fallback for older systems, but it’s currently inactive by default.
Once you have time, I’d appreciate it if you could review and test the current implementation to confirm whether this structure aligns with your intended direction — or if you’d recommend further adjustments or refinements.

Thanks again for your guidance — I’ll be glad to apply any changes you suggest)).

Ph0en1x-dev:
I’ve also tried to investigate both of these visual issues in more depth — the clipping and the slight blurriness when icons are upscaled.

For the clipping, I attempted to traverse and modify the Clip region of the taskbar’s XAML elements, as well as to detach it via parent/child hierarchy traversal. However, this consistently caused access violations or Explorer crashes — likely because the geometry objects are COM-managed and not writable at runtime.

As for the blurriness, I experimented with adjusting the scaling mode (e.g., BitmapScalingMode::NearestNeighbor) and applying transforms at different hierarchy levels, but the system still enforces linear interpolation when scaling beyond 1.0.

If you have any insights or preferred approaches — such as whether wrapping each icon in a Canvas or using composition-level scaling might be a viable solution — I’d really appreciate your thoughts before moving further in that direction.

m417z:
Clipping: A hack that perhaps can be done is to make the container larger, and then fake everything else: the work area space (occupied by maximized windows), the actual visual taskbar height, etc. For the work area space, looking at the vertical taskbar and taskbar-on-top mods can help.

Blurriness: The "Taskbar height and icon size" mod can increase the icons and upscale them too, looking at it can help.

Regarding the PR, I didn't have time to look at the latest changes, but from a quick glance, looks like everything is addressed for now. Is it ready to be merged from your perspective?

m417z:
> The legacy MeasureOverride hook was not removed at this stage — it remains in the code as a fallback initialization path for potential compatibility with older Windows 11 builds.
> (I’ll wait for your feedback to decide whether it should be fully removed in the final version.)

I see no reason not to remove it, I don't see how it can help with compatibility. But if you prefer to keep it for some reason, I don't mind.

From my perspective, it can be merged. Let me know when it's ready.

Ph0en1x-dev:
Thanks a lot for reviewing and for the additional hints — I’ll definitely take a look at the “vertical taskbar” and “Taskbar height and icon size” mods to explore those approaches for handling clipping and scaling.

Regarding the legacy MeasureOverride hook — understood. I’ve removed it completely.

From my side, everything is ready for merge. If any feedback, compatibility issues, or improvement ideas come up later from users, I’ll be happy to address and iterate on them right away.

Thanks again for the guidance and collaboration throughout this process!

mods/taskbar-dock-animation.wh.cpp
```diff
@@ -0,0 +1,780 @@
...
+    winrt::weak_ref<FrameworkElement> element;
+    double originalCenterX = 0.0;
+};
+
+struct DockAnimationContext {
+    bool isInitialized = false;
+    winrt::weak_ref<FrameworkElement> taskbarFrame;
+    std::vector<TaskbarIconInfo> icons;
+    winrt::event_token pointerMovedToken;
+    winrt::event_token pointerExitedToken;
```
m417z: Since you've switched to using hooks, you should be able to remove `pointerMovedToken` and `pointerExitedToken` and the relevant code that uses them.

m417z:
Thank you for the contribution and the patience!

PR #2669, opened 2025-11-04
New mod to hide home and gallery section from explorer

rinosaur681:


mods/hide-home-gallery-explorer.wh.cpp
```diff
@@ -0,0 +1,349 @@
...
+    return NULL;
+}
+
+// Enumerate all top-level Explorer windows (CabinetWClass)
+static std::vector<HWND> GetExplorerWindows() {
+    std::vector<HWND> result;
+    EnumWindows([](HWND hwnd, LPARAM lParam)->BOOL {
+        wchar_t cls[256] = {};
+        GetClassNameW(hwnd, cls, 255);
+        if (wcscmp(cls, L"CabinetWClass") == 0 && IsWindowVisible(hwnd)) {
```
m417z: There might be multiple explorer.exe instances. Make sure you only handle windows that belong to the current process. You can use `GetWindowThreadProcessId`.

Divc09:
The explorer constantly refreshes and goes back to the home directory with this mod. Needs fixing.

PR #2680, opened 2025-11-09
Click on empty taskbar space v2.0

m1lhaus:
Changelog:

- concept of triggers completely reworked (incompatible with v1.x settings)
- user can add up to 50 triggers
- added keyboard modifiers for each trigger - resolves https://github.com/m1lhaus/windhawk-mods/issues/22 and resolves https://github.com/m1lhaus/windhawk-mods/issues/1
- added single left/right button click triggers - resolves https://github.com/m1lhaus/windhawk-mods/issues/27 and resolves https://github.com/m1lhaus/windhawk-mods/issues/24
- smart handling of taskbar context menu on right-click triggers
- media keys not working bug fixed - fixes https://github.com/m1lhaus/windhawk-mods/issues/30

### DIFF
--- a/mods/taskbar-empty-space-clicks.wh.cpp
+++ b/mods/taskbar-empty-space-clicks.wh.cpp
@@ -1183,7 +1183,7 @@ static struct
 class COMAPI
 {
 public:
-    COMAPI() : m_isInitialized(false), m_isCOMInitialized(false), m_needsCOMUninit(false), m_isUIAInitialized(false), m_isDEInitialized(false),
+    COMAPI() : m_isInitialized(false), m_isCOMInitialized(false), m_isUIAInitialized(false), m_isDEInitialized(false),
                m_pUIAutomation(nullptr), m_pDeviceEnumerator(nullptr) {}
 
     // init COM for UIAutomation and Volume control
@@ -1191,25 +1191,10 @@ public:
     {
         if (!m_isCOMInitialized)
         {
-            const auto hr = CoInitializeEx(NULL, COINIT_APARTMENTTHREADED);
-            if (SUCCEEDED(hr))
+            if (SUCCEEDED(CoInitializeEx(NULL, COINIT_APARTMENTTHREADED))) // COM was most likely already initialized in the GUI thread, but just to be sure
             {
                 m_isCOMInitialized = true;
-                if (hr == S_OK) // COM was most likely already initialized in the GUI thread, in case not, we will need to uninitialize it later
-                {
-                    LOG_INFO(L"COM initilized");
-                    m_needsCOMUninit = true;
-                }
-                else if (hr == S_FALSE)
-                {
-                    LOG_INFO(L"COM already initialized on this thread");
-                    m_needsCOMUninit = true; // each successful CoInitializeEx must be balanced with CoUninitialize including S_FALSE
-                }
-                else
-                {
-                    LOG_INFO(L"COM already initialized in incompatible concurrency model");
-                    m_needsCOMUninit = false;
-                }
+                LOG_INFO(L"COM initilized");
             }
             else
             {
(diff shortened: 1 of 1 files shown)


mods/taskbar-empty-space-clicks.wh.cpp
```diff
@@ -2847,5 +3489,5 @@ void Wh_ModUninit()
             UnsubclassTaskbarWindow(hSecondaryWnd);
         }
     }
-    g_comInitializer.Uninit();
+    g_comAPI.Uninit();
```
m417z: This doesn't seem correct. COM is initialized lazily on the UI thread, but isn't initialized here in `Wh_ModUninit` which runs in the Windhawk thread.

From [the docs](https://learn.microsoft.com/en-us/windows/win32/api/combaseapi/nf-combaseapi-couninitialize):

> A thread must call CoUninitialize once for each successful call it has made to the [CoInitialize](https://learn.microsoft.com/en-us/windows/desktop/api/objbase/nf-objbase-coinitialize) or [CoInitializeEx](https://learn.microsoft.com/en-us/windows/desktop/api/combaseapi/nf-combaseapi-coinitializeex) function

Since all usages of `g_comAPI` are done in the UI thread (right?), I think it'd be best to just remove `CoInitialize` and `CoUninitialize`, as I assume the Explorer UI thread already initializes COM.

Regarding the `Release` calls ("force underlying ptr to get released" parts), they also should be called in the UI thread. Since you're subclassing `g_hTaskbarWnd`, you can send it a custom message to have it run the uninitialization before unsubclassing.

m1lhaus: You are right. That never occured to me that now the Uninit would be done from a different thread. Anyway as you said, the COM is already initialized from GUI thread, but I implemented both Init and UnInit just in case as you proposed.

mods/taskbar-empty-space-clicks.wh.cpp
```diff
@@ -1183,18 +1183,28 @@ static struct
...
-                LOG_INFO(L"COM initilized");
+                if (hr == S_OK) // COM was most likely already initialized in the GUI thread, in case not, we will need to uninitialize it later
+                {
+                    LOG_INFO(L"COM initilized");
+                    m_needsCOMUninit = true;
+                }
+                else
+                {
+                    LOG_INFO(L"COM already initialized");
+                    m_needsCOMUninit = false;
```
m417z: From [the documentation](https://learn.microsoft.com/en-us/windows/win32/api/combaseapi/nf-combaseapi-coinitializeex):

> To uninitialize the COM library gracefully on a thread, each successful call to [CoInitialize](https://learn.microsoft.com/en-us/windows/desktop/api/objbase/nf-objbase-coinitialize) or CoInitializeEx, including any call that returns S_FALSE, must be balanced by a corresponding call to [CoUninitialize](https://learn.microsoft.com/en-us/windows/desktop/api/combaseapi/nf-combaseapi-couninitialize)

So it was correct before adding `m_needsCOMUninit`.

m1lhaus: Jesus, they wrote it to "BTW section" 🙃

mods/taskbar-empty-space-clicks.wh.cpp
```diff
@@ -1200,9 +1200,14 @@ class COMAPI
...
+                else if (hr == S_FALSE)
+                {
+                    LOG_INFO(L"COM already initialized on this thread");
+                    m_needsCOMUninit = true; // each successful CoInitializeEx must be balanced with CoUninitialize including S_FALSE
+                }
                 else
                 {
-                    LOG_INFO(L"COM already initialized");
+                    LOG_INFO(L"COM already initialized in incompatible concurrency model");
                     m_needsCOMUninit = false;
```
m417z: I don't think there's any other result which is successful, yet isn't `S_OK` or `S_FALSE`. So it seems a bit redundant, but not incorrect.

m1lhaus: Correct, I did not realize that `RPC_E_CHANGED_MODE=0x80010106L` ends up being a negative value.

PR #2681, opened 2025-11-10
Command Prompt Folder Larp 1.1

kawapure:
Fixed a bug where the actual path was clobbered for some commands, resulting in an error messaging stating that the path is not valid and no command running.

### DIFF
--- a/mods/cmd-folder-larp.wh.cpp
+++ b/mods/cmd-folder-larp.wh.cpp
@@ -208,7 +208,7 @@ HRESULT __cdecl StringCchPrintfW_hook(LPWSTR szBuffer, size_t size, LPCWSTR szTe
         LPCWSTR szFirstStringArgument = va_arg(vaMut, LPCWSTR);
 
         if (
-            !IsBadStringPtrW(szFirstStringArgument, MAX_PATH) &&
+            (szTemplate[0] == L'%' && szTemplate[1] == L's') &&
             wcscmp(szFirstStringArgument, g_pszCurrentDirectory) == 0
         )
         {


mods/cmd-folder-larp.wh.cpp
```diff
@@ -148,38 +156,98 @@ bool ApplyPathReplacementToString(LPWSTR szTarget, DWORD nBufferLength)
...
+        // The directory string is only ever the first varadic argument in
+        // CMD's implementation. It's probably unsafe to generally assume
+        // a 64-bit word argument, but it's fine under the narrow filter
+        // applied here.
+        va_list vaMut; // va_arg shifts the internal pointer of the list,
+        va_start(vaMut, szTemplate);        // so a mutable copy is used.
+        LPCWSTR szFirstStringArgument = va_arg(vaMut, LPCWSTR);
+
+        if (
+            !IsBadStringPtrW(szFirstStringArgument, MAX_PATH) &&
```
m417z: `IsBadStringPtr` is rarely the right way to to do validation, see:
https://devblogs.microsoft.com/oldnewthing/20060927-07/?p=29563
Why is it necessary here? How about validating `szTemplate` instead, e.g. that the first `%` character is followed by `s`, or some other validation that makes sense here?

kawapure: I know; I was just too lazy to write code to parse the template string (even though `sz[0] == L'%' && sz[1] == L's'` alone would have technically sufficed). This is a narrow path, only taken as the callee of a specific function which only passes a few distinct argument sets, so I found it more convenient to use this "dangerous" function. Even if it could cause improper behaviour (which I have not observed) I would be surprised if an access violation would ever follow in application, even if it is possible in theory.

I'll replace it with the extremely primitive template validation, which is fine for all possible template strings passed to this function.

PR #2728, opened 2025-11-23
Add Explorer Ctrl+N to New File mod

lieyanbang:
# Explorer Ctrl+N to New File

Intercepts Explorer’s **Ctrl+N** (open new window) and replaces it with creating an empty file in the current directory, then entering rename mode.

**How it works**

Uses Windhawk to inject a `WH_KEYBOARD_LL` low-level keyboard hook into explorer.exe to intercept Ctrl+N.
Since the desktop cannot navigate folders, files are created directly in the desktop path.

### DIFF
--- a/mods/explorer-ctrln-newfile.wh.cpp
+++ b/mods/explorer-ctrln-newfile.wh.cpp
@@ -87,15 +87,21 @@ static bool IsExplorerTopLevel(HWND hwnd) {
     return wcscmp(cls, L"CabinetWClass") == 0 || wcscmp(cls, L"ExploreWClass") == 0;
 }
 
+static bool IsCurrentProcessWindow(HWND hwnd) {
+    DWORD pid = 0;
+    GetWindowThreadProcessId(hwnd, &pid);
+    return pid == GetCurrentProcessId();
+}
+
 static HWND HitTestExplorerByCursor() {
     POINT pt;
     GetCursorPos(&pt);
     HWND h = WindowFromPoint(pt);
     if (!h) return nullptr;
     HWND root = GetAncestor(h, GA_ROOT);
-    if (IsExplorerTopLevel(root)) return root;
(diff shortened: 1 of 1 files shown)


mods/explorer-ctrln-newfile.wh.cpp
```diff
@@ -0,0 +1,435 @@
...
+
+        if (info && isKeyUp && info->vkCode == 'N') g_nKeyDown = false;
+        if (info && isKeyUp && (info->vkCode == VK_LCONTROL || info->vkCode == VK_RCONTROL)) g_nKeyDown = false;
+
+        if (isKeyDown && info && info->vkCode == 'N' && !g_nKeyDown) {
+            const bool ctrl = (GetAsyncKeyState(VK_CONTROL) & 0x8000) != 0;
+            const bool shift = (GetAsyncKeyState(VK_SHIFT) & 0x8000) != 0;
+            if (ctrl && !shift) {
+                HWND fg = GetForegroundWindow();
+                if (IsExplorerTopLevel(fg) || fg == GetShellWindow()) {
```
m417z: There might be several explorer.exe instances. It's better to make sure that the window is part of the current process with `GetWindowThreadProcessId`.

mods/explorer-ctrln-newfile.wh.cpp
```diff
@@ -0,0 +1,435 @@
...
+    ComInit com;
+
+    HWND target = GetTargetExplorerHWND();
+    if (!target) {
+        Wh_Log(L"[Ctrln] No explorer window detected.");
+        return;
+    }
+
+    std::wstring dir;
+    if (!GetExplorerPathFromHWND(target, dir) || dir.empty()) dir = GetDesktopDir();
```
m417z: If a path is unsupported (e.g. Control Panel, I assume), it seems wrong to randomly create a file on the desktop.

PR #2737, opened 2025-11-25
Add Explorer Ctrl+Q to New Folder mod

TheShadyRainbow4:
Introduces a Windhawk mod that maps Ctrl+Q in Windows Explorer to create a new folder in the current directory,  The mod avoids shortcut conflicts, ensures stability with shell notifications and thread synchronization, and programmatically selects and renames the new folder.

### DIFF
--- a/mods/explorer-ctrlq-new-folder.wh.cpp
+++ b/mods/explorer-ctrlq-new-folder.wh.cpp
@@ -1,5 +1,5 @@
 // ==WindhawkMod==
-// @id              explorer-ctrlq-newfolder
+// @id              explorer-ctrlq-new-folder
 // @name            Explorer Ctrl+Q to New Folder
 // @description     Press Ctrl+Q in Explorer and create a new folder in current path, including desktop.
 // @version         1.2


mods/explorer-ctrlq-newfile.wh.cpp
```diff
@@ -0,0 +1,463 @@
+// ==WindhawkMod==
+// @id              explorer-ctrlq-newfile
+// @name            Explorer Ctrl+Q to New File
```
m417z: It probably makes more sense to call it:
```suggestion
// @id              explorer-ctrlq-newfolder
// @name            Explorer Ctrl+Q to New Folder
```

Same for the mod id

mods/explorer-ctrlq-newfile.wh.cpp
```diff
@@ -0,0 +1,463 @@
...
+// @name:zh-CN      Create new file with Ctrl+Q
```
m417z: This is supposed to be the name in Chinese (zh-CN). If you're not using it, remove it.

mods/explorer-ctrlq-newfile.wh.cpp
```diff
@@ -0,0 +1,463 @@
...
+    psb->Release();
+    if (FAILED(hr) || !psv) {
+        wb->Release();
+        return;
+    }
+
+    IShellFolder* pParent = nullptr;
+    PCUITEMID_CHILD pChild = nullptr;
+    if (BindParentAndChildPIDL(fullPath, &pParent, &pChild)) {
+        // This is where it was crashing - added safety
```
m417z: Why does it crash? It's preferable to have a proper fix rather than using `Sleep` which just reduces the chances for a crash.

TheShadyRainbow4: It crashed due to file explorer not displaying the new folder right away and trying to rename the file before it existed. I will remove the comment. It should not crash! I would love it if more people tested it in different environments though so I can continue to update it :)

mods/explorer-ctrlq-newfile.wh.cpp
TheShadyRainbow4: Removed! Sorry that was a left over of a previous one. Thanks for pointing it out :) Much appreciated

mods/explorer-ctrlq-newfile.wh.cpp
m417z: I mean, if it crashes without the Sleep call, that's not good. Even if the folder doesn't exist for some reason, it should just abort the rename operation, not crash.

SamHasler:
What's the problem with the built-in `Ctrl` + `Shift` + `N` to create a new folder?

PR #2746, opened 2025-11-26
Add Mod: Taskbar Music Lounge

ghost:
A sleek, pill-shaped music ticker for your Windows 11 taskbar. It sits unobtrusively in the corner, scrolling your current track and providing quick media controls.

### DIFF
--- a/mods/taskbar-music-lounge.wh.cpp
+++ b/mods/taskbar-music-lounge.wh.cpp
@@ -5,7 +5,7 @@
 // @version         3.0
 // @author          Hashah2311
 // @github          https://github.com/Hashah2311
-// @include         windhawk.exe
+// @include         explorer.exe
 // @compilerOptions -lole32 -ldwmapi -lgdi32 -luser32 -lwindowsapp -lshcore -lgdiplus
 // ==/WindhawkMod==
 
@@ -96,6 +96,48 @@ typedef struct _WINDOWCOMPOSITIONATTRIBDATA {
     SIZE_T SizeOfData;
 } WINDOWCOMPOSITIONATTRIBDATA;
 typedef BOOL(WINAPI* pSetWindowCompositionAttribute)(HWND, WINDOWCOMPOSITIONATTRIBDATA*);
+
+// --- Z-Band API ---
+enum ZBID {
+    ZBID_DEFAULT = 0,
+    ZBID_DESKTOP = 1,
+    ZBID_UIACCESS = 2,
+    ZBID_IMMERSIVE_IHM = 3,
+    ZBID_IMMERSIVE_NOTIFICATION = 4,
+    ZBID_IMMERSIVE_APPCHROME = 5,
+    ZBID_IMMERSIVE_MOGO = 6,
+    ZBID_IMMERSIVE_EDGY = 7,
+    ZBID_IMMERSIVE_INACTIVEMOBODY = 8,
+    ZBID_IMMERSIVE_INACTIVEDOCK = 9,
+    ZBID_IMMERSIVE_ACTIVEMOBODY = 10,
+    ZBID_IMMERSIVE_ACTIVEDOCK = 11,
+    ZBID_IMMERSIVE_BACKGROUND = 12,
+    ZBID_IMMERSIVE_SEARCH = 13,
+    ZBID_GENUINE_WINDOWS = 14,
+    ZBID_IMMERSIVE_RESTRICTED = 15,
+    ZBID_SYSTEM_TOOLS = 16,
+    ZBID_LOCK = 17,
+    ZBID_ABOVELOCK_UX = 18,
+};
+
+typedef HWND(WINAPI* pCreateWindowInBand)(
+    DWORD dwExStyle,
+    LPCWSTR lpClassName,
+    LPCWSTR lpWindowName,
+    DWORD dwStyle,
+    int x,
+    int y,
+    int nWidth,
+    int nHeight,
+    HWND hWndParent,
+    HMENU hMenu,
+    HINSTANCE hInstance,
+    LPVOID lpParam,
+    DWORD dwBand
+);
+
+typedef BOOL(WINAPI* pSetWindowBand)(HWND hWnd, HWND hwndInsertAfter, DWORD dwBand);
+typedef BOOL(WINAPI* pGetWindowBand)(HWND hWnd, PDWORD pdwBand);
 
 // --- Configurable State ---
 struct ModSettings {
@@ -370,20 +412,27 @@ void DrawMediaPanel(HDC hdc, int width, int height) {
 // --- Window Procedure ---
 #define IDT_POLL_MEDIA 1001
 #define IDT_ANIMATION  1002
-#define IDT_ZORDER     1003
+#define APP_WM_CLOSE   WM_APP
 
 LRESULT CALLBACK MediaWndProc(HWND hwnd, UINT msg, WPARAM wParam, LPARAM lParam) {
     switch (msg) {
         case WM_CREATE: 
             UpdateAppearance(hwnd); // Apply DWM Rounding + Acrylic
             SetTimer(hwnd, IDT_POLL_MEDIA, 1000, NULL); 
-            SetTimer(hwnd, IDT_ZORDER, 2000, NULL);      
             return 0;
 
         case WM_ERASEBKGND: 
             return 1;
 
+        case WM_CLOSE:
+            return 0;
+
+        case APP_WM_CLOSE:
+            DestroyWindow(hwnd);
+            return 0;
+
         case WM_DESTROY:
+            g_SessionManager = nullptr;
             PostQuitMessage(0);
             return 0;
 
@@ -427,9 +476,6 @@ LRESULT CALLBACK MediaWndProc(HWND hwnd, UINT msg, WPARAM wParam, LPARAM lParam)
                 } else {
                     KillTimer(hwnd, IDT_ANIMATION); 
                 }
-            }
-            else if (wParam == IDT_ZORDER) {
-                SetWindowPos(hwnd, HWND_TOPMOST, 0, 0, 0, 0, SWP_NOMOVE | SWP_NOSIZE | SWP_NOACTIVATE);
             }
             return 0;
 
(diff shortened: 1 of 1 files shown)


mods/taskbar-music-lounge.wh.cpp
```diff
@@ -0,0 +1,722 @@
...
+    size_t ytmPos = title.find(L" - YouTube Music");
+    if (ytmPos != wstring::npos) {
+        *pResult = title.substr(0, ytmPos);
+        return FALSE;
+    }
+    if (title.find(L"YouTube Music") != wstring::npos && title.find(L" - ") != wstring::npos) {
+        *pResult = title;
+        return FALSE;
+    }
+    if (title.find(L"Spotify") != wstring::npos && title.find(L" - ") != wstring::npos) {
```
m417z: If I understand correctly, the mod only supports YouTube Music and Spotify, it makes sense to mention it in the readme.

By the way, you might prefer using `GlobalSystemMediaTransportControlsSessionManager` for a universal solution. You can take a look at this pull request for a reference:
https://github.com/m417z/my-windhawk-mods/pull/34

mods/taskbar-music-lounge.wh.cpp
```diff
@@ -0,0 +1,722 @@
...
+// --- Helper: Load Settings ---
+void LoadSettings() {
+    g_Settings.width = Wh_GetIntSetting(L"PanelWidth");
+    g_Settings.height = Wh_GetIntSetting(L"PanelHeight");
+    g_Settings.fontSize = Wh_GetIntSetting(L"FontSize");
+    g_Settings.offsetX = Wh_GetIntSetting(L"OffsetX");
+    g_Settings.offsetY = Wh_GetIntSetting(L"OffsetY");
+    g_Settings.autoTheme = Wh_GetIntSetting(L"AutoTheme") != 0;
+    
+    PCWSTR textHex = Wh_GetStringSetting(L"TextColor");
```
m417z, line 184: A call to `Wh_FreeStringSetting` to free the string is missing, here and below.

Anixx:
Thanks, this clears it up.

mods/taskbar-music-lounge.wh.cpp
ghost: > If I understand correctly, the mod only supports YouTube Music and Spotify, it makes sense to mention it in the readme.

Noted, I will make the changes accordingly.

> you might prefer using GlobalSystemMediaTransportControlsSessionManager for a universal solution.

I needed some advice regarding that, Implementing GSMTC in C++ requires using WinRT interfaces, wont this add significant complexity, new libraries, and potential stability issues??

m417z: > wont this add significant complexity

Depends on the implementation, generally it shouldn't.

> new libraries

Sure, but that shouldn't be an issue.

> and potential stability issues

That, too, depends on the implementation, but the WinRT APIs themselves aren't unstable or something, they are used by Windows.

ghost: Thanks. Also, I have another question, I want to use DWM native rounding which available only on Win11, should I use that or stay windows version neutral?

ghost: @m417z

m417z: It's up to you. Windows 11 has rounded design and Windows 10 doesn't so I think it'd be a good fit, but if you want to leave it up to the user to you can add an option for that.

SpacEagle17:
Works flawlessly! ty

And idk if this could be looked into, I kinda hid it below my issue report haha
image

m417z:
> Also, since I use the `Taskbar auto-hide speed` mod, the hiding speed of the music player no longer matches the taskbar, so a setting to alter that would be appreciated, if at all possible :)

It was discussed here, you might want to try the proposed solution: https://github.com/ramensoftware/windhawk-mods/commit/2eddf6953b7aaee9ad5eb050bb329e135f4c7798#commitcomment-173045991

SpacEagle17:
> It was discussed here, you might want to try the proposed solution: [2eddf69#commitcomment-173045991](https://github.com/ramensoftware/windhawk-mods/commit/2eddf6953b7aaee9ad5eb050bb329e135f4c7798#commitcomment-173045991)

And once again, all changes mentioned there made this perfect! Yep that and the AppUserModelID should for sure be in the next update of this mod! Works all as expected and perfectly!
Just 1 small nitpick/suggestion:

Also with the taskbar styler mod and a modified taskbar to make it narrower, for example the glass theme one does not even have to disable widgets (visible in my video shared earlier in this thread). Currently, disabling the widget in Windows settings and enabling it again overrides the mod's way of detecting it. Ideally an option could be added to overwrite it within the mod

Cinabutts:
Prerequisite: This is a continuation of my "Issue" here: [#2828 ](https://github.com/ramensoftware/windhawk-mods/issues/2828) 
I used Gemini 3 Pro Preview(so be cautious, but it seems fairly capable, I also have it cross reference multiple other mods just fyi)

Hey @Hashah2311, **1:** I want to say if you haven't read the "issue" I created before, I am incredibly grateful for this mod man, I have been looking for something exactly like this as a Rainmeter skin and just wanted to say thank you and sorry if any of this sounds assertive or anything like im taking over your amazing work you likely spent a LOT more time into than me just doing this.
**2:** I am posting this here as I felt it might reach you sooner. Also in case the other users above see this that have posted about issues I may have fixed in my version.

Anyway, I've spent some time extensively reworking the mod, This version fixes a fair amount of issues others have experienced (like auto-hide lag and it remaining visible in fullscreen apps) and improves general stability.

Could you please review the code below and consider pushing it as an official update?

========================================================
Tested on: Windows 11 Pro Insider Preview Build 26100.ge_release.240331-1435

image
Screenshot 2025-12-24 023003

------------------------------------------------------------------------

**Summary of changes:**

To fix the Auto-Hide, Fullscreen, and High-DPI issues, the logic was overhauled from a "guessing" system to a "listening" system. Here is the breakdown:

*   **Reworked Window Positioning & Smart Docking:**
    *   **Core Sync:** Replaced the timer-based positioning with a centralized **`SyncPositionWithTaskbar`** function driven by a **`WinEventHook`**. The widget now "listens" for the taskbar's exact movement coordinates, staying glued pixel-perfectly even during custom animations.
    *   **Smart Docking (Startup/Shutdown):** Implemented a "Parked" state. On startup, shutdown, or if `explorer.exe` crashes, the widget unobtrusively "docks" to the bottom screen edge (2px visible) instead of floating at (0,0) or vanishing. It slides up automatically when the Taskbar is ready.

*   **Integrated Multi-State Handling:**
    *   **Smart "Game Mode" Detection:** Implemented a passive hybrid system (OS Signal + Geometry Check) to detect D3D Fullscreen and Borderless games. This allows the widget to hide during games without triggering Anti-Cheat software.
    *   **Master Toggle:** Added `EnableGameDetection` to completely disable the hiding logic if desired.
    *   **Bidirectional Transitions:** Added a toggleable **Slide Animation**. The widget smoothly slides down off-screen when entering a game and slides back up when returning to the desktop.
    *   **App Whitelist:** Added an **`IgnoredApps`** setting. This allows specific apps (like `firefox.exe`) to run in Fullscreen (F11) while keeping the widget visible.

*   **Stability & Cleanup:**
    *   **True High-DPI Support:** Added dynamic `ScaleTransform` logic combined with `SetProcessDpiAwarenessContext`. The UI now automatically scales its size and text to look crisp and correctly proportioned on 4K/1440p monitors.
    *   **Settings Layout:** Reorganized the settings menu to flow logically (Visuals -> Transitions -> Detection -> Exceptions).
    *   **Optimization:** Modified the internal loops to reduce CPU overhead when idle.
    *   **Dependencies:** Added `-lshell32` and `-lpsapi` to the compiler options to support the new detection and whitelist features.

[mod.wh.cpp.zip](https://github.com/user-attachments/files/24324266/mod.wh.cpp.zip)

------------------------------------------------------------------------
**Personal TODOs:**
_Here you'll find ideas I'll eventually work on with the same ai, whether the dev gets to em first is up to him, This is what I'll likely be working on next(obviously probably not all at once), Going in order of priority._
- Replace the hex input "Manual Text Color" to allow RGB or possibly RGBA, For simplicity sake.(Also is this thing even working right? I can't find a hex online that comes out the same color lol)
- Add rounded corners for Album cover. As requested by another user. I much agree.
- Smarter game detection, ie if it knows I'm playing a game, keep that in memory for X seconds so that it doesn't hog up any resources(possibly unnecessary, idk how to judge the performance soo hard to tell)
- The Smart docking feature via the Explorer crashing should possibly also be a toggle(Depending on user feedback). I realized this after the fact but to me it makes since as I mean it's in the name, **Taskbar** Music Lounge
- Vertical taskbar support(requires possible full redesign, The current "Panel Width/Height" adds complexity to this.) Would like for this to be a simple toggle but in doing so it'd change the Panel Width/Height thus leading to a conundrum. Are these able to be overridden even?

Cinabutts:
Hey @Hashah2311 thanks for pushing some of my code, here's my latest version if you're interested. It's main changes is regarding the background, in this version you are being able to set a custom background color when auto hide is off, or if you leave it default 0,0,0 it sticks with the original light/dark theme and i've added the option to invert the mods perceived mode. 
**NOTE**: It is a bit clunky though. Could be better. And this version is still based on my own fork, not the updated version above fyi.

My next plan is to add a toggle-able slim progress bar at the top so when the mod auto hides it still shows this progress bar, thus in my mind helps negate the fact that when the taskbar is hidden via auto hide and the widget gets docked, when I hover over the widget it doesn't allow the taskbar to reappear.(maybe that'll be fixed in the later versions?)
This progress bar would always be visible whilst toggled on fyi so for instance I could see the progress in-game.

[mod.wh.cpp.zip](https://github.com/user-attachments/files/24334056/mod.wh.cpp.zip)

image
image
image

SpacEagle17:
I personally don't think a progress bar for auto hiding is necessary at all. You can keep such changes for your local edit

Cinabutts:
> I personally don't think a progress bar for auto hiding is necessary at all. You can keep such changes for your local edit

probably right, in my mind it just makes since because when it's docked and the taskbar is hidden the widget as said is docked so it still shows a few pixels where it once was, hovering over this whilst auto-hide is on doesn't allow the taskbar to reappear, soo might as well make it useful i feel. but i get it. kinda visually distracting, thus why i said it'd be toggle-able.

PR #2750, opened 2025-11-28
Add Two-sided Snapping mod for Windows 11

Anixx:
This mod disables corner snapping while enabling two-sided snapping in Windows 11, restoring the behavior of Windows 10.

### DIFF
--- a/mods/two-sided-snapping.wh.cpp
+++ b/mods/two-sided-snapping.wh.cpp
@@ -28,7 +28,7 @@ HRESULT WINAPI windowsudkshellcommon_SLGetWindowsInformationDWORDHook(PCWSTR pws
 {
     HRESULT hr = SLGetWindowsInformationDWORDFunc(pwszValueName, pdwValue);
 
-    if (SUCCEEDED(hr) &&(!wcsncmp(pwszValueName, L"Shell-Windowing-LimitSnappedWindows", 35)))
+    if (SUCCEEDED(hr) && !wcscmp(pwszValueName, L"Shell-Windowing-LimitSnappedWindows"))
         *pdwValue = 1;
 
     return hr;


mods/two-sided-snapping.wh.cpp
```diff
@@ -0,0 +1,47 @@
...
+*/
+// ==/WindhawkModReadme==
+
+static HRESULT(WINAPI *SLGetWindowsInformationDWORDFunc)(PCWSTR pwszValueName, DWORD* pdwValue) = NULL;
+
+HRESULT WINAPI windowsudkshellcommon_SLGetWindowsInformationDWORDHook(PCWSTR pwszValueName, DWORD* pdwValue)
+{
+    HRESULT hr = SLGetWindowsInformationDWORDFunc(pwszValueName, pdwValue);
+
+    if (!wcsncmp(pwszValueName, L"Shell-Windowing-LimitSnappedWindows", 36))
```
m417z: Any reason to use `wcsncmp` and not `wcscmp` here?

Anixx: Possibly, because the function can be called with incorrect data, not a null-terminating string. This was the choice in the original EP code.

Anixx: P.S. I just checked the original and it uses 35 character comparison, not 36. Perhaps, the last character can vary, so we compare not to the last character.

m417z: I see that that's the original code (https://github.com/valinet/ExplorerPatcher/pull/1080), but it doesn't look correct. This usage of `wcsncmp` means that any prefix will match, for example `Shell-Windowing-LimitSnappedWindows!@#$%`

Also, I'd add a condition to check that the function actually succeeded, otherwise you might be accessing an invalid string.

So basically:

```cpp
if (SUCCEEDED(hr) && !wcscmp(pwszValueName, L"Shell-Windowing-LimitSnappedWindows"))
```

Anixx: My guess is the string's end may vary by Windows version, in that case it makes sense to check only the beginning. I also added to the description that this mod is intended only for Win11 (the check for Win11 was in the original code plus Win10 users report it does not work for them).

Anixx: P.S. After reading the link you provided, I agree that this must be a mistake, so I fixed the code following your suggestion.

PR #2773, opened 2025-12-06
F1 Blocker 0.3

d0gkiller87:


mods/f1-blocker.wh.cpp
```diff
@@ -2,45 +2,109 @@
...
+    ) {
+      return TRUE;
+    }
+    return TranslateAcceleratorW_Original( hWnd, hAccTable, lpMsg );
+  }
+
+  using ShellExecuteW_t = decltype( &ShellExecuteW );
+  ShellExecuteW_t ShellExecuteW_Original;
+
+  BOOL WINAPI ShellExecuteW_Hook( HWND hWnd, LPCWSTR lpOperation, LPCWSTR lpFile, LPCWSTR lpParameters, LPCWSTR lpDirectory, INT nShowCmd ) {
```
m417z: `ShellExecuteW` return `HINSTANCE`.

d0gkiller87: Nice catch👀

PR #2775, opened 2025-12-07
Add Classic Explorer Status Bar mod

Anixx:
This mod adds a classic status bar to Windows Explorer, displaying item count, free disk space, and total size of selected files.


Anixx:
Seems a check bug, compiles well locally.

PR #2781, opened 2025-12-09
Add virtual desktop helper mod

u2x1:
## Summary
  - add the new `virtual-desktop-helper` Windhawk mod that mirrors [VD.ahk](https://github.com/FuPeiJiang/VD.ahk/tree/v2_port#readme) functionality in C++
  - expose configurable hotkeys for moving the focused window or switching desktops (1‑9) with per-build COM GUID handling

  ## Testing
  - tested on `Windows 11 26100.7171`

### DIFF
--- a/mods/virtual-desktop-helper.wh.cpp
+++ b/mods/virtual-desktop-helper.wh.cpp
@@ -5,7 +5,7 @@
 // @version         1.0.0
 // @author          u2x1
 // @github          https://github.com/u2x1
-// @include         explorer.exe
+// @include         windhawk.exe
 // @compilerOptions -lole32 -loleaut32 -luuid
 // ==/WindhawkMod==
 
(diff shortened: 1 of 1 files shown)


mods/virtual-desktop-helper.wh.cpp
```diff
@@ -0,0 +1,814 @@
...
+//=============================================================================
+
+BOOL Wh_ModInit() {
+    Wh_Log(L"Virtual Desktop Helper mod initializing...");
+
+    // Load settings first
+    LoadSettings();
+
+    // Initialize COM
+    HRESULT hr = CoInitializeEx(nullptr, COINIT_APARTMENTTHREADED);
```
m417z: This doesn't look correct. It makes more sense to initialize COM in the hotkey thread, since that's where it's used.

u2x1:
works on my machine with these changes

PR #2802, opened 2025-12-14
Change Qt App Icon 1.0.0

aubymori:
This mod allows you to change the window icon of all windows for a Qt application.

**Example**: *IDA 9.2 with its pre-6.0 icon*
Example: IDA 9.2 with its pre-6.0 icon

### DIFF
--- a/mods/change-qt-app-icon.wh.cpp
+++ b/mods/change-qt-app-icon.wh.cpp
@@ -71,7 +71,6 @@ LRESULT WINAPI SendMessageW_hook(
                     GetCurrentProcess(), hQWindows, &mi, sizeof(mi));
                 s_uQWindowsBase = (ULONG_PTR)mi.lpBaseOfDll;
                 s_uQWindowsEnd  = (ULONG_PTR)mi.lpBaseOfDll + mi.SizeOfImage;
-                FreeLibrary(hQWindows);
             }
         }
 


mods/change-qt-app-icon.wh.cpp
```diff
@@ -0,0 +1,159 @@
...
+used at all times for an application.
+
+**Example**: *IDA 9.2 with its pre-6.0 icon*
+
+![Example: IDA 9.2 with its pre-6.0 icon](https://raw.githubusercontent.com/aubymori/images/refs/heads/main/change-qt-app-icon-example.png)
+
+## How to use
+1. Go to the "Advanced" tab of this mod.
+2. Find the "Custom process inclusion list" option.
+3. Add the name of the EXE you want to disable tray icons for.
```
m417z: "disable tray icons for" - leftover from a different mod.

mods/change-qt-app-icon.wh.cpp
```diff
@@ -0,0 +1,159 @@
...
+        {
+            HMODULE hQWindows = GetModuleHandleW(L"qwindows.dll");
+            if (hQWindows)
+            {
+                MODULEINFO mi;
+                GetModuleInformation(
+                    GetCurrentProcess(), hQWindows, &mi, sizeof(mi));
+                s_uQWindowsBase = (ULONG_PTR)mi.lpBaseOfDll;
+                s_uQWindowsEnd  = (ULONG_PTR)mi.lpBaseOfDll + mi.SizeOfImage;
+                FreeLibrary(hQWindows);
```
m417z: This looks like a mistake.

mods/change-qt-app-icon.wh.cpp
```diff
@@ -0,0 +1,159 @@
...
+        }
+
+        if (s_uQWindowsBase)
+        {
+            /* Only change the icon if the call is coming from qwindows.dll.
+               That way, we don't mess up things like the native open/save dialogs. */
+            ULONG_PTR uRetAddr = (ULONG_PTR)__builtin_return_address(0);
+            if (uRetAddr >= s_uQWindowsBase && uRetAddr <= s_uQWindowsEnd)
+            {
+                DestroyIcon((HICON)lParam);
```
m417z: I don't think `WM_SETICON` moves ownership to Windows, are you sure Qt doesn't use/free that icon later on?

mods/change-qt-app-icon.wh.cpp
aubymori: Fixed.

mods/change-qt-app-icon.wh.cpp
aubymori: How so?

mods/change-qt-app-icon.wh.cpp
aubymori: Now that I think about it, it's probably better to leave the icon in memory and let Qt free it itself rather than having it make an erroneous `DestroyIcon` call. Fixed.

mods/change-qt-app-icon.wh.cpp
m417z: You used `GetModuleHandle`, not `LoadLibrary`. `GetModuleHandle` returns a handle to a mapped module without incrementing its reference count.

aubymori: Ah, didn't know it didn't increase the ref count. Fixed.

PR #2803, opened 2025-12-15
Add Mod: Windows 11 Better Ultrawide Taskbar

roeseth:
Add a mod to center taskbar and system tray as a single unit to provide better usability in ultrawide screen.
Also offer an offset only mode to just offset the system tray.

### DIFF
--- a/mods/windows-11-beter-ultrawide-taskbar.wh.cpp
+++ b/mods/windows-11-beter-ultrawide-taskbar.wh.cpp
@@ -1,5 +1,5 @@
 // ==WindhawkMod==
-// @id              windows-11-beter-ultrawide-taskbar.wh
+// @id              windows-11-beter-ultrawide-taskbar
 // @name            Windows 11 Better Ultrawide Taskbar
 // @description     Centers the taskbar and system tray as a single unit
 // @version         1.0
--- a/mods/windows-11-beter-ultrawide-taskbar.wh.cpp
+++ b/mods/windows-11-beter-ultrawide-taskbar.wh.cpp
@@ -2,13 +2,18 @@
 // @id              windows-11-beter-ultrawide-taskbar
 // @name            Windows 11 Better Ultrawide Taskbar
 // @description     Centers the taskbar and system tray as a single unit
-// @version         1.0
+// @version         1.1
 // @author          Molko
 // @github          https://github.com/roeseth
 // @include         explorer.exe
 // @architecture    x86-64
-// @compilerOptions -lcomctl32 -lole32 -loleaut32 -lruntimeobject -Wl,--export-all-symbols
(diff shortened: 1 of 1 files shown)


m417z:
Thanks for the submission. Here's how it looks on my computer with the default settings, I assume the edges shouldn't look like this:

image

Also, you're using XAML Diagnostics (`InitializeXamlDiagnosticsEx`), which makes the mod incompatible with other mods or customization tools that use it. Specifically, it makes the mod incompatible with [Windows 11 Taskbar Styler](https://windhawk.net/mods/windows-11-taskbar-styler).

You can keep it as is, in which case I think it'd be a good idea to add a notice about it in the readme. Alternatively, you might want to update the implementation such that XAML Diagnostics API isn't used. There are several mod examples which can serve as an example, for example [Start button always on the left](https://windhawk.net/mods/taskbar-start-button-position):
https://github.com/ramensoftware/windhawk-mods/blob/862ba36c9284acfd3cd617faffe37aea501db928/mods/taskbar-start-button-position.wh.cpp#L390-L407

mods/windows-11-beter-ultrawide-taskbar.wh.cpp
```diff
@@ -0,0 +1,1043 @@
...
+
+    // Clean up any pending LayoutUpdated event subscription
+    if (g_layoutUpdatedToken)
+    {
+        auto taskbarFrame = g_taskbarFrameWeakRef.get();
+        if (taskbarFrame)
+        {
+            try
+            {
+                taskbarFrame.LayoutUpdated(g_layoutUpdatedToken);
```
m417z, line 1004: This should be run from the taskbar (UI) thread.

Something like:
```cpp
HWND hTaskbarWnd = FindCurrentProcessTaskbarWnd();
if (hTaskbarWnd)
{
    RunFromWindowThread(
        hTaskbarWnd, [](void *pParam)
        { /* cleanup code here */ }, 0);
}
```

Also generally, I suggest to at least print errors instead of having empty `catch` blocks.

TheGamer1445891:
@roeseth i found a bug!
when you change the position of the tasklist buttons, the system tray glitches and is now blocking the taskbar! (only happens if the system tray is flipped (using taskbar styler mod))

https://github.com/user-attachments/assets/94d7f1f7-e68f-4502-a27d-f9a0f1c769e9

roeseth:
Hi @TheGamer1445891，
Yea I just freshly re-installed a new 24H2 which broke my mod so I was working on fixing other compatibility issues by rewriting the code with some more features. Could you tell me what taskbar styler settings are you using and the exact step I should do to reproduce the issue?

TheGamer1445891:
@crazyboyybs you should share your fix here (i forgot to ping you a long time ago)

PR #2815, opened 2025-12-18
linux-style-window-drag.wh.cpp

djestick:
Implements Linux-style window moving (Win + Left Click drags window from anywhere).

### DIFF
--- a/mods/linux-style-window-drag.wh.cpp
+++ b/mods/linux-style-window-drag.wh.cpp
@@ -1,0 +1,148 @@
+// ==WindhawkMod==
+// @id              linux-style-window-drag
+// @name            Linux-style Window Drag
+// @description     Enables moving windows by clicking anywhere while holding the Win key, similar to GNOME or KDE. Includes a fix for the Start Menu.
+// @version         1.0
+// @author          djestick
+// @github          https://github.com/djestick/linux-style-window-drag-on-windows
+// @include         explorer.exe
+// @compilerOptions -luser32
+// ==/WindhawkMod==
+
+#include <windows.h>
+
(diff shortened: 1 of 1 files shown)


mods/linux-style-window-drag.wh.cpp
```diff
@@ -0,0 +1,148 @@
+// ==WindhawkMod==
+// @id              linux-style-window-drag
+// @name            Linux-style Window Drag
+// @description     Enables moving windows by clicking anywhere while holding the Win key, similar to GNOME or KDE. Includes a fix for the Start Menu.
+// @version         1.0
+// @author          djestick
+// @github          https://github.com/djestick/linux-style-window-drag-on-windows
+// @include         explorer.exe
```
m417z, line 8: Please see https://github.com/ramensoftware/windhawk/wiki/Mods-as-tools:-Running-mods-in-a-dedicated-process

mods/linux-style-window-drag.wh.cpp
```diff
@@ -0,0 +1,148 @@
...
+                        }
+                    }
+                }
+            }
+        }
+
+        // 2. ДВИЖЕНИЕ
+        if (wParam == WM_MOUSEMOVE && g_isDragging) {
+            bool isWinDown = (GetAsyncKeyState(VK_LWIN) & 0x8000) || 
+                             (GetAsyncKeyState(VK_RWIN) & 0x8000);
```
m417z, line 93: How about adding an option to customize the hotkey? I think Alt is a more common key for this, for example in [AltDrag](https://stefansundin.github.io/altdrag/).

PR #2819, opened 2025-12-19
feat(virtual-desktop-helper): v2.0.0 with tiling and pin support

u2x1:
- Add window tiling (manually triggered by hotkey)
- Add window pinning to all desktops
- Add previous desktop toggle
- Add per-desktop focus/layout memory

### DIFF
--- a/mods/virtual-desktop-helper.wh.cpp
+++ b/mods/virtual-desktop-helper.wh.cpp
@@ -1254,12 +1254,7 @@ void StopHotkeyThread() {
     PostThreadMessage(g_threadId, WM_QUIT, 0, 0);
   }
   if (g_hThread) {
-    // Wait longer (3 seconds) since COM calls can be slow
-    DWORD waitResult = WaitForSingleObject(g_hThread, 3000);
-    if (waitResult == WAIT_TIMEOUT) {
-      Wh_Log(L"Warning: Hotkey thread did not exit in time, terminating");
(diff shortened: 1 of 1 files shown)


mods/virtual-desktop-helper.wh.cpp
```diff
@@ -1,819 +1,1300 @@
...
+  
+  if (g_threadId) {
+    PostThreadMessage(g_threadId, WM_QUIT, 0, 0);
+  }
+  if (g_hThread) {
+    // Wait longer (3 seconds) since COM calls can be slow
+    DWORD waitResult = WaitForSingleObject(g_hThread, 3000);
+    if (waitResult == WAIT_TIMEOUT) {
+      Wh_Log(L"Warning: Hotkey thread did not exit in time, terminating");
+      TerminateThread(g_hThread, 0);
```
m417z: `TerminateThread` is not a safe operation, and it's preferable to avoid it if possible. See:
https://devblogs.microsoft.com/oldnewthing/20150814-00/?p=91811

Did you add it because it's slow, or because it becomes stuck forever?

u2x1: it's a workaround in case something is still preventing the mod from uninitializing... but i've added a few fixes to the hotkey thread, so maybe we could remove it.

u2x1: and it did get stuck forever before

PR #2825, opened 2025-12-20
Dark mode context menus 1.3.3

MGGSK:
- Fixed https://github.com/MGGSK/DarkMenus/issues/16
- Fixed https://github.com/MGGSK/DarkMenus/issues/14

### DIFF
--- a/mods/dark-menus.wh.cpp
+++ b/mods/dark-menus.wh.cpp
@@ -272,9 +272,15 @@ LRESULT CALLBACK DefWindowProcW_Hook(HWND hWnd, UINT uMsg, WPARAM wParam, LPARAM
 decltype(&SetMenuInfo) SetMenuInfo_Original;
 WINBOOL WINAPI SetMenuInfo_Hook(HMENU hMenu, LPCMENUINFO lpInfo)
 {
-    LPMENUINFO lpMenuInfo = const_cast<LPMENUINFO>(lpInfo);
-    lpMenuInfo->hbrBack = nullptr; //Disable custom menu backgrounds because they are broken in dark mode. (See https://github.com/MGGSK/DarkMenus/issues/16)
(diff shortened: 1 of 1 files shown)


mods/dark-menus.wh.cpp
```diff
@@ -269,6 +269,14 @@ LRESULT CALLBACK DefWindowProcW_Hook(HWND hWnd, UINT uMsg, WPARAM wParam, LPARAM
     return DefWindowProcW_Original(hWnd, uMsg, wParam, lParam);
 }
 
+decltype(&SetMenuInfo) SetMenuInfo_Original;
+WINBOOL WINAPI SetMenuInfo_Hook(HMENU hMenu, LPCMENUINFO lpInfo)
+{
+    LPMENUINFO lpMenuInfo = const_cast<LPMENUINFO>(lpInfo);
```
m417z: I'd avoid this cast, since the data might reside in read-only memory, which will cause a crash.

How about:
```cpp
decltype(&SetMenuInfo) SetMenuInfo_Original;
WINBOOL WINAPI SetMenuInfo_Hook(HMENU hMenu, LPCMENUINFO lpInfo)
{
    //Disable custom menu backgrounds because they are broken in dark mode. (See https://github.com/MGGSK/DarkMenus/issues/16)
    alignas(MENUINFO) BYTE buffer[256];
    if (!(lpInfo->fMask & MIM_BACKGROUND) || lpInfo->cbSize > sizeof(buffer)) {
        return SetMenuInfo_Original(hMenu, lpInfo);
    }

    memcpy(buffer, lpInfo, lpInfo->cbSize);
    LPMENUINFO pCopy = reinterpret_cast<LPMENUINFO>(buffer);
    pCopy->fMask &= ~MIM_BACKGROUND;
    return SetMenuInfo_Original(hMenu, pCopy);
}
```

PR #2840, opened 2025-12-22
Dark mode context menus 1.4

MGGSK:
- Fixed https://github.com/MGGSK/DarkMenus/issues/18
- Added support for mdi windows
- Code cleanup
- Improved readme

### DIFF
--- a/mods/dark-menus.wh.cpp
+++ b/mods/dark-menus.wh.cpp
@@ -237,6 +237,8 @@ bool CALLBACK UAHWndProc(HWND hWnd, UINT uMsg, WPARAM wParam, LPARAM lParam, LRE
                 glyph = L'\ue923'; //Maximize icon
             else if (offset == 1)
                 glyph = L'\ue8bb'; //Close icon
+            else
+                return false;
 
             HFONT hOldFont = SelectFont(drawingInfo->um.hdc, g_iconFont);
 
@@ -256,12 +258,14 @@ bool CALLBACK UAHWndProc(HWND hWnd, UINT uMsg, WPARAM wParam, LPARAM lParam, LRE
     }
 
     case WM_THEMECHANGED:
+    case WM_DESTROY:
         if (g_menuBarTheme)
         {
             CloseThemeData(g_menuBarTheme);
             g_menuBarTheme = nullptr;
         }
         return false;
+        
     default:
         return false;
     }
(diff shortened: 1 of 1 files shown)


mods/dark-menus.wh.cpp
```diff
@@ -278,9 +380,9 @@ WINBOOL WINAPI SetMenuInfo_Hook(HMENU hMenu, LPCMENUINFO lpInfo)
         return SetMenuInfo_Original(hMenu, lpInfo);
 
     memcpy(buffer, lpInfo, lpInfo->cbSize);
-    LPMENUINFO pCopy = reinterpret_cast<LPMENUINFO>(buffer);
-    pCopy->hbrBack = brBackground;
-    return SetMenuInfo_Original(hMenu, pCopy);
+    auto* infoCopy = reinterpret_cast<LPMENUINFO>(buffer);
+    infoCopy->hbrBack = CreateSolidBrush(DARK_MENU_COLOR); //Fixes https://github.com/MGGSK/DarkMenus/issues/18
```
m417z, line 388: This seem to create a new brush every time, and it's never freed. There's already `DARK_MENU_BRUSH`, why not use it?

mods/dark-menus.wh.cpp
```diff
@@ -138,98 +149,126 @@ void DrawUAHMenuNCBottomLine(HWND hWnd)
...
         return true;
     }
+
     case WM_THEMECHANGED:
-    {
-        CloseThemeData(g_menuTheme);
-        g_menuTheme = nullptr;
+        if (g_menuBarTheme)
+        {
+            CloseThemeData(g_menuBarTheme);
```
m417z, line 264: How about closing it on `WM_DESTROY` too? Otherwise there's a leak for each window.

mods/dark-menus.wh.cpp
```diff
@@ -138,98 +149,126 @@ void DrawUAHMenuNCBottomLine(HWND hWnd)
...
+
+            wchar_t glyph;
+            if (drawingInfo->umi.iPosition == 0)
+                glyph = L'\ue700'; //Menu icon
+            else if (offset == 3)
+                glyph = L'\ue921'; //Minimize icon
+            else if (offset == 2)
+                glyph = L'\ue923'; //Maximize icon
+            else if (offset == 1)
+                glyph = L'\ue8bb'; //Close icon
```
m417z, line 239: What if none of the conditions are true? Perhaps better to return in this case.

mods/dark-menus.wh.cpp
MGGSK: Hi @m417z. I am creating a new brush because reusing it here causes https://github.com/MGGSK/DarkMenus/issues/18.

PR #2844, opened 2025-12-23
Dark mode context menus 1.4.1

MGGSK:
- Fixed https://github.com/MGGSK/DarkMenus/issues/19

### DIFF
--- a/mods/dark-menus.wh.cpp
+++ b/mods/dark-menus.wh.cpp
@@ -515,11 +515,13 @@ void Wh_ModUninit()
         DeleteObject(DARK_MENU_ITEM_BACKGROUND_HOVER);
         DeleteObject(DARK_MENU_ITEM_BACKGROUND_SELECTED);
 
-        DeleteObject(g_iconFont);
+		if(g_iconFont)
+            DeleteObject(g_iconFont);
         CloseThemeData(g_menuBarTheme);
         FreeLibrary(g_hUxtheme);
     }
 }
+
 
 
 


mods/dark-menus.wh.cpp
```diff
@@ -338,6 +368,7 @@ LRESULT CALLBACK DefWindowProcW_Hook(HWND hWnd, UINT uMsg, WPARAM wParam, LPARAM
 decltype(&DefFrameProcA) DefFrameProcA_Original;
 LRESULT CALLBACK DefFrameProcA_Hook(HWND hWnd, HWND hMdiClient, UINT uMsg, WPARAM wParam, LPARAM lParam)
 {
+    return DefFrameProcA_Original(hWnd, hMdiClient, uMsg, wParam, lParam);
```
m417z: This, and `return false;` above, look like temporary changes.

MGGSK: > This, and `return false;` above, look like temporary changes.

Thanks. That was a temporary change that I forgot to remove.

m417z: What about this line that I commented on?
```cpp
return DefFrameProcA_Original(hWnd, hMdiClient, uMsg, wParam, lParam);
```

mods/dark-menus.wh.cpp
```diff
@@ -240,6 +264,13 @@ bool CALLBACK UAHWndProc(HWND hWnd, UINT uMsg, WPARAM wParam, LPARAM lParam, LRE
             else
                 return false;
 
+            if(!g_iconFont)
+            {
+                g_iconFont = CreateFontW(10, NULL, NULL, NULL, FW_NORMAL, FALSE, FALSE, FALSE,
```
m417z, line 269: This might never run, but you're freeing `g_iconFont` unconditionally on mod unload.

Issue #1704, opened 2025-03-23
[Windows 11 Taskbar Styler] - Incorrect window handle found by class name Shell_TrayWnd


Video-Nomad:
Windhawk gets the taskbar handle by using:

https://github.com/ramensoftware/windhawk-mods/blob/9e22c833002db2e8c8c002ea098d25805d09167c/mods/windows-11-taskbar-styler.wh.cpp#L2807 

which simply returns the topmost window with this class name. This can be an incorrect handle if there is another software that is using this class name for custom system tray functionality. For example [YASB](https://github.com/amnweb/yasb), [Zebar ](https://github.com/glzr-io/zebar) and anything made with [Managed Shell](https://github.com/cairoshell/ManagedShell) will have this issue as they all create a window with `Shell_TrayWnd` class name to intercept system tray messages for their custom widgets.

A better way would be to use `FindWindowEx` and check if the process name is `explorer.exe`. If it's not, just find the next window with this class name. 

Python example:
```py
import ctypes
from ctypes import wintypes

user32 = ctypes.windll.user32
kernel32 = ctypes.windll.kernel32
psapi = ctypes.windll.psapi

FindWindowEx = user32.FindWindowExW
FindWindowEx.argtypes = [
    wintypes.HWND,
    wintypes.HWND,
    wintypes.LPCWSTR,
    wintypes.LPCWSTR,
]
FindWindowEx.restype = wintypes.HWND

PROCESS_QUERY_INFORMATION = 0x0400
PROCESS_VM_READ = 0x0010

TARGET_CLASS = "Shell_TrayWnd"
TARGET_PROCESS = "explorer.exe"


def get_process_name(hwnd: int) -> str | None:
    """Retrieve the process name given a window handle."""
    pid = wintypes.DWORD()
    user32.GetWindowThreadProcessId(hwnd, ctypes.byref(pid))
    h_process: str = kernel32.OpenProcess(
        PROCESS_QUERY_INFORMATION | PROCESS_VM_READ, False, pid.value
    )
    if not h_process:
        return None
    buffer = ctypes.create_unicode_buffer(512)
    if psapi.GetModuleBaseNameW(h_process, None, buffer, 512) == 0:
        kernel32.CloseHandle(h_process)
        return None
    process_name: str = buffer.value  # pyright: ignore[reportAny]
    kernel32.CloseHandle(h_process)
    return process_name


def find_windows_by_class(class_name: str) -> dict[int, str]:
[...]

PR #2872, opened 2025-12-27
Add Taskbar Start Button Corner Fix mod

alchemyyy:
This mod fixes an issue on multi-monitor setups with left-aligned start menus where it seems sticky corners, or perhaps something else, occludes the start menu button in the very corner of the screen.

There is a good amount of code here dedicated to hooking a pane on the taskbar correctly (waiting for it to exist), and ensuring this is only ever done once, in a way that avoids endless polling. 

The mod works via a mouse hook installed into this magical root InputSite pane on the taskbar, and checking if a mouse down event happens in the corner of the screen, where the issue is.

### DIFF
--- a/mods/taskbar-start-button-corner-fix.wh.cpp
+++ b/mods/taskbar-start-button-corner-fix.wh.cpp
@@ -77,31 +77,74 @@ bool g_menuOpenAtMouseDown = false;
 // UI Automation (initialized on InputSite thread)
 IUIAutomation* g_pAutomation = nullptr;
 bool g_comInitialized = false;
-HWND g_cleanupWindow = nullptr;
-const wchar_t* g_cleanupWindowClass = L"StartButtonFix_Cleanup";
+HWND g_inputSiteWnd = nullptr;
 
-#define WM_CLEANUP_COM (WM_USER + 100)
+// Helper to run code on a window's thread
+using RunFromWindowThreadProc_t = void(WINAPI*)(void* parameter);
 
-// Cleanup window procedure - runs on InputSite thread
-LRESULT CALLBACK CleanupWindowProc(HWND hWnd, UINT uMsg, WPARAM wParam, LPARAM lParam) {
-    switch (uMsg) {
-        case WM_CLEANUP_COM:
-            if (g_pAutomation) {
-                g_pAutomation->Release();
-                g_pAutomation = nullptr;
-                if (g_comInitialized) {
-                    CoUninitialize();
-                    g_comInitialized = false;
-                }
-                Wh_Log(L"UI Automation cleaned up on InputSite thread");
-            }
-            return 0;
+bool RunFromWindowThread(HWND hWnd,
+                         RunFromWindowThreadProc_t proc,
+                         void* procParam) {
+    static const UINT runFromWindowThreadRegisteredMsg =
+        RegisterWindowMessage(L"Windhawk_RunFromWindowThread_" WH_MOD_ID);
 
-        case WM_CLOSE:
-            DestroyWindow(hWnd);
-            return 0;
+    struct RUN_FROM_WINDOW_THREAD_PARAM {
+        RunFromWindowThreadProc_t proc;
+        void* procParam;
+    };
+
+    DWORD dwThreadId = GetWindowThreadProcessId(hWnd, nullptr);
+    if (dwThreadId == 0) {
+        return false;
     }
-    return DefWindowProc(hWnd, uMsg, wParam, lParam);
+
+    if (dwThreadId == GetCurrentThreadId()) {
+        proc(procParam);
+        return true;
+    }
+
+    HHOOK hook = SetWindowsHookEx(
+        WH_CALLWNDPROC,
+        [](int nCode, WPARAM wParam, LPARAM lParam) -> LRESULT {
+            if (nCode == HC_ACTION) {
+                const CWPSTRUCT* cwp = (const CWPSTRUCT*)lParam;
+                static const UINT msg =
+                    RegisterWindowMessage(L"Windhawk_RunFromWindowThread_" WH_MOD_ID);
+                if (cwp->message == msg) {
+                    RUN_FROM_WINDOW_THREAD_PARAM* param =
+                        (RUN_FROM_WINDOW_THREAD_PARAM*)cwp->lParam;
+                    param->proc(param->procParam);
+                }
+            }
+
+            return CallNextHookEx(nullptr, nCode, wParam, lParam);
+        },
+        nullptr, dwThreadId);
+    if (!hook) {
+        return false;
+    }
+
+    RUN_FROM_WINDOW_THREAD_PARAM param;
+    param.proc = proc;
+    param.procParam = procParam;
+    SendMessage(hWnd, runFromWindowThreadRegisteredMsg, 0, (LPARAM)&param);
+
+    UnhookWindowsHookEx(hook);
+
+    return true;
+}
+
+// Cleanup UI Automation (called on InputSite thread)
+void WINAPI CleanupUIAutomation(void*) {
+    if (g_pAutomation) {
+        g_pAutomation->Release();
+        g_pAutomation = nullptr;
+        if (g_comInitialized) {
+            CoUninitialize();
+            g_comInitialized = false;
+        }
+        Wh_Log(L"UI Automation cleaned up on InputSite thread");
+    }
 }
 
 // Initialize UI Automation (call from InputSite thread only)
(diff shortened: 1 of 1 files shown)


mods/taskbar-start-button-corner-fix.wh.cpp
```diff
@@ -0,0 +1,431 @@
...
+    }
+    CleanupUIAutomation();
+    Wh_Log(L"Uninitialized");
+}
+
+BOOL Wh_ModSettingsChanged(BOOL* bReload) {
+    LoadSettings();
+    *bReload = FALSE;
+    return TRUE;
+}
```
m417z: The simpler variant can be used here:
```suggestion
void Wh_ModSettingsChanged() {
    LoadSettings();
}
```

mods/taskbar-start-button-corner-fix.wh.cpp
```diff
@@ -0,0 +1,431 @@
...
+    if (g_mouseHook) {
+        UnhookWindowsHookEx(g_mouseHook);
+        g_mouseHook = NULL;
+    }
+    if (g_taskbarMutex) {
+        ReleaseMutex(g_taskbarMutex);
+        CloseHandle(g_taskbarMutex);
+        g_taskbarMutex = NULL;
+    }
+    CleanupUIAutomation();
```
m417z: `Wh_ModInit` and `Wh_ModUninit` might run in different threads, see [Mod lifetime](https://github.com/ramensoftware/windhawk/wiki/Mod-lifetime). `CoInitializeEx` only applies for the `Wh_ModInit` thread.

And generally, it's better to initialize and uninitialize COM objects which are used in the UI thread in that thread.

mods/taskbar-start-button-corner-fix.wh.cpp
```diff
@@ -0,0 +1,431 @@
...
+            // Taskbar is ours - now find InputSite
+            inputSiteWnd = NULL;
+            EnumChildWindows(taskbarWnd, FindInputSiteCallback, (LPARAM)&inputSiteWnd);
+
+            if (inputSiteWnd) {
+                break;  // Found both taskbar and InputSite
+            }
+        }
+
+        Sleep(500);
```
m417z: An alternative approach to avoid pooling is to hook `CreateWindowInBand`:
https://github.com/ramensoftware/windhawk-mods/blob/f5c9e38460918ba1fa4b643dedab4ecf42faf03c/mods/taskbar-volume-control.wh.cpp#L1552-L1597

Your approach is fine, too, just letting you know.

BTW, your implementation doesn't seem to handle secondary taskbars. Not sure if they have the same issue.

mods/taskbar-start-button-corner-fix.wh.cpp
alchemyyy: Done! Thanks!

mods/taskbar-start-button-corner-fix.wh.cpp
alchemyyy: Fixed!

mods/taskbar-start-button-corner-fix.wh.cpp
```diff
@@ -366,66 +309,47 @@ void LoadSettings() {
...
     }
 
-    // Spawn worker thread to wait for taskbar and install hook
-    g_workerThread = CreateThread(NULL, 0, WorkerThread, NULL, 0, NULL);
-    if (!g_workerThread) {
-        Wh_Log(L"Failed to create worker thread");
-        return FALSE;
+    // Check for existing InputSite window (taskbar may already exist)
+    HWND taskbarWnd = FindWindowW(L"Shell_TrayWnd", NULL);
+    if (taskbarWnd) {
```
m417z, line 423: Make sure that it's part of the current process. You can use a function like this to account for [other programs that use this class name](https://github.com/ramensoftware/windhawk-mods/issues/1704#issuecomment-2746369867):

https://github.com/ramensoftware/windhawk-mods/blob/f5c9e38460918ba1fa4b643dedab4ecf42faf03c/mods/taskbar-auto-hide-per-monitor.wh.cpp#L228-L247

mods/taskbar-start-button-corner-fix.wh.cpp
```diff
@@ -174,6 +161,7 @@ bool IsStartMenuOpen() {
 
 // Toggle Start menu via UI Automation
 void ToggleStartMenu() {
+    InitUIAutomation();
```
m417z: Automation is never released now. Is it a deliberate leak?

mods/taskbar-start-button-corner-fix.wh.cpp
alchemyyy: Thank you so much, updated to use this. Works fantastically.

For whatever bizarre reason, this isn't an issue with secondary taskbars. I tested this on a 3 monitor setup with a secondary taskbar on the right-most monitor, and it all seems to work fine.

mods/taskbar-start-button-corner-fix.wh.cpp
alchemyyy: Awesome! Done.

mods/taskbar-start-button-corner-fix.wh.cpp
alchemyyy: Thank you so much for your patience, my brain stopped working there. I've added a timer callback to get UIAutomation to get released from the right thread.

mods/taskbar-start-button-corner-fix.wh.cpp
```diff
@@ -0,0 +1,434 @@
...
+    }
+    return DefWindowProc(hWnd, uMsg, wParam, lParam);
+}
+
+// Initialize UI Automation (call from InputSite thread only)
+bool InitUIAutomation() {
+    if (g_pAutomation) return true;
+
+    HRESULT hr = CoInitializeEx(NULL, COINIT_APARTMENTTHREADED);
+    if (FAILED(hr) && hr != RPC_E_CHANGED_MODE && hr != S_FALSE) return false;
```
m417z: * `S_FALSE` is success, so the check isn't needed.
* Calling `CoUninitialize` after getting `RPC_E_CHANGED_MODE` is incorrect, and generally it shouldn't happen.

mods/taskbar-start-button-corner-fix.wh.cpp
```diff
@@ -0,0 +1,434 @@
...
+
+    DWORD threadId = GetWindowThreadProcessId(hWnd, NULL);
+    g_mouseHook = SetWindowsHookExW(WH_MOUSE, MouseHookProc, NULL, threadId);
+    if (!g_mouseHook) {
+        Wh_Log(L"Failed to install mouse hook (error: %lu)", GetLastError());
+        return;
+    }
+
+    // Create cleanup window on the InputSite thread via a timer callback
+    SetTimer(hWnd, (UINT_PTR)CreateCleanupWindow, 0, [](HWND hWnd, UINT, UINT_PTR idEvent, DWORD) {
```
m417z: From [the docs](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-settimer):

> This window must be owned by the calling thread.

For a way to run code in a window's thread, you can use `RunFromWindowThread`:

https://github.com/ramensoftware/windhawk-mods/blob/f5c9e38460918ba1fa4b643dedab4ecf42faf03c/mods/taskbar-start-button-position.wh.cpp#L327-L377

That will also make it unnecessary to have a cleanup window, the InputSite window can be used for that.

alchemyyy: Perfect! Done. Thank you so much again for the patience.

mods/taskbar-start-button-corner-fix.wh.cpp
alchemyyy: Whoops, fixed.

mods/taskbar-start-button-corner-fix.wh.cpp
```diff
@@ -0,0 +1,454 @@
...
+
+This mod hooks `CreateWindowInBand` to detect when the InputSite window is created,
+then installs a thread-specific mouse hook on its thread. This hook intercepts mouse
+clicks at the corner region and invokes the Start menu via UI Automation when the
+real button doesn't handle the click.
+
+## Requirements
+
+- Windows 11 64-bit (tested on 25H2)
+- Windhawk v1.4 or later
```
m417z, line 38: FYI, the `@license` metadata tag is only supported from Windhawk v1.5. Windhawk v1.4 is really old, released 2 years ago, with about 1% still using it from my data.

PR #2873, opened 2025-12-27
Add Taskbar Thumbnail Hover Fix mod

alchemyyy:
This mod fixes a bug in Windows 11 where taskbar thumbnail windows, when repositioned under the cursor, do not update their focus. This seems to be because they are simply waiting for mouse events to happen, with no additional check.

The use-case for this mod is to be able to rapidly close sets of windows form the thumbnails.

The ramifications from the simulated mouse inputs are unknown to me, but there likely are some. I would love to avoid doing this altogether if there is another solution.

EDIT
written to pair with this: https://github.com/m417z/my-windhawk-mods/pull/43

### DIFF
--- a/mods/taskbar-thumbnail-hover-fix.wh.cpp
+++ b/mods/taskbar-thumbnail-hover-fix.wh.cpp
@@ -28,8 +28,26 @@ Only the explorer.exe instance that owns the taskbar will activate this mod.
 #include <windhawk_utils.h>
 
 static HANDLE g_workerThread = nullptr;
-static HWINEVENTHOOK g_winEventHook = nullptr;
 static DWORD g_taskbarThreadId = 0;
+
+// Find the thread ID that created Shell_TrayWnd across all processes
+DWORD FindTaskbarThreadId() {
+    DWORD taskbarThreadId = 0;
+
+    EnumWindows(
+        [](HWND hWnd, LPARAM lParam) -> BOOL {
+            wchar_t className[32];
+            if (GetClassNameW(hWnd, className, ARRAYSIZE(className)) &&
+                wcscmp(className, L"Shell_TrayWnd") == 0) {
+                *reinterpret_cast<DWORD*>(lParam) = GetWindowThreadProcessId(hWnd, nullptr);
+                return FALSE;
+            }
+            return TRUE;
+        },
+        reinterpret_cast<LPARAM>(&taskbarThreadId));
+
+    return taskbarThreadId;
+}
 
 void TriggerHoverRefresh() {
     INPUT inputs[2] = {};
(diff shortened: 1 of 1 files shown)


mods/taskbar-thumbnail-hover-fix.wh.cpp
```diff
@@ -0,0 +1,218 @@
...
+    if (!root) {
+        root = hwnd;
+    }
+
+    WCHAR className[256];
+    if (GetClassName(root, className, ARRAYSIZE(className))) {
+        if (wcsstr(className, L"Xaml") != nullptr) {
+            return true;
+        }
+    }
```
m417z, line 128: I'd also check that the window belongs to the taskbar thread.

mods/taskbar-thumbnail-hover-fix.wh.cpp
```diff
@@ -0,0 +1,218 @@
...
+        g_mutex = nullptr;
+        return 1;
+    }
+
+    Wh_Log(L"Hook installed successfully");
+
+    // Message pump for WinEventHook
+    MSG msg;
+    while (g_running) {
+        DWORD result = MsgWaitForMultipleObjects(0, nullptr, FALSE, 100, QS_ALLINPUT);
```
m417z: Instead of polling each 100ms, you can post a message to the thread to exit. Example:

https://github.com/ramensoftware/windhawk-mods/blob/7a97b948c1747274d849d2c53f6f88be9920f95f/mods/taskbar-background-helper.wh.cpp#L667-L682

https://github.com/ramensoftware/windhawk-mods/blob/7a97b948c1747274d849d2c53f6f88be9920f95f/mods/taskbar-background-helper.wh.cpp#L874-L879

mods/taskbar-thumbnail-hover-fix.wh.cpp
alchemyyy: done!

mods/taskbar-thumbnail-hover-fix.wh.cpp
alchemyyy: beautiful. thank you! done.

mods/taskbar-thumbnail-hover-fix.wh.cpp
```diff
@@ -27,12 +27,9 @@ Only the explorer.exe instance that owns the taskbar will activate this mod.
 
 #include <windhawk_utils.h>
 
-#define MUTEX_NAME L"Local\\TaskbarThumbnailHoverFix"
-
 static HANDLE g_workerThread = nullptr;
-static HANDLE g_mutex = nullptr;
 static HWINEVENTHOOK g_winEventHook = nullptr;
```
m417z: This can be a local variable now.

mods/taskbar-thumbnail-hover-fix.wh.cpp
```diff
@@ -92,55 +95,40 @@ void CALLBACK WinEventProc(
...
-
-        Sleep(500);
-    }
+    return hTaskbarWnd;
+}
 
-    if (!g_running) {
+DWORD WINAPI WorkerThreadProc(LPVOID lpParam) {
+    // Check if this process owns the taskbar
+    HWND taskbarWnd = FindCurrentProcessTaskbarWnd();
```
m417z: Without polling, this will likely fail on process startup. One alternative to polling is hooking window creation.

alchemyyy: Oh you're absolutely right. It does seem to work fine but I shut my head off here again. Updated with what should *hopefully* be a pretty good solution. I thought it would be nice to avoid a mutex for this mod by leveraging the relative infrequency and conditions of window destruction events.

mods/taskbar-thumbnail-hover-fix.wh.cpp
alchemyyy: done

mods/taskbar-thumbnail-hover-fix.wh.cpp
```diff
@@ -61,6 +79,41 @@ bool IsCursorOverThumbnail() {
         root = hwnd;
     }
 
+    // First time we encounter a window destruction, verify we're in the taskbar owner process
+    if (g_taskbarThreadId == 0) {
+        DWORD taskbarThreadId = FindTaskbarThreadId();
```
m417z, line 84: Minor optional notes for the future:

* `FindTaskbarThreadId` could easily return the process ID too, which would make it unnecessary to use `OpenThread` below.
* You could also use `FindTaskbarThreadId` on mod init to exit early in case the taskbar is already owned by an existing process.

PR #2879, opened 2025-12-28
Add cpl-classic-view-lite

Anixx:


mods/cpl-classic-view-lite.wh.cpp
```diff
@@ -0,0 +1,81 @@
...
+- categ: false
+  $name: Display the Control Panel categories as normal folders as well
+  $description: Determines whether the Control Panel categories should also be displayed as usual folders.
+*/
+// ==/WindhawkModSettings==
+
+#include <windhawk_utils.h>
+
+HRESULT (*CreateCommonFrame_orig)(void *, UINT, GUID, void **);
+HRESULT CreateCommonFrame_hook(void * psf, UINT uLayoutType, GUID guid, void **ppv)
```
m417z: The calling convention is missing here, which is likely to cause a crash for 32-bit targets. Please add calling conventions or, if the mod doesn't support 32-bit, limit it to 64-bit, which will also make compilation faster. See:

Also, for `guid` of type `struct _GUID const &` (according to the symbol), I'd expect it to be `GUID*` rather than `GUID`.

Anixx: Is it OK now?

PR #2883, opened 2025-12-29
Mod: No MS Edge WebView2 for Copilot

Xetrill:


mods/xs-nomsedgewebview4copilot.wh.cpp
```diff
@@ -0,0 +1,96 @@
...
+    LPVOID lpEnvironment,
+    LPCWSTR lpCurrentDirectory,
+    LPSTARTUPINFOW lpStartupInfo,
+    LPPROCESS_INFORMATION lpProcessInformation
+)
+{
+    const wchar_t *cmd = lpCommandLine ? lpCommandLine : L"";
+    Wh_Log(L"[CreateProcessW] Cmdline: '%s'", cmd);
+
+    if (wcsstr(cmd, L"msedgewebview2.exe") != nullptr) {
```
m417z: Is a case-sensitive match good enough here? Also, is there anything specific to Copilot in the command line in these cases? It can help avoid breaking the apps in the future if they start using Webview2 for core functionality.

Xetrill: Made the string matching case-insensitive.

Sadly, no there's nothing to pin WebView usage to Copilot specifically, at least not from the command line.
Here's an example for reference:
`"C:\Program Files (x86)\Microsoft\EdgeWebView\Application\143.0.3650.96\msedgewebview2.exe" --embedded-browser-webview=1 --webview-exe-name=OneDrive.exe --webview-exe-version=25.222.1112.0002 --user-data-dir="C:\Users\Xetrill\AppData\Local\Microsoft\OneDrive\EBWebView" --noerrdialogs --embedded-browser-webview-dpi-awareness=1 --disable-features=msEnhancedTrackingPreventionEnabled --mojo-named-platform-channel-pipe=17380.31640.7996082180259808924 /pfhostedapp:4ea411c1477e6df8dae43fb3a80aa9e2d611b5cd`

PR #2889, opened 2025-12-30
Add Mod: taskbar-idle-clean-desktop.wh.cpp

BREIQ:
Hides the taskbar and minimizes all windows after a period of user inactivity.

Features:

    Smart detection: Won't trigger during full-screen games, videos, or presentations.

    Auto-restore: Taskbar reappears instantly upon mouse or keyboard activity.

    Boot protection: Ensures the taskbar is visible on startup or when the mod initializes.

### DIFF
--- a/mods/taskbar-idle-clean-desktop.wh.cpp
+++ b/mods/taskbar-idle-clean-desktop.wh.cpp
@@ -2,7 +2,7 @@
 // @id              taskbar-idle-clean-desktop
 // @name            Clean Desktop on Idle
 // @description     Hides the taskbar and minimizes all windows after a period of user inactivity. Includes smart detection.
-// @version         1.8
+// @version         1.9
 // @author          Breiq
 // @include         windhawk.exe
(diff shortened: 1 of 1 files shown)


mods/taskbar-idle-clean-desktop.wh.cpp
```diff
@@ -0,0 +1,127 @@
+// ==WindhawkMod==
+// @id              taskbar-idle-clean-desktop
+// @name            Clean Desktop on Idle
+// @description     Hides the taskbar and minimizes all windows after a period of user inactivity. Includes smart detection.
+// @version         1.7
+// @author          Breiq
+// @include         explorer.exe
```
m417z: See: https://github.com/ramensoftware/windhawk/wiki/Mods-as-tools:-Running-mods-in-a-dedicated-process

m417z: Your update is not per the wiki instructions, it only implements step 3, not steps 1 and 2.

PR #2895, opened 2025-12-31
IDA Aliased Fonts 1.0

kawapure:


mods/ida-aliased-fonts.wh.cpp
```diff
@@ -0,0 +1,188 @@
...
+            WH_DISASM_RESULT disasm3 = { 0 };
+            Wh_Disasm((void *)((size_t)pCurAddr + disasm2.length), &disasm3);
+            Wh_Log(L"We should be done: %S; %S", disasm2.text, disasm3.text);
+        }
+    }
+}
+
+// The mod is being initialized, load settings, hook functions, and do other
+// initialization stuff if required.
+#include <cstdio>
```
m417z, line 130: For future updates: looks like this include is unused.

kawapure: Haha, oops. The IDE likes to throw random includes in. I never use `<cstdio>`, only `<stdio.h>`, so that's how I know.

Issue #2010, opened 2025-06-09
[Translucent Window] Dark/black background artifacts even for excluded apps (e.g. Office)


qkdan:
@Undisputed00x 

When the mod is enabled, Many MS Office controls and views are inverted from how they should be in dark mode (e.g. Outlook message pane is white instead of black, Excel workbook editing area is black instead of white)

Disabling Translucent Windows puts Office back to expected appearance 

Nothing was logged by the mod with debug logging enabled, so no log to share unfortunately

Windows Version: Windows 11 Enterprise 24H2 26100.4061, Experience Pack 1000.26100.84.0

Office Version: 365 2505 build 18827.20140 CTR

Mod version: 1.5.0

Windhawk version: 1.6.1

Undisputed00x:
The Windows Theme Optimization setting is intended to be used with a selected effect to avoid visual bugs and modifies parts of the Windows theme. You can target programs to apply specific settings to them by applying a program rule to the mod settings, or you can even use Windhawk's exception list to deny injection to those programs entirely.

myrajurek:
This is still an issue even using the process rules to turn off all settings or adding the process path to the exclusion list.
Disabling the entire mod returns excel to its original state. Also it should be noted that the transparency does work on the of applications display such as the standard menu bar and file options , but when a work-book cell in in its default state it is black, changing the color of the cell still works as intended (Ex: you can still change a cell to white) its only when the cell is "empty" does it display as black. Text Still displays correctly when input .  Additionally all dialog boxes pop up  with a black background and the text renders incorrectly in their menus. The issue really arises (for me at least) when importing workbooks. Because some of them already have theming like white cells with black text, you cannot see them. you can only fix this by manually modifying the cell. Also when imputing text into a cell it normally appears in the cell as you type it, this preview while you type before you actually modify the cell is in black, and cannot be seen over the default black cells. extraneous information but: I have has this issue with google sheets in the past using dark mode extensions. In most spreadsheet apps, the spreadsheet is rendered differently than the rest of the app.
<img width="582" height="525" alt="Image" src="https://github.com/user-attachments/assets/7b8d656a-b37a-4562-9680-e117fe993564" />
The dropdown boxes here have text, though its very hard to see.  
Microsoft Excel for Microsoft 365 MSO (Version 2506 Build 16.0.18925.20076) 64-bit
Windows 11 Home 24H2 OS build	26100.4652
Windows Feature Experience Pack 1000.26100.128.0
[...]

PR #2901, opened 2025-12-31
Translucent Windows v1.7.0

Undisputed00x:
* Expand translucent effects to flyouts (popup menus, tooltips, dropdown menu) (thanks to [TranslucentFlyouts](https://github.com/ALTaleX531/TranslucentFlyouts) project )
* Update readme, provide more useful information.
* Add a setting option to modify system colors and custom rendering option settings in process rules.
* Some custom theme additions/tweaks.

### DIFF
--- a/mods/translucent-windows.wh.cpp
+++ b/mods/translucent-windows.wh.cpp
@@ -50,6 +50,9 @@ Restart explorer.exe to change the background color.⚠️
 * ⚠️The background effects do not affect most modern windows (UWP/WinUI), 
 apps with different front-end rendering (e.g Qt, Electron, Chromium etc.. programs) and native windows with hardcoded colors.⚠️
 
+⚠️In case parts of Windows UI colors remain modified after disabling the mod, 
+delete the HKEY_CURRENT_USER\Control Panel\Colors registry key and reboot.⚠️
+
 * ❕The blur effect may show a bleeding effect at the edges of a window when maximized or snapped to the edge of the screen. 
 This is caused by default by the AccentBlur API.❕
 
@@ -5721,4 +5724,5 @@ BOOL Wh_ModSettingsChanged(BOOL* bReload)
     *bReload = TRUE;
     return TRUE;
 }
+
 


mods/translucent-windows.wh.cpp
```diff
@@ -52,6 +71,13 @@ maximized or snapped to the edge of the screen, this is caused by default.
       $name: Windows theme custom rendering
       $description: >-
        Modifies parts of the Windows theme using the Direct2D graphics API.
+        ✨It is recommended to enable this with background translucent effects.
+    - SysColors: FALSE
```
m417z: This doesn't seem to be used in the code.

mods/translucent-windows.wh.cpp
```diff
@@ -1120,14 +1208,12 @@ constexpr INT SysColorElements[] = {
...
-    if (hTh != nullptr)
-        hTh = nullptr;
+    if (g_hTheme != nullptr)
+        g_hTheme = nullptr;
 
-    hTh = OpenThemeData(NULL, themeclass); 
+    g_hTheme = OpenThemeData(NULL, themeclass); 
 }
 
 VOID RevertSysColors()
```
m417z, line 1223: I wonder if this can be improved, since I saw several reports of users who disabled the mod but still see the black artifacts. I believe some reported that reboot didn't help too.

Here's a report just from today:
https://www.reddit.com/r/Windhawk/comments/1q1uh7b/black_background_and_text/

Do you have an idea as of why it may happen?

mods/translucent-windows.wh.cpp
```diff
@@ -1137,12 +1223,60 @@ VOID RevertSysColors()
...
+    return GetSysColor_orig(nIndex);
+}
+
+COLORREF WINAPI HookedGetSysColor(int nIndex) {    
+    return GetCustomSysColor(nIndex);
+}
+
+HBRUSH WINAPI HookedGetSysColorBrush(int nIndex) {
+    COLORREF color = GetCustomSysColor(nIndex);
+    return CreateSolidBrush(color);
```
m417z: The original `GetSysColorBrush` returns a system-managed brush that callers don't delete. This hook creates a new brush every call which will never be freed.

mods/translucent-windows.wh.cpp
Undisputed00x: In the report you shared, the user probably skipped the alpha text blending option, which also changes the color of the text. For some reason, many users skip this option, despite the fact that even the title mentions text. I have integrated the text option into the custom theme rendering because of this.

But you're right, although so far I've seen 2 reports on Windhawk's discord about changed system colors persisting even after a reboot, which is really strange and I don't know why this could happen, as the mod's modifications are in memory.

mods/translucent-windows.wh.cpp
Undisputed00x: Done.

mods/translucent-windows.wh.cpp
m417z: I don't know either. I tried asking AI. Below is its attempt at explaining this. I didn't have time to try it out, but the theory with `SetSysColors` + settings change causing permanent changes sound interesting and worth checking.

It might be completely wrong, too, as AI answers sometimes are.

> **The code itself does NOT directly cause persistent color changes.** The `SetSysColors` API used at [translucent-windows.wh.cpp:1236](https://github.com/Undisputed00x/windhawk-mods/blob/71a6a66cc0667cbf584bfd5cfea18bf47b642dc7/mods/translucent-windows.wh.cpp#L1236) and [translucent-windows.wh.cpp:1308](https://github.com/Undisputed00x/windhawk-mods/blob/71a6a66cc0667cbf584bfd5cfea18bf47b642dc7/mods/translucent-windows.wh.cpp#L1308) does not write to the registry - it only affects the current Windows session, as confirmed by [Microsoft's official documentation](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setsyscolors) and [pinvoke.net](https://www.pinvoke.net/default.aspx/user32.SetSysColors).
> 
> However, there's a likely explanation for user reports:
> 
> ### The Probable Cause: Windows Settings Synchronization
> 
> When the mod is active and has called `SetSysColors`, the current system colors are modified. If a user then:
> 
> 1.  **Opens Windows Settings > Personalization > Colors**
> 2.  **Changes ANY color/theme setting** (even unrelated ones like accent color)
> 3.  **Or simply opens the old Control Panel appearance settings**
> 
> Windows may **read the current color values** (which are the mod's modified colors) and **write them to the registry** at `HKEY_CURRENT_USER\Control Panel\Colors`, making them permanent.
> 
> This is a known Windows behavior pattern documented in various forums - the [Windows 10 Forums](https://www.tenforums.com/customization/146702-registry-color-keys.html) and [Eleven Forum](https://www.elevenforum.com/t/default-registry-values-for-colors-windows-11.15580/) discuss how the registry and live colors interact.
> 
> ### The `RevertSysColors()` Limitation
> 
> The restoration function at [translucent-windows.wh.cpp:1224-1240](https://github.com/Undisputed00x/windhawk-mods/blob/71a6a66cc0667cbf584bfd5cfea18bf47b642dc7/mods/translucent-windows.wh.cpp#L1224-L1240) uses `GetThemeSysColor()` to restore colors from the current theme, **not** from the user's original `Control Panel\Colors` registry values. If the registry was already overwritten (via the scenario above), this restoration just reads from the theme defaults, which may differ from what the user had before.
> 
> ### Recommendation
> 
> The mod should save the original registry values from `HKEY_CURRENT_USER\Control Panel\Colors` on initialization and restore those exact values on unload, rather than relying on theme colors.

Undisputed00x: This didn't help, also the AI ​​confused the RevertSysColors() and the recommended solution.
Let's leave it as is, since there are a small number of users who have reported it, which probably have modified something on their end that causes it.
Deleting the HKEY_CURRENT_USER\Control Panel\Colors resolves this, so i'll add this workaround suggestion in the mod's readme in case someone faces this issue.

MAUSOL95:
I hope you'll take my feedback into account and fix the Office issue. I managed to get it working, but I did what I could to get it officially released. I hope it can be fixed in a future update.

hotcofeee:
“Hey, there is a new bug after the update. When you try to watch any video in full screen, the window borders are still visible, and it’s not actually full screen.”

Undisputed00x:
> “Hey, there is a new bug after the update. When you try to watch any video in full screen, the window borders are still visible, and it’s not actually full screen.”

Does this also happen in Gecko (Mozilla) browsers? Currently, you can either set up rules for this process or exclude it completely from the Windhawk's mod exclusion list.

MAUSOL95:
> > Espero que tenga en cuenta mis comentarios y solucione el problema de Office. Logré hacerlo funcionar, pero hice lo que pude para que se lanzara oficialmente. Espero que se pueda solucionar en una futura actualización..
> 
> > Cambio: Comentó SetSysColors() para evitar cambios globales
> 
> Reemplazó la API SetSysColors con captadores SysColors y FillRect, pero eso no es suficiente. Ver: [#2010 (Comentario)](https://github.com/ramensoftware/windhawk-mods/issues/2010#issuecomment-3544379735)

Well, I was able to fix it. Check the code; I only edited it from version 1.6.3, but version 1.7.0 wasn't available. However, I did manage to prevent Office from being affected.
Captura de pantalla 2026-01-04 172259
image
[Translucent.Windows.Fix.txt](https://github.com/user-attachments/files/24424842/Translucent.Windows.Fix.txt)

MAUSOL95:
I want to inform you that I fixed the dark pages issue in Office version 1.7.0
[Translucent.Windows.1.7.0.fix.txt](https://github.com/user-attachments/files/24425529/Translucent.Windows.1.7.0.fix.txt)

Luna-Nocturna:
> > The new update seems to force some blur or shadow effect around right click menu's. I don't have those translurent by default as sadly I can't enable "Windows theme custom rendering"due to the app exclusion limitations described in the new FAQ. As such my right click menu's now look like the following:
> > image
> > Is there a way to downgrade to an earlier version? As I already couldn't use more of the features as the exclusions weren't working I was fine with it and none of the new features apply to me anyway.
> 
> It was conflicting with the mod -dark context menu As these menu are already dark in windows by default.

But there's no options to enable/disable it for menu's. So how do I disable it to avoid the conflict?

Aside from that it also does this with menu's from apps now (Example Notepad++)

image

It makes everything white/gray as if it's unselectable until you hover over it with the mouse.

Again is there a way to downgrade? I rather stick on version 1.6.x without all these new problems.

Undisputed00x:
> The new update seems to force some blur or shadow effect around right click menu's. I don't have those translurent by default as sadly I can't enable "Windows theme custom rendering"due to the app exclusion limitations described in the new FAQ. 

@Luna-Nocturna  There is a new option called "New System Colors" that overrides process exceptions. In this case, you can enable the Custom Theme Rendering setting.

PR #2916, opened 2026-01-02
Taskbar Clock Customization v1.7.1

m417z:
* Added an option to specify text for `%media_info%` when no media is playing.
* Added the `%n%` pattern as a shorter alternative for `%newline%`.
* Fixed `%battery%` to show 100% when fully charged. 99% was previously displayed in this case.
* Fixed network and GPU metrics not accounting for new adapters or processes.
* Fixed the mod preventing explorer.exe from exiting if media info patterns are used.


utkuvrs:
Looks good to me

PR #2923, opened 2026-01-03
Add "Confirm Closing Multiple Tabs in File Explorer" mod

AromaKitsune:
This mod adds a confirmation dialog that spawns when you attempt to close a File Explorer window
with multiple tabs open, preventing accidental closure of all tabs.

Preview

### DIFF
--- a/mods/confirm-closing-multiple-explorer-tabs.wh.cpp
+++ b/mods/confirm-closing-multiple-explorer-tabs.wh.cpp
@@ -19,6 +19,7 @@ with multiple tabs open, preventing accidental closure of all tabs.
 */
 // ==/WindhawkModReadme==
 
+#include <windhawk_utils.h>
 #include <windows.h>
 #include <commctrl.h>
 #include <cstdio>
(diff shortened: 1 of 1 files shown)


mods/confirm-closing-multiple-explorer-tabs.wh.cpp
```diff
@@ -0,0 +1,199 @@
...
+    return TRUE;
+}
+
+// Mod initialization
+BOOL Wh_ModInit() {
+    // Hook creation of new windows
+    Wh_SetFunctionHook((void*)CreateWindowExW, (void*)CreateWindowExW_Hook, (void**)&CreateWindowExW_Original);
+
+    // Attach to any currently open Explorer windows
+    EnumWindows(EnumWindowsCallback, 0);
```
m417z, line 186: Does this work? I suspect that it doesn't because normally, `SetWindowSubclass` can't be called for windows in a different thread.

You can use `WindhawkUtils::SetWindowSubclassFromAnyThread`, see:

Then, you can call `WindhawkUtils::RemoveWindowSubclassFromAnyThread` on unload.

mods/confirm-closing-multiple-explorer-tabs.wh.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+// Subclass procedure to intercept WM_CLOSE
+LRESULT CALLBACK ExplorerSubclassProc(HWND hWnd, UINT uMsg, WPARAM wParam, LPARAM lParam, UINT_PTR uIdSubclass, DWORD_PTR dwRefData) {
+    if (uMsg == WM_CLOSE) {
+        int tabCount = GetTabCount(hWnd);
+
+        // Only trigger if 2 or more tabs are open
+        if (tabCount >= 2) {
+            // Play Asterisk sound
+            MessageBeep(MB_ICONASTERISK);
```
m417z: The message already plays that sound.

mods/confirm-closing-multiple-explorer-tabs.wh.cpp
AromaKitsune: Done. I got the confirmation dialog working after enabling a mod while the File Explorer windows are already open.

mods/confirm-closing-multiple-explorer-tabs.wh.cpp
AromaKitsune: Removed.

m417z:
Using AI for help is fine, nothing wrong about that. It can be more challenging to review work that was made with AI with little to no understanding from the author, especially if review comments are forwarded back to AI that again does something wrong, which creates a not-so-productive feedback loop.

In this case, it was fairly smooth.

PR #2926, opened 2026-01-03
Create Word Local AutoSave

communism420:
Enables AutoSave functionality for local documents in Microsoft Word by sending Ctrl+S

### DIFF
--- a/mods/word-local-autosave.wh.cpp
+++ b/mods/word-local-autosave.wh.cpp
@@ -28,6 +28,7 @@ short delay.
 - Configurable delay before saving
 - Optional minimum interval between saves to prevent excessive disk writes
 - Works with any locally saved Word document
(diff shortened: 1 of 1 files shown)


mods/word-local-autosave.wh.cpp
```diff
@@ -0,0 +1,294 @@
...
+
+    // Check if there was recent input
+    if (g_lastInputTime == 0) {
+        return;
+    }
+
+    Wh_Log(L"Performing auto-save...");
+
+    // Send Ctrl+S
+    SendCtrlS();
```
m417z, line 159: The mod simulates Ctrl+S, but it doesn't verify that Word is still the foreground window.

communism420: Fixed it

PR #2956, opened 2026-01-06
Add Windows 10 taskbar clock seconds mod

halfofknowledge-cpu:


Anixx:
Windhawk mods shold not make permanent changes to the registry settings. Your mod changes a registry value.

halfofknowledge-cpu:
I don't understand anixx

Anixx:
Your mod makes changes to the registry. Windhawk mods should not do so.

Anixx:
Above is the link to the mod that you can use for your mod. You only change AltTabSettings to ShowSecondsInSystemClock and it adds seconds to system clock.

Issue #2957, opened 2026-01-06
[Taskbar Thumbnail Size] Use absolute sizing settings make large thumbnails when there is a vertically long window


adusamr:
I'm assuming that the intended behavior of use absolute sizing settings with preserve aspect ratio is making the thumbnail fit inside a target rectangle satisfying (minWidth, minHeight) <= target <= (maxWidth, maxHeight), and if resizing is unecessary, preserve the original thumbnail size.

As the title says, when there is a vertically long window the thumbnails become large. I think this is a bug.

I made a fork that works as I expected:
```C++
// ==WindhawkMod==
// @id              taskbar-thumbnail-size-fork
// @name            Taskbar Thumbnail Size fork
// @description     Customize the size of the new taskbar thumbnails in Windows 11
// @version         1.1
// @author          m417z
// @github          https://github.com/m417z
// @twitter         https://twitter.com/m417z
// @homepage        https://m417z.com/
// @include         explorer.exe
// @architecture    x86-64
// @compilerOptions -lole32 -loleaut32 -lruntimeobject
// ==/WindhawkMod==

// Source code is published under The GNU General Public License v3.0.
//
// For bug reports and feature requests, please open an issue here:
// https://github.com/ramensoftware/windhawk-mods/issues
//
// For pull requests, development takes place here:
// https://github.com/m417z/my-windhawk-mods

// ==WindhawkModReadme==
/*
# Taskbar Thumbnail Size

Customize the size of the new taskbar thumbnails in Windows 11.

By default, this mod uses simple percentage-based scaling. You can optionally
enable absolute sizing mode to set specific pixel dimensions for thumbnails.

For older Windows versions, the size of taskbar thumbnails can be changed via the registry:

* [Change Size of Taskbar Thumbnail Previews in Windows
  11](https://www.elevenforum.com/t/change-size-of-taskbar-thumbnail-previews-in-windows-11.6340/)
  (before Windows 11 version 24H2)
* [How to Change the Size of Taskbar Thumbnails in Windows
  10](https://www.tenforums.com/tutorials/26105-change-size-taskbar-thumbnails-windows-10-a.html)

![Demonstration](https://i.imgur.com/nGxrBYG.png)
*/
// ==/WindhawkModReadme==

// ==WindhawkModSettings==
/*
- size: 150
  $name: Thumbnail size (percentage)
  $description: Percentage of the original size. Used when absolute sizing is disabled.
- useAbsoluteSize: false
[...]

PR #2966, opened 2026-01-07
Fix taskbar-thumbnail-size.wh.cpp preserve-aspect-ratio bounds prioritization

alchemyyy:
When "preserve aspect ratio" is selected, the respected bounds are now the width and height maximums. Minimum bounds are attempted to be respected, but not guaranteed.

Also performed some general cleanup.

Fixes https://github.com/ramensoftware/windhawk-mods/issues/2957


m417z:
Please submit the pull request here:
https://github.com/m417z/my-windhawk-mods/pulls

PR #2976, opened 2026-01-08
Added Explorer Up → New Window

TobbeLino:
Explorer Up → New Window

Opens the parent folder in a **new** Explorer window when you:
- **Middle-click** on the Up button
- **Ctrl + Left-click** on the Up button
- Press **Ctrl + Alt + Up** keyboard shortcut

Useful when you want to keep your current folder open while exploring the parent.

### DIFF
--- a/mods/explorer-up-new-window.wh.cpp
+++ b/mods/explorer-up-new-window.wh.cpp
@@ -2,10 +2,10 @@
 // @id              explorer-up-new-window
 // @name            Explorer Up → New Window (Ctrl-click or Middle-click)
 // @description     Ctrl or middle-click on explorer up-button opens parent in a new window.
-// @version         1.2
+// @version         1.3
 // @author          Tobias Lind
 // @github          https://github.com/TobbeLino
-// @include         explorer.exe
+// @include         windhawk.exe
 // @compilerOptions -lole32 -loleaut32 -luiautomationcore -lshell32 -luser32 -luuid
 // @license         MIT
 // ==/WindhawkMod==
@@ -25,6 +25,7 @@ Useful when you want to keep your current folder open while exploring the parent
 
 #define WIN32_LEAN_AND_MEAN
 #include <windows.h>
+#include <shellapi.h>
 #include <UIAutomation.h>
 #include <OleAuto.h>
 #include <string>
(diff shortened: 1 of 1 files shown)


mods/explorer-up-new-window.wh.cpp
```diff
@@ -0,0 +1,534 @@
...
+
+// ---------- Windhawk entry points ----------
+BOOL Wh_ModInit() {
+    Wh_Log(L"Init (Up → New Window, UIA invoke + Ctrl-latch)");
+
+    g_mouseHook = SetWindowsHookExW(WH_MOUSE_LL, MouseLLProc, nullptr, 0);
+    if (!g_mouseHook) Wh_Log(L"[err] SetWindowsHookEx WH_MOUSE_LL failed: %lu", GetLastError());
+
+    g_kbdHook = SetWindowsHookExW(WH_KEYBOARD_LL, KbdLLProc, nullptr, 0);
+    if (!g_kbdHook) Wh_Log(L"[err] SetWindowsHookEx WH_KEYBOARD_LL failed: %lu", GetLastError());
```
m417z: From [the documentation](https://learn.microsoft.com/en-us/windows/win32/winmsg/lowlevelkeyboardproc):

mods/explorer-up-new-window.wh.cpp
```diff
@@ -0,0 +1,534 @@
...
+}
+
+void Wh_ModUninit() {
+    if (g_mouseHook) { UnhookWindowsHookEx(g_mouseHook); g_mouseHook = nullptr; }
+    if (g_kbdHook)   { UnhookWindowsHookEx(g_kbdHook);   g_kbdHook   = nullptr; }
+    g_suppress = SuppressBtn::None;
+    ResetInProgress();
+
+    // Clean up thread-local COM objects for current thread
+    CleanupThreadUia();
```
m417z: That's not the correct thread for cleanup.

mods/explorer-up-new-window.wh.cpp
```diff
@@ -0,0 +1,611 @@
+// ==WindhawkMod==
+// @id              explorer-up-new-window
+// @name            Explorer Up → New Window (Ctrl-click or Middle-click)
+// @description     Ctrl or middle-click on explorer up-button opens parent in a new window.
+// @version         1.2
+// @author          Tobias Lind
+// @github          https://github.com/TobbeLino
+// @include         explorer.exe
```
m417z: See:

TobbeLino: Ok, I (read: Claude Opus 4.5) converted it to the tool mod pattern

SpacEagle17:
Hi there, ty for the mod :)
Would love to have the functionality for it to open in a new tab instead if that's possible, ty ^^

PR #3004, opened 2026-01-12
Taskbar desktop switcher

Frqmelikescheese:
Adds modern, customizable desktop switching buttons to the taskbar with configurable colors, sizes, positioning, and smooth hover effects. Allows seamless virtual desktop navigation with visual feedback.

### DIFF
--- a/mods/taskbar-desktop-switcher.wh.cpp
+++ b/mods/taskbar-desktop-switcher.wh.cpp
@@ -2,11 +2,30 @@
 // @id              taskbar-desktop-switcher
 // @name            Taskbar Desktop Switcher
 // @description     Modern, customizable desktop buttons with hover effects
-// @version         1.0
+// @version         1.1
 // @author          Frqme
-// @github         https://github.com/Frqmelikescheese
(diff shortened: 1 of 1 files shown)


m417z:
I tried it with 2 desktops. When I created additional two desktops (to have 4), it began working. That's not mentioned in the readme. In fact, there's no readme at all. Please add a readme (mod description) that says what the mod is about and how to use it.

mods/taskbar-desktop-switcher.wh.cpp
```diff
@@ -0,0 +1,285 @@
+// ==WindhawkMod==
+// @id              taskbar-desktop-switcher
+// @name            Taskbar Desktop Switcher
+// @description     Modern, customizable desktop buttons with hover effects
+// @version         1.0
+// @author          Frqme
+// @github         https://github.com/Frqmelikescheese
+// @include         explorer.exe
```
m417z, line 8: See https://github.com/ramensoftware/windhawk/wiki/Mods-as-tools:-Running-mods-in-a-dedicated-process

PR #3006, opened 2026-01-12
Add taskbar visibility toggle mod

roosmsg:
## Summary
- Add `taskbar-visibility-toggle` mod to toggle taskbar visibility with Ctrl+Esc.
- Sticky state persists (shown/hidden) until toggled again.
- Optional setting to keep window sizes fixed (requires Windows auto-hide enabled).

## How it works
- Hooks taskbar hide/unhide timers and tray unhide flow.
- Uses a low-level keyboard hook to intercept Ctrl+Esc and block Start menu.
- Updates auto-hide state when needed to keep window sizing consistent.

## Testing
- [x] Manual: toggled Ctrl+Esc; taskbar hides/shows as expected.
- [x] Manual: with Windows auto-hide enabled + “Keep window sizes fixed”.


m417z:
I tried this mod on my Windows 11, and it didn't work reliably for me. Most of the time, Ctrl+Esc just opened/closed the start menu. It seems that the `state != ToggleState::ForcedHidden` condition isn't true most of the time for me. Does it work for you? Or am I missing something?

Also, I noticed that it doesn't work when Windhawk is on the foreground since Windhawk is running as administrator. I found it confusing, you might want to add a note about it.

PR #3022, opened 2026-01-14
taskbar-disappearing-tray-icons-fix.wh.cpp Patch GDI handle leak and disable passive mode by default

alchemyyy:
There's a GDI handle leak that *appears* to be coming from LoadImageW in a suspiciously identical amount of icons that would be in my "quick settings tray". This adds hooks to LoadImageW, CreateIconIndirect, CopyIcon, and CreateIconFromResourceEx as possible culprits for handle leaks.

These hooks only run during passive mode. I'm making a tradeoff for reliability since a one-time rebroadcasting is unlikely to cause 10000 GDI handles to be created. Passive mode is also now off by default.

### DIFF
--- a/mods/taskbar-disappearing-tray-icons-fix.wh.cpp
+++ b/mods/taskbar-disappearing-tray-icons-fix.wh.cpp
@@ -357,6 +357,7 @@ void Wh_ModUninit() {
     Wh_Log(L"Uninitialized");
 }
 
-void Wh_ModSettingsChanged() {
-    LoadSettings();
+BOOL Wh_ModSettingsChanged(BOOL* bReload) {
+    *bReload = TRUE;
+    return TRUE;
 }


mods/taskbar-disappearing-tray-icons-fix.wh.cpp
```diff
@@ -58,34 +64,138 @@ Only the explorer.exe instance that owns the taskbar will activate this mod.
...
 
 void LoadSettings() {
     int initialDelay = Wh_GetIntSetting(L"initialDelaySeconds");
     int interval = Wh_GetIntSetting(L"intervalSeconds");
-    
+
     g_settings.initialDelaySeconds.store(initialDelay < 1 ? 5 : initialDelay);
     g_settings.intervalSeconds.store(interval < 1 ? 5 : interval);
-    g_settings.disablePassiveBroadcast.store(Wh_GetIntSetting(L"disablePassiveBroadcast") != 0);
+    g_settings.passiveBroadcast.store(Wh_GetIntSetting(L"passiveBroadcast") != 0);
```
m417z, line 198: Changing the settings won't cause the hooks to be set/removed. You can either set/remove hooks when the settings change, or, the easier option, use [the second variant of `Wh_ModSettingsChanged`](https://github.com/ramensoftware/windhawk/wiki/Creating-a-new-mod#wh_modsettingschanged) and reload the mod.

alchemyyy: Totally skipped over that, fixed!

PR #3031, opened 2026-01-14
Click on empty taskbar space v2.5

m1lhaus:
Diff might look a little bit messy. I was moving around some parts of the code enclosing it to regions to make navigation easier. So yeah, sorry for that 🥹

Changelog:
- added "Example presets" chapter to mod's description
- extended "Troubleshooting and Hints" chapters in mod's description
- Ctrl+Alt+Tab action reworked - now supports consecutive triggers without window constantly re-opening
- added new action "Switch virtual desktop"
- bugfixed file logger for Onedrive users
- bugfixed touch/tap detection on Windows 10 taskbar

### DIFF
--- a/mods/taskbar-empty-space-clicks.wh.cpp
+++ b/mods/taskbar-empty-space-clicks.wh.cpp
@@ -624,6 +624,12 @@ static const DWORD g_injectedClickID = 0xEADBEAF1u; // magic number to identify 
 static const UINT g_explorerPatcherContextMenuMsg = RegisterWindowMessageW(L"Windows11ContextMenu_{D17F1E1A-5919-4427-8F89-A1A8503CA3EB}");
 static const UINT g_uninitCOMAPIMsg = RegisterWindowMessageW(L"Windhawk_UnInit_COMAPI_empty-space-clicks");
 
+// Private API for window band (z-order band).
+// https://blog.adeltax.com/window-z-order-in-windows-10/
+using GetWindowBand_t = BOOL(WINAPI *)(HWND hWnd, PDWORD pdwBand);
+GetWindowBand_t pGetWindowBand = nullptr;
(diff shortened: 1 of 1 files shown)


mods/taskbar-empty-space-clicks.wh.cpp
```diff
@@ -1655,34 +2103,180 @@ HWND FindDesktopWindow()
...
+    return param.hWnd;
+}
+
+// Finds Task Switching window corresponding to current OS version
+HWND FindTaskSwitchingWindow()
+{
+    LOG_TRACE();
+
+    const std::wstring className = (g_osVersion == WIN_11) ? L"XamlExplorerHostIslandWindow" : L"MultitaskingViewFrame"; // dialog class on Win10
+    const std::wstring windowName = L"Task Switching";
```
m417z: I'm afraid that might be locale-specific. I have [a repo for such checks](https://github.com/m417z/windows-language-files), but I don't have a local clone right now, and GitHub doesn't search in it since it's too large.

In any case, you can check it if you wish, and I can also suggest an alternative way to identify the window - by using a thread name and a window band:
https://github.com/ramensoftware/windhawk-mods/blob/910d825f1c812f1f50da53a5d28c226581e0bf2e/mods/taskbar-on-top.wh.cpp#L1592-L1628

m1lhaus: For Windows 10 the thread description returns empty so I hope the class name is good enough identifier. For Windows 11 all works. Thx for the code snippet.

m417z: Note that without `GetWindowBand`, which you didn't include, you'll be also matching the virtual desktop switcher:

image

m1lhaus: Thank you for the hint. I wasn't even aware of this since I don't use virtual desktops. I added the check.

PR #3101, opened 2026-01-23
Explorer Tab Utility Window Hider - A New Mod

bcrtvkcs:
# Explorer Tab Utility Window Hider - Windhawk Mod
This is a Windhawk mod that prevents w4po's Explorer Tab Utility program (https://github.com/w4po/ExplorerTabUtility) window from appearing and being visible. The program continues to run in the background, but its window never appears.

## Problem
Explorer Tab Utility normally runs in the background, but sometimes its window becomes visible at computer startup and needs to be manually closed.

## Solution
This Windhawk mod completely blocks the Explorer Tab Utility window. The program continues to run in the background with all its functionality, but its configuration window never appears.

### DIFF
--- a/mods/explorer-tab-utility-window-hider.wh.cpp
+++ b/mods/explorer-tab-utility-window-hider.wh.cpp
@@ -2,7 +2,7 @@
 // @id              explorer-tab-utility-window-hider
 // @name            Explorer Tab Utility Window Hider
 // @description     Hides the Explorer Tab Utility window completely, allowing it to run only in the background
-// @version         1.0.1
+// @version         1.0.2
 // @author          BCRTVKCS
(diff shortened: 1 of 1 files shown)


mods/explorer-tab-utility-window-hider.wh.cpp
```diff
@@ -0,0 +1,241 @@
...
+*/
+// ==/WindhawkModReadme==
+
+#include <windhawk_utils.h>
+
+// Flag to track if we've found the main window
+HWND g_mainWindow = NULL;
+
+// Check if a window belongs to the main application window class
+bool IsMainAppWindow(HWND hWnd) {
```
m417z, line 38: Also check that the window belong to the current process.

PR #3154, opened 2026-01-28
Add new mod: VLC Discord Rich Presence (vlc-discord-rpc)

ciizerr:
This PR adds a new mod **VLC Discord Rich Presence** (`vlc-discord-rpc`).

### Mod Description
This mod seamlessly integrates VLC Media Player with Discord to display rich playback status, media metadata (movies, TV shows, anime), and quality tags (e.g., 4K, HDR).

### Key Features
*   **Smart Recognition:** Automatically parses media titles to identify Show Name, Season, Episode, or Movie titles.
*   **Quality Tags:** displaying tags for resolution (4K, 1080p) and HDR status.
*   **Interactive Buttons:** Adds a "Search This" button redirecting to Google, IMDb, or YouTube.
*   **Customizable:** Users can switch between "Default" and "Dark" icon themes and configure the Search Provider via Mod Settings.

### Verification
*   Checked that `@id` matches the filename.
*   Included valid `@github` metadata.
*   Verified functionality with VLC Web Interface enabled (Port 8080).

### DIFF
--- a/mods/vlc-discord-rpc.wh.cpp
+++ b/mods/vlc-discord-rpc.wh.cpp
@@ -2,7 +2,7 @@
 // @id              vlc-discord-rpc
 // @name            VLC Discord Rich Presence
 // @description     Shows your playing status, quality tags (4K/HDR), and interactive buttons on Discord.
-// @version         1.0.1
+// @version         1.0.3
 // @author          ciizerr
 // @github          https://github.com/ciizerr
 // @include         vlc.exe


mods/vlc-discord-rpc.wh.cpp
```diff
@@ -0,0 +1,500 @@
+// ==WindhawkMod==
+// @id              vlc-discord-rpc
+// @name            VLC Discord Rich Presence
+// @description     Shows your playing status, quality tags (4K/HDR), and interactive buttons on Discord.
+// @version         1.0.0
+// @author          ciizerr
+// @github          https://github.com/ciizerr
+// @include         vlc.exe
```
m417z, line 8: Does it have to run as part of vlc.exe? What if there's more than one vlc.exe process?
See also: https://github.com/ramensoftware/windhawk/wiki/Mods-as-tools:-Running-mods-in-a-dedicated-process

mods/vlc-discord-rpc.wh.cpp
```diff
@@ -0,0 +1,500 @@
...
+                            }
+                        }
+
+                        if (isConnected) {
+                            std::string state = isPlaying ? "Playing" : "Paused";
+                            if (query.empty()) query = "VLC Media Player";
+                            std::string btnUrl = GenerateButtonUrl(query, myProvider, myCustomUrl);
+                            
+                            std::string js = "{\"cmd\":\"SET_ACTIVITY\",\"args\":{\"pid\":" + NumToStr(GetCurrentProcessId()) + ",\"activity\":{";
+                            js += "\"details\":\"" + top + "\",";
```
m417z: What if `top` contains quotes or newlines? Same for other strings which are added to `js`.

mods/vlc-discord-rpc.wh.cpp
```diff
@@ -0,0 +1,500 @@
...
+
+    // ASSET NAMES
+    std::string assetLarge = myTheme + "vlc_icon";
+    std::string assetPlay  = myTheme + "play_icon";
+    std::string assetPause = myTheme + "pause_icon";
+    std::string assetStop  = myTheme + "stop_icon";
+
+    // NETWORK
+    HINTERNET hSession = WinHttpOpen(L"VLC-CPP/1.0", WINHTTP_ACCESS_TYPE_DEFAULT_PROXY, WINHTTP_NO_PROXY_NAME, WINHTTP_NO_PROXY_BYPASS, 0);
+    HINTERNET hConnect = WinHttpConnect(hSession, L"127.0.0.1", 8080, 0);
```
m417z: How about improving error handling? What happens if the connection fails?

mods/vlc-discord-rpc.wh.cpp
ciizerr: I verified the behavior with multiple VLC instances. The VLC HTTP Interface (port 8080) behaves as a singleton—only the first instance successfully binds to the port. Subsequent instances do not expose data on that port.

Therefore, using `@include` vlc.exe is safe:

If multiple VLC instances run, they all query localhost:8080.
They all receive the status of the first/primary instance.
This prevents race conditions where Discord might flip-flop between different videos.

Regarding the decision to use `@include` vlc.exe:

Since this mod's sole purpose is to bridge VLC data to Discord, running it as a global background tool (when VLC is closed) would consume unnecessary system resources (CPU/RAM) for no benefit.

By using `@include` vlc.exe, we ensure the mod follows the lifecycle of the application:
Starts only when the user needs it (opens VLC).
Stops immediately when the user is done (closes VLC).
This seems like the most resource-efficient approach for the end user.

mods/vlc-discord-rpc.wh.cpp
ciizerr: Thanks for addressing this issue.

I have implemented a SanitizeString helper function to handle this safely. Instead of complex escaping (which can sometimes fail with nested JSON), I chose a robust sanitization approach:

Quotes: Double quotes (") are automatically replaced with single quotes ('). This preserves readability (e.g., The "Best" Movie becomes The 'Best' Movie) while guaranteeing the JSON structure cannot be broken.

Control Characters: Backslashes and non-printable characters are stripped out.

I tested this specifically with edge cases like The "Best" Movie and The New \ Movie, and the payload now transmits correctly to Discord without crashing.

mods/vlc-discord-rpc.wh.cpp
ciizerr: Regarding error handling and connection drops:

I have completely refactored the Worker() loop to be self-healing.

Lazy Connection: The WinHTTP connection handles are now managed inside the main loop, not just at initialization.

Failure Detection: If WinHttpSendRequest fails (e.g., VLC is closed or restarted), the mod detects the failure immediately.

Auto-Recovery: Upon failure, the existing hConnect handle is closed and invalidated (NULL). The loop then enters a 'retry state' with a 2-second backoff delay.

Result: If the user restarts VLC while the mod is running, the mod automatically reconnects without requiring a reload.

PR #3160, opened 2026-01-29
Add Global Suffix Remover mod

thewerthon:
Adds a new mod that removes redundant suffixes (e.g., " - Google Chrome") from window titles globally. It implements hooks for SetWindowTextW to handle generic applications and FindResourceExW to specifically handle File Explorer logic.


m417z:
I updated the mod (and renamed it), please try Remove Taskbar Window Suffixes v1.1, let me know if anything is missing.

PR #3164, opened 2026-01-29
Added disable open/close anims mod

Lockframe:
Please let me know if you know a way to have this mod not use `@include *` with certain exclusions, without it being too slow to disable the animations.

### DIFF
--- a/mods/disable-open-close-anims.wh.cpp
+++ b/mods/disable-open-close-anims.wh.cpp
@@ -6,9 +6,6 @@
 // @author          Lockframe
 // @github          https://github.com/Lockframe
 // @include         *
-// @exclude         lsass.exe
-// @exclude         csrss.exe
-// @exclude         smss.exe
 // @compilerOptions -ldwmapi -luser32
 // ==/WindhawkMod==
 
@@ -222,3 +219,4 @@ BOOL Wh_ModInit() {
     Wh_SetFunctionHook((void*)AnimateWindow, (void*)AnimateWindow_Hook, (void**)&AnimateWindow_Original);
     return TRUE;
 }
+


mods/disable-open-close-anims.wh.cpp
```diff
@@ -0,0 +1,225 @@
...
+// @author          Lockframe
+// @github          https://github.com/Lockframe
+// @include         *
+// @exclude         lsass.exe
+// @exclude         csrss.exe
+// @exclude         smss.exe
+// @exclude         wininit.exe
+// @exclude         services.exe
+// @exclude         spoolsv.exe
+// @exclude         dwm.exe
```
m417z: Excluding these doesn't hurt, but they're already excluded by default in Windhawk:
https://github.com/ramensoftware/windhawk/wiki/Injection-targets-and-critical-system-processes#the-predefined-lists-of-excluded-processes
They're excluded for the mod even if they're removed from Windhawk global exclusions:
https://github.com/ramensoftware/windhawk/wiki/Injection-targets-and-critical-system-processes#mods-excluded-system-processes

mods/disable-open-close-anims.wh.cpp
```diff
@@ -0,0 +1,225 @@
...
+// ==/WindhawkModReadme==
+
+#include <windhawk_utils.h>
+#include <dwmapi.h>
+
+// ---------------------------------------------------------------------------
+// CONFIGURATION
+// ---------------------------------------------------------------------------
+#define RESTORE_ANIM_DELAY_MS 50
+#define TIMER_ID_RESTORE 0xDEAD
```
m417z: What if this id is already in use? There are two options I can think of:
* Generating a unique timer id with SetTimer by calling it as: `timerId = SetTimer(NULL, 0, , RESTORE_ANIM_DELAY_MS, RestoreAnimTimerProc)`. For remembering the timer id for that window, you can use a map or `SetProp`.
* Using a more unique id. I like copying the current timestamp from https://www.epochconverter.com/ which is highly unlikely to collide with anything.

mods/disable-open-close-anims.wh.cpp
m417z: I was commenting on all exclusions, not only the last 4 which GitHub decided to show in the preview above the comment.

PR #3169, opened 2026-01-30
Add: Breadcrumb Addressbar Open Newtab via middleclick

osmanonurkoc:
🐧 Motivation: The Linux Feel
For users coming from Linux environments (KDE Dolphin, GNOME Files, Thunar), middle-clicking a parent folder in the address bar to open it in a new tab is pure muscle memory. Windows 11 introduced Tabs, but missed this shortcut.

This mod bridges that gap. It allows you to middle-click any folder name in the breadcrumb bar to instantly open that location in a new background tab.

⚙️ How it Works (Technical Deep Dive)
Windows 11 Explorer (specifically versions using XAML Islands) presents unique challenges for automation. Standard window handles (HWND) do not correspond directly to breadcrumb buttons.

This mod uses a sophisticated "Universal UI Scraping" approach:

Threaded Hook: Installs a Low-Level Mouse Hook (WH_MOUSE_LL) on a dedicated native thread to detect middle-clicks without freezing the UI.
UI Automation (UIA): Identifies the clicked element (e.g., "Program Files") using Microsoft's UI Automation framework.
Heuristic Path Discovery: Instead of relying on specific Automation IDs (which Microsoft changes often), it scans the active window for any element containing a text string that looks like a file path (e.g., "C:\Users...").
Focus & Navigation: - Extracts the full path.
Truncates it to the clicked folder.
Simulates: Ctrl+T (New Tab) -> Alt+D (Focus Address Bar) -> Paste -> Enter.
The "Alt+D" step is crucial to ensure focus isn't lost to the file view.

### DIFF
--- a/mods/explorer-breadcrumb-mclick-newtab.wh.cpp
+++ b/mods/explorer-breadcrumb-mclick-newtab.wh.cpp
@@ -6,7 +6,7 @@
 // @author          osmanonurkoc
 // @github          https://github.com/osmanonurkoc
 // @include         explorer.exe
-// @compilerOptions -luiautomationcore -lole32 -loleaut32 -lshlwapi
+// @compilerOptions -luiautomationcore -lole32 -loleaut32 -luser32 -lshlwapi
 // ==/WindhawkMod==
 
 // ==WindhawkModReadme==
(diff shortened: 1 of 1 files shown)


osmanonurkoc:
> It can't be the solution because as I said, nothing happens, i.e. a tab isn't opened.
> 
>  0fUYiNbFii.mp4

https://jumpshare.com/s/0NxKTgD9bwOhHsTi6Lwk
I don't understand the problem, because it works for me. Are you using a different mod or configuration that might affect Explorer? I'll try to fix it.

osmanonurkoc:
> It doesn't work because `FindPathInWindow` returns `C:\`.

It should retrieve the address from the AddressEditBox and return the correct path. But you probably don't have that. Could you please tell me your OS version? So I can reproduce the error.

osmanonurkoc:
Fixed. I rewrited path parse logic completely.

mods/explorer-breadcrumb-mclick-newtab.wh.cpp
```diff
@@ -0,0 +1,474 @@
...
+                            HWND hForeground = GetForegroundWindow();
+                            pAutomation->ElementFromHandle(hForeground, &pRoot);
+                            if (pRoot) {
+                                fullPath = GetPathFromUI(pAutomation, pRoot);
+                                pRoot->Release();
+                            }
+                        }
+
+                        if (!fullPath.empty()) {
+                            size_t found = fullPath.rfind(clickedName);
```
m417z: This implementation results in incorrect behavior in some cases:

* If the clicked folder name is a substring of another folder, it won't work correctly. For example:
  `C:\Users\User\my_Users_backup`
  Middle-clicking on "Users" will result in `C:\Users\User\my_Users` which doesn't exist. This can be fixed by looking for a substring that also starts with `\` and ends with `\` or is at the end of the string.
* If a folder name exists more than once, for example:
  `C:\Users\User\Users`
   Clicking on such folder which is not the last folder will behave incorrectly too.

osmanonurkoc: I'll look into it. Thx for feedback.

osmanonurkoc: switched UIA + Input Injection method from complex COM to 100% accuracy. I tested all old problematic scenarios. And it works flawlessly.

PR #3174, opened 2026-01-31
Add Block Win+V mod

JiangFengning:


mods/block-win-v.wh.cpp
```diff
@@ -0,0 +1,40 @@
...
+# Block Win+V / 阻止 Win+V
+
+This mod blocks the Win+V keyboard shortcut (Clipboard History) in Windows.
+
+此模块阻止 Windows 中的 Win+V 键盘快捷键（剪贴板历史记录）。
+*/
+// ==/WindhawkModReadme==
+
+BOOL(*pOriginalRegisterHotKey)(HWND hWnd, int id, UINT fsModifiers, UINT vk);
+BOOL RegisterHotKeyHook(HWND hWnd, int id, UINT fsModifiers, UINT vk)
```
m417z: The calling convention is missing here, which is likely to cause a crash for 32-bit targets. Please add calling conventions. See:

mods/block-win-v.wh.cpp
```diff
@@ -0,0 +1,40 @@
+// ==WindhawkMod==
+// @id              block-win-v
+// @name            Block Win+V / 阻止 Win+V
+// @description     Blocks the Win+V keyboard shortcut (Clipboard History) / 阻止 Windows 的 Win+V 键盘快捷键（剪贴板历史记录）
```
m417z: Note that you can have a separate name/description per-language. For example:
https://github.com/ramensoftware/windhawk-mods/blob/161fe6155b4e2de349f68456abe63b55027bc78c/mods/run-box-no-ime.wh.cpp#L3-L6

PR #3179, opened 2026-01-31
Better file sizes in Explorer details v1.5

m417z:
* When showing folder sizes, the sizes are now shown in Tiles and Content folder view, not only in Details view. In addition, folder sizes are shown in the status bar, and in the details pane (enabled via OldNewExplorer). Note: An explorer restart might be required to apply these changes.
* Fixed the "Use IEC terms" option for locales, such as Korean, which don't have a space, e.g. "50MB" and not "50 MB".
* Fixed explorer freezing in some rare cases, possibly related to some network drives.

### DIFF
--- a/mods/explorer-details-better-file-sizes.wh.cpp
+++ b/mods/explorer-details-better-file-sizes.wh.cpp
@@ -2,7 +2,7 @@
 // @id              explorer-details-better-file-sizes
 // @name            Better file sizes in Explorer details
 // @description     Optional improvements: show folder sizes, use MB/GB for large files (by default, all sizes are shown in KBs), use IEC terms (such as KiB instead of KB)
-// @version         1.4.11
+// @version         1.5
 // @author          m417z
 // @github          https://github.com/m417z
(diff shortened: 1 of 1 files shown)


mode0192:
I’m reporting a reproducible issue that seems to be caused by this mod on my system.

I can reliably reproduce the following:
- Windows Search eventually gets stuck loading / becomes unusable.
- The problem does not happen immediately. It shows up after using the PC for a while.
- If I disable this mod, then restart the Windows Search service and Explorer, Search works again.
- If I enable the mod again, the issue comes back later.

I don’t yet know which exact code path is responsible, but the issue appears tied to this mod.

For context:
- I mainly use this mod for MB/GB formatting for large files.
- I’m not specifically testing the newer extra features.

Settings I am using in the mod:
- Show folder sizes -> Enabled via "Everything" integration
- Mix files and folders when sorting by size -> on
- Use MB/GB for large files -> on
- Use IEC terms -> off

The explorer-details-better-file-sizes version is 1.5

Screenshot 2026-04-05 064831

PR #3187, opened 2026-02-01
Make several attempts.

Anixx:
Make several attempts if unsuccessful.

### DIFF
--- a/mods/classic-theme-enable.wh.cpp
+++ b/mods/classic-theme-enable.wh.cpp
@@ -218,7 +218,10 @@ extern "C" NTSTATUS NTAPI NtOpenSection(
     IN POBJECT_ATTRIBUTES ObjectAttributes
 );
 
-BOOL TrySetThemeSectionSecurity(int attemptNumber) {
+HANDLE g_hThread = NULL;
+volatile BOOL g_bStopThread = FALSE;
+
+BOOL TrySetThemeSectionSecurity() {
     DWORD sessionId;
     ProcessIdToSessionId(GetCurrentProcessId(), &sessionId);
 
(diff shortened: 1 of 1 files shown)


mods/classic-theme-enable.wh.cpp
```diff
@@ -209,61 +212,74 @@ it will require the UAC authorization). Alternatively you can use the
...
+
+    for (int attempt = 1; attempt <= MAX_ATTEMPTS; attempt++) {
+        if (TrySetThemeSectionSecurity(attempt)) {
+            Wh_Log(L"Success on attempt %d\n", attempt);
+            return TRUE;
+        }
+        
+        if (attempt < MAX_ATTEMPTS) {
+            Wh_Log(L"Attempt %d failed, waiting %d ms before retry...\n", attempt, RETRY_DELAY);
+            Sleep(RETRY_DELAY);
```
m417z, line 280: If injected before winlogon.exe starts running, this will block its start. If I understand correctly, these retry attempts won't help, since you're waiting for winlogon.exe to create the section, but winlogon.exe won't run before `Wh_ModInit` returns.

A possible solution can be something like:
```cpp
BOOL Wh_ModInit() {
    if (TrySetThemeSectionSecurity()) {
        return FALSE;  // no need to keep the mod loaded
    }

    CreateThread(..., ThreadForRetries, ...);
    return TRUE;
}
```

Anixx: It injects winlogon.exe only because it is the only process with appropriate privilegies for the task. It has nothing to do with what winlogon does. This modified mod seemingly works better than the original with only one attempt.

m417z: Maybe it's unintended, but the `Windows\ThemeSection` section is owned by winlogon.exe.

image

Anixx: I've put it into a separate thread.

PR #3191, opened 2026-02-01
New Mod: reMarkable Polish

SteffanDonal:
Adds new mod - reMarkable App Polish

TYVM for Windhawk 💜

### DIFF
--- a/mods/remarkable-polish.wh.cpp
+++ b/mods/remarkable-polish.wh.cpp
@@ -14,7 +14,9 @@
 /*
 # reMarkable App Polish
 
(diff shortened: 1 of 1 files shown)


mods/remarkable-polish.wh.cpp
```diff
@@ -0,0 +1,344 @@
...
+    return DefSubclassProc(hWnd, uMsg, wParam, lParam);
+}
+
+static void AttachSubclass(HWND hWnd) {
+    if (!hWnd || !IsWindow(hWnd)) {
+        Wh_Log(L"[AttachSubclass] invalid hwnd=%p", hWnd);
+        return;
+    }
+
+    if (SetWindowSubclass(hWnd, TargetSubclassProc, 1, 0)) {
```
m417z, line 284: The subclass is never removed, which will cause a crash when the mod is unloaded. Do you think it's worth handling, or is it a short-lived window?

SteffanDonal: It _is_ a pretty short-lived window, but I figured I'd have a crack at handling it properly. Should be resolved now, barring any silly mistakes (tested on my end at least)

PR #3194, opened 2026-02-01
Double f2: timeless method

leaumar:
As requested by @nmatt, Double F2 could work without even taking time into account. I tested this on some files and found no problemms, changing the settings also works fine.

Changelog:
- minor refactor
- option to loop selection based on current selection instead of hotkey timing


mods/explorer-double-f2-rename-extension.wh.cpp
m417z, line 466: I just noticed that the mod defines `Wh_ModInit` as returning `void`, which is incorrect, it should be `BOOL` instead. It might cause the mod to fail to loaded if the return register happens to be zero. Please fix it in the next update.

leaumar: I'm trying to use the github web ide for this but wow the ux is 💩

PR #3226, opened 2026-02-07
Classic Minimize/Maximize Animations 1.1.0

aubymori:
- Hook SystemParametersInfo to handle SPI_(GET|SET)ANIMATION
  - Enabling/disabling "Animate windows when minimizing and maximizing" with the mod enabled will now control the mod's animation, not the system's
- Call SystemParametersInfo to automatically disable the DWM window animations at init and to enable them at uninit if MinAnimate is still enabled

### DIFF
--- a/mods/classic-min-max-animations.wh.cpp
+++ b/mods/classic-min-max-animations.wh.cpp
@@ -1080,9 +1080,19 @@ VS_FIXEDFILEINFO *GetModuleVersionInfo(HMODULE hModule, UINT *puPtrLen)
 
 BOOL Wh_ModInit(void)
 {
-    /* Disable window animations */
     ANIMATIONINFO ai;
     ai.cbSize = sizeof(ai);
+
+    if (Wh_GetIntValue(L"MinAnimate", -1) == -1)
+    {
+        SystemParametersInfoW(SPI_GETANIMATION, sizeof(ai), &ai, FALSE);
+        if (ai.iMinAnimate)
+        {
+            Wh_SetIntValue(L"MinAnimate", 1);
+        }
+    }
+    
+    /* Disable window animations */
     ai.iMinAnimate = FALSE;
     SystemParametersInfoW(SPI_SETANIMATION, sizeof(ai), &ai, TRUE);
 


mods/classic-min-max-animations.wh.cpp
```diff
@@ -343,6 +339,82 @@ BOOL WINAPI SetWindowPos_hook(
...
+        case SPI_SETANIMATION:
+        {
+            LPANIMATIONINFO lpai = (LPANIMATIONINFO)pvParam;
+            if (!lpai || lpai->cbSize != sizeof(*lpai))
+            {
+                *pfResult = FALSE;
+                return true;
+            }
+
+            Wh_SetIntValue(L"MinAnimate", lpai->iMinAnimate != 0);
```
m417z, line 377: It will be set only when settings are changed, right? So if I have animations enabled, then enabled the mod, and then disable it, my animations won't be restored?

aubymori: > It will be set only when settings are changed, right? So if I have animations enabled, then enabled the mod, and then disable it, my animations won't be restored?

I'm not sure exactly what you're referring to, but `Wh_ModUninit` does call SPI to enable the DWM animations again if the mod has MinAnimate enabled.

m417z: I mean, what will trigger a `SystemParametersInfo(SPI_SETANIMATION)` call which will end up calling `Wh_SetIntValue(L"MinAnimate")`, which is necessary for it to be restored?

aubymori: > I mean, what will trigger a `SystemParametersInfo(SPI_SETANIMATION)` call which will end up calling `Wh_SetIntValue(L"MinAnimate")`, which is necessary for it to be restored?

Let me just explain the behavior of the entire update.

`IsMinAnimate` calls `Wh_GetIntValue` with a default value of 1, so if the user has not configured the mod previously, the mod's classic animation will be enabled, and DWM animations will be re-enabled upon uninit.

Unticking the selected checkbox in the below screenshot will disable `MinAnimate`, disabling the mod's classic animation and making so that DWM animations will not be re-enabled upon uninit.

image

Basically, this update just makes it act more like versions of Windows that actually had this animation.

m417z: > DWM animations will be re-enabled upon uninit

OK, and what if animations were disabled before using the mod? The mod will enable them on uninit?

> Unticking the selected checkbox in the below screenshot will disable `MinAnimate`

Only while the mod is loaded.

I had a similar situation with the "Taskbar auto-hide when maximized" mod which enabled auto-hide, and users complained that the mod made permanent changes even after being unloaded. I updated the mod to avoid making persistent changes. In your case, if that's not possible, perhaps it's worth remembering the initial user preference on init, something like:

```cpp
if (Wh_GetIntValue(L"MinAnimate", -1) == -1) {
  SystemParametersInfoW(SPI_GETANIMATION, ...);
  Wh_SetIntValue(L"MinAnimate", lpai->iMinAnimate != 0);
}
```

aubymori: > OK, and what if animations were disabled before using the mod? The mod will enable them on uninit?

Yes. Admittedly, this isn't ideal behavior, but the main concern that caused me to do this is migrating existing users from 1.0.2 and before that have the setting disabled already. I'm not really sure how to do it nicely...

m417z: Since these users disabled this option manually, it's reasonable to assume that it should also be enabled manually. Up to you, but either way I think it's worth mentioning it in the readme.

aubymori: My idea right now is that users with animations disabled and `MinAnimate` unset will have animations with the mod, and animations will continue to be disabled once disabling the mod, assuming they did not set `MinAnimate` while it was enabled.

aubymori: Actually, I can't do that nicely due to multiple processes using the mod. I also can't really read the real value at startup nicely for the same reason... I think the behavior has to be kept as is.

aubymori: No, wait, I just managed to enact that behavior. Check out the latest commit.

PR #3235, opened 2026-02-08
Update Taskbar Fade to v2.0 (Performance Improvement + Readme)

lukvbp:
This update addresses the feedback from the initial release regarding performance and documentation.

Changelog v2.0:

Performance / Idle Logic: I refactored the main loop to address the CPU usage concern. The mod now separates its logic: heavy window geometry updates are throttled to 1Hz (1 check/s), while lightweight mouse detection remains at 15Hz to ensure the taskbar wakes up instantly without lag.
Note: I decided to keep the mouse detection at 15Hz because testing showed that lowering it introduced noticeable wake-up lag. However, because I stripped the heavy API calls out of this loop, the 15Hz check is now lightweight (effectively just GetCursorPos), reducing CPU cost.

Documentation: Added a full WindhawkModReadme block with a description, features, and configuration guide.

Improvements: Switched to a Cubic Ease-Out curve for a smoother feel. The animation is now time-based (deterministic), ensuring consistent speed regardless of system frame rate. Also Implemented a new pipeline to handle state transitions, added monitor hotplug detection to handle multi-monitor setups correctly and made some other adjustments.

Thanks for the feedback on the previous PR and for your work on Windhawk, I really appreciate it!

### DIFF
--- a/mods/taskbar-fade.wh.cpp
+++ b/mods/taskbar-fade.wh.cpp
@@ -80,7 +80,7 @@ struct {
     DWORD idleTimeoutMS;
     int fadeDuration;
     int idleFadeDuration;
-    int frameTimeMs; 
+    int frameTimeMs;
     double targetDimAlpha;
 } g_cache;
 
@@ -107,7 +107,7 @@ void LoadSettings() {
     if (rawTimeout < 1) rawTimeout = 1;
     if (rawTransp < 0) rawTransp = 0;
     if (rawTransp > 100) rawTransp = 100;
-    
+
     // FPS Clamp: 15 to 120
     if (rawFPS < 15) rawFPS = 15;
     if (rawFPS > 120) rawFPS = 120;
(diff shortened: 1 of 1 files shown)


mods/taskbar-fade.wh.cpp
```diff
@@ -299,172 +405,48 @@ void WhTool_ModUninit() {
...
-
+    RestoreTaskbarStyle(FindWindow(L"Shell_TrayWnd", NULL));
     HWND hSec = NULL;
     while ((hSec = FindWindowEx(NULL, hSec, L"Shell_SecondaryTrayWnd", NULL)) != NULL) {
-        EnsureLayeredStyle(hSec, false);
+        RestoreTaskbarStyle(hSec);
     }
 }
 
-// --- WINDHAWK TOOL BOILERPLATE (DO NOT EDIT) ---
```
m417z, line 312: As the note says, please don't edit the code below which is taken from:

lukvbp:
Sorry, my bad. Fixed.

PR #3249, opened 2026-02-10
Add Caffeine mod

almas-cp:
Tray icon utility to keep your PC awake. Left-click to toggle, right-click for timed options (15m–4h or indefinite). Shows countdown on hover, notifies on expiry, and persists state across restarts.

### DIFF
--- a/mods/caffeine.wh.cpp
+++ b/mods/caffeine.wh.cpp
@@ -123,6 +123,7 @@ these flags.
 
 
 #define WM_TRAYICON (WM_USER + 1)
+#define WM_SETTINGS_CHANGED (WM_USER + 2)
 #define ID_TRAYICON 1002
 #define ID_TIMER_COUNTDOWN 2001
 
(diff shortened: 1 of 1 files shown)


mods/caffeine.wh.cpp
```diff
@@ -0,0 +1,1296 @@
...
+
+bool g_isToolModProcessLauncher;
+HANDLE g_toolModProcessMutex;
+
+
+void WINAPI EntryPoint_Hook() {
+    Wh_Log(L"Tool mod process entry point hooked. Starting message loop.");
+
+    MSG msg;
+    while (GetMessage(&msg, nullptr, 0, 0)) {
```
m417z: Originally, `EntryPoint_Hook` just calls `ExitThread(0);`. Please restore [the original code](https://github.com/ramensoftware/windhawk/wiki/Mods-as-tools:-Running-mods-in-a-dedicated-process) and adjust the mod.

The problem with this approach is that different mod callbacks are called from different threads. So for example, `WhTool_ModSettingsChanged` is called from a different thread, but it calls functions that assume that the original thread is running, such as `SetThreadExecutionState`. The same problem occurs in `WhTool_ModUninit`, but it doesn't matter that much because the process exits shortly after that.

Change the mod such that `WhTool_ModInit` creates a new thread, and `WhTool_ModSettingsChanged` communicates that that thread (or to the tray window) that settings are changed. `WhTool_ModUninit` can do the same, or it can be just empty, there's no much need to apply anything since the process will exit in a moment anyway.

PR #3254, opened 2026-02-11
Explorer Copy-as-path Remove quotes

ilyfairy:


m417z:
How does it differ from https://windhawk.net/mods/remove-quotes-from-ctrl-shift-c in terms of functionality?

ilyfairy:
> How does it differ from https://windhawk.net/mods/remove-quotes-from-ctrl-shift-c in terms of functionality?

Thanks for the pointer — I didn't realize there’s already a mod covering this.

ilyfairy:
> How does it differ from https://windhawk.net/mods/remove-quotes-from-ctrl-shift-c in terms of functionality?


I reopened this because after testing more carefully, I found that the existing mod doesn't actually work correctly for me. When I first replied, I assumed it covered the same functionality, but that doesn't seem to be the case.

With `remove-quotes-from-ctrl-shift-c`, the pasted output looks like this:

C:\Windows\System"
"C:\Windows\System32"
"C:\Windows\SystemApps

But with my implementation, the pasted output is correct:

C:\Windows\System
C:\Windows\System32
C:\Windows\SystemApps

So the existing mod doesn't produce the expected result in this case, while my implementation does.

Also, I can't get in touch with the original mod author. Their repository (https://github.com/draxas/windhawk-mods) has issues disabled, so I couldn't report the problem there either.

PR #3265, opened 2026-02-12
Disable Thumbnail Minimum Size 1.3: Fix Windows 11 compatibility

### LINKED ISSUE #3203: [disable-thumbnail-minimum-size] Doesn't work as claimed
Author: @Leymonaide

Effect:
Image

Leymonaide:
Closes #3203

### DIFF
--- a/mods/disable-thumbnail-minimum-size.wh.cpp
+++ b/mods/disable-thumbnail-minimum-size.wh.cpp
@@ -229,7 +229,7 @@ BOOL Wh_ModInit() {
 
     // The duplicate functions were removed from shell32 in Windows 11, so we
     // will only attempt to hook them on Windows 10.
-    if (osvi.dwBuildNumber < 22000)
+    if (osvi.dwBuildNumber < 26100)
     {
         if (!WindhawkUtils::HookSymbols(shell32, shell32Hooks, ARRAYSIZE(shell32Hooks)))
         {


mods/disable-thumbnail-minimum-size.wh.cpp
```diff
@@ -224,8 +224,10 @@ BOOL Wh_ModInit() {
 
     if (!WindhawkUtils::HookSymbols(shell32, shell32Hooks, ARRAYSIZE(shell32Hooks)))
     {
+        // This failure is non-fatal as the functionality is mostly contained in
+        // windows.storage.dll, and the functions are removed from shell32 in
+        // Windows 11.
```
m417z: A better solution would be to avoid calling `WindhawkUtils::HookSymbols` for Windows 11, because the first call to `WindhawkUtils::HookSymbols` is slow, and if it fails, then it's slow every time. More details below if you're interested.

## More details

When `WindhawkUtils::HookSymbols` is called, the mod tries to download symbols from the online cache server, then from Microsoft servers. Upon success, the offsets are cached and the next `WindhawkUtils::HookSymbols` calls are very fast. That's the happy flow, no problem here.

If `WindhawkUtils::HookSymbols` doesn't succeed - no internet, no symbols were published, or the target binary doesn't have the requested symbols - then the process is retried on every call of `WindhawkUtils::HookSymbols`, i.e. for every new process. That's not good as is means that every process that's launched looks for symbols in the online cache server and then in Microsoft's servers. It causes a delay, and may cause the progress dialog to pop up if the initialization takes more than 3 seconds (by default).

The latest version of Windhawk tries to reduce the impact by recording the failure and retrying every several hours, and not every time. But still, there's a negative impact on performance.

What the mod can do in this case is to detect environments which are unsupported and fail gracefully without having `WindhawkUtils::HookSymbols` needlessly go online only to return false.

mods/disable-thumbnail-minimum-size.wh.cpp
```diff
@@ -222,12 +227,17 @@ BOOL Wh_ModInit() {
         return FALSE;
     }
 
-    if (!WindhawkUtils::HookSymbols(shell32, shell32Hooks, ARRAYSIZE(shell32Hooks)))
+    // The duplicate functions were removed from shell32 in Windows 11, so we
+    // will only attempt to hook them on Windows 10.
+    if (osvi.dwBuildNumber < 22000)
```
m417z: I checked it, and it was present in versions before 26100 (e.g. 10.0.22621.6489). So a better check would be:
```suggestion
    if (osvi.dwBuildNumber < 26100)
```

Meteony:
ITS PERFECT THX

PR #3222, opened 2026-02-06
Add Taskbar Auto-Hide Instant Show mod

Bo0ii:
## New Mod: Taskbar Auto-Hide Instant Show

Removes the built-in delay before the auto-hidden taskbar starts appearing and adds custom animation types.

### Features
- **Zero delay** - Taskbar reacts instantly when mouse hits the screen edge (hooks SetTimer to eliminate the ~300-500ms Windows delay)
- **5 animation types**: Slide (with speedup), Elastic (spring overshoot), Bounce, Fade (opacity), Slide + Fade
- **Configurable** - Animation speed, duration, and frame rate settings
- **DWM synced** - Fade animations use DwmFlush() for flicker-free rendering
- **Compatible** - Windows 10 & 11, with ExplorerPatcher support

### Preview

**Elastic**
Elastic

**Fade**
Fade

**Slide + Fade**
Slide + Fade

PR #3279, opened 2026-02-13
Taskbar Auto-Hide Instant Show v2.1 Bug Fix

Bo0ii:
## Summary
- Fix fade cleanup flash on hide (alpha set before removing WS_EX_LAYERED)
- Add animation cancellation to prevent state corruption on rapid show/hide
- Suppress pre-animation flash by updating TrayUI state while window is hidden
- Add mouse edge detection thread for monitors with no visible windows

those are bugs people reported in my private repo and here #3222

### DIFF
--- a/mods/taskbar-autohide-instant-show.wh.cpp
+++ b/mods/taskbar-autohide-instant-show.wh.cpp
@@ -2,7 +2,7 @@
 // @id              taskbar-autohide-instant-show
 // @name            Taskbar Auto-Hide Instant Show
 // @description     Removes the delay before the taskbar appears with custom animation types (elastic, bounce, fade, etc.)
-// @version         2.0
+// @version         2.1
 // @author          Bo0ii
 // @github          https://github.com/Bo0ii
 // @homepage        https://github.com/Bo0ii/windhawk-mods
@@ -100,6 +100,12 @@ For all other modes, use *Show/Hide duration (ms)* to control timing.
   $description: >-
     Enable this if you use ExplorerPatcher or similar to use the old
     Windows 10 taskbar on Windows 11.
+- edgeDetection: false
+  $name: Edge detection for empty monitors
+  $description: >-
+    Enable this if the taskbar does not appear when hovering the screen
+    edge on a monitor with no visible windows. Creates a background thread
+    that polls the cursor position.
 */
 // ==/WindhawkModSettings==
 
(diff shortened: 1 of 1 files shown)


mods/taskbar-autohide-instant-show.wh.cpp
```diff
@@ -719,6 +875,8 @@ BOOL Wh_ModInit() {
 
     g_initialized = true;
 
+    StartEdgeCheckThread();
```
m417z: Why is this workaround needed? Is it a Windows bug, or is it needed because something the mod does?

Bo0ii: It's a Windows OS limitation. When no visible (non-minimized) windows exist on a monitor, Windows never triggers TrayUI::SlideWindow() at all. Since the mod only hooks that function, the taskbar won't show on empty monitors without this workaround. The thread polls the cursor position and posts WM_TIMER to trigger the existing show logic.

m417z: I don't see it. How can I reproduce it?

https://github.com/user-attachments/assets/3d83df0f-660d-46f0-b078-e15d778d1d4d

Bo0ii: It's a Windows limitation, not caused by the mod. To reproduce:

  1. Enable taskbar auto-hide
  2. Minimize all windows on a monitor (or use a second monitor with no windows)
  3. Move mouse to the bottom screen edge
  4. The taskbar won't show  Windows never calls TrayUI::SlideWindow when no visible (non-minimized) windows exist on that monitor

This happens with or without the mod. The mod only hooks SlideWindow, so if Windows never calls it, there's nothing to intercept. The edge detection thread works around this by detecting the cursor at the screen edge and posting WM_TIMER with ID 3 (the same unhide timer Windows uses internally).

note : not all computers have this issue i had to isntall a new VM with some curppted files of windows to reproduce the bug. this isnt a mod problem but user experince problem some users have diff windows integration and curppoted files so the mod behaves differntly but now it should ( I HOPE ) works with all 

check the issue in my repo https://github.com/Bo0ii/windhawk-mods/issues/2

m417z: That's what I did, the recording shows the second monitor with no windows. If it doesn't happen on all computers, then perhaps most users don't need it. I'm not sure creating a new thread which that runs all the time by default is justified. Perhaps make it optional, what do you think?

Bo0ii: Good point, I'll make it an optional setting disabled by default. That way users who experience the issue can enable it without adding overhead for everyone else. it happens usally on 60hz seond monitor

PR #3280, opened 2026-02-13
Invisible Window Borders v1.1.0 Bug Fix

Bo0ii:
## Summary
- Add ExcludeList setting for comma-separated process names
- Mod skips initialization entirely for excluded processes
- Useful for Chromium-based apps (Edge, Chrome) that render their own borders

### DIFF
--- a/mods/invisible-borders.wh.cpp
+++ b/mods/invisible-borders.wh.cpp
@@ -2,7 +2,7 @@
 // @id              invisible-borders
 // @name            Invisible Window Borders
 // @description     Makes window borders invisible while keeping rounded corners
-// @version         1.0.0
+// @version         1.1.0
 // @author          Bo0ii
 // @github          https://github.com/Bo0ii
 // @homepage        https://github.com/Bo0ii/windhawk-mods
@@ -17,7 +17,7 @@
 
 Makes window borders invisible while preserving rounded corners.
 
-![Comparison](https://raw.githubusercontent.com/Bo0ii/windhawk-mods/main/invisible-borders/onoff.png)
+![Comparison](onoff.png)
 
 ## Features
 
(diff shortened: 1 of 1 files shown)


mods/invisible-borders.wh.cpp
```diff
@@ -17,7 +17,7 @@
 
 Makes window borders invisible while preserving rounded corners.
 
-![Comparison](https://raw.githubusercontent.com/Bo0ii/windhawk-mods/main/invisible-borders/onoff.png)
+![Comparison](onoff.png)
```
m417z, line 20: Seems like a mistake. Restore the screenshot link.

mods/invisible-borders.wh.cpp
```diff
@@ -185,6 +238,17 @@ BOOL Wh_ModInit() {
...
 
+    std::vector<std::wstring> excludeList;
+    LPCWSTR excludeStr = Wh_GetStringSetting(L"ExcludeList");
+    ParseExcludeList(excludeStr, excludeList);
+    Wh_FreeStringSetting(excludeStr);
+    g_isExcludedProcess = CheckExcludedProcess(excludeList);
+
+    if (g_isExcludedProcess) {
+        Wh_Log(L"Process excluded from invisible borders");
+        return FALSE;
```
m417z, line 249: Since the mod is unloaded here, you don't need the `g_isExcludedProcess` global variable and the check in `IsValidWindow`.

mods/invisible-borders.wh.cpp
Bo0ii: That's a mistake from copying the file. The upstream had the full raw GitHub URL, our local dev copy had just onoff.png. Needs to be restored. thanks

mods/invisible-borders.wh.cpp
Bo0ii: Since return FALSE from Wh_ModInit unloads the mod entirely, the hooks never fire. The global variable and the IsValidWindow check are dead code. Should remove both. thanks for that as well

PR #3285, opened 2026-02-14
Create better-taskbar-styler.wh.cpp

batata123muitoboa-dot:
Taskbar Styler (FANMADE) is a fixed Windows 11 Taskbar Styler, because i found some bugs on the original one, so i made a new one to fix it :)


batata123muitoboa-dot:
> What are the bugs? It makes more sense to fix the existing mod if there are any issues. You can submit pull requests here: https://github.com/m417z/my-windhawk-mods

like, the surface theme is "getting out" of my screen, and the docklike i just didnt like it at all
and sorry, im a new developer so i dont know some things

PR #3289, opened 2026-02-14
Add Icon Customizer mod

FireBlade211:
# Icon Customizer
Allows you to override almost 100 Windows icons to anything you like.

## Override an icon
To change an icon, open the mod settings and change the icons you like by entering the path to an ICO, EXE, or DLL file.
Optionally you can also specify an index, for example:

```
C:\Windows\System32\shell32.dll,32
```

## How it works
This mod adds a hook for `SHGetStockIconInfo`, which Windows apps call whenever they need to retrieve a stock icon. It then
 checks if the icon for the specified ID is overriden and gives Windows the new icon.

## Note
This mod will only work on apps that load icons themselves, and will not work on built-in Windows dialogs such as message boxes.
This is because those dialogs do not utilize the `SHGetStockIconInfo` function.


mods/icon-customizer.wh.cpp
```diff
@@ -0,0 +1,889 @@
...
+    settings.SETTINGS = Wh_GetStringSetting(L"SETTINGS");
+    settings.DRIVEHDDVD = Wh_GetStringSetting(L"DRIVEHDDVD");
+    settings.DRIVEBD = Wh_GetStringSetting(L"DRIVEBD");
+    settings.MEDIAHDDVDROM = Wh_GetStringSetting(L"MEDIAHDDVDROM");
+    settings.MEDIAHDDVDR = Wh_GetStringSetting(L"MEDIAHDDVDR");
+    settings.MEDIAHDDVDRAM = Wh_GetStringSetting(L"MEDIAHDDVDRAM");
+    settings.MEDIABDROM = Wh_GetStringSetting(L"MEDIABDROM");
+    settings.MEDIABDR = Wh_GetStringSetting(L"MEDIABDR");
+    settings.MEDIABDRE = Wh_GetStringSetting(L"MEDIABDRE");
+    settings.CLUSTEREDDRIVE = Wh_GetStringSetting(L"CLUSTEREDDRIVE");
```
m417z, line 857: Strings loaded with `Wh_GetStringSetting` should be freed with `Wh_FreeStringSetting`. Alternatively, you can use `WindhawkUtils::StringSetting`: https://github.com/ramensoftware/windhawk/wiki/Development-tips

mods/icon-customizer.wh.cpp
```diff
@@ -0,0 +1,982 @@
...
+    Wh_FreeStringSetting(settings.MEDIABDR);
+    Wh_FreeStringSetting(settings.MEDIABDRE);
+    Wh_FreeStringSetting(settings.CLUSTEREDDRIVE);
+}
+
+// The mod settings were changed, reload them.
+void Wh_ModSettingsChanged() {
+    Wh_Log(L"SettingsChanged");
+
+    LoadSettings();
```
m417z, line 987: The strings should be freed here too.

Issue #2936, opened 2026-01-04
[Translucent Windows] Notepad invisible text


AltyFox:
@Undisputed00x 

Issues with notepad.exe
Worked wonderfully before, and created a nice, beautiful effect on notepad, however now text is completely invisible with no configuration changes.

<img width="1408" height="658" alt="Image" src="https://github.com/user-attachments/assets/4046e68d-88c5-4da5-a5ad-f39ef5323910" />

<img width="1408" height="658" alt="Image" src="https://github.com/user-attachments/assets/d4603359-9444-4add-af70-a6b18d4a30aa" />

AltyFox:
Closing, as I just read the FAQ about HDR.

I have a new HDR display

PR #3311, opened 2026-02-16
Translucent Windows v.1.7.4

Undisputed00x:
* Fix black background system colors in some programs (e.g. Office) using mods process ruled settings
* Minor custom theme additions

### DIFF
--- a/mods/translucent-windows.wh.cpp
+++ b/mods/translucent-windows.wh.cpp
@@ -2,7 +2,7 @@
 // @id              translucent-windows
 // @name            Translucent Windows
 // @description     Enables native translucent effects in Windows 11
-// @version         1.7.3
+// @version         1.7.4
 // @author          Undisputed00x
 // @github          https://github.com/Undisputed00x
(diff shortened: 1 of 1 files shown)


artama08:
thanks for your hard work, I really appreciate, 

But I got an issue that when I enable this option, the text on notepad was gone and invisible on the actual view, but when I take the screenshot/snipping it's fine,

artama08:
> > thanks for your hard work, I really appreciate,
> > But I got an issue that when I enable this option, the text on notepad was gone and invisible on the actual view, but when I take the screenshot/snipping it's fine,
> 
> I assume you have HDR enabled. This is a bad side effect of extending translucent effects to the client area where uncomposed text is rendered like in Notepad. Unfortunately, Windows Notepad uses DirectWrite rendering for its text, and the mod only modifies GDI-rendered text to render correctly in HDR.
> 
> Similar issue: #2936

ah I see, i'll try to disable the HDR option,
thankya

PR #3317, opened 2026-02-16
Taskbar Volume Control v1.3

m417z:
* Added an option to scroll anywhere on the screen while holding configurable combination of modifier keys to change the system volume.
* Added an option to scroll at the taskbar position even when a full screen window covers the taskbar.
* Added an option for custom regions along the taskbar where scrolling will control the system volume. For example, "100-200" (pixels) or "20%-50%" (percentage of taskbar length).
* Improved the volume change step option, which is now compatible with all volume indicators.
* Improved the "No automatic mute toggle" option, which is now compatible with all volume indicators.


brunoosti:
windows key alone with scroll is activating start menu after release.

PR #3327, opened 2026-02-17
Update Magnifier Headless mod to version 1.0

bcrtvkcs:
## Huge Code Refactor - v1.0 Initial Release

refactor: Remove over-engineering and fix Windhawk compliance issues

- Remove 7 unnecessary API hooks (`AnimateWindow`, `BringWindowToTop`,
  `SetForegroundWindow`, `UpdateLayeredWindow`, `SetLayeredWindowAttributes`,
  `SetWindowRgn`, `DwmSetWindowAttribute`) — the 4 core hooks
  (`CreateWindowExW`, `ShowWindow`, `SetWindowPos`, `SetWindowLongPtrW`) fully
  cover all window visibility paths for Magnify.exe
- Remove `CallWndProc` hook + window subclassing infrastructure (150+ lines)
  that duplicated what the API hooks already handle
- Remove LRU cache (16-entry tick-based eviction for a process with ~5
  windows)
- Remove Safe wrapper functions with retry mechanisms (`SafeSetParent Sleep+retry`, `SafeSetWindowLongPtrW`, `SafeGetClassName`, `SafeIsWindow`)
- Remove `AutoCriticalSection` class and CRITICAL_SECTION — no longer needed without subclassing or cache
- Remove `CSpNotify` and `MSCTFIME` from `IsMagnifierWindow` detection — these are system-wide accessibility/IME windows, hiding them breaks accessibility features in an accessibility tool
- Remove GDI+ helper window detection — not Magnifier-specific
- Remove MSVC-only #pragma comment(lib) — Windhawk uses Clang/mingw-w64
- Remove redundant #include directives (`windhawk_api.h` is auto-included)
- Remove `@compilerOptions -lcomctl32` (only needed for removed subclassing)
- Fix `EnumWindows` to scope to current process only via `GetCurrentProcessId`
- Rewrite `WindhawkModReadme` section to be user-focused instead of implementation docs
- Bump version to 1.0

fix: Hide Magnifier Touch window from Alt+Tab
The touch overlay was still appearing in Alt+Tab because `ShowWindow_Hook` was letting show commands pass through for touch overlay windows.
- `ShowWindow_Hook` blocks show commands for touch overlays (same as magnifier)
- `SetWindowPos_Hook` adds `SWP_HIDEWINDOW` flag for touch overlays
- `CreateWindowExW_Hook` calls `ShowWindow` (`SW_HIDE`) after creating touch overlay
- `CreateWindowExW_Hook` strips `WS_VISIBLE` from touch overlay before creation

### DIFF
--- a/mods/magnifier-headless.wh.cpp
+++ b/mods/magnifier-headless.wh.cpp
@@ -2,13 +2,12 @@
 // @id              magnifier-headless
 // @name            Magnifier Headless Mode
 // @description     Blocks the Magnifier window creation, keeping zoom functionality with win+"-" and win+"+" keyboard shortcuts.
-// @version         0.9.5.4
+// @version         1.0
 // @author          BCRTVKCS
 // @github          https://github.com/bcrtvkcs
 // @twitter         https://x.com/bcrtvkcs
 // @homepage        https://grdigital.pro
 // @include         magnify.exe
-// @compilerOptions -lcomctl32
 // ==/WindhawkMod==
 
 // ==WindhawkModReadme==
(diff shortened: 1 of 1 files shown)


mods/magnifier-headless.wh.cpp
```diff
@@ -388,549 +135,146 @@ LONG_PTR WINAPI SetWindowLongPtrW_Hook(HWND hWnd, int nIndex, LONG_PTR dwNewLong
...
-
-    // Mark initialization as complete (atomic operation)
-    InterlockedExchange(&g_lInitialized, 1);
+    g_bInitialized = TRUE;
 
-    // Enumerate and hide any existing magnifier windows that were created before mod loaded
-    Wh_Log(L"Magnifier Headless: Enumerating existing windows to hide any magnifier-related windows...");
-    EnumWindows(EnumWindowsProc_HideMagnifierWindows, 0);
+    // Hide any magnifier windows that already exist
+    EnumWindows(EnumWindowsProc_HideMagnifier, 0);
```
m417z: This won't do anything, since `EnumWindowsProc_HideMagnifier` tries to use `ShowWindow_Original` which is always null in `Wh_ModInit`, because hooks aren't set yet. You can either use `ShowWindow`, or run `EnumWindows` from `Wh_ModAfterInit`,

mods/magnifier-headless.wh.cpp
```diff
@@ -388,549 +135,146 @@ LONG_PTR WINAPI SetWindowLongPtrW_Hook(HWND hWnd, int nIndex, LONG_PTR dwNewLong
...
-        g_hSubclassedMagnifierWnd = NULL;
-    }
-
-    // Thread-safe cleanup
-    HWND hHostWnd = NULL;
-    {
-        AutoCriticalSection lock(&g_csGlobalState);
-        hHostWnd = g_hHostWnd;
+    if (g_hHostWnd) {
+        DestroyWindow(g_hHostWnd);
```
m417z: That's incorrect. From the documentation:
> A thread cannot use DestroyWindow to destroy a window created by a different thread.

Also, you fail to undo other things, such as the state of the hidden windows, or undoing the `RegisterClassW` call. If restoring the state is not supported, perhaps it's better to just exit the process.

bcrtvkcs:
fix: address Windhawk developer review issues (m417z)
- Move EnumWindows(EnumWindowsProc_HideMagnifier) from Wh_ModInit to
  a new Wh_ModAfterInit callback. ShowWindow_Original is null during
  Wh_ModInit because hooks aren't installed until after it returns;
  Wh_ModAfterInit is called once hooks are active and trampolines valid.

- Replace DestroyWindow in Wh_ModUninit with TerminateProcess. Per
  Windows docs, DestroyWindow cannot be called cross-thread. Also,
  restoring all state (hidden windows, RegisterClassW, reparented
  children) is not feasible, so terminating the process lets magnify.exe
  restart cleanly without the mod.

PR #3337, opened 2026-02-18
Update taskbar-dock-animation.wh.cpp

Ph0en1x-dev:
Added an option to disable animation for system taskbar buttons (Start, Widgets, Task View, Search).

Added multilingual support.

Added a smoothing (lerp) setting for the animation (not to be confused with the animation curve types).

Added dirty-checking to prevent expensive repeated scans of the Windows UI element tree.

### DIFF
--- a/mods/taskbar-dock-animation.wh.cpp
+++ b/mods/taskbar-dock-animation.wh.cpp
@@ -1465,14 +1465,23 @@ void Wh_ModBeforeUninit() {
 void Wh_ModSettingsChanged() {
     Wh_Log(L"DockAnimation: Settings changed.");
     LoadSettings();
-    try {
-        for (auto& pair : g_contexts) {
-            auto& ctx = pair.second;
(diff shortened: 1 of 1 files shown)


mods/taskbar-dock-animation.wh.cpp
```diff
@@ -807,11 +1468,11 @@ void Wh_ModSettingsChanged() {
     try {
         for (auto& pair : g_contexts) {
             auto& ctx = pair.second;
+            ResetAllIconScales(ctx.icons);
```
m417z: Does it work? I believe it should run from the UI thread, like in `Wh_ModBeforeUninit`.

Ph0en1x-dev: Good catch — ResetAllIconScales in Wh_ModSettingsChanged currently runs outside the UI thread. I'll move it to RunFromWindowThread like in Wh_ModBeforeUninit to ensure thread safety.

PR #3354, opened 2026-02-20
Create win10-taskbar-per-app-volume-control.wh.cpp

Anixx:
Adjust sound volume per app by scrolling over Win10 taskbar

### DIFF
--- a/mods/win10-taskbar-per-app-volume-control.wh.cpp
+++ b/mods/win10-taskbar-per-app-volume-control.wh.cpp
@@ -1,0 +1,393 @@
+// ==WindhawkMod==
+// @id              win10-taskbar-per-app-volume-control
+// @name            Taskbar Volume Control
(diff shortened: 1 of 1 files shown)


mods/win10-taskbar-per-app-volume-control.wh.cpp
```diff
@@ -0,0 +1,393 @@
+// ==WindhawkMod==
+// @id              win10-taskbar-per-app-volume-control
+// @name            Taskbar Volume Control
```
m417z, line 3: I only noticed after merging, but there's already a mod with that name. Please change the name to make clear it's a Win10-only per-app volume control mod.

Also, it'd be nice to mention [Taskbar Volume Control Per-App](https://windhawk.net/mods/taskbar-volume-control-per-app) for the Win11 taskbar.

PR #3368, opened 2026-02-22
change name

Anixx:


mods/win10-taskbar-per-app-volume-control.wh.cpp
```diff
@@ -1,8 +1,8 @@
 // ==WindhawkMod==
 // @id              win10-taskbar-per-app-volume-control
-// @name            Taskbar Volume Control
+// @name            Win10 Taskbar Volume Control
```
m417z: Maybe:
```suggestion
// @name            Win10 Taskbar Volume Control Per-App
```

Otherwise it can be confused with [Taskbar Volume Control](https://windhawk.net/mods/taskbar-volume-control) which is compatible with the Win10 taskbar.

Anixx: Ah, done.

PR #3378, opened 2026-02-24
New Mod: Notepad Multi-Select"

hamomagdy724:
Implement multi-caret support in Notepad with Alt+Shift functionality for typing, backspacing, and rendering. Includes thread-safe subclassing for window injection.

### DIFF
--- a/mods/notepad-multi-select.wh.cpp
+++ b/mods/notepad-multi-select.wh.cpp
@@ -1,6 +1,6 @@
 // ==WindhawkMod==
 // @id              notepad-multi-select
-// @name            Notepad++ feature
+// @name            Notepad Multi-Select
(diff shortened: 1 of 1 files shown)


mods/notepad-multi-select.wh.cpp
```diff
@@ -0,0 +1,1038 @@
+// ==WindhawkMod==
+// @id              notepad-multi-select
+// @name            Notepad++ feature
```
m417z: This name makes it seem like a mod for Notepad++. I think the previous name (Notepad Multi-Select) was better.

hamomagdy724: I edited @name to Notepad Multi-select

PR #3379, opened 2026-02-24
Create Fix Windows Tools folder

Anixx:


mods/fix-windows-tools-view.wh.cpp
```diff
@@ -0,0 +1,259 @@
...
+                       (void*)RegQueryValueExWHook, (void**)&pOriginalRegQueryValueExW);
+    Wh_SetFunctionHook((void*)CreateWindowExW,
+                       (void*)CreateWindowExWHook, (void**)&pOriginalCreateWindowExW);
+    return TRUE;
+}
+
+void Wh_ModUninit(void) {
+    for (HWND hwnd : g_subclassedDefViews) {
+        if (IsWindow(hwnd))
+            RemoveWindowSubclass(hwnd, DefViewSubclassProc, 0);
```
m417z: From the documentation:
> You cannot use the subclassing helper functions to subclass a window across threads.

You can use `SendMessage` instead with a custom message, which will call `RemoveWindowSubclass` when handled. You can use `RegisterWindowMessage` for a custom message.

Anixx: Fixed.

mods/fix-windows-tools-view.wh.cpp
```diff
@@ -0,0 +1,276 @@
...
+LRESULT CALLBACK DefViewSubclassProc(HWND hWnd, UINT uMsg, WPARAM wParam,
+                                      LPARAM lParam, UINT_PTR uIdSubclass, DWORD_PTR dwRefData) {
+    HWND hwndCabinet = GetAncestor(hWnd, GA_ROOT);
+
+    if (uMsg == WM_REMOVE_SUBCLASS && WM_REMOVE_SUBCLASS != 0) {
+        auto it = g_windows.find(hwndCabinet);
+        if (it != g_windows.end() && it->second.pFolderView && it->second.bagNumber > 0) {
+            SaveViewMode(it->second.pFolderView, it->second.bagNumber);
+            it->second.pFolderView->Release();
+            it->second.pFolderView = nullptr;
```
m417z: Here and in `WM_DESTROY` (which seems to be the same code BTW), I'd expect the entry to be deleted from `g_windows` too.

Anixx: It it OK now?

PR #3381, opened 2026-02-24
Update taskbar-fade to v2.1

lukvbp:
Cleaned the code a bit
Added the ability to hide desktop icons when idling
Added the ability to disable the fade while something is focused


kaanaldemir:
This one is great and one step close to one solution for all for me. Would you consider fading the cursor like the desktop icons too? With 2.1 my OLED monitor turns to pitch black at idle only if I move the cursor to the edge. 

I use black wallpaper but another option for making background black at idle might be useful for some too?

I don't know if these are possible to do but I wanted to share some ideas that come to my mind.

Anyway I wanted to say thanks because I cant get used to taskbar hiding so this is the perfect alternative for me.

PR #3382, opened 2026-02-24
Create Windows 10 Taskbar Context Menu Fix for Win11 24H2+

Anixx:
Create Windows 10 Taskbar Context Menu Fix for Win11 24H2+


Anixx:
But would not architecture x86-64 tag exclude ARM?

PR #3384, opened 2026-02-24
Logon & Sleep Fade Restorer 1.2

Ingan121:
* Add support for fading on sleep, hibernation, and monitor off
* Add support for setting different duration values on logon/unlock and logoff/shutdown, and optionally keeping the default DWM fade on logon/unlock
* Add support for using the kernel-mode fade implementation instead of the custom logic in this mod (on supported versions)

### DIFF
--- a/mods/win7-login-fade.wh.cpp
+++ b/mods/win7-login-fade.wh.cpp
@@ -465,11 +465,8 @@ HWND CreateOverlayWindow() {
     wc.lpszClassName = OVERLAY_WIN_CLASS;
     wc.hbrBackground = (HBRUSH)GetStockObject(BLACK_BRUSH);
     if (!RegisterClassW(&wc)) {
-        int gle = GetLastError();
-        if (gle != ERROR_CLASS_ALREADY_EXISTS) {
-            Wh_Log(L"RegisterClassW failed, GLE=%d", gle);
-            return NULL;
-        }
+        Wh_Log(L"RegisterClassW failed, GLE=%d", GetLastError());
+        return NULL;
     }
     return CreateWindowExW(WS_EX_TOPMOST | WS_EX_TOOLWINDOW, OVERLAY_WIN_CLASS, L"", WS_POPUP, 0, 0, 0, 0, NULL, NULL, wc.hInstance, NULL);
 }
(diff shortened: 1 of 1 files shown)


mods/win7-login-fade.wh.cpp
```diff
@@ -178,26 +858,141 @@ BOOL Wh_ModInit() {
...
 }
 
-void Wh_ModSettingsChanged() {
+BOOL Wh_ModSettingsChanged(BOOL* bReload) {
+    BOOL prevSleepFadeEnabled = g_settings.sleepFadeEnabled;
     LoadSettings();
+    if (!g_isWinlogon && prevSleepFadeEnabled != g_settings.sleepFadeEnabled) {
+        // The global injection is enabled or disabled, need to unload the mod
+        // (Reloading unloaded mod on mod setting change is automatically handled by Windhawk)
+        return FALSE;
```
m417z, line 993: This keeps the mod disabled. It probably works because having `g_settings.sleepFadeEnabled == true` here is unlikely as it wouldn't be loaded in mod init, but a better check would be:

```cpp
if (!g_isWinlogon && !g_settings.sleepFadeEnabled) {
   return FALSE;
}
```

mods/win7-login-fade.wh.cpp
```diff
@@ -93,76 +315,534 @@ int GetDeviceRefreshRate() {
...
+
+HWND CreateOverlayWindow() {
+    WNDCLASSW wc = { 0 };
+    wc.lpfnWndProc = OverlayWndProc;
+    wc.hInstance = GetModuleHandleW(NULL);
+    wc.lpszClassName = OVERLAY_WIN_CLASS;
+    wc.hbrBackground = (HBRUSH)GetStockObject(BLACK_BRUSH);
+    if (!RegisterClassW(&wc)) {
+        int gle = GetLastError();
+        if (gle != ERROR_CLASS_ALREADY_EXISTS) {
```
m417z: Can it ever be `ERROR_CLASS_ALREADY_EXISTS`? See also: https://github.com/ramensoftware/windhawk-mods/pull/3398#discussion_r2855807243.

PR #3390, opened 2026-02-25
Add Classic Tray Keyboard Layout Indicator

Anixx:
This file implements a tray indicator for keyboard layout, allowing users to switch layouts and view the current layout in the system tray.

### DIFF
--- a/mods/classic-tray-keyboard-layout-indicator.wh.cpp
+++ b/mods/classic-tray-keyboard-layout-indicator.wh.cpp
@@ -1,0 +1,404 @@
+// ==WindhawkMod==
+// @id              classic-tray-keyboard-layout-indicator
+// @name            Classic Tray Keyboard Layout Indicator
+// @description     Tray indicator for keyboard layout, similar to internat.exe from Windows 2000
+// @version         1.0.0
+// @author          Anixx
+// @github          https://github.com/Anixx
+// @include         explorer.exe
+// @compilerOptions -lgdi32 -luser32 -lshell32
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Keyboard Layout Tray Indicator
+
+![Screenshot](https://i.imgur.com/L4i2WhE.png)
+
+Displays current keyboard layout in the system tray as a two-letter code, 
+in the style of Windows 2000. This is an alternative to the default indicator.
+
+- **Left click** — cycle through installed layouts
+- **Right click** — pick layout from list
+*/
+// ==/WindhawkModReadme==
+
+#include <windows.h>
+#include <shellapi.h>
+#include <vector>
+#include <string>
+
+#define WM_TRAYICON     (WM_USER + 1)
+#define WM_MOD_QUIT     (WM_USER + 2)
+#define WM_REFRESH      (WM_USER + 3)
+
+#define TRAY_ICON_UID   7654
+#define ID_TIMER        1
+#define IDM_LAYOUT_BASE 1000
+
+#define MUTEX_NAME      L"Global\\KbdLayoutTray_WH_Mutex_7654"
+
+static HWND            g_hwnd       = nullptr;
+static HKL             g_currentHKL = nullptr;
+static HINSTANCE       g_hInstance  = nullptr;
+static HWINEVENTHOOK   g_eventHook  = nullptr;
+static HANDLE          g_thread     = nullptr;
+static HANDLE          g_mutex      = nullptr;
+static HICON           g_hIcon      = nullptr;
+static UINT            g_msgTaskbar = 0;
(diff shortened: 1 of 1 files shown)


mods/classic-tray-keyboard-layout-indicator.wh.cpp
```diff
@@ -0,0 +1,404 @@
+// ==WindhawkMod==
+// @id              classic-tray-keyboard-layout-indicator
+// @name            Classic Tray Keyboard Layout Indicator
+// @description     Tray indicator for keyboard layout, similar to internat.exe from Windows 2000
+// @version         1.0.0
+// @author          Anixx
+// @github          https://github.com/Anixx
+// @include         explorer.exe
```
m417z, line 8: Does it have to be part of explorer.exe? See:

Anixx: <stdin>:560:5: error: use of undeclared identifier 'WhTool_ModSettingsChanged'; did you mean 'Wh_ModSettingsChanged'?
  560 |     WhTool_ModSettingsChanged();
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~
      |     Wh_ModSettingsChanged
<stdin>:555:6: note: 'Wh_ModSettingsChanged' declared here
  555 | void Wh_ModSettingsChanged() {
      |      ^
1 error generated.

m417z: Define an empty `WhTool_ModSettingsChanged`:

```cpp
void WhTool_ModSettingsChanged() {
  // No action needed...
}
```

Anixx: This mod already has mutex. It would need refactoring. Also, the Windhawk-injected version for now looks less stable.

m417z: There are other reasons besides multiple instances, for example:
* explorer.exe stability
* The target explorer.exe that acquires the mutex can be short-lived, and once it exits, the mod will stop working

What refactoring is needed besides adding the code and renaming/adding callbacks?

Why does it look less stable? What does it mean?

Anixx: > The target explorer.exe that acquires the mutex can be short-lived, and once it exits, the mod will stop working

This has already been dealt with. 

> What refactoring is needed besides adding the code and renaming/adding callbacks?

Removing the mutex, for instance.

> Why does it look less stable? What does it mean?

Testing it.

Also, it takes 2MB as a separate process...

Anixx: What about having two mods, one as separate process and one as injecting into Explorer? Some people cannot use many third-party tools on workplace, so having everything hidden in Explorer is a good for them.

m417z: > This has already been dealt with.

To some extent - if I understand correctly, the previous process will take the mod instance with it, and the next instance will start, taking its place. Can be jumpy but it works.

Regarding the waiting, disabling the mod will likely cause crashes because some instances will be stuck waiting for the mutex.

Overall, it just doesn't feel like the right design.

> What about having two mods, one as separate process and one as injecting into Explorer?

That will be even more confusing for users. If you insist, we can merge it as is, I just think that it's a less optimal, less robust design.

> Some people cannot use many third-party tools on workplace

You're already using Windhawk, so I don't see how it's an issue.

Anixx: Well, it seems, neither approach will work as needed, because the mod needs elevated privilegies to catch layout switching in elevated apps.

m417z: I wonder if a non-elevated UIAccess process will be able to do that too. You can try this:
* Use the tool mod snippet
* Replace target process with `osk.exe`
* Enable the mod
* Run `osk.exe`
* Another `osk.exe` should be spawned to host the mod, see if the mod works

It's not a solution yet since you need to run `osk.exe` manually, but let's see if it even works.

Anixx: Nope. Still not enough provilegies.

Anixx: Could we have the same approach, but allowing to spawn elevated interactive Windhawk processes?

m417z: You can spawn an elevated child process from within the Windhawk service process, but:
* You'll have to spawn it in the right session, perhaps for every user, since the service runs in an isolated session (session 0)
* It won't work for the portable Windhawk version

PR #3398, opened 2026-02-25
Add Windows 11 Start Menu Power Buttons mod

ahzvenol:
This mod adds customizable one-click Shutdown, Restart, Sign out, Sleep, and Hibernate buttons to the Windows 11 Start menu, replacing the default power flyout.

Users can customize which buttons to display and their order via the mod settings.

### DIFF
--- a/mods/win11-power-buttons.wh.cpp
+++ b/mods/win11-power-buttons.wh.cpp
@@ -46,7 +46,7 @@ Adds customizable one-click Shutdown, Restart, Sign out, Sleep, and Hibernate bu
 
 - confirm_before_action: false
   $name: Confirm before power actions
-  $name:zh-CN: 执行电源操作前确认
+  $name:zh-CN: 执行前确认
   $description: Show a confirmation dialog before executing power actions.
   $description:zh-CN: 在执行电源操作前显示确认对话框。
 


mods/win11-power-buttons.wh.cpp
```diff
@@ -0,0 +1,847 @@
...
+// @architecture    x86-64
+// @compilerOptions -lcomctl32 -lole32 -loleaut32 -lruntimeobject -Wl,--export-all-symbols -luser32 -lwtsapi32 -lpowrprof
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Windows 11 Start Menu Power Buttons
+
+Adds customizable one-click Shutdown, Restart, Sign out, Sleep, and Hibernate buttons to the Windows 11 Start menu, replacing the default power flyout.
+
```
m417z, line 22: How about adding a screenshot?

mods/win11-power-buttons.wh.cpp
```diff
@@ -0,0 +1,847 @@
...
+// Thread function for the hidden proxy window
+static DWORD WINAPI ProxyWindowThread(LPVOID) {
+    InitProxyMessage();
+
+    WNDCLASSEXW wcex = { sizeof(WNDCLASSEX) };
+    wcex.lpfnWndProc = ProxyWndProc;
+    wcex.hInstance = GetModuleHandle(NULL);
+    wcex.lpszClassName = PROXY_WINDOW_CLASS;
+
+    if (!RegisterClassExW(&wcex)) {
```
m417z, line 368: The class is never unregistered. This may cause crashes if the mod is loaded later at a different address, because the old `ProxyWndProc` value will no longer be valid.

Also, for that reason, it's better not to proceed in case of `ERROR_CLASS_ALREADY_EXISTS`, which shouldn't happen if the class is unregistered on unload.

mods/win11-power-buttons.wh.cpp
```diff
@@ -0,0 +1,847 @@
...
+    }, hCoreWnd);
+}
+
+void Wh_ModUninit() {
+    g_unloading = true;
+
+    if (g_proxyWindow) {
+        PostMessage(g_proxyWindow, WM_QUIT, 0, 0);
+        g_proxyWindow = NULL;
+    }
```
m417z, line 910: Wait for the thread to exit, otherwise the mod might be unloaded while the thread is still running.

mods/win11-power-buttons.wh.cpp
```diff
@@ -0,0 +1,847 @@
...
+
+// Attempts to initialize UWP hooks. Retries if the environment isn't ready.
+static void InitWithRetry(HWND hwnd) {
+    if (!g_initialized.exchange(true)) {
+        LoadSettings();
+        BuildButtons();
+    }
+
+    if (!StartVisibilityMonitoring()) {
+        SetTimer(hwnd, 101, 100, RetryTimerProc);
```
m417z: I'd use a more unique timer id. I like to just grab [the current timestamp](https://www.epochconverter.com/): 1772058423.

m417z (commented):
Very nice, well done! How about adding an option to ask for confirmation? Otherwise it's easy to click on these buttons accidentally.

mods/win11-power-buttons.wh.cpp
```diff
@@ -0,0 +1,857 @@
...
+    // Bypass UIPI to allow messages from lower-integrity processes
+    ChangeWindowMessageFilterEx(g_proxyWindow, g_proxyMessage, MSGFLT_ALLOW, NULL);
+
+    MSG msg;
+    while (GetMessage(&msg, NULL, 0, 0)) {
+        TranslateMessage(&msg);
+        DispatchMessage(&msg);
+    }
+
+    DestroyWindow(g_proxyWindow);
```
m417z: I think the window is already destroyed at this point. Does `DestroyWindow`/`IsWindow` return TRUE here?

ahzvenol: Yes, it does return TRUE.

PR #3403, opened 2026-02-26
Add mod to fix taskbar keyboard layout indicator

Anixx:
This mod fixes the keyboard layout indicator text orientation when using the classic Windows taskbar from Windows 10 running under Windows 11 24H2.

### DIFF
--- a/mods/fix-legacy-taskbar-tray-input-indicator.wh.cpp
+++ b/mods/fix-legacy-taskbar-tray-input-indicator.wh.cpp
@@ -5,6 +5,7 @@
 // @version         1.0.0
 // @author          Anixx
 // @github          https://github.com/Anixx
+// @architecture    x86-64
 // @include         explorer.exe
 // @compilerOptions -lgdi32 -luser32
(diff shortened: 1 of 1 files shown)


mods/fix-legacy-taskbar-tray-input-indicator.wh.cpp
```diff
@@ -0,0 +1,120 @@
+// ==WindhawkMod==
+// @id              fix-legacy-taskbar-tray-input-indicator
+// @name            Fix language indicator in Win10 taskbar under Win11 24H2+
+// @description     Fixes text orientation in the keyboard layout indicator in Win10 taskbar running under Win11 24H2+
+// @version         1.0.0
+// @author          Anixx
+// @github          https://github.com/Anixx
+// @include         explorer.exe
```
m417z, line 9: This mod can be limited to x86-64.

Anixx: Done!

PR #3405, opened 2026-02-26
Update vlc-discord-rpc to 1.1.2: Bug fix & Improvement

ciizerr:
**What's New:**

_Fixes & Performance Improvements_

* High CPU Usage Fix: Resolved a performance issue introduced in v1.1.1 where the metadata cleaner caused high CPU load. Implemented a state-change cache so the heavy metadata scrubber only runs when the source media actually changes.

_Advanced Filtering & Processing_

* Dynamic Filter Syncing: The metadata cleaner now remotely fetches and locally caches community-maintained junk word and tag filters. This ensures the cleaner stays highly accurate and up-to-date over time without requiring mod updates.

_Enhanced Customization_

* Modular Mod Settings: Added dedicated toggles in the Windhawk settings to individually enable or disable Quality Tags (`ShowQualityTags`) and the Metadata Cleaner (`clean media titles`) based on user preference.

### DIFF
--- a/mods/vlc-discord-rpc.wh.cpp
+++ b/mods/vlc-discord-rpc.wh.cpp
@@ -720,19 +720,17 @@ int DetectActivityType(const std::string& filename, const std::string& quality) 
 // 4. REMOTE FILTER UPDATER & CACHING
 // =============================================================
 
-std::string GetCacheFilePath() {
-    char path[MAX_PATH];
-    if (SUCCEEDED(SHGetFolderPathA(NULL, CSIDL_APPDATA, NULL, 0, path))) {
-        std::string dir = std::string(path) + "\\VLC_RPC";
-        CreateDirectoryA(dir.c_str(), NULL);
-        return dir + "\\junklist.txt";
+std::wstring GetCacheFilePathW() {
+    WCHAR storagePath[MAX_PATH];
+    if (Wh_GetModStoragePath(storagePath, ARRAYSIZE(storagePath))) {
+        return std::wstring(storagePath) + L"\\junklist.txt";
     }
-    return "";
+    return L"";
 }
 
-bool IsCacheValid(const std::string& path) {
-    struct stat result;
-    if (stat(path.c_str(), &result) == 0) {
+bool IsCacheValid(const std::wstring& path) {
+    struct _stat result;
+    if (_wstat(path.c_str(), &result) == 0) {
         time_t now = time(nullptr);
         // 21600 seconds = 6 hours
         if (difftime(now, result.st_mtime) < 21600) {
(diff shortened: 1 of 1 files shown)


mods/vlc-discord-rpc.wh.cpp
```diff
@@ -665,17 +712,163 @@ int DetectActivityType(const std::string& filename, const std::string& quality)
...
+    
+    // 1. Check if we have a valid cache that is < 24 hours old
+    if (!cachePath.empty() && IsCacheValid(cachePath)) {
+        LoadFiltersFromFile(cachePath);
+        if (g_filtersLoaded) return; // Success!
+    }
+
+    // 2. Fetch from GitHub if no cache or cache expired
+    std::wstring host = L"raw.githubusercontent.com";
+    std::wstring path = L"/ciizerr/vlc-discord-rpc-archive/main/assets/filters.txt";
```
m417z: Include it in the mod, which makes the code reviewable without external dependencies.

ciizerr: Thanks for the review! I understand the concern about external dependencies.

The reason I fetch filters.txt remotely is because release tags, piracy labels, and junk filename keywords change all the time. Keeping the list on GitHub lets it stay updated without users needing a full mod update just for small dictionary changes.

That said, I completely understand the need for transparency. I can:

- Expand the hardcoded fallback list inside file so it works well even without remote fetching.
- Clearly link the public filters.txt file in the modREADME section so it's fully visible.
- Add a setting (like “Disable Remote Filters”) so users can choose to use only the local list.

If that works, I’ll implement it in the next commit.
let me know.

mods/vlc-discord-rpc.wh.cpp
```diff
@@ -665,17 +717,168 @@ int DetectActivityType(const std::string& filename, const std::string& quality)
...
+// 4. REMOTE FILTER UPDATER & CACHING
+// =============================================================
+
+std::string GetCacheFilePath() {
+    char path[MAX_PATH];
+    if (SUCCEEDED(SHGetFolderPathA(NULL, CSIDL_APPDATA, NULL, 0, path))) {
+        std::string dir = std::string(path) + "\\VLC_RPC";
+        CreateDirectoryA(dir.c_str(), NULL);
+        return dir + "\\junklist.txt";
+    }
```
m417z, line 727: Use `Wh_GetModStoragePath` for storing mod files.

mods/vlc-discord-rpc.wh.cpp
m417z: I see that you did these steps. Also, please use `Wh_GetUrlContent`, and have a simple, isolated function which doesn't touch global state to receive the text from the file (online or local) and returns a struct with these four vectors: newSites, newTlds, newTags, newWords. This will make it easier for me and others to review.

ciizerr:
All review feedback applied! The mod now uses `Wh_GetModStoragePath` to sandbox the cache, `Wh_GetUrlContent` to vastly simplify the download logic, and all text parsing has been cleanly isolated into a pure 

`ParseFilters`
 function returning a struct before touching global state. Strict Local Mode is also now enabled by default so the mod operates entirely offline out-of-the-box."

PR #3407, opened 2026-02-26
Add Taskbar Split by Monitor mod

lighfu:
## Summary

- Adds a new mod that forces each monitor's taskbar to only display windows located on that specific monitor
- Hooks `RegQueryValueExW` and `RegGetValueW` to enforce multi-monitor taskbar settings (`MMTaskbarMode`, `MMTaskbarEnabled`)
- Supports two modes: per-monitor only (default) and main + per-monitor
- Compatible with Windows 10, Windows 11, and Windows 11 24H2 (ShellHost.exe)

## Test plan

- [x] Tested on Windows 11 Home with dual monitors
- [x] Verified each taskbar shows only its monitor's windows
- [x] Verified settings change (filter mode) applies without explorer restart
- [x] Verified original settings restore on mod disable

### DIFF
--- a/mods/taskbar-split-by-monitor.wh.cpp
+++ b/mods/taskbar-split-by-monitor.wh.cpp
@@ -6,7 +6,7 @@
 // @description:ja  各モニターのタスクバーにそのモニター上のウィンドウのみを表示します
 // @version         1.1.0
 // @author          ajisaiflow
-// @github          lighfu
+// @github          https://github.com/lighfu
 // @include         explorer.exe
 // @include         ShellHost.exe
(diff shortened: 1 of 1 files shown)


m417z:
How is the mod different from changing the Windows taskbar settings? And why does it both update the registry, and hook it?

lighfu:
Thanks for the review. I've been managing all my Windows display settings through Windhawk, so I overlooked that this is already available as a built-in Windows setting. That said, there might be others like me who miss it for the same reason — but I understand this doesn't justify a dedicated mod. Closing this PR.

PR #3412, opened 2026-02-26
Volume Mixer Minimize Button

Jevil7452:
Adds a Minimize button to the Volume Mixer

### DIFF
--- a/mods/sndvol-minimize.wh.cpp
+++ b/mods/sndvol-minimize.wh.cpp
@@ -1,0 +1,41 @@
+// ==WindhawkMod==
+// @id              sndvol-minimize
+// @name            Volume Mixer Minimize Button
+// @description     Adds a Minimize button to the Volume Mixer
+// @version         1
+// @author          Jevil7452
+// @github          https://github.com/Jevil7452
+// @include         sndvol.exe
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+This mod adds the Minimize button to the Volume Mixer and Volume Control Options dialogs.
+
+Before:
+
+![1](https://i.imgur.com/MG6eeLL.png)
+![2](https://i.imgur.com/2ZCGB2t.png)
+
+After:
+
+![3](https://i.imgur.com/1pNiTNi.png)
+![4](https://i.imgur.com/cu6j0JR.png)
+*/
+// ==/WindhawkModReadme==
+
+#include <windows.h>
+
+LRESULT CALLBACK CallWndProc(int nCode, WPARAM wParam, LPARAM lParam) {
+    CWPSTRUCT* pwp = (CWPSTRUCT*)lParam;
+    if (pwp->message == WM_CREATE) {
+        LONG_PTR style = GetWindowLongPtrW(pwp->hwnd, GWL_STYLE);
+        SetWindowLongPtrW(pwp->hwnd, GWL_STYLE, style | WS_MINIMIZEBOX);
+    }
+    return CallNextHookEx(NULL, nCode, wParam, lParam);
+}
+
+BOOL Wh_ModInit(void) {
+    SetWindowsHookExW(WH_CALLWNDPROC, CallWndProc, NULL, GetCurrentThreadId());
+    return TRUE;
+}
--- a/mods/sndvol-minimize.wh.cpp
+++ b/mods/sndvol-minimize.wh.cpp
@@ -25,19 +25,80 @@ After:
 // ==/WindhawkModReadme==
 
 #include <windows.h>
+#include <tlhelp32.h>
+#include <vector>
+
+std::vector<HHOOK> g_hooks;
 
 LRESULT CALLBACK CallWndProc(int nCode, WPARAM wParam, LPARAM lParam) {
-    if (nCode < 0)
+    if (nCode < 0) 
         return CallNextHookEx(NULL, nCode, wParam, lParam);
+
     CWPSTRUCT* pwp = (CWPSTRUCT*)lParam;
     if (pwp->message == WM_CREATE) {
         LONG_PTR style = GetWindowLongPtrW(pwp->hwnd, GWL_STYLE);
-        SetWindowLongPtrW(pwp->hwnd, GWL_STYLE, style | WS_MINIMIZEBOX);
+        if (!(style & WS_CHILD)) {
+            SetWindowLongPtrW(pwp->hwnd, GWL_STYLE, style | WS_MINIMIZEBOX);
+        }
     }
+
     return CallNextHookEx(NULL, nCode, wParam, lParam);
 }
 
-BOOL Wh_ModInit(void) {
-    SetWindowsHookExW(WH_CALLWNDPROC, CallWndProc, NULL, GetCurrentThreadId());
+BOOL CALLBACK EnumWindowsProc(HWND hWnd, LPARAM lParam) {
+    DWORD pid = 0;
+    GetWindowThreadProcessId(hWnd, &pid);
+    if (pid != GetCurrentProcessId()) {
+        return TRUE;
+    }
+
+    LONG_PTR style = GetWindowLongPtrW(hWnd, GWL_STYLE);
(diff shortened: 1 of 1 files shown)


mods/sndvol-minimize.wh.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+![3](https://i.imgur.com/1pNiTNi.png)
+![4](https://i.imgur.com/cu6j0JR.png)
+*/
+// ==/WindhawkModReadme==
+
+#include <windows.h>
+
+LRESULT CALLBACK CallWndProc(int nCode, WPARAM wParam, LPARAM lParam) {
+    CWPSTRUCT* pwp = (CWPSTRUCT*)lParam;
```
m417z, line 37: Handle `nCode < 0` per documentation:
```suggestion
    if (nCode < 0)
        return CallNextHookEx(NULL, nCode, wParam, lParam);
    CWPSTRUCT* pwp = (CWPSTRUCT*)lParam;
```

m417z:
### Submission review

I'd prefer to have all points addressed, but at the very least, addressing 2 is easy, and addressing 3 is preferable both to support mid-session injection and improve stability - currently the mod assumes that `Wh_ModInit` runs on the main thread, but it's not always the case.

---

**1. The hook is never removed — this crashes the process on disable/unload.** `Wh_ModInit` calls `SetWindowsHookExW` but discards the returned `HHOOK`, and there is no `Wh_ModUninit`. Windhawk only removes its *own* function/symbol hooks automatically — a `SetWindowsHookEx` hook is a raw Win32 hook that the mod must remove itself. When the mod is disabled, updated, or reloaded, the DLL is unmapped but the `WH_CALLWNDPROC` hook stays installed pointing into the now-freed mod code; the next window message dispatched on that thread jumps into unmapped memory → crash (or worse). Store the handle in a global and unhook it on teardown:

```cpp
HHOOK g_hook;

BOOL Wh_ModInit(void) {
    g_hook = SetWindowsHookExW(WH_CALLWNDPROC, CallWndProc, NULL,
                              GetCurrentThreadId());
    return g_hook != NULL;
}

void Wh_ModUninit(void) {
    if (g_hook) {
        UnhookWindowsHookEx(g_hook);
    }
}
```

For reference, [caps-ime-switcher](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/caps-ime-switcher.wh.cpp#L346) and [monitor-sleep-button](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/monitor-sleep-button.wh.cpp) both store the `HHOOK` and unhook it in `Wh_ModUninit`.

**2. `WS_MINIMIZEBOX` is being OR-ed onto *every* window, including child controls — where that bit is actually `WS_GROUP`.** The hook fires on `WM_CREATE` for every window created on the thread: the top-level dialogs *and* every child control inside them. But `WS_MINIMIZEBOX` (`0x00020000`) is the same bit as `WS_GROUP`, and for a child window the system interprets it as `WS_GROUP`. So the mod is stamping `WS_GROUP` onto every control, which changes keyboard group navigation (arrow-key cycling, radio-button groups) inside the dialogs. Filter to top-level windows before touching the style:

```cpp
LONG_PTR style = GetWindowLongPtrW(pwp->hwnd, GWL_STYLE);
if (!(style & WS_CHILD)) {
    SetWindowLongPtrW(pwp->hwnd, GWL_STYLE, style | WS_MINIMIZEBOX);
}
```

This also naturally scopes the change to the windows you actually care about.

**3. Doesn't work when enabled while the Volume Mixer is already open, `Wh_ModInit` thread is assumed.** The approach (thread-scoped hook + `WM_CREATE`) only affects windows created *after* init, and only on the init thread. Two consequences:
- If sndvol.exe is already running when the mod is enabled, `WM_CREATE` for the existing window has already fired, so nothing happens to it.
- Worse, when Windhawk injects into an already-running sndvol.exe, `Wh_ModInit` runs on a transient injection thread, not the UI thread — so the thread-local hook lands on the wrong (and short-lived) thread and has no effect at all until sndvol.exe is relaunched.

Don't run `SetWindowsHookExW` in `Wh_ModInit`. Instead, hook `CreateWindowExW`, and also walk the existing top-level windows at `Wh_ModAfterInit` and apply the style, e.g. `EnumWindows` filtered by `GetCurrentProcessId()` (or `EnumThreadWindows`), then OR in `WS_MINIMIZEBOX` there too. It works today only because sndvol.exe is normally launched fresh with the mod already enabled.

<p>

- Adding `WS_MINIMIZEBOX` requires `WS_SYSMENU` to be present for the button to show. The sndvol dialogs have it, so this is fine as-is, but if you later broaden the target set it's worth OR-ing `WS_SYSMENU` too so the minimize box is guaranteed visible.

</p>
</details>

m417z:
### Submission review

---

**Replace the `WH_CALLWNDPROC` + thread-snapshot machinery with a single `CreateWindowExW` function hook.** The current design enumerates every thread with `CreateToolhelp32Snapshot` and installs a `WH_CALLWNDPROC` hook on each one just to catch `WM_CREATE`. A single function hook on `CreateWindowExW` that ORs `WS_MINIMIZEBOX` into `dwStyle` before calling the original does the same job and is strictly better:

- It catches windows created on **any** thread, including threads created after `Wh_ModInit` — the current snapshot only hooks threads that already exist at init, so a window created later on a new thread is missed.
- It applies the style **before the frame is first painted**, so the button shows up with no redraw workaround needed (see next point).
- It drops `HookAllProcessThreads`, `g_hooks`, and the `Wh_ModUninit` unhook loop entirely — no per-thread bookkeeping.

This is the established pattern for exactly this kind of change; see [`notepad-clientedge.wh.cpp`](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/notepad-clientedge.wh.cpp#L28-L73), which hooks `CreateWindowExW` and ORs a style flag into a creation call. Prefer `WindhawkUtils::SetFunctionHook()` for the type-safe form. You can still filter to `!(dwStyle & WS_CHILD)` (top-level) as you do today.

**The already-open case needs `SWP_FRAMECHANGED`.** Both `EnumWindowsProc` and the `WM_CREATE` handler change `GWL_STYLE` with a bare `SetWindowLongPtrW`. For a window whose frame is already drawn — the exact scenario the last commit targets ("minimize button appears when it's enabled while the dialog is already open") — the frame is cached, and per [MSDN](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setwindowlongptrw) a frame-style change "will not take effect until you call the `SetWindowPos` function ... with the `SWP_FRAMECHANGED` flag." So the button may not actually appear until something else forces a non-client recalc. After changing the style on an already-visible window, force the recalc:

```cpp
SetWindowLongPtrW(hWnd, GWL_STYLE, style | WS_MINIMIZEBOX);
SetWindowPos(hWnd, nullptr, 0, 0, 0, 0,
             SWP_NOMOVE | SWP_NOSIZE | SWP_NOZORDER | SWP_NOACTIVATE | SWP_FRAMECHANGED);
```

If you move new-window handling to the `CreateWindowExW` hook (recommended above), this only matters for the `Wh_ModAfterInit`/`EnumWindows` pass that fixes windows already open when the mod is enabled — keep that pass and add the `SWP_FRAMECHANGED` there.

<p>

Minor polish — none of this affects users in the common case, so it's your call.

- **Reversibility on disable.** When the mod is disabled, any window that's currently open keeps the `WS_MINIMIZEBOX` you added — the change isn't undone. It's minor (the windows are transient and disappear when sndvol closes), but strictly speaking a mod's effects should revert on disable. You could restore the original styles in `Wh_ModUninit` (re-`EnumWindows` and clear the bit) if you want it fully clean.

- **Narrow to the actual dialog(s).** You currently add the button to every top-level window in the process. That's harmless here because the flag is a no-op on caption-less popups/tooltips and sndvol is tiny, but if you want to be precise you could check the window class (e.g. only the mixer / Volume Control Options dialogs) rather than all non-child windows.

</p>
</details>

PR #3433, opened 2026-02-28
Add inclusion/exclusion list for "Use window icons"

### LINKED ISSUE #3432: [Disable grouping on the taskbar] Add an inclusion/exclusion list for the "Use window icons" option
I would like to request a new option that allows partially enabling/disabling the "Use window icons" feature for specific programs.

For my use case, I want to disable this feature for Explorer so I can quickly locate it on the taskbar, but I also want to enable it for chat applications so I can quickly identify different chat windows.

@m417z What do you think about this? I have already tried implementing this feature locally. I will submit a PR shortly, and if you are willing, you can check and merge it.

ahzvenol:
This PR adds an option to partially enable/disable "Use window icons" for specific programs.

- When "Use window icons" is enabled, listed programs are excluded
- When "Use window icons" is disabled, only listed programs use window icons

Closes #3432

### DIFF
--- a/mods/taskbar-grouping.wh.cpp
+++ b/mods/taskbar-grouping.wh.cpp
@@ -2,7 +2,7 @@
 // @id              taskbar-grouping
 // @name            Disable grouping on the taskbar
 // @description     Causes a separate button to be created on the taskbar for each new window
-// @version         1.3.10
+// @version         1.3.11
 // @author          m417z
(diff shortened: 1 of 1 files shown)


m417z:
Thanks for the PR. 
Development takes place here:

Please post the pull request in that repository.

PR #3461, opened 2026-03-03
Add Better Dialogs mod

FireBlade211:
# Better Dialogs
This mod improves certain Windows dialogs. It also re-translates messages for those dialogs to make sure that apps that expect the original dialogs still work properly.

More dialogs coming soon! (or maybe not so soon, just in the future)

### DIFF
--- a/mods/better-dialogs.wh.cpp
+++ b/mods/better-dialogs.wh.cpp
@@ -272,7 +272,7 @@ int WINAPI MessageBoxW_Hook(HWND hWnd, LPCWSTR lpText, LPCWSTR lpCaption, UINT u
 
         PCWSTR caption = lpCaption != NULL ? lpCaption : L"Error";
 
-        HMODULE hComCtl = LoadLibraryW(L"comctl32.dll");
+        HMODULE hComCtl = LoadLibraryExW(L"comctl32.dll", nullptr, LOAD_LIBRARY_SEARCH_SYSTEM32);
         if (hComCtl) {
             // TaskDialog_t pTaskDialog = (TaskDialog_t)GetProcAddress(hComCtl, "TaskDialog");
 
@@ -519,7 +519,7 @@ int WINAPI MessageBoxIndirectW_Hook(const MSGBOXPARAMSW *lpmbp) {
 
         PCWSTR caption = lpmbp->lpszCaption != NULL ? lpmbp->lpszCaption : L"Error";
 
-        HMODULE hComCtl = LoadLibraryW(L"comctl32.dll");
+        HMODULE hComCtl = LoadLibraryExW(L"comctl32.dll", nullptr, LOAD_LIBRARY_SEARCH_SYSTEM32);
         if (hComCtl) {
             // TaskDialog_t pTaskDialog = (TaskDialog_t)GetProcAddress(hComCtl, "TaskDialog");
 


mods/better-dialogs.wh.cpp
```diff
@@ -0,0 +1,981 @@
...
+            }
+            break;
+        }
+
+        if (help)
+            buttons |= TDCBF_HELP;
+
+        PCWSTR caption = lpCaption != NULL ? lpCaption : L"Error";
+
+        HMODULE hComCtl = LoadLibraryW(L"comctl32.dll");
```
m417z: Use `LoadLibraryExW(L"<name>.dll", nullptr, LOAD_LIBRARY_SEARCH_SYSTEM32)` here and in other places. See https://github.com/ramensoftware/windhawk-mods/issues/2063 for details.

mods/better-dialogs.wh.cpp
```diff
@@ -0,0 +1,981 @@
...
+// @author          FireBlade
+// @github          https://github.com/FireBlade211
+// @include         *
+// @compilerOptions -lcomdlg32 -luser32 -lshell32 -lole32 -luuid -lcomctl32
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Better Dialogs
+This mod improves certain Windows dialogs. It also re-translates messages for those dialogs to make sure that apps that expect the original dialogs still work properly.
```
m417z, line 15: It's difficult to understand what the mod does just from the name, description, and readme. One has to look at the settings to see which dialogs are customized. I suggest to update the mod description and readme to include the customized dialogs. Adding a screenshot can be a nice addition.

mods/better-dialogs.wh.cpp
```diff
@@ -0,0 +1,981 @@
...
+# Here you can define settings, in YAML format, that the mod users will be able
+# to configure. Metadata values such as $name and $description are optional.
+# Check out the documentation for more information:
+# https://github.com/ramensoftware/windhawk/wiki/Creating-a-new-mod#settings
+- messageTaskDlg: true
+  $name: Replace Message Boxes with Task Dialogs
+  $description: Replace Windows message boxes with task dialogs.
+- modernFolderBrowser: true
+  $name: Use modern folder picker dialog
+  $description: Replace the old folder picker dialog with the modern version.
```
m417z, line 37: This looks similar to https://windhawk.net/mods/modernize-folder-picker-dialog, have you seen this mod?

mods/better-dialogs.wh.cpp
```diff
@@ -0,0 +1,981 @@
...
+            int nButton = 0;
+            HRESULT hr = pTaskDialogIndirect(&tdc, &nButton, NULL, NULL);
+
+            if (SUCCEEDED(hr))
+            {
+                return nButton;
+            }
+        }
+
+        return 0;
```
m417z: It makes sense to fall back to `MessageBoxW_Original` here.

mods/better-dialogs.wh.cpp
FireBlade211: Yes, I have seen that mod, but this mod also routes the messages and events to keep compatibility.

mods/better-dialogs.wh.cpp
m417z: Not all calls were updated.

PR #3470, opened 2026-03-04
New method, support for 24h2 and later

Anixx:


mods/legacy-sound-flyout.wh.cpp
```diff
@@ -16,36 +17,286 @@ Enables legacy sound volume flyout on Windows 10 taskbar (running under either W
...
+}
 
-    return pOriginalRegQueryValueExW(hKey, lpValueName, lpReserved, lpType, lpData, lpcbData);
+// -------------------------------------------------------
+// Находим toolbar
+// -------------------------------------------------------
+
+static HWND FindTrayToolbar() {
+    HWND hTray = FindWindowW(L"Shell_TrayWnd", nullptr);
+    if (!hTray) return nullptr;
```
m417z, line 198: Also check that it's part of the current process.

Anixx: done

PR #3498, opened 2026-03-07
Photos Fullscreen Mode

mak7im01:
# Photos Fullscreen Mode

## How it works

The mod intercepts the creation of the Photos app main window and automatically applies the maximized window style, forcing the application to open in fullscreen.

## Compatibility

- Windows 10
- Windows 11
- Photos app (Microsoft.Photos.exe / Photos.exe)

## Notes

After installing this mod, the Photos app will automatically open in fullscreen mode. You can still manually resize or minimize the window if needed

### DIFF
--- a/mods/photos-fullscreen.wh.cpp
+++ b/mods/photos-fullscreen.wh.cpp
@@ -1,8 +1,8 @@
 // ==WindhawkMod==
 // @id              photos-fullscreen
-// @name            Photos Fullscreen Mode
+// @name            Windows Photos Fullscreen Mode
 // @description     Forces Windows Photos app to open in fullscreen mode instead of windowed mode
-// @version         1.0.2
+// @version         1.0.3
 // @author          mak7im01
 // @github          https://github.com/mak7im01
 // @include         Microsoft.Photos.exe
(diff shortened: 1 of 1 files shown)


mods/photos-fullscreen.wh.cpp
```diff
@@ -0,0 +1,185 @@
...
+
+## Notes
+
+After installing this mod, the Photos app will automatically open in fullscreen mode. You can still manually resize or minimize the window if needed.
+*/
+// ==/WindhawkModReadme==
+
+// ==WindhawkModSettings==
+/*
+- AutoMaximize: true
```
m417z: This option doesn't seem to be useful, as without it, the mod just doesn't do anything, right? The mod can just be disabled in this case.

mak7im01: You're right. I deleted the setting

mods/photos-fullscreen.wh.cpp
```diff
@@ -0,0 +1,177 @@
...
+ShowWindow_t ShowWindow_Original;
+
+// Settings
+struct {
+    BOOL autoMaximize;
+} g_settings;
+
+// Load settings
+void LoadSettings() {
+    g_settings.autoMaximize = Wh_GetIntSetting(L"AutoMaximize");
```
m417z: It's still being loaded, now the mod just won't work.

mak7im01: Settings have been completely removed

PR #3499, opened 2026-03-07
Add taskbar desktop roman indicator mod

simon-ami:
Adds a Windows 11 taskbar clock mod that shows the current virtual desktop number with configurable numbering, spacing, weight, and polling interval settings.

### DIFF
--- a/mods/taskbar-desktop-indicator.wh.cpp
+++ b/mods/taskbar-desktop-indicator.wh.cpp
@@ -32,11 +32,11 @@ Displays the current virtual desktop as a number or marker in the Windows 11 tas
 
 ## Screenshot
 
-Workspace markers mode (example):
-![Workspace markers mode](https://raw.githubusercontent.com/simon-ami/images/main/windhawk/taskbar-desktop-indicator/marker-indicator.png)
+![Workspace markers mode](https://raw.githubusercontent.com/simon-ami/images/main/windhawk/taskbar-desktop-indicator/marker-indicator.png) \
+_Workspace markers mode (example)_
 
-Number mode (example):
-![Number mode](https://raw.githubusercontent.com/simon-ami/images/main/windhawk/taskbar-desktop-indicator/roman-number.png)
+![Number mode](https://raw.githubusercontent.com/simon-ami/images/main/windhawk/taskbar-desktop-indicator/roman-number.png) \
+_Number mode (example)_
 
 ## Notes
 


mods/taskbar-desktop-roman-indicator.wh.cpp
```diff
@@ -0,0 +1,1473 @@
...
+* Configurable polling interval
+* Fast desktop change detection
+
+## Notes
+
+* Windows 11 only
+* The indicator is shown on the date line when available
+* The current desktop is detected by polling Explorer's virtual desktop state
+* Lower poll intervals react faster but increase registry reads in Explorer;
+  50-100 ms is usually fine, while going much lower is rarely worth it
```
m417z: How about adding a screenshot to the readme?

mods/taskbar-desktop-indicator.wh.cpp
```diff
@@ -0,0 +1,1473 @@
...
+    if (g_winVersion >= WinVersion::Win11 && !g_taskbarViewDllLoaded) {
+        if (HMODULE taskbarViewModule = GetTaskbarViewModuleHandle()) {
+            if (!g_taskbarViewDllLoaded.exchange(true) &&
+                HookTaskbarViewDllSymbols(taskbarViewModule)) {
+                Wh_ApplyHookOperations();
+            }
+        }
+    }
+
+    UpdateAllClockEntries(true);
```
m417z, line 2251: That doesn't seem to work - the mod doesn't apply when enabled, explorer has to be restarted after enabling the mod. Either improve it to make it work, or document it as a known limitation.

mods/taskbar-desktop-roman-indicator.wh.cpp
```diff
@@ -0,0 +1,1473 @@
...
+  $options:
+    - smaller: Smaller
+    - normal: Normal
+    - larger: Larger
+- inactiveMarkerOpacity: 22
+  $name: Inactive marker opacity
+  $description: Opacity percentage for non-active workspace markers. 100 matches the active marker, lower values make inactive markers dimmer.
+- pollIntervalMs: 100
+  $name: Poll interval (ms)
+  $description: How often Explorer checks for desktop changes. Lower values react faster but do more registry reads; 50-100 ms is usually fine, and values much lower than that are rarely worthwhile.
```
m417z: Instead of polling, you can use `CurrentVirtualDesktopChanged`. Unfortunately, it's undocumented, but it's used by various projects, and samples can be found online. Let me know if you'd like to use it, or if you'd like to keep the polling.

mods/taskbar-desktop-indicator.wh.cpp
simon-ami: Couldn’t reproduce locally, but I added an explicit live-clock refresh bootstrap after enable/settings change in https://github.com/simon-ami/windhawk-mods/commit/bec196c0d91f9cb0b17d0c9ac288c1d8f81c312f to make the mod less dependent on existing UI lifecycle timing.

mods/taskbar-desktop-roman-indicator.wh.cpp
simon-ami: Thanks, I've implemented `CurrentVirtualDesktopChanged` as the primary update path in https://github.com/simon-ami/windhawk-mods/commit/3d7da144f487e553721f39236eefcfe092fc4981, based on public examples, and kept optional polling as a fallback. On my machine the notification path now updates instantly on its own; polling is disabled by default and can be enabled if needed.

mods/taskbar-desktop-indicator.wh.cpp
simon-ami: I later noticed my testing had been helped by the Taskbar Clock Customization mod also being installed. The actual missing piece seems to have been a Win11 clock-update trigger, so I added the same clock watcher nudge in https://github.com/simon-ami/windhawk-mods/commit/4c8d3bf152f7c870a9bda5d94aafc3818478839d. After that, the mod applied immediately on its own without depending on the other mod.

mods/taskbar-desktop-roman-indicator.wh.cpp
simon-ami: Added screenshots to the readme in https://github.com/simon-ami/windhawk-mods/commit/6b3deea75db40d1c13f592f68c5b36e14f61cde7. I included examples for both workspace markers mode and number mode.

mods/taskbar-desktop-roman-indicator.wh.cpp
```diff
@@ -0,0 +1,2262 @@
+// ==WindhawkMod==
+// @id              taskbar-desktop-roman-indicator
+// @name            Taskbar Desktop Roman Indicator
```
m417z: Looks great! A minor suggestion before we merge the mod: Perhaps it'd be better to just call it "Taskbar Desktop Indicator". It's not limited to Roman numerals, and now it's not even the default. What do you think?

Also, the description implies that that's the only option.

mods/taskbar-desktop-indicator.wh.cpp
```diff
@@ -0,0 +1,2262 @@
...
+* Configurable left and right padding
+* Configurable spacing between indicator characters
+* Configurable indicator weight and size
+* Notification-based desktop change detection
+* Optional polling fallback
+
+## Screenshot
+
+Workspace markers mode (example):
+![Workspace markers mode](https://raw.githubusercontent.com/simon-ami/images/main/windhawk/taskbar-desktop-indicator/marker-indicator.png)
```
m417z: I think it'll look better with either a newline:
```markdown
Workspace markers mode (example): \
![Workspace markers mode](https://raw.githubusercontent.com/simon-ami/images/main/windhawk/taskbar-desktop-indicator/marker-indicator.png)
```

Or with a footer text, for example:
```markdown
![Workspace markers mode](https://raw.githubusercontent.com/simon-ami/images/main/windhawk/taskbar-desktop-indicator/marker-indicator.png) \
_Workspace markers mode (example)_
```

Same for the other screenshot below.

mods/taskbar-desktop-roman-indicator.wh.cpp
simon-ami: Yes, the renaming and docs alignment was done just now too

mods/taskbar-desktop-indicator.wh.cpp
simon-ami: Ah yes, didn't notice on my tiny screen here :D
Adjusted screenshot formatting in https://github.com/simon-ami/windhawk-mods/commit/90d1b446d971680b9dc74e3165172622c29f7cb4 to use footer-style captions for both images.

PR #3506, opened 2026-03-09
Discord Balloon Notifications mod

repensky:
Converts Discord's modern Windows toast notifications into balloon tips that appear from the system tray

### DIFF
--- a/mods/discord-balloon-notifs.wh.cpp
+++ b/mods/discord-balloon-notifs.wh.cpp
@@ -862,7 +862,7 @@ static char* DownloadEmojiJson(DWORD* outSize) {
     if (!hInternet) return nullptr;
     
     HINTERNET hUrl = InternetOpenUrlW(hInternet,
-        L"https://raw.githubusercontent.com/repensky/local-wh-mods/refs/heads/main/discord-emojis.json", //Downloads Discord emoji database
+        L"https://raw.githubusercontent.com/repensky/local-wh-mods/ac919c0ffe39ea171086c80a589c83bcabf432b5/discord-emojis.json", //Downloads Discord emoji database
         nullptr, 0,
         INTERNET_FLAG_RELOAD | INTERNET_FLAG_NO_CACHE_WRITE | INTERNET_FLAG_SECURE,
         0);


mods/discord-balloon-notifs.wh.cpp
```diff
@@ -0,0 +1,2089 @@
...
+
+static char* DownloadBasicGlyphJson(DWORD* outSize) {
+    *outSize = 0;
+    
+    HINTERNET hInternet = InternetOpenW(L"WindhawkDiscordBalloon/1.0",
+        INTERNET_OPEN_TYPE_PRECONFIG, nullptr, nullptr, 0);
+    if (!hInternet) return nullptr;
+    
+    HINTERNET hUrl = InternetOpenUrlW(hInternet,
+        L"https://raw.githubusercontent.com/repensky/local-wh-mods/refs/heads/main/discord-basic-glyphs.json", //Downloads basic glyphs list
```
m417z: Does it have to be downloaded externally?

mods/discord-balloon-notifs.wh.cpp
```diff
@@ -0,0 +1,2089 @@
...
+// Download the JSON file from GitHub
+static char* DownloadEmojiJson(DWORD* outSize) {
+    *outSize = 0;
+    
+    HINTERNET hInternet = InternetOpenW(L"WindhawkDiscordBalloon/1.0",
+        INTERNET_OPEN_TYPE_PRECONFIG, nullptr, nullptr, 0);
+    if (!hInternet) return nullptr;
+    
+    HINTERNET hUrl = InternetOpenUrlW(hInternet,
+        L"https://raw.githubusercontent.com/repensky/local-wh-mods/refs/heads/main/discord-emojis.json", //Downloads Discord emoji database
```
m417z: The mod has no indication in the options or in the readme that it uses external dependencies. For transparency, make it clear that this is the case, and make it possible to disable external dependencies for offline usage.

mods/discord-balloon-notifs.wh.cpp
```diff
@@ -0,0 +1,2089 @@
...
+    }
+    
+    CloseHandle(hDownloadMutex);
+    return 0;
+}
+
+static void GetEmojiCachePath() {
+    WCHAR winDir[MAX_PATH];
+    GetWindowsDirectoryW(winDir, MAX_PATH);
+    swprintf_s(g_emojiCachePath, L"%s\\Temp\\windhawk_discord_emojis.json", winDir);
```
m417z: Use `Wh_GetModStoragePath` for storing mod files.

mods/discord-balloon-notifs.wh.cpp
```diff
@@ -0,0 +1,2089 @@
...
+
+static void TryHookLoadXml(void* pXmlDocInstance) {
+    if (g_loadXmlHooked || !pXmlDocInstance) return;
+
+    void* pDocIO = nullptr;
+    HRESULT hr = ((IUnknown*)pXmlDocInstance)->QueryInterface(IID_IXmlDocumentIO, &pDocIO);
+    if (FAILED(hr) || !pDocIO) return;
+
+    void** ioVtbl = *(void***)pDocIO;
+    if (PatchVtableEntry(ioVtbl, 6, (void*)LoadXml_Hook, (void**)&LoadXml_Orig)) {
```
m417z, line 1804: VTable hooks aren't removed when the mod is unloaded, which is likely to lead to a crash.

mods/discord-balloon-notifs.wh.cpp
```diff
@@ -0,0 +1,2089 @@
...
+        if (hSnap != INVALID_HANDLE_VALUE) {
+            PROCESSENTRY32W pe = {};
+            pe.dwSize = sizeof(pe);
+            
+            if (Process32FirstW(hSnap, &pe)) {
+                do {
+                    if (_wcsicmp(pe.szExeFile, L"Discord.exe") == 0) {
+                        HANDLE hProc = OpenProcess(PROCESS_TERMINATE, FALSE, pe.th32ProcessID);
+                        if (hProc) {
+                            TerminateProcess(hProc, 0);
```
m417z, line 499: Will it also terminate the current process? If it will, then the loop will stop and other Discord processes might keep running.

mods/discord-balloon-notifs.wh.cpp
```diff
@@ -0,0 +1,2089 @@
...
+    // Start emoji loading in background
+    HANDLE hEmojiThread = CreateThread(nullptr, 0, EmojiLoadThread, nullptr, 0, nullptr);
+    if (hEmojiThread) CloseHandle(hEmojiThread);
+    
+    // Start basic glyph loading in background
+    HANDLE hGlyphThread = CreateThread(nullptr, 0, BasicGlyphLoadThread, nullptr, 0, nullptr);
+    if (hGlyphThread) CloseHandle(hGlyphThread);
+
+    g_hThread = CreateThread(nullptr, 0, BalloonThread, nullptr, 0, nullptr);
+    Sleep(200);
```
m417z: Why is this Sleep needed?

mods/discord-balloon-notifs.wh.cpp
repensky: Since the JSON file content is not that large for basic glyphs, I have incorporated it into the mod's code instead on the latest change, limiting external dependency to Discord's emoji database which is only downloaded and stored for subsequent use on first run (if convert emojis to shortcodes toggle is enabled).

mods/discord-balloon-notifs.wh.cpp
repensky: I have added in readme and a new toggle which includes the mention of downloading emoji's database if this particular toggle is on.

mods/discord-balloon-notifs.wh.cpp
repensky: I have switched to ``Wh_GetModStoragePath`` on the latest change.

mods/discord-balloon-notifs.wh.cpp
repensky: I have also addressed this in the latest change.

mods/discord-balloon-notifs.wh.cpp
repensky: Good catch - it does terminate the current process, making the loop stop early. I have updated the code to skip the current process during the loop to ensure all Discord related processes are terminated, then finally terminate the current process.

mods/discord-balloon-notifs.wh.cpp
repensky: It was added to allow the BalloonThread time to register the window class, create the hidden balloon window (``g_hBalloonWnd``) and enter its message loop. Without ``Sleep(200)`` in this code, it may cause a race condition where the thread may try to use ``g_hBalloonWnd`` before the balloon thread has created it. However, I have switched to a more robust approach in the latest change, which is to use an event that signals if it is ready to be used which may work better on varying systems.

mods/discord-balloon-notifs.wh.cpp
m417z: Since it's not expected to change, use the link for that specific commit:
`https://raw.githubusercontent.com/repensky/local-wh-mods/ac919c0ffe39ea171086c80a589c83bcabf432b5/discord-emojis.json`

repensky: Updated the mod code to use that specific link.

PR #3519, opened 2026-03-10
Hide Start Button

ptrkhh:
Hides the start button from the taskbar. The start menu can still be opened
via Win key, Ctrl+Esc key or touchscreen gestures.

Only Windows 11 is supported.

Based on "Start button always on the left" (taskbar-start-button-position) mod.

### DIFF
--- a/mods/hide-start-button.wh.cpp
+++ b/mods/hide-start-button.wh.cpp
@@ -268,7 +268,7 @@ bool HookTaskbarDllSymbols() {
     if (!module)
         return false;
 
-    WindhawkUtils::SYMBOL_HOOK hooks[] = {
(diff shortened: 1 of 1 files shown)


mods/hide-start-button.wh.cpp
```diff
@@ -0,0 +1,418 @@
...
+            startButton.Opacity(0);
+            startButton.IsHitTestVisible(false);
+
+            double startButtonWidth = startButton.ActualWidth();
+            if (startButtonWidth > 0) {
+                Thickness margin = startButton.Margin();
+                margin.Right = -startButtonWidth;
+                startButton.Margin(margin);
+            } else {
+                startButton.SizeChanged(
```
m417z: This callback is never unregistered. Doesn't it cause a crash when the mod is unloaded?

PR #3525, opened 2026-03-10
Add No SumatraPDF Drag Cursor mod

ItsProfessional:
Removes the drag cursor which is shown when panning a document by dragging with the cursor.
SumatraPDF Drag Cursor

### DIFF
--- a/mods/no-sumatrapdf-drag-cursor.wh.cpp
+++ b/mods/no-sumatrapdf-drag-cursor.wh.cpp
@@ -12,7 +12,8 @@
 
 // ==WindhawkModReadme==
 /*
-# No SumatraPDF drag cursor
+# No SumatraPDF Drag Cursor
+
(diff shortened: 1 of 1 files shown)


mods/no-sumatrapdf-drag-cursor.wh.cpp
```diff
@@ -0,0 +1,71 @@
...
+#define STDCALL  __cdecl
+#else
+#define STDCALL  __stdcall
+#endif
+
+HCURSOR g_hCursor{0};
+
+using LoadCursor_t = decltype(&LoadCursor);
+LoadCursor_t LoadCursor_orig;
+HCURSOR STDCALL LoadCursor_hook(HINSTANCE hInstance, LPCWSTR lpCursorName) {
```
m417z: Both are WinAPI functions, just use `WINAPI` instead of `STDCALL`.

PR #3568, opened 2026-03-16
Keep Rainmeter Always on Desktop Mod

bcrtvkcs:
# Keep Rainmeter Always on Desktop

Keeps Rainmeter windows to stay on desktop. Rainmeter skin windows can unexpectedly appear on top of other windows during window switching events (e.g. Alt+Tab), application focus changes, or when triggered by external scripts. This mod forces all Rainmeter windows to remain below all other windows at all times, using a `EVENT_SYSTEM_FOREGROUND` event hook to push them to the bottom of the Z-order whenever any window comes to the foreground.

## Compatibility
- Windows 10 and Windows 11
- Targets `windhawk.exe` only

### DIFF
--- a/mods/keep-rainmeter-always-bottom.wh.cpp
+++ b/mods/keep-rainmeter-always-bottom.wh.cpp
@@ -2,7 +2,7 @@
 // @id              keep-rainmeter-always-bottom
 // @name            Keep Rainmeter Always on Desktop
 // @description     Keeps Rainmeter windows to stay on desktop.
-// @version         1.1.1
+// @version         1.1.2
 // @author          BCRTVKCS
 // @github          https://github.com/bcrtvkcs
 // @twitter         https://x.com/bcrtvkcs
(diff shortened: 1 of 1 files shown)


mods/keep-rainmeter-always-bottom.wh.cpp
```diff
@@ -0,0 +1,90 @@
+// ==WindhawkMod==
+// @id              keep-rainmeter-always-bottom
+// @name            Keep Rainmeter Always on Desktop
+// @description     Keeps Rainmeter windows to stay on desktop.
+// @version         1.0
+// @author          BCRTVKCS
+// @github          https://github.com/bcrtvkcs
+// @twitter         https://x.com/bcrtvkcs
+// @homepage        https://grdigital.pro
+// @include         explorer.exe
```
m417z: See https://github.com/ramensoftware/windhawk/wiki/Mods-as-tools:-Running-mods-in-a-dedicated-process

mods/keep-rainmeter-always-bottom.wh.cpp
```diff
@@ -0,0 +1,90 @@
...
+// @compilerOptions -luser32 -lkernel32
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Keep Rainmeter Always on Desktop
+
+Keeps Rainmeter windows to stay on desktop.
```
m417z: What happens without the mod? What does the mod fix/improve?

bcrtvkcs: > What happens without the mod? What does the mod fix/improve?

Rainmeter skin windows can unexpectedly appear on top of other windows during window switching events (e.g. Alt+Tab), application focus changes, or when triggered by external scripts. This mod forces all Rainmeter windows to remain below all other windows at all times, using a `EVENT_SYSTEM_FOREGROUND` event hook to push them to the bottom of the Z-order whenever any window comes to the foreground.

PR #3610, opened 2026-03-21
Add Taskbar Numberer for Windows 11 mod

rik-smeets:
A mod that display keyboard shortcut numbers to taskbar items like 7+ Taskbar Numberer.

Heavily based on [this PR](https://github.com/ramensoftware/windhawk-mods/pull/2260) by @jsfdez, so massive thanks to him. I there also included him as author and mentioned in Special Thanks. If he prefers to pick this one back up again by himself, please let us know here!

The most prominent change is fixing the issue that taskbar items of the same app origin still incremented the number (when task bar items are not combined). 

Screenshot of Taskbar Numberer for Windows 11 mod

### DIFF
--- a/mods/taskbar-numberer.wh.cpp
+++ b/mods/taskbar-numberer.wh.cpp
@@ -352,7 +352,6 @@ FrameworkElement CreateNumberOverlay(int number) {
         SetNumberPosition(textContainer);
 
         textContainer.Margin(Thickness{4, 2, 4, 2});
-        Canvas::SetZIndex(textContainer, 0);
 
         if (number > 10) {
             textContainer.Visibility(Visibility::Collapsed);
@@ -519,7 +518,10 @@ void UpdateButtonOverlay(FrameworkElement button, int number) {
 
             auto newOverlay = CreateNumberOverlay(number);
             if (newOverlay) {
-                iconPanelElement.as<Panel>().Children().Append(newOverlay);
+                auto children = iconPanelElement.as<Panel>().Children();
+                
+                // Insert the new overlay just before the last child to ensure it appears above the icon, but below any other elements like badges or progress indicators
+                children.InsertAt(children.Size() - 1, newOverlay);
                 buttonInfo.overlay = newOverlay;
             }
         }


rik-smeets:
@m417z That works great! Even though the solution might be obscure, it should definitely be more reliable this way. Thanks!

I'm looking into an option to show icons on all taskbars (on multiple monitors) via a setting, if that doesn't cause any problems that is.

Do you have any other things that should be added, tested or fixed before an initial release?

mods/taskbar-numberer.wh.cpp
```diff
@@ -0,0 +1,871 @@
+// ==WindhawkMod==
+// @id              taskbar-numberer
+// @name            Taskbar Numberer for Windows 11
+// @description     Displays keyboard shortcut numbers (1-9, 0) on taskbar items like 7+ Taskbar Numberer
+// @version         1.0.0
+// @author          Rik Smeets & jsfdez
```
m417z: Leave only your name, which is the name of the GitHub profile author and the maintainer of the mod. Additional credits can be specified in the readme (as they are specified now).

You can also link [the research blog post](https://tightcorner.substack.com/p/reverse-engineering-windows-11s-taskbar) in the readme.

mods/taskbar-numberer.wh.cpp
```diff
@@ -0,0 +1,871 @@
...
+// @github          https://github.com/rik-smeets
+// @include         explorer.exe
+// @architecture    x86-64
+// @compilerOptions -DWINVER=0x0A00 -lole32 -loleaut32 -lruntimeobject
```
m417z: The WINVER flag was relevant for very old Windhawk versions and is no longer needed.
```suggestion
// @compilerOptions -lole32 -loleaut32 -lruntimeobject
```

mods/taskbar-numberer.wh.cpp
```diff
@@ -0,0 +1,871 @@
...
+  - topLeft: Top left
+  - topRight: Top right
+  - bottomLeft: Bottom left
+  - bottomRight: Bottom right
+- numberSize: 10
+  $name: Number font size
+  $description: Size of the overlay numbers (8-16)
+- numberColor: "#FFFFFF"
+  $name: Number color
+  $description: Text color for the numbers (hex format "#RRGGBB" or "#AARRGBB")
```
m417z:
```suggestion
  $description: Text color for the numbers (hex format "#RRGGBB" or "#AARRGGBB")
```

mods/taskbar-numberer.wh.cpp
```diff
@@ -0,0 +1,871 @@
...
+- backgroundColor: "#80000000"
+  $name: Stroke/outline color
+  $description: Outline color around the numbers for better visibility (hex format "#RRGGBB" or "#AARRGBB")
```
m417z:
```suggestion
  $description: Outline color around the numbers for better visibility (hex format "#RRGGBB" or "#AARRGGBB")
```

mods/taskbar-numberer.wh.cpp
```diff
@@ -0,0 +1,871 @@
...
+        (child = FindChildByClassName(child, L"SystemTray.SystemTrayFrame")) &&
+        (child = FindChildByName(child, L"SystemTrayFrameGrid")) &&
+        (child = FindChildByName(child, L"ControlCenterButton"))) {
+        controlCenterButton = child;
+    }
+    if (!controlCenterButton) return false;
+    return controlCenterButton.ActualWidth() < 5;
+}
+
+bool RunFromWindowThread(HWND hWnd, RunFromWindowThreadProc_t proc, void* procParam) {
```
m417z: This function is unused.

mods/taskbar-numberer.wh.cpp
```diff
@@ -0,0 +1,871 @@
...
+
+            auto buttonsCopy = buttons;
+            dispatcher.RunAsync(
+                CoreDispatcherPriority::Normal,
+                DispatchedHandler([buttonsCopy]() {
+                    for (const auto& bi : buttonsCopy) {
+                        CleanupButtonOverlays(bi);
+                    }
+                })
+            );
```
m417z: This is run asynchronously, let's wait for it to complete, something like:
```cpp
            HANDLE event = CreateEvent(nullptr, TRUE, FALSE, nullptr);
            dispatcher.RunAsync(
                CoreDispatcherPriority::Normal,
                DispatchedHandler([&buttons, event]() {
                    for (const auto& bi : buttons) {
                        CleanupButtonOverlays(bi);
                    }
                    SetEvent(event);
                })
            );
            WaitForSingleObject(event, INFINITE);
            CloseHandle(event);
```

**Edit:** Perhaps simply this will work:
```cpp
            dispatcher.RunAsync(
                CoreDispatcherPriority::Normal,
                DispatchedHandler([&buttons, event]() {
                    for (const auto& bi : buttons) {
                        CleanupButtonOverlays(bi);
                    }
                    SetEvent(event);
                })
            ).get();
```

I also removed the `buttons` copying since it's no longer accessed in parallel.

This also allows us to remove `Sleep(100);` below.

mods/taskbar-numberer.wh.cpp
```diff
@@ -0,0 +1,871 @@
...
+// For grouped items each group is already a single button, so they naturally
+// get distinct numbers. This function is only needed for ungrouped items where
+// multiple windows of the same app each have their own button.
+void* GetTaskGroupFromTaskListButton(UIElement element) {
+    if (!TryGetItemFromContainer_TaskListWindowViewModel_Original ||
+        !TaskListWindowViewModel_get_TaskItem_Original ||
+        !TaskItem_ReportClicked_Original) {
+        return nullptr;
+    }
+
```
m417z: This can actually be removed since all the variables are guaranteed to be set.

```suggestion
```

mods/taskbar-numberer.wh.cpp
```diff
@@ -0,0 +1,871 @@
...
+    }
+};
+
+using XamlRootKey = void*;
+
+// Simplified globals
+std::atomic<bool> g_taskbarViewDllLoaded{false};
+std::atomic<bool> g_unloading{false};
+std::mutex g_overlayMutex;
+std::unordered_map<XamlRootKey, std::vector<ButtonInformation>> g_trackedButtonsByRoot;
```
m417z: Using `void*` as a key can be problematic if a taskbar is removed (e.g. an external monitor is detached). A better way would be to have a vector of `winrt::weak_ref<XamlRoot>` and `std::vector<ButtonInformation>`.

mods/taskbar-numberer.wh.cpp
```diff
@@ -0,0 +1,871 @@
...
+}
+
+// Only when user setting is enabled and Windows taskbar mode is set to "All taskbars"
+bool ShouldShowOnAllTaskbars() {
+    if (!g_settings.showOnAllTaskbars) return false;
+
+    try {
+        DWORD value = 0;
+        DWORD size = sizeof(value);
+        if (RegGetValueW(
```
m417z, line 575: Why is the registry check needed, why not just `return g_settings.showOnAllTaskbars`? If there are no secondary taskbars, you won't get events about secondary taskbars. Especially since this function is being called a lot.

mods/taskbar-numberer.wh.cpp
```diff
@@ -0,0 +1,871 @@
...
+        } else {
+            j++;
+        }
+    }
+    if (removedCount > 0) {
+        Wh_Log(L"RemoveExistingOverlays: Removed %d old overlays", removedCount);
+    }
+}
+
+void UpdateButtonOverlay(FrameworkElement taskListButtonElement, int number, void* buttonPointer) {
```
m417z: `buttonPointer` isn't needed here, you can use `taskListButtonElement`.

mods/taskbar-numberer.wh.cpp
rik-smeets: The registry check doesn't check whether there are secondary taskbars, it checks the way in which taskbar items are grouped. Windows has an option which only shows taskbar apps with an open window on secondary taskbars. 

If that is the case, numbering would be off on those task bars, so we shouldn't show task bar items there.

mods/taskbar-numberer.wh.cpp
rik-smeets: Fixed!

mods/taskbar-numberer.wh.cpp
rik-smeets: Fixed!

mods/taskbar-numberer.wh.cpp
rik-smeets: Fixed!

mods/taskbar-numberer.wh.cpp
rik-smeets: Fixed!

mods/taskbar-numberer.wh.cpp
rik-smeets: Fixed!

mods/taskbar-numberer.wh.cpp
rik-smeets: Thx, I removed it.

mods/taskbar-numberer.wh.cpp
rik-smeets: Thx, I updated it in the second way. Seems to work just fine :)

mods/taskbar-numberer.wh.cpp
rik-smeets: Fixed as well.

mods/taskbar-numberer.wh.cpp
rik-smeets: Removed!

m417z:
Thanks! And thanks @jsfdez for the initial development.

ovgolovin:
Thank you for releasing this plugin!
I have tested it and created an issue with the bug I found: https://github.com/ramensoftware/windhawk-mods/issues/3651

PR #3611, opened 2026-03-21
Word PDF Lossless Export v1.1.1 Update

JoeYe-233:
Avoid fetching for online symbol cache


m417z:
> It's due to network issue, but it's a general issue that affect all people here in Mainland China. Github Pages (github.io) seems to get blocked before, because of abuse, making it very unstable even today. Sometimes the TCP request gets lost along the way.

That's too bad, it means that all mods are affected, and the result is a 30 seconds delay and then downloading of the full PDB, instead of just downloading the tiny cache. Is it about DNS, or the GitHub IPs? If you add the entry to the hosts file (in my case `ramensoftware.github.io` resolves to `2606:50c0:8002::153`/`185.199.109.153`), does it work? Also, if you think there's a way to detect this kind of failure as opposed to just having no internet connection, let me know. It'd be better to skip the retries in this case.

PR #3613, opened 2026-03-22
First release of Nilesoft Animator Mod

Lockframe:
* Renamed from Animated Nilesoft Shell.
* Includes several bug fixes and changes compared to the discord version.

### DIFF
--- a/mods/alt-tab-per-monitor.wh.cpp
+++ b/mods/alt-tab-per-monitor.wh.cpp
@@ -2,7 +2,7 @@
 // @id              alt-tab-per-monitor
 // @name            Alt+Tab per monitor
 // @description     Pressing Alt+Tab shows all open windows on the primary display. This mod shows only the windows on the monitor where the cursor is.
-// @version         1.1.1
+// @version         1.2
 // @author          L3r0y
 // @github          https://github.com/L3r0yThingz
 // @include         explorer.exe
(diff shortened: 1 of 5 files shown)


mods/nilesoft-shell-animator.wh.cpp
```diff
@@ -0,0 +1,1301 @@
...
+        }
+    }
+}
+
+static DWORD WINAPI PersistentAnimationThread(LPVOID) {
+    // Bump to HIGHEST to combat CPU starvation and ensure smooth UI scheduling
+    SetThreadPriority(GetCurrentThread(), THREAD_PRIORITY_HIGHEST); 
+    
+    // Bind 1ms timer resolution to the thread lifecycle to prevent kernel scheduler thrashing
+    timeBeginPeriod(1);
```
m417z: The resolution is changed for the entire lifetime of the mod, potentially affecting power consumption. Will it work if it's enabled only for the actual animation? Something like:
```cpp
if (group) {
    timeBeginPeriod(1);
    RunAnimation(group);
    timeEndPeriod(1);
    FreeGroup(group);
}
```

Lockframe: I didn't think about that and power consumption, thanks for pointing it out

PR #3615, opened 2026-03-22
Initial release of AMD Radeon Software Window Fix Windhawk mod

St0RM53:
# AMD Radeon Software Window Fix

Forces the position, size, and maximized state of Radeon Software (RadeonSoftware.exe) every time its window is shown from the system tray. Optionally forces the startup page.

----------------------------------------------------------------------------------------------

AMD's Radeon Software runs persistently in the system tray. When you double-click
the tray icon to open it, the app reads its size from the registry and then centers
itself — ignoring any position you prefer, and doing so *after* any `SetWindowPos`
call, making simple hooks ineffective.

The behavior is especially frustrating on multi-monitor setups with extended desktop
and has been most likely present since the Adrenalin edition release from 2018. Since
AMD devs are unwilling to even acknowledge their incompetence or even attempt to fix
it, i've used some of my knowledge and a few AI coding tools to give a working solution.

This mod subclasses the main window to intercept `WM_WINDOWPOSCHANGING` — a message
fired *before* any move or resize takes effect — and overrides the position and size
there, which is the only intercept point that reliably wins against Qt's internal
layout logic.

Tested on version 26.3.1 (which introduced a metrics overlay bug), but should work on 
any of the recent (2025-2026) versions that use the Qt 6.x.x framework.

### DIFF
--- a/mods/amd-radeon-software-window-fix.wh.cpp
+++ b/mods/amd-radeon-software-window-fix.wh.cpp
@@ -73,6 +73,7 @@ Open the mod's **Settings** tab to configure:
 // ==/WindhawkModSettings==
 
 #include <windhawk_api.h>
+#include <windhawk_utils.h>
 #include <windows.h>
 #include <commctrl.h>
 #include <algorithm>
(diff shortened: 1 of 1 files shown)


mods/amd-radeon-software-window-fix.wh.cpp
```diff
@@ -0,0 +1,261 @@
...
+            Wh_Log(L"WM_WINDOWPOSCHANGING: forced to (%d,%d %dx%d)",
+                   wp->x, wp->y, wp->cx, wp->cy);
+        }
+    }
+    else if (uMsg == WM_SHOWWINDOW && wParam == TRUE) {
+        if (g_settings.forceStartupPage) {
+            WriteResumeLastPage();
+        }
+        if (g_settings.maximized) {
+            PostMessage(hWnd, WM_USER + 1, 0, 0);
```
m417z: `WM_USER + 1` could collide with Qt's own message handling. Prefer using `RegisterWindowMessage()` to create a unique message id.

mods/amd-radeon-software-window-fix.wh.cpp
```diff
@@ -0,0 +1,261 @@
...
+    GetWindowThreadProcessId(hWnd, &pid);
+
+    if (pid == GetCurrentProcessId()) {
+        wchar_t className[64] = {};
+        GetClassNameW(hWnd, className, _countof(className));
+
+        if (wcscmp(className, L"Qt683QWindowIcon") == 0) {
+            DWORD_PTR existing = 0;
+            if (!GetWindowSubclass(hWnd, SubclassProc, SUBCLASS_ID, &existing)) {
+                if (SetWindowSubclass(hWnd, SubclassProc, SUBCLASS_ID, 0)) {
```
m417z: The subclass is not removed when the mod is unloaded. If the window is usually short-lived, perhaps it's acceptable, but you might want to address this. The easiest way is to use the Windhawk APIs: `WindhawkUtils::SetWindowSubclassFromAnyThread` and `WindhawkUtils::RemoveWindowSubclassFromAnyThread`. Save the window handle, and remove the subclass on mod unload.

PR #3630, opened 2026-03-23
Added Context Menu Preloader mod

Lockframe:


mods/context-menu-preloader.wh.cpp
```diff
@@ -0,0 +1,481 @@
...
+    
+    PCWSTR customExtsRaw = Wh_GetStringSetting(L"CustomExtensions");
+    if (customExtsRaw) {
+        ParseExtensions(customExtsRaw);
+        Wh_FreeStringSetting(customExtsRaw);
+    }
+}
+
+BOOL Wh_ModInit() {
+    LoadSettings();
```
m417z, line 478: Settings aren't reloaded in `Wh_ModSettingsChanged` (which is not defined).

mods/context-menu-preloader.wh.cpp
```diff
@@ -0,0 +1,481 @@
+// ==WindhawkMod==
+// @id              context-menu-preloader
+// @name            Context Menu Preloader
+// @description     Instantly loads context menus. Includes RAM usage monitoring and proper thread cleanup.
```
m417z: > Includes RAM usage monitoring and proper thread cleanup

These are about debug logs and mod correctness, I don't think they should be part of the description. It could be similar to the first line of the readme instead.

PR #3642, opened 2026-03-25
Create better-uwp-clean.wh.cpp

smlfrysamuri:
Modified the uwp-clean mod to preserve RuntimeBroker.exe functionality, while still ending UWP processes that are not shell-critical using a more targeted approach. 

This primarily makes it so taskbar context menu functionality (i.e., right-clicking taskbar buttons and the resulting context menus) are preserved, as they rely on RuntimeBroker.exe.

### DIFF
--- a/mods/better-uwp-clean.wh.cpp
+++ b/mods/better-uwp-clean.wh.cpp
@@ -1,0 +1,122 @@
+// ==WindhawkMod==
+// @id                better-uwp-clean
+// @name            Clean Up UWP Processes (Improved)
+// @description    Automatically end UWP processes when no UWP apps are open, except for shell-critical processes.
+// @version         0.1
+// @author          Smlfrysamuri
+// @github          https://github.com/smlfrysamuri
+// @include         ApplicationFrameHost.exe
+// @include         TextInputHost.exe
+// @include         SystemSettingsAdminFlows.exe
+// @include         ctfmon.exe
+// @compilerOptions -lcomdlg32
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# (Better) UWP Clean
+This is a slightly modified version of the UWP Clean mod created by https://github.com/arukateru. Whereas the
+original mod operated in such a way that caused RuntimeBroker.exe to exit, this mod still disables UWP processes
+without killing RuntimeBroker. This preserves shell-critical processes that use UWP, most importantly, the ability to 
+right-click taskbar menu buttons and access their context menus, which relies on RuntimeBroker.
+*/
+// ==/WindhawkModReadme==
+
+#include <Windows.h>
+#include <string>
+#include <thread>
+#include <TlHelp32.h>
+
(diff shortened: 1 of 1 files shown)


mods/better-uwp-clean.wh.cpp
```diff
@@ -0,0 +1,122 @@
...
+// @include         TextInputHost.exe
+// @include         SystemSettingsAdminFlows.exe
+// @include         ctfmon.exe
+// @compilerOptions -lcomdlg32
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# (Better) UWP Clean
+This is a slightly modified version of the UWP Clean mod created by https://github.com/arukateru. Whereas the
```
m417z, line 18: I suggest to add a description for what the mod does that doesn't require looking at another mod. The other mod can be linked later as a side note.

Also, add its link (https://windhawk.net/mods/uwp-clean).

mods/better-uwp-clean.wh.cpp
```diff
@@ -0,0 +1,122 @@
...
+) {
+    switch (ul_reason_for_call) {
+    case DLL_PROCESS_ATTACH:
+        std::thread(CheckProcesses).detach();
+        break;
+    default:
+        break;
+    }
+    return TRUE;
+}
```
m417z, line 122: I know that the reference mod does this too, but prefer using Wh_ModInit:

```suggestion
BOOL Wh_ModInit() {
    std::thread(CheckProcesses).detach();
    return TRUE;
}
```

Also, currently, there's no handling for mod unload. If possible, add proper cleanup in `Wh_ModUninit`.

mods/better-uwp-clean.wh.cpp
```diff
@@ -0,0 +1,122 @@
...
+        bool wduiRunning           = IsProcessRunning(L"SecHealthUI.exe");
+        bool storeRunning          = IsProcessRunning(L"WinStore.App.exe");
+
+        if (!systemSettingsRunning && !wduiRunning && !storeRunning) {
+            // Only self-terminate safe, non-shell-critical processes.
+            // Never terminate RuntimeBroker.exe — it backs taskbar context menus.
+            if (currentProc == L"ApplicationFrameHost.exe" ||
+                currentProc == L"TextInputHost.exe"        ||
+                currentProc == L"SystemSettingsAdminFlows.exe" ||
+                currentProc == L"ctfmon.exe") {
```
m417z, line 102: Since the mod targets only these processes, I'm not sure why this check is needed.

mods/better-uwp-clean.wh.cpp
```diff
@@ -0,0 +1,122 @@
...
+            return true;
+        }
+    } while (Process32NextW(hSnap, &pe32));
+
+    CloseHandle(hSnap);
+    return false;
+}
+
+// Terminate a specific process by name instead of the current host process
+void TerminateProcessByName(const std::wstring& processName) {
```
m417z, line 53: This function is unused.

PR #3644, opened 2026-03-26
add: dwrite-font-fallback-override

RagibHasin:


mods/dwrite-font-fallback-override.wh.cpp
```diff
@@ -0,0 +1,363 @@
...
+void Wh_ModUninit() {
+    Wh_Log(L"DirectWrite Font Fallback Override (C++/WinRT) - Uninitializing");
+
+    // // Uninitialize WinRT
+    // uninit_apartment();
+}
+
+void Wh_ModSettingsChanged() {
+    Wh_Log(L"Settings changed, reloading...");
+    LoadSettings();
```
m417z: `LoadSettings` reloads rules into `g_fallbackRules` but `g_fallback` is never reset to `nullptr`. The cached fallback object (built with old rules) continues to be used forever.

mods/dwrite-font-fallback-override.wh.cpp
```diff
@@ -0,0 +1,363 @@
...
+using GetSystemFontFallback_t = HRESULT(
+    STDMETHODCALLTYPE*)(IDWriteFactory8* self, IDWriteFontFallback1** fallback);
+
+GetSystemFontFallback_t g_originalGetSystemFontFallback = nullptr;
+
+STDMETHODIMP GetSystemFontFallback_Hook(IDWriteFactory8* self,
+                                        IDWriteFontFallback1** fallback) {
+    Wh_Log(L"GetSystemFontFallback");
+
+    if (g_fallback == nullptr) {
```
m417z: Multiple threads calling DirectWrite concurrently can all enter the initialization block simultaneously, leaving `g_fallback` in an inconsistent state. That's especially important since this mod targets all processes. Add a synchronization mechanism.

mods/dwrite-font-fallback-override.wh.cpp
```diff
@@ -0,0 +1,363 @@
...
+    // IDWriteFactory6 :  7 : 49
+
+    auto pGetSystemFontFallback =
+        get_item_in_vtable(factory, 26);  // 1st of IDWriteFactory2
+    auto pCreateTextFormat0 =
+        get_item_in_vtable(factory, 15);  // 13th of IDWriteFactory
+    auto pCreateTextLayout0 =
+        get_item_in_vtable(factory, 18);  // 16th of IDWriteFactory
+    auto pCreateTextFormat6 =
+        get_item_in_vtable(factory, 55);  // 7th of IDWriteFactory6
```
m417z: `IDWriteFactory6` is only supported from Windows 10 Build 16299. What happens on older builds? I assume it will crash, which is especially bad since the mod targets all processes. Add a check for it, for example by querying for `IDWriteFactory6` and checking that the interface is supported.

mods/dwrite-font-fallback-override.wh.cpp
```diff
@@ -0,0 +1,363 @@
...
+        // Try to set custom fallback on the text layout
+        com_ptr<IDWriteTextFormat> format;
+        format.copy_from(*textFormat);
+
+        auto format3 = format.try_as<IDWriteTextFormat3>();
+        if (format3) {
+            Wh_Log(L"");
+            com_ptr<IDWriteFontFallback> customFallback;
+            self->GetSystemFontFallback(customFallback.put());
+            Wh_Log(L"");
```
m417z: These seem to be leftover testing logs.

mods/dwrite-font-fallback-override.wh.cpp
```diff
@@ -0,0 +1,363 @@
...
+        return FALSE;
+    }
+    Wh_SetFunctionHook(pCreateTextLayout0, (void*)CreateTextLayout0_Hook,
+                       (void**)&g_originalCreateTextLayout0);
+    Wh_Log(L"Successfully hooked CreateTextLayout0");
+
+    if (!supportsFactory6) {
+        Wh_Log(L"System does not support IDWriteFactory6");
+        Wh_Log(L"Did not attempt hooking CreateTextFormat6");
+        return TRUE;
```
m417z, line 341: Why not `return FALSE;`?

RagibHasin: It would still work as intended by the mod in this case I think? I don't have a Windows 10 machine on hand to test though.

m417z: If the idea is to keep the mod, there are similar issues with other interfaces such as `IDWriteFactory5`, introduced in Windows 10 Build 15063, etc.

RagibHasin: Which minimum version of Windows does Windhawk support?

m417z: Windows 7. Regardless, it's an easy check, and it's OK to just return false for older versions, but I prefer to avoid crashing all processes even for old supported Windows versions.

RagibHasin: This mod does not hook anything from `IDWriteFactory5`, only from `IDWriteFactory`, `IDWriteFactory2`, and `IDWriteFactory6`. Of these, only `IDWriteFactory6` is conditionally available, is that not so? And this code now already checks for availability of `IDWriteFactory6`, would that no be sufficient?

m417z: OK. `IDWriteFactory2` is available from Windows 8.1 but that's old enough.

PR #3647, opened 2026-03-26
Update auto-time-sync-on-startup.wh.cpp

communism420:
- Fixed the session-run guard logic to use a real volatile registry subkey instead of a persistent value under `HKCU\Software`.
- Added cleanup for the legacy persistent guard value that could make the mod think the sync had already been done after reboot.
- Version bumped to `1.1`.

### DIFF
--- a/mods/auto-time-sync-on-startup.wh.cpp
+++ b/mods/auto-time-sync-on-startup.wh.cpp
@@ -2,7 +2,7 @@
 // @id              auto-time-sync-on-startup
 // @name            Auto Time Sync On Startup
 // @description     Requests a Windows time synchronization when Explorer starts after sign-in.
-// @version         1.0
+// @version         1.1
 // @author          communism420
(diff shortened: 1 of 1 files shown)


mods/auto-time-sync-on-startup.wh.cpp
```diff
@@ -155,17 +158,38 @@ bool SleepWithStopCheck(DWORD durationMs) {
...
-                                  &disposition);
+    result = RegCreateKeyExW(parentKey,
+                             kRunGuardSubKey,
+                             0,
+                             nullptr,
+                             REG_OPTION_VOLATILE,
+                             KEY_QUERY_VALUE | KEY_SET_VALUE,
+                             nullptr,
+                             &key,
+                             &disposition);
```
m417z, line 191: In case you like aesthetics: you probably can just check `disposition` here, and return based on that. That makes the value below unnecessary.

PR #3649, opened 2026-03-26
Create lockscreen-wallpaper.wh.cpp

Siddharthxd99:


mods/lockscreen-wallpaper.wh.cpp
```diff
@@ -0,0 +1,447 @@
...
+    RegDeleteVal(HKEY_LOCAL_MACHINE, kCSPKey, kCSPStat);
+    Sleep(150);   // small pause so Windows sees the delete before the rewrite
+
+    // Method 1: Group Policy (SYSTEM can write this)
+    bool gpo = RegWriteSz(HKEY_LOCAL_MACHINE, kGPOKey, kGPOVal, jp);
+
+    // Method 2: PersonalizationCSP
+    bool c1  = RegWriteSz   (HKEY_LOCAL_MACHINE, kCSPKey, kCSPPath, jp);
+    bool c2  = RegWriteSz   (HKEY_LOCAL_MACHINE, kCSPKey, kCSPUrl,  jp);
+    bool c3  = RegWriteDword(HKEY_LOCAL_MACHINE, kCSPKey, kCSPStat, 1);
```
m417z, line 178: One of the main advantages of Windhawk is that mods don't do persistent changes to the system, unlike other solutions which modify the registry or system files. If something goes wrong, a mod can always be disabled to revert the changes.

Contrary to this, this mod makes changes to system registry entries, which doesn't align with Windhawk principles. One way to have a compatible implementation is to hook registry read functions and return different values. There are several examples for this:
https://github.com/search?q=repo%3Aramensoftware%2Fwindhawk-mods+REGQUERYVALUEEXW&type=code

Siddharthxd99: when will this mod be added in windhawk?

m417z: It will not be added in the current form for the reason I outlined.

Siddharthxd99: could you please explain the reason I didn't get it sorry

m417z: Windhawk is a tool that lets you customize Windows without permanently changing your computer's settings or important files. Mods are safe because you can simply turn them off if there is a problem. However, this specific mod mentioned here is different because it actually changes your system's permanent records (the registry).

Siddharthxd99: thank you for the explanation I'll see what I can do to resolve but if I be honest I don't think we can change the lockscreen wallpaper without registry and this is very necessary I feel because other softwares like wallpaper engine and etc take money to let you change the lockscreen and most of the people doesn't wanna spend their money so, I wanted to make a free alternative so let's see what I can do to fix the issue

PR #3650, opened 2026-03-26
Add one-hair mod

Acercandr0:


mods/one-hair.wh.cpp
```diff
@@ -0,0 +1,321 @@
+// ==WindhawkMod==
+// @id              one-hair
+// @name            One Hair
+// @description     A hair on your screen. Always on top.
+// @version         1.0
+// @author          Acercandr0
+// @github          https://github.com/Acercandr0
+// @include         explorer.exe
```
m417z: See: https://github.com/ramensoftware/windhawk/wiki/Mods-as-tools:-Running-mods-in-a-dedicated-process

Acercandr0: Done!

m417z: I see that your snippet has some changes compared to the code in the wiki. Please copy the snippet as is.

mods/one-hair.wh.cpp
```diff
@@ -0,0 +1,499 @@
...
+        return FALSE;
+    }
+
+    WaitForSingleObject(g_hWindowReady, 5000);
+    return TRUE;
+}
+
+void WhTool_ModSettingsChanged()
+{
+    int newInterval = Wh_GetIntSetting(L"moveInterval");
```
m417z: Looks like settings are loaded when they're changed, but not when the mod is loaded.

mods/one-hair.wh.cpp
```diff
@@ -0,0 +1,499 @@
...
+}
+
+// ─────────────────────────────────────────────
+//  Tool mod implementation
+// ─────────────────────────────────────────────
+BOOL WhTool_ModInit()
+{
+    Wh_Log(L"WhTool_ModInit called");
+
+    g_hExitEvent = CreateEvent(nullptr, TRUE, FALSE, nullptr);
```
m417z: `g_hExitEvent` isn't used and can be removed.

PR #3653, opened 2026-03-27
Updated Taskbar ContentPresenter Injector mod

Lockframe:
* Fixed 25H2 support by implementing deferred injection
* Improved structure tracking logic and performance
* Added multi-monitor support
* The mod now injects into certain system tray elements as well

Initial deferred injection patch by [Acercandro](https://github.com/acercandr0)

### DIFF
--- a/mods/taskbar-content-presenter-injector.wh.cpp
+++ b/mods/taskbar-content-presenter-injector.wh.cpp
@@ -40,24 +40,26 @@ Due to how Windows 11 calculates taskbar dimensions, if you want to change the W
 
 #include <atomic>
 #include <mutex>
-#include <map>
-#include <set>
 
 using namespace winrt::Windows::UI::Xaml;
 
 std::atomic<bool> g_taskbarViewDllLoaded = false;
 
 // --- Global Trackers ---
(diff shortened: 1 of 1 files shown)


mods/taskbar-content-presenter-injector.wh.cpp
```diff
@@ -36,40 +39,44 @@ This mod acts as an addon to the [Windows 11 Taskbar Styler mod](https://windhaw
...
-struct TrackedPanelRef {
-    winrt::weak_ref<Controls::Panel> ref;
+// --- Global Trackers ---
+struct PendingHook {
+    winrt::weak_ref<Controls::Panel> panelRef;
+    winrt::event_token token;
 };
 
-std::vector<TrackedPanelRef> g_trackedPanels;
+std::map<void*, winrt::weak_ref<Controls::Panel>> g_trackedPanels;
```
m417z: Using `void*` here and below as a key can be problematic if the element is destroyed. Then the address may be reused by a different element. A better way would be to have a vector of `winrt::weak_ref` as it was before.

PR #3655, opened 2026-03-27
add scrambled eggs mod

dell-optiplex-790:


mods/scrambled-eggs.wh.cpp
```diff
@@ -0,0 +1,543 @@
...
+// @version         1.0
+// @author          dell-optiplex-790
+// @include         *
+// @compilerOptions -lgdi32
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Scrambled Eggs
+This is a fork of the Replace Text mod that completely scrambles the text on your computer. Do not attempt to use your computer like this, unless you want a really bad time.
```
m417z, line 14: How about adding a screenshot to show what that means? Or do you prefer for it to be a surprise?

dell-optiplex-790: I prefer it to be a surprise.

PR #3658, opened 2026-03-28
New Mod: Force Kill Active Window

zunaidFarouque:
Initial submission of the Force Kill Active Window mod.

Architectural highlights:
* System-Level Termination: Utilizes `TerminateProcess` to instantly kill unresponsive foreground windows, bypassing standard polite closure.
* Safe Injection: Logic is restricted to the primary `explorer.exe` shell to prevent duplicate hidden threads.
* Process Blacklist: Includes a user-configurable comma-separated blacklist to protect critical applications from accidental termination.
* Thread-Safe Live Settings: Implements `std::mutex` locks and custom thread messages (`WM_RELOAD_HOTKEY`) to safely unbind and rebind the hotkey when the user modifies settings, preventing access violations.
* Strict Teardown Lifecycle: Uses `WaitForSingleObject` during `Wh_ModUninit` to guarantee the background thread is dead before the DLL unloads, ensuring Explorer stability.

### DIFF
--- a/mods/force-kill-active-window.wh.cpp
+++ b/mods/force-kill-active-window.wh.cpp
@@ -12,23 +12,16 @@
 /*
 # Force Kill Active Window
 
-Instantly terminate any frozen or unresponsive application with a keyboard
-shortcut.
+Instantly terminate any frozen or unresponsive application with a keyboard shortcut.
 
-By default, pressing **Ctrl+Alt+F4** will execute a hard system kill on the
-currently active foreground window, bypassing the standard Windows polite
-closure request.
+By default, pressing **Ctrl+Alt+F4** will execute a hard system kill on the currently active foreground window, bypassing the standard Windows polite closure request.
 
 ### ⚠️ Warning
-This performs a hard system termination (`TerminateProcess`). The target
-application will **not** be given a chance to save your work. Use strictly as a
-last resort.
+This performs a hard system termination (`TerminateProcess`). The target application will **not** be given a chance to save your work. Use strictly as a last resort.
 
 ### Features
-* **Customizable Hotkey:** Change the trigger shortcut in the settings. Updates
-apply instantly.
-* **Process Blacklist:** Define apps (like `winword.exe` or `chrome.exe`) that
-the mod is forbidden from killing to protect your most critical work.
+* **Customizable Hotkey:** Change the trigger shortcut in the settings. Updates apply instantly.
+* **Process Blacklist:** Define apps (like `winword.exe` or `chrome.exe`) that the mod is forbidden from killing to protect your most critical work.
 */
 // ==/WindhawkModReadme==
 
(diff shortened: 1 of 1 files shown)


mods/force-kill-active-window.wh.cpp
```diff
@@ -0,0 +1,242 @@
+// ==WindhawkMod==
+// @id              force-kill-active-window
+// @name            Force Kill Active Window
+// @description     Press Ctrl+Alt+F4 to forcefully kill the active foreground window.
+// @version         1.0
+// @author          vfxturjo
+// @include         explorer.exe
```
m417z: See https://github.com/ramensoftware/windhawk/wiki/Mods-as-tools:-Running-mods-in-a-dedicated-process

zunaidFarouque: Refactored to run as a dedicated tool process using the wrapper. Thanks for the guidance!

mods/force-kill-active-window.wh.cpp
```diff
@@ -0,0 +1,429 @@
...
+        wcscmp(argv[i], L"-service-stop") == 0) {
+      isExcluded = true;
+      break;
+    }
+  }
+
+  for (int i = 1; i < argc - 1; i++) {
+    if (wcscmp(argv[i], L"-tool-mod") == 0) {
+      isToolModProcess = true;
+      if (wcscmp(argv[i + 1], L"force-kill-active-window") == 0) {
```
m417z: Please keep the snippet from the wiki as is, there's no need to make changes. It will help with the maintenance. Thanks.

mods/force-kill-active-window.wh.cpp
```diff
@@ -0,0 +1,429 @@
...
+      return VK_F1 + fNum - 1;
+  }
+  return VK_F4;
+}
+
+void LoadSettings() {
+  // Lock the mutex before modifying global settings
+  std::lock_guard<std::mutex> lock(g_settingsMutex);
+
+  g_modifiers = 0;
```
m417z, line 88: It'd be better to add `MOD_NOREPEAT` too, especially for a hotkey that kills processes.

mods/force-kill-active-window.wh.cpp
```diff
@@ -0,0 +1,429 @@
...
+  $description: The main key to press (e.g., F4, Q, X, 1).
+- req_ctrl: true
+  $name: Require Ctrl key
+- req_alt: true
+  $name: Require Alt key
+- req_shift: false
+  $name: Require Shift key
+- req_win: false
+  $name: Require Windows key
+- protected_apps: "devenv.exe, winword.exe, excel.exe"
```
m417z: Adding `explorer.exe` by default might be a good idea too.

zunaidFarouque: Thank you for the tip! added explorer.exe in the list.

mods/force-kill-active-window.wh.cpp
zunaidFarouque: Thank you. Added the modifier :)

mods/force-kill-active-window.wh.cpp
zunaidFarouque: Thank you.
Copied it without modification this time.

PR #3660, opened 2026-03-28
New Mod Submission: Taskbar Z-order Override

Meteony:
Control whether the taskbar stays always on top, always at the bottom, or behaves like a normal window (ported from 7+ Taskbar Tweaker). 

Mimic normal window:
image

Always on top (for auto-hide users primarily):
image

Always at bottom: 
image

### DIFF
--- a/mods/taskbar-z-order-override.wh.cpp
+++ b/mods/taskbar-z-order-override.wh.cpp
@@ -25,7 +25,7 @@ Makes the taskbar behave like a regular window - like how it was possible pre-Vi
 
 ### Always on top
 
-Keeps the taskbar above normal windows, no matter in games or in browser fullscreen. 
+Keeps the taskbar above normal windows - no matter in games or browser fullscreen. 
 
 This is mainly useful for people who use **auto-hide** and want the taskbar to stay accessible consistently.
 
@@ -77,8 +77,6 @@ enum class TaskbarZOrder {
 
 namespace {
 
-constexpr DWORD kDesktopBand = 1;
-
 using SetWindowPos_t = BOOL(WINAPI*)(HWND, HWND, int, int, int, int, UINT);
 
 using SetWindowBand_t = BOOL(WINAPI*)(HWND, HWND, DWORD);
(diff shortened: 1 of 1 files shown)


mods/taskbar-z-order-override.wh.cpp
```diff
@@ -0,0 +1,387 @@
...
+    return;
+  }
+
+  g_state.taskbarThreadId =
+      GetWindowThreadProcessId(g_state.taskbarWnd, nullptr);
+
+  SetLastError(0);
+  auto prev = reinterpret_cast<WNDPROC>(
+      SetWindowLongPtrW(g_state.taskbarWnd,
+                        GWLP_WNDPROC,
```
m417z: `GWLP_WNDPROC` doesn't work well if more than one mod subclasses the window. Use `WindhawkUtils::SetWindowSubclassFromAnyThread`, see the wiki for more details: https://github.com/ramensoftware/windhawk/wiki/Development-tips

mods/taskbar-z-order-override.wh.cpp
```diff
@@ -0,0 +1,387 @@
+// ==WindhawkMod==
+// @id              taskbar-z-order-override
+// @name            Taskbar Z-Order Override
+// @description     Control whether the taskbar stays always on top, always at the bottom, or behaves like a normal window
+// @version         0.5
+// @author          meteoni
+// @github          https://github.com/Meteony
+// @include         explorer.exe
+// @architecture    x86-64
+// ==/WindhawkMod==
```
m417z, line 11: How about adding a readme? The GitHub submission has nice description and screenshots which could also be added as a readme.

mods/taskbar-z-order-override.wh.cpp
```diff
@@ -0,0 +1,387 @@
...
+                          reinterpret_cast<void**>(&SetWindowBand_Original))) {
+    Wh_Log(L"Failed to set SetWindowBand function hook");
+    return FALSE;
+  }
+
+  return TRUE;
+}
+
+void Wh_ModAfterInit() {
+  InstallTaskbarSubclass();
```
m417z: What if the mod is loaded before the taskbar is created? One way to handle it is to hook `CreateWindowExW`, there are several mods that can be used as a reference, for example Taskbar Scroll Actions.

mods/taskbar-z-order-override.wh.cpp
```diff
@@ -0,0 +1,387 @@
...
+
+  return CallWindowProcW(g_state.originalTaskbarProc, hwnd, msg, wParam, lParam);
+}
+
+static void InstallTaskbarSubclass() {
+  if (g_state.originalTaskbarProc) {
+    return;
+  }
+
+  g_state.taskbarWnd = FindWindowW(L"Shell_TrayWnd", nullptr);
```
m417z: Have you considered supporting secondary taskbars too?

mods/taskbar-z-order-override.wh.cpp
```diff
@@ -191,21 +180,23 @@ static BOOL WINAPI SetWindowPos_Hook(HWND hWnd, HWND hWndInsertAfter, int X,
...
+// e.g. opening the Star* menu -> taskbar raised to band 6. 
+// Each call bumps the taskbar to the top of the specified band even 
+// if the band number is identical. 
 //
-// kDesktopBand = 1: undocumented value for normal desktop band.
+// We block that change entirely to make sure Windows doesn't  
+// mess with our own order. 
+//
+// dwBand = 1: normal desktop band?
+// dwBand = 6: normal taskbar band?
```
m417z, line 191: Regarding the question marks - there's a great documentation here:
https://blog.adeltax.com/window-z-order-in-windows-10/

PR #3661, opened 2026-03-28
New mod Windows 11 Start Menu Buttons

Salyts:
This mod replaces the default bottom row of the Windows 11 Start menu with custom buttons.
Each button has three fields:

1. **Name** — tooltip text.
2. **Icon** — a [Segoe Fluent Icons / Segoe UI Symbol](https://learn.microsoft.com/en-us/windows/apps/design/iconography/segoe-ui-symbol-font) glyph, for example `\uE7E8`. 

3. **Action** — what the button does.

### Action formats

- `cmd:explorer.exe` — runs through `cmd.exe`.
- `shell:shutdown /r` — runs console window.
- `"C:\Program Files"` — opens a folder.
- `"C:\Program Files\Google\Chrome\Application\chrome.exe"` — оpens a file or app.
- `~Downloads` — searches for a folder by name in common system locations.
- `~chrome.exe` — searches for a file by name in common system locations.
- `https://www.youtube.com/` — opens a website.
- `{ "Restart", "\uE777", "shell:shutdown /r /f /t 0" }, { "Shut down", "\uE7E8", "shell:shutdown /s /f /t 0" }` — a submenu with multiple items.

### DIFF
--- a/mods/windows-11-start-menu-buttons.wh.cpp
+++ b/mods/windows-11-start-menu-buttons.wh.cpp
@@ -268,6 +268,8 @@ static void LoadSettings() {
 
     std::lock_guard<std::mutex> lock(g_settingsMutex);
     g_settings = newSettings;
+
+    Wh_Log(L"Settings Loaded: CloseStartMenu = %d", newSettings.closeStartMenu);
 }
 
 static std::wstring GetSettingStringIndex(const wchar_t* key, int index) {
@@ -388,13 +390,19 @@ static inline void CloseStartMenuAfterClick() {
         auto dq = winrt::Windows::System::DispatcherQueue::GetForCurrentThread();
         if (dq) {
             dq.TryEnqueue(winrt::Windows::System::DispatcherQueuePriority::Low, []() {
-                SendEscapeKey();
+                std::thread([]() {
+                    Sleep(100); 
+                    SendEscapeKey();
+                }).detach();
             });
             return;
         }
     } catch (...) {}
 
-    SendEscapeKey();
+    std::thread([]() {
+        Sleep(100);
+        SendEscapeKey();
+    }).detach();
 }
 
 // -------------------- executing actions --------------------
@@ -642,7 +650,7 @@ static DWORD WINAPI ProxyWindowThread(LPVOID) {
     }
 
     UnregisterClassW(PROXY_WINDOW_CLASS, GetModuleHandleW(nullptr));
-    g_proxyWindow = NULL;
+    g_proxyWindow = NULL; 
     return 0;
 }
 
@@ -1067,6 +1075,8 @@ static void WINAPI InitOnWindowThread(PVOID param) {
 
 BOOL Wh_ModInit() {
     Wh_Log(L"Init");
+    LoadSettings();
+    BuildButtons();
 
     if (IsExplorerProcess()) {
         StartProxyThread();


mods/windows-11-start-menu-buttons.wh.cpp
```diff
@@ -0,0 +1,1131 @@
...
+            PostMessageW(g_proxyWindow, WM_CLOSE, 0, 0);
+        if (g_proxyThread) {
+            WaitForSingleObject(g_proxyThread, 1000);
+            CloseHandle(g_proxyThread);
+            g_proxyThread = nullptr;
+        }
+        return;
+    }
+
+    StopVisibilityMonitoring();
```
m417z: This, and similar UI code in `Wh_ModSettingsChanged`, doesn't seem correct, as these callbacks don't run on the UI thread. The "Windows 11 Start Menu Power Buttons", which you possibly took inspiration from, uses `CallOnWindowThread` to correctly run UI code on the UI thread.

mods/windows-11-start-menu-buttons.wh.cpp
```diff
@@ -0,0 +1,1131 @@
...
+/*
+# Windows 11 Start Menu Buttons
+
+This mod replaces the default bottom row of the Windows 11 Start menu with custom buttons.
+Each button has three fields:
+
+1. **Name** — tooltip text.
+2. **Icon** — a [Segoe Fluent Icons / Segoe UI Symbol](https://learn.microsoft.com/en-us/windows/apps/design/iconography/segoe-ui-symbol-font) glyph, for example `\uE7E8`. 
+
+3. **Action** — what the button does.
```
m417z, line 24: How about a screenshot?

mods/windows-11-start-menu-buttons.wh.cpp
```diff
@@ -0,0 +1,1131 @@
...
+// -------------------- proxy window --------------------
+
+static LRESULT CALLBACK ProxyWndProc(HWND hWnd, UINT message, WPARAM wParam, LPARAM lParam) {
+    if (message == WM_COPYDATA) {
+        const COPYDATASTRUCT* cds = reinterpret_cast<const COPYDATASTRUCT*>(lParam);
+        if (cds && cds->dwData == kCopyDataMagic && cds->lpData && cds->cbData >= sizeof(wchar_t)) {
+            const wchar_t* data = static_cast<const wchar_t*>(cds->lpData);
+            size_t len = cds->cbData / sizeof(wchar_t);
+            if (len > 0 && data[len - 1] == L'\0') {
+                ExecuteActionText(std::wstring(data));
```
m417z: The proxy window accepts `WM_COPYDATA` from any process and executes the payload as shell commands. Combined with `ChangeWindowMessageFilterEx` lowering the UIPI barrier, a low-integrity or sandboxed process can send a message to execute arbitrary commands, which is a potential security issue. Any process on the system can call `FindWindowW(PROXY_WINDOW_CLASS, PROXY_WINDOW_NAME)` and send commands.

It could be improved such that instead of sending the command, an index of the operation will be sent. Since explorer.exe has access to the settings too, it will be able to execute it, but the options will be limited to user-configured commands.

mods/windows-11-start-menu-buttons.wh.cpp
```diff
@@ -0,0 +1,1131 @@
...
+
+static void InitWithRetry(HWND hwnd) {
+    if (!g_initialized.exchange(true)) {
+        LoadSettings();
+        BuildButtons();
+    }
+    if (hwnd) g_retryHwnd = hwnd;
+
+    if (!StartVisibilityMonitoring())
+        SetTimer(hwnd, 1772058423, 100, RetryTimerProc);
```
m417z: This id is also used by the Windows 11 Start Menu Power Buttons mod. Please use a unique id to avoid conflicts if both mods are used.

mods/windows-11-start-menu-buttons.wh.cpp
Salyts: okey

mods/windows-11-start-menu-buttons.wh.cpp
m417z: BTW this is just the timestamp of the time the id was generated (e.g. via https://www.epochconverter.com/).

mods/windows-11-start-menu-buttons.wh.cpp
Salyts: added everything

mods/windows-11-start-menu-buttons.wh.cpp
```diff
@@ -634,8 +635,6 @@ static DWORD WINAPI ProxyWindowThread(LPVOID) {
         return 1;
     }
 
-    ChangeWindowMessageFilterEx(g_proxyWindow, WM_COPYDATA, MSGFLT_ALLOW, nullptr);
```
m417z: Does it work without `ChangeWindowMessageFilterEx`?

Salyts: already removed

m417z: And does the mod work for you? It doesn't seem to work for me.

Salyts: work

Salyts: What exactly is the problem?

m417z: I installed the mod to try it, the buttons appear but do nothing.

m417z: I think that the problem is that explorer.exe doesn't load the settings on init, only when settings are changed.

Salyts: Try disabling "Close Start Menu after button click" - this is a bug I'll try to fix.

PR #3663, opened 2026-03-28
Change Tray Icons 1.0.0

aubymori:
This mod will change all tray icons for an application to a specific other icon.

**Example**: Steam with its pre-2011 icon
Example: *Steam with its pre-2011 icon

### DIFF
--- a/mods/change-tray-icons.wh.cpp
+++ b/mods/change-tray-icons.wh.cpp
@@ -5,7 +5,7 @@
 // @version         1.0.0
 // @author          aubymori
 // @github          https://github.com/aubymori
-// @compilerOptions -lshell32
+// @compilerOptions -lshell32 -lgdi32
 // ==/WindhawkMod==
 
 // ==WindhawkModReadme==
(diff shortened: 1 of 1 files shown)


mods/change-tray-icons.wh.cpp
```diff
@@ -0,0 +1,123 @@
...
+)
+{
+    if (g_hIcon
+    && lpData
+    && lpData->cbSize >= offsetof(NOTIFYICONDATAW, hIcon)
+    && (lpData->uFlags & NIF_ICON))
+    {
+        /* Probably best to toss the old icon out. */
+        if (lpData->hIcon)
+            DestroyIcon(lpData->hIcon);
```
m417z: I believe that `Shell_NotifyIconW` copies the icon, and so I don't think that's correct, since the program may reuse the icon or destroy it.

mods/change-tray-icons.wh.cpp
```diff
@@ -0,0 +1,123 @@
...
+        || (pBackslash && 0 == wcsicmp(pBackslash + 1, pszPath)))
+        {
+            LPCWSTR pszIconPath = Wh_GetStringSetting(L"icons[%d].icon", i);
+            g_hIcon = (HICON)LoadImageW(
+                NULL,
+                pszIconPath,
+                IMAGE_ICON,
+                GetSystemMetrics(SM_CXSMICON),
+                GetSystemMetrics(SM_CYSMICON),
+                LR_LOADFROMFILE | LR_DEFAULTCOLOR);
```
m417z, line 116: Optional suggestion: For better DPI support, I found the following to work better:
```cpp
	HICON icon = nullptr;
	LoadIconWithScaleDown(
		hinst,
		pszName,
		GetSystemMetricsForDpi(SM_CXSMICON, dpi),
		GetSystemMetricsForDpi(SM_CYSMICON, dpi),
		&icon);
```
But it doesn't support loading from file, and `LoadImageW` doesn't support scale down logic.

Still, just replacing `GetSystemMetrics` with `GetSystemMetricsForDpi` with the primary monitor DPI should be an improvement.

PR #3664, opened 2026-03-28
Added Shell Animation Disabler mod

Lockframe:


mods/shell-animation-disabler.wh.cpp
```diff
@@ -0,0 +1,201 @@
...
+    if (_wcsicmp(exeName, L"explorer.exe") == 0) g_CurrentApp = AppType::Explorer;
+    else if (_wcsicmp(exeName, L"StartMenuExperienceHost.exe") == 0) g_CurrentApp = AppType::Start;
+    else if (_wcsicmp(exeName, L"SearchHost.exe") == 0) g_CurrentApp = AppType::Search;
+    else if (_wcsicmp(exeName, L"ShellHost.exe") == 0) g_CurrentApp = AppType::ShellHost;
+    else if (_wcsicmp(exeName, L"ShellExperienceHost.exe") == 0) g_CurrentApp = AppType::ShellExperience;
+
+    LoadSettings();
+
+    if (g_CurrentApp == AppType::Explorer) {
+        RestartTargetProcesses();
```
m417z: If there's more than one explorer.exe process, each will run this. A simple fix is to also check that `FindCurrentProcessTaskbarWnd` isn't null.
You can use this implementation:
https://github.com/ramensoftware/windhawk-mods/blob/9a65e3674e7c1c8f8265f346bd561a76bce6667c/mods/icon-resource-redirect.wh.cpp#L2029-L2048

PR #3684, opened 2026-03-30
Add Smooth Scroll for Win32

crazyboyybs:
Adds smooth scrolling with spring physics to legacy Win32 applications. Only modifies legacy controls — modern WinUI windows are not affected.

### DIFF
--- a/mods/smooth-scroll-win32.wh.cpp
+++ b/mods/smooth-scroll-win32.wh.cpp
@@ -242,42 +242,25 @@ static int LVLineH(HWND h) {
 }
 
 // ---------------------------------------------------------------------------
-// Globals (declared early — needed by UIA lazy init)
-// ---------------------------------------------------------------------------
-
-static CRITICAL_SECTION g_cs;
-
-// ---------------------------------------------------------------------------
 // UI Automation (lazy init + pattern caching)
 //
-// g_uia is a process-wide singleton protected by g_cs.
-// COM is initialized per-thread; we track the result to know whether
-// to call CoUninitialize.
+// g_uia is per-thread (thread_local) because COM STA objects should not
+// be shared across apartment threads.  COM is initialized once per thread
+// and kept alive for the lifetime of the thread — CoUninitialize is not
+// called, which is acceptable:
+// https://github.com/microsoft/windows-rs/issues/1169#issuecomment-925107412
 // ---------------------------------------------------------------------------
 
-static IUIAutomation* g_uia = nullptr;
-static bool g_uiaInit = false;
-static thread_local bool g_comInitialized = false;
+static thread_local IUIAutomation* g_uia = nullptr;
+static thread_local bool g_uiaInit = false;
 
 static IUIAutomation* UIA() {
-    // Ensure COM is initialized on the calling thread.
-    if (!g_comInitialized) {
-        HRESULT hr = CoInitializeEx(nullptr, COINIT_APARTMENTTHREADED);
-        g_comInitialized = SUCCEEDED(hr);
-    }
-
-    // Protect singleton creation. CRITICAL_SECTION is re-entrant on Windows,
-    // so this is safe even if the caller already holds g_cs.
-    EnterCriticalSection(&g_cs);
-    if (!g_uiaInit) {
-        g_uiaInit = true;
-        CoCreateInstance(__uuidof(CUIAutomation), nullptr,
-                         CLSCTX_INPROC_SERVER,
-                         __uuidof(IUIAutomation), (void**)&g_uia);
-    }
-    IUIAutomation* result = g_uia;
-    LeaveCriticalSection(&g_cs);
-    return result;
+    if (g_uiaInit) return g_uia;
+    g_uiaInit = true;
+    CoInitializeEx(nullptr, COINIT_APARTMENTTHREADED);
+    CoCreateInstance(__uuidof(CUIAutomation), nullptr, CLSCTX_INPROC_SERVER,
+                     __uuidof(IUIAutomation), (void**)&g_uia);
+    return g_uia;
 }
 
 static IUIAutomationScrollPattern* GetPattern(HWND h) {
@@ -335,6 +318,7 @@ struct State {
     bool uiaOk = true;
 };
 
+static CRITICAL_SECTION g_cs;
 static std::map<HWND, State> g_st;
 static UINT g_sysLines = 3;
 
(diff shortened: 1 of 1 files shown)


mods/smooth-scroll-win32.wh.cpp
```diff
@@ -0,0 +1,652 @@
...
+        State& sh = it->second;
+        switch (sh.method) {
+
+        case Method::PixelLV: {
+            sh.accH += dH;
+            int px = (int)sh.accH;
+            if (px) {
+                sh.accH -= px;
+                LeaveCriticalSection(&g_cs);
+                SendMessageW(hw, LVM_SCROLL, (WPARAM)(-px), 0);
```
m417z, line 478: Doesn't `-px` reverse the scroll? I'd expect it to be positive.

mods/smooth-scroll-win32.wh.cpp
```diff
@@ -0,0 +1,652 @@
...
+            }
+            break;
+        }
+
+        case Method::LineScroll: {
+            sh.accH += dH;
+            int ln = (int)sh.accH;
+            if (ln) {
+                sh.accH -= ln;
+                UINT cmd = (ln > 0) ? SB_LINELEFT : SB_LINERIGHT;
```
m417z, line 489: Here too, this seems to reverse the scroll. If intentional, perhaps it should default to not be reversed, and have an option to reverse it. Otherwise it won't be consistent with other apps which the mod doesn't target.

mods/smooth-scroll-win32.wh.cpp
```diff
@@ -0,0 +1,652 @@
...
+    EnterCriticalSection(&g_cs);
+    for (auto& p : g_st) {
+        if (p.second.timer && IsWindow(p.first))
+            KillTimer(p.first, TID);
+        Release(p.second);
+    }
+    g_st.clear();
+    LeaveCriticalSection(&g_cs);
+    DeleteCriticalSection(&g_cs);
+    if (g_uia) { g_uia->Release(); g_uia = nullptr; }
```
m417z: * `g_uia` is never initialized on the Windhawk thread, so it'll always be null here.
* It won't be freed for other threads.

mods/smooth-scroll-win32.wh.cpp
```diff
@@ -0,0 +1,652 @@
...
+// UI Automation (lazy init + pattern caching)
+// ---------------------------------------------------------------------------
+
+static thread_local IUIAutomation* g_uia = nullptr;
+static thread_local bool g_uiaInit = false;
+
+static IUIAutomation* UIA() {
+    if (g_uiaInit) return g_uia;
+    g_uiaInit = true;
+    CoInitializeEx(nullptr, COINIT_APARTMENTTHREADED);
```
m417z: `CoInitializeEx` might fail. Record its return value to know whether to uninitialize it.

mods/smooth-scroll-win32.wh.cpp
crazyboyybs: I negated the add in the horizontal push, which makes the timer work correctly for both axes without touching the timer code

`if (vert) s.sV.Push(add, g_cfg.springK);
else      s.sH.Push(-add, g_cfg.springK);`

Also fixed the undo in the SetTimer failure path (+= add instead of -= add)

mods/smooth-scroll-win32.wh.cpp
crazyboyybs: g_uia and g_uiaInit are now regular global objects (not thread_local). Initialization is protected by g_cs (which is reentrant on Windows). The Release in Wh_ModUninit now effectively releases the object. Only g_comInitialized remains thread_local because CoInitializeEx is per-threaded.

mods/smooth-scroll-win32.wh.cpp
crazyboyybs: Now the HRESULT is stored and only marks g_comInitialized = true if SUCCEEDED(hr)

mods/smooth-scroll-win32.wh.cpp
crazyboyybs: This has been fixed too

mods/smooth-scroll-win32.wh.cpp
```diff
@@ -241,20 +241,43 @@ static int LVLineH(HWND h) {
...
+// ---------------------------------------------------------------------------
+
+static CRITICAL_SECTION g_cs;
+
 // ---------------------------------------------------------------------------
 // UI Automation (lazy init + pattern caching)
+//
+// g_uia is a process-wide singleton protected by g_cs.
+// COM is initialized per-thread; we track the result to know whether
+// to call CoUninitialize.
```
m417z: But it's not called anywhere :-)
I think we can keep it as is, as the other alternative is to initialize and uninitialize it for each operation, and it seems that having it always initialized is acceptable:
https://github.com/microsoft/windows-rs/issues/1169#issuecomment-925107412

So let's just fix the comment.

mods/smooth-scroll-win32.wh.cpp
```diff
@@ -241,20 +241,43 @@ static int LVLineH(HWND h) {
...
 // ---------------------------------------------------------------------------
 
-static thread_local IUIAutomation* g_uia = nullptr;
-static thread_local bool g_uiaInit = false;
+static IUIAutomation* g_uia = nullptr;
```
m417z: I don't think sharing an object like this between STA threads is safe. Moreover, Microsoft seems to recommend running UI Automation in a separate MTA thread:
https://learn.microsoft.com/en-us/windows/win32/winauto/uiauto-threading#:~:text=Related%20topics-,UI%20Automation%20and%20the%20UI%20Thread,handlers%20on%20the%20same%20thread.

That's might require some non-trivial changes to the mod. What do you think?

mods/smooth-scroll-win32.wh.cpp
crazyboyybs: I've fixed the comment

mods/smooth-scroll-win32.wh.cpp
crazyboyybs: You're right, sharing an STA COM object across threads isn't safe. I've reverted `g_uia` and `g_uiaInit` back to `thread_local` so each thread creates and owns its own `IUIAutomation` instance, keeping everything within the same apartment.

For cleanup, the `ScrollPattern` pointers cached in `g_st` are explicitly released in `Wh_ModUninit`. The per-thread `IUIAutomation` instances aren't released there (can't access `thread_local` from the windhawk thread), but they live until the process exits, which is acceptable, like the COM guidance you linked.

Regarding the dedicated MTA thread approach from the Microsoft docs, I agree that's the "proper" solution, but it adds significant complexity (worker thread, message passing, synchronization) for what is currently a simple use case, Explorer and a few classic Win32 apps where scrolling happens on the main UI thread. I think the `thread_local` approach is a reasonable tradeoff for now.

Do you think this is acceptable for v1, or would you prefer the MTA thread approach?

PR #3691, opened 2026-03-30
Add systemtray-only-show mod

roflsunriz:
## New mod: System Tray Only (systemtray-only-show)

Hides all taskbar elements except the system tray (notification area) and displays it as a small floating window in the bottom-right corner of the screen.

### Motivation

When using a third-party launcher (e.g. ObjectDock) as a taskbar replacement and hiding the Windows taskbar, apps pinned to the taskbar or Start menu can be replaced by the launcher itself — but system tray (notification area) resident apps become inaccessible.

The practical workaround is either keeping the taskbar always visible (which defeats the purpose of the launcher setup) or tediously toggling its visibility whenever tray access is needed.

This mod was developed to fill that gap: it strips the taskbar down to just the system tray and presents it as a minimal floating overlay, letting you maintain a clean launcher-based desktop while keeping system tray apps fully accessible.

### How it works

The mod activates automatically on load — no third-party dock or external trigger is required. It hooks into TrayUI::WndProc and TaskbarHost::FrameHeight, removes the AppBar space reservation, and keeps only the system tray visible as a floating overlay.

### Features

- Auto-hide with configurable delay (mouse-leave → hide, hover bottom-right corner → show)
- Optional clock display
- Optional Show Desktop button
- Configurable background opacity (0–100)

### Requirements

- Windows 11 25H2 (build 26200)

### Notes

Maximized windows use the full screen height because AppBar space reservation is removed while the mod is active.

### Screenshot
スクリーンショット 2026-03-31 102949

### DIFF
--- a/systemtray-only-show.wh.cpp
+++ b/systemtray-only-show.wh.cpp
@@ -228,6 +228,120 @@ FrameworkElement FindDescendantByName(FrameworkElement element, PCWSTR name) {
     return nullptr;
 }
 
+FrameworkElement FindDescendantByClassName(FrameworkElement element,
+                                           PCWSTR className) {
+    int childrenCount = Media::VisualTreeHelper::GetChildrenCount(element);
+
+    for (int i = 0; i < childrenCount; i++) {
+        auto child = Media::VisualTreeHelper::GetChild(element, i)
+                         .try_as<FrameworkElement>();
+        if (!child) {
+            continue;
+        }
+        if (winrt::get_class_name(child) == className) {
+            return child;
+        }
+        auto result = FindDescendantByClassName(child, className);
+        if (result) {
+            return result;
+        }
+    }
+
+    return nullptr;
+}
+
+void LogElementIdentity(PCWSTR label, FrameworkElement element) {
+    if (element) {
+        Wh_Log(L"%s: name='%s' class='%s'", label, element.Name().c_str(),
+               winrt::get_class_name(element).c_str());
+    } else {
+        Wh_Log(L"%s: not found", label);
+    }
+}
+
+FrameworkElement FindTrayFrameElement(FrameworkElement xamlRootContent,
+                                      bool logDiagnostics) {
+    if (!xamlRootContent) {
+        if (logDiagnostics) {
+            Wh_Log(L"Tray frame lookup skipped: XamlRoot content unavailable");
+        }
+        return nullptr;
+    }
+
+    FrameworkElement systemTrayFrameGrid =
+        FindDescendantByName(xamlRootContent, L"SystemTrayFrameGrid");
+    if (systemTrayFrameGrid) {
+        return systemTrayFrameGrid;
+    }
+
+    FrameworkElement systemTrayFrame =
+        FindDescendantByClassName(xamlRootContent, L"System Tray.SystemTrayFrame");
+    if (!systemTrayFrame) {
+        systemTrayFrame =
+            FindDescendantByClassName(xamlRootContent, L"System Tray.System TrayFrame");
+    }
+
+    if (systemTrayFrame) {
+        FrameworkElement fallbackGrid =
+            FindDescendantByClassName(systemTrayFrame,
+                                      L"Windows.UI.Xaml.Controls.Grid");
+        if (fallbackGrid) {
+            if (logDiagnostics) {
+                Wh_Log(L"Using fallback tray grid under SystemTrayFrame");
+                LogElementIdentity(L"SystemTrayFrame", systemTrayFrame);
+                LogElementIdentity(L"FallbackTrayGrid", fallbackGrid);
+            }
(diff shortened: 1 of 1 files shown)


mods/systemtray-only-show.wh.cpp
```diff
@@ -0,0 +1,1574 @@
+// ==WindhawkMod==
+// @id              systemtray-only-show
+// @name            System Tray Only
+// @name:ja-JP      システムトレイのみ表示
+// @description     Hides everything except the system tray, turning the taskbar into a small floating tray in the bottom-right corner. Works standalone without any third-party dock. (Windows 11 25H2)
+// @description:ja-JP システムトレイ以外を非表示にし、タスクバーを右下角の小さなフローティングトレイに変換します。サードパーティのドックなしで単独動作します。（Windows 11 25H2）
+// @version         0.14
+// @author          roflsunriz
+// @github          https://github.com/roflsunriz
+// @license         MIT
```
m417z: The mod contains a large amount of code published under The GNU General Public License v3.0.

mods/systemtray-only-show.wh.cpp
```diff
@@ -0,0 +1,1574 @@
...
+/*
+# System Tray Only
+
+Hides all taskbar elements except the system tray (notification area) and
+displays it as a small floating window in the bottom-right corner of the screen.
+
+The mod activates automatically on load — no third-party dock or external
+trigger is required. It removes the AppBar space reservation so that maximized
+windows use the full screen, and keeps only the system tray visible as a
+small floating overlay.
```
m417z: How about adding a screenshot?

mods/systemtray-only-show.wh.cpp
```diff
@@ -0,0 +1,1574 @@
...
+        Wh_Log(L"Taskbar view module not loaded yet");
+
+        HMODULE kernelBaseModule = GetModuleHandle(L"kernelbase.dll");
+        auto pKernelBaseLoadLibraryExW =
+            (decltype(&LoadLibraryExW))GetProcAddress(kernelBaseModule,
+                                                      "LoadLibraryExW");
+        WindhawkUtils::SetFunctionHook(pKernelBaseLoadLibraryExW,
+                                       LoadLibraryExW_Hook,
+                                       &LoadLibraryExW_Original);
+    }
```
m417z: You don't seem to be hooking Taskbar.View.dll symbols, so this code and the whole "Taskbar.View.dll load detection" region seems to be unnecessary.

mods/systemtray-only-show.wh.cpp
```diff
@@ -0,0 +1,1574 @@
...
+
+    if (!unloading && !g_settings.showClock) {
+        FrameworkElement clockElement = nullptr;
+        EnumChildElements(systemTrayFrameGrid, [&](FrameworkElement c) {
+            auto cn = winrt::get_class_name(c);
+            if (cn == L"SystemTray.OmniButton") {
+                auto automationId =
+                    Automation::AutomationProperties::GetAutomationId(c);
+                if (automationId == L"SystemTrayClockButton" ||
+                    c.Name() == L"ClockButton") {
```
m417z: FYI this doesn't work for me.

Image

mods/systemtray-only-show.wh.cpp
roflsunriz: Am I can attach screenshot here? anyway here it is:
スクリーンショット 2026-03-31 102949

mods/systemtray-only-show.wh.cpp
roflsunriz: Thanks. The XAML tree in your screenshot still contains `SystemTrayFrameGrid`, so this may be an initialization-timing issue or a build-specific taskbar difference rather than a completely different element tree.

This mod is currently targeted at Windows 11 25H2 build 26200. Could you share:
- your exact Windows build number
- whether it fails only after Explorer/Windhawk auto-load at startup, or also after manually re-enabling the mod
- whether taskbar alignment / multiple monitors / tablet mode-like settings are involved
- if possible, a full XAML tree screenshot starting from `taskbarframe`
- relevant Windhawk log lines around `Wh_ModAfterInit`, `Transform complete`, and `SystemTrayFrameGrid not found`

If your build differs, I can either gate the mod more strictly or add a fallback path for that tree variant.
I tested this mod on 26200.7623 and 26200.8037 and both works fine.

mods/systemtray-only-show.wh.cpp
roflsunriz: I'm going to delete unnecessary code block.

mods/systemtray-only-show.wh.cpp
roflsunriz: I'm going to change "MIT LICENCE" to "GPL-3.0-or-later".

PR #3695, opened 2026-03-30
Add Auto Titlebar Colors Mod 

Louis047:
Added `auto-custom-titlebar-colors.cpp` mod which combines the following mods:
- Auto Dark Titlebars by Asteski
- Windows 11 Custom Titlebar Colors by Th3Fanbus

Supports both HEX and RGB values.

### DIFF
--- a/mods/auto-custom-titlebar-colors.wh.cpp
+++ b/mods/auto-custom-titlebar-colors.wh.cpp
@@ -4,7 +4,7 @@
 // @description     Auto-switches titlebar dark/light mode with the Windows theme, with separate custom colours for active/inactive windows in both modes
 // @version         1.0.0
 // @author          Lone
-// @github			https://github.com/Louis047
+// @github          https://github.com/Louis047
 // @include         *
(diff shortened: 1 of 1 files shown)


mods/auto-custom-titlebar-colors.wh.cpp
```diff
@@ -0,0 +1,551 @@
...
+// -----------------------------------------------------------------------------
+
+BOOL Wh_ModInit()
+{
+    Wh_Log(L"=== Auto Dark Titlebar with Custom Colours - Init [PID %d] ===",
+        GetCurrentProcessId());
+
+    if (IsProcessExcluded()) {
+        Wh_Log(L"Process excluded - skipping hooks");
+        return TRUE;
```
m417z: Why not use `@exclude` like you did with `devenv.exe`?

Louis047: my bad, didn't notice it, will change it soon 👍

PR #3637, opened 2026-03-24
Add Translucent Flyouts Controller mod (Windhawk settings bridge for TranslucentFlyouts)

GID0317:
This mod adds a Windhawk-based settings UI for TranslucentFlyouts. It lets users configure Global, DropDown, Menu, and Tooltip behavior from Windhawk instead of editing registry values manually.

This mod writes configuration values to the TranslucentFlyouts registry path because that is how TranslucentFlyouts is designed to load its settings. It does not patch unrelated Windows features or modify system behavior outside of TranslucentFlyouts configuration. TranslucentFlyouts must already be installed and running.

This build is intended for Windows 11 22H2 or later and has been tested on Windows 11 25H2.

Features included:
- Full settings control for Global, DropDown, Menu, and Tooltip categories
- Category reset actions (Global, DropDown, Menu, Tooltip, or All)
- Fast apply via setting change
- Optional hard reload mode (detach/attach) when needed

References:
- TranslucentFlyouts by @ALTaleX531
- TranslucentFlyoutsConfig by @Satanarious

<p align="center">
  <table>
    <tr>
      <td>
        
      </td>
      <td>
        
      </td>
    </tr>
  </table>
</p>

PR #3697, opened 2026-03-30
[translucent-flyouts-controller] Compatibility for the Translucent Flyouts Config app and support for block lists in Advanced Functions controls

GID0317:
This PR adds compatibility behavior for the Translucent Flyouts Config app and introduces Advanced Functions block list controls.
#3637

### DIFF
--- a/mods/translucent-flyouts-controller.wh.cpp
+++ b/mods/translucent-flyouts-controller.wh.cpp
@@ -657,24 +657,21 @@
   $name: Controller
 
 - advancedFunctions:
-  - enabled: false
-    $name: Enable Advanced Functions
-    $description: Turn on advanced process block/disabled controls. These options are for advanced users only
   - processBlockList: ""
     $name: Process Block List
-    $description: Advanced users only. Applied only when Enable Advanced Functions is on. Comma, semicolon, or newline separated process names to block TranslucentFlyouts loading (for example explorer.exe). This can require restarting Translucent Flyouts or Windows and may cause high CPU usage if misconfigured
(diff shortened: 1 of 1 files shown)


mods/translucent-flyouts-controller.wh.cpp
```diff
@@ -1548,12 +1775,27 @@ static void LoadSettings()
...
+    g_settings.tooltipProcessDisabledList = ParseProcessListSettingWithLegacy(L"advancedFunctions.tooltipProcessDisabledList", L"controller.tooltipProcessDisabledList");
+    g_settings.dropDownProcessDisabledList = ParseProcessListSettingWithLegacy(L"advancedFunctions.dropDownProcessDisabledList", L"controller.dropDownProcessDisabledList");
+    const bool advancedFunctionsExplicitlyEnabled = (Wh_GetIntSetting(L"advancedFunctions.enabled") != 0);
+    const bool advancedFunctionsHasEntries =
+      !g_settings.processBlockList.empty() ||
+      !g_settings.processDisabledList.empty() ||
+      !g_settings.menuProcessDisabledList.empty() ||
+      !g_settings.tooltipProcessDisabledList.empty() ||
+      !g_settings.dropDownProcessDisabledList.empty();
+    g_settings.advancedFunctionsEnabled = advancedFunctionsExplicitlyEnabled || advancedFunctionsHasEntries;
```
m417z: The setting descriptions all say "Applied only when Enable Advanced Functions is on", but the code auto-enables if any list has entries. I think it'd be clearer to just remove the "enable" option and simply keep the warning that these options are for advanced usage.

PR #3706, opened 2026-03-31
File Explorer Toolbar Position Fix v1.0

tetawaves:
Fixes issue with explorer toolbars not setting the position correctly

### DIFF
--- a/mods/fix-file-explorer-toolbar-position.wh.cpp
+++ b/mods/fix-file-explorer-toolbar-position.wh.cpp
@@ -12,7 +12,15 @@
 // ==WindhawkModReadme==
 /*
 # File Explorer Toolbar Position Fix
-Fixes issue with explorer toolbars not setting the position correctly when opening explorer.
+This mod fixes an issue with file explorer toolbars not setting the position correctly when opening explorer.
+
+**Before**  
+After opening explorer, all of the toolbars will be forced to be on a new line, ignoring the previous position and style.
+![Before](https://i.imgur.com/v8bj787.png)
+
+**After**  
+Now the toolbars will respect the previous position and style when opening explorer.
+![After](https://i.imgur.com/zGqzAfz.png)
 */
 // ==/WindhawkModReadme==
 


mods/fix-file-explorer-toolbar-position.wh.cpp
```diff
@@ -0,0 +1,90 @@
...
+// @author          xalejandro
+// @github          https://github.com/tetawaves
+// @include         explorer.exe
+// @architecture    x86-64
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# File Explorer Toolbar Position Fix
+Fixes issue with explorer toolbars not setting the position correctly when opening explorer.
```
m417z: How about adding before/after screenshots to clarify what the mod does?

kfh83: i don't think theres a good way to exemplify the fix in screenshots, the best example of this mod's changes is if you have the links toolbar next to classicexplorer, it would reset its position to the bottom-most when explorer is reopened:
image

m417z: With or without screenshots, I think it'd be better to add more information to the readme.
Who might need this mod? Is it relevant for the stock Win11? Stock Win10? Is it only useful for users of a tool such as Classic Explorer? What issue does it solve? Etc.

kfh83: I agree.

tetawaves: done

m417z: According to the screenshot, the mod is designed to work with Classic Explorer. Can you please mention it? I think it's also worth mentioning in the description, perhaps even rename the mod to "Classic Explorer Toolbar Position Fix", since it's not really relevant for most File Explorer users.

tetawaves: I'm pretty sure the issue happens with all toolbars, unless you meant another thing with "Classic Explorer".

aubymori: > I'm pretty sure the issue happens with all toolbars, unless you meant another thing with "Classic Explorer".

Yeah, it should be the same thing with any non-Classic Explorer toolbar. It is primarily useful for Classic Explorer but can be applied to other toolbars.

PR #3709, opened 2026-04-01
Taskbar-reorder mod, but works with left-click

Xenalin:


Xenalin:
I needed it mainly for my own purposes, but decided to share it in case it helped anyone else. It could be made into an option of the original if desired. The experience is about as buggy as the original, occasionally requiring some "finesse". 

Overcoming having to override the native win11 behavior in order to use the LMB was the main accomplishment here.

PR #3717, opened 2026-04-01
New mod Lock program to Monitor

EveryVerseDOTca:
First upload


EveryVerseDOTca:
Ok thanks, ya it was all with AI coding, I have basic understanding, but probably not enough to not be guided through a little to correct it.

I just found your mods and my mod is very helpful to me so I wanted to share if it would be for others. No big deal. 

Thanks again :)

PR #3728, opened 2026-04-02
Fixes for "Taskbar Music Lounge" better TB autohide & code cleanup

Chaython:
### 🛠️ Fork Changelog (v4.0.3)
* **Fixed:** Widget no longer leaves a visible "lip" when the taskbar auto-hides.
* **Fixed:** Resolved visibility loop bugs causing flickering.
* **Fixed:** Added DPI scaling support for hidden taskbars (4K monitor fix).
* **Fixed:** Removed `<vector>` and `<atomic>` as they were called but never used.

Switched out Hashah2311 as his github page is no longer existing at the URL

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,7 +1,7 @@
 ---
 name: Bug report
 about: Report a bug in a Windhawk mod
-title: ''
+title: '[Mod Name] Bug description'
 labels: mod-bug
 assignees: ''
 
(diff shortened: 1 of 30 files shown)


Chaython:
> Hello, I ask my question here because I don't know where else ... I made some changes on the Taskbar Music Lounge. Here : #3701 Do you think that it is possible to make it as a v5 of the original module ? Has the author dont't looks like to work on it anymore. Can I update this mod ? After your changes, or it is better to have a separated one ? Thanks

Mine is fixes yours is functionality.
I think best practice would be making the multiple album art view a setting rather than enforced. Rather than having multiple branches / things of the same/similar functionality

PR #3767, opened 2026-04-10
add Discord Transparency Resize Fix

Gameknight963:
Adds a mod that fixes Discord being non-resizable when using Vencord's window transparency + native titlebar option, by setting WS_THICKFRAME on Discord's window.

## Changelog

* Initial release

## Mod authorship

This mod was created by:

- - [x] Manually by the submitter (with AI assistance)
- - [ ] Claude Code
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/resizable-discord.wh.cpp
+++ b/mods/resizable-discord.wh.cpp
@@ -2,7 +2,7 @@
 // @id              resizable-discord
 // @name            Discord Transparency Resize Fix
 // @description     Forces discord to be resizable with vencord window transparency
-// @version         1.0.0
+// @version         1.0.1
 // @author          gameknight963
 // @include         discord.exe
 // @github          https://github.com/Gameknight963
@@ -12,7 +12,7 @@
 /*
 This mod forces resizing on Discord when using window transparency + native titlebar.
 
-How it works: It enables WS_THICKFRAME for proper resizing behavior,
+How it works: It enables WS_THICKFRAME by hooking ,
 which Discord disables because it uses a custom titlebar, and Electron handles
 the resize logic through software in this custom titlebar mode.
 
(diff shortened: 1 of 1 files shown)


mods/resizable-discord.wh.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+    LONG_PTR style = GetWindowLongPtr(hWnd, GWL_STYLE);
+    SetWindowLongPtr(hWnd, GWL_STYLE, style | WS_THICKFRAME);
+}
+
+BOOL CALLBACK EnumWindowsProc(HWND hWnd, LPARAM) { HookWindow(hWnd); return TRUE; }
+
+BOOL Wh_ModInit()
+{
+    EnumWindows(EnumWindowsProc, 0);
```
m417z, line 51: * `EnumWindows` modifies windows across all processes. `EnumWindows` enumerates every top-level window on the desktop, not just Discord's. Since `Chrome_WidgetWin_1` is the class name for all Chromium-based apps (Chrome, Edge, Slack, VS Code, etc.), `HookWindow` will set `WS_THICKFRAME` on windows belonging to other applications. The standard pattern is to filter by process ID:

    ```cpp
    DWORD pid;
    GetWindowThreadProcessId(hWnd, &pid);
    if (pid != GetCurrentProcessId()) return;
    ````
* Fix only applies once at initialization. The mod calls `EnumWindows` in `Wh_ModInit()` and never again. If Discord creates new windows after the mod loads, they won't get the fix. The mod should hook `CreateWindowExW` to intercept window creation, or implement another way to handle newly created windows.

Gameknight963: thanks for the feedback! should be fixed now, not sure how I didn't catch the initialization issue. I think it's because discord doesn't destroy it's window when you close it. It only breaks if it fully relaunches which I didn't test for some reason

PR #3776, opened 2026-04-11
Update Center New Windows

crazyboyybs:
## Changelog

* SetWindowPos grace period — after a window is first centred, a 2-second grace period blocks any SetWindowPos call that tries to move it elsewhere. This fixes a known issue with Firefox-based browsers (Firefox, Floorp, Zen, LibreWolf and others) that restore their saved session position via SetWindowPos ~700 ms after ShowWindow, overriding the centred position.
* Vertical offset setting — shifts the centred position up or down by a fixed number of pixels. Useful for third-party docks or status bars that are not reflected in the system work area.
* DPI-unaware app fix — windows and dialogs that belong to processes using system scaling (e.g. Audacity, older Win32 apps) are now correctly centred at all scale levels. The previous calculation mixed physical monitor coordinates with virtual window coordinates, placing windows above and to the left of true center.

### DIFF
--- a/mods/center-new-windows.wh.cpp
+++ b/mods/center-new-windows.wh.cpp
@@ -1,0 +1,1154 @@
+// ==WindhawkMod==
+// @id              center-new-windows
+// @name            Center New Windows
+// @name:pt-BR      Centralizar Novas Janelas Automaticamente
+// @name:es         Centrar Nuevas Ventanas Automáticamente
+// @description     Opens all new windows centered on the screen, including UWP and WinUI apps
+// @description:pt-BR Abre todas as novas janelas centralizadas na tela, incluindo apps UWP e WinUI
+// @description:es  Abre todas las ventanas nuevas centradas en la pantalla, incluyendo apps UWP y WinUI
+// @version         1.0.0
+// @author          crazyboyybs
+// @github          https://github.com/crazyboyybs
+// @include         *
+// @compilerOptions -luser32 -ldwmapi
+// @license         MIT
+// ==/WindhawkMod==
+
+// ==WindhawkModSettings==
+/*
+- fadeTrickEnabled: false
+  $name: Enable Fade Trick
+  $name:pt-BR: Ativar Fade Trick
+  $name:es: Activar Fade Trick
+  $description: >
+    When enabled, the window is made transparent before being shown, then
+    repositioned while invisible, becoming visible only after the delay below.
+    Use this if an app overrides the centered position after ShowWindow,
+    causing a visible jump.
+  $description:pt-BR: >
+    Quando ativado, a janela é tornada transparente antes de ser exibida e
+    reposicionada enquanto invisível, tornando-se visível apenas após o
+    atraso abaixo. Use isso se um aplicativo sobrescreve a posição
+    centralizada após o ShowWindow, causando um salto visível.
+  $description:es: >
+    Cuando está activado, la ventana se vuelve transparente antes de
+    mostrarse y se reposiciona mientras es invisible, haciéndose visible
+    solo después del retraso indicado. Úsalo si una aplicación anula la
+    posición centrada después de ShowWindow, causando un salto visible.
+- fadeDelayMs: 25
+  $name: Fade delay (ms)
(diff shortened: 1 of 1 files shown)


mods/center-new-windows.wh.cpp
```diff
@@ -785,105 +920,108 @@ static bool IsFrameMaximized(HWND frame) {
...
                     }
                 }
             }
-        } else if (!(uFlags & SWP_NOMOVE)) {
+        }
+    }
+
+    // ── Regular window grace period (all processes) ──────────────────
+    if (!(uFlags & SWP_NOMOVE)) {
+        RegTrackedWindow* rt = RegFindWindow(hWnd);
```
m417z: This is not thread-safe. You're already using `SetProp` with value `1`, perhaps set a timestamp instead, and check here whether `REG_GRACE_MS` passed. If not, calculate and use `centeredX`, `centeredY` (which may also change depending on the new size). WDYT, will it work? Alternatively, if you prefer to avoid moving the window for simplicity, you can just set `SWP_NOMOVE`.

mods/center-new-windows.wh.cpp
```diff
@@ -752,14 +845,46 @@ static void CenterWindow(HWND hwnd) {
...
+    const int waLeft   = static_cast<int>(originX + (mi.rcWork.left   - originX) / scaleX);
+    const int waTop    = static_cast<int>(originY + (mi.rcWork.top    - originY) / scaleY);
+    const int waRight  = static_cast<int>(originX + (mi.rcWork.right  - originX) / scaleX);
+    const int waBottom = static_cast<int>(originY + (mi.rcWork.bottom - originY) / scaleY);
+
+    int userOffsetY;
+    {
+        AcquireSRWLockShared(&g_srw);
+        userOffsetY = g_settings.centerOffsetY;
+        ReleaseSRWLockShared(&g_srw);
```
m417z: Instead of using a lock for just one variable, you can define it as `std::atomic<int>` instead of `int`, which both more efficient and more convenient.

mods/center-new-windows.wh.cpp
crazyboyybs: Implemented using SetProp/GetProp with a timestamp stored directly on the HWND, as you suggested. The centered position is recomputed on each blocked call by delegating to CenterWindow() (which already handles DPI scaling, shadow offsets, and monitor detection) and then passing SWP_NOMOVE to block the app's move while still allowing size/z-order changes.

mods/center-new-windows.wh.cpp
crazyboyybs: Done, centerOffsetY is now a std::atomic<int> outside the Settings struct. All read sites use .load(memory_order_relaxed) and LoadSettings uses .store(..., relaxed). The SRWLOCK is still used for the rest of Settings since those fields (excludeClasses, excludeProcesses) require coordinated reads of the full struct

mods/center-new-windows.wh.cpp
```diff
@@ -1015,13 +984,33 @@ BOOL WINAPI HookedSetWindowPos(HWND hWnd, HWND hWndInsertAfter,
...
+            DWORD storedTick = static_cast<DWORD>(reinterpret_cast<ULONG_PTR>(prop));
+            DWORD now        = static_cast<DWORD>(GetTickCount64());
+            if (now - storedTick < static_cast<DWORD>(REG_GRACE_MS)) {
+                // Grace period still active — call CenterWindow which already
+                // handles DPI scaling, shadow offsets, and monitor detection
+                // correctly. We suppress our own hook during this call so it
+                // doesn't recurse, then immediately return the original call
+                // with SWP_NOMOVE to lock the position.
+                g_inHook = true;
+                CenterWindow(hWnd);
```
m417z, line 1007: In case `pOrigSetWindowPos` below resizes the window, perhaps you'd want to call it first, and then call `CenterWindow`?

crazyboyybs: Good catch. Fixed, pOrigSetWindowPos with SWP_NOMOVE is now called first so any resize takes effect, then CenterWindow reads the final size via GetWindowRect and repositions correctly

PR #3781, opened 2026-04-11
Add Accent Highlight Color mod

FireBlade211:
# Accent Highlight Color
Makes the highlight color use your system accent color in Win32 apps instead of the
default blue.

**Before**:  
Notepad with the text "test" selected in blue

**After**:  
Notepad with the text "test" selected in the current accent color

## Changelog

* Changelog item 1...
* Changelog item 2...

## Mod authorship

This mod was created by:

- - [ ] Claude Code
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/accent-hl-color.wh.cpp
+++ b/mods/accent-hl-color.wh.cpp
@@ -198,7 +198,8 @@ void UpdateAccentCache()
 HBRUSH WINAPI GetSysColorBrush_Hook(int nIndex)
 {
     Wh_Log(L"System brush hook triggered");
-    UpdateAccentCache();
+    if (nIndex == COLOR_HIGHLIGHT || nIndex == COLOR_HIGHLIGHTTEXT)
+        UpdateAccentCache();
 
     if (nIndex == COLOR_HIGHLIGHT)
     {
--- a/mods/accent-hl-color.wh.cpp
+++ b/mods/accent-hl-color.wh.cpp
@@ -25,6 +25,7 @@ default blue.
 
 #include <dwmapi.h>
 #include <cmath>
+#include <atomic>
 
 using GetSysColor_t = decltype(&GetSysColor);
 using GetSysColorBrush_t = decltype(&GetSysColorBrush);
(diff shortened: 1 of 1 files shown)


mods/accent-hl-color.wh.cpp
```diff
@@ -0,0 +1,227 @@
...
+    {
+        DeleteObject(g_highlightBrush);
+        g_highlightBrush = NULL;
+    }
+
+    if (g_highlightTextBrush)
+    {
+        DeleteObject(g_highlightTextBrush);
+        g_highlightTextBrush = NULL;
+    }
```
m417z: `UpdateAccentCache()` deletes `g_highlightBrush` / `g_highlightTextBrush` when the accent color changes. But another thread may have just received the old `HBRUSH` from `GetSysColorBrush_Hook` and is still using it (e.g., mid-`FillRect`). The stock `GetSysColorBrush` returns system-owned brushes that callers expect to remain valid.

One approach: never delete old brushes during the mod's lifetime - only delete them in `Wh_ModUninit`. When the accent changes, create a new brush and atomically swap the pointer (you can use `std::atomic<HBRUSH>`).

mods/accent-hl-color.wh.cpp
```diff
@@ -0,0 +1,227 @@
...
+}
+
+HBRUSH WINAPI GetSysColorBrush_Hook(int nIndex)
+{
+    Wh_Log(L"System brush hook triggered");
+    UpdateAccentCache();
```
m417z: It can be optimized to called only for a matching `nIndex`. Also, `DwmGetColorizationColor` can still be called quite often, I think it's worth improving it, e.g. by calling it at most once a second and caching the last result.

mods/accent-hl-color.wh.cpp
```diff
@@ -0,0 +1,231 @@
...
+    g_cachedText.store(
+        GetTextColorForAccent(argb),
+        std::memory_order_release
+    );
+}
+
+HBRUSH WINAPI GetSysColorBrush_Hook(int nIndex)
+{
+    Wh_Log(L"System brush hook triggered");
+    UpdateAccentCache();
```
m417z: Like in the other hook:
```suggestion
    if (nIndex == COLOR_HIGHLIGHT || nIndex == COLOR_HIGHLIGHTTEXT)
        UpdateAccentCache();
```

PR #3786, opened 2026-04-13
Update Better Dialogs mod to v1.1

FireBlade211:
# Better Dialogs v1.1
This update to Better Dialogs adds new Color and Font dialogs, as well as hooks both ANSI and Unicode variants of functions. **Note:** Thanks to [#1](https://github.com/FireBlade211/windhawk-mods/pull/1) for help with making these new dialogs.

Improved Color Dialog  
  
Improved Font Dialog 

## Changelog

* Modernized Color dialog
* Modernized Font dialog
* ANSI functions hooked
* Legacy file pickers redirected to modern file dialog
* Fixed bug where message boxes with **Warning** icon would become Task Dialogs with an **Error** icon instead of the **Warning** icon

### DIFF
--- a/mods/better-dialogs.wh.cpp
+++ b/mods/better-dialogs.wh.cpp
@@ -1607,7 +1607,6 @@ static LRESULT CALLBACK FP_WndProc(HWND hwnd, UINT msg, WPARAM wParam, LPARAM lP
         int halfW = (FP_CLIENTW-FP_PAD*2-FP_GAP)/2;
         mk(L"BUTTON", L"OK", BS_DEFPUSHBUTTON, 0, FP_PAD, FP_BTN_Y, halfW, 38, IDOK);
         mk(L"BUTTON", g_cpCancel, BS_PUSHBUTTON, 0, FP_PAD+halfW+FP_GAP, FP_BTN_Y, halfW, 38, IDCANCEL);
->>>>>>> 44ce70ca10503b6ae3ce4bb9f31df628aa70f437
         return 0; }
 
     case WM_ERASEBKGND: return TRUE;
--- a/mods/better-dialogs.wh.cpp
+++ b/mods/better-dialogs.wh.cpp
@@ -2192,7 +2192,7 @@ static LRESULT CALLBACK FP_WndProc(HWND hwnd, UINT msg, WPARAM wParam, LPARAM lP
 		return 0;
 	}
 
-	if (FLAG(d->flags, CF_ENABLEHOOK) && d->pcfHook)
+	if (d && FLAG(d->flags, CF_ENABLEHOOK) && d->pcfHook)
 		return d->pcfHook(hwnd, msg, wParam, lParam);
 
 	return DefWindowProcW(hwnd, msg, wParam, lParam);
--- a/mods/better-dialogs.wh.cpp
+++ b/mods/better-dialogs.wh.cpp
@@ -915,7 +915,6 @@ static const COLORREF g_basicColors[48] = {
 
 static int CP_GridY()    { return CP_SPEC_Y + CP_SPEC_H + 48; }
 static int CP_CustGridY(){ return CP_GridY() + CP_BROWS * CP_CIRC_STEP + 38; }
-static int CP_BtnY()     { return (CP_CustGridY() + CP_CIRC_STEP) - 24; }
 static int CP_BtnY()     { return CP_CustGridY() + 2 * CP_CIRC_STEP + 28; }
 static int CP_ClientW()  { return CP_INFO_X + 160; }
 static int CP_ClientH()  { return CP_BtnY() + 40 + CP_PAD; }
@@ -1223,7 +1222,7 @@ static LRESULT CALLBACK CP_WndProc(HWND hwnd, UINT msg, WPARAM wParam, LPARAM lP
         mk(L"BUTTON", L"OK", BS_DEFPUSHBUTTON, 0, btnX, btnY, 80, 24, IDOK);
         mk(L"BUTTON", g_cpCancel, BS_PUSHBUTTON, 0, btnX, btnY + 32, 80, 24, IDCANCEL);
         mk(L"BUTTON", L"+", BS_PUSHBUTTON, 0, CP_SPEC_X+szCust.cx+10, CP_CustGridY()-28, 24, 22, 4010);
-        int btnY = CP_BtnY(), halfW = (CP_ClientW()-CP_PAD*2-10)/2;
+        btnY = CP_BtnY(), halfW = (CP_ClientW()-CP_PAD*2-10)/2;
         mk(L"BUTTON", L"OK", BS_DEFPUSHBUTTON, 0, CP_PAD, btnY, halfW, 34, IDOK);
         mk(L"BUTTON", g_cpCancel, BS_PUSHBUTTON, 0, CP_PAD+halfW+10, btnY, halfW, 34, IDCANCEL);
         d->specPixels = (DWORD*)calloc(CP_SPEC_W * CP_SPEC_H, 4);
@@ -1350,7 +1349,6 @@ BOOL WINAPI ChooseColorW_Hook(LPCHOOSECOLORW lpcc) {
 #define FP_PREV_Y   (FP_LIST_Y + FP_LIST_H + 20)
 #define FP_PREV_H   64
 #define FP_EFF_Y    (FP_PREV_Y + FP_PREV_H + 16)
-#define FP_BTN_Y    (FP_EFF_Y)
 #define FP_BTN_Y    (FP_EFF_Y + 34)
 #define FP_CLIENTW  (FP_SIZ_X + FP_SIZ_W + FP_PAD)
 #define FP_CLIENTH  (FP_BTN_Y + 42 + FP_PAD)
--- a/mods/better-dialogs.wh.cpp
+++ b/mods/better-dialogs.wh.cpp
@@ -1289,12 +1289,34 @@ static LRESULT CALLBACK CP_WndProc(HWND hwnd, UINT msg, WPARAM wParam, LPARAM lP
                     wsprintfW(b,L"%d",GetGValue(d->result));SetDlgItemTextW(hwnd,IDC_CP_G,b);
                     wsprintfW(b,L"%d",GetBValue(d->result));SetDlgItemTextW(hwnd,IDC_CP_B,b);
                     d->updating=FALSE;CP_RenderSpectrum(d);CP_RenderVBar(d);CP_UploadBitmaps(d);InvalidateRect(hwnd,NULL,FALSE);}}
-            else if(id>=IDC_CP_R&&id<=IDC_CP_B){int r=GetDlgItemInt(hwnd,IDC_CP_R,NULL,FALSE),g=GetDlgItemInt(hwnd,IDC_CP_G,NULL,FALSE),b=GetDlgItemInt(hwnd,IDC_CP_B,NULL,FALSE);
-                r=r>255?255:r;g=g>255?255:g;b=b>255?255:b;CP_RGBtoHSV(RGB(r,g,b),d->hue,d->sat,d->val);
-                d->result=CP_HSVtoRGB(d->hue,d->sat,d->val);d->updating=TRUE;WCHAR buf[16];
-                wsprintfW(buf,L"#%02X%02X%02X",GetRValue(d->result),GetGValue(d->result),GetBValue(d->result));
-                SetDlgItemTextW(hwnd,IDC_CP_HEX,buf);d->updating=FALSE;CP_RenderSpectrum(d);CP_RenderVBar(d);CP_UploadBitmaps(d);InvalidateRect(hwnd,NULL,FALSE);}}
-        return 0; }
+            else if (id>=IDC_CP_R&&id<=IDC_CP_B)
+            {
+                int r = GetDlgItemInt(hwnd, IDC_CP_R, NULL, FALSE);
+                int g = GetDlgItemInt(hwnd, IDC_CP_G, NULL, FALSE);
+                int b = GetDlgItemInt(hwnd,IDC_CP_B,NULL,FALSE);
+                r = r > 255 ? 255 : r;
+                g = g > 255 ? 255 : g;
+                b = b > 255 ? 255 : b;
+
+                CP_RGBtoHSV(RGB(r,g,b),d->hue,d->sat,d->val);
+
+                d->result=CP_HSVtoRGB(d->hue,d->sat,d->val);
+                d->updating=TRUE;
+
+                WCHAR buf[16] = L"";
+
+                wsprintfW(buf, L"#%02X%02X%02X", GetRValue(d->result), GetGValue(d->result), GetBValue(d->result));
+                SetDlgItemTextW(hwnd,IDC_CP_HEX,buf);
+
+                d->updating=FALSE;
+                CP_RenderSpectrum(d);
+                CP_RenderVBar(d);
+                CP_UploadBitmaps(d);
+                InvalidateRect(hwnd,NULL,FALSE);
+            }
+        }
+        return 0;
+    }
     case WM_DESTROY: CP_DestroyD2D(d); if(d->specPixels){free(d->specPixels);d->specPixels=nullptr;} if(d->vbarPixels){free(d->vbarPixels);d->vbarPixels=nullptr;} PostQuitMessage(0); return 0;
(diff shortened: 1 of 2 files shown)


mods/better-dialogs.wh.cpp
```diff
@@ -822,104 +877,1584 @@ LPITEMIDLIST WINAPI SHBrowseForFolderW_Hook(LPBROWSEINFOW lpbi) {
...
+                    wsprintfW(b,L"%d",GetGValue(d->result));SetDlgItemTextW(hwnd,IDC_CP_G,b);
+                    wsprintfW(b,L"%d",GetBValue(d->result));SetDlgItemTextW(hwnd,IDC_CP_B,b);
+                    d->updating=FALSE;CP_RenderSpectrum(d);CP_RenderVBar(d);CP_UploadBitmaps(d);InvalidateRect(hwnd,NULL,FALSE);}}
+            else if(id>=IDC_CP_R&&id<=IDC_CP_B){int r=GetDlgItemInt(hwnd,IDC_CP_R,NULL,FALSE),g=GetDlgItemInt(hwnd,IDC_CP_G,NULL,FALSE),b=GetDlgItemInt(hwnd,IDC_CP_B,NULL,FALSE);
+                r=r>255?255:r;g=g>255?255:g;b=b>255?255:b;CP_RGBtoHSV(RGB(r,g,b),d->hue,d->sat,d->val);
+                d->result=CP_HSVtoRGB(d->hue,d->sat,d->val);d->updating=TRUE;WCHAR buf[16];
+                wsprintfW(buf,L"#%02X%02X%02X",GetRValue(d->result),GetGValue(d->result),GetBValue(d->result));
+                SetDlgItemTextW(hwnd,IDC_CP_HEX,buf);d->updating=FALSE;CP_RenderSpectrum(d);CP_RenderVBar(d);CP_UploadBitmaps(d);InvalidateRect(hwnd,NULL,FALSE);}}
+        return 0; }
+    case WM_DESTROY: CP_DestroyD2D(d); if(d->specPixels){free(d->specPixels);d->specPixels=nullptr;} if(d->vbarPixels){free(d->vbarPixels);d->vbarPixels=nullptr;} PostQuitMessage(0); return 0;
```
m417z: Why is the code so compact, having multiple statements in the same line? That makes it difficult to read and review the code.

m417z: That line was only an example, there are many similarly packed lines.

FireBlade211: Sorry for that, [#1](https://github.com/FireBlade211/windhawk-mods/pull/1) made all the drawing code.

m417z:
Please format the code to make it more readable and more easily reviewable, either manually, or with a formatter. You can do it in Windhawk too by pressing Alt+Shift+F.

Also, here's an AI-assisted review, I filtered though items which seem to be correct:

### Correctness

*  **`PostQuitMessage(0)` in custom-dialog `WM_DESTROY` poisons the host's message loop.** Both [`CP_WndProc`](vscode-webview://1qomlt1ktdjirjto09cjqkh7skmmds6mvcliq4sdhj7pd7ngtfk4/mods/better-dialogs.wh.cpp) (line ~669 in diff) and [`FP_WndProc`](vscode-webview://1qomlt1ktdjirjto09cjqkh7skmmds6mvcliq4sdhj7pd7ngtfk4/mods/better-dialogs.wh.cpp) (line ~1047) call `PostQuitMessage(0)`. `PostQuitMessage` posts `WM_QUIT` to the **thread's** message queue, not the modal dialog's. After your nested `GetMessageW` loop returns, the `WM_QUIT` will still be in the queue (or worse, your loop may swallow it before the host's main loop sees `WM_QUIT` and shuts down). The standard pattern is to loop while `IsWindow(hwnd)` (or a `done` flag), not to call `PostQuitMessage`. As written, calling `ChooseColorW`/`ChooseFontW` will sometimes terminate the host process's message loop after the dialog closes.

*  **Duplicate `IDOK`/`IDCANCEL` (and "+") child controls.** In `CP_WndProc::WM_CREATE`, two pairs of OK/Cancel buttons are created (lines 593--594 and 598--599 in the diff), and two "+" buttons with id `4010` (lines 591 and 595). `FP_WndProc::WM_CREATE` has the same problem (lines 975--976 and 978--979). Duplicate child IDs cause `GetDlgItem`, accelerators and default-button handling to misbehave; visually you also get four buttons stacked. Looks like mid-edit dead code that wasn't deleted.

*  **`ModernFileDialog_Impl` swallows COM-init failure and returns "cancelled".** Line ~1315 in the diff: `if (FAILED(hrCo)) return FALSE;`. `CoInitialize` can fail with `RPC_E_CHANGED_MODE` on threads that are already MTA. Returning `FALSE` makes the host think the user cancelled. The wrapper's `if (!ret && GetLastError() != 0)` fallback won't fire because `GetLastError()` is unrelated to `CoInitialize` failure. Either fall back to the `_Original` here, or check the mode beforehand.

*  **Capture-strings hack opens the original dialogs offscreen on first use.** `CP_LoadStrings` and `FP_LoadStrings` call the **real** `ChooseColorW_Original` / `ChooseFontW_Original` with `CC_ENABLEHOOK`/`CF_ENABLEHOOK` and a hook proc that immediately `EndDialog`s after grabbing labels. For `ChooseFont`, this enumerates every font on the system before being closed, so the very first invocation of the modern font picker is delayed by a full font enumeration (and there's a brief flash at `(-32000, -32000)` plus z-order shuffling). Use `LoadStringW`/known comdlg32 string IDs, or just hardcode English fallback labels --- the current approach is fragile.

*  **`ChooseColorW_Hook`/`ChooseFontW_Hook` only check `*_ENABLETEMPLATE`, not `*_ENABLEHOOK`.** Apps that pass `CC_ENABLEHOOK` / `CF_ENABLEHOOK` expect their hook procedure to receive `WM_INITDIALOG`, `WM_NOTIFY` (`WM_CHOOSECOLOR_*`), etc. Your replacement window never invokes the caller's hook, so those apps will silently break. Either fall back to the original when `*_ENABLEHOOK` is set, or run the caller's hook against your replacement (the latter is impractical given the different control IDs/messages).

*  **`OFN_HIDEREADONLY` / `OFN_READONLY` semantics inverted.** Per [GetOpenFileName docs](https://learn.microsoft.com/windows/win32/api/commdlg/ns-commdlg-openfilenamew), the read-only checkbox is shown by default and hidden only when `OFN_HIDEREADONLY` is set; `OFN_READONLY` controls the *initial state*. The current code only adds the checkbox `if (lpofn->Flags & OFN_READONLY)` (line ~1401), which is backwards. It should be:

    ```
    if (!(lpofn->Flags & OFN_HIDEREADONLY))
        pCustomize->AddCheckButton(IDC_MODERN_READONLY, L"Open as read-only",
                                   !!(lpofn->Flags & OFN_READONLY));

    ```

*  **`lpfnHook` on OFN is silently ignored.** Apps using `OFN_ENABLEHOOK` get callbacks like `CDN_FOLDERCHANGE`, `CDN_SELCHANGE`, `CDN_FILEOK`, etc. The modern path doesn't invoke them at all, so apps that rely on the hook (preview panes, custom validation, etc.) will misbehave. At minimum, fall back to `_Original` when `OFN_ENABLEHOOK` is set.

*  **`MessageBoxIndirectA_Hook` `_alloca` with no length check.** `MessageBoxA_Hook` caps `lenText > 4096 || lenCap > 1024`, but `MessageBoxIndirectA_Hook` (line ~1879) calls `_alloca(lenText * sizeof(WCHAR))` directly. An attacker-controlled (or just unusually large) message text overflows the stack. Add the same length check before allocating.

*  **`GetLastError()` fallback in ANSI/W file-dialog hooks is unreliable.** `if (!ret && GetLastError() != 0) return GetOpenFileNameA_Original(lpofnA);` runs after `CoTaskMemFree` etc., any of which can clobber `LastError`. You'll randomly fall back to the legacy dialog after the user already cancelled the modern one. Use `CommDlgExtendedError()` semantics (return `FALSE` with no error → user cancelled), or have `ModernFileDialog_Impl` return a tri-state (success / cancel / error) and decide at the call site.

### Security & Safety

*  **Window-class registration leaks across mod reload.** `g_cpClassAtom` / `g_fpClassAtom` are registered with `RegisterClassExW` in process scope but never unregistered. On Windhawk reload (`Wh_ModUninit` → `Wh_ModInit`), `RegisterClassExW` will fail with `ERROR_CLASS_ALREADY_EXISTS`, but the wndproc inside the host now points into freed mod memory --- next call into the dialog crashes. Either unregister in `Wh_ModUninit` (and refuse to unload while a dialog is up), or use a unique class name suffix per init.

*  **No `Wh_ModUninit` cleanup at all** for the new state (D2D factories are per-`ColorPickerData`, but registered classes, hook installations etc. are leaked on unload --- and any modal currently up will crash when its wndproc disappears).

### Design

*  **Capturing localized strings via the original dialog is the wrong abstraction.** It assumes the original is hookable through the `CC_ENABLEHOOK`/`CF_ENABLEHOOK` ABI (which is what your hook is intercepting) --- it works only because you call `_Original` directly, but it adds a hidden cold-start cost. Hardcode English + a short translation table, or load comdlg32 string resources.

### Conventions

*  **README is duplicated** (diff lines 31--46): the ChooseColor and ChooseFont bullets appear twice, and the screenshots are inserted in the middle of the "Dialogs changed" list. Looks like a botched merge.

mods/better-dialogs.wh.cpp
```diff
@@ -92,894 +144,3067 @@ void LoadSettings() {
...
 
-// The mod is being initialized, load settings, hook functions, and do other
-// initialization stuff if required.
-BOOL Wh_ModInit() {
-    Wh_Log(L"Init");
+	if (!hwnd) return FALSE;
+	if (hwndOwner) EnableWindow(hwndOwner, FALSE);
+
+	MSG msg;
+	while (GetMessageW(&msg, NULL, 0, 0))
```
m417z: I tested it and my program seems to be stuck after the dialog is closed, possibly because this loop is never ended, because `EnableWindow(..., TRUE)` isn't called, or both. Did it work correctly in your tests?

I tried to find and suggest a localized fix, but it seems like more than a 2-lines fix.

Below is an AI-assisted suggestion with reference links you might find useful.

I also saw that clicking on Cancel and clicking on the X button behaves differently, you might want to make sure both work correctly.

<details><summary>Details</summary>
<p>

Raymond Chen's "The dialog manager" series (parts 4-7) covers this end-to-end. Here's the definitive pattern from [Part 5](https://devblogs.microsoft.com/oldnewthing/20050404-48/?p=35983):

**State struct:**

```
typedef struct DIALOGSTATE {
    BOOL fEnded;
    int  iResult;
} DIALOGSTATE;

```

**Setting the flag (replaces EndDialog):**

```
void EndManualModalDialog(HWND hdlg, int iResult)
{
    DIALOGSTATE *pds = reinterpret_cast<DIALOGSTATE*>
        (GetWindowLongPtr(hdlg, DWLP_USER));
    if (pds) {
        pds->iResult = iResult;
        pds->fEnded = TRUE;
        PostMessage(hdlg, WM_NULL, 0, 0);  // wake up GetMessage (part 6)
    }
}

```

**The modal loop:**

```
int DoModal(HWND hwnd)
{
    DIALOGSTATE ds = { 0 };
    HWND hdlg = CreateDialogParam(g_hinst, MAKEINTRESOURCE(1),
                hwnd, DlgProc, reinterpret_cast<LPARAM>(&ds));
    if (!hdlg) return -1;

    EnableWindow(hwnd, FALSE);
    MSG msg;
    msg.message = WM_NULL;
    while (!ds.fEnded && GetMessage(&msg, NULL, 0, 0)) {
        if (!IsDialogMessage(hdlg, &msg)) {
            TranslateMessage(&msg);
            DispatchMessage(&msg);
        }
    }
    if (msg.message == WM_QUIT) {
        PostQuitMessage((int)msg.wParam);  // re-post for outer loop
    }
    EnableWindow(hwnd, TRUE);
    DestroyWindow(hdlg);
    return ds.iResult;
}

```

Key details from the series:

-   **[Part 6](https://devblogs.microsoft.com/oldnewthing/20050405-46/?p=35973)**: The `PostMessage(hdlg, WM_NULL, 0, 0)` is essential --- when `fEnded` is set from inside `DispatchMessage`, the loop will naturally re-check, but if `GetMessage` is blocked waiting for a message, it needs a kick to wake up and re-evaluate the condition.
-   **[Part 3 of the Modality series](https://devblogs.microsoft.com/oldnewthing/20050218-00/?p=36413)**: `PostQuitMessage` is **wrong** for terminating modal loops --- `WM_QUIT` propagates to the outer loop and kills the application. The `WM_QUIT` re-post guard handles the case where *someone else* posts quit during the modal loop.
-   **Window destruction happens AFTER the loop exits**, not inside `WM_COMMAND`. This is exactly how `DialogBox`/`EndDialog` works --- `EndDialog` just sets a flag; `DialogBox` destroys the window after the loop.

Wine's implementation of `EndDialog` in `dlls/user32/dialog.c` confirms the same pattern --- it sets a `DF_END` flag and posts `WM_NULL` with the comment `/* unblock dialog loop */`.

So the PR's color/font pickers need to be restructured: stop calling `DestroyWindow` from inside `WM_COMMAND`, set a flag instead, and let the modal loop function (`CP_ShowDialog`/`FP_ShowDialog`) handle destruction after the loop exits.

</p>
</details>

m417z:
Other notes:

* Font picker creates duplicate OK/Cancel buttons.** In `FP_WndProc` WM\_CREATE, there are TWO pairs of OK/Cancel buttons created --- a small pair at lines 2044-2045 and a wider pair at lines 2048-2049. Both use the same IDs (IDOK/IDCANCEL). This is clearly leftover from trying two layouts. One pair needs to be removed.
* **`vDlgs`, `nFontDlgCount`, `nColorDlgCount` are not thread-safe.** These globals are read/written from arbitrary threads (since the hooked functions can be called from any thread). `std::vector::push_back` and `std::erase` are not atomic. A concurrent color picker open from two threads would corrupt `vDlgs`. At minimum, use `std::mutex` for `vDlgs` and `std::atomic<int>` for the counters.
* **`Wh_ModUninit` calls `DestroyWindow` on windows from other threads.** `DestroyWindow` fails if the window belongs to a different thread. The cleanup loop at line 3195-3196 is unreliable. For cross-thread destruction, you'd need `PostMessage(hwnd, WM_CLOSE, 0, 0)` instead (but that's async, so the unload sequence needs coordination).

FireBlade211:
Alright, I fixed the issues. I also tried to make custom hooks supported, but I gave up when it kept throwing access violations and just reverted it back to the old behavior of just using the normal Windows dialog when a hook is present.

m417z:
How did you address the `while (GetMessageW(&msg, NULL, 0, 0))` loop issue? There were no changes there.

FireBlade211:
Sorry, forgot about that. I applied the fix to the second message loop. (the first one already had it)

m417z:
Now it crashes for me with a simple color picker. How are you testing the mod?

FireBlade211:
Alright, I fixed it. Turns out an access violation was happening due to an uninitialized pointer in the window procedure.

m417z:
Did you test all other dialogs to make sure they work?

FireBlade211:
Ok turns out the **Font** dialog also had the same issue. I fixed it and now they work. Also how I'm testing it, I just built this C++ with MSVC:

```cpp
#include <windows.h>

int WINAPI wWinMain(_In_ HINSTANCE hInstance, _In_opt_ HINSTANCE hPrevInstance, _In_ LPWSTR lpCmdLine, _In_ int nCmdShow)
{
    SetThreadUILanguage(MAKELANGID(LANG_ENGLISH, SUBLANG_ENGLISH_US));

    DWORD dwDlg = MessageBoxW(NULL,
        L"Which dialog do you want to show?\n\nClick YES to show the Color dialog.\nClick NO to show the Font dialog.",
        L"Dialog Test", MB_YESNOCANCEL | MB_ICONQUESTION);

    if (dwDlg == IDYES)
    {
        CHOOSECOLORW cc = {};
        cc.lStructSize = sizeof(cc);

        ChooseColorW(&cc);
    }
    else if (dwDlg == IDNO)
    {
        CHOOSEFONTW cf = {};
        cf.lStructSize = sizeof(cf);

        ChooseFontW(&cf);
    }

    return 0;
}
```

m417z:
Form a quick test, clicking on the X button in the color dialog doesn't re-enable the owner dialog. It also feels to me that it could use more testing, but it's up to you. Let me know when it's ready.

FireBlade211:
Alright, now the window restores itself to enabled when the dialog closes.

PR #3809, opened 2026-04-15
taskbar-z-order-override 0.8

Meteony:
Add UWP fullscreen workaround for always on top mode.

### DIFF
--- a/mods/taskbar-z-order-override.wh.cpp
+++ b/mods/taskbar-z-order-override.wh.cpp
@@ -25,7 +25,7 @@ Makes the taskbar behave like a regular window - like how it was possible pre-Vi
 
 ### Always on top
 
-Keeps the taskbar above normal windows, no matter in games or in browser fullscreen. 
+Keeps the taskbar above normal windows - no matter in games or browser fullscreen. 
 
 This is mainly useful for people who use **auto-hide** and want the taskbar to stay accessible consistently.
 
@@ -77,8 +77,6 @@ enum class TaskbarZOrder {
 
 namespace {
 
-constexpr DWORD kDesktopBand = 1;
-
 using SetWindowPos_t = BOOL(WINAPI*)(HWND, HWND, int, int, int, int, UINT);
 
 using SetWindowBand_t = BOOL(WINAPI*)(HWND, HWND, DWORD);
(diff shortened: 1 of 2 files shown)


mods/taskbar-z-order-override.wh.cpp
```diff
@@ -179,24 +180,37 @@ static BOOL WINAPI SetWindowPos_Hook(HWND hWnd, HWND hWndInsertAfter, int X,
...
+    if (g_state.mode == TaskbarZOrder::Top) {
+      /* Prevent demotion to band 0; otherwise allow Top mode to
+      cooperate with UWP fullscreen. Not so easy for other modes */
+      dwBand = (dwBand == 0) ? 1 : dwBand;
+      return SetWindowBand_Original(hWnd, hwndInsertAfter, dwBand);
+    }
+
     return TRUE;
-  }
+}
```
m417z:
```suggestion
  }
```

mods/taskbar-z-order-override.wh.cpp
```diff
@@ -179,24 +180,37 @@ static BOOL WINAPI SetWindowPos_Hook(HWND hWnd, HWND hWndInsertAfter, int X,
...
                                       DWORD dwBand) {
-  // Wh_Log(L"Blocked Taskbar band change -> %lu", dwBand);
-  if (IsTrackedTaskbarWindow(hWnd)){
+  if (IsTrackedTaskbarWindow(hWnd)) {
+    //Wh_Log(L"Attempted Taskbar band change -> %lu", dwBand);
+
+    if (g_state.mode == TaskbarZOrder::Top) {
+      /* Prevent demotion to band 0; otherwise allow Top mode to
+      cooperate with UWP fullscreen. Not so easy for other modes */
+      dwBand = (dwBand == 0) ? 1 : dwBand;
```
m417z: `0` is `ZBID_DEFAULT` and `1` is `ZBID_DESKTOP` according to:
https://blog.adeltax.com/window-z-order-in-windows-10/
`ZBID_DESKTOP` is for regular windows, so aren't both the same?

Meteony: I wrote this preventatively because I only saw windows demoting taskbar back to band 1 after exiting full screen, and I thought 0 would be one level below desktop. Didn't realize both are identical.

PR #3812, opened 2026-04-15
Add Disable Windows Shortcuts Mod

Louis047:
Adds the `disable-windows-shortcuts` mod, which allows selectively disabling Windows keyboard shortcuts with individual toggles. 

### Implementation Details
This mod uses a **hybrid hooking approach** to ensure maximum compatibility with third-party apps while still being able to block stubborn hardcoded OS hotkeys:
1. **`RegisterHotKey` Hook (`explorer.exe` only):** 
   Used for standard shell shortcuts (e.g., `Win+D`, `Win+E`, `Win+S`). By returning `ERROR_HOTKEY_ALREADY_REGISTERED`, this prevents Explorer from registering the hotkeys, successfully freeing them up so third-party applications (like AutoHotkey, PowerToys, etc.) can bind to them.
   *(Note: Requires an Explorer restart to apply. A prompt is included.)*
2. **`WH_KEYBOARD_LL` Hook (`dwm.exe` only):** 
   Used strictly for hardcoded OS shortcuts that completely bypass `RegisterHotKey` (e.g., `Win+A`, `Win+Tab`, `Win+Arrows`, and the Office Hub `Ctrl+Shift+Alt+Win` shortcut). The mod runs this inside `dwm.exe` to catch these keys at a low level.
   *(Note: This hook is dynamically unloaded from `dwm.exe` if the user hasn't disabled any hardcoded keys, ensuring zero background overhead when unneeded. Hardcoded keys cannot be freed up for 3rd-party apps).*
   
## Mod authorship
This mod was created by:
- - [ ] Claude Code
- - [ ] ChatGPT
- - [ ] Gemini
- - [x] Another AI (please specify): OpenCode (Gemini 3.1 Pro Preview) Diagnosed issues manually and did thorough testing of the mod with different approaches.

### DIFF
--- a/mods/disable-windows-shortcuts.wh.cpp
+++ b/mods/disable-windows-shortcuts.wh.cpp
@@ -21,7 +21,7 @@ For this mod to successfully block window snapping (Win+Arrows) and Task View (W
 3. Under **Process inclusion list**, ensure `dwm.exe` is added (or `*` is used to include all processes)
 4. Click **Save**. Windhawk will automatically restart to apply the new settings.
 
-*Note: Changes to standard shortcuts (like Win+E) require an Explorer restart to take effect. You will be prompted automatically.*
+*Note: Changes to standard shortcuts (like Win+E) require an Explorer restart to take effect. You will be prompted automatically. If you completely disable or remove this mod from Windhawk, you must manually restart Explorer to restore those standard shortcuts.*
 
 ## Features
 - Disable any Windows key combination
--- a/mods/disable-windows-shortcuts.wh.cpp
+++ b/mods/disable-windows-shortcuts.wh.cpp
@@ -925,6 +925,21 @@ void Wh_ModUninit()
         CloseHandle(g_restartExplorerPromptThread);
         g_restartExplorerPromptThread = nullptr;
     }
+
+    if (g_isExplorer)
+    {
+        // When the mod is disabled or removed, the DLL is unloaded instantly.
+        // We use a detached PowerShell process to prompt for an Explorer restart
+        // so it survives the DLL unloading and doesn't block the Windhawk thread.
+        WCHAR commandLine[] = L"powershell.exe -WindowStyle Hidden -Command \"$wshell = New-Object -ComObject Wscript.Shell; $result = $wshell.Popup('The Disable Windows Shortcuts mod has been unloaded.\\n\\nTo fully restore standard Windows shortcuts (like Win+E), Explorer needs to be restarted.\\n\\nRestart Explorer now?', 0, 'Windhawk - Disable Windows Shortcuts', 4132); if ($result -eq 6) { Stop-Process -Name explorer -Force; Start-Process explorer }\"";
+        STARTUPINFO si = { .cb = sizeof(si) };
+        PROCESS_INFORMATION pi{};
+        if (CreateProcess(nullptr, commandLine, nullptr, nullptr, FALSE, CREATE_NO_WINDOW, nullptr, nullptr, &si, &pi))
+        {
+            CloseHandle(pi.hThread);
+            CloseHandle(pi.hProcess);
+        }
+    }
 }
 
 void Wh_ModSettingsChanged()
--- a/mods/disable-windows-shortcuts.wh.cpp
+++ b/mods/disable-windows-shortcuts.wh.cpp
@@ -663,10 +663,10 @@ void PromptForExplorerRestart()
         int button;
         if (SUCCEEDED(TaskDialogIndirect(&taskDialogConfig, &button, nullptr, nullptr)) && button == IDYES)
         {
-            WCHAR commandLine[] = L"cmd.exe /c \"taskkill /F /IM explorer.exe & start explorer\"";
+            WCHAR commandLine[] = L"cmd.exe /c \"taskkill /F /IM explorer.exe & explorer.exe\"";
             STARTUPINFO si = { .cb = sizeof(si) };
             PROCESS_INFORMATION pi{};
-            if (CreateProcess(nullptr, commandLine, nullptr, nullptr, FALSE, 0, nullptr, nullptr, &si, &pi))
+            if (CreateProcess(nullptr, commandLine, nullptr, nullptr, FALSE, CREATE_NO_WINDOW, nullptr, nullptr, &si, &pi))
             {
                 CloseHandle(pi.hThread);
                 CloseHandle(pi.hProcess);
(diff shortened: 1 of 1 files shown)


mods/disable-windows-shortcuts.wh.cpp
```diff
@@ -0,0 +1,739 @@
...
+                // Forces the OS to see an unassigned keystroke, cancelling the Start Menu pop-up
+                INPUT inputs[2] = {};
+                inputs[0].type = INPUT_KEYBOARD;
+                inputs[0].ki.wVk = 0xFF;
+                inputs[1].type = INPUT_KEYBOARD;
+                inputs[1].ki.wVk = 0xFF;
+                inputs[1].ki.dwFlags = KEYEVENTF_KEYUP;
+                SendInput(2, inputs, sizeof(INPUT));
+
+                return 1; // Suppress physical key
```
m417z, line 803: I thought that the idea is that you can reassign a hotkey. For example, with the previous mod, I could disable a system hotkey such as Win+F, and then configure another app to use it. With this mod, it's always blocked, which is less useful.

Louis047: I think this snippet is added for only disabling office hotkey and im sure rest disabled windows shortcuts can be utilized for third party apps without any conflicts. I'll still give it a test and let you know 👍

Louis047:
@m417z I completely understand the concern about reassigning hotkeys. Ideally, I would love for them to be completely freed up so other apps can claim them via `RegisterHotKey`
 
However, the reason I abandoned the previous approach (hooking RegisterHotKey in explorer.exe) is that it fundamentally fails to block a large portion of modern Windows 10/11 shortcuts. For example, Win+A (Quick Settings), Win+K (Cast), Win+Tab (Task View), and all Win+Arrow combinations (Window Snapping) are completely unaffected by the `RegisterHotKey` hook because the OS processes them at a lower level (internally via DWM/win32k) rather than registering them through standard Explorer API calls. Furthermore, the old hook would fail entirely if an Administrator window was focused due to UIPI.
 
Using a `WH_KEYBOARD_LL` hook inside `dwm.exe` was the only way I could successfully and universally intercept all system shortcuts, including window snapping. 
 
If we revert to the `RegisterHotKey` method, the mod loses the ability to block half of the most requested shortcuts. Since you are much more familiar with Windows internals, do you know of a better undocumented function/API we could hook (perhaps inside `dwm.exe` or `explorer.exe`) that would cancel the shell's internal handling of these keys, without swallowing the keystroke entirely from third-party apps?

mods/disable-windows-shortcuts.wh.cpp
Louis047: Hey there, I left a comment below, please do check it out as I have covered some crucial points related to the current mod and the reason for this approach.

m417z:
> Furthermore, the old hook would fail entirely if an Administrator window was focused due to UIPI.

What do you mean? If a shortcut isn't registered, then it isn't functional regardless of the focused window.

> do you know of a better undocumented function/API we could hook (perhaps inside `dwm.exe` or `explorer.exe`) that would cancel the shell's internal handling of these keys, without swallowing the keystroke entirely from third-party apps?

I don't know how shortcuts like Win+Left are handled, and which process is responsible for them. You can try implement system-wide logging for `RegisterHotKey`, perhaps to a log file to capture early on-boot registrations, to see if anything is logged.

Other than that, perhaps you go use a hybrid approach - use the `RegisterHotKey` key for hotkeys that support it (like in the previous mod version), and only use the keyboard hook for hotkeys which require it.

Louis047:
> What do you mean? If a shortcut isn't registered, then it isn't functional regardless of the focused window.

During my testing, for some weird reasons, the shortcuts were only blocked when focused on a non-admin window. When focused on an admin window, the shortcuts are not blocked. This is fixed already so need to bother about this.

Louis047:
I found that there are a few hardcoded shortcuts that cannot normally be blocked with `RegistryHotkey` method so I have added them to the low-level keyboard hook method. Here's the list of shortcuts affected by it:

- Win+A
- Win+K
- Win+U
- Win+P
- Win+V
- Win+Up
- Win+Down
- Win+Left
- Win+Right

I have to test this further as these are the shortcuts so far are not being blocked by `RegistryHotkey` method. So they simply cannot be used for third-party apps. Correct me if I'm wrong.

Louis047:
Update: Fixed an issue with `Win+A` shortcut key which is for quick settings/action center. When trying to open it after blocking the shortcut, it doesn't open at all indicating that the related process is blocked as well and that it treats the mouse click event for opening quick settings panel as a shortcut trigger so it get blocked.

mods/disable-windows-shortcuts.wh.cpp
```diff
@@ -0,0 +1,910 @@
...
+    HMODULE hUser32 = GetModuleHandleW(L"user32.dll");
+    if (hUser32)
+    {
+        void* pRegisterHotKey = (void*)GetProcAddress(hUser32, "RegisterHotKey");
+        if (pRegisterHotKey)
+            Wh_SetFunctionHook(pRegisterHotKey, (void*)RegisterHotKey_Hook, (void**)&RegisterHotKey_Original);
+    }
+
+    if (g_isDWM)
+        StartHookThread();
```
m417z, line 908: How about unloading the mod in dwm.exe if none of the special hotkeys are blocked?

mods/disable-windows-shortcuts.wh.cpp
```diff
@@ -0,0 +1,910 @@
...
+
+    LoadSettings();
+
+    // Hook RegisterHotKey in all processes
+    HMODULE hUser32 = GetModuleHandleW(L"user32.dll");
+    if (hUser32)
+    {
+        void* pRegisterHotKey = (void*)GetProcAddress(hUser32, "RegisterHotKey");
+        if (pRegisterHotKey)
+            Wh_SetFunctionHook(pRegisterHotKey, (void*)RegisterHotKey_Hook, (void**)&RegisterHotKey_Original);
```
m417z: Does it need to be hooked in dwm.exe?

Louis047:
I hope this does it 👍

mods/disable-windows-shortcuts.wh.cpp
```diff
@@ -854,6 +854,17 @@ void StopHookThread()
     }
```
m417z, line 853: Wait for the thread to complete, like it's done for the `g_restartExplorerPromptThread` thread.

Louis047: Done, I'll update the PR body just in case to not mislead users.

Louis047:
I found that some windows shortcuts require the explorer to restart in order to fully recover the shortcuts back to windows defaults while removing the mod. `Win+E` is one of the victims. It would be a good safety measure to implement this in order to not face any issues.

mods/disable-windows-shortcuts.wh.cpp
```diff
@@ -0,0 +1,947 @@
...
+RegisterHotKey_t RegisterHotKey_Original;
+
+BOOL WINAPI RegisterHotKey_Hook(HWND hWnd, int id, UINT fsModifiers, UINT vk)
+{
+    if (ShouldBlockHotkey(fsModifiers, vk))
+    {
+        // If the hotkey is a known hardcoded shell key (like Win+A for Action Center),
+        // we MUST let Explorer successfully register it. If we fake a failure here, 
+        // Explorer components fail to initialize and the user can't even open them 
+        // with a manual mouse click on the taskbar tray icons!
```
m417z, line 608: I found this comment surprising, and I couldn't reproduce it. but allowing these hotkeys doesn't really matter since they're blocked with the keyboard hook anyway.

PR #3816, opened 2026-04-16
Add Edge Snap Throw mod

Getrektbynoob15:
This file implements the Edge Snap Throw functionality, allowing windows to snap to screen edges with customizable settings.

## Mod authorship

This mod was created by:

- [x] Manually by the submitter (with help of AI assistance)

## Notes

Requires the **Slick Window Arrangement** mod to work. This mod hooks into
the sliding animation that Slick Window Arrangement produces, and triggers
native Windows snapping when the window hits a screen edge with enough velocity.

Snap zones:
- Left / right edge: 50% snap
- Top edge: maximize
- Corners: 25% quarter snap (no Snap Assist popup)

### DIFF
--- a/mods/edge-snap-throw.wh.cpp
+++ b/mods/edge-snap-throw.wh.cpp
@@ -2,7 +2,7 @@
 // @id              edge-snap-throw
 // @name            Edge Snap Throw
 // @description     Slide a window into a screen edge and it snaps like native Windows snap
-// @version         1.6.0
+// @version         1.9.0
 // @author          getrektbynoob20
 // @github          https://github.com/getrektbynoob20
 // @include         *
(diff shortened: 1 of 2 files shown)


mods/edge-snap-throw.wh.cpp
```diff
@@ -0,0 +1,441 @@
...
+// ── Hook SetWindowPos ─────────────────────────────────────────────────────────
+// This hook runs on the window's UI thread (Slick Window Arrangement's slide
+// timer is a thread timer processed by the UI thread's message loop).
+// That means SetWindowSubclass can safely be called here.
+BOOL WINAPI SetWindowPosHook(HWND hWnd, HWND hWndInsertAfter,
+    int X, int Y, int cx, int cy, UINT uFlags)
+{
+    const UINT slideFlags = SWP_NOSIZE | SWP_NOZORDER | SWP_NOACTIVATE | SWP_NOOWNERZORDER;
+
+    if ((uFlags & slideFlags) == slideFlags && !(uFlags & SWP_NOMOVE)) {
```
m417z, line 369: Checking for `slideFlags` doesn't guarantee sliding. I'm not sure there's an elegant solution here, but you can check the caller of `SetWindowPos`. Here's an example:

https://github.com/ramensoftware/windhawk-mods/blob/bf450c23f6810c142a3e5a4445b9afa296df24b3/mods/windows-11-file-explorer-styler.wh.cpp#L4902-L4921

mods/edge-snap-throw.wh.cpp
```diff
@@ -0,0 +1,441 @@
...
+                                std::lock_guard<std::mutex> rlock(g_restoreMutex);
+                                g_restoreMap[hWnd] = { rc, false };
+                            }
+                            SetWindowSubclass(hWnd, RestoreSubclassProc, 0, 0);
+                        }
+
+                        // Stop the slide and snap on a worker thread so we
+                        // don't block the UI thread with Sleep() calls
+                        auto* param = new SnapThreadParam{ hWnd, zone };
+                        CreateThread(nullptr, 0, SnapThreadProc, param, 0, nullptr);
```
m417z: Why not use `SetTimer` instead? Then you could also replace `g_tracks` with `thread_local` variable, and remove the mutex.

mods/edge-snap-throw.wh.cpp
```diff
@@ -0,0 +1,441 @@
...
+
+            return TRUE;
+        }
+    }
+    else if (uMsg == WM_EXITSIZEMOVE) {
+        {
+            std::lock_guard<std::mutex> lock(g_restoreMutex);
+            g_restoreMap.erase(hWnd);
+        }
+        RemoveWindowSubclass(hWnd, RestoreSubclassProc, 0);
```
m417z: `WM_EXITSIZEMOVE` is sent when dragging is done, but dragging is likely done before the subclass is set, which means that it might stay there forever, until the user drags that window again. That also means that each such window's process will crash when the mod is unloaded, since the subclass isn't removed on unload.

Getrektbynoob15: Hey, thanks for the review! Fixed everything you pointed out added the caller check with `__builtin_return_address(0)` to make sure we're only acting on calls from `slick-window-arrangement`, swapped `CreateThread` for `SetTimer` and made `g_tracks` thread_local so the mutex is gone, and fixed the subclass issue so it cleans itself up right after restoring in `WM_MOVING` and properly removes on unload too.

PR #3829, opened 2026-04-17
[Disable Windows Shortcuts]: Fix few hardcoded windows shortcut keys with a new approach

Louis047:
## Changelog
- Fixes few shortcut keys that are hard-coded into windows with a new approach
- Refactored settings into two separate groups for the updated shortcuts

### DIFF
--- a/mods/disable-windows-shortcuts.wh.cpp
+++ b/mods/disable-windows-shortcuts.wh.cpp
@@ -50,7 +50,7 @@ For this mod to successfully block window snapping (Win+Arrows) and Task View (W
 - Win+Q is redundant with Win+S (both open Search)
 
 ## Known Limitations
-- A small number of system shortcuts (Win+A, Win+K, Win+P, Win+U, Win+Tab, and Win+Arrows) are hardcoded into the OS at a low level. While this mod successfully blocks them from performing their default Windows actions, they cannot be freed up to be registered by third-party apps (like AutoHotkey or PowerToys). When these specific shortcuts are disabled, they simply do nothing.
+- A small number of system shortcuts (Win+A, Win+K, Win+N, Win+P, Win+U, Win+Tab, and Win+Arrows) are hardcoded into the OS at a low level. While this mod successfully blocks them from performing their default Windows actions, they cannot be freed up to be registered by third-party apps (like AutoHotkey or PowerToys). When these specific shortcuts are disabled, they simply do nothing.
 */
 // ==/WindhawkModReadme==
 // ==WindhawkModSettings==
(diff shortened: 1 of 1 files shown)


m417z:
Can you explain the issue? I tried blocking Win+N with the release version and it seems to be working fine. The notification center can still be opened from the tray by clicking on the clock.

Louis047:
Hmm that's odd, i had this issue in the release version, here's a video showcase:

https://github.com/user-attachments/assets/58aaffe0-73c5-4bef-b19d-964fd665e4d1

m417z:
I assume that you're using a third party tool (YASB or similar?) that opens the notification area by simulating Win+N instead of opening it natively like the native taskbar does. Using a hook fixes that, but it's not relevant for most users with the native taskbar.

Louis047:
Yup, you were right about the bar, I'm using YASB. The notification opens in the taskbar but not in yasb. But either way, this might fix potential conflicts with third party apps where they trigger the notification center panel. Let me know your opinion on this, anything to change etc.

m417z:
I think that the default shouldn't require dwm.exe, which is more limiting, more complicated to set up, and not needed for most users.

How about, for these hotkeys (Win+N, Win+A, any others?), provide 3 options: Off, Disable hotkey, Block hotkey. And add a description explaining that disabling the hotkey means that apps that simulate it are affected too, and that blocking it means that apps can still simulate it, but it can't be used for anything else and requires the dwm.exe step.

Louis047:
> How about, for these hotkeys (Win+N, Win+A, any others?), provide 3 options: Off, Disable hotkey, Block hotkey. And add a description explaining that disabling the hotkey means that apps that simulate it are affected too, and that blocking it means that apps can still simulate it, but it can't be used for anything else and requires the dwm.exe step.

Sounds like a good idea to me. Will work on it once I get back home.

mods/disable-windows-shortcuts.wh.cpp
```diff
@@ -14,56 +14,86 @@
...
-## Known Limitations
-- A small number of system shortcuts (Win+A, Win+K, Win+P, Win+U, Win+Tab, and Win+Arrows) are hardcoded into the OS at a low level. While this mod successfully blocks them from performing their default Windows actions, they cannot be freed up to be registered by third-party apps (like AutoHotkey or PowerToys). When these specific shortcuts are disabled, they simply do nothing.
 */
 // ==/WindhawkModReadme==
 // ==WindhawkModSettings==
 /*
-- DisableWinA: false
+- _Separator: false
+  $name: ⬇️ Stubborn Shortcuts (3-Tier Options) ⬇️
+  $description: (Read the 'Advanced / Stubborn Shortcuts' section in Details for info)
```
m417z: Having a simulated separator like this is confusing. You can use groups instead, for an example, see:
https://windhawk.net/mods/translucent-flyouts-controller

Louis047:
image

@m417z i hope this styling is fine, let me know if further improvement is needed. I kind of want to change the number-based toggle into a dropdown like you have cited from [this](https://windhawk.net/mods/translucent-flyouts-controller) mod

m417z:
You can do it by using a string, example:
https://github.com/ramensoftware/windhawk-mods/blob/f8301dd5b08cb37bc3f3983d7732dc7c5a570548/mods/taskbar-grouping.wh.cpp#L62-L69

Louis047:
Tested the new changes for notification center with third party apps (like YASB) with `Block Hotkey` option and it works as expected. Block the hotkey but the notification center opens on manual trigger. 

Yet to know the result for this issue: #3842 @prochan2

PR #3856, opened 2026-04-22
Taskbar Auto-Hide Instant Show v2.2

Bo0ii:
- Add **No animation** and **Overshoot** options
- Upgrade **Slide + Fade** to quintic ease for silkier motion
- Default **Edge detection** on; replace one-shot trigger with a throttled retry loop so the taskbar reliably appears over fullscreen apps and foreground windows that swallow the edge event

### DIFF
--- a/mods/taskbar-autohide-instant-show.wh.cpp
+++ b/mods/taskbar-autohide-instant-show.wh.cpp
@@ -1,8 +1,8 @@
 // ==WindhawkMod==
 // @id              taskbar-autohide-instant-show
 // @name            Taskbar Auto-Hide Instant Show
-// @description     Removes the delay before the taskbar appears with custom animation types (elastic, bounce, fade, etc.)
-// @version         2.1
+// @description     Removes the delay before the taskbar appears with custom animation types (none, slide, elastic, bounce, fade, slide+fade, overshoot)
+// @version         2.2
 // @author          Bo0ii
 // @github          https://github.com/Bo0ii
 // @homepage        https://github.com/Bo0ii/windhawk-mods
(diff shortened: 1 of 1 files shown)


m417z:
> * Default **Edge detection** on

Why was it changed? We discussed it in a previous PR. I don't think it should be enabled by default. That's not the main purpose of the mod, it changes the default Windows behavior and creates an extra background thread for all users, even those who don't need this option or don't know about it.

Bo0ii:
You're right, I'll revert the default back to false. The retry-loop change (replacing the one-shot latch) still lands, so users who do enable edge detection get the reliability fix, but it stays opt-in, and no background thread
runs for anyone who hasn't asked for it. Pushing the fix now.

The reason I experimented with that again is bcz, when having the terminal opened or sometimes Chrome the taskbar stops coming up and u need to force it by minimizing Chrome, then it works. This happens 10% of the time not a major issue.

Bo0ii:
reverted the default back

mods/taskbar-autohide-instant-show.wh.cpp
```diff
@@ -1,8 +1,8 @@
 // ==WindhawkMod==
 // @id              taskbar-autohide-instant-show
 // @name            Taskbar Auto-Hide Instant Show
-// @description     Removes the delay before the taskbar appears with custom animation types (elastic, bounce, fade, etc.)
-// @version         2.1
+// @description     Removes the delay before the taskbar appears with custom animation types (elastic, bounce, fade, overshoot, snap, swing, silky slide+fade, etc.)
```
m417z: What do snap and swing refer to?

Bo0ii: was experimenting on some animations and decided to remove them last minute. Thanks for noticing that

PR #3758, opened 2026-04-09
Add Disk Usage Bar Color mod v1.1.0

dirtyrazkl:
This mod allows for changing the colors of the Windows 11 disk usage bars under This PC in File Explorer. Supports separate colors for normal usage and nearly-full (warning) state. File copy/move progress bars are not affected.

This mod was developed with AI assistance (GitHub Copilot). I reviewed and understand all the code.

## Changelog

If the submission is an update to an existing mod, include the changelog below:

* Changelog item 1...
* Changelog item 2...

## Mod authorship

If the submission is a new mod, please fill the form below.

This mod was created by:

- - [ ] Manually by the submitter (with or without AI assistance)
- - [ ] Claude Code
- - [ ] ChatGPT
- - [ ] Gemini
- - [x] Another AI (please specify): GitHub Copilot 
- - [ ] Other (please specify):

Please select the appropriate option. Your selection will not affect acceptance criteria, but will help reviewers understand the context of the code and provide relevant feedback.

PR #3861, opened 2026-04-22
 Enhanced Disk Usage

bbmaster123:
Merge of my disk details mod and Disk Usage Bar Color https://github.com/ramensoftware/windhawk-mods/pull/3758
image

has now been tested across various resolutions and scaling settings, but nothing above 4k at the moment. I have only tested in English so far, and will likely not work in other languages without adding translated strings to parse. I did add support for both . and , 

made with ai assistance (Gemini)

## Changelog

* Changelog item 1...
* Changelog item 2...

## Mod authorship

This mod was created by:

- - [ ] Claude Code
- - [ ] ChatGPT
- - [ ] Gemini


mods/enhanced-disk-usage.wh.cpp
```diff
@@ -0,0 +1,545 @@
...
+                           }
+        std::wstring cf = CleanNumericString(fs), ct = CleanNumericString(ts);
+        double fv, tv; wchar_t fu[16], tu[16];
+        
+        if (swscanf(cf.c_str(), L"%lf %15s", &fv, fu) == 2 && swscanf(ct.c_str(), L"%lf %15s", &tv, tu) == 2) {
+            std::wstring us = std::wstring(StrFormatByteSizeW((ULONGLONG)std::max(0.0, (tv * GetUnitMultiplier(tu) - fv * GetUnitMultiplier(fu))), fu, 16));
+            
+            if (g_removeSpace) us.erase(std::remove(us.begin(), us.end(), L' '), us.end());
+            std::wstring fS = g_formatString; size_t p1 = fS.find(L"%s"), p2 = fS.find(L"%s", p1+2), p3 = fS.find(L"%s", p2+2);
+            std::wstring s1 = fS.substr(0, p1) + fs + fS.substr(p1+2, p2-p1-2), s2 = g_boldUsed ? MakeBoldText(us) : us, s3 = fS.substr(p2+2, p3-p2-2) + ts + fS.substr(p3+2);
```
m417z: Long lines such as this one make the mod difficult to read and review. I'd guess that you didn't write these parts yourself, right? Consider making it more readable. One way is to auto-format the code in Windhawk (right click menu in the editor).

bbmaster123: it was even more condensed originally. I anticipated that being a readability issue, and manually formatted the parts that gemini generated, creating a new line at the end of each statement after the semi-colon.

I didn't realize the editor had built in formatting like vs, but of course that makes sense that it exists! I didn't even think to check aha
thanks for the heads up! Will auto-format now

mods/enhanced-disk-usage.wh.cpp
```diff
@@ -0,0 +1,669 @@
...
+    }
+
+    // prevent navpane being styled
+    if (pRect->left < (int)(32 * scale))
+        return false;
+    return true;
+}
+
+static RECT g_lastBarRect = {0};
+static HDC g_lastBarDC = NULL;
```
m417z: Use `thread_local` since this is per-thread data. Each Explorer window runs in a separate thread and has its own state.

mods/enhanced-disk-usage.wh.cpp
```diff
@@ -0,0 +1,669 @@
...
+                    return false;
+                }
+            }
+            walk = GetParent(walk);
+        }
+    }
+
+    // prevent navpane being styled
+    if (pRect->left < (int)(32 * scale))
+        return false;
```
m417z, line 690: This heuristic looks fragile, have you tried having a more specific check for the target window? e.g. instead of excluding class names such as `#32770`, why not target the expected class name?

mods/enhanced-disk-usage.wh.cpp
```diff
@@ -0,0 +1,669 @@
...
+
+bool FindSpaceStats(const std::wstring& t, std::wstring& f, std::wstring& tot) {
+    std::wstring nt = t;
+    for (auto& c : nt) {
+        if (c == 0xA0)
+            c = L' ';
+    }
+
+    // Search for the split points used by Explorer
+    size_t split = nt.find(L" free of ");
```
m417z: This will only work for English Windows versions.

mods/enhanced-disk-usage.wh.cpp
```diff
@@ -0,0 +1,669 @@
...
+    return false;
+}
+
+int WINAPI DrawTextW_Hook(HDC hdc, LPCWSTR psz, int cch, LPRECT prc, UINT fmt) {
+    if (!psz || !prc)
+        return DrawTextW_Orig(hdc, psz, cch, prc, fmt);
+    std::wstring t(psz, cch == -1 ? wcslen(psz) : cch);
+    std::wstring fs, ts;
+
+    if (FindSpaceStats(t, fs, ts)) {
```
m417z, line 796: `FindSpaceStats` is called regardless of the target. Could you check that the `hdc` target window matches the expected window?

mods/enhanced-disk-usage.wh.cpp
```diff
@@ -0,0 +1,669 @@
...
+                            int cch,
+                            LPRECT prc,
+                            UINT fmt,
+                            LPDRAWTEXTPARAMS pDtp) {
+    if (!psz || !prc)
+        return DrawTextExW_Orig(hdc, psz, cch, prc, fmt, pDtp);
+    std::wstring t(psz, cch == -1 ? wcslen(psz) : cch);
+    std::wstring fs, ts;
+
+    if (FindSpaceStats(t, fs, ts)) {
```
m417z, line 975: `FindSpaceStats` is called regardless of the target. Could you check that the `hdc` target window matches the expected window?

mods/enhanced-disk-usage.wh.cpp
```diff
@@ -0,0 +1,669 @@
+// ==WindhawkMod==
+// @id              enhanced-disk-usage
+// @name            Enhanced Disk Usage
+// @description     Enables the ability to customize the disk drive tiles in explorer. Several customizations are available for enhancing the
```
m417z: The description is cut off.

m417z:
For me, the format doesn't fit with the default settings.

https://github.com/user-attachments/assets/92a7b300-cd2c-45dc-b09a-1d2dae4dc976

bbmaster123:
thanks for the feedback!
I will try to integrate all of of that. I also want this to be robust and bug-free. 
I'm already using it on my main system

> Use thread_local

Will do!

> This heuristic looks fragile, have you tried having a more specific check for the target window?

yes I did try, however I was finding unrelated elements being styled such as the horizontal scrollbar in dialog windows such as #32270, this is my best attempt so far to style only the disk bars, but I'm open of course to other implementations 

> This will only work for English Windows versions.

my plan was to make sure the mod was stable and then add strings for other languages, also not married to this implementation if there's a better way.

 > FindSpaceStats is called regardless of the target. Could you check that the hdc target window matches the expected window?
 
 I'm sure I can figure that out!
 
>  The description is cut off.

whoops my bad!

I'm also going to try to the fix border outline when set to 0, and try to add track height settings, separate from disk bar height

>  the format doesn't fit with the default settings

Maybe I can find a way to line wrap the text if its too long... Otherwise I could change the default format so that it does fit, or I can ellipse the cut off text, hmmm...
This would fit by default:
`100 GB free | 100/200 GB total`

EDIT: coming along nicely:
image

mods/enhanced-disk-usage.wh.cpp
```diff
@@ -0,0 +1,1061 @@
...
+
+        Wh_SetFunctionHook(
+            (void*)GetProcAddress(uxtheme, "DrawThemeBackground"),
+            (void*)HookedDrawThemeBackground,
+            (void**)&DrawThemeBackground_Orig);
+    }
+
+    HMODULE user32 = GetModuleHandle(L"user32.dll");
+    if (user32) {
+        void* pOpenDpi = (void*)GetProcAddress(user32, "OpenThemeDataForDpi");
```
m417z: `OpenThemeDataForDpi` is exported from `uxtheme.dll`, not `user32.dll`. Have you verified what `GetProcAddress` returns here?

mods/enhanced-disk-usage.wh.cpp
```diff
@@ -0,0 +1,1061 @@
...
+                          (c - L'a'));
+        } else
+            res += c;
+    }
+    return res;
+}
+
+static bool IsValidDiskBarWindow(HWND hwnd) {
+    if (!hwnd)
+        return true;
```
m417z, line 389: Is it intentional that `IsValidDiskBarWindow` returns true when there's no window?

mods/enhanced-disk-usage.wh.cpp
```diff
@@ -0,0 +1,1061 @@
...
+  $name: Enable Glossy Overlay
+- leftInset: 0
+  $name: Bar Inset Left
+- rightInset: 0
+  $name: Bar Inset Right
+- topInset: 0
+  $name: Bar Inset Top
+- bottomInset: 0
+  $name: Bar Inset Bottom
+- formatString: "%s free I %s used\n%s total"
```
m417z: I assume it was supposed to be `|` and not `I` (capital i), right?

Also, newlines aren't supported by the settings editor, replace `\n` with `\\n`.

mods/enhanced-disk-usage.wh.cpp
```diff
@@ -0,0 +1,1061 @@
...
+                }
+            }
+        }
+    }
+
+    return DrawTextExW_Orig(hdc, psz, cch, prc, fmt, pDtp);
+}
+
+static BOOL CALLBACK RefreshExplorerCallback(HWND hwnd, LPARAM lParam) {
+    wchar_t cls[MAX_PATH];
```
m417z, line 1139: Skip windows which aren't part of the current process.

mods/enhanced-disk-usage.wh.cpp
```diff
@@ -0,0 +1,1061 @@
...
+
+- barNormalStart: "#2ECC71"
+  $name: Bar Color Gradient Start
+  $description: set start and end to the same value if you do not want a gradient
+- barNormalEnd: "#27AE60"
+  $name: Bar Color Gradient End
+- barFullStart: "#E74C3C"
+  $name: Bar Full Color Gradient Start
+- barFullEnd: "#C0392B"
+  $name: Bar Full Color Graient End
```
m417z:
```suggestion
  $name: Bar Full Color Gradient End
```

mods/enhanced-disk-usage.wh.cpp
```diff
@@ -0,0 +1,1061 @@
...
+disk's usage bar, as well as the details that appear below.
+
+- custom colors with transparency for disk usage, track (background/unused), and
+outline
+- separate disk colors for when drive is near full
+- linear gradient support with configurable direction
+- rounded corners
+- glossy overlay toggle option (for a more Windows Aero-ish looking aesthetic)
+- height/width controls (inset) controls for disk bar and track
+- custom disk usage text with font size adjustment, muti-line support,
```
m417z:
```suggestion
- custom disk usage text with font size adjustment, multi-line support,
```

mods/enhanced-disk-usage.wh.cpp
bbmaster123: ah, thank you for pointing this out! I had not checked, and it had been returning a null address. I've corrected this and it now returns an actual value.

mods/enhanced-disk-usage.wh.cpp
bbmaster123: in this case, apparently yes, as according ai, its first drawn offscreen into a double buffer, and needs to be handled there. I have not attempted to skip the buffer and draw directly if that's possible, but I did check that returning false with the current code causes the text to revert to default. Same if I remove the check altogether.
as always, open to suggestions! :)

mods/enhanced-disk-usage.wh.cpp
bbmaster123: yes correct, my mistake. I was checking out other characters to use in place of the pipe character, since its not perfectly centered vertically. I meant to revert this, thanks for catching it!

mods/enhanced-disk-usage.wh.cpp
bbmaster123: whoops, yes will do!

mods/enhanced-disk-usage.wh.cpp
m417z: Currently, having a file/folder with a name such as "Backup 100MB to 200MB" or "1 to 2 m" triggers the mod, which is of course not optimal. I don't have a simple fix, but I'd try to either hook a more specific function, or harden the check with a set of events that only occur when drawing the driver bars. Both require some research. For now, I can merge it as is, but it'd be better to have it addressed in the future.

bbmaster123:
@m417z 
cool! happy to see my first mod merged!
I just found and fixed one more issue, since this is merged how should I handle adding the fix? new PR? let me know, thanks! :)

m417z:
Congratulations :)
Yes, a new PR with the updated code. It's also mentioned in the repo readme:
https://github.com/ramensoftware/windhawk-mods#submitting-a-mod-update

PR #3865, opened 2026-04-22
Update - Explorer Details View Columns - version 1.1

ernisn:
## Changelog (Edit)

- You can now choose to set a width for each column or not.
- Refined width realization, fixed the problem of minimum width stuck at around 80.
- Width settings will be consistent now in case of multiple monitors with different DPI settings.
- New changes should take effect immediately now including opened folders.

---

## Changelog

- Width settings should be consistent now after changing monitors with different DPI.
- New changes take effect immediately now to all windows including opened ones.

### DIFF
--- a/mods/explorer-force-details-columns.wh.cpp
+++ b/mods/explorer-force-details-columns.wh.cpp
@@ -2,11 +2,11 @@
 // @id              explorer-force-details-columns
 // @name            Explorer Details View Columns
 // @description     Forces a fixed set of columns in File Explorer Details view. Columns and their order are configurable. Has no effect on other view modes.
-// @version         1.0
+// @version         1.1
 // @author          ernisn
 // @github          https://github.com/ernisn
 // @include         explorer.exe
-// @compilerOptions -lole32 -lshlwapi -lpropsys
+// @compilerOptions -lole32 -lshlwapi -lpropsys -lgdi32
 // @license         MIT
 // ==/WindhawkMod==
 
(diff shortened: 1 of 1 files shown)


mods/explorer-force-details-columns.wh.cpp
```diff
@@ -133,32 +166,79 @@ static void ApplyForcedColumns(void* pThis) {
...
         if (col.width <= 0)
             continue;
 
         CM_COLUMNINFO ci = {};
-        ci.cbSize  = sizeof(ci);
-        ci.dwMask  = CM_MASK_WIDTH;
-        ci.uWidth  = static_cast<UINT>(col.width);
+        ci.cbSize = sizeof(ci);
+        ci.dwMask = CM_MASK_WIDTH;
+        ci.uWidth = ScaleToDpi(col.width);
```
m417z: What about per-monitor DPI? It might be better to use `GetDpiForWindow` here if you can get the window handle.

mods/explorer-force-details-columns.wh.cpp
```diff
@@ -133,32 +166,79 @@ static void ApplyForcedColumns(void* pThis) {
...
+                                            reinterpret_cast<void**>(&pWBA))) && pWBA) {
+            IServiceProvider* pSP = nullptr;
+            if (SUCCEEDED(pWBA->QueryInterface(IID_IServiceProvider_,
+                                               reinterpret_cast<void**>(&pSP))) && pSP) {
+                IShellBrowser* pSB = nullptr;
+                if (SUCCEEDED(pSP->QueryService(SID_STopLevelBrowser, IID_IShellBrowser,
+                                                reinterpret_cast<void**>(&pSB))) && pSB) {
+                    IShellView* pSV = nullptr;
+                    if (SUCCEEDED(pSB->QueryActiveShellView(&pSV)) && pSV) {
+                        ApplyForcedColumns(pSV);
```
m417z: Before applying, check that the enumerated window belongs to the current process.

PR #3872, opened 2026-04-23
osu!Profile in Taskbar - Update: 3.9.1

Antoshika:
# 🟣 osu!Profile in Taskbar
Version: 3.9.1
## Changelog

* After restarting the computer, the mod displayed an "error" message, which was resolved by "enabling/disabling" the mod. Now, after restarting the PC, the mod works correctly.

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [X] Gemini

### DIFF
--- a/mods/osu-tray-profile.wh.cpp
+++ b/mods/osu-tray-profile.wh.cpp
@@ -2,7 +2,7 @@
 // @id              osu-tray-profile
 // @name            osu!Profile in Taskbar
 // @description     Displays PP, rank, and avatar from rhythm game osu! (standard mode) next to the system tray
-// @version         3.9
+// @version         3.9.1
 // @author          antoshika
 // @github          https://github.com/Antoshika
 // @include         windhawk.exe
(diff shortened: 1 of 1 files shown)


mods/osu-tray-profile.wh.cpp
```diff
@@ -445,7 +527,7 @@ void UiThreadFunc() {
             WS_EX_LAYERED | WS_EX_TOOLWINDOW,
             wc.lpszClassName, L"OsuStats",
             WS_POPUP,
-            0, 0, 200, 50,
+            -1000, -1000, 200, 50, // Стартуем за пределами экрана!
```
m417z: This might cause it to appear at a different monitor, to the top-left of the primary monitor. Why not show it only when it's positioned?

Antoshika: > This might cause it to appear at a different monitor, to the top-left of the primary monitor. Why not show it only when it's positioned?

this means that it will first run off-screen and then return to its place, for correct launch

m417z: I mean, if you have multiple monitors, (-1000, -1000) might not be off-screen.

Example:

image

Antoshika: ok, i'll fix it

PR #3893, opened 2026-04-25
New mod "Explorer Rename Overwrite"

triatomic:
# Explorer Rename Overwrite

When renaming a file in Explorer to a name that already exists, this mod overwrites the existing file instead of showing the "file already exists" prompt or auto-numbering.

Works by hooking `IFileOperation::RenameItem` / `RenameItems`. On collision, the existing file is cleared first per the user's chosen disposition:
- **recycle** (default) — sent to the Recycle Bin, optionally renamed with a suffix first so restoring it won't clobber the new file.
- **keep** — renamed in place (e.g. `report.txt` → `report.txt_old`), kept in the same directory with a deliberately broken file association.
- **permanent** — irreversibly deleted.

All suffixes, the master enable toggle, and a directory-protection guard are exposed as Windhawk settings.

## Mod authorship

This mod was created by:

- - [x] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/explorer-rename-overwrite.wh.cpp
+++ b/mods/explorer-rename-overwrite.wh.cpp
@@ -2,7 +2,7 @@
 // @id              explorer-rename-overwrite
 // @name            Explorer Rename Overwrite
 // @description     When renaming a file in Explorer to a name that already exists, overwrite the existing file instead of being blocked by the "file already exists" prompt.
-// @version         1.1.0
+// @version         1.1.1
 // @author          tria
 // @github          https://github.com/triatomic
 // @include         explorer.exe
@@ -47,6 +47,11 @@ removed first (to the Recycle Bin by default, or permanently if configured).
   treatment as the recycle path. Defaults to `_old`.
 
 ## Changelog
+
+### 1.1.1
+- Fix `keep` disposition: notify the shell of the in-place rename so
+  Explorer's view doesn't transiently show both files with the suffix
+  until the window is refreshed.
 
 ### 1.1.0
 - New `keep` disposition: rename the overwritten file in place (default
@@ -231,7 +236,12 @@ bool KeepInPlace(const std::wstring& path) {
     if (g_settings.keepInPlaceSuffix.empty()) return false;
     std::wstring dst = PickAppendedPath(path);
     if (dst.empty()) return false;
-    return MoveFileExW(path.c_str(), dst.c_str(), 0) != 0;
+    if (MoveFileExW(path.c_str(), dst.c_str(), 0) == 0) return false;
+    // Tell the shell about the rename so Explorer's view doesn't end up with
+    // a stale entry pointing at the old name (which makes the new file
+    // appear with the "_old" suffix until the user refreshes).
+    SHChangeNotify(SHCNE_RENAMEITEM, SHCNF_PATHW, path.c_str(), dst.c_str());
+    return true;
 }
 
 bool DeleteToRecycleBin(const std::wstring& path) {
--- a/mods/explorer-rename-overwrite.wh.cpp
+++ b/mods/explorer-rename-overwrite.wh.cpp
@@ -1,0 +1,421 @@
+// ==WindhawkMod==
+// @id              explorer-rename-overwrite
+// @name            Explorer Rename Overwrite
+// @description     When renaming a file in Explorer to a name that already exists, overwrite the existing file instead of being blocked by the "file already exists" prompt.
+// @version         1.1.0
+// @author          tria
+// @github          https://github.com/triatomic
+// @include         explorer.exe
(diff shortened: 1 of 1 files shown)


triatomic:
image

Settings Page

mods/explorer-rename-overwrite.wh.cpp
```diff
@@ -0,0 +1,431 @@
...
+- Settings: `enabled`, `disposition` (`recycle` / `permanent`),
+  `onlyFiles`, `renameBeforeRecycle`, `recycleSuffix`.
+*/
+// ==/WindhawkModReadme==
+
+// ==WindhawkModSettings==
+/*
+- enabled: true
+  $name: Enable overwrite-on-rename
+  $description: Master switch. When disabled, the mod does nothing and Explorer uses its default rename behavior. Useful for temporarily turning the feature off without unloading the mod.
```
m417z: Why would this option be useful? It seems to duplicate the ability to disable the mod. There should be no visible difference between unloading a mod or having it do nothing.

mods/explorer-rename-overwrite.wh.cpp
```diff
@@ -0,0 +1,431 @@
...
+    }
+
+    std::wstring toRecycle = target;
+    if (g_settings.renameBeforeRecycle &&
+        !g_settings.recycleSuffix.empty() &&
+        !(attr & FILE_ATTRIBUTE_DIRECTORY)) {
+        std::wstring suffixed = PickSuffixedPath(target);
+        if (!suffixed.empty() &&
+            MoveFileExW(target.c_str(), suffixed.c_str(),
+                        MOVEFILE_REPLACE_EXISTING) != 0) {
```
m417z: Since `PickSuffixedPath` returns a path that doesn't exist, it's safer to avoid using `MOVEFILE_REPLACE_EXISTING` here.

mods/explorer-rename-overwrite.wh.cpp
```diff
@@ -0,0 +1,431 @@
...
+    op.fFlags = FOF_ALLOWUNDO | FOF_NO_UI | FOF_NOCONFIRMATION |
+                FOF_NOERRORUI | FOF_SILENT;
+    return SHFileOperationW(&op) == 0 && !op.fAnyOperationsAborted;
+}
+
+bool DeletePermanently(const std::wstring& path) {
+    DWORD attr = GetFileAttributesW(path.c_str());
+    if (attr == INVALID_FILE_ATTRIBUTES) return true;  // already gone
+    if (attr & FILE_ATTRIBUTE_DIRECTORY) {
+        return RemoveDirectoryW(path.c_str()) != 0;
```
m417z: `RemoveDirectoryW` only removes empty directories.

mods/explorer-rename-overwrite.wh.cpp
```diff
@@ -0,0 +1,431 @@
...
+               target.c_str());
+        return false;
+    }
+
+    if (g_settings.disposition == Disposition::Permanent) {
+        return DeletePermanently(target);
+    }
+
+    if (g_settings.disposition == Disposition::Keep) {
+        return KeepInPlace(target);
```
m417z, line 320: `target` might exist with a different case. Then, I think it's better to preserve it.

For example:

Rename: Src.txt -> Dest.txt
Existing file: DEST.txt - will be renamed to Dest.txt_old (instead of DEST.txt_old)

m417z (commented):
Nice mod, thanks for the submission. I left a couple of optional suggestions, let me know what you think.

mods/explorer-rename-overwrite.wh.cpp
triatomic: Maybe you just want to temporarily disable it & not unload it from memory.

mods/explorer-rename-overwrite.wh.cpp
triatomic: Directory function is very untested tbh; might aswell just keep it for files.

mods/explorer-rename-overwrite.wh.cpp
m417z: Why would that be useful? Why would a user want to do that?

triatomic: Maybe you don't want to overwrite the old with the new version of a downloaded file but rather want to keep both.
I mean you can also just unload the mod it so ; yeah that button might be unecessary. I just felt its "more convenient" then unloading from the memory

m417z: For the user, it adds another way to do the same thing. There's no observable difference between an unloaded mod and a mod that is loaded and does nothing.

Please remove this option and address the other comments. Let me know when the mod is ready for review.

PR #3901, opened 2026-04-25
Disable Windows Web Search 1.0

davidevol:
* Initial release.

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [x] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/disable-windows-web-search.wh.cpp
+++ b/mods/disable-windows-web-search.wh.cpp
@@ -1,0 +1,164 @@
+// ==WindhawkMod==
+// @id              disable-windows-web-search
+// @name            Disable Windows Web Search
+// @description     Turns off Bing results in Windows Search
+// @version         1.0
+// @author          davidevol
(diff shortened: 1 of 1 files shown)


mods/disable-windows-web-search.wh.cpp
```diff
@@ -0,0 +1,164 @@
...
+        nullptr,
+        &hKey,
+        nullptr
+    );
+
+    if (status != ERROR_SUCCESS) {
+        return false;
+    }
+
+    status = RegSetValueExW(
```
m417z, line 87: One of the benefits of Windhawk mods is that they don't make persistent changes to the system - all changes are done in memory. It makes it safe to try mods, knowing that disabling them will revert all changes.

This registry modification violates this principle.

davidevol: I'm going to rewrite the mod to according to the principles of Windwahk, Thank you for your willingness.

Issue #3880, opened 2026-04-24
[Disable Windows Shortcuts] Add `Win+Q`, `Win+Ctrl+F`, `Win+/`


AromaKitsune:
Mod author: @Louis047 

There are other keyboard shortcuts to disable.

Standard shortcuts:
* `Win+Q` - Search
* `Win+Ctrl+F` - Find Computers

Special shortcut:
* `Win+/` - IME reconversion

Louis047:
@AromaKitsune Hey, thanks for the suggestion! 

For `Win+Q`, I deliberately excluded it as it carries the same functionality as of `Win+S`. I'll add it back as people are requesting for it. I found another shortcut which configures keyboard input switch `Alt+Shift` It'll be added in the standard shortcut list as well. 

I'll work on this once I get free time 👍

AromaKitsune:
Yes please, disabling `Win+Q` would allow me to assign this shortcut for [EverythingToolbar](https://github.com/srwi/EverythingToolbar), effectively "replacing" Search, which is way better for searching files lol, while I use `Win+S` for opening the Settings app via AutoHotkey (using `Run "ms-settings:"` since I also disabled `Win+I`).

Louis047:
> disabling Win+Q would allow me to assign this shortcut for [EverythingToolbar](https://github.com/srwi/EverythingToolbar), effectively "replacing" Search, which is way better for searching files lol

Yeah fair :) 

I personally use `FlowLauncher` with `Everything Alpha` which works the best for me. I have completely blocked search and start menu `Block Start Menu and Hosts` mod as I don't use them at all.

Louis047:
F, I just noticed that `Win+Space` doesn't get blocked. Currently working on it. I have done implementing the shortcuts you have requested and they work as expected (Tested only for `Find Computers` and `Win+Q` search)

Louis047:
@AromaKitsune Added and fixed the issues. Made a PR. I encourage you to test it as well to ensure everything works 👍

AromaKitsune:
@Louis047 Here are the results

- `Win+Q` - works
- `Win+Ctrl+F` - works
- `Win+/` - works (Block hotkey)
- `Alt+Shift` - works
- `Win+Space` - works

Louis047:
@AromaKitsune Thanks for the confirmation, so its confirmed the mod works successfully!

RadekPilich:
[...]

PR #3906, opened 2026-04-26
Update Disable Windows Shortcuts Mod

Louis047:
## Changelog

- Added new shortcuts (#3880)
  - Standard: `Win+Q`, `Win+Ctrl+F`, `Alt+Shift`, `Ctrl+Esc`, `Win`
  - Special: `Win+/` 
- Fixed an issue where after re-enabling the mod, it won't apply the settings. It shows an explorer restart prompt to re-apply the settings correctly now
- Fixed mod un-initialization and re-initialization delay
- Fixed `Win+Space` which refused to get blocked in the earlier versions
- Bump Version to `1.1.1`

## Mod authorship

This mod was updated by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini
- - [x] Another AI (please specify): Antigravity (Gemini 3.1 Pro (High))

### DIFF
--- a/mods/disable-windows-shortcuts.wh.cpp
+++ b/mods/disable-windows-shortcuts.wh.cpp
@@ -2,7 +2,7 @@
 // @id              disable-windows-shortcuts
 // @name            Disable Windows Shortcuts
 // @description     Selectively disable Windows keyboard shortcuts with individual toggles
-// @version         1.1.0
+// @version         1.1.1
 // @author          Lone
 // @github          https://github.com/Louis047
 // @include         explorer.exe
@@ -27,7 +27,7 @@ To handle these properly, this mod provides **three options** for them in a dedi
 - **2 - Block hotkey:** Blocks the physical keystroke but tricks Windows into thinking it was registered. This allows simulating apps to work while physically blocking the key, but **requires injecting into `dwm.exe`**.
 
 ## ⚠️ Important `dwm.exe` Installation Step ⚠️
-If you use the **"Block hotkey"** option, or if you disable window snapping (Win+Arrows) or Task View (Win+Tab), you **must** allow Windhawk to inject into the Desktop Window Manager (`dwm.exe`):
+If you use the **"Block hotkey"** option, or if you disable window snapping (Win+Arrows), Task View (Win+Tab), or Switch keyboard layout (Win+Space, Alt+Shift), you **must** allow Windhawk to inject into the Desktop Window Manager (`dwm.exe`):
 1. Open Windhawk and go to **Settings**
 2. Click on **Advanced settings** at the bottom
 3. Under **Process inclusion list**, ensure `dwm.exe` is added (or `*` is used to include all processes)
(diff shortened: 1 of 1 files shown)


mods/disable-windows-shortcuts.wh.cpp
```diff
@@ -941,6 +1021,20 @@ BOOL Wh_ModInit()
             if (pRegisterHotKey)
                 Wh_SetFunctionHook(pRegisterHotKey, (void*)RegisterHotKey_Hook, (void**)&RegisterHotKey_Original);
         }
+
+        // Check if Explorer has already registered standard hotkeys.
+        // We use Win+R as a probe. If it fails, Explorer is mid-session and already owns it.
+        if (!IsFirstTimeInit() && RegisterHotKey_Original)
```
m417z: Did you mean to check for `IsFirstTimeInit()` instead of `!IsFirstTimeInit()` here? If not, what's the idea behind the check?

mods/disable-windows-shortcuts.wh.cpp
```diff
@@ -941,6 +1021,20 @@ BOOL Wh_ModInit()
...
+        {
+            if (!RegisterHotKey_Original(NULL, 0x1337, MOD_WIN | MOD_NOREPEAT, 'R'))
```
m417z: Just call `RegisterHotKey`. Hooks are only set after the init function returns.

mods/disable-windows-shortcuts.wh.cpp
```diff
@@ -659,6 +687,28 @@ BOOL WINAPI RegisterHotKey_Hook(HWND hWnd, int id, UINT fsModifiers, UINT vk)
...
+    LSTATUS status = RegCreateKeyExW(HKEY_CURRENT_USER, L"Software\\Windhawk\\DisableWindowsShortcuts", 0, NULL, REG_OPTION_NON_VOLATILE, KEY_READ | KEY_WRITE, NULL, &hKey, NULL);
+    if (status == ERROR_SUCCESS)
+    {
+        DWORD initialized = 0;
+        DWORD size = sizeof(initialized);
+        status = RegQueryValueExW(hKey, L"Initialized", NULL, NULL, (LPBYTE)&initialized, &size);
+        if (status != ERROR_SUCCESS)
+        {
+            initialized = 1;
+            RegSetValueExW(hKey, L"Initialized", 0, REG_DWORD, (const BYTE*)&initialized, sizeof(initialized));
```
m417z: Use `Wh_SetIntValue` instead. Mods in the portable version shouldn't write to the registry.

mods/disable-windows-shortcuts.wh.cpp
Louis047: The idea behind using `!IsFirstTimeInit()` is to intentionally prevent the mod from prompting the user to restart Explorer immediately upon their very first installation of the mod. Since by default all the mod's settings are 'off', throwing a restart prompt right after they click 'Install' is annoying and unnecessary. The prompt will only show up subsequently when they actually toggle a setting or if Windhawk/Explorer restarts mid-session.

Louis047: Actually, it is necessary as when disabling and re-enabling mod, this ensures the following:
- During the mod disable event, it prompts the user for explorer restart to revert back the changes to defaults so that the users can use the disabled shortcuts normally
- During the mod enable event, prompts the user for explorer restart to re-apply the modified shortcuts

Louis047: I have also fixed the explorer restart issue where sometimes, it takes a while for mod un-initialization and re-initialization, will make the commits soon after further testing. Added new shortcuts as well. Updated the PR description.

m417z: > Since by default all the mod's settings are 'off'...

So why not check that instead? They can be all off at other times too.

> will make the commits soon after further testing

Let me know when it's ready for review.

Louis047: I've committed the changes already yesterday, you can check it out.

PR #3910, opened 2026-04-26
Disable Taskbar Thumbnails v1.2

m417z:
* Added an option to disable the virtual desktop switcher flyout that appears when hovering over the Task View button on the taskbar.
* Added an option to disable thumbnails on hover while having a list on click.
* Fixed compatibility with new insider Windows 11 builds.


alexpmorris:
great addition with the click to list! even better, how about making this a toggle? if there's a list, first click shows it, second click hides it again. 

also, maybe still show the window title if there's only a single window, since clicking pops up the window if there's only one, making it tricky to choose the correct one for non-grouped same icons.

m417z:
Maybe in the future, feel free to create an issue for it.

PR #3911, opened 2026-04-26
Add Restore Windows 10 Emojis For Windows 11 Mod

Louis047:
- Safely restores the classic Windows 10 flat emojis on Windows 11.
- Downloads and installs fresh Windows 10 and 11 font files directly into C:\Windows\Fonts.
- Bypasses protected file locks by toggling the Windows Registry between these new files instead of touching system files.
- Clears the font cache in the background to apply changes to the system instantly. (although some apps do require a restart on the event of mod install/uninstall)
- Runs completely asynchronously to prevent any Windhawk uninitialization hangs or crashes.

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini
- - [x] Another AI (please specify): Antigravity (Gemini 3.1 Pro (High))

### DIFF
--- a/mods/restore-win-10-emoji-for-win-11.wh.cpp
+++ b/mods/restore-win-10-emoji-for-win-11.wh.cpp
@@ -1,0 +1,218 @@
+// ==WindhawkMod==
+// @id              restore-win-10-emoji-for-win-11
+// @name            Restore Windows 10 Emojis for Windows 11
+// @description     Replaces the Windows 11 Fluent emojis with the classic Windows 10 flat emojis safely via Windows Registry redirection.
+// @version         1.0
+// @author          Lone
+// @github          https://github.com/Louis047
+// @include         explorer.exe
+// @compilerOptions -lurlmon -ladvapi32 -lole32 -lshell32
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Restore Windows 10 Emojis for Windows 11
+
+This mod replaces the Windows 11 fluent emojis with the classic Windows 10 flat emojis safely via Windows Registry redirection.
+
+How it works:
+- The mod downloads fresh copies of both the Windows 10 and Windows 11 emoji fonts.
+- It installs these fonts as distinct files in the system fonts directory to bypass any system file locks.
+- It seamlessly toggles the font registry between these two files when the mod is enabled or disabled.
+- Administrator privileges (UAC prompt) are requested once during installation and uninstallation to modify the registry and clear the font cache.
+- The Emoji Picker instantly applies the changes in the background without requiring an Explorer restart.
+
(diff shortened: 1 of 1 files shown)


m417z:
It makes more sense for it to be a script or a standalone tool than a Windhawk mod, for multiple reasons:
* There's no reason for it to be injected into explorer.exe.
* There's no reason for it to be constantly loaded.
* It does permanent changes to system files, which is against the Windhawk principles.
* It downloads files from external sources, which makes it difficult to review and trust.

Each item by itself can be discussed if it justifies an exception, but all these reasons together show how it's not a good fit for being a Windhawk mod.

Anixx:
So, this is Gemini. Why not tho check it?

Sand216:
Can't you permanently do this by hand? Just uninstall 11's font and install 10's?

ageographynerd19:
- Get 7-zip
- Get a Windows 10 iso
- Mount it
- Open up sources\install.wim in 7-zip
- Navigate to 1\Windows\Fonts\
- Find seguiemj.ttf 
- Drag it to your desktop
- Right click install
:thinking:

Louis047:
@m417z @Anixx @Sand216 @ageographynerd19 

Hello y'all, I apologize for the confusion. I made this mod for the sole reason of automating the process via a single click for the install and uninstall at an instant to apply the changes without a system reboot. And also, this works two-ways: Installing the mod inserts and replaces Windows 10 Emoji Pack system-wide. Uninstalling/Disabling the mod reverts back to Windows 11 Emoji Pack on the flow. I thought it would be better to manage it this way as windhawk enables the users to manage and customize their windows expleirence one click away.

I think that this mod helps in propagating the windows 10 emojis even after a windows update which is known to replace the system files. Please correct me if I'm wrong here


And yeah, your reasons for the mod being a script and that it going against the windhawk principles is very reasonable. But I hope my explanation gives the correct explanation for why its useful from the traditional script method. If not, I'll gladly close the PR and move back to script method. So let me know of your thoughts.


@Sand216 Completely unrelated, may I get an ETA for new Vibranium11 update which fixes the dark context menu issue I posted recently in discord ? Thanks in advance 😄

Louis047:
Hey, still waiting for your responses y'all so that I can proceed with either ways for this mod 👍

As for the question about the source for `ttf`, I tried to obtain those from `winbindex` but it allows the users to download only `exe`, `dll` and `sys` files. So I had to use a third-party as the source which I found randomly in github and seems to be legitimate. So, I would like to host the files from my github for this. 

And as for modifying the critical system files (in this case, only replacing the `seguiemj.ttf`), as @m417z said, it's bad to keep the process running in background just for a simple replacing task. That's why I would like to hear your different approaches for this to make it worthy as a mod. Not just a script.

PR #3927, opened 2026-04-28
Update - Explorer Details View Columns - version 1.2

ernisn:
## Changelog

- Fixed the problem causing the settings to fallback to default during mod initiating.
- Refined width realization and DPI calculation, fixed the problem of minimum width being too wide.
- Duplicated property keys are handled only once if the user added same property multiple times in settings.
- Added verbose logging option for debug in case needed.
- Fixed some other problems.

### DIFF
--- a/mods/explorer-force-details-columns.wh.cpp
+++ b/mods/explorer-force-details-columns.wh.cpp
@@ -2,7 +2,7 @@
 // @id              explorer-force-details-columns
 // @name            Explorer Details View Columns
 // @description     Forces a fixed set of columns in File Explorer Details view. Columns and their order are configurable. Has no effect on other view modes.
-// @version         1.1
+// @version         1.2
 // @author          ernisn
 // @github          https://github.com/ernisn
 // @include         explorer.exe
@@ -34,6 +34,7 @@ For a full list of available Shell property names, see: https://learn.microsoft.
 
 **Note:**
 - Any property name not recognised by Windows will be skipped.
+- Duplicate property entries will only take the first occurrence.
 - To allow a column to be freely resized, turn off "Force Width".
 - Changes should take effect immediately to opened folders, but if the opened window is on **another monitor with a differnt DPI settings**, the width will be updated **after re-opening the folder once**.
 */
@@ -61,6 +62,9 @@ For a full list of available Shell property names, see: https://learn.microsoft.
     - width: 60
   $name: Columns
   $description: "Columns to show in order in Details view. Width is in logical pixels at 100% DPI scaling."
+- verbose_logging: false
+  $name: Verbose Logging
+  $description: Enable detailed logs for debugging. Logs are written to Windhawk's log output. Leave off for normal use.
 */
 // ==/WindhawkModSettings==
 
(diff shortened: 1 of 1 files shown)


mods/explorer-force-details-columns.wh.cpp
```diff
@@ -61,6 +62,9 @@ For a full list of available Shell property names, see: https://learn.microsoft.
     - width: 60
   $name: Columns
   $description: "Columns to show in order in Details view. Width is in logical pixels at 100% DPI scaling."
+- verbose_logging: false
+  $name: Verbose Logging
+  $description: Enable detailed logs for debugging. Logs are written to Windhawk's log output. Leave off for normal use.
```
m417z: `Wh_Log` logs are already disabled by default, and if somebody enables logs, they might as well get the verbose logs. I'm not sure it warrants another user-facing option in the settings.

Especially for errors, I'd expect to have them logged at all times when logs are enabled.

mods/explorer-force-details-columns.wh.cpp
```diff
@@ -360,7 +476,15 @@ BOOL Wh_ModInit() {
 }
 
 void Wh_ModAfterInit() {
-    Wh_Log(L"AfterInit - applying to existing windows");
+    Wh_Log(L"AfterInit (loaded=%d, count=%zu)",
+           g_settingsLoaded ? 1 : 0, g_columns.size());
+
+    // If settings weren't ready during ModInit after restart, try one more time
+    if (!g_settingsLoaded) {
```
m417z: Why would settings not be ready during `Wh_ModInit`?

mods/explorer-force-details-columns.wh.cpp
```diff
@@ -91,47 +96,104 @@ struct ColumnEntry {
...
+    int duplicatesSkipped = 0;
+    int unrecognisedSkipped = 0;
+    bool sawAnyEntry = false;
 
     for (int i = 0; ; i++) {
         PCWSTR rawProp = Wh_GetStringSetting(L"columns[%d].property", i);
-        if (!rawProp || rawProp[0] == L'\0') {
+
+        // Distinguish "settings not ready / no more entries" from "empty string"
+        // Wh_GetStringSetting returns nullptr if the index is out of range, or an empty string if the field exists but is blank
```
m417z: `Wh_GetStringSetting` never returns `nullptr`, it returns an empty string in both cases.

mods/explorer-force-details-columns.wh.cpp
ernisn: I was trying to figure out why sometimes the settings are not applied for a short time and miss understood the reason. I have changed this part back.

mods/explorer-force-details-columns.wh.cpp
```diff
@@ -167,10 +189,24 @@ static UINT ScaleToDpi(int logicalPx, UINT dpi) {
 // ---------------------------------------------------------------------------
 
 static void ApplyForcedColumns(void* pThis) {
+    // Check empty list 
+    if (g_columns.empty()) {
+        Wh_Log(L"ApplyForcedColumns: empty column list");
+        LoadSettings();
```
m417z, line 195: `LoadSettings` is already called both on init and when settings change, there should be no reason to call it at other times. Is there a reason you added it?

In any case, `g_columns.empty()` shouldn't normally happen unless the user clears all settings, and calling `LoadSettings` again does no harm, just redundant. You might want to delete it in the next update if it's indeed unnecessary.

Issue #3926, opened 2026-04-28
[taskbar-desktop-indicator] Compatibility with Newer Win11 Builds


Meteony:
Mod author: @simon-ami 
Newer builds of windows 11 introduced a bunch of changes to the taskbar (new systray.dll, etc). The old hooking method doesn't find the right handles without fixing. As in the picture enabling the mod doesn't show the indicators in the taskbar . No error logs were generated.  

<img width="661" height="248" alt="Image" src="https://github.com/user-attachments/assets/4a85609b-e372-4d25-a17c-1bc8d0792e60" />

(reference patch)
https://github.com/m417z/my-windhawk-mods/commit/ac90ab21bff67e8e41a044e26edb22c0a653c424

Meteony:
<img width="912" height="567" alt="Image" src="https://github.com/user-attachments/assets/604f48bc-a0f7-4483-a284-3b54bd746565" />
I submitted a fix. Built and tested on both ends. 
https://github.com/simon-ami/windhawk-mods/pull/2

simon-ami:
Hi Meteony, I'm super tight on time rn but hope to be able to take a look at this soon.

mattsigal:
I am pretty sure this also effects other taskbar related mods. Since my computer updated, the following no longer work:

* Taskbar height and icon size
* Taskbar tray icon spacing and grid
* Taskbar tray system icon tweaks

Would this fix also apply here?

m417z:
@mattsigal the mods you listed were already updated to support `SystemTray.dll`. I assume that your issue it the unavailability of debug symbols for your Windows 11 build. See this issue and my comment here: https://github.com/ramensoftware/windhawk/issues/922#issuecomment-4366820200.

mattsigal:
Will do, thanks for your reply!

PR #3932, opened 2026-04-28
Add Taskbar Virtual Desktop Switcher v1.0

sb4ssman:
## Summary

Adds a new mod: **Taskbar Virtual Desktop Switcher** (`taskbar-vd-switcher`).

Injects a grid of numbered buttons into the system tray — one per virtual desktop — for direct switching without opening Task View. Buttons auto-arrange into rows based on taskbar height and update live as desktops are added, removed, or switched.

## Features

- **Five placement positions** within `SystemTrayFrameGrid` (after clock, before clock, before OmniButton, before notification icons, after Show Desktop)
- **Auto-sizing grid** — row count detected from taskbar height, or set manually
- **Live updates** via `IVirtualDesktopNotificationService` — highlighting, button count, and desktop names all update without restart
- **Tooltips** — hover shows the desktop's display name (read from registry `Desktops\{GUID}\Name`, falls back to "Desktop N")
- **Label formats** — numbers, roman numerals, filled/empty dots, or custom comma-separated labels
- **Full visual customization** — active/inactive background color, text color, font size, corner radius, opacity, shine gradient, border color/thickness, padding, bold active label
- **Hide when single** — optionally suppress the bar when only one desktop exists

## Technical notes

- Hooks `IconView::IconView` from `Taskbar.View.dll` for injection timing (pattern from vertical-omnibutton)
- `SystemTrayFrameGrid` uses Grid column layout; injection inserts a `ColumnDefinition` and shifts existing element columns rather than relying on Children insertion order
- `SwitchToDesktop` uses COM vtable calls with build-specific IIDs (pattern from taskbar-empty-space-clicks)
- Registry reading for desktop state from session-scoped `VirtualDesktopIDs` / `CurrentVirtualDesktop` (pattern from taskbar-desktop-indicator)
- Supports explorer builds 22000–26100+

### DIFF
--- a/mods/taskbar-vd-switcher.wh.cpp
+++ b/mods/taskbar-vd-switcher.wh.cpp
@@ -1,0 +1,1311 @@
+// ==WindhawkMod==
+// @id              taskbar-vd-switcher
+// @name            Taskbar Virtual Desktop Switcher
+// @description     Injects clickable buttons into the taskbar — one per virtual desktop — for direct switching. Grid layout adapts to taskbar height.
+// @version         1.0
+// @author          sb4ssman
+// @github          https://github.com/sb4ssman
+// @include         explorer.exe
+// @architecture    x86-64
+// @compilerOptions -lole32 -loleaut32 -lruntimeobject -lversion -luuid
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Taskbar Virtual Desktop Switcher
+
+Adds numbered buttons to the system tray — one per virtual desktop. Click to switch directly.
+
+Buttons auto-arrange into a grid when the taskbar is tall enough for multiple rows.
+
+## Settings
+- **Position** — five positions within the system tray
+- **Size** — button width × height in pixels; spacing between buttons
+- **Rows** — 0 = auto (detected from taskbar height), or a fixed count
+- **Active color** — hex color for the current-desktop button (e.g. `#4488FF`)
+- **Inactive color** — hex color for other buttons (empty = system default)
+- **Opacity** — 0–100; lower values let the taskbar show through
+- **Shine effect** — subtle gradient highlight (applies when a color is set)
+- **Label format** — numbers, roman numerals, dots, or custom comma-separated labels
+- **Padding** — left and right margin around the button grid (px)
+- **Text color** — foreground color for active and inactive buttons
+- **Font size** — button label size in pt
+- **Corner radius** — rounded corners (px)
+- **Active bold** — bold the current desktop's label
+- **Border** — border color and thickness
+- **Hide when single** — hide the bar when only one desktop exists
+- **Tooltips** — hover a button to see the desktop name
+
+## Known limitations
+- Multi-monitor: only the primary taskbar gets buttons.
+*/
+// ==/WindhawkModReadme==
+
+// ==WindhawkModSettings==
+/*
+- position: "afterClock"
+  $name: Position
+  $description: Where in the system tray to inject the VD buttons
+  $options:
+  - "afterClock": "After clock (before Show Desktop)"
+  - "beforeClock": "Before clock (after OmniButton)"
+  - "beforeOmni": "Before OmniButton (wifi/vol/bat)"
+  - "beforeIcons": "Before notification icons"
+  - "afterShowDesktop": "After Show Desktop strip"
+
+- buttonWidth: 20
+  $name: Button width (px)
+
+- buttonHeight: 22
+  $name: Button height (px)
+
+- buttonSpacing: 2
+  $name: Button spacing (px)
+  $description: Gap between buttons in the grid
+
+- buttonRows: 0
+  $name: Button rows (0 = auto from taskbar height)
+
+- activeColor: "#4488FF"
+  $name: Active desktop color (hex, empty = system default)
+
+- inactiveColor: ""
+  $name: Inactive button color (hex, empty = system default)
+
+- buttonOpacity: 100
+  $name: Button opacity (0–100)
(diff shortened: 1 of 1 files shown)


mods/taskbar-vd-switcher.wh.cpp
```diff
@@ -0,0 +1,1321 @@
...
+
+HMODULE WINAPI LoadLibraryExW_Hook(LPCWSTR path, HANDLE file, DWORD flags) {
+    HMODULE h = LoadLibraryExW_Original(path, file, flags);
+    if (h && path && !g_taskbarViewDllLoaded) {
+        const wchar_t* base = wcsrchr(path, L'\\');
+        base = base ? base + 1 : path;
+        if (_wcsicmp(base, L"Taskbar.View.dll") == 0) {
+            // Taskbar.View.dll
+            WindhawkUtils::SYMBOL_HOOK taskbarViewHooks[] = {{
+                {LR"(public: __cdecl winrt::SystemTray::implementation::IconView::IconView(void))"},
```
m417z: In newer builds, this function was moved to `SystemTray.dll`, see:
https://github.com/ramensoftware/windhawk-mods/issues/3926

Some mods were updated to support both older and newer builds. Please consider updating your mod as well.

m417z:
Thanks for the submission, great mod.

Two things I noticed in a quick test session:
1. Sometimes only half of the button is clickable, see video below
2. Disabling the mod crashes explorer

https://github.com/user-attachments/assets/a83a1377-e77e-48dd-b32f-a52fbe035856

mods/taskbar-vd-switcher.wh.cpp
```diff
@@ -0,0 +1,1321 @@
...
+#if defined(_M_X64)
+    {
+        const BYTE* b = (const BYTE*)TaskbarHost_FrameHeight_Original;
+        if (b[0]==0x48 && b[1]==0x83 && b[2]==0xEC && b[4]==0x48 &&
+            b[5]==0x83 && b[6]==0xC1 && b[7]<=0x7F)
+            offset = b[7];
+        else
+            Wh_Log(L"Unsupported TaskbarHost::FrameHeight");
+    }
+#endif
```
m417z: Use this updated code with 0x10 default offset and ARM64 support:
https://github.com/m417z/my-windhawk-mods/blob/e1261d85c2f42006b0dc355fbbc3a8d71a078585/mods/taskbar-multirow.wh.cpp#L261-L291

mods/taskbar-vd-switcher.wh.cpp
```diff
@@ -0,0 +1,1321 @@
...
+        if (disp) {
+            auto _ = disp.RunAsync(winrt::Windows::UI::Core::CoreDispatcherPriority::Normal,
+                [grid, parent, col, hSelf]() {
+                    DoUninitRemove(parent, grid, col);
+                    if (hSelf) FreeLibrary(hSelf);
+                });
+            return;
+        }
+    } catch (...) {}
+    if (hSelf) FreeLibrary(hSelf);
```
m417z: That might be related to the crashes. Why call `FreeLibrary` here?

mods/taskbar-vd-switcher.wh.cpp
```diff
@@ -0,0 +1,1321 @@
...
+        ApplyAllSettingsOnWindowThread();
+
+    std::thread([]() {
+        for (int i = 0; i < 5 && !g_unloading; i++) {
+            Sleep(2000);
+            if (g_buttonGrid) break;
+            Wh_Log(L"[AfterInit] Retry %d", i + 1);
+            ApplyAllSettingsOnWindowThread();
+        }
+    }).detach();
```
m417z: Is this really necessary? It will cause a crash if the mod is unloaded at this time.

sb4ssman:
Thanks for the feedback. I will investigate.

Salyts:
@sb4ssman, Great mod. How about adding the options “Next to the Start button” and “Above the Start button” to the position settings? It would look great with the “Hide Start button” mod enabled.

sb4ssman:
> @sb4ssman, Great mod. How about adding the options “Next to the Start button” and “Above the Start button” to the position settings? It would look great with the “Hide Start button” mod enabled.

Thank you! I haven't have much time to spend on mods, but I think I've got settings that will let you achieve the effect you want. Working on getting it polished and published.

sb4ssman:
Updated the PR branch with v1.5.

This pass addresses the earlier stability/clickability concerns and adds the requested Start-area placement options:
- hardened injection timing so the tray XAML is loaded before modifying it
- improved unload/settings-change cleanup for notification threads and WinRT revokers
- moved desktop switching work off the taskbar UI thread to avoid reentrant XAML updates during clicks
- added Start-adjacent modes: left of Start, over Start, and right of Start with reserved space
- added the master button/sliver layout options and tightened grid positioning/logging

m417z:
The non-clickable buttons are still an issue: https://github.com/ramensoftware/windhawk-mods/pull/3932#issuecomment-4361837429.

I can merge it without the fix, but getting it fixed in a future update will be nice.

sb4ssman:
Moving the mouse away and back seems to relieve it. Fixing it is a priority. I've got a few more mods on deck too.

sb4ssman:
Hi @m417z — this mod was submitted under a placeholder ID before the name was finalized. I've now resubmitted as **Virtual Desktop Switcher** (`virtual-desktop-switcher`) in PR #4484 with a cleaner ID, updated documentation, full settings table, and several improvements since v1.5. Would you be able to disable or remove `taskbar-vd-switcher` from the store once #4484 is reviewed? Thanks!

m417z:
Hi, the id can't be changed. The rest can be updated.

PR #3941, opened 2026-04-30
✨ feat: add win11 round corner killer

HowXu:
## Changelog

* Changelog item 1...
* Changelog item 2...

## Mod authorship

This mod was created by:

- - [ ✔] Manually by the submitter (with or without AI assistance)
- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini

Add a new mod whick modify the windows round corners. May not work in some early versions, but on my device windows11 26100.3775 work well

### DIFF
--- a/mods/win11-round-corner-killer.wh.cpp
+++ b/mods/win11-round-corner-killer.wh.cpp
@@ -23,7 +23,7 @@ std::vector<PatchedFloat> g_patchedFloats;
 const float kNearZeroRadius = 0.001f; // 0f - 1f is also used 
 
 // locate .rdata
-IMAGE_SECTION_HEADER* FindSection(HMODULE hModule, const char* sectionName) {
+IMAGE_SECTION_HEADER* find(HMODULE hModule, const char* name) {
     auto dosHeader = (IMAGE_DOS_HEADER*)hModule;
     if (dosHeader->e_magic != IMAGE_DOS_SIGNATURE) return nullptr;
 
@@ -32,8 +32,8 @@ IMAGE_SECTION_HEADER* FindSection(HMODULE hModule, const char* sectionName) {
 
     auto section = IMAGE_FIRST_SECTION(ntHeaders);
     for (int i = 0; i < ntHeaders->FileHeader.NumberOfSections; i++, section++) {
-        // 比较段名 (如 ".rdata")
-        if (strncmp((const char*)section->Name, sectionName, 8) == 0) {
+        // compare
+        if (strncmp((const char*)section->Name, name, 8) == 0) {
             return section;
         }
     }
@@ -51,7 +51,7 @@ BOOL Wh_ModInit() {
     }
 
     // find .rdata
-    auto rdataSection = FindSection(hUdwm, ".rdata");
+    auto rdataSection = find(hUdwm, ".rdata");
     if (!rdataSection) {
         Wh_Log(L"Failed to locate .rdata section");
         return FALSE;
@@ -72,10 +72,10 @@ BOOL Wh_ModInit() {
                 // for unload
                 g_patchedFloats.push_back({ptr, *ptr});
                 
-                // change
+                // change the vaule
                 *ptr = kNearZeroRadius;
                 
-                // revover
+                // recover protect
                 VirtualProtect(ptr, sizeof(float), oldProtect, &oldProtect);
                 
                 Wh_Log(L"Patched float at %p", ptr);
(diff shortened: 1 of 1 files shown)


m417z:
How does it differ from https://windhawk.net/mods/disable-rounded-corners?

HowXu:
> How does it differ from https://windhawk.net/mods/disable-rounded-corners?

[disable-rounded-corners](https://windhawk.net/mods/disable-rounded-corners) use [Win11DisableRoundedCorners](https://github.com/valinet/Win11DisableRoundedCorners), which was updated 5 years ago. Win11DisableRoundedCorners was a work of valinet(one main author of ExplorerPatcher) and is one default way of EP's. However, it only changes the return value of GetEffectiveCornerStyle in uDWM.dll, the style of window round corner has been cached. Thus it will cause some problems, like in the right click context menu, there will be a white outline.

图片

I use another way [Here the repo](https://github.com/rich-ayr/win11-toggle-rounded-corners), directly modify the value in .rdata selction to disable the round corner, this work fine and there's no white outline.

Ideally, only the filled-corner functions use the 4.0f and 8.0f fields at .rdata, so such changes will be less damaging to dwm, and the repository, which has many users (including me), has not had crash issues for a long time, and can be considered more stable

m417z:
> in the right click context menu, there will be a white outline

Is this the only advantage of your mod?

> Ideally, only the filled-corner functions use the 4.0f and 8.0f fields at .rdata

Ideally, but there might be other usages, now or in the future. So this approach is more risky.

Also, have you tried the [Custom Window Corner Radius](https://windhawk.net/mods/custom-corner-radius) mod? Does it also have the white outline?

HowXu:
> 
> Is this the only advantage of your mod?
> 
Modifying window rounded corners is just a very small feature. Its implementation essentially only involves changing a certain value or modifying a certain function. From all perspectives, this Windhawk mod is more perfect — no need to install the entire ExplorerPatcher, no need to manually set up startup entries like win11-toggle-rounded-corners (thanks to Windhawk), and the visual presentation is absolutely flawless.

one more thing, the outline is really ugly.


> Ideally, but there might be other usages, now or in the future. So this approach is more risky.
> 

To be honest, no one can guarantee what will happen in the future. However, based on my development experience, whether it's ExplorerPatcher, Win11DisableRoundedCorners, or the Windhawk plugin built on top of it, directly making GetEffectiveCornerStyle return 0 is actually more likely to cause issues. In some places, whether the function returns zero directly affects branching logic — just like how we often write if (xxx == 0) else ....

My mod adjusts this value to 0.001f instead of 0. No one writes code that specifically checks whether a static field falls within a certain range (it's obviously fixed). Therefore, in the vast majority of cases, modifying this field only leads to quantitative changes, not qualitative ones. From the goal perspective, it doesn't aim to achieve perfect rounded corners — instead, it makes the corners extremely slightly rounded, visually approximating right angles. Even if these values are referenced elsewhere, it would only cause minor issues (like the cursor size being slightly reduced — just as an analogy; this specific issue doesn't actually occur).

Meanwhile, the lower risk I mentioned is also supported by the fact that the win11-toggle-rounded-corners project already has 300 stars yet almost no open issues.

> Also, have you tried the Custom Window Corner Radius mod? Does it also have the white outline?
> 

Before I created this mod, I was using Custom Window Corner Radius. That mod only allows lowering the corner radius to a minimum of 1f, so it can cause some minor issues with certain applications (on my machine, OOPZ had some problems). More specifically, Custom Window Corner Radius wasn't designed to eliminate rounded corners — it never claimed in its description that it could do so. It's meant for users who want to make corners slightly smaller or larger.

My goal, on the other hand, is to eliminate rounded corners entirely. Users download this mod, the corners disappear; turn it off, the corners come back. It's that simple. That said, I should probably mention in the description that this mod may conflict with Custom Window Corner Radius.

Finally, thank you for your reply.

m417z:
There are two reasons why I'm reluctant to merge this mod:

* User perspective - I don't think there should be two mods doing exactly the same thing, unless there's a good justification for it.
* Implementation - I agree about "no one can guarantee what will happen in the future", but with your approach, not only the mod may stop working, it may also corrupt something totally unrelated to the intended functionality, from other code that uses this constant, to some data that's not even a floating point value that contains these bytes.

I think that the best solution in this case is to update the existing "Disable rounded corners in Windows 11" mod by adding an option that allows to use the "Custom Window Corner Radius" approach which uses a tiny ~0.001 value. Basically, similar to your approach but with function hooking and not data scanning and patching.

HowXu:
Got it. Maybe I should open a pr to Disable rounded corners in Windows 11. Thank you for your reply. I will close this pr.

m417z:
You can submit a PR here: https://github.com/m417z/my-windhawk-mods
Should be fairly straightforward by merging the "Custom Window Corner Radius" mod functionality into "Disable rounded corners in Windows 11" with a small ~0.001 radius value.

PR #3969, opened 2026-05-02
Enhanced Disk Usage 1.0.1 - fixed middle click to open new file explorer windows, reordered settings for better flow

bbmaster123:
## Changelog

* Changelog item 1...
* Changelog item 2...

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini

middle click to open new file explorer windows wasn't working when "launch folder windows in a separate process" was not enabled. I usually have it enabled so I hadn't caught it until now. This update changes the way windows are refreshed, which avoids a deadlock in that situation.

I've also reordered the settings to flow better/grouped together which I think makes it a bit better overall.


bbmaster123:
wait 18 commits? I thought I was just trying to merge the 2 commits I made...did I mess up?

m417z:
It's fine as long as what you see in the "Files changed" tab is correct.

bbmaster123:
sorry I'll work on it a bit more and resubmit...

m417z:
Sure, let me know when it's ready.

bbmaster123:
@m417z 
sorry about that, should be good to go now. I don't believe it will need any more changes until I receive feedback from users. Thanks for helping me with this <3

m417z:
> 1px should = 1px, not (1px * scale)

Why? I think using (1px * scale) is more helpful. For example, if somebody uses multiple monitors with different DPIs (scales), using (1px * scale) will yield a similar result in all monitors, while not taking scale into account will make the corner radius and other sizes look different.

bbmaster123:
@m417z 
hmm good point, as a single monitor user I didn't consider this.
The original reason for the change was to improve the sharpness of the border when using a custom scaling factor (I use 169%). 

As an alternative, I can revert this change and instead add a custom scale factor setting, or probably more simple to just have finer grained border controls, which should give pretty much the same effect.

m417z:
> just have finer grained border controls

I think that's the simplest solution, e.g. you can allow to configure quarters of pixels, so a value of 12 will mean 3px on 100% DPI.

bbmaster123:
@m417z 
hey, I believe the update is now ready for a re-review when you have a chance :)

Issue #3657, opened 2026-03-27
[Logon & Sleep Fade Restorer] Mod prevents screen from turning off when session is locked after user is logged on


Skyyblaze:
@Ingan121 Thanks for the mod! I noticed a small but annoying bug with it. When Logon & Sleep Fade Restorer is enabled the screen won't turn off on the lock-screen when you lock a session after a user is logged on. The screen will fade to black after the Windows timeout and then the screen wakes up immediately again.

This does not happen when the screen turn-off time out is hit on the lock-screen before a user is logged on or if the system is idle without locking the session. In both of these scenarios the screen turns off correctly without waking up immediately.

I'm running version 1.3 of Logon & Sleep Fade Restorer on Windows 11 25H2 26200.8037.

Skyyblaze:
@Ingan121 Sorry for the late response! I was busy over Easter and then forgot to reply. Disabling the Lockscreen fixes the issue. It is also disabling the fade in this particular instance however if I lock the screen because I'm stepping away from my computer I won't see the fade anyhow. So it seems the problem is caused by the Lockscreen in this case.

Skyyblaze:
Thank you for the fix!

PR #3978, opened 2026-05-03
Logon & Sleep Fade Restorer

Ingan121:
* Fix monitor not turning off when the Windows 10+ lock screen is being displayed
* Fix monitor-off fade not working when a 32-bit app's window is focused

## Changelog

* Changelog item 1...
* Changelog item 2...

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/win7-login-fade.wh.cpp
+++ b/mods/win7-login-fade.wh.cpp
@@ -2,14 +2,15 @@
 // @id              win7-login-fade
 // @name            Logon & Sleep Fade Restorer
 // @description     Bring back the old logon screen and sleep fade effect
-// @version         1.3
+// @version         1.4
 // @author          Ingan121
 // @github          https://github.com/Ingan121
 // @twitter         https://twitter.com/Ingan121
 // @homepage        https://www.ingan121.com/
 // @include         winlogon.exe
+// @include         LogonUI.exe
+// @include         LockApp.exe
 // @include         *
-// @architecture    x86-64
 // @compilerOptions -lgdi32
 // ==/WindhawkMod==
 
@@ -47,6 +48,12 @@
 * Turning off the monitor with the power button action (or sleeping on Modern Standby devices) will not trigger the sleep fade effect, as it is handled internally by the kernel without notifying a user-mode component before initiating the action.
   * Monitor off initiated by the idle timer can have fade added as usual.
 * Only `Gamma (Kernel)` and `DWM (Original)` logon/logoff fade types are supported on Windows 8 for now, to avoid critical issues that I have observed. Early Windows 10 versions have not been thoroughly tested, so it is recommended to avoid `None` and `Gamma (Reimplemented)` modes on those versions as a precaution, until I can confirm their stability.
+* On 32-bit systems, only the monitor off and sleep/hibernation fades are supported, as this mod was never tested on 32-bit Windows systems.
+## Miscellaneous
+* To allow fading in the logon screen when turning off the monitor with the idle timer, you must allow Windhawk to inject into the **LogonUI.exe**.
+* To do so, add it to the process inclusion list in the advanced settings.
+
+![Advanced settings screenshot](https://i.imgur.com/LRhREtJ.png)
 */
 // ==/WindhawkModReadme==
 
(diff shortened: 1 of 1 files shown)


Ingan121:
Winlogon hooks are disabled on 32bit as this mod was never tested on a real 32bit Windows. (Testing this mod in a VM is hard as gamma doesn't work.)

Ingan121:
Fixes #3657

mods/win7-login-fade.wh.cpp
```diff
@@ -408,38 +422,64 @@ __int64 __fastcall SwitchDesktopWithFade_hook(HDESK hDesktop, DWORD duration, DW
...
+bool TurnOffMonitor(HWND initiatorWindow) {
+    NTSTATUS res = NtPowerInformation_original(ScreenOff, NULL, 0, NULL, 0);
+    Wh_Log(L"Called original NtPowerInformation with ScreenOff, result=0x%X", res);
+    if (res != 0) {
+        Wh_Log(L"NtPowerInformation failed to turn off the monitor");
+        if (initiatorWindow && IsWindow(initiatorWindow)) {
+            // Some sandboxed/immersive processes like LockApp.exe may fail to call NtPowerInformation(ScreenOff)
+            g_keepOrigWndProc.store(true);
+            if (!PostMessage(initiatorWindow, WM_SYSCOMMAND, SC_MONITORPOWER, 2)) {
+                g_keepOrigWndProc.store(false);
```
m417z: `PostMessage` only enqueues the message, so using a variable like this won't work. Maybe it makes sense to use `SendMessage` here.

mods/win7-login-fade.wh.cpp
```diff
@@ -408,38 +422,64 @@ __int64 __fastcall SwitchDesktopWithFade_hook(HDESK hDesktop, DWORD duration, DW
...
+LRESULT WINAPI OverlayWndProc(HWND hWnd, UINT uMsg, WPARAM wParam, LPARAM lParam) {
     switch (uMsg) {
         case WM_CREATE:
             Wh_Log(L"OverlayWndProc WM_CREATE");
             return 0;
         case WM_MONOFFTASK:
             Wh_Log(L"OverlayWndProc WM_MONOFFTASK");
             Sleep(100); // Needed to prevent flashing when monitor fade is started multiple times in the same process
-            RestoreGammaAndMonitorOff();
+            RestoreGammaAndMonitorOff((HWND)lParam);
```
m417z, line 473: Isn't `lParam` always zero here?

mods/win7-login-fade.wh.cpp
```diff
@@ -408,38 +422,64 @@ __int64 __fastcall SwitchDesktopWithFade_hook(HDESK hDesktop, DWORD duration, DW
...
     return result;
 }
+#endif
 
 #pragma region Hooks and functions for monitor off fading
 typedef NTSTATUS (NTAPI* NtPowerInformation_t)(POWER_INFORMATION_LEVEL, PVOID, ULONG, PVOID, ULONG);
 NtPowerInformation_t NtPowerInformation_original;
 
+typedef HWND (WINAPI* CreateWindowInBand_t)(DWORD dwExStyle, LPCWSTR lpClassName, LPCWSTR lpWindowName, DWORD dwStyle, int X, int Y, int nWidth, int nHeight, HWND hWndParent, HMENU hMenu, HINSTANCE hInstance, LPVOID lpParam, DWORD dwBand);
+CreateWindowInBand_t CreateWindowInBand;
```
m417z: Seems to be unused.

mods/win7-login-fade.wh.cpp
```diff
@@ -468,20 +509,34 @@ HWND CreateOverlayWindow() {
...
 
-        HWND overlayWnd = CreateOverlayWindow();
+        HWND overlayWnd = NULL;
+
+        if (initiatorWindow && IsWindow(initiatorWindow)) {
+            wchar_t title[256];
+            GetWindowTextW(initiatorWindow, title, 256);
+            // Disable the overlay mechanism for LockApp.exe as it often gets suspended when the monitor is off or after unlocking
+            // which makes hiding the overlay at the right time hard. If failed, it will get stuck on the screen, as LockApp is kept suspended even after unlocking
+            if (wcscmp(title, L"Windows Default Lock Screen") != 0) {
```
m417z: Titles are often localized. Did you check it? If it's a known limitation, I think it's worth documenting it if finding another way is not easy.

mods/win7-login-fade.wh.cpp
Ingan121: Yeah I confused them

mods/win7-login-fade.wh.cpp
Ingan121: It was still English in my Korean environment, and since the title is not visible at all in normal usage, I think it's the same as the `Program Manager` title from `explorer.exe`. Changed to exe check anyway as I was not 100% sure if it's still the case in all languages, and I gave up on analyzing the WinRT binaries to check if it's really hardcoded.

PR #3983, opened 2026-05-04
Support SystemTray.dll for taskbar desktop indicator

### LINKED ISSUE #3926: [taskbar-desktop-indicator] Compatibility with Newer Win11 Builds

simon-ami:
Adds compatibility for newer Windows 11 builds where the taskbar SystemTray symbols moved to SystemTray.dll.

The mod now tries SystemTray.dll first, then falls back to Taskbar.View.dll and ExplorerExtensions.dll for older builds. It also keeps the existing post-init refresh/update path intact and guards the hook attempt so it won't install the same hooks twice.

Tested on my older Taskbar.View.dll path without regressions. @Meteony tested the SystemTray.dll path on a newer build and confirmed it works.

Fixes #3926

## Changelog

* Added support for newer Windows 11 builds where taskbar SystemTray symbols are in SystemTray.dll.
* Kept fallback support for older Taskbar.View.dll and ExplorerExtensions.dll builds.

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/taskbar-desktop-indicator.wh.cpp
+++ b/mods/taskbar-desktop-indicator.wh.cpp
@@ -2,7 +2,7 @@
 // @id              taskbar-desktop-indicator
 // @name            Taskbar Desktop Indicator
 // @description     Displays the current virtual desktop as a number or marker in the Windows 11 taskbar clock area
-// @version         1.2.1
+// @version         1.2.2
 // @author          Simon Benedict
 // @github          https://github.com/simon-ami
 // @include         explorer.exe
@@ -244,7 +244,7 @@ struct VirtualDesktopNotificationObject {
 WinVersion g_winVersion = WinVersion::Unsupported;
 WORD g_explorerBuildNumber = 0;
 WORD g_explorerRevisionNumber = 0;
-std::atomic<bool> g_taskbarViewDllLoaded = false;
+std::atomic<bool> g_systemTrayModuleHooked = false;
 std::atomic<bool> g_unloading = false;
 std::atomic<int> g_currentDesktopNumber = 1;
 std::atomic<int> g_pollIntervalMs = 0;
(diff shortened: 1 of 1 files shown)


mods/taskbar-desktop-indicator.wh.cpp
```diff
@@ -2043,18 +2043,35 @@ HRESULT WINAPI BadgeIconContent_get_ViewModel_Hook(LPVOID pThis, LPVOID pArgs) {
...
         module = GetModuleHandleW(L"ExplorerExtensions.dll");
     }
 
     return module;
 }
 
-bool HookTaskbarViewDllSymbols(HMODULE module) {
+bool HookSystemTraySymbols(HMODULE module) {
+    // Hooks SystemTray.dll when available; falls back to the modules below.
     // Taskbar.View.dll, ExplorerExtensions.dll
```
m417z: Add `SystemTray.dll` to the list. The verification failed on that, it should be fixed now.

mods/taskbar-desktop-indicator.wh.cpp
```diff
@@ -2318,11 +2353,12 @@ BOOL Wh_ModInit() {
 
     HookExplorerExeSymbols();
 
-    if (HMODULE taskbarViewModule = GetTaskbarViewModuleHandle()) {
-        g_taskbarViewDllLoaded = true;
-        if (!HookTaskbarViewDllSymbols(taskbarViewModule)) {
-            return FALSE;
+    if (HMODULE systemTrayModule = GetSystemTrayModuleHandle()) {
+        if (!TryHookSystemTrayModule(systemTrayModule)) {
+            Wh_Log(L"System tray symbols not found in initial module");
```
m417z, line 2356: I think it makes sense to keep the `return FALSE` here.

mods/taskbar-desktop-indicator.wh.cpp
```diff
@@ -2212,15 +2242,20 @@ void RefreshLiveTaskbarClock() {
...
+
+    if (GetSystemTrayModuleHandle() != module) {
+        return;
+    }
+
+    Wh_Log(L"Loaded %s", moduleName);
+
+    if (TryHookSystemTrayModule(module)) {
+        Wh_ApplyHookOperations();
+        RefreshLiveTaskbarClock();
```
m417z: Since the module was just loaded at this point, I don't believe refreshing is necessary, as there's nothing to refresh.

mods/taskbar-desktop-indicator.wh.cpp
simon-ami: Done, restored the explicit SystemTray.dll entry. I verified it passes with the updated validator from main.

mods/taskbar-desktop-indicator.wh.cpp
simon-ami: Done, kept the init failure behavior when the module is already loaded but the symbols can't be hooked.

mods/taskbar-desktop-indicator.wh.cpp
simon-ami: Done, removed the refresh from the module-load path.

m417z:
Thanks.

PR #3985, opened 2026-05-04
Added window border customizer mod

Lockframe:
## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [X] Gemini

Not sure if I should keep or remove the safeguard since it interferes with rapid mod toggling, let me know what you think.

### DIFF
--- a/mods/window-border-customizer.wh.cpp
+++ b/mods/window-border-customizer.wh.cpp
@@ -4,6 +4,7 @@
 // @description     Replaces DWM stock borders with custom ARGB translucent borders.
 // @version         1.0.0
 // @author          Lockframe
+// @github          https://github.com/Lockframe
 // @include         dwm.exe
 // @include         explorer.exe
 // @architecture    x86-64
--- a/mods/window-border-customizer.wh.cpp
+++ b/mods/window-border-customizer.wh.cpp
@@ -4,10 +4,10 @@
 // @description     Replaces DWM stock borders with custom ARGB translucent borders.
 // @version         1.0.0
 // @author          Lockframe
-// @github          https://github.com/Lockframe
 // @include         dwm.exe
 // @include         explorer.exe
 // @architecture    x86-64
+// @compilerOptions -lwevtapi
 // ==/WindhawkMod==
 
 // ==WindhawkModReadme==
@@ -29,7 +29,7 @@ and make sure that `dwm.exe` is in the list.
 ---
 
 ## Contributors
-[m417z](github.com/m417z/)
+[m417z](https://github.com/m417z/), 
 Krisatisa
 */
 // ==/WindhawkModReadme==
@@ -87,11 +87,6 @@ Krisatisa
     - dark_color: "#80202020"
       $name: Dark Color
   $name: Tooltip border
-
-- advanced:
-    - enable_debug_logs: false
-      $name: Enable Debug Logs
-  $name: Advanced
 */
 // ==/WindhawkModSettings==
 
@@ -99,6 +94,7 @@ Krisatisa
 #include <windhawk_utils.h>
 #include <windows.h>
 #include <sddl.h>
+#include <winevt.h>
 #include <unordered_map>
 #include <atomic>
 #include <cmath>
@@ -229,7 +225,6 @@ struct BorderSettings {
     BorderCategorySettings maximized;
     BorderCategorySettings contextMenu;
     BorderCategorySettings tooltip;
-    bool enableDebugLogs;
 };
 
 SRWLOCK g_SettingsLock = SRWLOCK_INIT;
@@ -256,22 +251,6 @@ SRWLOCK g_EnumLock = SRWLOCK_INIT;
 std::atomic<bool> g_ModUnloading = false;
 std::atomic<ULONGLONG> g_LastEnumTick{0};
 
-void LogIfEnabled(const wchar_t* format, ...) {
-    bool shouldLog = false;
-    AcquireSRWLockShared(&g_SettingsLock);
-    shouldLog = g_Settings.enableDebugLogs;
-    ReleaseSRWLockShared(&g_SettingsLock);
-
-    if (shouldLog) {
-        va_list args;
-        va_start(args, format);
-        wchar_t buffer[512];
-        vswprintf(buffer, ARRAYSIZE(buffer), format, args);
-        va_end(args);
-        Wh_Log(L"%s", buffer);
-    }
-}
-
 static DWM_COLOR_VALUE MakeColorFromHEX(PCWSTR hex) {
     DWM_COLOR_VALUE color = {1.0f, 1.0f, 1.0f, 1.0f};
     if (!hex) return color;
@@ -287,6 +266,7 @@ static DWM_COLOR_VALUE MakeColorFromHEX(PCWSTR hex) {
             
             if (pDwmGetColorizationColor && SUCCEEDED(pDwmGetColorizationColor(&colorizationColor, &opaqueBlend))) {
                 color.a = ((colorizationColor >> 24) & 0xFF) / 255.0f;
+                if (color.a <= 0.0f) color.a = 1.0f / 255.0f;
                 color.r = ((colorizationColor >> 16) & 0xFF) / 255.0f;
                 color.g = ((colorizationColor >> 8) & 0xFF) / 255.0f;
                 color.b = (colorizationColor & 0xFF) / 255.0f;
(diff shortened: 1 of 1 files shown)


mods/window-border-customizer.wh.cpp
```diff
@@ -0,0 +1,752 @@
...
+This mod needs to hook into `dwm.exe` to work. Please navigate to Windhawk's
+Settings > Advanced settings > More advanced settings > Process inclusion list,
+and make sure that `dwm.exe` is in the list.
+
+![Advanced settings screenshot](https://i.imgur.com/LRhREtJ.png)
+
+---
+
+## Contributors
+[m417z](github.com/m417z/)
```
m417z:
```suggestion
[m417z](https://github.com/m417z)
```

mods/window-border-customizer.wh.cpp
```diff
@@ -0,0 +1,752 @@
...
+        AcquireSRWLockExclusive(&g_CacheLock);
+        g_WindowCache.clear();
+        ReleaseSRWLockExclusive(&g_CacheLock);
+        
+        InterlockedExchangePointer((PVOID volatile*)&g_SharedState, NULL);
+    }
+
+    SharedThemeState* sharedToUnmap = (SharedThemeState*)InterlockedExchangePointer((PVOID volatile*)&g_SharedState, NULL);
+    if (sharedToUnmap) {
+        Sleep(50);
```
m417z: Why is the sleep needed?

mods/window-border-customizer.wh.cpp
```diff
@@ -0,0 +1,752 @@
...
+
+    HANDLE hFile = CreateFileW(filePath.c_str(), GENERIC_READ | GENERIC_WRITE, FILE_SHARE_READ | FILE_SHARE_WRITE, nullptr,
+                               OPEN_ALWAYS, FILE_ATTRIBUTE_NORMAL, nullptr);
+    if (hFile != INVALID_HANDLE_VALUE) {
+        DWORD bytesRead;
+        if (GetLastError() == ERROR_ALREADY_EXISTS) {
+            if (ReadFile(hFile, &lastTick, sizeof(lastTick), &bytesRead, nullptr) &&
+                bytesRead == sizeof(lastTick)) {
+                if (currentTick - lastTick < 5000) {
+                    Wh_Log(L"CRASH LOOP DETECTED: DWM restarted within 15 seconds. Disabling mod.");
```
m417z: Here's a more reliable way of detecting a dwm crash:
https://gist.github.com/m417z/58d2fffd846c358b961f8c70d4f46c0d

mods/window-border-customizer.wh.cpp
```diff
@@ -0,0 +1,752 @@
...
+    g_hStopEvent = CreateEventW(NULL, TRUE, FALSE, NULL);
+    InitSharedMemory();
+
+    if (IsExplorerProcess()) {
+        g_hExplorerMutex = CreateMutexW(NULL, FALSE, L"Windhawk_CustomBorders_ExplorerMutex");
+        if (GetLastError() == ERROR_ALREADY_EXISTS) {
+            Wh_Log(L"[Explorer Hook] Secondary explorer. Skip theme thread.");
+            CloseHandle(g_hExplorerMutex);
+            g_hExplorerMutex = NULL;
+            return TRUE;
```
m417z: What's the purpose of the process in this case? If it doesn't do anything, it makes more sense to unload the mod by returning FALSE.

mods/window-border-customizer.wh.cpp
```diff
@@ -0,0 +1,752 @@
...
+    - color: "#80FFFFFF"
+      $name: Light Color
+    - dark_color: "#80202020"
+      $name: Dark Color
+  $name: Tooltip border
+
+- advanced:
+    - enable_debug_logs: false
+      $name: Enable Debug Logs
+  $name: Advanced
```
m417z: Windhawk already has logging control, why not just use it with `Wh_Log`? In fact, `LogIfEnabled` is never even called.

mods/window-border-customizer.wh.cpp
```diff
@@ -0,0 +1,752 @@
...
+    }
+
+    bool isDarkTheme = false;
+    bool isFocused = true; 
+
+    SharedThemeState* shared = (SharedThemeState*)InterlockedCompareExchangePointer((PVOID volatile*)&g_SharedState, NULL, NULL);
+    if (shared) {
+        isDarkTheme = (shared->isDark == 1);
+        if (hwnd && popupKind == WindowKind::Normal) {
+            isFocused = (hwnd == shared->focusedHwnd);
```
m417z: Just using `GetForegroundWindow()` doesn't work in dwm.exe?

mods/window-border-customizer.wh.cpp
```diff
@@ -0,0 +1,752 @@
...
+    Tooltip,
+};
+
+struct CacheEntry {
+    WindowKind kind;
+    HWND hwnd;
+    ULONGLONG lastVerifyTick;
+};
+
+std::unordered_map<void*, CacheEntry> g_WindowCache;
```
m417z, line 191: The key - the `void*` value - can be reused if the memory is freed and a new object is allocated at that address.

mods/window-border-customizer.wh.cpp
Lockframe: It's to prevent crashes during mod unload due to concurrent rendering threads in `SetBorderParameters_Hook`, just in case one of them captured a valid pointer just before it was nullified

mods/window-border-customizer.wh.cpp
Lockframe: It'd keep trying to re-inject into explorer while spamming system error logs, creating a loop that'd use unnecessary CPU cycles. Returning it true keeps it loaded but dormant

mods/window-border-customizer.wh.cpp
Lockframe: Calling `GetForegroundWindow()` inside dwm returns `null` due to UIPI, which is why I used explorer instead, since it runs in the active user session and can properly track/catch events, like theme switching, and writes them to memory for dwm to read instead.

mods/window-border-customizer.wh.cpp
m417z: It'll only re-inject when settings change, which doesn't happen often.

mods/window-border-customizer.wh.cpp
m417z: dwm runs in the user session. I was curious so I did a very simple test, and it seems to be working:

image

I suggest to revisit that, as it can make the mod much simpler and potentially more stable.

Lockframe: Since DWM runs as 'DWM-1' user, I managed to get it to detect focused/unfocused windows, but it can't detect theme changes and instead only applies light-mode colors, so I still had to use explorer for that.

image

mods/window-border-customizer.wh.cpp
m417z: I didn't spend enough time to understand how it works and why the sleep is needed, but it looks fragile. How about using a shared dll section instead? I believe it should simplify the code as well. Here's an example:
https://github.com/ramensoftware/windhawk-mods/blob/c76d030cf7c5848b97696ce764c1f5ec7db9d049/mods/taskbar-primary-on-secondary-monitor.wh.cpp#L162-L166

Lockframe: Not sure if that would work. Doesn't windows block PE section sharing across different integrity levels? The current approach applies a custom security descriptor to force windows to allow memory sharing between DWM-1 user profile and the actual user profile

m417z: I don't think a shared memory section is limited in any way by integrity levels.

Lockframe: I've attached a version of the mod with the shared dll method, although it tracks cross-process states, the actual risk is an intra-process thread race condition, so if `g_SharedState` is unmapped while a thread is mid-read, it'd cause an Access Violation error and crash dwm.exe, which is why the sleep is there, to give currently executing hooks enough time to exit the function before the memory is cleared.

Let me know if you think this should be the released version and not the committed one.

[window-border-customizer-shareddll-method.txt](https://github.com/user-attachments/files/27437992/window-border-customizer-shareddll-method.txt)

m417z: Why do you still need `g_SharedState` and `InitSharedMemory` with the shared dll method?

> if `g_SharedState` is unmapped while a thread is mid-read, it'd cause an Access Violation error

I expect each process to have its own mapped view, such that unmapping a view in one process doesn't impact other processes. I didn't look at the code in detail - I don't have the time right now, and I hope it will become unnecessary if the simpler shared dll method works.

PR #4000, opened 2026-05-06
Add AutoScroll Win32

crazyboyybs:
Simulates browser-style autoscroll (middle mouse button) in Win32 applications such as File Explorer, with a WinUI-style visual indicator at the scroll origin.

### How to use

1. **Hold** the **middle mouse button** anywhere in a scrollable area.
2. Move the cursor **up or down** (and optionally sideways) from the origin point -- the farther from the origin, the faster the scroll.
3. **Release** the middle button to stop.

## Mod authorship

This mod was created by:

- - [x] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/autoscroll-win32.wh.cpp
+++ b/mods/autoscroll-win32.wh.cpp
@@ -15,6 +15,11 @@
 
 // ==WindhawkModReadme==
 /*
+
+![Preview](https://raw.githubusercontent.com/crazyboyybs/assets/refs/heads/main/auto%20scroll%20preview.gif)
+
+---
+
 ## AutoScroll Win32
 
 Simulates browser-style autoscroll (middle mouse button) in Win32 applications
@@ -36,8 +41,8 @@ exactly like real mouse-wheel events.
 ### Adding more applications
 
 By default the mod targets only explorer.exe. To enable it in other
-applications, add their executable names in the Windhawk mod settings for
-this mod under "Include processes".
+applications, add their executable names in the **Advanced** tab of the mod
+settings in Windhawk under "Include processes".
 
 ---
 
@@ -63,8 +68,8 @@ eventos reais da roda do mouse.
 ### Adicionar mais aplicativos
 
 Por padrao, o mod afeta apenas o explorer.exe. Para ativa-lo em outros
-aplicativos, adicione os nomes dos executaveis nas configuracoes do mod no
-Windhawk em "Include processes".
+aplicativos, adicione os nomes dos executaveis na aba **Advanced** das
+configuraces do mod no Windhawk em "Include processes".
 
 ---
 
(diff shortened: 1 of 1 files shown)


mods/autoscroll-win32.wh.cpp
```diff
@@ -0,0 +1,1013 @@
...
+static void IndicatorThreadProc() {
+    static const wchar_t kClass[] = L"WH_AutoscrollIndicator";
+
+    WNDCLASSEXW wc{};
+    wc.cbSize        = sizeof(wc);
+    wc.lpfnWndProc   = IndicatorWndProc;
+    wc.hInstance     = GetModuleHandleW(nullptr);
+    wc.lpszClassName = kClass;
+    if (!RegisterClassExW(&wc)
+        && GetLastError() != ERROR_CLASS_ALREADY_EXISTS) {
```
m417z: Don't continue on `ERROR_CLASS_ALREADY_EXISTS`, it might belong to a mod instance that's already freed, with a `lpfnWndProc` pointer that's no longer valid.

Also, set `hInstance` to the mod dll instance. Example:
https://github.com/ramensoftware/windhawk-mods/blob/46e9b560b3ed69055810d70c273df2356d670e28/mods/desktop-live-overlay.wh.cpp#L484-L493

mods/autoscroll-win32.wh.cpp
```diff
@@ -0,0 +1,1013 @@
...
+
+Fully compatible. AutoScroll Win32 posts standard WM_MOUSEWHEEL and
+WM_MOUSEHWHEEL messages, which Smooth Scroll Win32 intercepts and animates
+exactly like real mouse-wheel events.
+
+### Adding more applications
+
+By default the mod targets only explorer.exe. To enable it in other
+applications, add their executable names in the Windhawk mod settings for
+this mod under "Include processes".
```
m417z: Clarify that it's in the "Advanced" tab of the mod.

mods/autoscroll-win32.wh.cpp
```diff
@@ -0,0 +1,1013 @@
...
+
+// ===========================================================================
+// Scroll thread
+// ===========================================================================
+static void ScrollThreadProc() {
+    while (g_threadRunning.load(std::memory_order_relaxed)) {
+
+        HWND hwnd = g_targetHwnd.load(std::memory_order_acquire);
+        if (!hwnd) {
+            Sleep(10);
```
m417z: That seems quite inefficient. Up to you, but I'd consider at least making sure this thread is active only when an explorer window is on the foreground, or, better, only while actually scrolling.

mods/autoscroll-win32.wh.cpp
```diff
@@ -0,0 +1,1013 @@
...
+    pSink->AddLine(tip);
+    pSink->AddLine(b2);
+    pSink->EndFigure(D2D1_FIGURE_END_OPEN);
+    pSink->Close();
+    pSink->Release();
+    return S_OK;
+}
+
+static HRESULT InitIndicatorResources(IndicatorResources& res) {
+    UINT dpi = GetDpiForSystem();
```
m417z: How about per-monitor DPI support?

mods/autoscroll-win32.wh.cpp
```diff
@@ -0,0 +1,1013 @@
...
+configuracion del mod en Windhawk bajo "Include processes".
+*/
+// ==/WindhawkModReadme==
+
+// ==WindhawkModSettings==
+/*
+- deadzone: 8
+  $name: Zona Morta (px)
+  $name:en: Dead Zone (px)
+  $name:es: Zona Muerta (px)
```
m417z, line 109: Make English the default, like with the mod metadata.

mods/autoscroll-win32.wh.cpp
```diff
@@ -0,0 +1,1013 @@
...
+// @version         1.0.0
+// @author          crazyboyybs
+// @github          https://github.com/crazyboyybs
+// @include         explorer.exe
+// @compilerOptions -luser32 -lgdi32 -ld2d1 -ladvapi32
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+## AutoScroll Win32
```
m417z, line 23: A screenshot/gif would be nice.

m417z:
The mod is often stuck on unload because many threads are waiting on `GetMessageW` which the mod hooks. Up to you if you want to address this.

mods/autoscroll-win32.wh.cpp
crazyboyybs: Now we fail hard on any `RegisterClassExW` error instead of silently continuing on `ERROR_CLASS_ALREADY_EXISTS`, a class left over from a previously freed mod instance would carry a stale `lpfnWndProc` pointing to unmapped memory, which would crash on the first message. `hInstance` is now set using `GetModuleHandleExW` with `GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS` so it always resolves to the mod's own DLL handle rather than the host process executable. The same handle is used on `CreateWindowExW ` and `UnregisterClassW`

mods/autoscroll-win32.wh.cpp
crazyboyybs: Fixed in all three language variants of the readme

mods/autoscroll-win32.wh.cpp
crazyboyybs: Replaced the `Sleep(10)` polling loop with `WaitForSingleObject` on an auto-reset event (`g_scrollEvent`). The scroll thread now sleeps at zero CPU cost until `StartAutoscroll` calls `SetEvent`, and `Wh_ModUninit` signals the same event to unblock the thread for clean shutdown. Additionally, `Wh_ModAfterInit` now guards against `CreateEventW` failure, previously a null handle would cause `WaitForSingleObject(nullptr, INFINITE)` to return `WAIT_FAILED` immediately, putting the thread in a tight busy spin

mods/autoscroll-win32.wh.cpp
crazyboyybs: Added per-monitor DPI support

mods/autoscroll-win32.wh.cpp
crazyboyybs: Metadata now uses English as the default

mods/autoscroll-win32.wh.cpp
crazyboyybs: A gif was added

crazyboyybs:
> The mod is often stuck on unload because many threads are waiting on `GetMessageW` which the mod hooks. Up to you if you want to address this.

The root cause was hooking GetMessageW, which is a blocking call. The fix is to remove the GetMessageW/A hooks entirely. Coverage is not lost: the standard Win32 message loop is `GetMessage` to `TranslateMessage` to `DispatchMessage`, and `WM_MBUTTONDOWN` is intercepted at `DispatchMessage`, which is a short-lived, non-blocking call that returns as soon as the `WndProc` finishes. `PeekMessageW/A` hooks cover `PeekMessage`-driven loops. `DispatchMessageW` is now the critical fatal hook, and `GetMessageW` is no longer touched

PR #4002, opened 2026-05-07
Add Remove Taskbar Window Prefixes mod

ryanspice:
﻿
## Local test status

- Windhawk local mod compile: PASS
- Taskbar prefix removal behavior: PASS
- Verified target case: ChatGPT PWA titles like ChatGPT - Windhawk PWA Title Fix
- Expected taskbar result: Windhawk PWA Title Fix
"@ | Add-Content ".\README.md" -Encoding UTF8

@"
Adds a taskbar-title prefix removal mod.

The mod is intended for apps and PWAs that place repeated app names at the beginning of window titles, such as ChatGPT - Conversation title, causing grouped taskbar labels to be less useful.

Features:
- Built-in ChatGPT prefix mode
- Universal first-separator prefix mode
- Simple literal prefix rules
- Advanced regex rules
- Optional process-specific matching by process name, full path, or App ID

Local validation:
- Compiles in Windhawk
- Tested against ChatGPT PWA taskbar titles
- Verified ChatGPT - ... prefix removal from taskbar labels

### DIFF
--- a/mods/remove-taskbar-window-prefixes.wh.cpp
+++ b/mods/remove-taskbar-window-prefixes.wh.cpp
@@ -1,0 +1,738 @@
+﻿// ==WindhawkMod==
+// @id              remove-taskbar-window-prefixes
+// @name            Remove Taskbar Window Prefixes
+// @description     Remove prefixes such as "ChatGPT - " from taskbar window titles, with built-in and custom per-app rules
+// @version         0.1.5
+// @author          MrSpice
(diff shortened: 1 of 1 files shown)


m417z:
You can already achieve it with the Remove Taskbar Window Suffixes mod by using the Custom title modification rules. Since prefixes are less common, and since it's already possible, I'm not sure a dedicated mod is needed for this.

PR #4005, opened 2026-05-07
Add UWP Assets Redirect

### LINKED ISSUE #2972: [Resource Redirect] Replacing icons/resources of UWP programs
How can I replace the icon of a UWP app?

I tried redirecting the .exe file present in the WindowsApps folder, as it's the one appearing in Task Manager (even though I know it isn't the actual executable), but nothing happened.
The apps I tried without luck are Windows Terminal and Apple Music.

@m417z

atferrys:
# UWP Assets Redirect
Replace UWP app assets (such as icons) without worrying about updates
or modifying system files permissions.

You can configure redirections in three ways:
- Add folder redirections for WindowsApps, SystemApps, and Custom paths
- Use one or more theme files (more details on how to structure them in the mod README)
- Selecting an Icon theme from the available list directly from the Settings tab (more details in the mod README)

This mod is heavily inspired by Resource Redirect, they share similar goals, and some of the functionality
was taken and adapted from there (as noted in the code). They also share the same license.

This is my first Windhawk mod, so any suggestions on how to improve the code or functionality
are more than welcome.

Resolves https://github.com/ramensoftware/windhawk-mods/issues/2972.

Mod and Icon themes contributions are welcome [here](https://github.com/atferrys/uwp-assets-redirect).

## Example: Before and After
Before and after comparison of some applications

## Changelog

N/A (new mod submission)

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/uwp-assets-redirect.wh.cpp
+++ b/mods/uwp-assets-redirect.wh.cpp
@@ -64,6 +64,9 @@ A quicker and easier approach is to use something like [TileGen](https://tilegen
 generates the required assets from a single image or an `.ico` file.
 _Using an .ico file is recommended, as it already contains multiple resolutions._
 
+_Keep in mind that even if File Explorer doesn't show the non-redirected assets after you've applied them,
+they're still present. This is only a visual issue caused by how File Explorer loads folder contents._
+
 # Theme paths
 Theme paths can be set in the settings.
 Each theme path can be a folder with custom assets files and a `theme.ini` file that contains redirection rules, or the `.ini` theme file itself.
--- a/mods/uwp-assets-redirect.wh.cpp
+++ b/mods/uwp-assets-redirect.wh.cpp
@@ -1773,7 +1773,7 @@ void LoadSettings() {
 
 }
 
-void Wh_ModInit() {
+BOOL Wh_ModInit() {
 
     LoadSettings();
 
@@ -1788,6 +1788,8 @@ void Wh_ModInit() {
     RefreshIcons(true);
 
     Wh_Log(L"UWP Assets Redirect has been initialized.");
+
+    return TRUE;
 
 }
 
--- a/mods/uwp-assets-redirect.wh.cpp
+++ b/mods/uwp-assets-redirect.wh.cpp
@@ -1624,6 +1624,7 @@ void LoadRedirections(std::unordered_map<std::wstring, std::wstring>& redirectio
 
                 if (response->statusCode != 200) {
                     Wh_Log(L"Failed to download Icon theme: Request failed with code %d.", response->statusCode);
+                    Wh_FreeUrlContent(response);
                     return false;
                 }
 
--- a/mods/uwp-assets-redirect.wh.cpp
+++ b/mods/uwp-assets-redirect.wh.cpp
@@ -1671,12 +1671,60 @@ void LoadRedirections(std::unordered_map<std::wstring, std::wstring>& redirectio
             const auto storage_path = std::filesystem::path{storage_path_buffer} / "icon-themes";
             std::filesystem::create_directories(storage_path);
 
-            auto icon_theme_path = storage_path / replace(icon_theme, L'/', L'\\');
+            // Workaround for: https://github.com/ramensoftware/windhawk-mods/pull/4005#discussion_r3213756330
+            // Taken and adapted from: https://github.com/m417z/my-windhawk-mods/commit/cdeeca27376e98cecea2e2f8b637902c86af9699
+            const std::wstring icon_theme_path_key = std::format(L"icon_theme_path_{}", icon_theme);
+
+            WCHAR stored_icon_theme_path[MAX_PATH];
+            Wh_GetStringValue(
+                icon_theme_path_key.c_str(),
+                stored_icon_theme_path,
+                ARRAYSIZE(stored_icon_theme_path)
+            );
 
             std::error_code error_code;
+            std::filesystem::path icon_theme_path;
+
+            if(*stored_icon_theme_path) {
+                icon_theme_path = storage_path / stored_icon_theme_path;
+                if (std::filesystem::is_directory(icon_theme_path, error_code)) {
+                    Wh_Log(L"Theme files for %s already downloaded.", icon_theme.c_str());
+                    return icon_theme_path;
+                }
+            }
+
+            // Find a usable icon theme folder name.
+            const std::wstring icon_theme_folder = replace(icon_theme, L'/', L'\\');
+
+            for (int suffix = 0; suffix < 100; suffix++) {
+
+                if (suffix == 0) {
+                    icon_theme_path = storage_path / icon_theme_folder;
+                } else {
+                    icon_theme_path = storage_path / std::format(L"{}_{}", icon_theme_folder, suffix + 1);
+                }
+
+                if (!std::filesystem::is_directory(icon_theme_path, error_code)) {
+                    // Icon theme folder doesn't exist, we can use it.
+                    break;
+                }
+
+                // Icon theme folder exists, try to remove it.
+                Wh_Log(L"Icon theme folder already exists, trying to remove: %s", icon_theme_path.c_str());
+                std::filesystem::remove_all(icon_theme_path, error_code);
+
+                if (!std::filesystem::is_directory(icon_theme_path, error_code)) {
+                    // Successfully removed icon theme folder.
+                    break;
+                }
+
+                Wh_Log(L"Failed to remove Icon theme folder, trying next usable name...");
+
+            }
+
             if (std::filesystem::is_directory(icon_theme_path, error_code)) {
-                Wh_Log(L"Theme files for %s already downloaded.", icon_theme.c_str());
-                return icon_theme_path;
+                Wh_Log(L"Failed to download Icon theme: Unable to find a usable theme folder name.");
+                return L"";
             }
 
             const auto lock_file = storage_path / std::format(L"{}_lock", replace(icon_theme, L'/', L'_'));
@@ -1694,6 +1742,8 @@ void LoadRedirections(std::unordered_map<std::wstring, std::wstring>& redirectio
 
             if (download_theme(icon_theme, temp_zip, temp_extract, icon_theme_path)) {
                 Wh_Log(L"Theme files for %s downloaded and extracted successfully.", icon_theme.c_str());
+                // Store the Icon theme folder name.
+                Wh_SetStringValue(icon_theme_path_key.c_str(), icon_theme_path.c_str());
             } else {
                 Wh_Log(L"Failed to download and extract icon theme.");
                 icon_theme_path.clear();
--- a/mods/uwp-assets-redirect.wh.cpp
+++ b/mods/uwp-assets-redirect.wh.cpp
@@ -1,0 +1,1822 @@
+// ==WindhawkMod==
+// @id              uwp-assets-redirect
+// @name            UWP Assets Redirect
+// @description     Replace UWP app assets (such as icons) without worrying about updates or modifying system files permissions.
+// @version         1.0
+// @author          ferrys
+// @github          https://github.com/atferrys
+// @license         GPL-3.0
+// @include         explorer.exe
+// @include         StartMenuExperienceHost.exe
+// @include         SearchHost.exe
+// @include         ShellExperienceHost.exe
+// @include         ShellHost.exe
+// @include         RuntimeBroker.exe
+// @include         Taskmgr.exe
+// @exclude         windhawk.exe
+// @compilerOptions -lcomctl32 -lole32 -loleaut32
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# UWP Assets Redirect
+Replace UWP app assets (such as icons) without worrying about updates
+or modifying system files permissions.
+
+## Example: Before and After
+![Before and after comparison of some applications](https://raw.githubusercontent.com/atferrys/uwp-assets-redirect/main/docs-assets/example-before-after.png)
+
+# Icon themes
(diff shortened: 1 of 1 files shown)


mods/uwp-assets-redirect.wh.cpp
```diff
@@ -0,0 +1,1822 @@
...
+// @github          https://github.com/atferrys
+// @license         GPL-3.0
+// @include         explorer.exe
+// @include         StartMenuExperienceHost.exe
+// @include         SearchHost.exe
+// @include         ShellExperienceHost.exe
+// @include         ShellHost.exe
+// @include         RuntimeBroker.exe
+// @include         Taskmgr.exe
+// @exclude         windhawk.exe
```
m417z: This has no effect, nothing causes `windhawk.exe` to be included.

mods/uwp-assets-redirect.wh.cpp
```diff
@@ -0,0 +1,1822 @@
...
+
+        }
+
+        if(!redirectPath.empty()) {
+
+            Wh_Log(L"[Redirect Attempt] %s -> %s", originalPath.c_str(), redirectPath.c_str());
+
+            ObjectName->Buffer = (PWSTR) redirectPath.c_str();
+            ObjectName->Length = (USHORT) (redirectPath.length() * sizeof(WCHAR));
+            ObjectName->MaximumLength = ObjectName->Length;
```
m417z: The function is not supposed to modify `ObjectAttributes`. Instead of modifying it, create a local copy and use it.

mods/uwp-assets-redirect.wh.cpp
```diff
@@ -0,0 +1,1822 @@
...
+        }
+
+        Wh_Log(L"Loaded %i redirections from shared cache.", redirections.size());
+        g_redirections = std::move(redirections);
+
+    }
+
+}
+
+void Wh_ModInit() {
```
m417z: `Wh_ModInit` must return `BOOL`.

mods/uwp-assets-redirect.wh.cpp
```diff
@@ -0,0 +1,1822 @@
...
+
+}
+
+BOOL Wh_ModSettingsChanged(BOOL* bReload) {
+
+    Wh_Log(L"Reloading configuration...");
+
+    if(!DoesCurrentProcessOwnTaskbar()) {
+        *bReload = TRUE;
+        return TRUE;
```
m417z: Why reload it?

mods/uwp-assets-redirect.wh.cpp
```diff
@@ -0,0 +1,1822 @@
...
+                const WH_URL_CONTENT* response = Wh_GetUrlContent(theme_url.c_str(), &options);
+
+                if (!response) {
+                    Wh_Log(L"Failed to download Icon theme: Wh_GetUrlContent returned NULL.");
+                    return false;
+                }
+
+                if (response->statusCode != 200) {
+                    Wh_Log(L"Failed to download Icon theme: Request failed with code %d.", response->statusCode);
+                    return false;
```
m417z, line 1655: `Wh_FreeUrlContent` is missing here.

mods/uwp-assets-redirect.wh.cpp
```diff
@@ -0,0 +1,1822 @@
...
+            }
+
+            const auto storage_path = std::filesystem::path{storage_path_buffer} / "icon-themes";
+            std::filesystem::create_directories(storage_path);
+
+            auto icon_theme_path = storage_path / replace(icon_theme, L'/', L'\\');
+
+            std::error_code error_code;
+            if (std::filesystem::is_directory(icon_theme_path, error_code)) {
+                Wh_Log(L"Theme files for %s already downloaded.", icon_theme.c_str());
```
m417z: This approach caused some issues with Resource Redirect, and I changed the implementation to avoid reusing a folder unless the mod created it in the past. The relevant change:
https://github.com/m417z/my-windhawk-mods/commit/cdeeca27376e98cecea2e2f8b637902c86af9699

The issue with the previous approach:
* The user selects a theme, it's downloaded and extracted to `<themes>\<theme-id>`
* When the mod is removed, the storage folder is removed too, but some files might be in use, so it may stay with those files.
* The user installs the mod again, which sees that `<themes>\<theme-id>` exists and tries to use it, but it's nearly empty at this point.

m417z (commented):
Thanks for the submission, looks great!

mods/uwp-assets-redirect.wh.cpp
atferrys: Got it, I thought it could be included if the user added something like `*` in the include list

mods/uwp-assets-redirect.wh.cpp
atferrys: I tried changing it to this, but now some `Custom` redirections are failing and falling back to the original assets even if they exist, am I missing something? @m417z 

```cpp
NTSTATUS NTAPI NtCreateFile_Hook(
    PHANDLE FileHandle,
    ACCESS_MASK DesiredAccess,
    POBJECT_ATTRIBUTES ObjectAttributes,
    PIO_STATUS_BLOCK IoStatusBlock,
    PLARGE_INTEGER AllocationSize,
    ULONG FileAttributes,
    ULONG ShareAccess,
    ULONG CreateDisposition,
    ULONG CreateOptions,
    PVOID EaBuffer,
    ULONG EaLength
) {

    std::wstring originalPath;

    if(ObjectAttributes && ObjectAttributes->ObjectName) {
        UNICODE_STRING* ObjectName = ObjectAttributes->ObjectName;
        originalPath.assign(ObjectName->Buffer, ObjectName->Length / sizeof(WCHAR));
    }

    // Check if we should redirect
    if(!originalPath.empty()) {

        std::wstring redirectPath;

        const wchar_t* path = originalPath.c_str();
        size_t pathLength = originalPath.length();

        for(const auto& pair : g_redirections) {

            const wchar_t* pattern = pair.first.c_str();
            size_t patternLength = pair.first.length();

            // If there are no wildcards in pattern, simply
            // use the end of the pattern as the match_end_index
            size_t match_end_index = patternLength;

            if(match(pattern, patternLength, path, pathLength, match_end_index)) {
                redirectPath = pair.second + originalPath.substr(match_end_index);
                break;
            }

        }

        if(!redirectPath.empty()) {

            Wh_Log(L"[Redirect Attempt] %s -> %s", originalPath.c_str(), redirectPath.c_str());

            POBJECT_ATTRIBUTES RedirectedObjectAttributes = ObjectAttributes;
            UNICODE_STRING* ObjectName = RedirectedObjectAttributes->ObjectName;

            ObjectName->Buffer = (PWSTR) redirectPath.c_str();
            ObjectName->Length = (USHORT) (redirectPath.length() * sizeof(WCHAR));
            ObjectName->MaximumLength = ObjectName->Length;

            NTSTATUS result = NtCreateFile_Original(
                FileHandle,
                DesiredAccess,
                RedirectedObjectAttributes,
                IoStatusBlock,
                AllocationSize,
                FileAttributes,
                ShareAccess,
                CreateDisposition,
                CreateOptions,
                EaBuffer,
                EaLength
            );

            if(NT_SUCCESS(result)) {
                Wh_Log(L"[Redirect Success] Redirected to: %s", redirectPath.c_str());
                return result;
            }

            Wh_Log(L"[Redirect Fail] Failed with code 0x%08X. Rolling back to original: %s", result, originalPath.c_str());

        }

    }

    return NtCreateFile_Original(
        FileHandle,
        DesiredAccess,
        ObjectAttributes,
        IoStatusBlock,
        AllocationSize,
        FileAttributes,
        ShareAccess,
        CreateDisposition,
        CreateOptions,
        EaBuffer,
        EaLength
    );
}
```

mods/uwp-assets-redirect.wh.cpp
atferrys: Because otherwise, the processes that read from the shared cache will reload the previous cache instead of the new one. I can't know what process will be injected first, so I can't make every process clear the cache on reload.

What happens is:
- If the main process gets injected first, it instantly clears the cache and then takes its time to write the new data
- The other processes see that the cache hasn't been written to yet and wait a few seconds for that to happen (or directly load the updated cache)

Since I can't be sure that the other processes' config will be reloaded after the main one clears the cache,
I make them reload the mod so that by the time they get re-initialized, I'm sure that the main config has *at least* cleared the old cache

atferrys:
@m417z, one quick question before merging, I noticed that on different Windows versions, some apps use a different capitalization for their assets folders.
Do you think there will be a noticeable performance impact if I use `std::tolower` in my pattern matching function? 

Something like this:
```cpp
bool mismatch = false;
for (size_t k = 0; k < literal_len; ++k) {
         if (std::tolower(str[match_pos + k]) != std::tolower(pattern[rest_start + k])) {
              mismatch = true;
              break;
         }
}
```
and
```cpp
if (std::tolower(pattern[pattern_index]) != std::tolower(str[string_index])) return false;
```

My concern was that the matching would occur up to an average of 30 redirections (for the default Windows apps) if none of them match.

mods/uwp-assets-redirect.wh.cpp
m417z: This line copies the pointer, not the struct itself:
`POBJECT_ATTRIBUTES RedirectedObjectAttributes = ObjectAttributes;`
You need to copy `OBJECT_ATTRIBUTES` itself according to its `Length` field.

Here's an example which does something similar:
https://github.com/ramensoftware/windhawk-mods/blob/681c473a5d25049e49fc88918ef94e51757e8d1d/mods/dark-menus.wh.cpp#L420-L425

mods/uwp-assets-redirect.wh.cpp
m417z: > I make them reload the mod so that by the time they get re-initialized [...]

So the mod reloading is just to spend some time? How is it different from just sleeping?

> I'm sure that the main config has at least cleared the old cache

How can you be sure?

Generally, the caching seems to be quite fragile, and I'm not sure how much it benefits performance. The loading is done on process startup and when settings change, which are not as common. For example, `NtCreateFile` calls are much more common, so optimizing that function will probably have more impact.

If you'd like to keep the caching because you measured and saw a real impact, or just because you feel like it, I'd suggest to have a better synchronization mechanism. An idea for a simple improvement is having an id for each cached settings, which is updated after the settings are written to the cache. Non-main processes will wait for that id to become non-zero on init, and to change on each `Wh_ModSettingsChanged` call.

There are still possible races such as a non-main mod reading broken config while the main process is updating it, but with the suggested check it should be less likely.

m417z:
> Do you think there will be a noticeable performance impact if I use `std::tolower` in my pattern matching function?

I don't know how `std::tolower` is implemented, its Unicode support, etc. Since it's quite unlikely that you'll be dealing with non-ASCII folder names, you can use a simple inline function like:
```cpp
inline WCHAR ToLowerAscii(WCHAR c) {
    if (c >= L'A' && c <= L'Z') {
        return c + (L'a' - L'A');
    }
    return c;
}
```
With this implementation, it's unlikely that there will be a noticeable impact.

Other possible optimizations:
* Instead of char-by-char lowering, you can lower the patterns on load, and only lower the path once.
* You can determine for each pattern on load whether it contains actual pattern characters. If not, you can simply use `wcscmp` or similar, which will be faster. Not sure it's worth the hassle.

mods/uwp-assets-redirect.wh.cpp
atferrys: It still fails with the same two apps:
```cpp
alignas(POBJECT_ATTRIBUTES) BYTE buffer[256];
memcpy(buffer, ObjectAttributes, ObjectAttributes->Length);
auto* RedirectedObjectAttributes = reinterpret_cast<POBJECT_ATTRIBUTES>(buffer);

UNICODE_STRING* ObjectName = RedirectedObjectAttributes->ObjectName;
ObjectName->Buffer = (PWSTR) redirectPath.c_str();
ObjectName->Length = (USHORT) (redirectPath.length() * sizeof(WCHAR));
ObjectName->MaximumLength = ObjectName->Length;
```

I don't see anything strange in the logs, though

m417z: When you `memcpy` the `OBJECT_ATTRIBUTES`, you duplicate the struct, but the `ObjectName` field inside the copy still holds the same pointer value - pointing at the caller's original `UNICODE_STRING`. Try something like:

```cpp
alignas(OBJECT_ATTRIBUTES) BYTE buffer[256];
memcpy(buffer, ObjectAttributes, ObjectAttributes->Length);
auto* RedirectedObjectAttributes = reinterpret_cast<POBJECT_ATTRIBUTES>(buffer);

UNICODE_STRING ObjectName;
ObjectName.Buffer = (PWSTR) redirectPath.c_str();
ObjectName.Length = (USHORT) (redirectPath.length() * sizeof(WCHAR));
ObjectName.MaximumLength = ObjectName.Length;

redirectedObjectAttributes->ObjectName = &ObjectName;
```

atferrys: Thank you, that works!

mods/uwp-assets-redirect.wh.cpp
atferrys: I've implemented the suggested changes, and it seems to be working fine. Let me know if there's something else I can improve there

> The loading is done on process startup and when settings change, which are not as common.

I know it doesn’t happen that often, but considering that, depending on the number of redirections, it could end up reading up to 30 Appx manifests every time, I think it’s worth implementing at least some sort of caching mechanism. 

Even though by default only a couple of common processes are included, the user could inject all the WindowsApps executables, or even worse, include every process. In that case, if every process caused the reading of 30 files on its startup, I think it could become an issue very soon.

Maybe it's not that much of an issue and I'm just being "paranoid" on too many file reads :>

PR #4011, opened 2026-05-08
New mod "Files 2 Folders"

triatomic:
# Files 2 Folders

Inspired by [Files 2 Folder](https://www.dcmembers.com/skwire/download/files-2-folder/) by Jody Holmes (Skwire Empire).

When one or more items are selected in an Explorer window, adds a **Files 2 Folder...** entry to the right-click context menu. Choosing it opens a dialog with four options for moving the selection into a new subfolder:

1. Move the selection into a subfolder with a fixed name (e.g. `folder`).
2. Move each file into a subfolder named after the file (without extension): `Good.bat` → `.\Good\Good.bat`.
3. Move each file into a subfolder named after its extension (e.g. all `.txt` files into `.\txt\`).
4. Move the selection into a subfolder named with the current date/time, formatted using the date format from settings.

Files 2 Folder dialog

## Fast vs. slow mode

By default the mod uses `MoveFileExW` directly, which is essentially instant for same-volume moves. Enable **Slow mode (shell-integrated)** in settings to route moves through `IFileOperation` instead — that gives the standard Windows progress dialog, **Ctrl+Z undo**, UAC elevation prompts, and conflict-resolution dialogs, at the cost of significantly slower operation on large selections.

## Workaround for other file managers

A configurable global hotkey (default **Ctrl+Alt+F**) makes the mod usable from any file manager (Total Commander, Directory Opus, Files, XYplorer, Q-Dir, etc.):

1. Select files in the other program.
2. Press **Ctrl+C** (or **Ctrl+X**) to put the file paths on the clipboard.
3. Press the hotkey. The dialog appears, and the files get moved into the chosen subfolder under their parent directory.

Light/dark dialog theme follows the Windows app theme by default and can be forced via settings.

## Mod authorship

This mod was created by:

- - [x] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/files-2-folders.wh.cpp
+++ b/mods/files-2-folders.wh.cpp
@@ -2,7 +2,7 @@
 // @id              files-2-folders
 // @name            Files 2 Folders
 // @description     Move one or more selected files in Explorer into a subfolder (named, by extension, by name, or by date), with a workaround hotkey for other file managers
-// @version         1.0.1
+// @version         1.2
 // @author          tria
 // @github          https://github.com/triatomic
 // @include         explorer.exe
@@ -208,11 +208,15 @@ void Wh_ModSettingsChanged() {
 // ============================================================
 static const UINT F2F_MENU_CMD = 0xBF20;  // unlikely to clash with shell ids
 
-// State for the currently-tracked menu
-static HWND  g_currentMenuHwnd        = nullptr;
-static bool  g_currentMenuEligible    = false;
-static std::vector<std::wstring> g_currentSelection;
-static std::wstring g_currentFolder;
+// State for the currently-tracked menu. thread_local because TrackPopupMenuEx
+// and the matching PostMessageW(WM_COMMAND) always run on the same UI thread,
+// and the PostMessageW hook itself fires on every thread in the process —
+// without thread_local, an unrelated thread's PostMessageW call could race
+// with the Explorer UI thread mid-write.
+static thread_local HWND  g_currentMenuHwnd        = nullptr;
+static thread_local bool  g_currentMenuEligible    = false;
+static thread_local std::vector<std::wstring> g_currentSelection;
+static thread_local std::wstring g_currentFolder;
 
 // ============================================================
 //  Path / name helpers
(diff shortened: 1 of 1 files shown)


mods/files-2-folders.wh.cpp
```diff
@@ -0,0 +1,1371 @@
...
+        if (fgTid && fgTid != myTid) {
+            attached = AttachThreadInput(fgTid, myTid, TRUE) != 0;
+        }
+        SetWindowPos(hDlg, HWND_TOP, 0, 0, 0, 0,
+                     SWP_NOMOVE | SWP_NOSIZE | SWP_SHOWWINDOW);
+        BringWindowToTop(hDlg);
+        SetForegroundWindow(hDlg);
+        SetActiveWindow(hDlg);
+        SetFocus(hDlg);
+        if (attached) AttachThreadInput(fgTid, myTid, FALSE);
```
m417z: This is quite aggressive and [may cause problems](https://devblogs.microsoft.com/oldnewthing/20080801-00/?p=21393). Why is it needed?

mods/files-2-folders.wh.cpp
```diff
@@ -0,0 +1,1371 @@
...
+// ============================================================
+//  Custom command id we inject into the shell context menu
+// ============================================================
+static const UINT F2F_MENU_CMD = 0xBF20;  // unlikely to clash with shell ids
+
+// State for the currently-tracked menu
+static HWND  g_currentMenuHwnd        = nullptr;
+static bool  g_currentMenuEligible    = false;
+static std::vector<std::wstring> g_currentSelection;
+static std::wstring g_currentFolder;
```
m417z: These global variables may be accessed from multiple threads. The easiest fix is to define them as `thread_local`.

mods/files-2-folders.wh.cpp
triatomic: Dropped.

mods/files-2-folders.wh.cpp
triatomic: Defined as thread_local

PR #4017, opened 2026-05-09
Create file-explorer-details-autofit-columns.wh.cpp

armaninyow:
## Mod authorship

This mod was created by:

- - [x] Claude
- - [ ] ChatGPT
- - [ ] Gemini


armaninyow:
# File Explorer Details Auto-Fit Columns

Automatically fits all visible column widths to their content whenever a folder is opened or refreshed in `Details view`, eliminating truncated text and ellipsis (`...`).

File Explorer Details Auto-Fit Columns

## Triggers

- Opening or navigating to a folder
- Minimizing and restoring the Explorer window
- `Ctrl + R`
- `F5`
- Right-click context menu → Refresh
- The refresh button in the toolbar

Only affects `Details view`. Other view modes (Icons, Tiles, List, etc.) are untouched.

## Settings

`Refresh Delay (ms)`: How long the mod waits before auto-fitting columns. The default is 400ms, which gives Explorer enough time to finish loading files before measuring content width. If you notice columns fitting too early in large folders, increase this value. If you want a snappier response in small folders, decrease it.

mods/file-explorer-details-autofit-columns.wh.cpp
```diff
@@ -0,0 +1,337 @@
...
+        // Update the stored IShellView for this tab
+        if (hwndTab) {
+            auto it = g_tabShellViews.find(hwndTab);
+            if (it != g_tabShellViews.end() && it->second)
+                it->second->Release();
+            pShellView->AddRef();
+            g_tabShellViews[hwndTab] = pShellView;
+        }
+
+        // Subclass top-level and all children, tagged with their tab
```
m417z: Is this really necessary? It'd be much better to only subclass the needed window.

mods/file-explorer-details-autofit-columns.wh.cpp
```diff
@@ -0,0 +1,337 @@
...
+// Updated every time UIActivate fires for that tab
+// ---------------------------------------------------------------------------
+
+static std::unordered_map<HWND, IShellView*> g_tabShellViews;
+
+// ---------------------------------------------------------------------------
+// Map: all subclassed windows
+// ---------------------------------------------------------------------------
+
+static std::unordered_map<HWND, HWND> g_windowToTab; // any hwnd -> its ShellTabWindowClass
```
m417z: The global variables may be accessed from different threads. Consider adding synchronization for thread safety.

armaninyow: > The global variables may be accessed from different threads. Consider adding synchronization for thread safety.

In practice, UIActivate and the subclass proc both run on Explorer's main UI thread so concurrent access is unlikely, but a CRITICAL_SECTION has been added for correctness.

mods/file-explorer-details-autofit-columns.wh.cpp
armaninyow: > Is this really necessary? It'd be much better to only subclass the needed window.

This was the result of iterative debugging to find which windows receive refresh commands. Now that those are identified (ShellTabWindowClass, ReBarWindow32, ToolbarWindow32), subclassing has been narrowed down to only those specific window classes.

mods/file-explorer-details-autofit-columns.wh.cpp
m417z: Each Explorer window runs in a separate thread.

armaninyow: > Each Explorer window runs in a separate thread.

My bad. Each Explorer window runs in its own thread, so concurrent access to the shared maps is a real possibility, not just a theoretical one. The CRITICAL_SECTION is therefore a necessary guard, not just a precautionary one.

mods/file-explorer-details-autofit-columns.wh.cpp
```diff
@@ -321,6 +367,8 @@ void Wh_ModAfterInit() {}
 
 void Wh_ModUninit() {
     Wh_Log(L"Uninit");
+
+    EnterCriticalSection(&g_cs);
     for (auto& [hwnd, _] : g_windowToTab) {
         KillTimer(hwnd, AUTOFIT_TIMER_ID);
         WindhawkUtils::RemoveWindowSubclassFromAnyThread(hwnd, ExplorerSubclassProc);
```
m417z: Calling `RemoveWindowSubclassFromAnyThread` while holding the lock can cause a deadlock. Copy/move the window handles to a local variable and then call `RemoveWindowSubclassFromAnyThread` while the lock is not held.

armaninyow: > Calling `RemoveWindowSubclassFromAnyThread` while holding the lock can cause a deadlock. Copy/move the window handles to a local variable and then call `RemoveWindowSubclassFromAnyThread` while the lock is not held.

Done!

PR #4026, opened 2026-05-09
Windows 11 Notification Center Styler v1.5

m417z:
* Added themes: LiquidGlass, BetterControl11, LayerMicaUI, Borderless.
* Updated themes: WindowGlass, Fluid.
* Added style variables that allow to observe a control's property value and use it in other controls or styles. Refer to the mod description for details.
* Added a way to match any number of intermediate parent controls, for example: `ParentClass > * > Class#Name`.
* Added a way to match root controls, for example: `:root > Class#Name`.
* Added the following properties to `WindhawkBlur`: `TintLuminosityOpacity`, `TintSaturation`, `NoiseOpacity`, `NoiseDensity`, and `FallbackColor`. The implementation was contributed by [Jhen](https://github.com/Lockframe). For details, refer to [the styling guide](https://github.com/ramensoftware/windows-11-notification-center-styling-guide#windhawkblur-effect-as-color).
* Improved `ThemeResource` handling for `WindhawkBlur`.
* Improved property matching which now works for more properties.
* Added XAML syntax support for resource variables.
* Improved resource variable handling and theme resource refresh logic.

### DIFF
--- a/mods/windows-11-notification-center-styler.wh.cpp
+++ b/mods/windows-11-notification-center-styler.wh.cpp
@@ -2,7 +2,7 @@
 // @id              windows-11-notification-center-styler
 // @name            Windows 11 Notification Center Styler
 // @description     Customize the Notification Center and Action Center with themes contributed by others or create your own
-// @version         1.4
+// @version         1.5
 // @author          m417z
 // @github          https://github.com/m417z
 // @twitter         https://twitter.com/m417z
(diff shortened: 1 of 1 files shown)


kavibhuva:
The notification styler is amazing however I noticed that the calendar when scrolled down to the next month it will still read as the current month and not change to the month you are viewing. would be great to see that happen.

Otherwise like how the volume bar is styled and the player.

m417z:
Which theme are you using? Please provide more details in a new issue here:
https://github.com/ramensoftware/windows-11-notification-center-styling-guide/issues

PR #4046, opened 2026-05-11
Update vlc-discord-rpc to v1.1.5

ciizerr:
### New
- Music activity on Discord now shows:  
  - **Listening to [Song title]**  
  - **Listening to [Artist]**  
  - **Listening to [Album]**  
- VLC port is detected in most setups, so you don’t need to enter the port number manually.  
- The mod now reads your VLC Lua HTTP password directly, so you can use any password without extra setup.  

### Fixes
- Toast notifications no longer repeat or display the wrong track.  
- The image scraper now provides more accurate artwork instead of random results.  
- Port detection works correctly across different VLC configurations, including non‑default ports.  

### Improvements
- Added `wh_log()` entries to make debugging and tracking media simpler.  
- Artwork handling has been refined for sharper and more consistent images.  

Special thanks to **@josephct** for suggesting the music activity feature and for identifying the VLC port issue that could stop Rich Presence from working.

### DIFF
--- a/mods/vlc-discord-rpc.wh.cpp
+++ b/mods/vlc-discord-rpc.wh.cpp
@@ -2,7 +2,7 @@
 // @id              vlc-discord-rpc
 // @name            VLC Discord Rich Presence
 // @description     Shows your playing status, quality tags (4K/HDR), and interactive buttons on Discord.
-// @version         1.1.2
+// @version         1.1.3
 // @author          ciizerr
 // @github          https://github.com/ciizerr
 // @homepage        https://vlc-rpc.vercel.app/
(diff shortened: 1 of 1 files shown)


mods/vlc-discord-rpc.wh.cpp
```diff
@@ -1,135 +1,180 @@
...
+| **Notifications** | Shows a Windows toast when a new track starts |
+| **Search Button** | Picks which site the status button links to |
+| **"Listening to..." label** | Controls what shows in the activity name for music |
 
-## Configuration
-**Show Cover Art:** Toggle to enable/disable the fetching and uploading of cover art. If disabled, the mod will use the standard VLC icon.
+## Filter Updates
 
-**Show Chapter / Audio Language:** You can independently hide the current chapter number or audio language if you prefer a cleaner layout that only shows Season and Episode.
+To keep title cleaning accurate, the mod can download a small community-maintained filter list from GitHub (refreshed every 6 hours). If you'd rather make zero external requests, enable **Local Filters Only** in settings — this falls back to the built-in word list plus anything you've added yourself.
```
m417z: "zero external requests" is not accurate. The mod still:

-   Scrapes `www.bing.com` for poster URLs (`FindExternalArtwork`)
-   Uploads local album art to `uguu.se` (`UploadToUguu`)
-   Downloads `vlc-discord-icon.ico` from raw.githubusercontent.com when notifications fire

Clarify this section. I'll also be useful to have an explicit list of external resources the mod sends requests to by default.

ciizerr: Thanks for Pointing out — my earlier explanation wasn’t quite right.  
I’ve updated the section so it’s clear that **Local Filters Only** just disables the online filter list download.  
With that option on, the mod falls back to the built‑in word list plus any custom filters you’ve added yourself.  

I also added a new section listing the external requests the mod makes by default:

- `www.bing.com`  
- `uguu.se`  
- `raw.githubusercontent.com`

PR #4058, opened 2026-05-13
Create taskbar-ctrl-rightclick-open-file-location.wh.cpp

triatomic:
## Summary

New mod: hold a configurable modifier (default **Ctrl**) while clicking a taskbar button to:

- Open the running program's file location in Explorer (default)
- Copy the executable path to the clipboard
- Launch a new elevated instance (UAC prompt)

Works on the **classic taskbar** (`MSTaskListWClass`):

- Stock Windows 10
- Windows 11 with **StartAllBack**
- Windows 11 with **ExplorerPatcher** (legacy taskbar mode)

The mod subclasses `MSTaskListWClass` so the right-click never reaches the taskbar's own context-menu handler (this is what makes it work on StartAllBack, which doesn't go through `taskbar.dll`'s `_ComputeJumpViewPosition`). Button-to-process resolution uses MSAA `AccessibleObjectFromPoint` and, when the accessible value doesn't expose an HWND directly (the StartAllBack case), a fallback that matches against the exe's `FileDescription` version-info field with alphanumeric normalization (handles e.g. `ConEmu (x64)` ↔ `ConEmu64.exe`, `MSI Afterburner` ↔ `MSIAfterburner.exe`).

## Settings

- **Modifier key** — Ctrl / Shift / Alt / Win / None
- **Mouse button** — Right (suppresses jump list) / Middle
- **Action** — Open location / Copy path / Run as admin
- **Hold time (ms)** — 0 for instant, or require the click to be held for N ms before firing

## Mod authorship

This mod was created by:

- - [x] Claude
- - [ ] ChatGPT
- - [ ] Gemini

## Test plan

- [x] Install on Windows 11 with StartAllBack: Ctrl+Right-click a taskbar button opens its `.exe` location in Explorer without showing the SAB jump list
- [x] Grouped buttons (multiple windows) resolve to the right `.exe`
- [x] Ungrouped / single-window buttons resolve correctly
- [x] Switching settings (modifier, mouse button, action, hold time) takes effect without restarting explorer
- [x] Disabling/unloading the mod cleanly removes the subclass — no taskbar crash

### DIFF
--- a/mods/taskbar-ctrl-rightclick-open-file-location.wh.cpp
+++ b/mods/taskbar-ctrl-rightclick-open-file-location.wh.cpp
@@ -1,0 +1,719 @@
+// ==WindhawkMod==
+// @id              taskbar-ctrl-rightclick-open-file-location
+// @name            Ctrl+Right-click taskbar to Open file location
+// @description     Holding Ctrl while right-clicking a taskbar button opens the running program's file location in Explorer (instead of showing the jump list). Works with the classic / StartAllBack / ExplorerPatcher taskbar (MSTaskListWClass).
+// @version         1.2.0
+// @author          triatomic
+// @github          https://github.com/triatomic
+// @include         explorer.exe
+// @architecture    x86-64
+// @compilerOptions -lpsapi -lshell32 -loleacc -luser32 -lole32 -loleaut32 -lversion -lcomctl32
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Ctrl+Right-click taskbar to Open file location
+
+Hold **Ctrl** and **right-click** a running app's taskbar button to open
+its executable in File Explorer (the file is selected — same result as the
+jump list's "Open file location" option). Without Ctrl, right-click behaves
+normally and the jump list / context menu appears.
+
+This version targets the **classic taskbar** (`MSTaskListWClass`) used by:
+
+- Stock Windows 10
+- Windows 11 with **StartAllBack**
+- Windows 11 with **ExplorerPatcher** (legacy taskbar mode)
+
+It subclasses `MSTaskListWClass` directly (so the right-click never
+reaches the taskbar's own context-menu handler) and uses Microsoft Active
+Accessibility (MSAA) to identify the button under the cursor — no
+dependency on private taskbar.dll symbols, resilient to Windows updates.
+
+The mod runs entirely inside `explorer.exe`, so the hook is removed
+automatically when Windhawk unloads it.
+*/
+// ==/WindhawkModReadme==
+
+// ==WindhawkModSettings==
+/*
+- modifier: ctrl
+  $name: Modifier key
+  $description: Which key must be held while clicking the taskbar button.
+  $options:
+  - ctrl: Ctrl
+  - shift: Shift
+  - alt: Alt
+  - win: Win
+  - none: None (just the click)
+- mouseButton: right
+  $name: Mouse button
+  $description: Which mouse button triggers the action.
+  $options:
+  - right: Right-click (suppresses the jump list / context menu)
+  - middle: Middle-click
+- action: openLocation
+  $name: Action
+  $description: What to do when the trigger fires.
+  $options:
+  - openLocation: Open file location in Explorer
+  - copyPath: Copy executable path to clipboard
+  - runAsAdmin: Launch a new elevated instance (UAC prompt)
+- holdMs: 0
+  $name: Hold time (ms)
+  $description: >-
+    If non-zero, the trigger must be held down for at least this many
+    milliseconds before firing. A shorter press does nothing. 0 = fire
+    instantly on press.
+*/
+// ==/WindhawkModSettings==
+
+#include <windhawk_utils.h>
+
+#include <psapi.h>
+#include <shlobj.h>
+#include <oleacc.h>
+#include <windowsx.h>
+
+#include <string>
+#include <vector>
+#include <atomic>
+
+// ---------------------------------------------------------------------------
+// Settings
+// ---------------------------------------------------------------------------
+
+enum class Modifier { None, Ctrl, Shift, Alt, Win };
+enum class MouseBtn { Right, Middle };
+enum class Action { OpenLocation, CopyPath, RunAsAdmin };
+
+struct Settings {
+    Modifier modifier = Modifier::Ctrl;
+    MouseBtn mouseButton = MouseBtn::Right;
+    Action action = Action::OpenLocation;
+    int holdMs = 0;
+};
+
+static Settings g_settings;
+
+static int VkForModifier(Modifier m) {
+    switch (m) {
+        case Modifier::Ctrl:  return VK_CONTROL;
+        case Modifier::Shift: return VK_SHIFT;
+        case Modifier::Alt:   return VK_MENU;
+        case Modifier::Win:   return VK_LWIN;  // also accept RWIN below
+        default:              return 0;
+    }
+}
+
+static bool IsModifierHeld(Modifier m) {
+    if (m == Modifier::None) return true;
+    int vk = VkForModifier(m);
+    if (vk == VK_LWIN) {
+        return (GetKeyState(VK_LWIN) & 0x8000) ||
+               (GetKeyState(VK_RWIN) & 0x8000);
+    }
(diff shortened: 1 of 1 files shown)


m417z:
Any thoughts about supporting the Windows 11 taskbar? I believe that's what most Windhawk users use.
Several mods demonstrate how to subclass it and receive events, for example "Taskbar Volume Control" or "Click on empty taskbar space".

https://github.com/ramensoftware/windhawk-mods/blob/22ec4510d1db5a2a0734427f8aff7592aa254bda/mods/taskbar-volume-control.wh.cpp#L1541-L1557

triatomic:
Updated to support native windows 11.

mods/taskbar-ctrl-rightclick-open-file-location.wh.cpp
```diff
@@ -0,0 +1,1126 @@
...
+}
+
+void Wh_ModSettingsChanged() {
+    Wh_Log(L">");
+    LoadSettings();
+}
+
+void Wh_ModAfterInit() {
+    if (g_settings.taskbarType != TaskbarType::Win11) {
+        AttachTaskbarSubclasses();
```
m417z: This won't find the taskbar if the mod is loaded on explorer.exe start. Have you tested the mod after a reboot or explorer.exe restart, not just when it's enabled while Explorer is already running?

mods/taskbar-ctrl-rightclick-open-file-location.wh.cpp
```diff
@@ -0,0 +1,1126 @@
...
+        if (IsWindow(hWnd)) {
+            if (pendingId) KillTimer(hWnd, pendingId);
+            WindhawkUtils::RemoveWindowSubclassFromAnyThread(
+                hWnd, TaskListSubclassProc);
+        }
+    }
+    g_subclassedWindows.clear();
+
+    if (g_pUIA) {
+        g_pUIA->Release();
```
m417z: It must be released in the thread that created it. You can post a custom worker-thread message (e.g. `WM_APP_RELEASE_UIA`) that releases `g_pUIA` and then post `WM_QUIT` - do all teardown on the worker thread before joining.

mods/taskbar-ctrl-rightclick-open-file-location.wh.cpp
```diff
@@ -0,0 +1,1126 @@
...
+    }
+    if (g_settings.taskbarType != TaskbarType::Classic) {
+        InstallCreateWindowHooks();
+    }
+    return TRUE;
+}
+
+void Wh_ModSettingsChanged() {
+    Wh_Log(L">");
+    LoadSettings();
```
m417z, line 1160: `Wh_ModSettingsChanged` does not re-apply hooks when `taskbarType` changes.

mods/taskbar-ctrl-rightclick-open-file-location.wh.cpp
```diff
@@ -0,0 +1,1126 @@
+// ==WindhawkMod==
+// @id              taskbar-ctrl-rightclick-open-file-location
+// @name            Ctrl+Right-click taskbar to Open file location
+// @description     Holding Ctrl while right-clicking a taskbar button opens the running program's file location in Explorer (instead of showing the jump list). Works with the classic / StartAllBack / ExplorerPatcher taskbar and the native Windows 11 taskbar.
+// @version         1.3.0
+// @author          tria
+// @github          https://github.com/triatomic
+// @include         explorer.exe
+// @architecture    x86-64
+// @compilerOptions -DWINVER=0x0A00 -lpsapi -lshell32 -loleacc -luser32 -lole32 -loleaut32 -lversion -lcomctl32
```
m417z:
```suggestion
// @compilerOptions -lpsapi -lshell32 -loleacc -luser32 -lole32 -loleaut32 -lversion -lcomctl32
```

mods/taskbar-ctrl-rightclick-open-file-location.wh.cpp
```diff
@@ -0,0 +1,1126 @@
...
+            outHwnd = FindHwndInAccChildren(pAcc, 0);
+        }
+    }
+
+    VariantClear(&vChild);
+    pAcc->Release();
+
+    // Native Win11 taskbar fallback. If MSAA gave us the XAML host class as
+    // the name (or nothing usable), ask UI Automation — it descends into
+    // XAML and returns the button's actual Name (e.g. "Adobe Photoshop").
```
m417z: MSAA is legacy technology. Why not always use UI Automation? Also, Windows 11 also has the HWND accessible, which is more reliable and more efficient than title mapping, use it. Not sure about StartAllBack, I'm not using it, but you can check with Microsoft Inspect tool.

mods/taskbar-ctrl-rightclick-open-file-location.wh.cpp
```diff
@@ -0,0 +1,1126 @@
...
+            g_subclassedWindows.push_back(hWnd);
+            Wh_Log(L"Subclassed MSTaskListWClass %p", hWnd);
+        }
+    }
+    return TRUE;
+}
+
+static BOOL CALLBACK EnumTrayWndProc(HWND hWnd, LPARAM lParam) {
+    WCHAR cls[64];
+    if (GetClassNameW(hWnd, cls, ARRAYSIZE(cls)) > 0 &&
```
m417z, line 848: Limit the enumeration to windows of the current process.

mods/taskbar-ctrl-rightclick-open-file-location.wh.cpp
```diff
@@ -0,0 +1,1126 @@
...
+            if (g_hookedInputSiteProc) return;
+        }
+        bridge = FindWindowExW(
+            hTaskbarRoot, bridge,
+            L"Windows.UI.Composition.DesktopWindowContentBridge", nullptr);
+    }
+}
+
+static BOOL CALLBACK EnumWin11TrayProc(HWND hWnd, LPARAM /*lParam*/) {
+    if (IsTaskbarRoot(hWnd)) TryAttachWin11ForTaskbar(hWnd);
```
m417z: Limit the enumeration to windows of the current process.

mods/taskbar-ctrl-rightclick-open-file-location.wh.cpp
triatomic: split this by having old taskbar use MSAA while modern taskbar use UIA

mods/taskbar-ctrl-rightclick-open-file-location.wh.cpp
```diff
@@ -510,6 +516,8 @@ static std::wstring GetTaskbarButtonNameViaUIA(POINT pt) {
         cur->get_CurrentClassName(&bClass);
         BSTR bName = nullptr;
         cur->get_CurrentName(&bName);
+        UIA_HWND uHwnd = nullptr;
+        cur->get_CurrentNativeWindowHandle(&uHwnd);
```
m417z, line 520: Does this work? If I understand correctly `get_CurrentNativeWindowHandle` returns the window handle of the button (explorer.exe). But there's also info about the target window:
```
AutomationId:	"Window: 0x10750"
```

You can use [inspect.exe](https://www.dropbox.com/s/5cfa754k000zt07/inspect.exe?dl=1) to see it.

mods/taskbar-ctrl-rightclick-open-file-location.wh.cpp
```diff
@@ -1076,45 +1148,49 @@ BOOL Wh_ModInit() {
...
+                Sleep(500);
+                if (!FindWindowW(L"Shell_TrayWnd", nullptr)) continue;
+                ApplyTaskbarAttachments();
+                if (!g_subclassedWindows.empty() || g_hookedInputSiteProc) {
+                    Wh_Log(L"Late taskbar attach succeeded after %d tries", i + 1);
+                    return;
+                }
+            }
+            Wh_Log(L"Late taskbar attach gave up");
+        }).detach();
```
m417z, line 1182: This will crash if the mod is unloaded while the thread is running. Generally, you shouldn't need a thread if you're hooking `CreateWindowExW`. You can look at how existing mods do it.

It's best to just hook the creation functions regardless of the mode in the settings. Then, when settings are changed, you don't need to set different hooks, which I mentioned in another comment.

Also, this won't work for newly added taskbars (e.g. attached monitor). Window creation hooks will.

PR #4060, opened 2026-05-14
Add Simple Window Switcher Mod

Louis047:
This Mod is a port of Simple Window Switcher from ExplorerPatcher, built with the help of source code [here](https://github.com/valinet/sws)

Incorporates the following features
- **Dynamic Grid Layout:** Flexible rows that adapt to window aspect ratios.
- **Live Thumbnails:** High-performance DWM previews of all open windows.
- **Full Navigation:** Support for arrow keys, Tab, mouse clicks, and scrolling.
- **Sticky Mode:** Stays open with Alt+Ctrl+Tab for easy multi-tasking.
- **System Sync:** Auto-detects Dark/Light mode and handles multi-monitor DPI.
- **Modern Aesthetics:** Supports Acrylic backgrounds and Windows accent colors.

## Mod authorship

This mod was created by:

- - [x] The submitter, with AI assistance (and co-author @Asteski)
- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini
- - [x] Another AI (please specify): Antigravity (Opus 4.6)

### DIFF
--- a/mods/simple-window-switcher.wh.cpp
+++ b/mods/simple-window-switcher.wh.cpp
@@ -64,14 +64,18 @@ Additional improvements made by [Asteski](https://github.com/Asteski).
   $options:
   - horizontal: Horizontal
   - vertical: Vertical
-- headerContentMode: horizontal
-  $name: Header Content Mode
-  $description: Layout of the task header icon and title.
+- headerContentOrientation: horizontal
+  $name: Header Content Orientation
+  $description: Orientation of the task header icon and title.
   $options:
   - horizontal: Horizontal
   - vertical: Vertical
-  - verticalLarge: Vertical with large icons
-  - horizontalLarge: Horizontal with large icons
+- iconSize: small
+  $name: Icon Size
+  $description: Size of the header icon.
+  $options:
+  - small: Small
+  - large: Large
 - rowHeight: 230
   $name: Row Height
   $description: Total height of each thumbnail row in pixels (before DPI scaling). Default 230 matches ExplorerPatcher.
@@ -173,7 +177,7 @@ struct WindowEntry {
     SIZE effectiveSourceSize;  // Source size after cropping invisible frame
 };
 struct Settings {
-    WCHAR theme[32]; WCHAR colorScheme[32]; WCHAR cornerPreference[32]; WCHAR scrollWheelBehavior[32]; WCHAR taskListOrientation[32]; WCHAR headerContentMode[32];
+    WCHAR theme[32]; WCHAR colorScheme[32]; WCHAR cornerPreference[32]; WCHAR scrollWheelBehavior[32]; WCHAR taskListOrientation[32]; WCHAR headerContentOrientation[32]; WCHAR iconSize[32];
     WCHAR borderColor[16];
     int opacity; int rowHeight; int rowWidth;
     int maxWidthPercent; int maxHeightPercent; int windowPadding; int showDelay;
@@ -214,15 +218,13 @@ static const PROPERTYKEY kPkeyAppUserModelId = {
 static bool ThemeIs(const WCHAR* v) { return wcscmp(g_settings.theme, v) == 0; }
 static bool ScrollIs(const WCHAR* v) { return wcscmp(g_settings.scrollWheelBehavior, v) == 0; }
 static bool LayoutIsVertical() { return wcscmp(g_settings.taskListOrientation, L"vertical") == 0; }
-static bool HeaderModeIs(const WCHAR* v) { return wcscmp(g_settings.headerContentMode, v) == 0; }
+static bool HeaderOrientationIs(const WCHAR* v) { return wcscmp(g_settings.headerContentOrientation, v) == 0; }
+static bool IconSizeIs(const WCHAR* v) { return wcscmp(g_settings.iconSize, v) == 0; }
 static bool HeaderIsVertical() {
-    return HeaderModeIs(L"vertical") || HeaderModeIs(L"verticalLarge");
-}
-static bool HeaderIsHorizontalLarge() {
-    return HeaderModeIs(L"horizontalLarge");
+    return HeaderOrientationIs(L"vertical");
 }
 static int GetHeaderIconSizePx() {
-    if (HeaderModeIs(L"verticalLarge") || HeaderModeIs(L"horizontalLarge")) {
+    if (IconSizeIs(L"large")) {
         return MulDiv(32, g_dpiX, 96);
     }
     return MulDiv(SWS_ICON_SIZE, g_dpiX, 96);
@@ -463,7 +465,7 @@ static bool IsAltTabWindow(HWND h) {
 
 // Window Enumeration
 
-static HICON TryGetAppIconFromAumid(HWND hWnd, bool largeIcon);
+static HICON TryGetAppIconFromAumid(HWND hWnd, int desiredSizePx);
 
 static BOOL CALLBACK EnumWindowsProc(HWND hWnd, LPARAM lParam) {
     auto* list = reinterpret_cast<std::vector<WindowEntry>*>(lParam);
@@ -483,7 +485,7 @@ static BOOL CALLBACK EnumWindowsProc(HWND hWnd, LPARAM lParam) {
     SendMessageTimeoutW(hWnd, WM_GETICON, ICON_BIG, 0, SMTO_ABORTIFHUNG | SMTO_BLOCK, 100, (DWORD_PTR*)&e.hIcon);
     if (!e.hIcon) SendMessageTimeoutW(hWnd, WM_GETICON, ICON_SMALL2, 0, SMTO_ABORTIFHUNG | SMTO_BLOCK, 100, (DWORD_PTR*)&e.hIcon);
     if (!e.hIcon) SendMessageTimeoutW(hWnd, WM_GETICON, ICON_SMALL, 0, SMTO_ABORTIFHUNG | SMTO_BLOCK, 100, (DWORD_PTR*)&e.hIcon);
-    if (!e.hIcon) e.hIcon = TryGetAppIconFromAumid(hWnd, HeaderModeIs(L"verticalLarge") || HeaderModeIs(L"horizontalLarge"));
+    if (!e.hIcon) e.hIcon = TryGetAppIconFromAumid(hWnd, GetHeaderIconSizePx());
     if (!e.hIcon) e.hIcon = (HICON)GetClassLongPtrW(hWnd, GCLP_HICON);
     if (!e.hIcon) e.hIcon = (HICON)GetClassLongPtrW(hWnd, GCLP_HICONSM);
     if (!e.hIcon) e.hIcon = LoadIconW(NULL, IDI_APPLICATION);
@@ -491,7 +493,7 @@ static BOOL CALLBACK EnumWindowsProc(HWND hWnd, LPARAM lParam) {
     return TRUE;
 }
 
-static HICON TryGetAppIconFromAumid(HWND hWnd, bool largeIcon) {
+static HICON TryGetAppIconFromAumid(HWND hWnd, int desiredSizePx) {
     if (!g_SHGetPropertyStoreForWindow) {
         return NULL;
     }
@@ -517,11 +519,14 @@ static HICON TryGetAppIconFromAumid(HWND hWnd, bool largeIcon) {
                 PIDLIST_ABSOLUTE pidl = NULL;
                 if (SUCCEEDED(SHParseDisplayName(appsFolderPath, NULL, &pidl, 0, NULL)) && pidl) {
                     SHFILEINFOW sfi = {};
-                    if (SHGetFileInfoW((LPCWSTR)pidl,
-                                       0,
-                                       &sfi,
-                                       sizeof(sfi),
-                                       SHGFI_PIDL | SHGFI_ICON | (largeIcon ? SHGFI_LARGEICON : SHGFI_SMALLICON))) {
+                    // Request larger icon for medium and above (>24px)
+                    UINT flags = SHGFI_PIDL | SHGFI_ICON;
+                    if (desiredSizePx > 24) {
+                        flags |= SHGFI_LARGEICON;  // 32x32
+                    } else {
+                        flags |= SHGFI_SMALLICON;  // 16x16
+                    }
+                    if (SHGetFileInfoW((LPCWSTR)pidl, 0, &sfi, sizeof(sfi), flags)) {
                         hIcon = sfi.hIcon;
                     }
                     CoTaskMemFree(pidl);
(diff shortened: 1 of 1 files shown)


Louis047:
Took me a lot of testing for the whole week. There's a chance where I might have missed very niche edge cases but the current state of the mod is stable for everyone to use.

Louis047:
The sole reason I created this mod is for the users who don't want to use another tool just for using this feature, so thought of centralizing this and windhawk was the best choice for it.

Louis047:
Hey @m417z , an update on the mod:

Me and @Asteski are working on adding a few improvements to the mod. So I request you to not merge this yet. Once things are done, I'll let you know.

Louis047:
Hey @m417z, we are finally done working on this and is ready to merge. Let me know if any changes needed for this mod 👍

mods/simple-window-switcher.wh.cpp
```diff
@@ -0,0 +1,2024 @@
...
+    if (hShell32) {
+        g_SHGetPropertyStoreForWindow =
+            (SHGetPropertyStoreForWindow_t)GetProcAddress(hShell32, "SHGetPropertyStoreForWindow");
+    }
+
+    return true;
+}
+
+// Registry flag to prevent restart prompt loop on init
+#define SWS_REG_PATH L"Software\\Windhawk\\SimpleWindowSwitcher"
```
m417z: Don't modify the registry, use APIs such as `Wh_SetIntValue` and `Wh_SetStringValue` instead.

mods/simple-window-switcher.wh.cpp
```diff
@@ -0,0 +1,2024 @@
+// ==WindhawkMod==
+// @id              simple-window-switcher
+// @name            Simple Window Switcher
+// @description     Replaces the default Alt+Tab with a lightweight window switcher inspired by ExplorerPatcher's Simple Window Switcher
+// @version         1.0
+// @author          Lone
+// @github          https://github.com/Louis047
+// @include         explorer.exe
```
m417z, line 9: See https://github.com/ramensoftware/windhawk/wiki/Mods-as-tools:-Running-mods-in-a-dedicated-process

mods/simple-window-switcher.wh.cpp
```diff
@@ -0,0 +1,2024 @@
...
+// @compilerOptions -ldwmapi -luxtheme -lgdi32 -lshlwapi -loleaut32 -lole32 -lcomctl32 -lgdiplus
+// @architecture    x86-64
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Simple Window Switcher
+A lightweight Alt+Tab replacement for Windows, ported from the
+[Simple Window Switcher](https://github.com/valinet/sws) project.
+Additional improvements made by [Asteski](https://github.com/Asteski).
```
m417z, line 18: How about adding a screenshot?

Louis047:
@m417z Asteski was able to implement the second one but for some reason, the first one doesn't work at all. I'm not sure why. I currently can't check it as I'm traveling and will only be back by the next day. 

Update: I just found a lead for this. One of the main culprit is hard-coding the `@architecture` removed it. But there's a race condition now. Working on the fix.

Louis047:
Fixed the race condition but now there's another issue: 

When focused on admin windows, the mod doesn't work. But when focused on non-admin windows/desktop, the mod works without any issue. I'm not exactly sure what's wrong here. Maybe the child process doesn't elevate to admin ?

m417z:
With either the current approach (explorer.exe) or the isolated process approach (running as windhawk.exe), the mod is not elevated, and so it can't receive hotkey events when an elevated process is focused. I don't think there's an easy solution to this.

PowerToys, for example, has a similar limitation:
* https://www.thewindowsclub.com/weve-detected-an-application-running-with-administrator-privileges
* https://github.com/microsoft/PowerToys/wiki/PowerToys-and-running-as-Administrator/3ccb9d40b04df9c5295cf54b0f0a01016a57d670

Louis047:
> With either the current approach (explorer.exe) or the isolated process approach (running as windhawk.exe), the mod is not elevated, and so it can't receive hotkey events when an elevated process is focused.

Well this is kinda odd. The old `RegisterHotkey` method with `explorer.exe` version worked perfectly, which required an explorer restart. It worked in both cases (non-admin/admin windows)

So, is it fine to use it this way ?

@m417z sorry for the multiple pings


Update: 

I finally fixed this, did the following:
1. Added elevation-aware process creation in `Wh_ModAfterInit`:
    - If the launcher process is already elevated → uses `CreateProcessInternalW` (child inherits elevation)
    - If not elevated → uses `ShellExecuteExW` with "runas" verb to request elevation via UAC
2. Added `-lshell32` linker flag for `ShellExecuteExW`

m417z:
> The old `RegisterHotkey` method with `explorer.exe` version worked perfectly, which required an explorer restart. It worked in both cases (non-admin/admin windows)

Do you mean that Explorer is able to intercept hotkeys even if the foreground window is elevated? It might be the case, signed Microsoft processes sometimes have extra capabilities. Might be worth understanding how it works.

> I finally fixed this, did the following:
> 
> 1. Added elevation-aware process creation in `Wh_ModAfterInit`:
>    ....

Nice, but note that with some UAC settings it will cause the UAC prompt to be shown, so I believe there should be an option to disable it.

Louis047:
> Do you mean that Explorer is able to intercept hotkeys even if the foreground window is elevated?

Yes, because `RegisterHotKey` is registered globally with the OS kernel, the system posts `WM_HOTKEY` directly to the registering thread, which naturally bypasses UIPI regardless of the foreground window's elevation.

I feel that this method is much more reliable than making this mod run as a separate process, which does come with the limitation explained above. 

If  you want, I can do this:
- Revert back to `RegisterHotkey` method with `explorer.exe` version
- If not, I can add a toggle inside the mod settings to elevate the mod which throws an UAC prompt. 

Let me know your thoughts on the above

@m417z sorry for yet another ping 😅 

Have to sort the issues asap.

Asteski:
> Nice, but note that with some UAC settings it will cause the UAC prompt to be shown, so I believe there should be an option to disable it.

That would be very confusing for users and actually discourage them from using the mod. How about this - merge pr with mod in its current state, and once we'll figure this out, we'll implement the workaround for this, as it seems that it'd require deeper understanding of how it works. 

Hooking it into explorer makes the mod most reliable and doesn't require implementing workarounds, which may be annoying for some users.

m417z:
Wait, I'm not sure I'm following. Why not make it a separate process (tool mod targeting windhawk.exe), and also use `RegisterHotKey`? There are examples for such mods:
https://github.com/search?q=repo%3Aramensoftware%2Fwindhawk-mods%20whtool%20RegisterHotkey&type=code

Louis047:
`RegisterHotKey` with `MOD_ALT + VK_TAB` is a reserved system hotkey - the call returns `FALSE` when made from any process other than `explorer.exe` (the shell process). This is documented in the [RegisterHotKey API docs](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-registerhotkey) which states that the OS reserves certain key combinations like Alt+Tab. The other tool mods you linked (force-kill, move-window-to-monitor, etc.) work because they register normal key combinations, not shell-reserved ones like Alt+Tab. That's why I had to switch to a `WH_KEYBOARD_LL` hook to intercept Alt+Tab from outside explorer.

I hope this explanation makes sense @m417z 

Update: I'm going ahead and test it with `RegisterHotkey`

Louis047:
I tested using `RegisterHotKey` in the tool mod, and all Alt+Tab variants fail with `ERROR_HOTKEY_ALREADY_REGISTERED (1409)` - `explorer.exe` already holds those registrations. Only Alt+Backtick succeeds. This is why the previous version used a `WH_KEYBOARD_LL` hook instead. Here are the logs:

```
RegisterHotKey results: Alt+Tab=FAIL(err=1409) Alt+Shift+Tab=FAIL(err=1409) Alt+Ctrl+Tab=FAIL(err=1409) Alt+Shift+Ctrl+Tab=FAIL(err=1409) Alt+Backtick=OK(err=0)
```

mods/simple-window-switcher.wh.cpp
```diff
@@ -0,0 +1,2030 @@
+// ==WindhawkMod==
+// @id              simple-window-switcher
+// @name            Simple Window Switcher
+// @description     Replaces the default Alt+Tab with a lightweight window switcher inspired by ExplorerPatcher's Simple Window Switcher
+// @version         1.0
+// @author          Lone
```
Asteski, line 6: add me as an additional author like this:

`// @author          Lone, Asteski`

while keeping github url to your profile only

m417z:
> I tested using `RegisterHotKey` in the tool mod, and all Alt+Tab variants fail with `ERROR_HOTKEY_ALREADY_REGISTERED (1409)`

So if I understand correctly, the main reason your explorer.exe targeting approach works is because the mod is loaded before Explorer gets the chance to run, and registers the hotkey first. Still, I think that running the mod in a separate process is a good idea for system stability.

We can try a hybrid solution: use the tool mod approach targeting windhawk.exe, and also target explorer.exe. On mod init, check if you're running in explorer.exe, and if so, only do a minimal handling of `RegisterHotKey` - hook it and block Alt+Tab registration (like Disable Windows Shortcuts does). What do you think?

Louis047:
> We can try a hybrid solution: use the tool mod approach targeting windhawk.exe, and also target explorer.exe. On mod init, check if you're running in explorer.exe, and if so, only do a minimal handling of RegisterHotKey - hook it and block Alt+Tab registration (like Disable Windows Shortcuts does). What do you think?

Yeah this seems like a good idea, currently testing it 👍 

And @Asteski I'll be adding you as an author in the next commit :)

Louis047:
@m417z Hey, I just had a thought about this. Usually when a mod that wants to be injected into `explorer.exe`, especially the ones that handle special keybinds (like the current mod for eg), they require an explorer restart to get the complete control over the keybind. 

With the hybrid approach, I feel that it's very similar to the `RegisterHotkey` method with only injected into `explorer.exe` with an explorer restart prompt. During my testing, I found out that the hybrid approach still requires an explorer restart for it to work.

Louis047:
Here's the video showcase with the hybrid method, it still feels very similar:


https://github.com/user-attachments/assets/bbaedf41-63c1-4a5a-b644-d4ccffc40b22

m417z:
It is similar in terms of user experience (e.g. requiring a restart), but it's a better solution in terms of stability and avoiding Explorer hangs/crashes. For example, your mod runs as part of the main (Desktop) thread and uses `Sleep` in some cases which hangs the whole desktop. It can be improved, but there are many such possible scenarios, and having the mod running in a separate process makes explorer unaffected regardless of what the mod does.

Louis047:
@m417z Well, you already know by now that even using this mod with the separate process method has its own caveat where it wouldn't just work with admin windows but is stable enough, so does the `RegisterHotkey` method with `explorer.exe` mod injection. 

Do you have a possible solution in mind ? This PR has been running for almost two weeks now 😴

m417z:
I guess I'm missing something, but my hybrid suggestion (https://github.com/ramensoftware/windhawk-mods/pull/4060#issuecomment-4513309950) is about using both `RegisterHotkey` and a separate process, so I believe it should work with admin windows.

Louis047:
> but my hybrid suggestion (https://github.com/ramensoftware/windhawk-mods/pull/4060#issuecomment-4513309950) is about using both RegisterHotkey and a separate process, so I believe it should work with admin windows.

But the logs prove otherwise. I have already stated that `RegisterHotkey` hooked with `explorer.exe` still requires a system restart to take control over Alt+Tab key for SWS.

Sure enough, the separate process method works very seamless as it doesn't even require a system restart but the only limitation is that it won't play well when there are admin windows focused. I have already rectified this by elevating the process prematurely but there comes another caveat: every time the user logs in to the system, windhawk throws an UAC prompt which doesn't sound good for me.

m417z:
Yes, but you need to restart explorer either way, so it doesn't make the tool mod approach worse than running within explorer.exe.

m417z:
It feels to me that we're talking past each other, feel free to DM on Discord and we can discuss this.

Louis047:
@m417z Hey I have added the requested changes, but now there's only one issue:

The UWP Icons doesn't display correctly, shown in the screenshot below

image

Louis047:
I think I have finally fixed the existing issues with the hybrid approach. Please do look into them and let me know if this finally settles it 👍 

Screenshot proof that UWP Icons work:

image

Also, the PR validation check fails as I have added @Asteski as author now.

mods/simple-window-switcher.wh.cpp
```diff
@@ -0,0 +1,2450 @@
...
+        }
+    }
+    return hIcon;
+}
+
+typedef HRESULT (WINAPI *SHGetPropertyStoreForWindow_t)(HWND, REFIID, void**);
+static SHGetPropertyStoreForWindow_t g_SHGetPropertyStoreForWindow = nullptr;
+static const PROPERTYKEY kPkeyAppUserModelId = {
+    {0x9F4C2855, 0x9F79, 0x4B39, {0xA8, 0xD0, 0xE1, 0xD4, 0x2D, 0xE1, 0xD5, 0xF3}},
+    5
```
m417z: Instead of defining it manually, you can include `<propkey.h>` and use `PKEY_AppUserModel_ID`.

mods/simple-window-switcher.wh.cpp
```diff
@@ -0,0 +1,2450 @@
...
+        int desiredSizePx = (int)lParam;
+        
+        Wh_Log(L"Explorer IPC: Received icon request for HWND %p, size %d", hWndTarget, desiredSizePx);
+        
+        std::wstring aumid;
+        
+        if (!g_SHGetPropertyStoreForWindow) {
+            HMODULE hShell32 = GetModuleHandleW(L"shell32.dll");
+            if (!hShell32) hShell32 = LoadLibraryW(L"shell32.dll");
+            if (hShell32) g_SHGetPropertyStoreForWindow = (SHGetPropertyStoreForWindow_t)GetProcAddress(hShell32, "SHGetPropertyStoreForWindow");
```
m417z: Why not just call `SHGetPropertyStoreForWindow`? It's available since Windows 7.

mods/simple-window-switcher.wh.cpp
```diff
@@ -0,0 +1,2450 @@
...
+
+    // --- explorer.exe path: hook RegisterHotKey only ---
+    if (_wcsicmp(exeName, L"explorer.exe") == 0) {
+        g_isExplorer = true;
+        Wh_Log(L"SWS: Loaded into explorer.exe, hooking RegisterHotKey");
+
+        if (!g_WM_SWS_GET_UWP_ICON) {
+            g_WM_SWS_GET_UWP_ICON = RegisterWindowMessageW(L"Windhawk_SWS_GetUwpIcon");
+        }
+        CreateThread(NULL, 0, ExplorerIpcThread, NULL, 0, NULL);
```
m417z: The thread handle is never closed.

mods/simple-window-switcher.wh.cpp
```diff
@@ -0,0 +1,2450 @@
...
+        if (promptWnd) PostMessage(promptWnd, WM_CLOSE, 0, 0);
+
+        HWND hIpc = FindWindowW(L"WindhawkSWS_IpcWindow", NULL);
+        if (hIpc) {
+            PostMessageW(hIpc, WM_QUIT, 0, 0);
+        }
+        for (auto& pair : g_uwpIconCache) {
+            if (pair.second) DestroyIcon(pair.second);
+        }
+        g_uwpIconCache.clear();
```
m417z, line 2433: Wait for `ExplorerIpcThread` to end before clearing the icons.

mods/simple-window-switcher.wh.cpp
```diff
@@ -0,0 +1,2450 @@
...
+- rowWidth: 0
+  $name: Row Width
+  $description: Width of each thumbnail tile in pixels (before DPI scaling). Set to 0 for automatic width based on window aspect ratio.
+- maxWidthPercent: 80
+  $name: Maximum Width (percentage of screen width)
+- maxHeightPercent: 80
+  $name: Maximum Height (percentage of screen height)
+- windowPadding: 20
+  $name: Window Padding
+  $description: Padding around the entire window grid (pixels, before DPI scaling).
```
m417z: This option is not used in the code.

mods/simple-window-switcher.wh.cpp
```diff
@@ -0,0 +1,2450 @@
+// ==WindhawkMod==
+// @id              simple-window-switcher
+// @name            Simple Window Switcher
+// @description     Replaces the default Alt+Tab with a lightweight window switcher inspired by ExplorerPatcher's Simple Window Switcher
+// @version         1.0
+// @author          Lone, Asteski
```
m417z: Windhawk mod ownership and trust model allows a single author - the submitter who's responsible for the mod. Additional credits can be added in the readme.

mods/simple-window-switcher.wh.cpp
Louis047: Check this comment: https://github.com/ramensoftware/windhawk-mods/pull/4060#discussion_r3292892659

dainiunietis:
> Alt+Backtick

Hello, I used ExplorerPatcher's Simple Window Switcher only because of Alt+Backtick, unfortunately it doesn't work at all at the moment, if it worked in your implementation it would be very good. Thank You

PR #4082, opened 2026-05-17
Add Open in Admin Terminal mod

aimagist:
## Changelog

* Add the Open in Admin Terminal mod.

## Mod authorship

This mod was created by aimagist.

### DIFF
--- a/mods/open-in-admin-terminal.wh.cpp
+++ b/mods/open-in-admin-terminal.wh.cpp
@@ -1,0 +1,500 @@
+// ==WindhawkMod==
+// @id              open-in-admin-terminal
+// @name            Open in Admin Terminal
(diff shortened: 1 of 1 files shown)


m417z:
One of the main advantages of Windhawk is that mods don't do persistent changes to the system, unlike other solutions which modify the registry or system files. If something goes wrong, a mod can always be disabled to revert the changes.

Contrary to this, this mod makes changes to system registry entries, which doesn't align with Windhawk principles. One way to have a compatible implementation is to hook registry read functions and return different values. There are several examples for this:
https://github.com/search?q=repo%3Aramensoftware%2Fwindhawk-mods+REGQUERYVALUEEXW&type=code

aimagist:
Thanks, that makes sense. I hadn’t accounted for these principles.

I’ll try to rework this.

PR #4091, opened 2026-05-17
DWM Custom Projection Border v1.0.0

erisune:
This mod allows you to customize or disable the screen capture indicator (by default a yellow border) that is drawn by DWM on top of a window or display that is being captured using WinRT, specifically `Windows.Graphics.Capture`.

Windows 10 doesn't have a toggle to disable this border, so this mod should be particularly useful on that version.

## Examples

**Pink border on Windows 11 26H1**:

Pink border on Windows 11 26H1

**Disabled border on Windows 10 21H2**:

Disabled border on Windows 10 21H2

## Changelog

* N/A

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/dwm-custom-projection-border.wh.cpp
+++ b/mods/dwm-custom-projection-border.wh.cpp
@@ -111,8 +111,8 @@ struct {
     unsigned int innerSize;
     MARGINS margins;
 } g_modSettings;
-bool g_outerBorderDrawn = false;
-bool g_colorConverted = false;
+thread_local bool g_outerBorderDrawn = false;
+thread_local bool g_innerBorderDrawn = false;
 
 bool IsHighContrastMode()
 {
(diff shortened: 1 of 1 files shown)


mods/dwm-custom-projection-border.wh.cpp
```diff
@@ -0,0 +1,361 @@
...
+struct {
+    bool disableBorder;
+    bool drawOuterOnly;
+    D3DCOLORVALUE outerColor;
+    D3DCOLORVALUE innerColor;
+    unsigned int outerSize;
+    unsigned int innerSize;
+    MARGINS margins;
+} g_modSettings;
+bool g_outerBorderDrawn = false;
```
m417z: Can this value be accessed from multiple threads? If it can, it'd be safer to mark it as `thread_local`.

mods/dwm-custom-projection-border.wh.cpp
```diff
@@ -0,0 +1,361 @@
...
+HRESULT CProjectionBorderVisual___AddBorderInstructions_hook(
+    void *pThis,
+    LPRECT lpRect,
+    UINT uSize,
+    D3DCOLORVALUE *pColor)
+{
+    if (!g_colorConverted) {
+        Convert_MilColorF_sRGB_To_MilColorF_scRGB(&g_modSettings.outerColor);
+        Convert_MilColorF_sRGB_To_MilColorF_scRGB(&g_modSettings.innerColor);
+        g_colorConverted = true;
```
m417z: Why not run it in `LoadSettings`?

mods/dwm-custom-projection-border.wh.cpp
```diff
@@ -0,0 +1,361 @@
...
+
+    if (g_modSettings.disableBorder) {
+        return S_OK;
+    }
+
+    if (g_modSettings.drawOuterOnly && g_outerBorderDrawn) {
+        return S_OK;
+    }
+
+    SetBorderProperties(lpRect, uSize, pColor);
```
m417z: The caller doesn't expect `lpRect` and `pColor` to be modified, it's safer to create a copy and modify it instead.

mods/dwm-custom-projection-border.wh.cpp
erisune: Hey Michael, thanks for the input. I did this because when `LoadSettings` is called we still don't know if the color should actually be converted (this is only needed on versions prior to Windows 11 26H1). Perhaps I should just detect the OS build on `Wh_ModInit` like some other mods do. I'll address the other points as well.

m417z: If it's only for builds which have that function, you can simply check if `CProjectionBorderVisual___AddBorderInstructions_orig` is null after calling `HookSymbols`.

erisune: Yeah, that's exactly the case. Thank you, I'll do that.

PR #4119, opened 2026-05-20
Create file-explorer-reopen-closed-tab.wh.cpp

armaninyow:
## Mod authorship

This mod was created by:

- - [x] Claude
- - [ ] ChatGPT
- - [ ] Gemini


armaninyow:
# File Explorer Reopen Closed Tab

Press `Ctrl+Shift+T` in any active File Explorer window to reopen the last closed tab — just like in a browser.

- Restored tabs open directly in the existing Explorer window, not as a new window.
- If no Explorer window is open, nothing happens.
- Up to 50 tabs are remembered per session.
- History clears when all Explorer windows are closed.

m417z:
I tried the mod, and it doesn't work for me. Does it work for you?

mods/file-explorer-reopen-closed-tab.wh.cpp
```diff
@@ -0,0 +1,418 @@
+// ==WindhawkMod==
+// @id              file-explorer-reopen-closed-tab
+// @name            File Explorer Reopen Closed Tab
+// @description     Reopens the last closed Explorer tab when Ctrl+Shift+T is pressed in an active Explorer window
+// @version         1.0.0
+// @author          Armaninyow
+// @github          https://github.com/armaninyow
+// @include         explorer.exe
```
m417z, line 8: See https://github.com/ramensoftware/windhawk/wiki/Mods-as-tools:-Running-mods-in-a-dedicated-process

armaninyow:
> I tried the mod, and it doesn't work for me. Does it work for you?

Yeah it does. It's probably the windows 11 build version. Mine is Version 25H2 (OS Build 26200.6899).

~~Do i need to specify which version only works on?~~ I will try to inject windhawk.exe

mods/file-explorer-reopen-closed-tab.wh.cpp
armaninyow: > See https://github.com/ramensoftware/windhawk/wiki/Mods-as-tools:-Running-mods-in-a-dedicated-process

Changes done. I hope it works in your end.

m417z: I tried the updated version, but it still doesn't work.

m417z:
> Mine is Version 25H2 (OS Build 26200.6899).

I was testing it on build 10.0.26200.8457.

armaninyow:
@m417z 

Eventually it didn't work on mine.

Back to injecting into explorer.exe with no tool mod boilerplate.

Let me know!

m417z:
Why does it have to be injected into Explorer?

armaninyow:
I'm not sure I understand the question. This is for Windows 11 File Explorer.

Injecting it to windhawk.exe worked for me at first. But after a restart, it didn't work anymore. So i reverted it back to explorer.exe. It's back to working again, even after a restart.

m417z:
> But after a restart, it didn't work anymore

I'm asking why. But from your answer I assume you don't know, right?

Honestly, I don't feel comfortable merging a mod which sometimes works and sometimes doesn't, although the idea is great.

armaninyow:
Yes. I don't really know how to answer it technically.

Too bad it doesn't work in your end. So i undestand if you don't feel merging this mod. I can keep this local for now.

armaninyow:
Hey @m417z , I had Claude give you a technical answer:

> The mod uses IShellWindows (a COM interface) to enumerate and track Explorer tabs, and SendMessage to ShellTabWindowClass inside Explorer to open new tabs. Both of these work from any process — but ShellTabWindowClass is a child window inside Explorer's process space, and the SendMessage call to it is more reliable when made from within the same process rather than cross-process.

> More importantly, the keyboard hook (WH_KEYBOARD_LL) and the GetPrimaryExplorerHwnd check both need to run continuously in a stable process. When running in windhawk.exe via the tool mod pattern, the dedicated process launch mechanism (CreateProcessInternalW) is fragile and depends on timing — if Windhawk starts before the shell is fully initialized on boot, the process either doesn't launch or launches into a broken state.

> Injecting into explorer.exe sidesteps all of that because Explorer is always running, always has the shell initialized, and the SendMessage to ShellTabWindowClass is intra-process.

armaninyow:
Hey, I'm using the tool mod pattern again, but kept injecting explorer.exe instead of windhawk.exe. The reason is that the mod sends a message to ShellTabWindowClass (a child window inside Explorer) to open new tabs, which works more reliably intra-process. The boilerplate handles the mutex and single-instance guarantee so it won't run multiple times even with multiple explorer.exe processes. Would appreciate if you let me know if it works on your end.

m417z:
I checked why it doesn't work for me. The first issue is that `CoCreateInstance` inside `OpenAsTab` fails with error `CO_E_NOTINITIALIZED`. After calling `CoInitializeEx` in `HookThread`, the function returns another error: `RPC_E_CANTCALLOUT_ININPUTSYNCCALL`, meaning that `OpenAsTab` can't be called while inside `LowLevelKeyboardProc` - which is a bad idea in any case since all keyboard activity is blocked while inside the hook. Run this function in a separate thread.

Additionally, querying all open tabs in the system each 750 ms is not optimal. I suggest to try and avoid that. If there's no solution, at least run the polling only when an explorer window is the foreground window. You can subscribe to `EVENT_SYSTEM_FOREGROUND` events, and enable/disable the polling accordingly.

armaninyow:
Both issues are fixed:

1. CO_E_NOTINITIALIZED — CoInitializeEx is now called at the start of HookThread and RestoreThread.

2. RPC_E_CANTCALLOUT_ININPUTSYNCCALL — OpenAsTab is no longer called inside LowLevelKeyboardProc. The hook now just posts a WM_APP message to a dedicated RestoreThread, which has its own COM-initialized message loop and handles the actual tab opening safely.

For the polling optimization, EVENT_SYSTEM_FOREGROUND is subscribed via SetWinEventHook, installed on HookThread so it has a message pump to receive the callbacks. The poll thread now checks a flag and skips the COM enumeration entirely when Explorer isn't the foreground window, sleeping 150ms instead.

Let me know if it works on your end!

mods/file-explorer-reopen-closed-tab.wh.cpp
```diff
@@ -0,0 +1,685 @@
...
+                break;
+            }
+            Sleep(1000);
+        }
+    }
+
+    while (g_bRunning) {
+        // Only poll when Explorer is in the foreground — saves CPU otherwise
+        if (!g_bExplorerForeground) {
+            Sleep(150);
```
m417z, line 417: You could do that with an event and avoid the looping altogether. You might want to improve that in one of the future updates.

m417z (commented):
It seems to be working now, thanks.

mods/file-explorer-reopen-closed-tab.wh.cpp
armaninyow: Thank you! And yes I will try and do that in the next updates.

Issue #2491, opened 2025-10-03
[Better file sizes in Explorer details] Missing folder size in Content View and Tiles View.


leomate:
I'm using Windhawk 1.6.1
with mods "Better file sizes in Explorer details" by @m417z 

I'm actually using this mod for long time since late 2024, and for now i just realized if the folder size didn't appear in Content & Tiles View mode. I think it's just a minor thing, but i just want to make sure to more refined.

Like when i set from details it would see the folder size like this

<img width="626" height="190" alt="Image" src="https://github.com/user-attachments/assets/1ba86265-a17a-45ed-8d97-84c0c2d7838d" />

When i set to Tiles View, i think the folder size is missing

<img width="707" height="166" alt="Image" src="https://github.com/user-attachments/assets/d5c2a087-ba4c-4335-a544-dd98ceffc329" />

Same story when i set to Content View

<img width="827" height="367" alt="Image" src="https://github.com/user-attachments/assets/644daab8-0ebd-4f8d-ba55-a0c426473a30" />

I hope u get what my idea :D

leomate:
Okay, just to add something i just notice now

When i had select folder, in classic bottom details pane also didn't show up the folder size informations. I think it need to be added by registry like the previous post

Folder details info
<img width="355" height="102" alt="Image" src="https://github.com/user-attachments/assets/33ab8f94-d532-47b1-9c63-4525f562a82b" />

File details info
<img width="340" height="104" alt="Image" src="https://github.com/user-attachments/assets/81c0fd8d-834a-4d40-b3b8-018c0ca5f429" />

AromaKitsune:
Updated these registry tweaks in my previous comment. I also added the date created (`System.DateCreated`) property to match the file details info.

With the `System.Size` property being added to the `PreviewDetails` value, it makes the status bar's and bottom details pane's total file size sum not disappear even if the multi-selection contains folders! 😄 Relevant issue post: https://github.com/ramensoftware/windhawk-mods/issues/2248

However, for status bar, it still disappears if 100+ items are selected. For bottom details pane, all details are hidden and replaced with the "Show more details..." button if 16+ items are selected, clicking that button shows those details again.
<img width="800" height="320" alt="Image" src="https://github.com/user-attachments/assets/a4da7c0a-fe8e-47a1-bd94-5e0b683672a9" />

I enabled showing a current folder's size in status bar (`StatusBar` value) too.
[...]

PR #4133, opened 2026-05-22
Add explorer-status-metadata v1.3.0

VitalSkib:
Add explorer-status-metadata v1.3.0

Appends rich file metadata (dimensions, date modified, file type,
duration, bitrate, fps) to the Windows Explorer status bar when
a single file is selected. Zero polling — purely reactive,
path-keyed cache with LRU eviction.

## Changelog
Initial release.

## Mod authorship

This mod was created by:
- [x] Claude

### DIFF
--- a/mods/explorer-status-metadata.wh.cpp
+++ b/mods/explorer-status-metadata.wh.cpp
@@ -4,7 +4,7 @@
 // @description     Appends rich metadata (dimensions, date, type, duration, bitrate, fps) to Explorer's status bar. Zero polling — purely reactive, path-keyed cache.
 // @version         1.3.0
 // @author          VitalS
-// @github          VitalSkib
+// @github          https://github.com/VitalSkib
 // @include         explorer.exe
 // @compilerOptions -lgdi32 -luser32 -lole32 -loleaut32 -luuid -lpropsys
 // ==/WindhawkMod==
--- a/mods/explorer-status-metadata.wh.cpp
+++ b/mods/explorer-status-metadata.wh.cpp
@@ -1,0 +1,839 @@
+// ==WindhawkMod==
+// @id              explorer-status-metadata
+// @name            Explorer Status Bar Metadata
+// @description     Appends rich metadata (dimensions, date, type, duration, bitrate, fps) to Explorer's status bar. Zero polling — purely reactive, path-keyed cache.
+// @version         1.3.0
+// @author          VitalS
+// @github          VitalSkib
+// @include         explorer.exe
+// @compilerOptions -lgdi32 -luser32 -lole32 -loleaut32 -luuid -lpropsys
+// ==/WindhawkMod==
+
+// ==WindhawkModSettings==
+/*
+- allowNetworkDrives: false
+  $name: Enable metadata on network drives (UNC paths)
+  $description: >
+    WARNING: May cause Explorer to freeze for several seconds if the network
+    is slow or unavailable. The metadata lookup is synchronous on the UI thread.
+    Only enable this if you are on a fast, reliable local network.
+
+- allowRemovableDrives: true
+  $name: Enable metadata on removable drives (USB, SD cards)
+  $description: >
+    Metadata lookup on slow removable media (e.g. old USB drives) may cause
+    a brief UI stutter. Fast USB 3.x drives are generally fine.
+
+- cacheEvictCount: 10
+  $name: Cache eviction count
+  $description: >
+    When the cache reaches 1000 entries, this many of the oldest entries
+    are removed. Range: 1-50. Lower values preserve more history;
+    higher values free more memory at once.
+*/
+// ==/WindhawkModSettings==
+
+// ==WindhawkModReadme==
+/*
+# Explorer Status Bar Metadata — v1.3.0
+
+Appends file metadata to the Windows Explorer status bar when a single file
+is selected: dimensions, modification date, file type, duration, bitrate, fps.
+
+## How it works
+- No timers, no polling. Metadata is computed only when the selected file changes.
+- Path-keyed LRU-style cache (1000 entries). Same file = instant return, zero COM calls.
+- Hooks `DrawTextW` and `GetTextExtentPoint32W` in the Explorer UI thread.
+- Custom binary parsers for formats the Windows Property System does not index:
+  SVG, HDR, EXR, TIFF/TX (LE+BE), TGA, WebM/MKV (EBML).
+
+## Settings
+- **Network drives** — disabled by default. Can freeze Explorer on slow networks.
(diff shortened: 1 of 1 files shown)


mods/explorer-status-metadata.wh.cpp
```diff
@@ -0,0 +1,839 @@
...
+static bool IsSelectionStatus(const wchar_t* s, UINT len)
+{
+    if (!s || len < 2) return false;
+    UINT checkLen = (len < 512u) ? len : 512u;
+
+    for (UINT i = 0; i + 7 < checkLen; ++i) {
+        wchar_t c = s[i];
+        // English: "selected" / "Selected"
+        if ((c == L's' || c == L'S') && wcsncmp(s + i + 1, L"elected", 7) == 0)
+            return true;
```
m417z: That's quite sloppy for multiple reasons, including:
* Other languages aren't supported.
* If the same string is used elsewhere, it will be replaced too. For example, try to rename a file to "selected".

Perhaps it'd be better to use the native Windows column functionality by hooking registry querying functions. Some information can be found here: https://github.com/ramensoftware/windhawk-mods/issues/2491#issuecomment-3369473511, and the "Better file sizes in Explorer details" can be used as a reference for registry querying hooks.

mods/explorer-status-metadata.wh.cpp
```diff
@@ -0,0 +1,839 @@
...
+{
+    outExplorerRoot = NULL;
+    HWND hFocus = NULL;
+    GUITHREADINFO gti = { sizeof(GUITHREADINFO) };
+    if (GetGUIThreadInfo(GetCurrentThreadId(), &gti))
+        hFocus = gti.hwndFocus ? gti.hwndFocus : gti.hwndActive;
+    if (!hFocus) hFocus = GetForegroundWindow();
+
+    for (HWND h = hFocus; h; h = GetParent(h)) {
+        auto* pSB = (IShellBrowser*)SendMessageW(h, CWM_GETISHELLBROWSER, 0, 0);
```
m417z: `CWM_GETISHELLBROWSER` is defined as `WM_USER+7`. Other windows might handle it differently and do some unintended action, so trying various windows like that is not safe.

mods/explorer-status-metadata.wh.cpp
```diff
@@ -0,0 +1,839 @@
...
+
+BOOL Wh_ModInit()
+{
+    Wh_Log(L"[Init] Explorer Status Bar Metadata v1.3.0 starting");
+
+    // Read settings (Windhawk guarantees these are available during Wh_ModInit)
+    g_allowNetworkDrives   = Wh_GetIntSetting(L"allowNetworkDrives")   != 0;
+    g_allowRemovableDrives = Wh_GetIntSetting(L"allowRemovableDrives") != 0;
+    int evictRaw           = Wh_GetIntSetting(L"cacheEvictCount");
+    g_cacheEvictCount      = (evictRaw >= 1 && evictRaw <= 50) ? (size_t)evictRaw : 10;
```
m417z, line 947: Settings are not reloaded (`Wh_ModSettingsChanged` is missing).

mods/explorer-status-metadata.wh.cpp
VitalSkib: Hi m417z,

I've been rewriting the mod using the `RegQueryValueExW` approach you suggested, but hit a wall and wanted to check if I'm missing something.

**What I tested:**

I hooked `RegQueryValueExW` and logged every `StatusBar` value read. When navigating folders, only `HKCR\Directory\StatusBar` is ever read — and it doesn't even exist by default (I get `ERROR_FILE_NOT_FOUND`, inject our `prop:` string, and it works for the folder itself).

When I **select files inside the folder** (JPEGs, videos, etc.) — nothing. No `StatusBar` read happens at all, for any key:
- `HKCR\SystemFileAssociations\image\StatusBar` — never read
- `HKCR\SystemFileAssociations\video\StatusBar` — never read  
- `HKCR\*\StatusBar` — never read

I also manually wrote a `prop:` string directly to `HKCR\CURRENT_USER\Software\Classes\SystemFileAssociations\image\StatusBar` in regedit — no effect on the status bar when selecting images.

**Conclusion:** Explorer doesn't go through `RegQueryValueExW` at all when determining what to show in the status bar for selected files. The only registry read is for `Directory` when the folder itself is the subject.

**Question:** Did you have a specific key or mechanism in mind when you suggested the registry approach? Or is the status bar for selected files driven by hardcoded logic inside Explorer that can only be reached via the GDI hooks after all?

Any pointer would be really appreciated.

VitalSkib: Quick update with more data. I added verbose logging for all property-list registry keys (StatusBar, InfoTip, TileInfo, PreviewDetails, FullDetails, QuickTip).
When selecting a file, Explorer reads InfoTip and PreviewDetails from SystemFileAssociations\.ext — but not StatusBar. The status bar text for a selected file never goes through RegQueryValueExW at all.
I also manually set InfoTip at HKLM\SOFTWARE\Classes\SystemFileAssociations\.jpg — it only affects the hover tooltip, not the status bar on selection.
So it seems the status bar for a selected file is built internally without any registry lookup. Is there a COM interface or a vtable hook that controls this? Or is DrawTextW/GetTextExtentPoint32W the only practical way after all?

m417z: Not sure, I haven't researched this area too much. @AromaKitsune any quick insights?

> Or is DrawTextW/GetTextExtentPoint32W the only practical way after all?

I'm sure it isn't, there must be a better point of injection, perhaps something that requires debug symbols, but not necessarily. It'd require some research. Even with DrawTextW/GetTextExtentPoint32W, it should be possible to find a way to limit it to the status bar with more certainty, perhaps with `WindowFromDC`, by hooking a status bar function that calls these functions, or via other means.

AromaKitsune: I've been messing around with the `StatusBar` value, it doesn't work with files. Probably this key only works with `HKLM\SOFTWARE\Classes\Directory` and `HKLM\SOFTWARE\Classes\Drive` keys.

VitalSkib:
Addressed all review comments:
> - `CWM_GETISHELLBROWSER` is now sent only to windows matching the current thread ID and known-safe classes
> - Cache upgraded to true LRU (hit moves entry to back of eviction queue)
> - SVG parser rewritten with whitespace-tolerant `ExtractAttribute` helper
> - `Wh_ModSettingsChanged` added
> - Strategy 3 (cross-window fallback) removed

VitalSkib:
I've updated the thread safety guard. 
Switched from Thread ID to Process ID to properly handle calls from shell worker threads (which caused silent failures after PC restarts).

PR #4144, opened 2026-05-23
Add filtering of injected events, inject dwm.exe now so to work in el…

Anixx:
…evated windows

Filter out injected events so to bypass keyboard redirectors, inject dwm.exe now so to work in elevated windows.

## Changelog

* Changelog item 1...
* Changelog item 2...

## Mod authorship

This mod was created by:

- - [x] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/key-click.wh.cpp
+++ b/mods/key-click.wh.cpp
@@ -2,17 +2,19 @@
 // @id              key-click
 // @name            Key Click
 // @description     Produces click sound on keypress, supports autorepeat
-// @version         1.1
+// @version         1.2
 // @author          Anixx
 // @github          https://github.com/Anixx
 // @compilerOptions -lwinmm
-// @include         windhawk.exe
+// @include         dwm.exe
 // ==/WindhawkMod==
 
 // ==WindhawkModReadme==
 /*
 Produces click sound on keypress, supports autorepeat.
 Modifier keys (Ctrl, Shift, Alt, Win) do not produce autorepeat sounds.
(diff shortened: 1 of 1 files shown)


m417z:
For stability, is it possible to target dwm.exe, but still keep using the tool mod? Such that the mod will spawn in a separate dwm.exe process. Have you tried that?

Also, dwm.exe is excluded by default. Add a note for users similar to this:
https://github.com/ramensoftware/windhawk-mods/blob/42e0209f2d49af8f06d3df8d1f19edb15ec1dbcd/mods/custom-corner-radius.wh.cpp#L36-L42

**Edit**: Even better, you can target `winlogon.exe`, and then the note won't be necessary.

Anixx:
Done! Regarding separate dwm.exe looks complicated.

Anixx:
Also, my concern is: having two dwm.exe processes may break those mods that inject DWM (for custom windows borders, etc).

m417z:
Have you seen my edit about winlogon.exe?

Anixx:
It does not work as part of Winlogon.exe. Only as part of dwm.exe.

m417z:
> It does not work as part of Winlogon.exe. Only as part of dwm.exe.

I forked v1.1 and replaced `windhawk.exe` with `winlogon.exe`, and it works for me with elevated windows. What doesn't work for you?

Anixx:
With winlogon.exe (non-toolmod injection) the mod does not work for me at all, an any windows. As a tooolmod it works but not in elevated windows. 

In elevated works only when injecting dwm.exe

PR #4149, opened 2026-05-23
Add Caps IME Switcher mod

ZeonXr:
## Caps IME Switcher

Brings macOS-style Caps Lock behavior to Windows.

- **Short-press Caps Lock** → cycle through input languages
- **Long-press Caps Lock** → toggle Caps Lock on/off

Switches input languages via `WM_INPUTLANGCHANGEREQUEST` directly — no simulated shortcuts.

Optionally keeps Simplified Chinese IMEs in Chinese mode, preventing accidental English input when switching to a Chinese IME.

## Changelog

* Initial release

## Mod authorship

This mod was created by:

- - [x] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/caps-ime-switcher.wh.cpp
+++ b/mods/caps-ime-switcher.wh.cpp
@@ -2,11 +2,11 @@
 // @id              caps-ime-switcher
 // @name            Caps IME Switcher
 // @description     Brings the macOS Caps Lock input source switching behavior to Windows.
-// @version         0.6
+// @version         0.7
 // @author          ZeonXr
 // @github          https://github.com/ZeonXr
-// @include         *
-// @compilerOptions -luser32 -lshell32 -limm32
+// @include         windhawk.exe
+// @compilerOptions -luser32 -lshell32
 // @license         MIT
 // ==/WindhawkMod==
 
@@ -14,15 +14,22 @@
 /*
 # Caps IME Switcher
 
+Brings the macOS Caps Lock input source switching behavior to Windows.
+
 Short-press Caps Lock to switch to the next input language.
 Long-press Caps Lock to toggle Caps Lock on/off.
 
-Switches input languages by posting WM_INPUTLANGCHANGEREQUEST
-directly to the focused window.
+**Pair with** IME Native Mode Lock (`ime-mode-lock`) — auto-locks
+IMEs to their native mode. Best experience when used together.
 
-Optionally keeps Simplified Chinese IMEs in Chinese mode, so switching
-to a Chinese IME always lands in native input mode instead of accidental
-English mode.
+---
+
+将 macOS 的 Caps Lock 输入法切换方式带到 Windows。
+
+短按 Caps Lock 切换下一个输入语言，长按触发大小写锁定。
+
+**推荐搭配** IME Native Mode Lock (`ime-mode-lock`) — 可在切换输入法时
+自动锁定中文模式，两个 mod 配合使用体验最佳。
 */
 // ==/WindhawkModReadme==
 
@@ -31,9 +38,6 @@ English mode.
 - LongPressMs: 300
   $name: Long press threshold (ms)
   $description: How long to hold Caps Lock (in milliseconds) before it toggles Caps Lock instead of switching input language
-- ForceChineseNativeMode: false
-  $name: Keep Simplified Chinese IMEs in Chinese mode
-  $description: Automatically switch Simplified Chinese IMEs to Chinese input mode when activated
 - PassThroughProcesses: ""
   $name: Caps Lock pass-through process list
   $description: Comma-separated process names where Caps Lock should not be intercepted (e.g. "Moonlight.exe,mstsc.exe"). Useful for remote desktop clients.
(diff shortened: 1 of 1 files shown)


mods/caps-ime-switcher.wh.cpp
```diff
@@ -0,0 +1,753 @@
+// ==WindhawkMod==
+// @id              caps-ime-switcher
+// @name            Caps IME Switcher
+// @description     Brings the macOS Caps Lock input source switching behavior to Windows.
+// @version         0.6
+// @author          ZeonXr
+// @github          https://github.com/ZeonXr
+// @include         *
```
m417z: Is targeting all processes truly necessary?

After the worker thread posts `WM_INPUTLANGCHANGEREQUEST` from `RequestNextInputSource`, is it possible to call `ImmGetDefaultIMEWnd(targetWindow)` and `SendMessageTimeoutW(imeWindow, WM_IME_CONTROL, IMC_SETCONVERSIONMODE, ...)` directly instead of hooking `DispatchMessageA`/`DispatchMessageW`?

If the mod can become a pure tool mod (`@include windhawk.exe` according to https://github.com/ramensoftware/windhawk/wiki/Mods-as-tools:-Running-mods-in-a-dedicated-process), it will become much simpler and safer.

ZeonXr: The `@include *` is only required when the user enables the `ForceChineseNativeMode` setting (disabled by default). When that setting is off, the mod functions as a pure Caps Lock input source switcher — no hooks are installed in application processes.

When `ForceChineseNativeMode` is enabled, we need to monitor `MSUIM.Msg.Private` and `WM_IME_NOTIFY` messages inside each application because the Microsoft Pinyin IME (and other Simplified Chinese IMEs) can spontaneously switch back to English mode. This happens, for example, when focus moves between controls, when certain applications reset the IME state, or when the user presses Shift. Without a `DispatchMessage` hook in the target process, these mode-change events are invisible to a tool-mod process — they are posted to the application's own window procedure, not broadcast globally.

Calling `ImmGetDefaultIMEWnd` + `SendMessageTimeoutW` from the tool-mod thread after `WM_INPUTLANGCHANGEREQUEST` can set the mode once, but cannot prevent subsequent spontaneous switches back to English. The existing `chinese-ime-fixed-mode` mod in this repository uses the same approach (`@include *` + DispatchMessage hook + MSUIM.Msg.Private) for the same reason.

If there is a way to receive these IME mode-change notifications from a tool-mod context (e.g., a global hook or event that I'm not aware of), I'd be happy to switch to a pure tool-mod design.

m417z: In this case, perhaps it's better to have two mods. One mod for what the description says:
> Short-press Caps Lock to switch to the next input language.
> Long-press Caps Lock to toggle Caps Lock on/off.

This mod won't inject into other programs, and is likely what most users will want.

And another mod for Chinese handling that targets all processes, only for those who need this specific functionality.

What do you think? Does it make sense?

ZeonXr: I agree this is a better approach. My original intent was to keep the number of mods minimal, but you're right that separating concerns is cleaner. I'll strip out the Chinese IME locking functionality and keep this mod focused solely on Caps Lock input source switching as a pure tool-mod.

Regarding the Chinese IME fixed mode: I noticed the repo already has `chinese-ime-fixed-mode`, but after testing it I found it's not quite complete. I'd like to submit a separate PR to improve it. However, I'm not the author of that mod, and my changes would likely involve significant refactoring and restructuring. Would that kind of substantial rework on someone else's mod be acceptable?

m417z: > Would that kind of substantial rework on someone else's mod be acceptable?

Only the author is able to submit updates to their mods. But for this mod, it looks like the author is no longer on GitHub ([@barrypp](https://github.com/barrypp)). It'd be best to submit a new mod.

ZeonXr:
I've extracted the IME mode locking functionality into a separate mod `ime-mode-lock` to keep each mod focused on a single responsibility.

The new mod has been submitted in PR #4233

mods/caps-ime-switcher.wh.cpp
```diff
@@ -0,0 +1,625 @@
...
+        }
+    }
+
+    if (g_workerThread) {
+        DWORD waitResult = WaitForSingleObject(g_workerThread, 5000);
+        if (waitResult == WAIT_TIMEOUT) {
+            Wh_Log(L"Worker thread did not exit in time");
+            // Safe here because ExitProcess(0) follows in Wh_ModUninit,
+            // which will clean up all thread resources.
+            TerminateThread(g_workerThread, 1);
```
m417z: Avoid using `TerminateThread`:

> [stop calling Terminate­Thread](https://devblogs.microsoft.com/oldnewthing/20150814-00/?p=91811). There are no valid use cases for it. Any time you call it, you will corrupt the target process

In this case, as the comment mentions, the process will exit anyway, and all threads will be terminated.

mods/caps-ime-switcher.wh.cpp
```diff
@@ -0,0 +1,625 @@
...
+        }
+        CloseHandle(g_workerThread);
+        g_workerThread = nullptr;
+    }
+}
+
+// Prevents the tool mod child process from running its normal entry point.
+// The worker thread keeps running; this effectively makes the process a
+// container for our keyboard hook logic.
```
m417z: For the code below, please keep it identical to the snippet in the wiki for easier review:

Issue #3973, opened 2026-05-03
[Disable Windows Shortcut] Popup shows on startup and when running Powertoys (due to "incompabilty")


TheGamer1445891:
Popup (the dialog that prompts you to restart explorer) shows on startup and when running Powertoys (due to "incompabilty") for some reason. (and some mods might also make the popup appear if it does something with the explorer)

the only way to maybe fix the popup appearing is by making an option to automatically restart explorer, and for fixing the powertoys incompabilty, idk (i didnt find any feature that does something with the windows shortcuts keys).

@Louis047

Louis047:
Hey @TheGamer1445891 Sorry I have been busy working on [this PR](https://github.com/ramensoftware/windhawk-mods/pull/4060) for the past two weeks now. It required a lot of changes to be done so couldn't really focus on this. Once the PR is merged, I'll start working on this after a short break. 

Once again, sorry for the long delay 😓

Louis047:
I did a preliminary test now and seems like you were right about powertoys, it spams explorer restart prompt during startup which is annoying.

TheGamer1445891:
I will close this issue

PR #4150, opened 2026-05-23
Update Disable Windows Shortcuts Mod

Louis047:
## Changelog

- Fixed an annoying issue for users with PowerToys installed where it triggers the explorer restart prompt every time during startup. Addresses #3973

### DIFF
--- a/mods/disable-windows-shortcuts.wh.cpp
+++ b/mods/disable-windows-shortcuts.wh.cpp
@@ -2,7 +2,7 @@
 // @id              disable-windows-shortcuts
 // @name            Disable Windows Shortcuts
 // @description     Selectively disable Windows keyboard shortcuts with individual toggles
-// @version         1.2.0
+// @version         1.1.1
 // @author          Lone
 // @github          https://github.com/Louis047
 // @include         explorer.exe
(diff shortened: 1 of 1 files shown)


mods/disable-windows-shortcuts.wh.cpp
```diff
@@ -1064,9 +1101,13 @@ BOOL Wh_ModInit()
...
         // We use Win+R as a probe. If it fails, Explorer is mid-session and already owns it.
-        if (!IsFirstTimeInit())
+        // We only do this for the main Explorer process, and only if it's been running for a while (>30s),
+        // to avoid false prompts on system startup or when secondary Explorers are launched.
+        if (IsMainExplorer() && IsExplorerUptimeLarge() && RegisterHotKey_Original && !IsFirstTimeInit())
         {
-            if (!RegisterHotKey(NULL, 0x1337, MOD_WIN | MOD_NOREPEAT, 'R'))
+            // We must use RegisterHotKey_Original to bypass our own hook, which would otherwise 
+            // return FALSE if the user has DisableWinR enabled, causing a false positive prompt.
+            if (!RegisterHotKey_Original(NULL, 0x1337, MOD_WIN | MOD_NOREPEAT, 'R'))
```
m417z: As we discussed in the previous pull request, you can use `RegisterHotKey` here because the hooks are applied only after `Wh_ModInit` returns.

m417z:
Perhaps this change should also be applied for Simple Window Switcher.

Louis047:
> Perhaps this change should also be applied for Simple Window Switcher.

Hmm, could you elaborate more on this 😅

m417z:
Here, it may have the same problem, no?

https://github.com/ramensoftware/windhawk-mods/blob/b6e013b94fe87b954b5fa2621b7388d150cfae22/mods/simple-window-switcher.wh.cpp#L2243-L2255

Louis047:
@m417z 

Good catch. SWS's current logic actually managed to dodge the PowerToys restart bug natively since the PID changes when Explorer restarts. However, you're absolutely right that the PID-tracking state machine is still vulnerable to other edge cases (like secondary folder windows silently breaking the prompt, or false prompts on a completely fresh install).

I agree with your suggestion. I'll replace the clunky LastPID logic entirely with the elegant `RegisterHotKey` probe approach we used in DWS. By probing Alt+Tab `(RegisterHotKey(NULL, 0x1337, MOD_ALT, VK_TAB))` alongside `IsMainExplorer()` and `IsExplorerUptimeLarge()`, we can perfectly detect early startup vs mid-session without relying on PIDs at all.

PR #4115, opened 2026-05-20
Dynamic Island for windows

devcode90:
### Description ✨

This pull request introduces **Dynamic Island for Windows**, a premium, fluid, and hardware-accelerated pill overlay inspired by the iPhone's Dynamic Island. Built from the ground up utilizing native Windows APIs and hardware-accelerated Direct2D rendering, it runs at a buttery-smooth 60 FPS with virtually zero impact on system resources.

### Features 🌟
- **Hardware Privacy Indicators:** Live pulsing dots indicating microphone (orange) or camera (green) activity with rate-limited, low-overhead polling.
- **Rich Media Integration:** Shows live audio waveform, album art, track details, and interactive playback controls.
- **Clipboard & Notifications:** Instant high-fidelity notifications and clipboard history with crisp 64px system icons extracted asynchronously.
- **System Dashboard / Game Overlay:** Real-time metrics including FPS, CPU, RAM, and disk utilization.
- **Fluid Spring Animations:** Natural transitions, smooth expansion on hover, and automatic splitting/merging behavior for concurrent events.

---

### Mod Authorship ✍️

- [Himanshu ] The submitter, without AI assistance
- [x] The submitter, with AI assistance
- [ ] Claude
- [ ] ChatGPT
- [x] Gemini (Antigravity)
- [ ] Another AI (please specify):
- [ ] Other (please specify):

PR #4152, opened 2026-05-23
Dynamic Island for Window

devcode90:
Follow-up fix to #4115.

The mod was not loading for public catalog users after installation. The root cause was that `windhawk.exe` is excluded from injection by default (to prevent recursion), so the tool-mod launcher was never executed and the Dynamic Island UI never appeared.

### Changes

- **`@include explorer.exe`** — `explorer.exe` always runs in the active user session (Session 1) and is always injected by default. This guarantees the launcher executes on startup for every user without any manual configuration.
- **Removed `sessionId == 0` check in `Wh_ModInit`** — not needed anymore since `explorer.exe` never runs in Session 0.
- **Dynamic settings launcher (Case 9)** — instead of passing a hardcoded `local@` URL, it now reads Windhawk's installation path from the registry (`SOFTWARE\Wow6432Node\Microsoft\Windows\CurrentVersion\Uninstall\Windhawk` → `DisplayIcon`) and opens it with `windhawk://mods/<WH_MOD_ID>`. This works correctly for both local drafts and the public catalog release.

### Mod Authorship ✍️
- [x] The submitter, without AI assistance
- [ ] Claude
- [ ] ChatGPT
- [ ] Gemini (Antigravity)
- [ ] Other (please specify):

### DIFF
--- a/mods/dynamic-island-for-windows.wh.cpp
+++ b/mods/dynamic-island-for-windows.wh.cpp
@@ -2,11 +2,10 @@
 // @id              dynamic-island-for-windows
 // @name            Dynamic Island for Windows
 // @description     A living, breathing pill overlay inspired by iPhone's Dynamic Island. Reacts to media, downloads, clipboard, battery, and more.
-// @version         1.0.0
+// @version         1.0.1
 // @author          Himanshu
 // @github          https://github.com/devcode90
 // @include         windhawk.exe
-// @architecture    x86-64
 // @compilerOptions -lole32 -loleaut32 -lshcore -ld2d1 -ldwrite -ldwmapi -lgdi32 -luser32 -lshell32 -lruntimeobject -lwindowscodecs -lavrt -lsetupapi
 // @license         MIT
 // ==/WindhawkMod==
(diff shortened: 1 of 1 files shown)


devcode90:
@m417z 

Hi @m417z, just to flag — after the merge it was discovered that the mod doesn't load for public catalog users because `windhawk.exe` is excluded from injection by default. I've opened a follow-up fix in #4152 that transitions the launcher to `explorer.exe` (the standard approach used by other tool mods). Sorry for the trouble and thank you for your patience!

m417z:
> The root cause was that `windhawk.exe` is excluded from injection by default (to prevent recursion)

That's not true, of course, many mods do that and they work fine. The reason it doesn't work is this line:
```
// @architecture    x86-64
```

Remove it.

Also, please test the mod before submitting an update.

devcode90:
@m417z  i have tested and now it is compiling without error.

m417z:
Undo the other changes to the mod, just remove the architecture line.

m417z:
And the version number of course.

devcode90:
@m417z i have made the changes.

PR #4159, opened 2026-05-24
Add Desktop root and expandable This PC to nav top

rodboev:
## Summary

This mod adds the Desktop namespace root and an expandable This PC entry to the top of Explorer's left nav.

Windows lets you pin folders to Quick Access, but the Desktop root (which contains Recycle Bin, Control Panel, etc.) can't be pinned. There is also no way to move the expandable This PC up; if it's pinned to Quick Access, it isn't expandable.

This mod adds one or both of these virtual folders in the top part of the nav, above Quick Access and before Home/Gallery.

## Options

- Whether each entry is expandable (This PC shows drives, for example)
- Whether This PC starts expanded when a window opens
- Ordering (Desktop above or below This PC)
- Hiding duplicate This PC/Desktop entries elsewhere in the nav pane
- Hiding native Home/Gallery entries (does not affect Quick Access)
- Removing separators individually
- Fixing clipped/pixelated chevrons

### DIFF
--- a/mods/add-virtual-folders-to-nav-top.wh.cpp
+++ b/mods/add-virtual-folders-to-nav-top.wh.cpp
@@ -2,7 +2,7 @@
 // @id              add-virtual-folders-to-nav-top
 // @name            Add This PC and Desktop to Nav Top
 // @description     Adds This PC and Desktop to the top of Explorer's nav
-// @version         1.1.17
+// @version         1.1.18
 // @author          Rod Boev
 // @github          https://github.com/rodboev
 // @include         *
@@ -236,6 +236,7 @@ struct TreeState {
     HTREEITEM boundaryItem = nullptr;
     HTREEITEM belowQAItem = nullptr;
     uint8_t pendingWork = 0;   // bitmask of PendingWork flags
+    int sepRetries = 0;
     bool ownsNscRef = false;
     bool triedHomeSpacer = false;
     HIMAGELIST savedStateImageList = nullptr;
@@ -632,7 +633,7 @@ static void DrawChevron(HDC hdc, const RECT *r, int partId, int stateId)
 static thread_local bool g_inTreePaint = false;
 static thread_local int g_inSubclassProc = 0;
 static thread_local int g_lastSepY = -1;
-static thread_local int g_sepRetries = 0;
+static thread_local int g_firstSepY = -1;
 
 static bool AreWeMutating()
 {
@@ -953,6 +954,7 @@ static void RestoreTree(HWND hTree)
     void *pNscRaw = ts->pNscTree;
     unsigned long enumFlags = ts->enumFlags;
     IShellItemFilter *pFilterRaw = ts->pFilter;
+    ts->pFilter = nullptr;
     if (pFilterRaw)
         pFilterRaw->AddRef();
     HIMAGELIST savedImgList = ts->savedStateImageList;
@@ -1001,11 +1003,6 @@ static void RestoreTree(HWND hTree)
         if (ownsRef)
             ts->ownsNscRef = false;
         ts->pNscTree = nullptr;
-        if (ts->pFilter)
-        {
-            ts->pFilter->Release();
-            ts->pFilter = nullptr;
-        }
     }
 
     if (ownsRef)
@@ -1023,6 +1020,12 @@ static void FullRebuildTree(HWND hTree)
     ComRef<INameSpaceTreeControl> pNsc((INameSpaceTreeControl *)ts->pNscTree);
     void *pNscRaw = ts->pNscTree;
     unsigned long enumFlags = ts->enumFlags;
+
+    if (ts->pFilter)
+    {
+        ts->pFilter->Release();
+        ts->pFilter = nullptr;
+    }
 
     // Remove HomeSpacer before roots — it's a separate namespace item.
     RemoveHomeSpacer(*ts, pNsc.get());
@@ -1112,10 +1115,8 @@ static void InsertHomeSpacer(HWND hTree)
     HTREEITEM hSpacer = nullptr;
     WCHAR spacerNames[1][64] = {};
     wcsncpy_s(spacerNames[0], 64, g_navItems[spacerId].label, _TRUNCATE);
-    ForEachLabeledSibling(hTree,
-        [&](HTREEITEM h) { return h == ts->hItems[NAV_DESKTOP] || h == ts->hItems[NAV_THISPC]; },
-        spacerNames, 1,
-        [&](HTREEITEM h, int) -> bool { hSpacer = h; return false; });
+    ForEachLabeledSibling(hTree, [&](HTREEITEM h) { return h == ts->hItems[NAV_DESKTOP] || h == ts->hItems[NAV_THISPC]; },
+        spacerNames, 1, [&](HTREEITEM h, int) -> bool { hSpacer = h; return false; });
 
     if (hSpacer)
     {
@@ -1235,6 +1236,8 @@ static void DrawSeparatorLine(HDC hdc, HWND hTree, int sepY)
     {
         FillRect(hdc, &sepRect, brush);
         DeleteObject(brush);
+        if (g_firstSepY == -1)
+            g_firstSepY = sepY;
         g_lastSepY = sepY;
     }
 }
@@ -1481,15 +1484,11 @@ static bool CleanupDups(HWND hTree, TreeState& ts)
     int childlessCount = 0;
     int sectionCount = 0;
 
-    ForEachLabeledSibling(hTree,
-        [&](HTREEITEM h) { return IsOurSection(ts, h); },
-        matchNames, expectedCount,
+    ForEachLabeledSibling(hTree, [&](HTREEITEM h) { return IsOurSection(ts, h); }, matchNames, expectedCount,
         [&](HTREEITEM h, int) -> bool {
             HTREEITEM hChild = (HTREEITEM)SendMessageW(hTree, TVM_GETNEXTITEM, TVGN_CHILD, (LPARAM)h);
-            if (!hChild)
-            { if (childlessCount < 8) toDeleteChildless[childlessCount++] = h; }
-            else
-            { if (sectionCount < 4) toDeleteSections[sectionCount++] = h; }
+            if (!hChild) { if (childlessCount < 8) toDeleteChildless[childlessCount++] = h; }
+            else { if (sectionCount < 4) toDeleteSections[sectionCount++] = h; }
             return true;
         });
 
@@ -1558,12 +1557,8 @@ static bool RemoveInherited(HWND hTree, TreeState& ts)
     for (int i = 0; i < targetCount; i++)
         wcsncpy_s(targetNames[i], 64, targets[i].name, _TRUNCATE);
 
-    ForEachLabeledSibling(hTree,
-        [&](HTREEITEM h) {
-            return IsOurSection(ts, h) || h == ts.hiddenDuplicate || h == ts.homeSpacerItem;
-        },
-        targetNames, targetCount,
-        [&](HTREEITEM h, int j) -> bool {
+    ForEachLabeledSibling(hTree, [&](HTREEITEM h) { return IsOurSection(ts, h) || h == ts.hiddenDuplicate || h == ts.homeSpacerItem; },
+        targetNames, targetCount, [&](HTREEITEM h, int j) -> bool {
             toDelete[delCount++] = { h, targets[j].id };
             return delCount < targetCount;
         });
@@ -1916,6 +1911,7 @@ LRESULT CALLBACK SubClassTreeWndProc_hook(HWND hWnd, UINT uMsg, WPARAM wParam, L
             SetWindowSubclass(hWnd, TreeInteractionProc, 0xAF01, 0);
 
         g_lastSepY = -1;
+        g_firstSepY = -1;
         g_inTreePaint = true;
 
         LRESULT result = SubClassTreeWndProc_orig(hWnd, uMsg, wParam, lParam, uIdSubclass, dwRefData);
@@ -1923,7 +1919,7 @@ LRESULT CALLBACK SubClassTreeWndProc_hook(HWND hWnd, UINT uMsg, WPARAM wParam, L
         g_inTreePaint = false;
 
         ts = GetTree(hWnd);
-        if (g_lastSepY >= 0 && g_sepColor != CLR_INVALID && g_sepRetries < 3)
+        if (g_lastSepY >= 0 && g_sepColor != CLR_INVALID && ts && ts->sepRetries < 3)
         {
             HDC hdc = GetDC(hWnd);
             if (hdc)
@@ -1931,28 +1927,37 @@ LRESULT CALLBACK SubClassTreeWndProc_hook(HWND hWnd, UINT uMsg, WPARAM wParam, L
                 RECT client;
                 GetClientRect(hWnd, &client);
                 int sampleX = client.right / 2;
-                COLORREF pixel = GetPixel(hdc, sampleX, g_lastSepY);
-                ReleaseDC(hWnd, hdc);
-                if (pixel != CLR_INVALID && pixel != g_sepColor)
+                bool mismatch = false;
+                int failY = g_lastSepY;
+                int checkYs[] = { g_firstSepY, g_lastSepY };
+                for (int y : checkYs)
                 {
-                    g_sepRetries++;
+                    if (y < 0) continue;
+                    COLORREF pixel = GetPixel(hdc, sampleX, y);
+                    if (pixel != CLR_INVALID && pixel != g_sepColor)
+                        { mismatch = true; failY = y; break; }
+                }
+                ReleaseDC(hWnd, hdc);
+                if (mismatch)
+                {
+                    ts->sepRetries++;
                     Wh_Log(L"[SEP-VERIFY] tree=%04X at (%d,%d): "
-                           L"expected=0x%06X got=0x%06X, retry %d",
-                           PTR4(hWnd), sampleX, g_lastSepY, g_sepColor, pixel, g_sepRetries);
-                    InvalidateRect(hWnd, nullptr, FALSE);
+                           L"expected=0x%06X, retry %d",
+                           PTR4(hWnd), sampleX, failY, g_sepColor, ts->sepRetries);
+                    RedrawWindow(hWnd, nullptr, nullptr, RDW_INVALIDATE | RDW_UPDATENOW);
                 }
                 else
-                    g_sepRetries = 0;
+                    ts->sepRetries = 0;
             }
         }
-        else if (g_lastSepY == -1 && g_sepColor != CLR_INVALID && g_sepRetries < 3
-                 && ts && !ts->pendingWork && HasOurItemsInTree(*ts) && ts->boundaryItem
-                 && !g_settings.removeSepBelowNav && !GetHiddenItem(*ts))
+        else if (g_lastSepY == -1 && g_sepColor != CLR_INVALID
+                 && ts && ts->sepRetries < 3 && !ts->pendingWork && HasOurItemsInTree(*ts)
+                 && ts->boundaryItem && !g_settings.removeSepBelowNav && !GetHiddenItem(*ts))
         {
-            g_sepRetries++;
+            ts->sepRetries++;
             Wh_Log(L"[SEP-VERIFY] tree=%04X boundary=%04X: expected separator but none drawn, retry %d",
-                   PTR4(hWnd), PTR4(ts->boundaryItem), g_sepRetries);
-            InvalidateRect(hWnd, nullptr, FALSE);
+                   PTR4(hWnd), PTR4(ts->boundaryItem), ts->sepRetries);
+            RedrawWindow(hWnd, nullptr, nullptr, RDW_INVALIDATE | RDW_UPDATENOW);
         }
 
         return result;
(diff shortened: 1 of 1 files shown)


rodboev:
Not sure if I should submit a changelog. This is a significant addition and a bit of a reworking to get the mod to show all changes immediately. It also adds another 5 toggles. I bumped to 1.1 but will leave a changelog out for now and simply update the PR.

RajStartsHere:
@rodboev 
I Have A Question, Is This Mod Compatible With These Mods:

image


- https://windhawk.net/mods/explorer-treeitem-tweaker
- https://windhawk.net/mods/explorer-navigation-pane-tweaks
- https://windhawk.net/mods/explorer-treeline-killer

Nothing More, Thanks!

rodboev:
> @rodboev I Have A Question, Is This Mod Compatible With These Mods:

- **Explorer TreeItem Tweaker:** Yes.
- **Explorer Navigation Tree Offset:** Yes.
- **Explorer TreeLine Killer:** No. This mod has more granular control over separators to let you control each one individually, which conflicts with the all-or-nothing behavior of TreeLine Killer. Nothing breaks, it just kind of half-works. (The toggles under Separators achieve the same result.)

@m417z I rebased the commits to one-per-version so the deploy script doesn't error out.

rodboev:
I am thinking of adding an option to add the contents of any user folder (like %UserProfile%\Links), so a user can group shortcuts into folders and have those shortcuts and/or folders with shortcuts show instantly on the nav.

Basically categorized Quick Access entries by folder, without any custom work. I tried to do this for some time with filesystem backed virtual folder objects but it never worked. Windhawk makes it possible.

m417z:
Thanks for the submission. Let me know when the mod is ready to be reviewed and merged.

rodboev:
> Thanks for the submission. Let me know when the mod is ready to be reviewed and merged.

It's ready. 😁

m417z:
### Submission review

Also:

> This mod injects only in processes that use `ExplorerFrame.dll`, so the include is set to `*` but it will not touch most processes.

Which other processes load `ExplorerFrame.dll` on init? The combination of including `*` but limiting to x86-64 is unusual too. I think it's better to either limit to explorer.exe, or support all apps and late-loading of `ExplorerFrame.dll`, but not this confusing combination.

---

Nice work — this fills a real gap (an expandable This PC / Desktop root pinned to the very top isn't possible without it), and the settings block is clean and fully in sync with the code. The concerns below are mostly about the multi-window / teardown threading model.

**1. Teardown removes the tree subclass — and drives COM — from the wrong thread.**
`Wh_ModUninit`, and the `TIER_REBUILD` / `TIER_REFRESH` branches of `Wh_ModSettingsChanged`, run on Windhawk's engine thread, **not** the Explorer UI thread that owns each `SysTreeView32`.

- `RestoreTree()` removes the auxiliary `TreeInteractionProc` subclass with raw `RemoveWindowSubclass(hTree, TreeInteractionProc, 0xAF01)` ([line 992](https://github.com/ramensoftware/windhawk-mods/blob/85684455fbfab1c1246ffa9e84727b923d9ce130/mods/add-virtual-folders-to-nav-top.wh.cpp#L992)). `RemoveWindowSubclass` only works when called on the window's owning thread, so from the engine thread it silently fails. The subclass then **survives the unload**, and the next message the tree receives jumps into a function pointer in the freed mod DLL → crash. You already do this correctly for the *parent* subclass at [line 2494](https://github.com/ramensoftware/windhawk-mods/blob/85684455fbfab1c1246ffa9e84727b923d9ce130/mods/add-virtual-folders-to-nav-top.wh.cpp#L2494) (`WindhawkUtils::RemoveWindowSubclassFromAnyThread`) — use the same here (and install `TreeInteractionProc` via `SetWindowSubclassFromAnyThread` for symmetry). This is a deadlock/crash on unload, so it's not optional.
- The same paths call `INameSpaceTreeControl::AppendRoot` / `RemoveRoot` / `Release` directly on objects that live on the UI thread. The `SendMessage`-based tree edits marshal fine, but the raw COM vtable calls touch UI/shell state cross-thread. You already have the right vehicle — `WM_DEFERRED_REBUILD` is posted to the tree and handled on its own thread — so the settings/uninit paths should post to it rather than calling `FullRebuildTree` / `RestoreTree` inline on the engine thread. See https://github.com/ramensoftware/windhawk/wiki/Development-tips.

**2. Shared global state is touched from multiple Explorer UI threads without synchronization.**
Each File Explorer window runs on its own thread, and your hooks (`SubClassTreeWndProc_hook`, `SepParentSubclassProc`) run on that window's thread. The globals `g_trees` (`std::unordered_map`), `g_subclassedParents`, `g_pendingRebuildTrees`, plus the plain-`static` guards `g_deferredOpInProgress` / `g_mutatingTree` / `g_sepColor`, are read and written from those threads (and from the engine thread on init/settings/uninit) with no locking. With two Explorer windows open, a `g_trees.try_emplace(...)` on one thread can rehash the map while another thread is inside `GetTree()` → use-after-free / crash. The comment at [line 330](https://github.com/ramensoftware/windhawk-mods/blob/85684455fbfab1c1246ffa9e84727b923d9ce130/mods/add-virtual-folders-to-nav-top.wh.cpp#L330) ("hooks fire on the UI thread", singular) reflects the wrong model — there are multiple UI threads. Guard the shared containers with a mutex (copy handles → release lock → do any cross-thread `SendMessage`, to avoid the lock-and-SendMessage deadlock), or move per-tree state into per-window storage. This is the single most common correctness theme in recent reviews; the recurring domain fact maintainers cite is *"each Explorer window runs on its own thread."*

**3. Modern file dialogs in other apps may silently not be hooked (late `ExplorerFrame.dll` load).**
`Wh_ModInit` returns `FALSE` when `ExplorerFrame.dll` isn't loaded yet ([line 2087](https://github.com/ramensoftware/windhawk-mods/blob/85684455fbfab1c1246ffa9e84727b923d9ce130/mods/add-virtual-folders-to-nav-top.wh.cpp#L2087)). With `@include *` the mod is injected at process start; a process that only loads ExplorerFrame *later* (the first time it shows a modern Open/Save dialog) already returned `FALSE` and won't be re-initialized except on a settings change, so its nav pane is left untouched. The README states *"Only 'modern' dialogs are affected,"* which implies they're handled. Either hook `LoadLibraryExW` and apply the hooks when ExplorerFrame loads (the standard late-load pattern), or soften the README to note dialogs are only affected in processes that already had ExplorerFrame loaded. (For `explorer.exe` itself the DLL is always present, so the core feature is fine — and gating the broad `@include *` on the DLL check is reasonable.)

**4. Partial overlap with an existing mod (informational).**
The core feature is genuinely new, but the bundled **Hide Home/Gallery** toggle overlaps [hide-home-gallery-explorer](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/hide-home-gallery-explorer.wh.cpp). It's a minor sub-feature here and your suppression approach (intercepting `AppendRoot`) is cleaner than that mod's periodic text scan, so I don't think it's a blocker — just flagging it so it's a conscious decision rather than an accidental duplication.

<p>

- **Use the type-safe hook helper for `DrawThemeBackground`.** [Line 2171](https://github.com/ramensoftware/windhawk-mods/blob/85684455fbfab1c1246ffa9e84727b923d9ce130/mods/add-virtual-folders-to-nav-top.wh.cpp#L2171) uses raw `Wh_SetFunctionHook((void*)pDTB, ...)` with `void*` casts and doesn't check the return value. `WindhawkUtils::SetFunctionHook(pDTB, DrawThemeBackground_hook, &DrawThemeBackground_orig)` is the idiomatic, type-checked form.
- **Remove the window props on teardown.** `SetPropW(hWnd, L"WH_NscTree"/L"WH_EnumFlags", ...)` is never matched by `RemovePropW` on unload/`WM_NCDESTROY`, so stale props linger on live windows after the mod is disabled. Harmless, but tidy them up where you tear the tree down.
- **Gate GDI+ startup on the feature.** `GdiplusStartup` runs unconditionally in `Wh_ModInit` even when `fixChevronDrawing` is off (the only GDI+ user is `DrawChevron`). Initialize it lazily / only when the setting is enabled.
- **Edge-case COM ref leak on unload.** If `RestoreTree` early-returns at [line 967](https://github.com/ramensoftware/windhawk-mods/blob/85684455fbfab1c1246ffa9e84727b923d9ce130/mods/add-virtual-folders-to-nav-top.wh.cpp#L967) (`roots.Create()` fails), an owned `pNscTree` ref is never released, and the straggler loop at [line 2499](https://github.com/ramensoftware/windhawk-mods/blob/85684455fbfab1c1246ffa9e84727b923d9ce130/mods/add-virtual-folders-to-nav-top.wh.cpp#L2499) only releases `pFilter`, not an owned `pNscTree`. Only matters on unload.
- **`TreeInteractionProc` largely duplicates logic already in `SubClassTreeWndProc_hook`** (both swallow `WM_RBUTTONDOWN`/hit-testing on the hidden duplicate). Consider folding it into the main hook to drop one subclass entirely (which also sidesteps item #1's removal path).

</p>
</details>

<p>

- **Pixel-readback separator verification is fragile.** After painting, the `WM_PAINT` handler `GetPixel`s the separator row and, on mismatch, calls `RedrawWindow(..., RDW_UPDATENOW)` up to 3× ([lines 1922–1961](https://github.com/ramensoftware/windhawk-mods/blob/85684455fbfab1c1246ffa9e84727b923d9ce130/mods/add-virtual-folders-to-nav-top.wh.cpp#L1922)), forcing synchronous re-paints. Anything drawn over the sample point (selection highlight, hover) reads as a "mismatch" and triggers needless redraws. It's bounded (3 retries) and the nav pane is small, so it's likely fine in practice — noting it as an FYI in case a more deterministic redraw trigger is available.
- **Per-paint full tree walks.** `FindSectionLayout`, `GetBaseItemHeight`, and `RedrawSeps` each walk the whole depth-1 list (several `SendMessage`s per item) on every `WM_PAINT`. Bounded by the nav pane size, but it's a fair bit of work per paint.
- **The suppress-in-`CDDS_PREPAINT` / redraw-in-`CDDS_POSTPAINT` + display-name matching approach is inherently version-sensitive** to ExplorerFrame internals. Worth noting that the localized-label matching itself is handled correctly — labels are derived from the PIDLs at runtime (`GetPidlDisplayName`), so they match the shell's current UI language rather than being hardcoded English. The fragility is the paint-stage interception, not the locale handling.

</p>
</details>

rodboev:
Thanks for the thorough review. This is a good 80/20 split between things that were genuinely missed and others that are battle-tested and easily explained.

> I think it's better to either limit to explorer.exe, or support all apps and late-loading of `ExplorerFrame.dll`, but not this confusing combination.

This was a regression. The mod works on enable/disable but exits early for processes that don't have ExplorerFrame.dll when they start and doesn't re-enter. The correct behavior is in place now with support for late-loading of ExplorerFrame.dll.

> The combination of including * but limiting to x86-64 is unusual too.

We need @include * for file dialogs in other apps, and only have a 64-bit build. Explorer is always 64-bit but other apps can be both.

> **1. Teardown removes the tree subclass — and drives COM — from the wrong thread.**

This oversight was introduced with the code used to suppress the context menu on synthetic separators. Instead of using the raw Win32 API the code now uses the cross-thread variant that was already in use elsewhere.

Wh_ModUninit now restores each tree with a WM_RESTORE_TREE message, so all COM calls and tree mutations execute on that tree's UI thread. (Orphaned trees whose window is gone still fall back to a best-effort release.)

> **2. Shared global state is touched from multiple Explorer UI threads without synchronization.**

The three insertion globals (g_ins, g_inCustomAppend, g_spacerInsertAfter) are now thread_local, as are the re-entrancy guards and pending-rebuild set. Multiple Explorer windows can live on different UI threads, and without per-thread copies two concurrent AppendRoot/FullRebuildTree calls overwrite each other's COM pointers.

The shared tree registry (g_trees) and parent-subclass tracking set (g_subclassedParents) are protected by a recursive mutex, released before any SendMessage or COM call to avoid deadlock.

Settings changes are marshaled from the engine thread to each tree's UI thread via SendMessage, so all COM calls and tree mutations execute on the correct STA thread.

> **3. Modern file dialogs in other apps may silently not be hooked (late `ExplorerFrame.dll` load).**

This was due to the early exit above. A LoadLibraryExW hook now detects when ExplorerFrame.dll loads in any process, installs and applies the ExplorerFrame hooks, then runs discovery. The detection fires once per process and is logged when it happens, so it's traceable without adding ongoing noise.

> **4. Partial overlap with an existing mod (informational).**

The main problem with the Hide Home/Gallery mod is that it is indiscriminate: it hides every text match including user-pinned items. This is different than what these toggles do, which is to clear the top nav only in order to make room for the new items we add. The Hide Home/Gallery toggles here can be used to do that while keeping those pinned items in Quick Access, providing flexibility for user preferences as to where Home and Gallery ultimately go and how they are grouped.

> **5. Engine-thread COM calls.**

Covered by the same SendMessage marshaling described in Finding 2. Wh_ModSettingsChanged no longer calls FullRebuildTree or RefreshNavPane directly from the engine thread; each tree receives a WM_SETTINGS_CHANGED message and dispatches its own tier on the owning UI thread.

* * *

Optional improvements:

> **Use the type-safe hook helper for `DrawThemeBackground`.**

Done. Uses `WindhawkUtils::SetFunctionHook` instead of void* casts.

> **Remove the window props on teardown.**

Done. `RemovePropW` for `WH_NscTree` and `WH_EnumFlags` in both `RestoreTree` and the `WM_NCDESTROY` handler.

> **Gate GDI+ startup on `fixChevronDrawing`.**

Skipped. The setting can change at runtime, so lazy init would need paired teardown and re-init logic. `GdiplusStartup` is called once at load and takes ~1ms; the complexity isn't worth it.

> **COM ref leak on `RestoreTree` early return.**

Done. The straggler loop in Wh_ModUninit now releases both pFilter and pNscTree for trees that RestoreTree didn't fully process.

> **Merge `TreeInteractionProc` into `SubClassTreeWndProc_hook`.**

Skipped for now. The two serve different roles in the dispatch chain: one is the inline hook on Explorer's own subclass, the other is an auxiliary Win32 subclass installed for interaction blocking. Merging them requires mapping exactly which messages reach which callback and in what order. The FromAnyThread fix is simpler and sufficient.

* * *

Functionality notes:

> **Pixel readback / SEP-VERIFY**

This proved necessary across many testing cycles. Explorer appears to sometimes paint over our custom separators within the same WM_PAINT cycle. Without the verification step, separators disappear until the next natural repaint (on hover or some other trigger.) The 3-retry cap prevents loops, but never hits more than 2 in practice, and a false positive only costs one repaint of a small area.

> **Per-paint tree walks**

The nav pane has at most ~15 depth-1 items, so the walks are bounded and fast. Caching would require invalidation logic for every tree mutation, adding complexity for no measurable gain.

> **Paint-stage interception fragility**

We have to swallow all prepaints to remove the separator lines between items in our section that Explorer normally draws. There is no way to suppress some and not others.

m417z:
If `Wh_ModInit` just hooks `LoadLibraryExW_hook`, `Wh_ModAfterInit` still enumerates all system windows, which will unnecessarily delay process launch. Skip it in this case.

In fact, even for explorer.exe which loads `ExplorerFrame.dll` statically, enumerating all windows on launch is a waste since the process has no windows yet. It can be skipped with this check:
https://github.com/ramensoftware/windhawk/wiki/Development-tips#avoid-mod-loading-on-process-startup

> We need `@include *` for file dialogs in other apps, and only have a 64-bit build. Explorer is always 64-bit but other apps can be both.

But then you'll customize file dialogs in other apps only if they're 64-bit, not if they're 32-bit. Perhaps I'm missing something?

Below is an updated Claude review. Some items might be repeated or incorrect. No need to comment on each, just address the actual issues if there are any.

Thanks.

---

Two residuals remain from that round, plus one small leak:

**1. Unload still has a narrow cross-thread use-after-free window.** The pervasive pattern is `TreeState* ts = GetTree(hWnd)` (which locks, returns a raw pointer, then unlocks at [line 249](https://github.com/ramensoftware/windhawk-mods/blob/58ed9cbb6b1f4a464eb8dcda76971ac5ae3901d3/mods/add-virtual-folders-to-nav-top.wh.cpp#L249)) followed by dereferencing `ts` with the lock no longer held — e.g. the `WM_PAINT` handler at [lines 1955–1959](https://github.com/ramensoftware/windhawk-mods/blob/58ed9cbb6b1f4a464eb8dcda76971ac5ae3901d3/mods/add-virtual-folders-to-nav-top.wh.cpp#L1955) reads `ts->pendingWork` unconditionally. During normal operation this is safe (each entry is only erased by its own thread's `WM_NCDESTROY`), but `Wh_ModUninit` calls `g_trees.clear()` on the engine thread ([line 2636](https://github.com/ramensoftware/windhawk-mods/blob/58ed9cbb6b1f4a464eb8dcda76971ac5ae3901d3/mods/add-virtual-folders-to-nav-top.wh.cpp#L2636)) while the mod's subclass hooks are still installed and *other* Explorer UI threads can still paint. With ≥2 windows on different threads, thread B can do `GetTree`→unlock→(engine thread runs `clear()`, freeing the node)→thread B dereferences the freed `TreeState*` → crash. It's narrow, but it's an unload-stability defect, so worth closing. Simplest fix: in `Wh_ModUninit`, after the `WM_RESTORE_TREE` round-trips, don't `clear()` the map — leave the (now-restored, COM-released) entries in place (a small one-time leak of empty `TreeState`s is harmless and the process is being detached from the mod anyway), or hold `g_treesMutex` across the whole `ts`-using body of `WM_PAINT` rather than just the `GetTree` call.

**2. `g_sepColor` / `g_logSepDraw` are still unsynchronized globals.** Finding #2 last round listed `g_sepColor` among the cross-thread state, but the thread_local/mutex changes didn't cover it — `g_sepColor` (written in `SepParentSubclassProc` at [line 1437](https://github.com/ramensoftware/windhawk-mods/blob/58ed9cbb6b1f4a464eb8dcda76971ac5ae3901d3/mods/add-virtual-folders-to-nav-top.wh.cpp#L1437), read in `RedrawSeps`) and `g_logSepDraw` are still plain `static` globals touched from every Explorer UI thread's paint path. This one is *benign* (a single aligned 32-bit write; worst case a separator paints one frame with another window's derived color, or a stray log line), so I've left it as the lower-priority of the two — but since it was explicitly named last round and wasn't addressed, flagging that it's still open. If you want it fully clean, make `g_sepColor` an `std::atomic<COLORREF>` (or derive it per-tree).

<p>

- **Per-thread `g_ins.filter` ref leak.** `AppendRoot_hook` caches the last filter into the `thread_local g_ins.filter` with an `AddRef` ([lines 571–576](https://github.com/ramensoftware/windhawk-mods/blob/58ed9cbb6b1f4a464eb8dcda76971ac5ae3901d3/mods/add-virtual-folders-to-nav-top.wh.cpp#L571)) but nothing ever `Release`s it — one COM ref is leaked per UI thread that ran an insert, and never reclaimed on unload. The per-tree `ts.pFilter` already holds the ref you actually use, so `g_ins.filter` looks redundant; if you keep it, release it in `Wh_ModUninit`.
- **Trim the `[TAG]` debug logging.** There's a large volume of `Wh_Log(L"[HOOK]…")` / `[APPEND]` / `[SEP-DRAW]` / `[DUP]` / `[CACHE]` etc. tracing. It's free at runtime (logging is off by default), so this is purely cosmetic, but the category-tag scaffolding reads as development debugging — consider thinning it before merge.

</p>
</details>

<p>

Non-critical observations about the feature behavior itself — these are FYI, and you already addressed most of them in the prior round.

- **Overall complexity.** At ~2.6k lines this is roughly 7× the closest comparable mod ([favorites-in-navpane](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/favorites-in-navpane.wh.cpp), ~390 lines, also an `AppendRoot`-hook nav-pane mod). A lot of that is the separator/spacer machinery and the rebuild-retry/dedup state. It clearly works, and you've justified the individual pieces — just noting that the surface area is large for a maintainer to reason about, and that future ExplorerFrame changes will land on a lot of paint-stage code.
- **Pixel-readback separator verification** ([WM_PAINT, lines 2038–2077](https://github.com/ramensoftware/windhawk-mods/blob/58ed9cbb6b1f4a464eb8dcda76971ac5ae3901d3/mods/add-virtual-folders-to-nav-top.wh.cpp#L2038)) and the suppress-all-in-`CDDS_PREPAINT` / redraw-in-`CDDS_POSTPAINT` approach remain inherently version- and theme-sensitive. You've explained why they're needed (Explorer overpaints within the same cycle; there's no way to suppress only some native separators), and the 3-retry cap bounds the cost — so this is an accepted trade-off, not a change request. Recorded here only so it's a conscious choice.
- **Hide Home/Gallery overlap** with [hide-home-gallery-explorer](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/hide-home-gallery-explorer.wh.cpp) — you made the case that your `AppendRoot`-suppression is more precise (doesn't touch pinned items) and serves the "make room at top" purpose; agreed it's not a duplication blocker.

</p>
</details>

rodboev:
> If `Wh_ModInit` just hooks `LoadLibraryExW_hook`, `Wh_ModAfterInit` still enumerates all system windows

Should be fixed now: `Wh_ModAfterInit` checks bails early if it's running on the process's initial thread now, so that no enumeration takes place.

> But then you'll customize file dialogs in other apps only if they're 64-bit, not if they're 32-bit.

At the moment 32-bit apps load 32-bit ExplorerFrame.dll which we don't hook. I should have been more clear that there is no 32-bit build at this time. It can be a future addition.

> 1. Unload still has a narrow cross-thread use-after-free window.

Got it. `Wh_ModUninit` no longer calls `g_trees.clear()`. The per-tree state is inert after restoration and harmless to leave behind until DLL unload.

> 2. `g_sepColor` / `g_logSepDraw` are still unsynchronized globals.

Yes, the repaint is benign even if we use another window's derived color since it is the same.

> Per-thread `g_ins.filter` ref leak.

Fixed by having the `WM_RESTORE_TREE` handler release `g_ins.filter` on the UI thread.

> The suppress-all-in-CDDS_PREPAINT / redraw-in-CDDS_POSTPAINT approach remain inherently version- and theme-sensitive.

We handle both light and dark themes, and before using color deltas, were sampling from Explorer-drawn separators. But it turns out the separators are not a theme resource, so the deltas are used instead now to save a repaint.

m417z:
> At the moment 32-bit apps load 32-bit ExplorerFrame.dll which we don't hook. I should have been more clear that there is no 32-bit build at this time. It can be a future addition.

I see. Is there a reason for that? I'd assume it's as simple as adding `ifdef` for the two symbols. Is anything else needed?

**Edit**: Just in case you're not familiar with it, you can use Windhawk Symbol Helper to easily show the symbols of a module.

rodboev:
There was a bit more to it than that, but it's in place now.

Also, I wasn't catching an issue earlier with cold loads, since most of my testing was in Explorer which is always running. As a result I had to use the same method as `favorites-in-navpane` and `classic-theme-explorer-search-fix` which load `ExplorerFrame.dll` in `Wh_ModInit` instead of hooking `LoadLibraryExW` to detect when the system loads it.

m417z:
Thanks. Looks good for merging. Further improvements can be made in future updates.

> As a result I had to [...] load `ExplorerFrame.dll` in `Wh_ModInit`

What issues did you have with the `LoadLibraryExW` hook method?

> `Wh_ModUninit` no longer calls `g_trees.clear()`.

It was restored in the last two commits. Not sure if it's a mistake.

rodboev:
> What issues did you have with the `LoadLibraryExW` hook method?

Explorer works fine since it's loaded all the time, but other ExplorerFrame.dll would not always load for other processes. Specifically, if the mod enabled before the process opens a file dialog, `LoadLibraryExW` hook never fired for the file dialog's `ExplorerFrame.dll` load. Loading it directly in `Wh_ModInit` via `LoadLibraryExW(..., LOAD_LIBRARY_SEARCH_SYSTEM32)` ensures hooks are installed before the dialog opens.

Without it, the mod would have to be enabled with the dialog open or ExplorerFrame.dll would not be detected. The hook is still the fallback if the direct load fails. If you had ideas for this I can try another route.

> > `Wh_ModUninit` no longer calls `g_trees.clear()`.
> 
> It was restored in the last two commits. Not sure if it's a mistake.

It's was inadvertently removed in "Skip startup enumeration..." and restored in "Support 32-bit processes...". `Wh_ModUninit` iterates and releases refs from `g_trees`, then `clear()` drops the stale entries so no dangling pointers remain. Removing it doesn't cause immediate issues that I can spot but my understand is that it's a defensive cleanup before DLL unload, not fixing an observable bug. 

The `std::unordered_map` destructor would do the same thing. The closest example in another mod is `explorer-navigation-pane-tweaks` at line 1454: `g_ExplorerStates.clear(); g_TreeStates.clear();`. This is the same kind of per-tree state map, same uninit context. Do you want it removed?

m417z:
> The hook is still the fallback if the direct load fails. If you had ideas for this I can try another route.

I looked at the code again:
```cpp
WindhawkUtils::SetFunctionHook(LoadLibraryExW, LoadLibraryExW_hook, &LoadLibraryExW_orig);
```
and the issue is likely that by specifying just `LoadLibraryExW`, you're hooking the function from kernel32.dll, but internal functions often call `LoadLibraryExW` from kernelbase.dll directly. `LoadLibraryExW` from kernel32 just calls the kernelbase function, so hooking it in kernelbase catches both, but not vice versa.

Example:
https://github.com/ramensoftware/windhawk-mods/blob/5183328f1bd44e572e823ad342b948a2025c253e/mods/taskbar-tray-system-icon-tweaks.wh.cpp#L1722-L1728

> Removing it doesn't cause immediate issues that I can spot but my understand is that it's a defensive cleanup before DLL unload

I looked in more detail and it was a mistake, it's fine to keep it.

BTW some users in Discord reported stability issues with the mod. I saw there's also one issue that was just created, but if you're interested, you can join [the Windhawk Discord server](https://discord.com/servers/windhawk-923944342991818753) for more feedback.

rodboev:
Thanks for the direction, I had trouble getting late-loading to work and it seems the last batch of changes caused that bigger regression.

The approach you recommended is now in place in #4282 and I'll try to get on Discord shortly.

PR #4167, opened 2026-05-24
Asteski's changes added

ExiledEye:
> Hello,
> 
> I've added additional processes to block:
> 
> ShellHost.exe -> The Action Center UI and process ("WIN + A" shortcut).
> ShellExperienceHost.exe -> The Calendar and Notifications UI and process ("WIN + N" key).
> MicrosoftStartFeedProvider.exe -> The Start Menu process, which provides news, weather, stock quotes, interest cards, recommended and etc.
> WidgetBoard.exe -> The Widgets board UI and process ("WIN + W" shortcut).
> WidgetService.exe -> The Widgets board background data, content and service infrastructure process.
> Additionally I've also updated description for some of the existing processes, specifically hotkey combinations with which they're triggered.
> 
> All of them were tested and are blocked successfully.


mods/block-windows-startmenu-and-hosts.wh.cpp
```diff
@@ -93,6 +128,33 @@ void TerminateConfiguredProcesses() {
...
+    if (g_blockWidgetService) {
+        TerminateProcessByName(L"WidgetService.exe");
+    }
+}
+
+// Background thread: watches for ShellHost.exe and kills it when the setting is on
+DWORD WINAPI ShellHostWatcherThread(LPVOID) {
+    while (WaitForSingleObject(g_watcherStopEvent, 1500) == WAIT_TIMEOUT) {
+        if (g_blockShellHost) {
+            TerminateProcessByName(L"ShellHost.exe");
```
m417z: Instead of running process enumeration all the time (which is quite slow and resource-intentsive) a better solution is for the mod to target `ShellHost.exe`, and, when starting as part of `ShellHost.exe`, just exit the process if blocked in the settings.

ExiledEye: You are right, but for some reason using the logic used to block the other host processes doesn't work for `ShellHost.exe` like it does for the others.
But ultimately the watcher thread is a last resort method, I will try different things: if no other method works, I will add a note to warn that `ShellHost.exe` blocking may be more resource demanding compared to the others (which actually saves resources by being blocked)

m417z: I don't understand why you'd want to keep the watcher thread, but I'll wait for your update.

ExiledEye: Removed the watcher thread, targeted shellhost.exe directly and it works!

mods/block-windows-startmenu-and-hosts.wh.cpp
```diff
@@ -157,12 +194,23 @@ void Wh_ModSettingsChanged() {
...
-    // 1. Load initial settings and kill running configured processes
     LoadSettings();
+
+    // ShellHost respawn block
+    if (_wcsicmp(GetCurrentProcessName(), L"ShellHost.exe") == 0) {
+        if (g_blockShellHost) {
+            Wh_Log(L"ShellHost.exe is blocked, exiting process.");
+            ExitProcess(0);
+        }
+        return TRUE;
```
m417z: The mod doesn't have to be loaded in this case, it's better to return `FALSE` here.

PR #4233, opened 2026-05-30
Add IME Native Mode Lock mod

ZeonXr:
# IME Native Mode Lock

Forces supported IMEs to stay in their native input mode.

By default, Windows automatically switches input modes when changing
window focus or switching between applications. This mod blocks that
behavior, keeping your IME in its native mode at all times.

## Supported IMEs
- **Simplified Chinese** (Microsoft Pinyin, etc.) — always stays in Chinese mode
- More languages can be added in future updates

**Pair with** Caps IME Switcher (`caps-ime-switcher`) — short-press
Caps Lock to switch input languages, long-press for Caps Lock.

## Changelog

* Initial release

## Mod authorship

This mod was created by:

- - [x] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/ime-mode-lock.wh.cpp
+++ b/mods/ime-mode-lock.wh.cpp
@@ -14,7 +14,8 @@
 /*
 # IME Native Mode Lock
 
-Forces supported IMEs to stay in their native input mode.
+Forces supported IMEs to stay in their native input mode, blocking
+both system-initiated and user-initiated mode switches.
 
 By default, Windows automatically switches input modes when changing
 window focus or switching between applications. This mod blocks that
@@ -29,7 +30,7 @@ Caps Lock to switch input languages, long-press for Caps Lock.
 
 ---
 
-强制支持的输入法保持在原生输入模式。
+强制支持的输入法保持在原生输入模式，同时阻止系统自动切换和用户主动切换。
 
 默认情况下，Windows 在切换窗口焦点或应用程序时会自动切换输入模式。
 此 mod 阻止该行为，始终保持输入法在原生输入模式。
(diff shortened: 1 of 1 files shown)


mods/ime-mode-lock.wh.cpp
```diff
@@ -0,0 +1,285 @@
...
+
+BOOL Wh_ModInit() {
+    DWORD sessionId;
+    if (ProcessIdToSessionId(GetCurrentProcessId(), &sessionId) &&
+        sessionId == 0) {
+        return FALSE;
+    }
+
+    LoadSettings();
+    return InstallHooks();
```
m417z, line 275: If `AnyProfileEnabled` returns false, it makes sense to return `FALSE` here to have the mod unload. It will be loaded again once settings are changed. Then it will also be unnecessary to reload on any `Wh_ModSettingsChanged` call.

Also, since `AnyProfileEnabled` is checked on mod loaded, there's no need to check it in other places.

mods/ime-mode-lock.wh.cpp
```diff
@@ -0,0 +1,285 @@
...
+By default, Windows automatically switches input modes when changing
+window focus or switching between applications. This mod blocks that
+behavior, keeping your IME in its native mode at all times.
+
+## Supported IMEs
+- **Simplified Chinese** (Microsoft Pinyin, etc.) — always stays in Chinese mode
+- More languages can be added in future updates
+
+**Pair with** Caps IME Switcher (`caps-ime-switcher`) — short-press
+Caps Lock to switch input languages, long-press for Caps Lock.
```
m417z, line 29: With the current version, what's the user-facing difference between this mod and [Chinese IME fixed mode](https://windhawk.net/mods/chinese-ime-fixed-mode)? Please add information about it to the readme.

ZeonXr: The key user-facing difference is that **Chinese IME fixed mode** only catches the IME's internal "auto fall-back" private message, so manual mode switching by the user still works. **IME Native Mode Lock** also intercepts `IMN_SETCONVERSIONMODE`, which fires on any mode change including user-initiated ones — so the IME is fully locked and the user can no longer switch out of it.

It's also more extensible: the profile-table design makes it easy to add support for other IMEs (Japanese, Korean) in future updates.

PR #4235, opened 2026-05-30
Add Tray hover expand mod for hidden icons

wygodad:
This mod allows the hidden tray icons flyout to open on hover instead of requiring a click. It includes settings for auto-closing, polling interval, collapse delay, and hit area padding.
It works through UI Automation (no shell function hooks), so it is relatively safe and resilient across Windows builds.
Known limitation: auto-collapse relies on detecting the flyout window class (TopLevelWindowForOverflowXamlIsland by default), which is exposed as a configurable setting in case it differs across Windows builds.
Tested on: Windows 11 25H2 (build 26200.8524)

## Changelog

N/A — this pull request introduces a new mod.
 
## Mod authorship

This mod was created by:

- - [x] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/tray-hover-expand.wh.cpp
+++ b/mods/tray-hover-expand.wh.cpp
@@ -1,0 +1,303 @@
+// ==WindhawkMod==
+// @id              tray-hover-expand
+// @name            Tray hover expand
+// @description     Open the hidden tray icons flyout on hover instead of clicking the chevron; optionally collapse it when the cursor leaves
+// @version         1.0.0
+// @author          wygodad
+// @github          https://github.com/wygodad
+// @include         explorer.exe
+// @compilerOptions -lole32 -loleaut32 -luuid -luiautomationcore
+// @license         MIT
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Tray hover expand
+
+Opens the hidden tray icons flyout (the "Show Hidden Icons" chevron) when you
+hover the cursor over it, instead of having to click. Optionally collapses it
+again once the cursor leaves the opened icons.
+
+It works through UI Automation, so it does not hook internal shell functions —
+it is relatively safe and resilient across Windows builds.
+
+## Notes
+- If the chevron button is not detected, the matching is name-based. The default
+  keywords cover English and Polish; other languages may need the button's name
+  added to the `keywords` list in the source.
+- If auto-collapse does not work, the flyout window class name may differ on your
+  build. Change it in the "Flyout window class" setting.
+*/
+// ==/WindhawkModReadme==
+
+// ==WindhawkModSettings==
+/*
+- autoClose: true
+  $name: Collapse when the cursor leaves
+  $description: After the cursor leaves the opened icons (and the chevron), the flyout closes itself.
+- pollInterval: 50
+  $name: Polling interval (ms)
+  $description: How often to check the cursor position. Lower = smoother, more CPU.
+- grace: 200
+  $name: Collapse delay (ms)
+  $description: How long the cursor must stay outside the area before the flyout closes (prevents flicker).
+- pad: 4
+  $name: Hit area padding (pixels)
+  $description: Enlarges the hover area around the chevron button.
+- flyoutClass: TopLevelWindowForOverflowXamlIsland
+  $name: Flyout window class
+  $description: Window class name of the opened flyout (used to tell whether the cursor is over the icons). Change it if auto-collapse does not work.
+*/
+// ==/WindhawkModSettings==
+
+#include <windows.h>
+#include <uiautomation.h>
+#include <string>
+#include <vector>
+#include <algorithm>
+
+// GUID-y na wszelki wypadek (gdyby linker MinGW ich nie dostarczył).
+#ifndef __IUIAutomation_FWD_DEFINED__
+#error "Brak nagłówków UI Automation"
+#endif
+
+struct Settings {
+    bool autoClose = true;
+    int pollInterval = 50;
+    int grace = 200;
+    int pad = 4;
+    std::wstring flyoutClass = L"TopLevelWindowForOverflowXamlIsland";
+    std::vector<std::wstring> keywords = {
+        L"ukryte ikony", L"hidden icons", L"rozwiń", L"overflow"
+    };
+};
+
+static Settings g_settings;
+static volatile bool g_running = false;
+static HANDLE g_thread = nullptr;
+
+static std::wstring ToLower(const std::wstring& s) {
+    std::wstring r = s;
+    std::transform(r.begin(), r.end(), r.begin(), ::towlower);
+    return r;
+}
+
+static bool NameMatches(const std::wstring& name) {
+    std::wstring low = ToLower(name);
+    for (const auto& k : g_settings.keywords) {
+        if (!k.empty() && low.find(ToLower(k)) != std::wstring::npos) return true;
+    }
+    return false;
+}
+
+// ---- Pomocnicze: pobranie wzorców UIA ----
+
+static ExpandCollapseState GetState(IUIAutomationElement* e) {
+    IUIAutomationExpandCollapsePattern* p = nullptr;
+    if (SUCCEEDED(e->GetCurrentPatternAs(
+            UIA_ExpandCollapsePatternId,
+            __uuidof(IUIAutomationExpandCollapsePattern), (void**)&p)) && p) {
+        ExpandCollapseState s = ExpandCollapseState_Collapsed;
+        p->get_CurrentExpandCollapseState(&s);
+        p->Release();
+        return s;
+    }
+    // Fallback: stan po widoczności okna flyoutu.
+    HWND f = FindWindowW(g_settings.flyoutClass.c_str(), nullptr);
+    return (f && IsWindowVisible(f)) ? ExpandCollapseState_Expanded
+                                     : ExpandCollapseState_Collapsed;
+}
+
+static void DoExpand(IUIAutomationElement* e) {
+    IUIAutomationExpandCollapsePattern* p = nullptr;
+    if (SUCCEEDED(e->GetCurrentPatternAs(
+            UIA_ExpandCollapsePatternId,
+            __uuidof(IUIAutomationExpandCollapsePattern), (void**)&p)) && p) {
+        p->Expand();
+        p->Release();
+        return;
+    }
+    IUIAutomationInvokePattern* inv = nullptr;
+    if (SUCCEEDED(e->GetCurrentPatternAs(
+            UIA_InvokePatternId,
+            __uuidof(IUIAutomationInvokePattern), (void**)&inv)) && inv) {
+        inv->Invoke();
+        inv->Release();
+    }
+}
+
+static void DoCollapse(IUIAutomationElement* e) {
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

2 will become less of an issue once 1 is implemented and the mod becomes isolated.
3 is not a must but would be nice.

---

Nice, self-contained UIA approach — it's reversible and doesn't hook the shell. A few integration issues to address before merge.

**1. This should be a "tool mod", not injected into `explorer.exe`.** The mod has no function hooks at all — it only uses UI Automation plus system-wide Win32 APIs (`FindWindowW`, `GetCursorPos`, `IsWindowVisible`), all of which work from any process. There's no reason to inject into `explorer.exe`; doing so means a copy runs in **every** `explorer.exe` instance, and each worker thread independently polls the cursor and fires `Expand`/`Collapse` on the same single flyout — they'll fight each other. Hosting it in the shell also means a bug here can destabilize the shell. This is a textbook fit for the [Mods as tools](https://github.com/ramensoftware/windhawk/wiki/Mods-as-tools:-Running-mods-in-a-dedicated-process) pattern, which also gives you single-instance behavior for free.

Recommended change:
- `@include windhawk.exe` instead of `explorer.exe`
- Rename `Wh_ModInit` / `Wh_ModSettingsChanged` / `Wh_ModUninit` → `WhTool_ModInit` / `WhTool_ModSettingsChanged` / `WhTool_ModUninit`
- Paste the tool-mod launcher boilerplate verbatim from the wiki

See [mods/keep-rainmeter-always-bottom.wh.cpp](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/keep-rainmeter-always-bottom.wh.cpp) for a working example (same shape: a worker thread doing cross-process window work, with the launcher boilerplate at the bottom).

**2. Thread teardown can orphan the worker and crash on disable.** In `Wh_ModUninit`:
```cpp
WaitForSingleObject(g_thread, 3000);
CloseHandle(g_thread);
```
The wait is bounded at 3 s, but you then `CloseHandle` and return regardless of whether the thread actually exited. `pollInterval` is only clamped at the low end (`>= 10`), so a user who sets, say, `pollInterval: 5000` makes `Sleep(g_settings.pollInterval)` outlast the 3 s wait — the wait times out, Windhawk unloads the mod DLL, and the still-sleeping thread then wakes into freed code → crash. A pathological cross-process UIA call (`FindAll` / `get_CurrentBoundingRectangle`) could also occasionally exceed 3 s. Fix: make the sleep interruptible and join unconditionally so uninit is both prompt *and* safe regardless of `pollInterval`:
```cpp
// create a manual-reset event in Wh_ModInit; signal it in Wh_ModUninit
WaitForSingleObject(g_stopEvent, g_settings.pollInterval);   // instead of Sleep()
...
// uninit:
g_running = false;
SetEvent(g_stopEvent);
WaitForSingleObject(g_thread, INFINITE);   // safe: the thread can no longer block long
CloseHandle(g_thread);
```

**3. Add a screenshot/GIF to the README.** This is a visible-effect mod, so a short GIF of the hover-to-expand behavior helps a lot in the catalog. The sibling mod [taskbar-tray-show-on-hover](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/taskbar-tray-show-on-hover.wh.cpp) embeds one (`https://i.imgur.com/...gif`); only `i.imgur.com` and `raw.githubusercontent.com` are allowed image hosts.

<p>

- **Data race on settings change.** `Wh_ModSettingsChanged` → `LoadSettings()` reassigns `g_settings` (including the `std::wstring` members) on the settings thread while the worker thread reads them every tick (`g_settings.flyoutClass.c_str()` in `IsFlyoutOpen`/`CursorOverFlyout`, `g_settings.trayIconAutomationId`, `g_settings.keywords`). Reassigning a `std::wstring` while another thread reads it is a data race that can crash. Since it only happens on a settings change it's not a blocker, but consider guarding `LoadSettings()` and the worker's reads with a small mutex, or snapshotting the strings the worker needs.

- **Prefer `std::atomic<bool>` over `volatile bool` for `g_running`.** `volatile` isn't a synchronization primitive; `std::atomic<bool>` is the idiom and pairs naturally with the stop-event fix above.

- **Redundant NULL checks on `Wh_GetStringSetting`.** It never returns `NULL` — it returns `L""` on unset/error — so `if (fc)` / `if (aid)` can never be false. As written, clearing the setting to empty would overwrite your defaults with `""` and break `FindWindowW`. Consider `WindhawkUtils::StringSetting` (RAII, no manual `Wh_FreeStringSetting`) and only override the default when the value is non-empty.

- **`IsFlyoutOpen()` runs `FindWindowW` on every tick** because `needState = (...) || g_settings.autoClose` and `autoClose` defaults to `true`. You only need the flyout-visibility check on the cursor-enter edge and while auto-collapsing; gating it more tightly avoids ~20 `FindWindowW` calls/sec when nothing is happening.

- **Code comments / `#error` message are in Polish.** The user-facing strings are already English (good); for a public mod, English source comments are preferred for maintainability.

</p>
</details>

<p>

- **Overlap check — this is distinct.** The closest existing mods are [taskbar-tray-show-on-hover](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/taskbar-tray-show-on-hover.wh.cpp) (auto-hides the *whole* tray area and reveals it on hover) and [taskbar-notification-icons-show-all](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/taskbar-notification-icons-show-all.wh.cpp) (forces all hidden icons to always show). Neither does "open the overflow flyout on hover", so this isn't a duplicate — worth a one-line mention in the README of how it differs, for users deciding between them.

- **`keywords` is hardcoded in the source.** Non-English/non-Polish users currently have to edit and recompile (or rely on the leftmost-icon heuristic). Exposing it as an array setting (`- keywords: [...]`, read with `keywords[%d]`) would let users add their locale's chevron name without touching the source — and lets you drop the README instruction to edit the list.

- **Cursor polling is the core mechanism and has no clean event-driven alternative** for "cursor entered this screen rect" (a low-level mouse hook would be worse). So the continuous 50 ms poll plus the periodic UIA `FindAll` (every 3 s) is an inherent, modest background cost — noting it as an FYI, not a change to make.

- **Chevron→flyout gap.** Moving the cursor from the chevron up into the flyout may briefly cross neither rect; the `grace` delay (default 200 ms) absorbs this, but a user setting `grace: 0` could see premature collapses. Fine as-is given the default.

</p>
</details>

wygodad:
Thanks for the thorough review! All points addressed in 1.4.0:

**Required:**
1. **Converted to a tool mod** — `@include windhawk.exe`, lifecycle renamed to `WhTool_*`, launcher boilerplate pasted verbatim (per `keep-rainmeter-always-bottom.wh.cpp`).
2. **Safe thread teardown** — the worker now sleeps on a manual-reset stop event (`WaitForSingleObject(g_stopEvent, pollInterval)`); uninit signals the event and joins with `INFINITE`, so teardown is prompt and safe regardless of `pollInterval`.
3. **Added an illustration to the README** (i.imgur.com).

**Optional items — all done as well:**
- Settings data race fixed: `LoadSettings()` builds a local struct and swaps it under an SRWLOCK; the worker keeps a private snapshot refreshed via a generation counter.
- `g_running` is now `std::atomic<bool>`.
- String settings: removed the impossible NULL checks; defaults are only overridden by non-empty values, so clearing a setting no longer breaks `FindWindowW`.
- `IsFlyoutOpen()` is gated: checked on the cursor-enter edge and while watching an open flyout; when idle it's throttled to one check per 500 ms (so a manually opened flyout still auto-closes).
- All source comments translated to English.
- `keywords` exposed as an array setting, so other locales can add their chevron name without recompiling; README updated accordingly.
- README now mentions how this differs from *taskbar-tray-show-on-hover* and *taskbar-notification-icons-show-all*.

Also added a Polish translation to the readme and `pl-PL` strings for the settings.

m417z:
### Submission review

---


**Tool-mod boilerplate deviates from the wiki snippet — the session-0 guard was dropped.** The maintainer asks that the [tool-mod boilerplate](https://github.com/ramensoftware/windhawk/wiki/Mods-as-tools:-Running-mods-in-a-dedicated-process) be pasted unchanged. Your `Wh_ModInit` is missing the leading session check that the wiki snippet opens with:

```cpp
BOOL Wh_ModInit() {
    DWORD sessionId;
    if (ProcessIdToSessionId(GetCurrentProcessId(), &sessionId) &&
        sessionId == 0) {
        return FALSE;
    }
    ...
```

Without it, the launcher logic can run in a session-0 `windhawk.exe` and spawn a tool process there, where there's no interactive taskbar for the mod to act on. Please restore that guard so the boilerplate matches the wiki byte-for-byte (the explanatory header comment from the snippet is worth keeping too).

P.S. Recommending `keep-rainmeter-always-bottom` as an example previously was a mistake, I'll improve the review.

<p>

- **Likely-unused link libraries.** You only do COM activation (`CoCreateInstance` → `ole32`) plus `VariantInit`/`SysFreeString` (`oleaut32`) and `CommandLineToArgvW` (`shell32`); the UIA interfaces are reached purely through `__uuidof`/COM. `-luuid` and `-luiautomationcore` appear unnecessary (no `uiautomationcore.dll` exports such as `UiaRaiseAutomationEvent` are called, and no named `IID_`/`CLSID_` GUID constants are referenced). You can probably drop both — just confirm it still links.
- **Two `FindWindowW` lookups per relevant tick.** `IsFlyoutOpen` and `CursorOverFlyout` each independently call `FindWindowW(flyoutClass)` + `IsWindowVisible`. When both run in the same tick you could resolve the flyout `HWND` once and reuse it. Negligible cost, but it removes a duplicated lookup.

</p>
</details>

<p>

- **The unconditional 3 s re-find looks redundant.** Every `REFIND_INTERVAL_MS` you release `pBtn` and re-run the full `FindAll(TreeScope_Subtree)`, even when the cached element is still valid. Position changes are already handled by the 750 ms `get_CurrentBoundingRectangle` refresh, and a stale/destroyed element makes that same call fail, which already triggers a re-find on the next loop. So you could make the re-find lazy (only when the rect query fails) and avoid a periodic cross-process UIA subtree walk that usually finds nothing new.

</p>
</details>

wygodad:
Thanks a lot for the detailed review, and for the follow-up on the example — much appreciated! All addressed in 1.4.2:

**Required:**
- Restored the session-0 guard at the top of `Wh_ModInit` and the canonical explanatory header comment, so the tool-mod boilerplate now matches the wiki verbatim.

**Optional polish — all done:**
- Dropped `-luuid` and `-luiautomationcore`; only COM activation via `__uuidof` is used, and it still compiles and links cleanly.
- The flyout window handle is now resolved once per tick (`GetVisibleFlyout`) and reused by both the click-inside detection and the auto-collapse check, removing the duplicated `FindWindowW`.

**Functionality note:**
- Made the re-find lazy — the periodic subtree walk is gone. The element is only re-found when there isn't a valid one; a stale/destroyed element makes `get_CurrentBoundingRectangle` fail, which clears it and re-finds promptly on the next tick. When the chevron is absent (e.g. no hidden icons), retries are throttled to the re-find interval instead of walking the tree every tick.

Tested on Windows 11 25H2 (build 26200.8524).

PR #4239, opened 2026-05-30
Files 2 Folders v2.0

triatomic:
## Changelog

* Fixed: the "Files 2 Folder..." menu entry now appears and works on the desktop (previously only inside Explorer folder windows).
* Added Silent mode (right-click menu): skips the dialog and immediately moves the selection using the default mode; suppresses the mod's own dialogs/error popups. The hotkey workaround is unaffected.
* "Default selected mode" is now a dropdown showing the actual mode names instead of a raw 1-4 number box.
* Fixed: the dialog now correctly pre-selects the configured "Default selected mode" (it was always reverting to the first option).
* Also handles selections in library/search views by deriving the destination from the selected items.
* Fixed: no longer inserts a doubled separator when "Place menu item near Cut/Copy" is on and Explorer already has a separator there.
* The menu entry and default subfolder name now use Windows' own localized strings (shell32: "Move to a folder...", "New folder"), so they appear in the OS language.
* Fixed-name / date modes no longer silently reuse an existing folder of that name — they number duplicates like Explorer ("New folder (2)", "(3)", ...).
* The "Files 2 Folder..." entry now also appears in the classic menu bar's Edit dropdown (after Cut/Copy).
* Settings descriptions rewritten for consistency (On/Off states on their own lines, unified quoting/terminology).
* Workaround hotkey: Win-based combos now work via a low-level keyboard hook (the technique PowerToys uses) and suppress the OS default, instead of failing to register. Note: bare Win+<letter> shortcuts (Win+F, Win+L, ...) are reserved by Windows and cannot be intercepted by any hook — pair Win with another modifier (e.g. Win+Shift+F); the mod now warns when a bare Win combo is selected.

## Mod authorship

This mod was created by:

- - [x] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/files-2-folders.wh.cpp
+++ b/mods/files-2-folders.wh.cpp
@@ -13,6 +13,8 @@
 /*
 # Files 2 Folders
 
+![Files 2 Folders demo](https://raw.githubusercontent.com/triatomic/windhawk-mods-assets/main/files-2-folders.gif)
+
 Inspired by [Files 2 Folder](https://www.dcmembers.com/skwire/download/files-2-folder/)
 by Jody Holmes (Skwire Empire).
 
--- a/mods/files-2-folders.wh.cpp
+++ b/mods/files-2-folders.wh.cpp
@@ -2,7 +2,7 @@
 // @id              files-2-folders
 // @name            Files 2 Folders
 // @description     Move one or more selected files in Explorer into a subfolder (named, by extension, by name, or by date), with a workaround hotkey for other file managers
-// @version         1.4
+// @version         1.6
 // @author          tria
 // @github          https://github.com/triatomic
 // @include         explorer.exe
@@ -20,7 +20,10 @@ When one or more items are selected in an Explorer window, adds a
 **Files 2 Folder...** entry to the right-click context menu. Choosing it opens
 a dialog with four options for moving the selection into a new subfolder:
 
-1. Move the selection into a subfolder with a fixed name (e.g. `folder`).
+1. Move the selection into a subfolder with a fixed name (defaults to the
+   localized Windows "New folder" name; e.g. `New folder`). If a folder of
+   that name already exists, the new one is numbered `New folder (2)`,
+   `New folder (3)`, etc. — the same way Explorer numbers duplicates.
 2. Move each file into a subfolder named after the file (without extension):
    `Good.bat` -> `.\Good\Good.bat`. With a single file selected, this just
    creates one folder around that file.
@@ -33,7 +36,12 @@ a dialog with four options for moving the selection into a new subfolder:
 Folders inside the selection are skipped for modes 2 and 3 (they only make
 sense for files); they are still moved for modes 1 and 4.
 
-This works in both Explorer folder windows and on the **desktop**.
+This works in both Explorer folder windows and on the **desktop**. The entry
+also appears in the classic menu bar's **Edit** dropdown (just after Cut/Copy)
+when the menu bar is shown.
+
+The menu entry and the default subfolder name use Windows' own localized
+strings, so they appear in your OS language automatically.
 
 ## Silent mode
 Enable **Silent mode (right-click menu)** in settings to skip the dialog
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

Mainly, note the multi-process explorer.exe issue. Either improve the mod to support it, or, less preferable, document it as a known limitation.

---

This is a solid, well-commented update — desktop support, the localized labels, the default-mode pre-selection fix, and the duplicate-numbering all look correct. A couple of things around the new failure popups are worth a look.

**Hotkey-failure popups repeat and can spam across Explorer instances.** The new modal `MessageBox` calls in `HotkeyThreadProc` — the `RegisterHotKey` "already in use" box ([L1566–L1584](https://github.com/ramensoftware/windhawk-mods/blob/6f501aea02957058d3701aa5631b18e93d18859a/mods/files-2-folders.wh.cpp#L1566-L1584)), the bare-Win warning ([L1543–L1554](https://github.com/ramensoftware/windhawk-mods/blob/6f501aea02957058d3701aa5631b18e93d18859a/mods/files-2-folders.wh.cpp#L1543-L1554)), and the LL-hook-failed box — run from the hotkey thread, which `Wh_ModSettingsChanged` → `RestartHotkeyThread` re-creates **on every settings change**. So a user whose combo conflicts with another app, or who picked a bare Win combo, gets the same modal dialog again every time they touch *any* setting. On top of that, a global hotkey is system-wide-unique and this mod runs in *every* `explorer.exe`: when more than one instance exists (separate-process folder windows, certain special folders, transiently during an Explorer restart), every instance after the first fails `RegisterHotKey` and pops its own dialog. v1.3 only `Wh_Log`'d here, which avoided all of this.

You added the popups deliberately ("make the failure visible") and that's reasonable for a genuine conflict — but consider keeping `Wh_Log` (already present) as the primary channel and either dropping the popups or showing them at most once (e.g. remember the last combo you warned about and only re-warn when it actually changes), and treating "already registered" as expected when another of your own instances already owns it. At minimum, don't re-pop on every settings reload.

**README has no visual.** This mod has a clearly visible effect (the injected menu entry and the mode dialog). A screenshot or short GIF of the right-click entry + dialog would help users a lot — only `i.imgur.com` / `raw.githubusercontent.com` are allowed image hosts.

<p>

- **`g_settings` read/written concurrently on a settings change.** `Wh_ModSettingsChanged` calls `LoadSettings()` *before* `RestartHotkeyThread()` ([L301–L304](https://github.com/ramensoftware/windhawk-mods/blob/6f501aea02957058d3701aa5631b18e93d18859a/mods/files-2-folders.wh.cpp#L301-L304)), so the still-running hotkey thread / active LL hook can read `g_settings` (including the `hotkeyChar` `std::wstring` in `BuildHotkeyVk`) while `LoadSettings` is reassigning it — a data race that can only happen during a settings change. Reordering to `StopHotkeyThread(); LoadSettings(); StartHotkeyThread();` closes the window for the hotkey path.
- **`DlgProc` statics are shared across dialog instances.** `static bool ready` / `static DlgState* s` / `static DlgTheme theme` ([L953–L959](https://github.com/ramensoftware/windhawk-mods/blob/6f501aea02957058d3701aa5631b18e93d18859a/mods/files-2-folders.wh.cpp#L953-L959)) are function-local statics, so a context-menu dialog and a hotkey dialog open at the same time (different threads) would clobber each other. Pre-existing for `s`/`theme`; `ready` joins them. Stashing state in `DlgState` (via `GWLP_USERDATA`) would make it robust.
- **Dangling doc reference.** The LL-hook comment says "See keyboardhook.md for the full risk write-up" ([L1483–L1484](https://github.com/ramensoftware/windhawk-mods/blob/6f501aea02957058d3701aa5631b18e93d18859a/mods/files-2-folders.wh.cpp#L1483-L1484)), but a single-file mod ships no such file — drop or inline it.
- **Menu label reuse.** The context-menu entry now uses shell32 string 30305 ("Move to a folder…"), which is the exact name of a *different* built-in shell command (move the selection into an existing folder). Some users may expect the built-in behavior. You clearly thought about this (the commit history shows the "(F2F)" disambiguation), so just flagging that the mod's own identity is now invisible in the menu.
- **Unload while a dialog is open.** `StopHotkeyThread` waits only 2 s then `CloseHandle`s and proceeds; if a F2F/MessageBox dialog is open on the hotkey thread at unload time, the thread is leaked and may run unmapped mod code afterward. Edge case (user unloads mid-dialog), and the new failure popups widen the surface slightly.

</p>
</details>

<p>

- **System-wide LL keyboard hook (opt-in).** When a Win combo is enabled, every keystroke system-wide passes through `LowLevelKeyboardProc`. The callback is correctly minimal — it just matches and `PostMessageW`s to the worker window, doing the real work off the hook ([L1502–L1518](https://github.com/ramensoftware/windhawk-mods/blob/6f501aea02957058d3701aa5631b18e93d18859a/mods/files-2-folders.wh.cpp#L1502-L1518)) — and it's the same technique PowerToys uses, so there's no cleaner alternative. Just an FYI of the cost; it's only active when the user opts into a Win combo.
- **No single-instance dedupe.** With multiple `explorer.exe` instances, each installs its own LL hook (extra per-keystroke overhead) and its own hotkey thread, on top of the `RegisterHotKey` collision noted above.
- **Fixed-name mode no longer reuses an existing folder.** Mode 1 now always numbers duplicates via `UniqueDest(..., splitExtension=false)` ([L733](https://github.com/ramensoftware/windhawk-mods/blob/6f501aea02957058d3701aa5631b18e93d18859a/mods/files-2-folders.wh.cpp#L733)). For *date* mode this is clearly right, but "fixed name" implies a deliberate, reusable name — a user repeatedly sending files to a folder called `archive` will now accumulate `archive`, `archive (2)`, `archive (3)`… instead of dropping everything into `archive`. It's intentional per the changelog; flagging in case the fixed-name case should keep the reuse behavior (the per-name / per-extension modes still accumulate).
- **Desktop virtual items.** When the selection on the desktop includes namespace items (This PC, Recycle Bin, the user's home-folder icon), `SHGDN_FORPARSING` yields a `::{GUID}` string rather than a path, so the move silently no-ops on those — and in silent mode there's no feedback at all. Minor, given how rarely those get "moved".

</p>
</details>

m417z:
Let me know when the mod is ready to be reviewed and merged.

triatomic:
should be ready for both. I have applied your last suggestions

Anixx:
Somehow, the option to group this with copy and paste disappeared?... Was useful.

PR #4247, opened 2026-05-31
Add Open in Admin Terminal mod

aimagist:
## Changelog

* New mod submission.
* Adds a classic Explorer context menu entry to open an elevated terminal in the current or selected folder.
* Supports folder background, folder item, drive root, Desktop, and optional navigation pane / Quick access targets.
* Supports Auto, Windows Terminal, PowerShell 7, Windows PowerShell, Command Prompt, WSL, Git Bash, WezTerm, Alacritty, ConEmu, and a custom command.

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [x] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/open-in-admin-terminal.wh.cpp
+++ b/mods/open-in-admin-terminal.wh.cpp
@@ -1,0 +1,1588 @@
+// ==WindhawkMod==
+// @id              open-in-admin-terminal
+// @name            Open in Admin Terminal
+// @description     Adds an Explorer classic context menu entry to open an elevated terminal in the current or selected folder.
+// @version         1.15
+// @author          aimagist
+// @github          https://github.com/aimagist
+// @include         explorer.exe
+// @compilerOptions -lole32 -loleaut32 -luuid -lshlwapi -lshell32 -lgdi32
+// @license         MIT
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Open in Admin Terminal
+
+**TL;DR:** Right-click any folder or drive in File Explorer and open an elevated terminal there — no more navigating after the fact.
+
+## With this mod you can:
+
+- Right-click a folder background and open an admin terminal in that location
+- Right-click a folder item and open an admin terminal inside it
+- Right-click a drive and open an admin terminal at its root
+- Optionally show the entry when right-clicking filesystem folders and drives in the navigation pane and Quick access
+- Choose your preferred terminal: Auto, Windows Terminal, PowerShell 7, Windows PowerShell, Command Prompt, WSL, Git Bash, WezTerm, Alacritty, ConEmu, or a custom command
+- Customize the context menu label, or let the mod use a smart default based on your terminal choice
+- Optionally append the terminal name to a custom label (e.g. "Open elevated (Windows Terminal)")
+
+## Preview
+
+![Normal usage — right-click folder background or folder item to open elevated Windows Terminal](https://i.imgur.com/qEgGpvc.gif)
+
+*Right-clicking a folder item to open an elevated terminal in that location (Windows Terminal)*
+
+![Context menu with Windows Terminal option](https://i.imgur.com/HziEs16.png) ![Context menu with PowerShell 7 option](https://i.imgur.com/CCE3vlP.png)
+
+![Mod settings showcase](https://i.imgur.com/8KgCdju.gif)
+
+*Settings panel — choose terminal, customize label, toggle terminal name appending*
+
+Screenshots may show earlier builds, but current releases use runtime classic-menu injection. The mod does not write Explorer context menu entries to the registry.
+
+## Files
+
+- `open-in-admin-terminal.wh.cpp` — Windhawk mod source
+
+## How to use
+
+1. Open Windhawk.
+2. Create a new mod.
+3. Paste in `open-in-admin-terminal.wh.cpp`.
+4. Compile and enable it.
+
+## Notes
+
+- On Windows 11, Explorer may place this entry under `Show more options` depending on your context menu setup.
+- The entry is injected only while Explorer's classic menu is open; disabling the mod leaves no registry cleanup behind.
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

Also, update or remove the "Files" and "How to use" sections in the readme. Normally, users just have to install the mod, not to compile it manually.

---

The items below are mostly about the launch mechanism and the custom-command option.

**1. Elevation is routed through a mandatory `powershell.exe` launcher.** Every launch — even `cmd`, `wt`, `wsl`, etc. — is wrapped in `powershell.exe -NoProfile -WindowStyle Hidden -EncodedCommand <base64 of "Start-Process … -Verb RunAs">` (`BuildStartProcess`, line 653). This has a few downsides for users:

- **Single point of failure:** if Windows PowerShell (`powershell.exe`) isn't available on `PATH`, *nothing* launches regardless of the selected terminal. Microsoft is steering Windows PowerShell 5.1 toward being an optional/removable component, so this is a real robustness concern.
- **Wrong UAC label:** the consent prompt names "Windows PowerShell" rather than the terminal the user actually chose, which is confusing.
- It also pulls in ~150 lines of escaping/encoding machinery (`EscapePS`, `BuildPSStringLiteral`, `QuoteProcessArgument`, `BuildPSArgumentArray`, `Base64Encode`, `EncodePowerShellCommand`) that depend on Windows PowerShell 5.1's specific `-ArgumentList` concatenation behavior.

Recommend elevating directly with `ShellExecuteEx` + the `runas` verb instead — `lpFile` = the resolved exe, `lpParameters` = the per-terminal argument string, `lpDirectory` = the target folder. That elevates without an intermediary process, shows the real terminal in the UAC prompt, removes the `powershell.exe` dependency, and lets you delete almost all of the PS-quoting code. See [keyboard-shortcut-actions.wh.cpp](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/keyboard-shortcut-actions.wh.cpp#L1255) for the `runas` pattern.

**2. The "Custom command" option never receives the folder path.** When `terminalChoice == "custom"`, `BuildCommand` (line 669) matches none of the `if (choice == …)` branches and falls through to `BuildStartProcess(exe, {}, {})` — no arguments and no working directory. So a custom terminal opens *elevated but at the default location*, not at the right-clicked folder, which defeats the mod's purpose for that choice. Additionally, a custom string that includes arguments (e.g. `wt.exe -d`) is passed wholesale as `-FilePath` and won't resolve. At minimum, set the working directory to the target. Ideally, support a placeholder so the user can position the path explicitly — Windows itself uses `%V` for the folder in its built-in "open terminal here" command templates; substituting `%V` (and/or `%1`) with the target path would be the idiomatic, predictable option. (If you switch to `ShellExecuteEx` with `lpDirectory` per item 1, the working-directory half of this is solved for free.)

**3. `debugLogging` setting + `DEBUG_LOG` macro reimplement something Windhawk already provides.** Windhawk lets the user enable/disable a mod's logging from the UI and already prefixes log lines with the mod name, and `Wh_Log` compiles to a cheap guarded check that's disabled by default — so it's safe to call directly even in the menu path. You can drop the `debugLogging` setting and the `DEBUG_LOG` wrapper and just call `Wh_Log` where needed.

<p>

- **`GetSettingString` (line 197)** does a raw `Wh_GetStringSetting` + `Wh_FreeStringSetting`; `WindhawkUtils::StringSetting` is the RAII idiom for this. Also note `Wh_GetStringSetting` never returns `NULL` (it returns `L""` when unset/on error), so the `if (PCWSTR s = …)` guard's fallback branch is effectively dead.
- **Hooks use raw `Wh_SetFunctionHook` with `reinterpret_cast<void*>`** (lines 1555, 1562). `WindhawkUtils::SetFunctionHook` is the type-safe equivalent. (Low priority — files-2-folders uses the raw form too.)
- **`-loleaut32`** in `@compilerOptions` doesn't appear to be used (the code uses plain `VARIANT` structs but no `oleaut32` API such as `SysAllocString`/`VariantClear`). Worth double-checking and removing if unused.
- The simpler variant of `Wh_ModSettingsChanged` can be used: `void Wh_ModSettingsChanged() {...}`.

</p>
</details>

<p>

- **`position: Default` placement is English-only.** `InsertAdminTerminalMenuItem` (line 1406) finds its insertion point by matching the substrings `"Open"`/`"Terminal"` in existing menu item text. On a non-English Windows UI nothing matches and it silently falls back to the top. It degrades gracefully, but the "Default" placement won't behave as intended in other languages.
- **Per-menu resolution cost.** Each right-click in a shell view resolves the target via `CoCreateInstance(CLSID_ShellWindows)` + an enumeration of shell windows. It's gated to shell-view / navigation-pane menus and mirrors how files-2-folders works, so there's no obviously cheaper route — just noting the small added latency on context-menu open.
- **Menu icon DPI.** `CreateMenuBitmapFromIcon` sizes the bitmap from `GetSystemMetrics(SM_CXMENUCHECK)` (system DPI). On mixed per-monitor-DPI setups the injected icon may be slightly mis-sized on a secondary monitor. Purely cosmetic.

</p>
</details>

aimagist:
Thanks for the review — addressed the requested launch/custom-command/logging/docs items in `e7665db`:

- Switched elevation to direct `ShellExecuteExW` with `runas`, so the UAC prompt is for the selected terminal executable.
- Removed the mandatory PowerShell launcher dependency and the related encoding/trampoline machinery.
- Fixed custom commands to parse quoted executable paths with trailing args, use the selected folder as the working directory, and expand `%V` / `%1` placeholders in parameters.
- Removed the custom `debugLogging` setting and `DEBUG_LOG` macro in favor of direct `Wh_Log` calls.
- Removed the README `Files` and manual `How to use` sections.

Validation run locally:
- Windhawk compliance script: compile check passed, no mechanical blockers.
- Catalog `scripts/compile_mod.py`: passed for x86, x64, and ARM64.

PR #4248, opened 2026-06-01
update neko-cat to v1.3.0

ciizerr:
## ✨ New Features
- **Renamed Mod:** From *Neko Cat* → *Desktop Companions*  
- **New Characters:** Sakura and Tomoyo join the lineup 🎉  
- **Diverse Litters:** Option to spawn a randomized mix of all downloaded characters at once  

## ⚡ Performance Improvements
- **High-Performance Texture Slicing:**  
  Replaced 38 separate PNG downloads with a single spritesheet → disk & network checks are **97% faster**!  
- **Dynamic RAM Cloning:**  
  Spritesheets are sliced in memory at startup for smooth, lag-free performance  

## 🛠 Fixes & Enhancements
- **Leak-Free Switcher:** Fixed memory leak when dynamically switching characters in settings  
- **Organized Settings:** Categories are now clean and user-friendly:  
  - Appearance  
  - Behavior  
  - Audio  
  - Advanced  
- **Window Behavior:** Prevent accidental maximization when double-clicking on a desktop companion  

## ⚠️ Important Note
Due to the reorganized settings layout, you may need to **reset or reconfigure your preferences** in the Windhawk settings panel after updating.

---

### 💖 Special Thanks
Big thanks to **@zkyhzxf (quantum)** on Discord for contributing the beautiful Sakura and Tomoyo icons!

### DIFF
--- a/mods/neko-cat.wh.cpp
+++ b/mods/neko-cat.wh.cpp
@@ -1,8 +1,8 @@
 // ==WindhawkMod==
 // @id              neko-cat
-// @name            Neko Cat
-// @description     Adds a desktop pet cat that runs around and follows your mouse
-// @version         1.2.1
+// @name            Desktop Companions
+// @description     Spawn multiple animated characters (neko cat, sakura, and tomoyo) on your screen that interact with your windows and follow your cursor.
+// @version         1.3.0
 // @author          ciizerr
 // @github          https://github.com/ciizerr
 // @include         windhawk.exe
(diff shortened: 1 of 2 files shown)


m417z:
### Submission review

---

This is a solid, self-contained tool mod and the threading/teardown is handled correctly (window creation, `SetTimer`, hotkey, and `delete`/`DestroyWindow` all stay on the worker thread; the old cross-thread `g_Nekos` iteration in `WhTool_ModSettingsChanged` is now correctly moved into the `WM_UPDATE_SETTINGS` handler — good). A couple of settings-vs-code issues stand out:

**1. `AppearanceGroup.theme` setting doesn't exist — `g_theme` is always empty, which breaks the default-theme download and the fallback.** In `LoadSettings` you read [`Wh_GetStringSetting(L"AppearanceGroup.theme")`](https://github.com/ramensoftware/windhawk-mods/blob/e053ac685f2c20c8f93b09eed0d804261e12e799/mods/neko-cat.wh.cpp#L1487), but the settings block has no `theme` key directly under `AppearanceGroup` — `theme` only exists inside the `characterList[]` array (`AppearanceGroup.characterList[%d].theme`, which you read separately and correctly). So `g_theme` is always `L""`, `g_assetPath` becomes `storagePath\` (trailing separator, no theme folder), and [`DownloadMissingAssets()` → `EnsureThemeDownloaded(L"")`](https://github.com/ramensoftware/windhawk-mods/blob/e053ac685f2c20c8f93b09eed0d804261e12e799/mods/neko-cat.wh.cpp#L327) returns immediately and downloads nothing.

In the common config the `g_customPets` / random loops mask this, but the [fallback in `LoadSprites`](https://github.com/ramensoftware/windhawk-mods/blob/e053ac685f2c20c8f93b09eed0d804261e12e799/mods/neko-cat.wh.cpp#L475) assumes `neko-cat\spritesheet.png` is always present — and it's only ever downloaded when random mode is on or when `neko-cat` is in the custom list. So with random off and a custom list that doesn't include `neko-cat` (e.g. only `sakura-icon`), if that theme fails to load there's nothing to fall back to → invisible pets. Fix: drop the dead `g_theme` / `AppearanceGroup.theme` machinery (it's superseded by `g_customPets` + `GetThemePathForPet`), and make the fallback reliable by always `EnsureThemeDownloaded(L"neko-cat")` before relying on it.

**2. "Number of Characters" (`character_count`) has no effect in the default mode.** The target count is `g_randomThemes ? g_catCount : (g_customPets.empty() ? g_catCount : g_customPets.size())` ([line 1343](https://github.com/ramensoftware/windhawk-mods/blob/e053ac685f2c20c8f93b09eed0d804261e12e799/mods/neko-cat.wh.cpp#L1343)). With random off and a non-empty character list (the out-of-the-box state), the count comes from the list length and `character_count` is silently ignored. A user who bumps "Number of Characters" to 5 while leaving random off sees no change — they have to add five list entries instead. Either make `character_count` apply in custom mode (e.g. cycle the list to fill N pets), or update the two `$description`s to make clear that count is driven by the list in custom mode and by "Number of Characters" only in random mode.

<p>

Minor polish — none of this affects users in normal use, so it's your call.

- **Settings-change data race.** `LoadSettings()` runs on Windhawk's settings-changed thread and mutates globals the worker thread reads (the `g_customPets` vector, the `g_nekoKeyStr`/`g_theme` strings, and the int settings). Two settings changes in quick succession could overlap `LoadSettings` 2 with the worker still handling `WM_UPDATE_SETTINGS` 1, which is a real data race on the vector/strings. Cleaner: in `WhTool_ModSettingsChanged`, just `PostThreadMessage(WM_UPDATE_SETTINGS)` and call `LoadSettings()` inside the handler so all settings access stays on the worker thread. (Only occurs on a settings change, hence optional.)
- **Dead NULL checks on `Wh_GetStringSetting`.** `Wh_GetStringSetting` never returns `NULL` (it returns `L""` when unset/on error), so `if (themeStr) … else { g_theme = L"neko-cat"; }` and the same pattern for `nekoKeyStr` ([lines 1487, 1524](https://github.com/ramensoftware/windhawk-mods/blob/e053ac685f2c20c8f93b09eed0d804261e12e799/mods/neko-cat.wh.cpp#L1524)) have dead `else` branches — check for an empty string instead. While there, `WindhawkUtils::StringSetting` (RAII) would replace the raw get + manual `Wh_FreeStringSetting`.

</p>
</details>

<p>

- **`rand()` is never seeded.** There's no `srand(...)`, so the new "Randomize Multiple Characters" feature picks the same theme arrangement on every launch, and spawn positions are identical each run too. Seed once in `WhTool_ModInit` (e.g. `srand(GetTickCount())`).
- **Audio is single-stream and process-wide.** `PlayAudio` uses `PlaySoundW(SND_ASYNC)` and `StopAudio` is `PlaySoundW(NULL, NULL, 0)`, so only one companion's sound is ever audible at a time, and hiding/trimming one companion silences all of them. That's inherent to `PlaySound`; just noting it given the mod now spawns multiple characters.
- **A settings change that introduces a new (uncached) theme blocks the animation loop.** `Wh_GetUrlContent` is synchronous and is called from the worker thread's `WM_UPDATE_SETTINGS` handler, so when a not-yet-downloaded theme is added, all pets freeze until the download finishes. First-run asset download (and the network dependency) is fine and consistent with other mods that use `Wh_GetUrlContent`, and pinning the URL to a commit hash is good — no action needed there.

</p>
</details>

ciizerr:
## ✅ Review Updates

Thanks for the thorough review! I’ve pushed a commit addressing all feedback:

- Removed unused `g_theme` variables and ensured the neko‑cat fallback always downloads.  
- “Number of Characters” now works correctly in Custom mode (cycles the list to fill the target count).  
- Fixed the data race by moving `LoadSettings()` into the `WM_UPDATE_SETTINGS` handler.  
- Cleaned up null checks and added `srand()` so everything is randomized properly on launch.  

Ready to merge 🎉

PR #4254, opened 2026-06-01
Update File Explorer Content Animation

crazyboyybs:
## Changelog

- GPU-accelerated fade - The fade effect now uses a DirectComposition overlay capturing the previous folder's content (including Mica/Acrylic). The overlay fades out on the GPU compositor thread, fully independent of CPU load from thumbnail rendering.
- Page transitions - Detects the navigation type and applies a different animation for each context:
Entrance (slide up) - Opening Explorer, new tabs, and general navigation.
Enter folder (short fast slide from right) - Navigating into subfolders.
Back slide (slide from left) - Going back via button, breadcrumb, or Up arrow.
Forward slide (slide from right) - Going forward after going back.
Reverse direction - Content can slide down instead of up.
- Navigation pane animation - The nav pane tree slides in from the left when opening Explorer or a new tab.
- Property page animation - Animates tab switching in folder and file Properties windows.
- Requires symbol resolution on explorerframe.dll (handled automatically by Windhawk).

### DIFF
--- a/mods/file-explorer-content-animation.wh.cpp
+++ b/mods/file-explorer-content-animation.wh.cpp
@@ -399,12 +399,11 @@ static constexpr ULONGLONG PENDING_EXPIRE_MS = 3000;
 // o mesmo root que ConsumePendingTransition usa. Mapeamento correto.
 // -----------------------------------------------
 
-// Mensagem customizada usada em Wh_ModBeforeUninit para solicitar que o
-// thread da janela remova o subclass com seguranca antes do DLL descarregar.
-static constexpr UINT WM_WH_REMOVE_SUBCLASS  = WM_USER + 0x5A;
-// Enviada para o thread da janela para tentar o hook via vtable
-// depois que a janela esta totalmente inicializada.
-static constexpr UINT WM_WH_TRY_VTABLE_HOOK  = WM_USER + 0x5B;
+// Mensagens customizadas registadas via RegisterWindowMessageW para evitar
+// colisao com WM_USER+N que CabinetWClass ou outros mods possam usar.
+// Valores no range 0xC000-0xFFFF, unicos por processo.
+static UINT WM_WH_REMOVE_SUBCLASS = 0;
+static UINT WM_WH_TRY_VTABLE_HOOK = 0;
 
 
 // Forward declarations -- definidos mais abaixo, usados no subclass proc
@@ -724,8 +723,8 @@ struct PendingCleanup {
 static std::vector<PendingCleanup> g_cleanupQueue;
 static std::mutex                  g_cleanupMtx;
 static std::condition_variable     g_cleanupCv;
-static std::thread                 g_cleanupThread;
-static std::thread                 g_initThread;     // D3D/DComp init em background
+static std::thread*                g_cleanupThread = nullptr;
+static std::thread*                g_initThread    = nullptr; // D3D/DComp init em background
 
 // Frame capturado no BrowseObject hook (conteudo antigo visivel e dimensionado).
 // Consumido em HookShowWindow para criar o overlay DComp.
@@ -1605,7 +1604,14 @@ static void DispatchAnimation(HWND hwnd, AnimInfo info) {
     {
         std::lock_guard<std::mutex> lk(g_animMtx);
         // Nao iniciar novas animacoes durante o descarregamento.
-        if (g_unloading.load(std::memory_order_relaxed)) return;
+        // Restaurar o estado da janela (HideViaRgn/BeginFade aplicados
+        // no hook antes de chamar esta funcao) para evitar que a janela
+        // fique invisivel ou semi-transparente permanentemente.
+        if (g_unloading.load(std::memory_order_relaxed)) {
+            RestoreRgn(hwnd);
+            if (info.useFade && !info.useDCompFade) EndFade(hwnd);
+            return;
+        }
         if (g_animating.count(hwnd)) return;
         g_animating.insert(hwnd);
     }
@@ -1732,6 +1738,13 @@ BOOL WINAPI HookSetWindowPos(HWND hwnd, HWND hwndAfter,
 BOOL Wh_ModInit() {
     LoadSettings();
 
+    // Registar mensagens customizadas no range 0xC000-0xFFFF para evitar
+    // colisao com WM_USER+N que CabinetWClass possa usar internamente.
+    WM_WH_REMOVE_SUBCLASS = RegisterWindowMessageW(
+        L"Windhawk_FileExplorerContentAnimation_RemoveSubclass");
+    WM_WH_TRY_VTABLE_HOOK = RegisterWindowMessageW(
+        L"Windhawk_FileExplorerContentAnimation_TryVtableHook");
+
     // Registar hooks ANTES de iniciar threads.
     // Se qualquer Wh_SetFunctionHook falhar, retornamos FALSE e o Windhawk
     // nao chama Wh_ModUninit. Se as threads ja estivessem a correr,
@@ -1751,14 +1764,16 @@ BOOL Wh_ModInit() {
     // Hooks registados com sucesso — agora iniciar threads de suporte.
 
     // Worker de cleanup DComp: dorme e processa a fila de cleanup.
-    g_cleanupThread = std::thread(CleanupWorkerProc);
+    // Heap-allocated (como g_initThread) para evitar std::terminate() em
+    // DLL_PROCESS_DETACH quando Explorer termina sem mod unload.
+    g_cleanupThread = new std::thread(CleanupWorkerProc);
 
-    // Iniciar D3D11/DComp em background desde Wh_ModInit (mais cedo que
-    // Wh_ModAfterInit) para maximizar a probabilidade de estar pronto
-    // antes da primeira navegacao do utilizador.
-    // Thread armazenado em g_initThread e joined em Wh_ModBeforeUninit
-    // para evitar UAF se o mod for descarregado durante a inicializacao.
-    g_initThread = std::thread([]() {
+    // Iniciar D3D11/DComp em background para maximizar a probabilidade de
+    // estar pronto antes da primeira navegacao do utilizador.
+    // Heap-allocated para evitar std::terminate() em DLL_PROCESS_DETACH
+    // (Explorer shutdown/restart -- ver std::thread destructor rules).
+    // Joined em Wh_ModBeforeUninit para evitar UAF durante mod unload.
+    g_initThread = new std::thread([]() {
         D3D_FEATURE_LEVEL level;
 
         // Tentar reutilizar um adapter DXGI ja carregado pelo Explorer para
@@ -1895,8 +1910,9 @@ static void TryHookBrowseObjectViaVtable() {
     static const GUID SID_STL =
         {0x4C96BE40, 0x915C, 0x11CF, {0x99, 0xD3, 0x00, 0xAA, 0x00, 0x4A, 0xE8, 0x37}};
 
-    // Wh_ModAfterInit roda no thread do Windhawk. Inicializar COM aqui
-    // para poder chamar CoCreateInstance com seguranca.
+    // Chamado de Wh_ModInit (main thread) ou WM_WH_TRY_VTABLE_HOOK
+    // (UI thread do Explorer). Inicializar COM aqui para poder chamar
+    // CoCreateInstance com seguranca em ambos os contextos.
     HRESULT hrCO = CoInitializeEx(NULL, COINIT_APARTMENTTHREADED);
     if (FAILED(hrCO) && hrCO != RPC_E_CHANGED_MODE) {
         Wh_Log(L"[VTABLE] CoInitializeEx falhou: 0x%08X", (unsigned)hrCO);
@@ -1984,8 +2000,10 @@ void Wh_ModBeforeUninit() {
     // Se o mod for descarregado durante a inicializacao (cold driver ~100-300ms),
     // sem este join o thread continuaria a escrever g_d3dDevice/g_dxgiDevice
     // enquanto Wh_ModUninit os liberta -> UAF.
-    if (g_initThread.joinable())
-        g_initThread.join();
+    if (g_initThread && g_initThread->joinable())
+        g_initThread->join();
+    delete g_initThread;
+    g_initThread = nullptr;
 
     // Remover subclasses antes do DLL descarregar para evitar crash com
     // proc invalido. Envia WM_WH_REMOVE_SUBCLASS para o thread da janela
@@ -2034,8 +2052,10 @@ void Wh_ModBeforeUninit() {
     // Join aqui garante que todos os refs de device AddRef'd sao libertados
     // antes de Wh_ModUninit libertar g_dcDevice.
     g_cleanupCv.notify_one();
-    if (g_cleanupThread.joinable())
-        g_cleanupThread.join();
+    if (g_cleanupThread && g_cleanupThread->joinable())
+        g_cleanupThread->join();
+    delete g_cleanupThread;
+    g_cleanupThread = nullptr;
 }
 
 void Wh_ModUninit() {
--- a/mods/file-explorer-content-animation.wh.cpp
+++ b/mods/file-explorer-content-animation.wh.cpp
@@ -266,6 +266,8 @@ resuelto en builds recientes del canal Insider/Beta.
 #include <algorithm>
 #include <atomic>
 #include <chrono>
+#include <climits>
+#include <cmath>
 #include <condition_variable>
 #include <mutex>
 #include <thread>
@@ -457,21 +459,17 @@ static LRESULT CALLBACK ExplorerSubclassProc(HWND hwnd, UINT msg,
     }
     // Solicitacao de remocao antes do DLL descarregar (enviada por Wh_ModBeforeUninit).
     if (msg == WM_WH_REMOVE_SUBCLASS) {
-        RemoveWindowSubclass(hwnd,
-            reinterpret_cast<SUBCLASSPROC>(ExplorerSubclassProc),
-            APPCOMMAND_SUBCLASS_ID);
+        WindhawkUtils::RemoveWindowSubclassFromAnyThread(hwnd, ExplorerSubclassProc);
         std::lock_guard<std::mutex> lk(g_subclassMtx);
         g_subclassedWindows.erase(hwnd);
         return 0;
     }
     if (msg == WM_DESTROY) {
-        // RemoveWindowSubclass evita que o sistema acesse o proc do DLL
-        // apos a janela ser destruida. Sem isso, a limpeza dependeria
-        // exclusivamente do Wh_ModBeforeUninit, o que cria uma janela de
-        // vulnerabilidade se o DLL for descarregado apos o WM_DESTROY.
-        RemoveWindowSubclass(hwnd,
-            reinterpret_cast<SUBCLASSPROC>(ExplorerSubclassProc),
-            APPCOMMAND_SUBCLASS_ID);
+        // RemoveWindowSubclassFromAnyThread garante que a subclass e
+        // removida com os parametros correctos (SubclassProcWrapper + id).
+        // Sem isso, a limpeza dependeria exclusivamente do auto-remove em
+        // WM_NCDESTROY, o que e seguro mas atrasa a remocao.
+        WindhawkUtils::RemoveWindowSubclassFromAnyThread(hwnd, ExplorerSubclassProc);
         std::lock_guard<std::mutex> lk(g_subclassMtx);
         g_subclassedWindows.erase(hwnd);
     }
@@ -495,17 +493,13 @@ static LRESULT CALLBACK PropSheetSubclassProc(HWND hwnd, UINT msg,
         }
     }
     if (msg == WM_WH_REMOVE_SUBCLASS) {
-        RemoveWindowSubclass(hwnd,
-            reinterpret_cast<SUBCLASSPROC>(PropSheetSubclassProc),
-            PROPSHEET_SUBCLASS_ID);
+        WindhawkUtils::RemoveWindowSubclassFromAnyThread(hwnd, PropSheetSubclassProc);
         std::lock_guard<std::mutex> lk(g_propSheetMtx);
         g_propSheets.erase(hwnd);
         return 0;
     }
     if (msg == WM_DESTROY) {
-        RemoveWindowSubclass(hwnd,
-            reinterpret_cast<SUBCLASSPROC>(PropSheetSubclassProc),
-            PROPSHEET_SUBCLASS_ID);
+        WindhawkUtils::RemoveWindowSubclassFromAnyThread(hwnd, PropSheetSubclassProc);
         std::lock_guard<std::mutex> lk(g_propSheetMtx);
         g_propSheets.erase(hwnd);
     }
@@ -746,9 +740,10 @@ struct CapturedFrame {
 // detachadas. Dorme ate o proximo item ficar pronto, processa em batch,
 // e drena a fila em shutdown (join ocorre antes de Wh_ModUninit libertar devices).
 //
-// Ordem de locks: sempre g_cleanupMtx liberado antes de adquirir g_dcFadeMtx.
-// StartDCompFade adquire g_dcFadeMtx e depois g_cleanupMtx (sequencialmente,
-// nunca aninhados) -- sem inversao de ordem possivel.
+// Ordem de locks: StartDCompFade adquire g_dcFadeMtx como lock externo (cobre
+// TODAS as chamadas DComp) e depois g_cleanupMtx (aninhado, para enfileirar).
+// CleanupWorkerProc liberta g_cleanupMtx (lk.unlock) antes de adquirir
+// g_dcFadeMtx para teardown -- nunca aninhados no worker, sem inversao.
 static void CleanupWorkerProc() {
     std::unique_lock<std::mutex> lk(g_cleanupMtx);
     for (;;) {
@@ -764,31 +759,20 @@ static void CleanupWorkerProc() {
             std::vector<PendingCleanup> drain = std::move(g_cleanupQueue);
             lk.unlock();
             for (auto& c : drain) {
-                DCompFadeAnim toDestroy{};
-                bool hasItem = false;
-                {
-                    std::lock_guard<std::mutex> lkf(g_dcFadeMtx);
-                    auto it = g_dcFades.find(c.hwndRoot);
-                    if (it != g_dcFades.end() &&
-                        it->second.generation == c.generation) {
-                        toDestroy = it->second;
-                        hasItem   = true;
-                        g_dcFades.erase(it);
-                    }
-                }
-                if (hasItem) {
-                    if (toDestroy.pTarget) {
-                        toDestroy.pTarget->SetRoot(nullptr);
-                        toDestroy.pTarget->Release();
-                    }
-                    if (toDestroy.pEffect)  toDestroy.pEffect->Release();
-                    if (toDestroy.pVisual)  toDestroy.pVisual->Release();
-                    if (toDestroy.pSurface) toDestroy.pSurface->Release();
+                std::lock_guard<std::mutex> lkf(g_dcFadeMtx);
+                auto it = g_dcFades.find(c.hwndRoot);
+                if (it != g_dcFades.end() &&
+                    it->second.generation == c.generation) {
+                    DCompFadeAnim a = it->second;
+                    g_dcFades.erase(it);
+                    if (a.pTarget) { a.pTarget->SetRoot(nullptr); a.pTarget->Release(); }
+                    if (a.pEffect)  a.pEffect->Release();
+                    if (a.pVisual)  a.pVisual->Release();
+                    if (a.pSurface) a.pSurface->Release();
                     if (c.dev) { c.dev->Commit(); c.dev->Release(); }
                 } else {
                     // Fade cancelado por nova navegacao (geracao diferente ou
                     // entrada ausente): resources ja libertados em StartDCompFade.
-                    // Libertar so o ref do device.
                     if (c.dev) c.dev->Release();
                 }
             }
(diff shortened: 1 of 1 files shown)


crazyboyybs:
@m417z Hey, about the Wh_FindFirstSymbol thing, is more of a fallback, I can remove it if you prefer it

m417z:
Why is the fallback needed? `WindhawkUtils::HookSymbols` uses `Wh_FindFirstSymbol` under the hood.

m417z:
### Submission review

Regarding `Wh_FindFirstSymbol` - I see that you're trying to match substrings. I don't think it's a good idea - if the function signature changes, the mod will likely crash.

Regarding the review below, 1 might not be important enough, up to you. Please address 2.

---

This is a large, careful rewrite — the subclass teardown (lock-copy-release + `SendMessageTimeoutW` with `SMTO_ABORTIFHUNG`), the long-lived cleanup worker, and the multi-layer symbol/vtable hooking are all done well. A few real issues to fix:

**1. Worker threads are detached and never joined on unload — UAF / crash window.** Two threads are spawned with `.detach()` and are not tracked or joined:

- The D3D11/DComp init thread in `Wh_ModInit` (the lambda ending at `.detach();`, line 1773). It runs `CreateDXGIFactory1` / `D3D11CreateDevice` / `DCompositionCreateDevice2`, which on a cold driver can take ~100–300 ms, and it never checks `g_unloading`. `Wh_ModBeforeUninit` joins only `g_cleanupThread` (line 2055), so if the mod is disabled (or reloaded) while this thread is still running, `Wh_ModUninit` releases and nulls `g_d3dDevice`/`g_dxgiDevice`/`g_dcDevice` (lines 2074–2076) while the thread is still writing them — leaking the freshly-created devices — and, after the DLL is freed, the thread is still executing code in unmapped pages → access violation. This window is narrow but absolutely reachable (toggle the mod off shortly after enabling, or a fast reload).
- The per-animation threads in `DispatchAnimation` (line 1626). These at least poll `g_unloading` each frame, but `Wh_ModBeforeUninit` only does `Sleep(100)` (line 2049) and hopes they've exited. A descheduled/starved animation thread can still be inside `RunSlideIn` (calling `SetWindowPos`/`RestoreRgn`) when the DLL unloads.

You already do the right thing for the cleanup worker — store the `std::thread` in a global and `join()` it in `Wh_ModBeforeUninit`. Extend the same pattern to the init thread (store it, set `g_unloading` first, then `join()`), and replace the `Sleep(100)` heuristic with a real wait for the in-flight animations to drain (e.g. join tracked threads, or wait on a condition variable until `g_animating` is empty). While there, make the device pointers `std::atomic` (or guard them) — they're written by the init thread and read by the UI-thread hooks (`Classify` line 1273, `HookBrowseObject` line 1534) with no synchronization.

**2. Settings UI strings default to Portuguese.** The unsuffixed `$name`/`$description` in the settings block (e.g. `$name: Duracao da animacao (ms)`, lines 178, 191, …) are Portuguese, with `$name:en` as the override. The unsuffixed value is the fallback for every locale that isn't explicitly listed, so a user on, say, German or French sees Portuguese rather than English. The convention (which your `@name`/`@description` metadata already follows correctly) is English as the unsuffixed default, with `$name:pt`/`$description:pt` for Portuguese. This is pre-existing from 1.0, but worth fixing while you're touching the block.

<p>

- **`CoInitializeEx`/`CoUninitialize` imbalance in `TryHookBrowseObjectViaVtable`.** The failure path balances `S_FALSE` (line 1898: `if (hrCO == S_OK || hrCO == S_FALSE) CoUninitialize();`), but the success path only handles `S_OK` (line 1939: `if (hrCO == S_OK) CoUninitialize();`). When this runs on the Explorer UI thread (the `WM_WH_TRY_VTABLE_HOOK` retry), COM is already initialized so `CoInitializeEx` returns `S_FALSE` — and on a successful hook the matching `CoUninitialize` is skipped, leaking one COM init ref. Make the two paths consistent.
- **`g_s.alphaFade` is read without the lock** at line 1534, even though `g_s.duration` two lines later (1572) is read under `g_sLock`. Snapshot it under the lock too for consistency.
- **`EnumDisplaySettingsW` is called on every animation** (line 1082) just to get the refresh rate. The monitor's refresh rate rarely changes; caching it (per `HMONITOR`, invalidated on display-change) would avoid the per-navigation call.
- **Possibly-unused `-ldcomp`.** `DCompositionCreateDevice2` is resolved via `GetProcAddress` (line 1766), not called directly, and the `IDCompositionXxx` interfaces are pure COM — so `-ldcomp` in `@compilerOptions` is likely dead. Worth confirming and dropping if so.
- **Duplicate `#include <vector>`** (lines 274–275).

</p>
</details>

<p>

- **Rapid navigation in the same window can cut a fade short.** Cleanup items are keyed only by `hwndRoot` (`g_cleanupQueue` / `g_dcFades[c.hwndRoot]`, lines 826–847). If you navigate twice quickly, nav-1's cleanup timer fires while nav-2's overlay is still fading, looks up `g_dcFades[root]`, finds nav-2's *current* visual, and destroys it mid-animation → the fade pops/disappears. (No leak or double-free — refcounts stay balanced — purely visual.) Tag each `DCompFadeAnim` with a generation id and have the cleanup item store and verify that id before tearing down, so a stale timer only frees the instance it was scheduled for.
- **The fade snapshot is a raw screen grab** (`GetDC(nullptr)` + `BitBlt` at the content's screen rect, lines 1552–1568). If anything overlaps the content area at navigation time (a popup, tooltip, or another window), those foreign pixels get baked into the overlay. There's no clean alternative for capturing the DWM-composited Mica/Acrylic result, so this is just an FYI about the trade-off.
- **`CreateTargetForHwnd` on Explorer's own top-level window** (line 895, `topmost=FALSE`). A window can only have one composition target per `topmost` value; if a given build of Explorer already owns a non-topmost target on `CabinetWClass`, this would fail. The code degrades gracefully (no overlay), so it's not a crash risk, but worth keeping in mind if the fade silently stops working on some build.

</p>
</details>

crazyboyybs:
> ### Submission review
> 
> 
> Regarding `Wh_FindFirstSymbol` - I see that you're trying to match substrings. I don't think it's a good idea - if the function signature changes, the mod will likely crash.
> 
> Regarding the review below, 1 might not be important enough, up to you. Please address 2.
> 
> This is a large, careful rewrite — the subclass teardown (lock-copy-release + `SendMessageTimeoutW` with `SMTO_ABORTIFHUNG`), the long-lived cleanup worker, and the multi-layer symbol/vtable hooking are all done well. A few real issues to fix:
> 
> **1. Worker threads are detached and never joined on unload — UAF / crash window.** Two threads are spawned with `.detach()` and are not tracked or joined:
> 
> * The D3D11/DComp init thread in `Wh_ModInit` (the lambda ending at `.detach();`, line 1773). It runs `CreateDXGIFactory1` / `D3D11CreateDevice` / `DCompositionCreateDevice2`, which on a cold driver can take ~100–300 ms, and it never checks `g_unloading`. `Wh_ModBeforeUninit` joins only `g_cleanupThread` (line 2055), so if the mod is disabled (or reloaded) while this thread is still running, `Wh_ModUninit` releases and nulls `g_d3dDevice`/`g_dxgiDevice`/`g_dcDevice` (lines 2074–2076) while the thread is still writing them — leaking the freshly-created devices — and, after the DLL is freed, the thread is still executing code in unmapped pages → access violation. This window is narrow but absolutely reachable (toggle the mod off shortly after enabling, or a fast reload).
> * The per-animation threads in `DispatchAnimation` (line 1626). These at least poll `g_unloading` each frame, but `Wh_ModBeforeUninit` only does `Sleep(100)` (line 2049) and hopes they've exited. A descheduled/starved animation thread can still be inside `RunSlideIn` (calling `SetWindowPos`/`RestoreRgn`) when the DLL unloads.
> 
> You already do the right thing for the cleanup worker — store the `std::thread` in a global and `join()` it in `Wh_ModBeforeUninit`. Extend the same pattern to the init thread (store it, set `g_unloading` first, then `join()`), and replace the `Sleep(100)` heuristic with a real wait for the in-flight animations to drain (e.g. join tracked threads, or wait on a condition variable until `g_animating` is empty). While there, make the device pointers `std::atomic` (or guard them) — they're written by the init thread and read by the UI-thread hooks (`Classify` line 1273, `HookBrowseObject` line 1534) with no synchronization.
> 
> **2. Settings UI strings default to Portuguese.** The unsuffixed `$name`/`$description` in the settings block (e.g. `$name: Duracao da animacao (ms)`, lines 178, 191, …) are Portuguese, with `$name:en` as the override. The unsuffixed value is the fallback for every locale that isn't explicitly listed, so a user on, say, German or French sees Portuguese rather than English. The convention (which your `@name`/`@description` metadata already follows correctly) is English as the unsuffixed default, with `$name:pt`/`$description:pt` for Portuguese. This is pre-existing from 1.0, but worth fixing while you're touching the block.
> 
> Optional improvements
> 
> 
> * **`CoInitializeEx`/`CoUninitialize` imbalance in `TryHookBrowseObjectViaVtable`.** The failure path balances `S_FALSE` (line 1898: `if (hrCO == S_OK || hrCO == S_FALSE) CoUninitialize();`), but the success path only handles `S_OK` (line 1939: `if (hrCO == S_OK) CoUninitialize();`). When this runs on the Explorer UI thread (the `WM_WH_TRY_VTABLE_HOOK` retry), COM is already initialized so `CoInitializeEx` returns `S_FALSE` — and on a successful hook the matching `CoUninitialize` is skipped, leaking one COM init ref. Make the two paths consistent.
> * **`g_s.alphaFade` is read without the lock** at line 1534, even though `g_s.duration` two lines later (1572) is read under `g_sLock`. Snapshot it under the lock too for consistency.
> * **`EnumDisplaySettingsW` is called on every animation** (line 1082) just to get the refresh rate. The monitor's refresh rate rarely changes; caching it (per `HMONITOR`, invalidated on display-change) would avoid the per-navigation call.
> * **Possibly-unused `-ldcomp`.** `DCompositionCreateDevice2` is resolved via `GetProcAddress` (line 1766), not called directly, and the `IDCompositionXxx` interfaces are pure COM — so `-ldcomp` in `@compilerOptions` is likely dead. Worth confirming and dropping if so.
> * **Duplicate `#include <vector>`** (lines 274–275).
> 
> Functionality notes

The Wh_FindFirstSymbol substring matching was removed and English is now the default for settings. The worker threads UAF/crash window was fixed (init thread tracked + joined, animation drain via condition variable, g_dcDevice made atomic). All optional improvements from the review were applied (CoInitializeEx/CoUninitialize imbalance, g_s.alphaFade unprotected read, -ldcomp linker flag removal), besides the monitor refresh rate caching which was deemed unnecessary right now. Rapid navigation fade destruction is now prevented via a generation ID on each DComp fade instance. Additionally during review, some other things were fixed and optimized, RunSlideIn early exits now properly clean up g_animating to avoid orphaned entries, dead contentCls code was removed, and the symbol hooks variable was renamed to explorerframe_dll_hooks to satisfy the PR validator

m417z:
This pass surfaced a few things the earlier review missed. Please fix 1 and 4, and check 2. 3 seems less critical, up to you. The items in the collapsed sections are optional.

Also, regarding `Wh_ApplyHookOperations` usage: Why do you need to run `WindhawkUtils::HookSymbols` again? And why not run `TryHookBrowseObjectViaVtable` only in `Wh_ModInit`? These two should make `Wh_ApplyHookOperations` unnecessary.

---

**1. The subclass teardown never actually removes the subclass → crash in `explorer.exe` on unload.** You install with `WindhawkUtils::SetWindowSubclassFromAnyThread(hwnd, ExplorerSubclassProc, …)` (line 531, and `PropSheetSubclassProc` at line 521), but that helper does **not** install your proc directly — internally it installs `detail::SubclassProcWrapper` with `uIdSubclass = (UINT_PTR)pfnSubclass` (see `windhawk_utils.h`, `SubclassProcWrapper` / `SetWindowSubclass(hWnd, SubclassProcWrapper, (UINT_PTR)pfnSubclass, …)`). Your teardown then calls the raw form:

```cpp
RemoveWindowSubclass(hwnd,
    reinterpret_cast<SUBCLASSPROC>(ExplorerSubclassProc),
    APPCOMMAND_SUBCLASS_ID);
```

Neither argument matches the installed subclass: the proc is `SubclassProcWrapper` (not `ExplorerSubclassProc`), and the id is `(UINT_PTR)ExplorerSubclassProc` (not `APPCOMMAND_SUBCLASS_ID` — that value was passed as `dwRefData`, not as the id). So `RemoveWindowSubclass` returns `FALSE` and removes nothing. The same applies to all four call sites (lines 460, 472, 498, 506).

Consequence: `Wh_ModBeforeUninit` sends `WM_WH_REMOVE_SUBCLASS`, the wrapper forwards it to your proc, your proc no-ops, and the subclass stays installed. After the DLL unloads, the window's subclass chain still points at `SubclassProcWrapper` (which lives in your now-freed module) → the next message dispatched to any still-open Explorer window is an access violation that takes down `explorer.exe`. (Normal window close happens to be safe only because the wrapper auto-removes itself on `WM_NCDESTROY` — that path is what masks the bug during casual testing.)

Fix: use the matching helper, `WindhawkUtils::RemoveWindowSubclassFromAnyThread(hwnd, ExplorerSubclassProc)` (and `…PropSheetSubclassProc`). It works from any thread, so you can keep the `WM_WH_REMOVE_SUBCLASS` handler (call it inside, runs inline on the window thread) or call it directly from `Wh_ModBeforeUninit` and drop the custom message entirely. See [taskbar-empty-space-clicks.wh.cpp](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/taskbar-empty-space-clicks.wh.cpp#L4179) for the install/remove pairing.

**2. DirectComposition device is used concurrently from two threads with no serialization.** `StartDCompFade` runs on the Explorer UI thread (called from `HookBrowseObject`) and creates surfaces/visuals/targets and calls `Commit()` (lines 909–979), while `CleanupWorkerProc` runs on the dedicated worker thread and calls `SetRoot(nullptr)` / `Release()` / `Commit()` on the same device and its objects (lines 780–787, 844–851). The only locks held are `g_dcFadeMtx` / `g_cleanupMtx` around the *map/queue* — the actual DComp API calls happen **outside** any lock (e.g. the prev-fade cancel at 897–902 and the worker teardown at 844–851 both `Release`/`Commit` with no mutex). DirectComposition objects are not internally synchronized; the application must serialize all calls on a device and its objects. Under rapid navigation a new fade's setup can overlap a prior fade's teardown on the same device → corrupted composition state / crash. Wrap every DComp device/visual/target/surface call (create, `SetEffect`, `SetRoot`, `Commit`, `Release`) in a single dedicated mutex, or marshal teardown back to the thread that created the objects.

**3. The `ShowWindow`/`SetWindowPos` hooks hide the window but skip the restore when unloading races a navigation.** `HookShowWindow` applies `HideViaRgn` (empty window region) or `BeginFade` (`WS_EX_LAYERED`, alpha 1) *before* `g_origSW`, then calls `DispatchAnimation` (lines 1709–1723). If `g_unloading` is already set (it stays live for the ~2 s drain wait in `Wh_ModBeforeUninit` while the hooks are still installed), `DispatchAnimation` early-returns at line 1644 **without** ever starting `RunSlideIn` — so nothing restores the region/layered style. The content pane is left clipped to an empty region (invisible) or stuck near-transparent, and since the hook is gone afterwards it won't self-heal until the window is closed. Easiest fix: in `DispatchAnimation`'s `g_unloading` early-return, undo the hide first (`RestoreRgn(hwnd); if (info.useFade && \!info.useDCompFade) EndFade(hwnd);`), or bail at the very top of both hooks when `g_unloading` is set, before touching the window. Same applies to `HookSetWindowPos` (line 1742).

**4. `EnumWindows` in `Wh_ModAfterInit` (1987) has no process filter.** It enumerates every top-level window on the desktop; `CabinetWClass` windows owned by *other* `explorer.exe` instances get a cross-process `SetWindowSubclassFromAnyThread` attempt. Filter with `GetWindowThreadProcessId(hwnd, &pid); if (pid == GetCurrentProcessId())` to avoid this.

<p>

- **Animation threads are still detached + drained via a 2 s CV timeout** (`.detach()` line 1677; `g_animDoneCv.wait_for(… 2000ms …)` line 2078). The CV closes most of the window the earlier review raised, but two residuals remain: the predicate is signalled (`g_animating.erase` + `notify_all`) *before* the thread finishes unwinding, and a starved thread past the 2 s timeout lets unload proceed with the thread still live. Tracking the `std::thread`s and `join()`ing them (as you already do for the init/cleanup threads) closes it completely. The maintainer already accepted the CV approach for these, so this is purely belt-and-suspenders.
- **Threads are started before the hook registrations that can `return FALSE`.** `g_cleanupThread` (1768) and `g_initThread` (1775) are spawned, then any of the three `Wh_SetFunctionHook` calls (1841–1857) can `return FALSE` from `Wh_ModInit`. On that path Windhawk won't call `Wh_ModUninit`, so the two threads are orphaned. Register the hooks first, then start the threads (or set `g_unloading` + join on the failure path). Hook failure here is near-impossible, hence optional.
- **`g_pendingMtx` is held across the whole screen capture + `StartDCompFade`.** The lock taken at line 1542 stays held to line 1629, covering the `BitBlt` and the entire DComp overlay build. That mutex is also taken on the UI thread in `ConsumePendingTransition` and the `WM_APPCOMMAND` handler, so concurrent navigations in other Explorer windows serialize behind one window's fade setup. Write the pending-nav entry under the lock, release it, then do the capture/fade.
- **Raw `Wh_SetFunctionHook` + `reinterpret_cast<void*>` for the three user32 hooks** (1841–1857). `WindhawkUtils::SetFunctionHook(ShowWindow, HookShowWindow, &g_origSW)` is type-safe and catches signature mismatches at compile time.
- **Missing direct includes.** `std::round` (line 1199) needs `<cmath>` and `INT_MIN` (1170) needs `<climits>`; both currently rely on transitive includes.

</p>
</details>

<p>

- **`Wh_ApplyHookOperations` ("very slow", per the header) is called on the Explorer UI thread** in the vtable-fallback path (`WM_WH_TRY_VTABLE_HOOK` → line 454). It's guarded by `if (\!g_origBrowseObject)` so it runs at most once per window until the hook lands, but it's a one-time stall on that window's message loop. Fine as a fallback when the PDB isn't available; just noting the cost.
- **The CPU fade fallback is a fade-*in* of the new content** (alpha ramps 1→255, lines 1231–1238), whereas the DComp path fades the *old* content out. Behaviorally different effects depending on whether DComp initialized in time; both look reasonable, just not identical.

</p>
</details>

mods/file-explorer-content-animation.wh.cpp
```diff
@@ -1985,40 +1953,30 @@ void Wh_ModAfterInit() {
...
-                },
-            };
-            WindhawkUtils::HookSymbols(hEF, explorerframe_dll_hooks, ARRAYSIZE(explorerframe_dll_hooks));
-            if (g_origBrowseObject)
-                Wh_Log(L"[INIT] BrowseObject hooked OK via HookSymbols (AfterInit)");
-        }
-        // Camada 2: vtable (sem PDB, sempre funciona)
-        if (!g_origBrowseObject)
-            TryHookBrowseObjectViaVtable();
+        TryHookBrowseObjectViaVtable();
```
m417z: Why not run it in `Wh_ModInit`?

crazyboyybs: > Why not run it in `Wh_ModInit`?

TryHookBrowseObjectViaVtable needs to be in AfterInit because it depends on IShellWindows COM to get a real instance of CShellBrowser, during Explorer startup no CabinetWClass window exists in Wh_ModInit. So Wh_ApplyHookOperations remains necessary (hooks registered outside of Wh_ModInit are not applied automatically)

m417z: > during Explorer startup no CabinetWClass window exists in Wh_ModInit

Then it won't exist in `Wh_AfterModInit` too, unless I'm missing something. On startup, the original process doesn't start until `Wh_AfterModInit` returns, see:
https://github.com/ramensoftware/windhawk/wiki/Mod-lifetime

crazyboyybs: > > during Explorer startup no CabinetWClass window exists in Wh_ModInit
> 
> Then it won't exist in `Wh_AfterModInit` too, unless I'm missing something. On startup, the original process doesn't start until `Wh_AfterModInit` returns, see: https://github.com/ramensoftware/windhawk/wiki/Mod-lifetime

You're right, I was wrong about AfterInit having windows available. Looking at the mod lifetime flowchart, the original process doesn't start until Wh_ModAfterInit returns, so no CabinetWClass windows exist there either during startup.
TryHookBrowseObjectViaVtable is now in Wh_ModInit along with HookSymbols. Both are tried there, and hooks registered in Wh_ModInit are applied implicitly. The Wh_ModAfterInit hook attempt block was removed entirely.
For the startup case where both fail (no explorerframe.dll loaded, no windows), the deferred WM_WH_TRY_VTABLE_HOOK path handles it when the first CabinetWClass appears via the CreateWindowExW hook , that's the only remaining call site for Wh_ApplyHookOperations

m417z:
Here's an additional round. Sorry for the multiple reviews. 1 is a fresh lesson from https://github.com/ramensoftware/windhawk-mods/issues/4260. 2 was missed, but can be a real issue.

---

**1. Global `std::thread` objects will crash Explorer on every shutdown/restart.**

`g_cleanupThread` and `g_initThread` are global `std::thread` objects, joined only in `Wh_ModBeforeUninit`:

```cpp
static std::thread g_cleanupThread;
static std::thread g_initThread;     // D3D/DComp init em background
```

`Wh_ModBeforeUninit`/`Wh_ModUninit` run on a normal mod unload, but **not** when the host process itself terminates (Explorer restart, sign-out, reboot). On process termination the OS detaches the DLL and the CRT runs the destructors of global objects during `DLL_PROCESS_DETACH`. Neither thread has been joined or detached at that point, so both are still `joinable()`, and `~std::thread()` on a joinable thread calls `std::terminate()` → the host aborts. This fires on *every* Explorer exit, not just mod unload.

Fix: make them heap pointers (trivial destructor → nothing runs at `DLL_PROCESS_DETACH`), and keep the real join+delete in `Wh_ModBeforeUninit`/`Wh_ModUninit`. By the time the process is terminating, the OS has already killed the worker threads, so there's nothing left to clean up. See [`taskbar-music-lounge.wh.cpp`](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/taskbar-music-lounge.wh.cpp#L809) for the exact pattern:

```cpp
static std::thread* g_cleanupThread = nullptr;
static std::thread* g_initThread    = nullptr;
...
g_cleanupThread = new std::thread(CleanupWorkerProc);
g_initThread    = new std::thread([]{ ... });
...
// Wh_ModBeforeUninit:
if (g_initThread && g_initThread->joinable()) g_initThread->join();
...
if (g_cleanupThread && g_cleanupThread->joinable()) g_cleanupThread->join();
```

(You can `delete` them after join, or just leak the now-finished thread objects — either is fine; the point is the *global* must not have a non-trivial destructor.)

**2. Custom messages in the `WM_USER` range are sent to Explorer's own windows.**

```cpp
static constexpr UINT WM_WH_REMOVE_SUBCLASS  = WM_USER + 0x5A;
static constexpr UINT WM_WH_TRY_VTABLE_HOOK  = WM_USER + 0x5B;
```

These are posted/sent to `CabinetWClass` windows (which belong to Explorer, not to the mod), and `ExplorerSubclassProc` handles both and returns `0` **without** calling `DefSubclassProc`. `WM_USER+N` values mean different things to different window classes — if `CabinetWClass` (or another mod / a future shell build) uses `WM_USER+0x5A`/`0x5B` internally, this silently swallows it. Register a unique message instead:

```cpp
static UINT g_msgRemoveSubclass;  // RegisterWindowMessage(L"...")
static UINT g_msgTryVtableHook;
// in Wh_ModInit:
g_msgRemoveSubclass = RegisterWindowMessageW(L"Windhawk_FECA_RemoveSubclass");
g_msgTryVtableHook  = RegisterWindowMessageW(L"Windhawk_FECA_TryVtableHook");
```

`RegisterWindowMessage` returns a process-wide-unique value in the `0xC000`–`0xFFFF` range that can never collide with a class's own `WM_USER` messages. (`WM_APPCOMMAND` handling is fine — that one correctly falls through to `DefSubclassProc`.)

<p>

- **Unload-time region leak.** In `HookShowWindow`/`HookSetWindowPos` you call `HideViaRgn(hwnd)` and then `DispatchAnimation`, but `DispatchAnimation` early-returns without restoring the region if `g_unloading` became true in the gap (it was only checked earlier in `HookShowWindow`). The window would be left with an empty region (invisible). It only happens during the unload race, but restoring the RGN on that early-return path would close it.

- **`g_origBrowseObject` is a plain pointer** read/written across threads (UI thread in `HookBrowseObject`, init thread / subclass proc when installing). Aligned pointer access is atomic in practice, but `std::atomic<BrowseObject_t>` makes the intent explicit and silences TSan-style complaints.

- **Stale comments.** A few comments still reference the old `Wh_ModAfterInit` flow now that the hook install moved into `Wh_ModInit` — e.g. `TryHookBrowseObjectViaVtable`'s `// Wh_ModAfterInit roda no thread do Windhawk` and the header block that says it's "Chamado de: Wh_ModInit ... / WM_WH_TRY_VTABLE_HOOK". Worth a quick pass so the comments match the code.

</p>
</details>

<p>

Non-critical observations about the feature behavior itself — no action required.

- **Screen capture for the fade reads composited screen pixels.** `GetDC(nullptr)` + `BitBlt` at the content's screen rect grabs whatever is on-screen there, so an overlapping window, or a partially off-screen / occluded Explorer window, will bleed into the fade frame. This looks deliberate — it's what lets you capture the live Mica/Acrylic backdrop, which `PrintWindow` wouldn't — so there's a real trade-off and probably no cleaner alternative. Just flagging the edge case.

- **`HookBrowseObject` does its COM walk + `BitBlt` synchronously on the UI thread** before each navigation (when fade is on). It's once-per-navigation, not per-frame, but it does add a little latency to the navigation itself. Inherent to capturing the *previous* content before it's torn down, so noting it rather than asking for a change.

- **Global `ShowWindow`/`SetWindowPos`/`CreateWindowExW` hooks** mean every such call in `explorer.exe` passes through your hook. The early-outs (`IsWindowVisible`, the `g_in*` guards) keep the common path cheap, and this is intrinsic to catching windows as they first appear, so there's no obvious better injection point — just be aware of the blast radius if any of those hooks ever misbehaves.

</p>
</details>

crazyboyybs:
> Here's an additional round. Sorry for the multiple reviews. 1 is a fresh lesson from #4260. 2 was missed, but can be a real issue.
> 
> **1. Global `std::thread` objects will crash Explorer on every shutdown/restart.**
> 
> `g_cleanupThread` and `g_initThread` are global `std::thread` objects, joined only in `Wh_ModBeforeUninit`:
> 
> ```c++
> static std::thread g_cleanupThread;
> static std::thread g_initThread;     // D3D/DComp init em background
> ```
> 
> `Wh_ModBeforeUninit`/`Wh_ModUninit` run on a normal mod unload, but **not** when the host process itself terminates (Explorer restart, sign-out, reboot). On process termination the OS detaches the DLL and the CRT runs the destructors of global objects during `DLL_PROCESS_DETACH`. Neither thread has been joined or detached at that point, so both are still `joinable()`, and `~std::thread()` on a joinable thread calls `std::terminate()` → the host aborts. This fires on _every_ Explorer exit, not just mod unload.
> 
> Fix: make them heap pointers (trivial destructor → nothing runs at `DLL_PROCESS_DETACH`), and keep the real join+delete in `Wh_ModBeforeUninit`/`Wh_ModUninit`. By the time the process is terminating, the OS has already killed the worker threads, so there's nothing left to clean up. See [`taskbar-music-lounge.wh.cpp`](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/taskbar-music-lounge.wh.cpp#L809) for the exact pattern:
> 
> ```c++
> static std::thread* g_cleanupThread = nullptr;
> static std::thread* g_initThread    = nullptr;
> ...
> g_cleanupThread = new std::thread(CleanupWorkerProc);
> g_initThread    = new std::thread([]{ ... });
> ...
> // Wh_ModBeforeUninit:
> if (g_initThread && g_initThread->joinable()) g_initThread->join();
> ...
> if (g_cleanupThread && g_cleanupThread->joinable()) g_cleanupThread->join();
> ```
> 
> (You can `delete` them after join, or just leak the now-finished thread objects — either is fine; the point is the _global_ must not have a non-trivial destructor.)
> 
> **2. Custom messages in the `WM_USER` range are sent to Explorer's own windows.**
> 
> ```c++
> static constexpr UINT WM_WH_REMOVE_SUBCLASS  = WM_USER + 0x5A;
> static constexpr UINT WM_WH_TRY_VTABLE_HOOK  = WM_USER + 0x5B;
> ```
> 
> These are posted/sent to `CabinetWClass` windows (which belong to Explorer, not to the mod), and `ExplorerSubclassProc` handles both and returns `0` **without** calling `DefSubclassProc`. `WM_USER+N` values mean different things to different window classes — if `CabinetWClass` (or another mod / a future shell build) uses `WM_USER+0x5A`/`0x5B` internally, this silently swallows it. Register a unique message instead:
> 
> ```c++
> static UINT g_msgRemoveSubclass;  // RegisterWindowMessage(L"...")
> static UINT g_msgTryVtableHook;
> // in Wh_ModInit:
> g_msgRemoveSubclass = RegisterWindowMessageW(L"Windhawk_FECA_RemoveSubclass");
> g_msgTryVtableHook  = RegisterWindowMessageW(L"Windhawk_FECA_TryVtableHook");
> ```
> 
> `RegisterWindowMessage` returns a process-wide-unique value in the `0xC000`–`0xFFFF` range that can never collide with a class's own `WM_USER` messages. (`WM_APPCOMMAND` handling is fine — that one correctly falls through to `DefSubclassProc`.)
> 
> Optional improvements
> 
> 
> * **Unload-time region leak.** In `HookShowWindow`/`HookSetWindowPos` you call `HideViaRgn(hwnd)` and then `DispatchAnimation`, but `DispatchAnimation` early-returns without restoring the region if `g_unloading` became true in the gap (it was only checked earlier in `HookShowWindow`). The window would be left with an empty region (invisible). It only happens during the unload race, but restoring the RGN on that early-return path would close it.
> * **`g_origBrowseObject` is a plain pointer** read/written across threads (UI thread in `HookBrowseObject`, init thread / subclass proc when installing). Aligned pointer access is atomic in practice, but `std::atomic<BrowseObject_t>` makes the intent explicit and silences TSan-style complaints.
> * **Stale comments.** A few comments still reference the old `Wh_ModAfterInit` flow now that the hook install moved into `Wh_ModInit` — e.g. `TryHookBrowseObjectViaVtable`'s `// Wh_ModAfterInit roda no thread do Windhawk` and the header block that says it's "Chamado de: Wh_ModInit ... / WM_WH_TRY_VTABLE_HOOK". Worth a quick pass so the comments match the code.
> 
> Functionality notes

Both global std::thread objects are now heap-allocated (std::thread*) to prevent std::terminate() during DLL_PROCESS_DETACH on Explorer shutdown/restart. They're joined and deleted in Wh_ModBeforeUninit as before, on process termination the OS has already killed the threads, so the leaked pointer is harmless.
The WM_USER+0x5A/0x5B custom messages were replaced with RegisterWindowMessageW which returns process-unique values in the 0xC000-0xFFFF range that cannot collide with CabinetWClass's own messages.
The unload-time region leak was already partially addressed in the previous round (g_unloading check at the top of both hooks, before HideViaRgn/BeginFade), but there was still a TOCTOU gap: if g_unloading becomes true after the hook's top check but before DispatchAnimation's entry check, the window already had HideViaRgn applied and the bare return didn't restore it. DispatchAnimation's g_unloading early-return now calls RestoreRgn + EndFade before returning, closing the gap completely.
The g_origBrowseObject atomic change was not applied, making it std::atomic would require restructuring how HookSymbols and Wh_SetFunctionHook write to it via void**, and there's no actual data race since all writes complete before hooks become active via Wh_ApplyHookOperations. Stale comments referencing the old Wh_ModAfterInit flow were updated

emvaized:
Thank you very much, this mod is wonderful! One of favourite Windhawk mods so far.

emvaized:
@crazyboyybs 
I have a small feature request for this wonderful mod: there is an option for Page transitions, which automatically detects the navigation type and applies different animations. 

It works great, but it only allows to see left/right transitions — would be great to also have an option for vertical up/down transitions!

SLITXX:
- Translucent Windows v1.7.4（今天更新后出现冲突）
  - 文件内容区域点击导航时闪烁
  - 关掉 Fade 选项后问题消失
  - 两个 mod 此前共存正常

PR #4255, opened 2026-06-01
Transparent Desktop Icons with Spotlight

drgutman:
Make desktop icons transparent when idle with an interactive spotlight effect. Hover over the desktop or select icons to reveal them with customizable transitions.

## Changelog

* Changelog item 1...
* Changelog item 2...

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [x] Gemini 3.1 Pro on aistudio

### DIFF
--- a/mods/transparent-desktop-icons-spotlight.wh.cpp
+++ b/mods/transparent-desktop-icons-spotlight.wh.cpp
@@ -29,7 +29,7 @@ customizable independent timeouts, blurs, and fade durations.
 * Independent Dual-GPU Blur processing for both the cursor spotlight and selected icons.
 * Interactive morphing shapes (square to circle) for both the spotlight and selected areas.
 * Persistent selection memory allows deselected icons to gracefully fade away.
-* Click-through architecture using WS_DISABLED ensures no interference with native OS interaction.
+* Click-through architecture using WS_EX_TRANSPARENT ensures no interference with native OS interaction.
 */
 // ==/WindhawkModReadme==
 
@@ -37,7 +37,7 @@ customizable independent timeouts, blurs, and fade durations.
 /*
 - idleOpacity: 128
   $name: Idle Opacity
-  $description: Opacity of overlay when idle (0 = fully opaque, 255 = fully transparent)
+  $description: How much the icons are dimmed when idle (0 = icons fully visible, 255 = icons fully hidden)
 - spotlightRadius: 200
   $name: Spotlight Radius
   $description: Size of the spotlight in pixels
@@ -50,9 +50,12 @@ customizable independent timeouts, blurs, and fade durations.
 - idleDelay: 2000
   $name: Spotlight Idle Timeout (ms)
   $description: Time in milliseconds before the spotlight begins to fade out
-- spotlightFadeDuration: 500
-  $name: Spotlight Fade Duration (ms)
-  $description: How long it takes for the spotlight to fade in and out
+- spotlightFadeInDuration: 500
+  $name: Spotlight Fade In Duration (ms)
+  $description: How long it takes for the spotlight to fade in when mouse enters
+- spotlightFadeOutDuration: 500
+  $name: Spotlight Fade Out Duration (ms)
+  $description: How long it takes for the spotlight to fade out when mouse leaves
 - mouseSmoothing: 10
   $name: Mouse Smoothing
   $description: Smoothing factor for spotlight following (0-20, 0 = no smoothing, 20 = very smooth)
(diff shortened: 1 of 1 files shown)


drgutman:
I switched to Work in progress because I found some bugs and it might take a while until I fix them.

m417z:
### Submission review

---

**1. Global `std::thread g_renderThread` aborts the host on Explorer restart / process exit.** `Wh_ModUninit` is **not** called when the host process terminates (Explorer restart, sign-out, reboot) — the OS just detaches the DLL and the CRT runs global destructors during `DLL_PROCESS_DETACH`. At that point `g_renderThread` is still `joinable()` (its underlying thread was killed by the OS, but the `std::thread` object doesn't know that), so `~thread()` calls `std::terminate()` and `explorer.exe` aborts. Since Explorer restarts are routine, users would hit this regularly. The fix is to give the global a trivial destructor — use a heap pointer (or a raw `HANDLE`, which is what the reference mod does for its [worker thread](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/desktop-live-overlay.wh.cpp#L879)), and keep the real join in `WM_DESTROY`:

```cpp
std::thread* g_renderThread = nullptr;
...
g_renderThread = new std::thread(RenderLoop);
...
// WM_DESTROY:
g_threadStop = true;
if (g_renderThread) {
    if (g_renderThread->joinable()) g_renderThread->join();
    delete g_renderThread;
    g_renderThread = nullptr;
}
```

At process exit the pointer is simply leaked (the OS has already terminated the thread), so nothing calls `std::terminate()`.

**2. README has no screenshot/GIF.** This is a highly visual effect (transparency + an interactive spotlight), and the README has no image. A screenshot or short GIF makes a big difference for users browsing the catalog. Only `i.imgur.com` and `raw.githubusercontent.com` are allowed image hosts.

**3. Unregister the window classes in `Wh_ModUninit`.** `RegisterOverlayWindowClass`/`RegisterMessageWindowClass` never `UnregisterClass`. If the mod is later reloaded into the same DLL base, the `RegisterClass` can fail with `ERROR_CLASS_ALREADY_EXISTS`, leaving `registered == false` and the overlay silently never created.

<p>

- **Remove unused libraries and includes.** The `@compilerOptions` line is copied verbatim from Desktop Live Overlay, which links those libs for weather/system-metrics features this mod doesn't have. None of `dwrite`, `dwmapi`, `wininet`, `pdh`, `powrprof`, `shlwapi` are used (I checked — no `DWrite*`/`Dwm*`/`InternetOpen`/`Pdh*`/`*PowerInformation`/`Path*`/`Str*` symbols anywhere). Trim to what you actually use, and drop the unused `#include <dwrite.h>` / `#include <dwmapi.h>`. (`-lwininet` is especially worth removing — the mod correctly contacts nothing external, so it shouldn't advertise otherwise.)
- **Use `WindhawkUtils::SetFunctionHook()`** instead of raw `Wh_SetFunctionHook` with `(void*)`/`(void**)` casts — it's type-safe and verifies the hook signature against the target.
- **`Wh_ModSettingsChanged`:** since you never request a reload (`*bReload = FALSE` always), the simpler `void Wh_ModSettingsChanged()` form is equivalent and clearer. Also, consider calling `LoadSettings()` directly there instead of routing through `WM_APP_SETTINGS_CHANGED` — `LoadSettings()` already takes `g_renderMutex` so it's thread-safe from any thread, and the current routing silently drops a settings change made before the overlay window exists (the `PostMessage` is skipped when `g_overlayWnd` is null).
- **Dead variable:** `g_hasSelection` is written (set `false` in a couple of places) but never read.
- The one-shot `SetTimer(nullptr, 0, 1000, ...)` in `CreateWindowExW_Hook` schedules a callback into mod code 1s later; if the mod is unloaded within that window the callback runs in freed memory. Narrow, but worth a guard or a different bootstrap.

</p>
</details>

<p>

- **Idle-skip still renders when the spotlight is invisible.** `stateChanged` triggers a render whenever `g_smoothedMousePos` moves >0.1px — but once `g_spotlightFade` has faded to 0, mouse movement no longer changes the output (no hole is drawn). So moving the cursor anywhere (e.g. inside a fullscreen app, with the overlay hidden behind it) still drives 60fps full-screen wallpaper redraws + `Present`. Gating the mouse-moved term on `g_spotlightFade > 0.001f` would make it genuinely idle and match the "0% GPU while idle" goal.
- **Multi-monitor / mixed-DPI.** The overlay is one window sized to a single desktop-parent rect, and `g_dpiScale` is a single value from one monitor. On a mixed-DPI multi-monitor setup the spotlight radius/blur and selection padding will be scaled by the wrong DPI on secondary monitors, and depending on the extent of the parent window the effect may not cover them at all. The reference mod handles this with `EnumDisplayMonitors` + virtual-screen coordinates (`SM_XVIRTUALSCREEN`); worth verifying on your own multi-monitor layout.
- **Static wallpaper only.** `PaintDesktop` captures the static wallpaper bitmap, so animated/live wallpapers (Wallpaper Engine, Lively, etc.) will show a frozen frame in the dimmed overlay that doesn't match what's actually behind the icons.
- **Continuous polling.** While enabled, the render thread wakes ~62×/s (`GetCursorPos`) and the UI timer polls every 50ms regardless of whether the desktop is even visible. It's cheap and acceptable, but could be paused when the desktop isn't the foreground.

</p>
</details>

drgutman:
Thank you for the review. I'll start implementing the fixes right away. Switched back to work in progress in the meantime, so it doesn't show as waiting for approval for nothing.

m417z:
Thanks.

One last fixup: remove these lines, they may cause instability if the class is reused from a previous instance of the mod:
`if (GetClassInfo(GetCurrentModuleHandle(), OVERLAY_WINDOW_CLASS, &wc)) return true;`
and:
`if (GetClassInfo(GetCurrentModuleHandle(), MESSAGE_WINDOW_CLASS, &wc)) return true;`

PR #4266, opened 2026-06-02
Snap Commander mod

Asteski:
## Changelog

* New mod submission: Snap Commander
* Move the active window to screen halves and corners with keyboard shortcuts, with preconfigured hotkeys ready to go.
* Inspired by Rectangle app from macOS, allows user to assign separate hotkey to every action, including global modifier key.
* Includes window gaps option and auto-snap rules, which forcibly triggers specific action on a newly opened window/app.
* Different vision for managing windows than Tiling Helper mod - more control over windows placement, although not an automated one like abovementioned.

## Mod authorship

This mod was created by:

- [x] Claude
- [ ] ChatGPT
- [ ] Gemini
- [ ] Other (please specify): 

### DIFF
--- a/mods/snap-commander.wh.cpp
+++ b/mods/snap-commander.wh.cpp
@@ -120,7 +120,7 @@ In order to use this mod, you must allow Windhawk to inject into the **dwm.exe**
 system process. To do so, add it to the process inclusion list in the advanced
 settings. If you do not do this, it will silently fail to inject.
 
-![Advanced settings screenshot](https://i.imgur.com/LRhREtJ.png
+![Advanced settings screenshot](https://i.imgur.com/LRhREtJ.png)
 
 */
 // ==/WindhawkModReadme==
--- a/mods/snap-commander.wh.cpp
+++ b/mods/snap-commander.wh.cpp
@@ -16,6 +16,8 @@
 Move the active window to specific screen areas using keyboard shortcuts, just like
 the [Rectangle](https://rectangleapp.com/) app on macOS.
 
+![snap-commander](https://raw.githubusercontent.com/Asteski/Windhawk-Mods/refs/heads/main/img/snap-commander/snap-commander.gif)
+
 ## Default shortcuts
 
 | Action          | Shortcut    |
--- a/mods/snap-commander.wh.cpp
+++ b/mods/snap-commander.wh.cpp
@@ -113,6 +113,14 @@ Enable a rule's **Don't resize** toggle to move the window into the chosen regio
 keeping its current size ÔÇö the unresized window is anchored to the matching corner or
 side of the region (e.g. *top-right* pins it to the region's top-right). This is ignored
 by the Minimize action.
+
+## ⚠ Important usage note ⚠
+
+In order to use this mod, you must allow Windhawk to inject into the **dwm.exe**
+system process. To do so, add it to the process inclusion list in the advanced
+settings. If you do not do this, it will silently fail to inject.
+
+![Advanced settings screenshot](https://i.imgur.com/LRhREtJ.png)
 */
 // ==/WindhawkModReadme==
 
--- a/mods/snap-commander.wh.cpp
+++ b/mods/snap-commander.wh.cpp
@@ -116,11 +116,17 @@ by the Minimize action.
 
 ## ⚠ Important usage note ⚠
 
-In order to use this mod with elevated processes, you must allow Windhawk to inject 
-into the **dwm.exe** system process and replace windhawk.exe as @included with dwm.exe.
-To do so, add it to the process inclusion list in the advanced settings, and modify 
-the mod code by forking it, and replacing "windhawk.exe" with "dwm.exe" in line 8.
-If you do not do this, it will silently fail to inject to elevated processes.
+In order to use this mod with elevated processes, it can be configured to target
+`dwm.exe`.
+
+First, navigate to Windhawk's Settings > Advanced settings > More advanced
+settings > Process inclusion list, and make sure that `dwm.exe` is in the list.
+
+![Advanced settings screenshot](https://i.imgur.com/LRhREtJ.png)
+
+Next, go to the mod's advanced tab, enable this option: "Ignore mod
+inclusion/exclusion lists", then add `dwm.exe` to the custom process inclusion
+list.
 
 */
 // ==/WindhawkModReadme==
--- a/mods/snap-commander.wh.cpp
+++ b/mods/snap-commander.wh.cpp
@@ -119,7 +119,7 @@ by the Minimize action.
 In order to use this mod with elevated processes, you must allow Windhawk to inject 
 into the **dwm.exe** system process and replace windhawk.exe as @included with dwm.exe.
 To do so, add it to the process inclusion list in the advanced settings, and modify 
-the mod code by forking it, and replacing "@include windhawk.exe" with "@include dwm.exe"
+the mod code by forking it, and replacing "windhawk.exe" with "dwm.exe" in line 8.
 If you do not do this, it will silently fail to inject to elevated processes.
 
 */
--- a/mods/snap-commander.wh.cpp
+++ b/mods/snap-commander.wh.cpp
@@ -113,6 +113,15 @@ Enable a rule's **Don't resize** toggle to move the window into the chosen regio
 keeping its current size - the unresized window is anchored to the matching corner or
 side of the region (e.g. *top-right* pins it to the region's top-right). This is ignored
 by the Minimize action.
+
+## ⚠ Important usage note ⚠
+
+In order to use this mod with elevated processes, you must allow Windhawk to inject 
+into the **dwm.exe** system process and replace windhawk.exe as @included with dwm.exe.
+To do so, add it to the process inclusion list in the advanced settings, and modify 
+the mod code by forking the mod. If you do not do this, it will silently fail 
+to inject to elevated processes.
+
 */
 // ==/WindhawkModReadme==
 
--- a/mods/snap-commander.wh.cpp
+++ b/mods/snap-commander.wh.cpp
@@ -119,8 +119,8 @@ by the Minimize action.
 In order to use this mod with elevated processes, you must allow Windhawk to inject 
 into the **dwm.exe** system process and replace windhawk.exe as @included with dwm.exe.
 To do so, add it to the process inclusion list in the advanced settings, and modify 
-the mod code by forking the mod. If you do not do this, it will silently fail 
-to inject to elevated processes.
+the mod code by forking it, and replacing "@include windhawk.exe" with "@include dwm.exe"
+If you do not do this, it will silently fail to inject to elevated processes.
 
 */
 // ==/WindhawkModReadme==
--- a/mods/snap-commander.wh.cpp
+++ b/mods/snap-commander.wh.cpp
@@ -2,7 +2,7 @@
 // @id              snap-commander
 // @name            Snap Commander
 // @description     Move the active window to screen halves and corners with keyboard shortcuts (inspired by Rectangle on macOS)
-// @version         1.0
+// @version         1.0.0
 // @author          Asteski
 // @github          https://github.com/Asteski
 // @include         dwm.exe
@@ -120,7 +120,8 @@ In order to use this mod, you must allow Windhawk to inject into the **dwm.exe**
 system process. To do so, add it to the process inclusion list in the advanced
 settings. If you do not do this, it will silently fail to inject.
 
-![Advanced settings screenshot](https://i.imgur.com/LRhREtJ.png)
+![Advanced settings screenshot](https://i.imgur.com/LRhREtJ.png
+
 */
 // ==/WindhawkModReadme==
 
--- a/mods/snap-commander.wh.cpp
+++ b/mods/snap-commander.wh.cpp
@@ -5,7 +5,7 @@
 // @version         1.0.0
 // @author          Asteski
 // @github          https://github.com/Asteski
-// @include         dwm.exe
+// @include         windhawk.exe
 // @compilerOptions -luser32 -ldwmapi
 // ==/WindhawkMod==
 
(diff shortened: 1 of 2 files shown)


m417z:
### Submission review

---

Nice, well-structured mod — the DWM frame-bounds handling, gap modes, and multi-monitor mapping are all thoughtful. The main blocker is that it declares itself a tool mod but isn't wired up as one; there's also a keyboard side effect worth fixing.

**1. `@include windhawk.exe` but the tool-mod boilerplate is missing — the mod runs inside Windhawk's own process(es).** The mod sets `@include windhawk.exe`, which is the "mods as tools" target, and it's a textbook fit (only `SetWindowsHookEx` / `SetWinEventHook` / `EnumWindows` / `SetWindowPos` — all system-wide APIs, no function hooks). But it uses plain `Wh_ModInit` / `Wh_ModSettingsChanged` / `Wh_ModUninit` and omits the entire launcher snippet (no `Wh_ModAfterInit`, no `-tool-mod` command-line dispatch, no single-instance mutex, no entry-point hook). As written it does **not** spawn a dedicated process — its DLL is injected directly into the live `windhawk.exe` process(es). Consequences: the global LL keyboard hook + WinEvent hook get installed inside the Windhawk UI/host process (and, without the session-0 guard, potentially the service process), a crash in the mod destabilizes Windhawk itself, there's no protection against multiple instances, and your home-grown thread/lifecycle code fights the framework's model.

   Fix: adopt the full boilerplate from the [Mods as tools wiki page](https://github.com/ramensoftware/windhawk/wiki/Mods-as-tools:-Running-mods-in-a-dedicated-process). Rename `Wh_ModInit` → `WhTool_ModInit`, `Wh_ModSettingsChanged` → `WhTool_ModSettingsChanged`, `Wh_ModUninit` → `WhTool_ModUninit`, and paste the launcher block verbatim.

**2. Global `WH_KEYBOARD_LL` hook leaks the modifier through to the foreground app — consider `RegisterHotKey` instead.** The hook swallows only the *letter* keydown (returns `1` at line 1096) while the held modifier's keydown/keyup still reach the system. With the default `Modifier = Alt`, that means each shortcut leaves Alt "tapped alone" from the foreground app's point of view, so `DefWindowProc` activates its menu bar after every press (Explorer, Notepad, and most classic Win32 apps); with `Modifier = Win`, releasing Win alone opens the Start menu. This affects the out-of-the-box experience.

   The two closest existing mods for this exact job — [move-window-to-monitor.wh.cpp](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/move-window-to-monitor.wh.cpp#L313) and [keyboard-shortcut-actions.wh.cpp](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/keyboard-shortcut-actions.wh.cpp#L700) — both use `RegisterHotKey`, which consumes the whole combo cleanly (no menu/Start side effect), keeps you out of the per-keystroke global input path, and is the idiomatic approach. It supports all your cases (modifier combos, `Win`, bare keys with `fsModifiers = 0`, exact-modifier semantics). The tradeoffs are real but minor: a combo can have only one global owner (log + skip if `RegisterHotKey` fails), and a few reserved combos (e.g. `Win`+`L`) can't be claimed. The auto-snap `SetWinEventHook` path is unaffected and stays as-is. If you keep the LL hook for a deliberate reason, please at least document the menu/Start-activation behavior in the README.

**3. README has no visual.** This is a very visible mod (windows snapping to regions), so a short screenshot or GIF in the README would help users understand it at a glance. Only `i.imgur.com` and `raw.githubusercontent.com` are allowed image hosts.

<p>

- **`"Rectangle:"` log prefix is a leftover.** Every `Wh_Log` call is prefixed with `Rectangle:` (e.g. lines 1258, 1345, 1350), but Windhawk already prefixes log lines with the mod name, and the mod is named "Window Manager" — so the prefix is both redundant and inconsistent. Drop it.
- **Unnecessary `NULL` checks / manual frees in `LoadSettings`.** `Wh_GetStringSetting` never returns `NULL` (it returns `L""` when unset), so the `modifier ? modifier : L"alt"` guards aren't needed. More importantly, the repeated `Wh_GetStringSetting` + `Wh_FreeStringSetting` pairs can be replaced with `WindhawkUtils::StringSetting` (RAII), which removes the manual free on every path and tidies the function considerably.
- **`g_modifierMask` can be a local.** It's only consumed inside `LoadSettings` (folded into `g_actionMods` at line 1237); it doesn't need to be a file-scope global.
- **`g_savedPlacements` is never pruned when a window is destroyed.** Entries are added on first snap (line 899) and removed only on *Restore* (line 909). `WinEventProc` already gets `EVENT_OBJECT_DESTROY` but only cleans `g_ruleHandledWindows`. Over a long session the map slowly grows, and—because Windows recycles `HWND` values—a recycled handle could let *Restore* move a brand-new window to a dead window's old placement. Consider enqueuing a cleanup work item on destroy so the worker thread (the sole owner of the map) can erase it.
- **Benign data race on the settings globals.** `g_actionKeys`/`g_actionMods` are written by `LoadSettings` (settings/init thread) and read by `MatchAction` on the keyboard-hook thread; the gap/center/resize globals are likewise read on the worker thread. It's harmless in practice (aligned scalars, only racing during a settings change) but technically a data race, and a settings change mid-keypress could momentarily match/swallow the wrong key. Not worth a lock, just noting it.

</p>
</details>

<p>

- **Distinct enough from existing mods.** It overlaps in spots with [tiling-helper](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/tiling-helper.wh.cpp) (automated tiling), [move-window-to-monitor](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/move-window-to-monitor.wh.cpp) (next/prev display), and [edge-snap-throw](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/edge-snap-throw.wh.cpp) / [two-sided-snapping](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/two-sided-snapping.wh.cpp), but the manual Rectangle-style keyboard snapper + auto-snap rules combination is a different enough vision. Noting for completeness; the PR already distinguishes it from Tiling Helper.
- **Only keydown is swallowed.** `WM_KEYUP`/`WM_SYSKEYUP` for the bound key still pass through (the hook matches only `WM_KEYDOWN`/`WM_SYSKEYDOWN`). Usually harmless, just be aware apps may see a key-up without a matching key-down.
- **Hard-coded pixels aren't DPI-scaled.** The 200×150 minimum in `ResizeFromCenter` (lines 877-878), and the gap/center-size pixel values, are raw device pixels rather than scaled per-monitor. Reasonable as-is, but on very high-DPI monitors the minimum is tiny — FYI.

</p>
</details>

m417z:
### Submission review

Also note that the readme seems to have some broken text, for example:
```
- `q` ÔÇö just the `Q` key
- `ctrl+q` ÔÇö `Ctrl` + `Q`
- `ctrl+shift+left` ÔÇö `Ctrl` + `Shift` + `Left arrow`
```

---

**Make this a tool mod instead of injecting into `dwm.exe`.** The mod has **no function hooks** at all — it only uses system-wide APIs that work from any process: `RegisterHotKey`, `GetForegroundWindow`, `SetWinEventHook`, `EnumWindows`, and `SetWindowPos`/`ShowWindow`/`SetWindowPlacement`. That's the textbook case for the ["Mods as tools: running mods in a dedicated process"](https://github.com/ramensoftware/windhawk/wiki/Mods-as-tools:-Running-mods-in-a-dedicated-process) pattern. Injecting into `dwm.exe` is the wrong host here:

- `dwm.exe` is the desktop compositor — a crash or hang in the mod (e.g. a `SetWindowPos` that blocks on an unresponsive target window) can take down the whole desktop, which is far more disruptive than a crash in a dedicated process.
- It forces the **"Important usage note"** asking users to manually whitelist `dwm.exe` in the advanced settings (Windhawk doesn't inject there by default). With a tool mod that friction disappears entirely — the framework launches the dedicated `windhawk.exe` process for you — so that whole README section and screenshot can be removed.
- The home-grown single-instance assumption (one `dwm.exe`, so hotkey ids don't collide) is exactly what the tool-mod launcher's mutex handles for you.

This is the established convention for hotkey-driven window managers in the catalog — see [`move-window-to-monitor.wh.cpp`](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/move-window-to-monitor.wh.cpp) and [`tiling-helper.wh.cpp`](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/tiling-helper.wh.cpp), both of which do `RegisterHotKey` + `GetForegroundWindow` + `SetWinEventHook` + window manipulation from `@include windhawk.exe`. Copy the launcher boilerplate from the wiki/those mods **verbatim** (rename `Wh_ModInit`/`Wh_ModSettingsChanged`/`Wh_ModUninit` → `WhTool_*`, paste the `Wh_ModInit`/`Wh_ModAfterInit` launcher snippet unchanged) and switch `@include` to `windhawk.exe`. One trade-off to be aware of, shared by those mods: a medium-integrity tool process can't reposition windows of elevated apps (UIPI) — acceptable and conventional.

**Only install the system-wide WinEvent hook when there's at least one auto-snap rule.** `HookThreadProc` unconditionally installs a global `SetWinEventHook(EVENT_OBJECT_DESTROY, EVENT_OBJECT_SHOW, …)`. `EVENT_OBJECT_SHOW` fires for *every* window shown in *every* process on the desktop, and the default configuration has **no rules**, so every user pays for a high-frequency global hook (which then takes two mutexes per event in `TryApplyRuleToWindow` only to early-out). Gate it on `g_appRules` being non-empty, and install/uninstall it from the hook thread when settings change (you already post `WM_APP_REGISTER_HOTKEYS` there — a similar message can toggle the WinEvent hook). When rules *do* exist the hook is the right mechanism, so this is purely about not paying for it when it's unused.

<p>

- **Unsynchronized reads of the layout globals.** `g_gapTop/Bottom/Left/Right`, `g_gapMode`, `g_centerWidth/Height`, and `g_resizeStepPercent` are written by `LoadSettings()` (on the settings-changed thread) and read by `SnapWindow()` on the worker thread with no synchronization. It's benign in practice (independent scalars; worst case one snap mixes old/new gap values during a settings change), but worth a note. If you want it tidy, snapshot them into the work item or guard them like the other shared state.
- **`MOD_F_*` duplicate the Win32 modifier bits.** The comment already notes `MOD_F_ALT/CTRL/SHIFT/WIN` are defined to equal `MOD_ALT`/`MOD_CONTROL`/`MOD_SHIFT`/`MOD_WIN` so the mask can go straight to `RegisterHotKey`. You could just use the Win32 constants directly and drop the parallel set.
- **Teardown timeouts proceed unconditionally.** `Wh_ModUninit` waits 2 s on each thread and then `CloseHandle`s / `clear()`s regardless of whether the wait timed out. The only way to hit the timeout is a target window's thread being hung inside a cross-thread `SetWindowPos`, so it's an edge case, but on timeout the subsequent `g_savedPlacements.clear()` could race the still-running worker. (Tool-mod teardown calls `ExitProcess` right after anyway, which sidesteps it.)

</p>
</details>

<p>

- **Default `Alt`+`<letter>` shortcuts are fairly intrusive.** Because the combo is consumed globally, the defaults claim `Alt+Q/W/E/R/T/H/Y/G/…` system-wide, which collide with common application menu/ribbon accelerators (e.g. `Alt+H` opens the Home ribbon in Office apps, `Alt+F` the File menu, etc.). Rectangle itself defaults to `Ctrl`+`Opt`+arrows for this reason. The behavior is documented, but you may want a less collision-prone default modifier (e.g. add `Win` or `Ctrl`).
- **Cross-DPI monitor moves are approximate.** `MoveToAdjacentDisplay` scales the window rect by the work-area ratio between monitors. On a mixed-DPI setup that won't perfectly preserve the perceived size, though it's a reasonable approximation and there's no obviously better choice.
- **Auto-snap fires on the first `EVENT_OBJECT_SHOW`.** Some apps finish sizing/positioning themselves shortly *after* their window is first shown, so an auto-snap rule can briefly race the app's own initial layout. Usually fine; just an FYI for apps that look mis-snapped on launch.
- **`ResizeFromCenter` hard-codes a 200×150 px minimum.** Fine on normal displays; just noting it isn't DPI- or monitor-size-relative.

</p>
</details>

Asteski:
@m417z please check now

m417z:
It doesn't handle admin apps again, right? Would it work if you target dwm.exe and also keep the tool mod snippet? i.e. only change the target process?

If it works, you can add it to the readme as optional instructions to change the target process via the mod advanced tab.

Asteski:
> It doesn't handle admin apps again, right? Would it work if you target dwm.exe and also keep the tool mod snippet? i.e. only change the target process?
> 
> If it works, you can add it to the readme as optional instructions to change the target process via the mod advanced tab.

@m417z  It doesn't but when changing @include to dwm.exe it works. I've updated readme with that information.

mods/snap-commander.wh.cpp
```diff
@@ -113,6 +113,15 @@ Enable a rule's **Don't resize** toggle to move the window into the chosen regio
...
 side of the region (e.g. *top-right* pins it to the region's top-right). This is ignored
 by the Minimize action.
+
+## ⚠ Important usage note ⚠
+
+In order to use this mod with elevated processes, you must allow Windhawk to inject 
+into the **dwm.exe** system process and replace windhawk.exe as @included with dwm.exe.
+To do so, add it to the process inclusion list in the advanced settings, and modify 
+the mod code by forking the mod. If you do not do this, it will silently fail 
+to inject to elevated processes.
```
m417z: I think the wording can be improved. How about:

```markdown
## ⚠ Important usage note ⚠

In order to use this mod with elevated processes, it can be configured to target
`dwm.exe`.

First, navigate to Windhawk's Settings > Advanced settings > More advanced
settings > Process inclusion list, and make sure that `dwm.exe` is in the list.

![Advanced settings screenshot](https://i.imgur.com/LRhREtJ.png)

Next, go to the mod's advanced tab, enable this option: "Ignore mod
inclusion/exclusion lists", then add `dwm.exe` to the custom process inclusion
list.
```

Asteski:
@m417z  done

PR #4272, opened 2026-06-02
Add mod: Task View / Desktop View Styler

mahtab-ali:
New mod: **Task View / Desktop View Styler** (`task-desktop-view-styler`).

Restyles the native Windows 11 Task View (the Win+Tab Desktop View / virtual-desktops switcher) **in place** by attaching to Explorer's live XAML visual tree via `InitializeXamlDiagnosticsEx` — the same non-invasive mechanism used by Windows 11 Taskbar Styler. It only sets `CornerRadius` / `Background` / `BorderThickness` on targeted elements: no overlay window, no input hooks, nothing written to disk. Disabling reverts on the next Task View open.

Features:
- Rounded corners on window cards, desktop cards, and their hover/selection backplates (radii kept in sync).
- Optional border removal and background tint.
- Clear the virtual-desktops bar backplate for a transparent look.
- Fully data-driven target lists with a built-in element-type logger for tuning per Windows build.

The PR contains a single file, `mods/task-desktop-view-styler.wh.cpp`. The `@github` metadata matches my GitHub profile (mahtab-ali). Licensed MIT.

### DIFF
--- a/mods/task-desktop-view-styler.wh.cpp
+++ b/mods/task-desktop-view-styler.wh.cpp
@@ -4,7 +4,7 @@
 // @description     Restyles the native Windows 11 Task View (Desktop View) — rounded corners, cleaner backdrop, transparent desktops bar — by editing its live XAML tree, no overlay, non-invasive
 // @version         4.5.0
 // @author          mahtab-ali
-// @github          https://github.com/mahtab-ali/taskview-clean-styler
+// @github          https://github.com/mahtab-ali
 // @include         explorer.exe
 // @compilerOptions -lole32 -loleaut32 -lruntimeobject -loleacc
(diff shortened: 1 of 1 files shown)


m417z:
Thanks for the submission. Windows 11 Taskbar Styler already allows to style them. Also, there can only be one XAML Diagnostics consumer. Using both mods means that one of the mods won't receive event notifications.

If you'd like to submit a specific style for the Task View, you can submit a theme here:
https://github.com/ramensoftware/windows-11-taskbar-styling-guide#themes

Alternatively, if you'd like to create a mod for simple styling of the Task View, it'd be better to use a different approach that doesn't use XAML Diagnostics, so that it can be used alongside Windows 11 Taskbar Styler.

PR #4276, opened 2026-06-03
Add mod: Hide Desktop Icon Text and Shortcut Arrows

kivsak:
Provides options to hide desktop icon labels (keeping folder names) and remove the shortcut arrow overlay.

**Features**
* **Hide Icon Text**: Hides text labels for files and shortcuts on the desktop while preserving folder names.
* **Hide Shortcut Arrows**: Removes the standard shortcut arrow overlay from icons.
* Both features can be toggled independently via the mod settings.

**Notes**
* The shortcut arrow removal affects the entire running Explorer session because the overlay resides in the shared system image list.
* Disabling the shortcut arrow feature (or the mod itself) restores the arrows only after Explorer restarts or the icon cache refreshes.

**Acknowledgments**
* Huge thanks to @m417z! While this is a completely new mod, their `desktop-icons-view` mod provided the perfect boilerplate for the desktop `SysListView32` window detection.

**Known limitation**
* Folder detection matches the drawn text against actual desktop folder names. If a file has exactly the same name as a folder on the desktop, that file's label will also be kept.

## Changelog

* Initial release (New mod with configurable settings)

## Mod authorship

This mod was created by @kivsak:

- [x] Claude


m417z:
### Submission review

Thanks for the submission! I think adding an option for each feature would indeed be helpful. The items in the collapsed sections are optional, so it's your call whether to address them.

---

**1. Consider gating the arrow removal behind a setting (default off), or splitting it out.** The mod bundles two unrelated tweaks (label hiding + arrow removal) with no `==WindhawkModSettings==` block, so a user can't take just the reversible text-hiding without also getting the irreversible arrow change. Given the issue above, at minimum let users opt out of the arrow part; ideally the two are separate mods.

<p>

- **Trim unused `@compilerOptions` libraries.** `-luuid` is linked even though the comment at the `iidImageList` GUID explicitly says it's defined locally *to avoid depending on uuid.lib* — drop `-luuid`. Also double-check `-luxtheme` and `-lcomctl32`: `DrawThemeTextEx` is resolved via `GetProcAddress`, the `ListView_*` calls are `SendMessage` macros, and `SHGetImageList` comes from shell32 via `GetProcAddress` — so those import libs may be unnecessary too. CI doesn't catch unused `-l` deps.
- **`@name` "Ultimate Desktop Minimizer"** is a bit grandiose for what the mod does; something descriptive like "Hide Desktop Icon Text and Shortcut Arrows" reads better in the catalog. Subjective.

</p>
</details>

<p>

- **The shortcut-arrow removal is not reversible.** `HideShortcutOverlay()` rewrites the process-wide system image list (`ReplaceIcon` + `SetOverlayImage`) so the link-overlay slot points at a transparent image. A mod's effects must disappear when it's disabled — here they don't: as you note in the README, the arrow only comes back after Explorer restarts or the icon cache refreshes, because there's no way to read the original overlay image back. This also makes the change global to all of Explorer (every shortcut everywhere), not the desktop-only tweak the mod is presented as. The reversible way to do this is to hook the *draw path* and strip the overlay at render time instead of mutating shared state — e.g. hook `IImageList::Draw` / `ImageList_DrawIndirect` (comctl32) and clear the `ILD_OVERLAYMASK` bits / `INDEXTOOVERLAYMASK(linkOverlay)` from `fStyle`/`dwFlags` for the link overlay. That suppresses the arrow only while the mod is loaded and restores it instantly on unload, with no shared-state surgery.

Two consequences of the current approach that the same fix removes:
- Each load adds 5 transparent icons (one per `SHIL_*` size) to the system image list via `ReplaceIcon(-1, …)` and they're never removed, so enabling/disabling/updating the mod repeatedly leaks images into the shared list and permanently repoints the overlay slot.
- There's no `Wh_ModUninit` restoration for the overlay at all — only the subclass and a repaint.
- **The label-hiding heuristic is string-matching, with caveats beyond the documented one.** `ShouldHideLabel` keeps a label iff its drawn text is in the desktop folder-name set. Besides the file-named-like-a-folder collision you already document, this also: keeps a *file* label whenever it equals any folder's display name; hides any non-label text drawn during the desktop paint (e.g. group headers in grouped/Details views); and is sensitive to localization and to pre-truncated/ellipsized strings the listview might pass. Given that the `DrawTextW`/`DrawThemeTextEx` hook point has no per-item context, name-matching is about the only option — just flagging the breadth as FYI.
- **WorkerW configurations aren't covered.** `IsDesktopFolderView`/`GetDesktopFolderView` only recognize the FolderView whose top ancestor is `Progman` or `GetShellWindow()`. When a wallpaper slideshow (or some "active desktop" configs) reparents `SHELLDLL_DefView` under a `WorkerW`, detection fails and the mod won't apply. This is inherited from the `desktop-icons-view` boilerplate, so it may be acceptable, but worth knowing.
- **`UpdateFolderCache()` runs a COM shell enumeration inside `WM_PAINT`** (throttled to once/sec). On a slow desktop folder (e.g. a redirected/network home folder) the first paint after a change could briefly stutter. Throttling makes this mostly a non-issue.
- **`SHChangeNotify(SHCNE_ASSOCCHANGED, …)` in `Wh_ModAfterInit`** is a system-wide hammer that forces every shell icon to recompose, causing a brief icon flicker across Explorer on enable. Expected given the chosen arrow mechanism; would go away with the draw-path approach in item 1.

</p>
</details>

kivsak:
Thank you for the detailed review and suggestions! 

I have updated the mod to address the core feedback:
1. Added a ==WindhawkModSettings== block with separate toggles for hiding desktop text (hide_text) and removing shortcut arrows (hide_arrows). Now users can use them independently.
2. Renamed the mod to "Hide Desktop Icon Text and Shortcut Arrows" as requested.
3. Added the -lcomctl32 library back to compiler options to fix the subclassing linker errors.

Regarding the draw-path alternative for arrows and WorkerW edge cases — thank you for the insights! I will keep them in mind for future improvements.

PR #4278, opened 2026-06-03
Add Redirect Settings to Control Panel mod

babamohammed2022:
# Redirect Settings to Control Panel

This mod intercepts modern `ms-settings:` URI protocols and forces Windows to open their classic Control Panel equivalents using native Windows components and legacy CLSIDs.

## Features
- Intercepts ShellExecuteExW, ShellExecuteW, and CreateProcessW calls
- Smart Personalization Detection via hwnd context awareness
- 200+ URI mappings to Control Panel applets
- Configurable fallback modes (ignore, control.exe, passthrough)
- Windows 11 compatibility mode for deprecated CLSIDs

## Compatibility
- Windows 10 21H2 (IoT Enterprise LTSC 2021) ✅
- Windows 10 22H2 ✅
- Windows 11 24H2 ✅ (but more limited compared to Windows 10)

## Limitations
- Windows 10: ~70% redirection success rate
- Windows 11: ~5% baseline restoration (many CLSIDs deprecated by Microsoft)

---

## Changelog

* Initial release - Redirect Settings to Control Panel v9.6.0
* 200+ ms-settings: URI mappings to classic Control Panel
* Smart Personalization detection with hwnd context awareness
* Windows 11 compatibility mode

## Mod authorship

This mod was created by:

- - [X] Claude
- - [X] ChatGPT
- - [X] Gemini

### DIFF
--- a/mods/settings-to-control-panel.wh.cpp
+++ b/mods/settings-to-control-panel.wh.cpp
@@ -2,14 +2,14 @@
 // @id             settings-to-control-panel
 // @name           Redirect Settings to Control Panel
 // @description    Forces classic Control Panel to open instead of Windows 10/11 Settings app using native components. Primarily designed for Windows 10; Windows 11 support is limited due to Microsoft's shell architecture changes.
-// @version        9.8.7
+// @version        9.8.8
 // @author         babamohammed
 // @github         https://github.com/babamohammed2022
 // @include        explorer.exe
 // ==/WindhawkMod==
 // ==WindhawkModReadme==
 /*
-# Redirect Settings → Control Panel (v9.8.7)
+# Redirect Settings → Control Panel (v9.8.8)
 
 This Windhawk mod intercepts modern `ms-settings:` URIs and redirects them to
 their corresponding classic Control Panel applets, relying exclusively on native
@@ -470,18 +470,8 @@ static void InitMappings() {
             L"shell:::{8E908FC9-BECC-40f6-915B-F4CA0E70D03D}"},
         {L"ms-settings:network-ethernet",
             L"shell:::{8E908FC9-BECC-40f6-915B-F4CA0E70D03D}"},
-        {L"ms-settings:network-vpn",
-            L"shell:::{8E908FC9-BECC-40f6-915B-F4CA0E70D03D}"},
-        {L"ms-settings:network-airplanemode",
-            L"shell:::{8E908FC9-BECC-40f6-915B-F4CA0E70D03D}"},
-        {L"ms-settings:network-mobilehotspot",
-            L"shell:::{8E908FC9-BECC-40f6-915B-F4CA0E70D03D}"},
-        {L"ms-settings:network-cellular",
-            L"shell:::{8E908FC9-BECC-40f6-915B-F4CA0E70D03D}"},
-        {L"ms-settings:datausage",
-            L"shell:::{8E908FC9-BECC-40f6-915B-F4CA0E70D03D}"},
         {L"ms-settings:network-proxy",                       L"inetcpl.cpl,,4"},
-        {L"settings:network-status", 
+        {L"ms-settings:network-status", 
             L"shell:::{7007ACC7-3202-11D1-AAD2-00805FC1270E}"},
         {L"ms-settings:network-status",
             L"shell:::{7007ACC7-3202-11D1-AAD2-00805FC1270E}"},
@@ -520,9 +510,14 @@ static void InitMappings() {
         {L"ms-settings:regionformatting",                    L"intl.cpl"},
         {L"ms-settings:language",                            L"intl.cpl"},
 
-        // ── Ease of Access ────────────────────────────────────────────────────
-        {L"ms-settings:easeofaccess",
-            L"shell:::{D555645E-D4F8-4c29-A827-D93C859C4F2A}"},
+        // ── Ease of Access ────────────────────────────────────────────────────,
+        // Ease Of Access entries
+        {L"ms-settings:easeofaccess", L"shell:::{D555645E-D4F8-4c29-A827-D93C859C4F2A}"},
+        {L"ms-settings:easeofaccess-narrator", L"shell:::{D555645E-D4F8-4c29-A827-D93C859C4F2A}\\pageNoVisual"},
+        {L"ms-settings:easeofaccess-magnifier", L"shell:::{D555645E-D4F8-4c29-A827-D93C859C4F2A}\\pageEasierToSee"},
+        {L"ms-settings:easeofaccess-speech", L"shell:::{D555645E-D4F8-4c29-A827-D93C859C4F2A"},
+        {L"ms-settings:easeofaccess-colorfilter", L"shell:::{D555645E-D4F8-4c29-A827-D93C859C4F2A}\\pageEasierToSee"},
+        {L"ms-settings:easeofaccess-display", L"shell:::{D555645E-D4F8-4c29-A827-D93C859C4F2A}\\pageEasierToSee"},
 
         // ── Recovery / Backup ─────────────────────────────────────────────────
         {L"ms-settings:backup",
@@ -753,8 +748,13 @@ static bool IsPersonalizationWindow(HWND hwnd) {
             return false;
         }
         if (c == L"cabinetwclass") {
-            Wh_Log(L"HWND: CabinetWClass -> inside Personalization");
-            return true;
+            if (t.find(L"personaliz") != std::wstring::npos) {
+                Wh_Log(L"HWND: CabinetWClass personalization confirmed");
+                return true;
+        }
+
+            Wh_Log(L"HWND: CabinetWClass but not personalization");
+            return false;
         }
         if (t.find(L"personaliz") != std::wstring::npos) {
             Wh_Log(L"HWND: title match 'personaliz' -> Personalization window");
@@ -802,6 +802,18 @@ static ResolveResult ResolveUri(const std::wstring& uri, HWND hwnd) {
         return {t, true};
     }
 
+    // ─── EDIT FOR EASE OF ACCESS IN PERSONALIZATION ───
+
+    if (g_settings.smartPersonalizationDetect && IsPersonalizationWindow(hwnd)) {
+        if (uri.find(L"ms-settings:easeofaccess") == 0) {
+            Wh_Log(L"SmartPersonalization: Accesso da finestra Personalizzazione, forzo radice Centro Accessibilità");
+            std::wstring t = ApplyWin11Filter(L"shell:::{D555645E-D4F8-4c29-A827-D93C859C4F2A}");
+            BounceGuardRecord(uri);
+            return {t, true};
+        }
+    }
+    // ─────────────────────────────────────────────────────────────────
+
     auto it = g_mappings.find(uri);
     if (it != g_mappings.end()) {
         if (BounceGuardIsBounce(uri)) {
@@ -839,7 +851,6 @@ static ResolveResult ResolveUri(const std::wstring& uri, HWND hwnd) {
     Wh_Log(L"Unmapped, passing through: %s", uri.c_str());
     return {L"", false};
 }
-
 // ── Precise control system detection ─────────────────────────────────────────
 
 static std::wstring BaseNameLower(const std::wstring& path) {
@@ -1064,7 +1075,7 @@ BOOL WINAPI CreateProcessW_hook(
 // ── Windhawk entry points ─────────────────────────────────────────────────────
 
 BOOL Wh_ModInit() {
-    Wh_Log(L"Redirect Settings to Control Panel v9.8.7 init");
+    Wh_Log(L"Redirect Settings to Control Panel v9.8.8 init");
 
     DetectWindowsVersion();
     LoadSettings();
--- a/mods/settings-to-control-panel.wh.cpp
+++ b/mods/settings-to-control-panel.wh.cpp
@@ -2,7 +2,7 @@
 // @id             settings-to-control-panel
 // @name           Redirect Settings to Control Panel
 // @description    Forces classic Control Panel to open instead of Windows 10/11 Settings app using native components. Primarily designed for Windows 10; Windows 11 support is limited due to Microsoft's shell architecture changes.
-// @version        10.0.2
+// @version        10.0.1
 // @author         babamohammed
 // @github         https://github.com/babamohammed2022
 // @include        explorer.exe
@@ -26,7 +26,6 @@ corresponding classic Control Panel applets using only native Windows components
 ## Features
 
 - Redirects numerous `ms-settings:` URIs to classic Control Panel
-- Smart personalization detection (desktop vs in-window)
 - Anti-loop protection
 - Configurable fallback modes
 
@@ -67,7 +66,7 @@ Control Panel target instead of the modern Settings app.
 
 - m417z – Code reviews and feedback
 - Anixx – Testing on Windows 11 23H2
-- dbilanoski – CLSID documentation  
+- dbilanoski – CLSID documentation
 */
 // ==/WindhawkModReadme==
 // ==WindhawkModSettings==
@@ -104,8 +103,6 @@ Control Panel target instead of the modern Settings app.
 #include <vector>
 #include <mutex>   
 #include <initguid.h>  
-#include <chrono>
-#include <thread>
 
 // Dynamic COM handling - no linking required
 typedef HRESULT (WINAPI *CoCreateInstance_t)(const GUID* rclsid, IUnknown* pUnkOuter, DWORD dwClsContext, const GUID* riid, void** ppv);
@@ -136,10 +133,6 @@ using ShellExecuteW_t = HINSTANCE(WINAPI*)(HWND, LPCWSTR, LPCWSTR, LPCWSTR, LPCW
 using ShellExecuteExW_t = BOOL(WINAPI*)(SHELLEXECUTEINFOW*);
 static ShellExecuteExW_t ShellExecuteExW_orig = nullptr;
 static ShellExecuteW_t ShellExecuteW_orig = nullptr;
-
-// Pending redirects guard per evitare chiamate duplicate
-static std::unordered_set<std::wstring> g_pendingRedirects;
-static std::mutex g_pendingMtx;
 
 // Core resolve logic struct - MUST be defined before use
 struct ResolveResult {
@@ -331,6 +324,13 @@ static const std::unordered_set<std::wstring> g_win11LoopClsids = {
     L"shell:::{9fe63afd-59cf-4419-9775-abcc3849f861}",
 };
 
+static const std::unordered_set<std::wstring> g_win11NoForceRedirect = {
+    L"ms-settings:defaultapps",
+    L"ms-settings:appsfeatures",
+    L"ms-settings:network",
+    L"ms-settings:bluetooth",
+    L"ms-settings:printers",
+};
 
 static bool IsClsidSafeOnWin11(const std::wstring& lowerTarget) {
     return g_win11SafeClsids.count(lowerTarget) > 0;
@@ -651,7 +651,7 @@ static bool HandleFallback(const std::wstring& uri) {
     }
 }
 
-// LaunchTarget - VERSIONE CORRETTA CON PULIZIA PENDING REDIRECTS
+// LaunchTarget - VERSIONE CORRETTA
 static void LaunchTarget(const std::wstring& command) {
     Wh_Log(L"Launching: %s", command.c_str());
 
@@ -756,8 +756,11 @@ static void LaunchTarget(const std::wstring& command) {
         CloseHandle(pi.hThread);
     }
 }
-
 // Personalization detection
+
+
+
+
 
 // Helper function to open sound panel
 static bool OpenSoundPanel() {
@@ -811,17 +814,7 @@ static std::wstring ResolvePersonalizationBackground(HWND hwnd) {
     Wh_Log(L"personalization-background -> Personalization root");
     return PERS_ROOT;
 }
-
 static ResolveResult ResolveUri(const std::wstring& uri, HWND hwnd) {
-    // Controlla se questo URI è già in elaborazione
-    {
-        std::lock_guard<std::mutex> lk(g_pendingMtx);
-        if (g_pendingRedirects.find(uri) != g_pendingRedirects.end()) {
-            Wh_Log(L"URI already being processed, blocking duplicate: %s", uri.c_str());
-            return {L"", true}; // Blocca completamente
-        }
-    }
-    
     if (uri == L"ms-settings:personalization-background") {
         if (BounceGuardIsBounce(uri)) {
             Wh_Log(L"Bounce-back on personalization-background - suppressing duplicate, target is opening");
@@ -829,12 +822,6 @@ static ResolveResult ResolveUri(const std::wstring& uri, HWND hwnd) {
         }
         std::wstring t = ApplyWin11Filter(ResolvePersonalizationBackground(hwnd));
         BounceGuardRecord(uri);
-        
-        // Marca come pendente
-        {
-            std::lock_guard<std::mutex> lk(g_pendingMtx);
-            g_pendingRedirects.insert(uri);
-        }
         return {t, true};
     }
 
@@ -856,12 +843,6 @@ static ResolveResult ResolveUri(const std::wstring& uri, HWND hwnd) {
 
         Wh_Log(L"Mapped: %s -> %s", uri.c_str(), t.c_str());
         BounceGuardRecord(uri);
-        
-        // Marca come pendente
-        {
-            std::lock_guard<std::mutex> lk(g_pendingMtx);
-            g_pendingRedirects.insert(uri);
-        }
         return {t, true};
     }
 
@@ -874,12 +855,6 @@ static ResolveResult ResolveUri(const std::wstring& uri, HWND hwnd) {
         if (g_isWin11 && IsClsidLoopOnWin11(uri)) {
             std::wstring t = ApplyWin11Filter(uri);
             Wh_Log(L"Win11 loop-CLSID intercepted: %s -> %s", uri.c_str(), t.c_str());
-            
-            // Marca come pendente
-            {
-                std::lock_guard<std::mutex> lk(g_pendingMtx);
-                g_pendingRedirects.insert(uri);
-            }
             return {t, true};
         }
     }
@@ -887,7 +862,6 @@ static ResolveResult ResolveUri(const std::wstring& uri, HWND hwnd) {
     Wh_Log(L"Unmapped, passing through: %s", uri.c_str());
     return {L"", false};
 }
-
 // Control system detection helpers
 static std::wstring BaseNameLower(const std::wstring& path) {
     size_t pos = path.rfind(L'\\');
@@ -928,14 +902,6 @@ static bool IsControlSystemCommand(const std::wstring& cmdLine) {
 
     std::wstring arg = ToLower(tokens[1]);
     return (arg == L"system" || arg == L"microsoft.system");
-}
-
-// Funzione per pulire i pending redirects dopo un delay
-static void CleanupPendingRedirect(const std::wstring& uri, DWORD delayMs) {
-    std::this_thread::sleep_for(std::chrono::milliseconds(delayMs));
-    std::lock_guard<std::mutex> lk(g_pendingMtx);
-    g_pendingRedirects.erase(uri);
-    Wh_Log(L"Cleaned up pending redirect for: %s", uri.c_str());
 }
 
 // Hook: ShellExecuteExW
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

---

The approach (hooking `ShellExecute*`/`CreateProcessW` in explorer so the redirect is fully reversible) is the right pattern for Windhawk. But the detection logic is too broad in a couple of places and breaks unrelated launches, and the mapping table is large but inaccurate. A few concrete things to fix:

**1. The `CreateProcessW` "control system" matcher fires on almost every `control.exe <applet>` call.** The match is `cmdLine` contains `"control"` **and** `"system"`, minus a short exclusion list. The problem: when explorer launches a Control Panel item it passes the full path, e.g. `C:\Windows\System32\control.exe appwiz.cpl` — that string contains `"system"` (from `System32`) and `"control"`, so it is intercepted and redirected to **System Properties**. The exclusion list (`mmsys.cpl`, `main.cpl`, `desk.cpl`, …) is itself proof of this — none of those names contain `"system"`, so they only needed excluding because the `System32` path is present. Anything *not* in that list (`appwiz.cpl`, `powercfg.cpl`, `firewall.cpl`, `wscui.cpl`, `colorcpl`, `sysdm.cpl`, …) gets wrongly sent to System Properties.

Match the actual command precisely instead of substring-scanning: parse the command line into argv, confirm the executable is `control.exe` and the lone argument is `system` (or `/name microsoft.system`).

Related: when the hook intercepts, it does `return TRUE` without ever populating `lpProcessInformation`. `CreateProcessW`'s contract is that on success it fills `hProcess`/`hThread`/`dwProcessId`/`dwThreadId`. A caller that doesn't zero-init that struct and then reads/closes `hProcess` will touch a garbage handle. If you keep the fake-success path, zero the struct (and don't hand back a handle you didn't create); better, redirect via the `ShellExecute*` hooks and leave `CreateProcessW` to call the original.

**2. Unmapped `shell:::{…}` targets are hijacked to Control Panel.** `IsShellClsid()` matches *any* `shell:::` string passed to `ShellExecuteExW`/`ShellExecuteW`, and an unmapped one falls into `HandleFallback`, whose default (`FallbackMode = 1`) launches `control.exe` **and** returns `intercept = true` — so the original action is suppressed. That means any `shell:::{clsid}` invoked from explorer that you didn't explicitly list silently opens Control Panel instead of doing what was asked. Only `ms-settings:` inputs should be subject to the fallback policy; an unknown `shell:::` target should fall through to the original (`return ShellExecute*_orig(...)`).

**3. `.msc` targets won't launch.** In `LaunchTarget`, `devmgmt.msc` / `compmgmt.msc` take the "run directly" branch and are passed straight to `CreateProcessW`. `CreateProcessW` does not do shell association resolution, so a bare `.msc` (which isn't an executable) fails — Device Manager / Computer Management never open (you'll see the `[ERROR] CreateProcess failed` log). Launch them as `mmc.exe <file>.msc`, or route `.msc`/`.cpl` through `ShellExecuteW`/the original instead of `CreateProcessW`.

**4. `FallbackMode` uses `$options` on a number setting.** `FallbackMode: 1` is inferred as a number and read with `Wh_GetIntSetting`, but `$options` dropdowns are for **string** settings — the stored option keys are compared as strings read via `Wh_GetStringSetting`. Compare with [dynamic-island-for-windows](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/dynamic-island-for-windows.wh.cpp#L98-L107), where both the default and the option keys are quoted strings. Either make it a string setting (`FallbackMode: "1"`, quote the `0/1/2` keys, read with `Wh_GetStringSetting`) or drop `$options` and describe the integer values in `$description`.

**5. Accuracy of the mapping table / overall effectiveness.** The README itself states ~70% success on Win10 and ~5% on Win11, and many of the 200+ entries point at unrelated applets — e.g. `ms-settings:privacy-microphone` → Ease of Access Center (`{D555645E…}`), `ms-settings:taskbar` → Notification Area Icons, the `ms-settings:gaming*` family → a bare `control.exe` (Control Panel root). For a mod whose entire purpose is *accurate* redirection, landing the user on the wrong page is worse than not redirecting. Please trim the table to mappings that are actually verified, and for pages with no real classic equivalent, prefer leaving them alone (fallback passthrough) over mapping them to an unrelated applet or a bare Control Panel window. This is also relevant to the mod's standing in the catalog — a mod that's openly ~5% effective on the current OS is a hard sell.

Since the PR discloses heavy AI assistance, a few of these read like AI artifacts worth a second look: the `control`+`system` substring heuristic, the duplicated `mmsys.cpl` exclusion, and the many semantically-wrong mappings. Worth a manual pass to confirm each mapping is intentional and correct.

<p>

- **Redundant log prefixes.** `Wh_Log` already prefixes every line with the mod name, so the `[v9.6.0]`, `[SETTINGS]`, `[DEBUG]`, `[HOOK]` … tags are noise — drop them.
- **`@compilerOptions -lshell32 -lkernel32`.** Both are linked by default, and the mod resolves these APIs through `GetProcAddress` anyway, so neither `-l` is needed.
- **`NormalizeUri` off-by-one.** `result.substr(pos + 15)` should be `pos + 14` (length of `"ms-settings://"`); as written it drops the first character after `//` (e.g. `ms-settings://display` → `ms-settings:isplay`). Narrow trigger (the `//` form is uncommon) but still wrong.
- **`(HINSTANCE)42`** as the fake `ShellExecuteW` success value is a magic number — a named constant (any value > 32) with a one-line comment reads better.
- **Settings reads race with `LoadSettings`.** `Wh_ModSettingsChanged` runs on an arbitrary thread and rewrites the `g_settings` fields while the hooks read them. The torn reads are harmless scalars and only happen on a settings change, so this is low priority, but `std::atomic` (or just accepting it) would tidy it up.
- **README visual.** The effect is visible (a Control Panel window opening instead of Settings); a short GIF/screenshot would help users understand what the mod does. Only `i.imgur.com` / `raw.githubusercontent.com` are allowed image hosts.

</p>
</details>

<p>

- **Personalization detection is a fuzzy, English-only heuristic.** `IsPersonalizationWindow` returns true for any `CabinetWClass` window (the class is shared by *all* Explorer and Control Panel windows) and otherwise matches the substring `"personaliz"` in the window title — which only works on English Windows (localized builds use translated titles). So "Desktop Background → wallpaper" detection can misfire when invoked from an unrelated Explorer window or on a non-English UI. It's a reasonable best-effort, just noting the limitation.
- **Scope is explorer-only.** Redirects fire only for `ShellExecute*`/`CreateProcessW` calls made *inside* `explorer.exe`. `ms-settings:` links opened from Start/Search (StartMenuExperienceHost, SearchHost) or from other apps won't be redirected. That's a sensible conservative scope, but worth stating in the README so users know what's covered.

</p>
</details>

babamohammed2022:
Hello @m417z, I've edited the code to fix the reported issues (command parsing, shell fallback, .msc launching, string settings, wrong mappings). I hope that the mod is now ready for Windhawk. Have a good day!

Note: The fixes were assisted by AI.

babamohammed2022 (commented):
Hello! The code has been updated.

m417z:
As the review below says, manual testing is expected. Please do extensive manual testing, and post here how the testing was done.

### Submission review

---

Thanks for the quick turnaround — most of the previous round is addressed: the `control system` matcher now parses argv precisely (no more `System32` false-match), unmapped `shell:::` targets fall through, `.msc` files launch via `mmc.exe`, `FallbackMode` is a quoted-string `$options` setting, `PROCESS_INFORMATION` is zeroed, the `NormalizeUri` off-by-one is fixed, the magic `(HINSTANCE)` value is named, and the redundant `@compilerOptions`/log prefixes are gone. Two things still need attention before merge:

**1. The `__display__` sentinel never resolves — those two redirects are broken.** Two entries map to the literal string `__display__`:
```cpp
{L"shell:::{26ee0668-...}\\3\\::{c555438b-...}", L"__display__"},   // line 316
{L"shell:::{f20df4e5-ea01-41a2-b02a-dcbd92d4696e}", L"__display__"}, // line 344
```
Nothing in `ResolveUri`/`ApplyWin11Compat`/`LaunchTarget` ever translates `__display__`, so it falls into the final `else` branch of `LaunchTarget` and runs **`control.exe __display__`** — control.exe doesn't recognize that token, so it opens the Control Panel root (or nothing), not Display settings. Map these to a real target (e.g. `L"desk.cpl"`, matching the other display entries on lines 185-188) or remove them. Easy to miss because the symptom is "wrong page opens," not a crash.

**2. Mapping accuracy — the gating concern from last round — is only partly addressed.** You trimmed the worst offenders (gaming family, `privacy → easeofaccess`, bare `control.exe` redirects), which is good. But the table still collapses whole `ms-settings:` categories onto a single generic CLSID, so the user lands on a page that isn't what they clicked:
   - every `ms-settings:easeofaccess-*` (narrator, magnifier, colorfilter, audio, …) → the same Ease of Access Center `{D555645E…}` (lines 261-276);
   - every `ms-settings:signinoptions-*` / `accounts-*` (fingerprint, face, dynamiclock, otherusers, …) → the same User Accounts `{60632754…}` (lines 234-249) — classic User Accounts has no fingerprint/face page;
   - every `ms-settings:network-*` → Network Connections `{8E908FC9…}` (lines 218-225);
   - `ms-settings:storagesense` / `storagepolicies` / `disksandvolumes` / `savelocations` → `{20D04FE0…}`, which is just **This PC** (lines 205-208), not any storage UI.

   As noted before, for a mod whose entire purpose is *accurate* redirection, landing on a generic/unrelated page is worse than not redirecting — please prefer fallback-passthrough for subpages that have no real classic equivalent rather than bucketing them onto a parent CLSID. This also ties back to catalog standing: the README still advertises **~5% effectiveness on Windows 11**, which is a hard sell for a merged mod. It's worth being honest with yourself about which mappings you've actually verified open the intended page, and dropping the rest.

Since the PR still discloses heavy AI assistance, the coarse category→single-CLSID buckets and the dangling `__display__` placeholder both read like AI artifacts — worth a manual pass to confirm each remaining mapping is intentional and verified.

<p>

- **Dead mapping entries.** `uri` is only ever populated for strings containing `ms-settings:` or `shell:::` (see `IsMsSettings`/`IsShellClsid` in the hooks), so the entries keyed `L"explorer.exe"`, `L"explorer"` (lines 158-159) and `L"control desktop"` (line 251) can never be looked up. Remove them.
- **Identity `shell::: → same shell:::` mappings add cost with no benefit.** Entries like lines 157, 317, 345-357 map a CLSID to itself, so the hook intercepts the call, spawns a *new* `explorer.exe shell:::{…}` process, and returns success — i.e. the same window the caller would have opened anyway, just via an extra process while suppressing the caller's own handling. Letting these fall through to the original (don't list them) is simpler and avoids the redundant spawn.
- **Type-safe hooks / skip the `GetProcAddress` dance.** `ShellExecuteExW`/`ShellExecuteW` are ordinary `shell32` exports the mod already links against, so you can hook them directly and type-checked: `WindhawkUtils::SetFunctionHook(ShellExecuteExW, ShellExecuteExW_hook, &ShellExecuteExW_orig)` — no `GetModuleHandle`/`GetProcAddress`/`void*` casts needed (lines 743-768).
- **`Wh_GetStringSetting` never returns NULL** — it returns `L""` on unset/error — so the `fallbackStr ?` guard (lines 92-94) is dead, and an empty value would silently give `fallbackMode = 0` (ignore) rather than the intended default `1`. `WindhawkUtils::StringSetting` (RAII) would also drop the manual `Wh_FreeStringSetting`.
- **README visual.** The effect is visible (a classic Control Panel window opening instead of Settings); a short GIF/screenshot would help users. Only `i.imgur.com` / `raw.githubusercontent.com` are allowed image hosts.
- **Settings race (still present, low priority).** `Wh_ModSettingsChanged` rewrites `g_settings` on an arbitrary thread while the hooks read it. The torn reads are harmless scalars and only on a settings change, so it's fine to leave — noting it for completeness.

</p>
</details>

<p>

- **`CreateProcessW` is hooked process-wide in explorer for a single narrow case.** The matcher is now precise, so the false-positive risk is gone, but every process creation in explorer still pays a tokenization pass just to catch `control system`. As suggested last round, handling System Properties through the `ShellExecute*` hooks and leaving `CreateProcessW` alone would be lighter — your call now that the match is correct.
- **`LaunchTarget` calls the *hooked* `CreateProcessW`, not `CreateProcessW_orig`.** Harmless today because none of the resolved targets tokenize to `control system`, but it means every redirect re-enters your hook, and the moment a mapping ever resolves to `control system` it would recurse. Calling `CreateProcessW_orig` from `LaunchTarget` would make it robust.
- **Personalization detection is a fuzzy, English-only heuristic.** `IsPersonalizationWindow` treats *any* `CabinetWClass` window as the Personalization window (that class is shared by all Explorer/Control Panel windows) and otherwise matches the substring `"personaliz"` in the title, which only holds on English Windows. So the "Desktop Background → wallpaper" branch can misfire from an unrelated Explorer window or on a localized UI. Reasonable best-effort — just noting the limitation.

</p>
</details>

babamohammed2022:
Hello @m417z. Regarding the testing methodology: I've tested on Windows 10 21H2 LTSC by opening each URI manually via Start → Run and the ms-settings: protocol, and verified the correct classic window opened. I personally confirmed: personalization (right-click desktop and from inside the window), mouse/touchpad, keyboard, sound, network, firewall, accounts, date/time, devices, printers, UAC prompts for Device Manager/Computer Management/Optional Features, and FallbackMode 0/1/2.
Regarding Windows 11: I don't currently have a Win11 machine available to test on. The last time I tested, redirects were partial as some CLSIDs fired correctly (network, firewall, accounts) and others didn't due to deprecated CLSIDs. I'd rather be honest about this than claim coverage I can't verify. I'm open to repositioning the mod as Windows 10 primary with best-effort Win11 support if that's acceptable.
Regarding CabinetWClass: I understand the limitation, it matches all Explorer windows, not just Personalization. I don't currently know how to make that detection more precise. I'd be happy to accept guidance on a better approach.

Anixx:
Clicking properties of This PC still brings an infinite loop with a clock animation (blue circle) on mouse cursor appearing and disappearing once a second and no properties dialog appearing. This is on Windows 11 23H2.

babamohammed2022:
> Clicking properties of This PC still brings an infinite loop with a clock animation (blue circle) on mouse cursor appearing and disappearing once a second and no properties dialog appearing. This is on Windows 11 23H2.

Hello! This is the feedback of the new update:
v9.8.4 pushed with all review issues addressed and tested on both platforms
The modification has been tested on:
- Windows 10 21H2 IoT Enterprise LTSC 2021 (Build 19044)
- Windows 11 24H2 (Build 26100)
Working on both:
- Personalization (root, wallpaper, colors via "explorer shell:::{ED834ED6}" syntax)
- System/About (sysdm.cpl on Win11, CLSID on Win10) <- this should address the previously reported issue
- Device Manager, Computer Management, Optional Features (UAC via ShellExecute)
- Sound, Volume Mixer, Network, Firewall, Accounts, Date/Time, Devices, Printers
- Fonts (direct path on Win11), Ease of Access, AutoPlay
- FallbackMode 0/1/2, kill switch, no false positives
- Anti-loop guards: cross-process, in-process, bounce-back detection

Known limitations on Win11:
- Default Apps has no working classic equivalent on 24H2 
- Desktop right-click → Personalize does not redirect on Win11 (the context menu
  uses a different code path than the CPL; CPL redirects like Personalization →
  Background/Colors work correctly on both Win10 and Win11)
- Some panels removed by Microsoft on Win11 have no classic equivalent

If you find other broken redirects on Win11 please let me know and I'll
investigate. I'll continue testing on new builds and update compatibility
where possible without introducing instability.

Anixx:
Why the "Trouleshooting" element in Control Panel brings the "All Control Panels elements" folder? By the way, legacy troubleshooting [still exists in Win 11](https://www.elevenforum.com/t/how-to-restore-the-old-troubleshooting.44606/#post-710096). 

You also can remove the license tag because it would be published under the same license (MIT) as Windhawk as a whole if you omit the license. 

The System Properties do not loop now infinitely, but there is a noticeable pause after invoking it.

Taskbar or notification settings do not redirect to the tray setup dialog
explorer shell:::{05d7b0f4-2121-4eff-bf6b-ed3f69b894d9}

Anixx:
> redirects like Personalization →
Background/Colors work correctly on both Win10 and Win11

What is meant? Where it can be tested?

Anixx:
Can you give some examples of redirects on Win11 that work? I do not see much of a change after I enable your mod.

Anixx:
Right clicking on the network icon in tray still brings the Settigs app with network settings, not the Control Panel network settings.

Anixx:
Clicking "Sound settings" in the speaker icon's context menu in tray brings the Settings app, not the Control Panel's sound properties.

In general, I see no effect from your mod. Nothing gets changed except the system properties dialog.

Anixx:
Cklicking this
изображение
brings this:
изображение
not this:
изображение
Clicking this:
изображение
brings this:
изображение
not this:
изображение

babamohammed2022:
> Why the "Trouleshooting" element in Control Panel brings the "All Control Panels elements" folder? By the way, legacy troubleshooting [still exists in Win 11](https://www.elevenforum.com/t/how-to-restore-the-old-troubleshooting.44606/#post-710096).
> 
> You also can remove the license tag because it would be published under the same license (MIT) as Windhawk as a whole if you omit the license.
> 
> The System Properties do not loop now infinitely, but there is a noticeable pause after invoking it.
> 
> Taskbar or notification settings do not redirect to the tray setup dialog explorer shell:::{05d7b0f4-2121-4eff-bf6b-ed3f69b894d9}

Thank you for the report. I have fixed the issue related to the Troubleshooting element in CPL (it was caused by a confused mapping) by using your documentation and I've removed the license and it's good that the system properties do not loop infinitely. However, the fact that it is slow is probably caused by the architecture of the Settings App in Windows 11. Furthermore, I'll be working on fixing the taskbar and notification settings issue if possible, since Windows 11 uses a different taskbar architecture so I need to document myself more on this. This is what gets opened now. I'll post the code with the applied fixes soon.
image

babamohammed2022:
> > redirects like Personalization →
> > Background/Colors work correctly on both Win10 and Win11
> 
> What is meant? Where it can be tested?

To open the classic personalization in Windows 11, press WIN+R and type shell:::{ED834ED6-4B5A-4bfe-8F11-A626DCB6A921} and, thanks to the mod, it will work exactly as it did in Windows 7/8/8.1 because elements from this CPL part are still present in Windows 11.
- Tested on Windows 11 24H2
image

Anixx:
My screenshots above were from the legacy taskbar (the one which is like Win10). It is included in Windows 11 but disabled by default. So, its architecture is the same as in Win10.

Anixx:
> To open the classic personalization in Windows 11, press WIN+R and type shell:::{ED834ED6-4B5A-4bfe-8F11-A626DCB6A921} and, thanks to the mod, it will work exactly as it did in Windows 7/8/8.1 because elements from this CPL part are still present in Windows 11.

Hmmm. Works without the mod as well for me... The mod does not change it.

babamohammed2022:
> My screenshots above were from the legacy taskbar (the one which is like Win10). It is included in Windows 11 but disabled by default. So, its architecture is the same as in Win10.

OK, thank you for the information. I'll use this info to investigate further on the behavior on Windows 10 too, I could have forgotten to add some mappings. Thanks.

> > To open the classic personalization in Windows 11, press WIN+R and type shell:::{ED834ED6-4B5A-4bfe-8F11-A626DCB6A921} and, thanks to the mod, it will work exactly as it did in Windows 7/8/8.1 because elements from this CPL part are still present in Windows 11.
> 
> Hmmm. Works without the mod as well for me... The mod does not change it.

This is unusual. While the CLSID redirects to the classic personalization, when pressing "Colors" or "Backgrounds" in both Windows 11 24H2 and Windows 10 21H2 IoT Enterprise LTSC 2021 it opened the Settings app. I'll double-check to be sure anyways.
Thank you for the detailed report.

Anixx:
Yes, your mod seems to fix the Colors and Backgrounds links.

Anixx:
I like the current behavior of your mod on click on the This PC's properties, but be informed that the old System Properties CPL page is still there in Windows 11 (but shows Windows 10 on the logo). I suggest you add an option to either open this

изображение

(As currently the case and I like it)

or this
изображение

More info is here:
https://winaero.com/open-classic-system-properties-in-windows-10-version-20h2/

babamohammed2022:
> I like the current behavior of your mod on click on the This PC's properties, but be informed that the old System Properties CPL page is still there in Windows 11 (but shows Windows 10 on the logo). I suggest you add an option to either open this
> 
> изображение
> (As currently the case and I like it)
> 
> or this изображение
> 
> More info is here: https://winaero.com/open-classic-system-properties-in-windows-10-version-20h2/

Regarding the option to choose between sysdm.cpl and classic shell:::{BB06C0E4} on Windows 11, I've decided not to add it for now. The CLSID was the main cause of the infinite loop on Win11, and re-introducing it as an option could bring back instability. On Windows 10 it works fine, so the mod already uses it there. I'll keep this in mind for a future update if a safe way to implement it is found and I'll keep in mind your suggestion. I'll post the fixed code soon. Thanks for the suggestion.

Anixx:
ExplorerPatcher redirects This PC properties to the Win10-style page. You can look into it to see how is it done.

babamohammed2022:
> ExplorerPatcher redirects This PC properties to the Win10-style page. You can look into it to see how is it done.

OK thank you for the suggestion. I will check this in a future update. I've now posted the code that should address some issues.
Fixes:
- Troubleshooting should now redirect to the legacy component in Windows 11 (it works properly on Windows 10)
- The license has been removed
- Notification settings should now redirect to explorer shell:::{05d7b0f4-2121-4eff-bf6b-ed3f69b894d9} (to test press WIN+R then explorer shell:::{05d7b0f4-2121-4eff-bf6b-ed3f69b894d9})
What could not be fixed:
- System tray redirects: I haven't managed to find a proper way to force-redirect Network settings and Audio settings to the CPL counterpart despite them existing. I'll check if there are some missing mappings that cause this or it's something else (this is the same issue for both Win11 and Win10). As confirmed earlier, personalization works properly along with system properties and this is a fundamental step ahead to enhance the mod. If any further documentation related to the legacy components is found, I'll try to improve the code further.

Anixx:
The sound settings and the network settings from the tray icons still lead to the Settings app, this did not change.

This link (accessibility) also leads to the Settings app, even when your mod is enabled:

изображение

The command `explorer shell:::{05d7b0f4-2121-4eff-bf6b-ed3f69b894d9}` (tray settings) works perfectly without your mod as well. The mod does not change this.

babamohammed2022:
> The sound settings and the network settings from the tray icons still lead to the Settings app, this did not change.
> 
> This link (accessibility) also leads to the Settings app, even when your mod is enabled:
> 
> изображение
> The command `explorer shell:::{05d7b0f4-2121-4eff-bf6b-ed3f69b894d9}` (tray settings) works perfectly without your mod as well. The mod does not change this.

Thanks for the report. I've managed to fix the Ease of Access redirect in the Personalization CPL, now this should function exactly as how it did in Windows 7. However, I'll be honest, I still have not managed to make the system tray part work correctly. The tray uses a different mechanism that bypasses the current hooks, and I haven't found a reliable way to intercept it yet. This is documented in the README as a known limitation Furthermore, if there are any issues feel free to report them.
image

Anixx:
Clicking this (Display properties)
изображение

leads to this:
изображение
and not to this:
изображение

Anixx:
> Thanks for the report. I've managed to fix the Ease of Access redirect in the Personalization CPL, now this should function exactly as how it did in Windows 7. 

It now leads to the page "Using computer withoud display", not to the general accessibility page.
To this:
изображение

Instead of this:
изображение

Anixx:
The background link stopped working even though it worked before:
изображение

babamohammed2022:
> Clicking this (Display properties) изображение
> 
> leads to this: изображение and not to this: изображение

OK thanks for pointing this out. Its name should be rundll32.exe display.dll,ShowAdapterSettings 0, I'll try to fix this too along with the background issue in Windows 11 (although I have not experienced this issue on 10 but I'll reach you out soon).

babamohammed2022:
> The background link stopped working even though it worked before: изображение



Added the requested Ease of Access and Display redirects.
They work correctly on Windows 10 21H2. On Windows 11 24H2 I can't get them to work, the internal
mechanism Windows 11 uses for those links bypasses API hooks and it may vary depending on your Windows 11 build.
Everything else works as before on both systems. The not working background issue has been fixed. If there are other issues please report them and I'll attempt to fix them.

Anixx:
Background does not work for me even though it worked before. One click does nothing, double-click opens Settings.

The Display link do not work at all, it is broken. If you cannot fix it, at least, leave in peace, keep redirect to Settings.
In [version 9.8.4](https://raw.githubusercontent.com/babamohammed2022/windhawk-mods/539c5d959618317aba118e91c44425c1f6fb567e/mods/settings-to-control-panel.wh.cpp) the Background link works (just tried again to ensure).

The Accessibility in 9.8.8 seems to work.

babamohammed2022:
> Background does not work for me even though it worked before. One click does nothing, double-click opens Settings.
> 
> The Display link do not work at all, it is broken. If you cannot fix it, at least, leave in peace, keep redirect to Settings.
> In [version 9.8.4](https://raw.githubusercontent.com/babamohammed2022/windhawk-mods/539c5d959618317aba118e91c44425c1f6fb567e/mods/settings-to-control-panel.wh.cpp) the Background link works (just tried again to ensure).
> 
> The Accessibility in 9.8.8 seems to work.

Thanks for reporting this issue. I will double-check the display part again because it worked on Windows 10 21H2 and I will restore the Background logic from the 9.8.4 version and keep the accessibility part. Are there any other issues to address?

babamohammed2022:
@Anixx Good evening. I've double-checked the issue related to the background and it should be fixed for both Windows 11 and Windows 10. The 2 redirects (Ease of Access and Display) should function properly on Windows 10 while it may vary on Windows 11 (since it didn't work on my Windows 11 24H2 but Ease of Access works on your 23H2). I think that the mod's functionality could be vastly limited by Windows 11's hybrid architecture which is less documented than Windows 11 so I think that making the mod Windows 10 primary could be a better choice for users with Windows 11 documented as partially supported because some components use DCOM activation that bypasses ShellExecute hooks entirely. This honest approach should set proper expectations for users. The mod already handles 90+ URI mappings reliably on Windows 10, and that's where it provides the most value. A known issue in Windows 10 is that when I press "Audio settings" on the Audio icon on the system tray settings is opened, if there are any documentations for this online let me know as I have not managed to find any. Thanks again for your help.

Anixx:
No, the Background link does not work now with the latest version. I recommend using Claude because other models are more likely to break code that you do not ask to fix. The Screen link works.

babamohammed2022:
@Anixx Can you share a screenshot of your mod settings? Knowing your configuration would help understand if 
we're experiencing the same issue or different ones.

Anixx:
изображение
The Win11 compatibility mode does not affect.

babamohammed2022:
> изображение The Win11 compatibility mode does not affect.

Could you tell me what happens precisely if you disable the "Smart Personalization Detection" mode compared to when it is enabled?

babamohammed2022:
@m417z @Anixx
Good afternoon. I've just released version 10.0.0 with the following updates:

- SmartPersonalizationDetection is now always active (removed from settings to reduce complexity and to address the background issue reported by Anixx)

- Added several new mappings (Gaming-gamebar → joy.cpl, Windows Update History, Network Places, System Settings, etc.) by using this documentation as a source https://gist.github.com/dbilanoski/3670be6d81b48cde29d40e48e41e0a48

- Improved LaunchTarget handling for control.exe /name commands

- Better Ease of Access redirection using CLSID

This is how the mod works based on Windows versions:
- Windows 10 (tested on 21H2 LTSC) → works almost completely, excluding some edge cases (tray icon audio/network redirections and taskbar settings which is absent in newer versions of Windows 10). These could be improved in the future if documentation about the system tray mechanism becomes available.
- Windows 11 (tested on 24H2) → limited support due to the more complex hybrid architecture. I've had difficulties making everything work properly on this version.

The mod now handles a large amount of URI mappings reliably on Windows 10, which is where it provides the most value. I'm available for any clarifications and I hope that the mod can be merged.
Thank you both for your reviews and testing.
Some examples: (Screenshots taken on Windows 10 21H2 LTSC)
image
image

Anixx:
The Background link still does not work.

Unlike the version 9.8.4.

изображение

I suggest to take that version as a basis and all all your new findings there.

Anixx:
Give an AI (like Claude) the both versions and describe what does not work in the latest one.

babamohammed2022:
> Give an AI (like Claude) the both versions and describe what does not work in the latest one.

I've used 9.8.4 as a base and this time it should be fixed as I've tried to keep the identical approach of this version while adding the new findings like joy.cpl. Let me know if it works and/or if there are any other issues to address.
This is a screenshot related to backgrounds
image
(Tested on Windows 10 21H2 LTSC)

Anixx:
Now all declared features work for me, including the Background link, good job!

Anixx:
When clicking Troubleshooting in the Control Panel, briefly another window is shown but then quickly disappears. Otherwise works well.

babamohammed2022:
> When clicking Troubleshooting in the Control Panel, briefly another window is shown but then quickly disappears. Otherwise works well.

OK thanks for reporting I will check this issue soon as well

babamohammed2022:
Hello @Anixx, I've tried to address the reported issue, let me know if everything still works properly and if the problem has been fixed. Thank you for your feedback.

Anixx:
It is worse now. The window sometimes aoppears, sometimes not, but what is worse, sometimes even Settings app is called. The previous variant was more stable.

babamohammed2022:
> It is worse now. The window sometimes aoppears, sometimes not, but what is worse, sometimes even Settings app is called. The previous variant was more stable.

Hello, I have reverted the code to the previous variant (10.0.1) with a minor description change

Anixx:
In version 10.0.0 (before the revert), the Troubleshooting started instantly, without additional windows. Could you incorporate that into the current version?

babamohammed2022:
> In version 10.0.0 (before the revert), the Troubleshooting started instantly, without additional windows. Could you incorporate that into the current version?

I will try to implement this fix

babamohammed2022:
> In version 10.0.0 (before the revert), the Troubleshooting started instantly, without additional windows. Could you incorporate that into the current version?

I've tried to fix it and it seems to work for me. If it still persists, I'll put this as a documented issue and revert to the previous behavior again.

Anixx:
No difference now. The window still appears on clicking the Troubleshooting. Can you simply copy and paste this code from version 10.0.0?

babamohammed2022:
> No difference now. The window still appears on clicking the Troubleshooting. Can you simply copy and paste this code from version 10.0.0?

OK, I have found the only major difference between the 2 versions was that it had a double mapping that somehow worked. Therefore, I will restore this.
`{L"ms-settings:troubleshoot", L"shell:::{C58C4893-3BE0-4B45-ABB5-A63E4B8C8651}\\historyPage"},`

Anixx:
Yes, works well now. @m417z looks like the issues are resolved.

babamohammed2022:
Good evening, @Anixx @m417z, if possible, please let me know if the mod requires improvements (like a more accurate description, code fixes, etc) or if it's good to merge. Greetings.

Issue #4280, opened 2026-06-03
[Add This PC and Desktop to Nav Top] Crashing Browser tabs + Breaks my Explorer


Veri7ion:
Mod author: @rodboev 
Mod: Add This PC and Desktop to Nav Top
Winver: 25H2 (OS Build 26200.8524)

After installing the mod and configuring the settings I started experiencing tab crash issues in my browser. Active tabs didn't crash but if I tried changing the current page or opening a new tab I'd get a tab crashed error (Zen browser, firefox fork). At first I thought it was maybe my browser so I fully closed it and tried re-opening it from my taskbar but it wouldn't even open anymore.

So I just did a full Windows restart instead.
After entering my password it took a very long time to get past the lock screen and I was then greeted with this error:

<img width="1154" height="973" alt="Image" src="https://github.com/user-attachments/assets/b4aff4d6-28de-47e1-ac8e-aecd8683a43a" />
 
My taskbar was also invincible and Explorer restarts didn't fix it.
Since I hadn't done much else on my PC yet I figured it must be the new Windhawk mod I just added.
So I restarted one more time to videotape it , disabling the mod and restarting my PC again to see if the issue was gone: https://youtu.be/NEX3rQ9DPdg

The mod settings I was using:

<img width="1119" height="1001" alt="Image" src="https://github.com/user-attachments/assets/bf78634b-db6d-412f-baf7-43bac4cd7f3e" />

Unfortunate because I think its a really cool mod!

For anyone else experiencing this issue this is how I fixed it:
Ctrl + Alt + Del → Task Manager    (because Ctrl + Shift + ESC wasn't working)
Click "Run new task" near the top right and enter: C:\Program Files\Windhawk\windhawk.exe
That opens the Windhawk window and you can now disable the mod and restart your PC

rodboev:
Looking into it...

MootjeeNL:
I have the same problem! :(

rodboev:
A last-minute change the injection strategy was likely responsible for these issues. PR #4282 has a new version that should fix this:

- [**v1.2.4:**](https://raw.githubusercontent.com/rodboev/windhawk-mods/9d92575562089fcefcd37502f87329a1c9d1236f/mods/add-virtual-folders-to-nav-top.wh.cpp) Can you try this version and report back?

* * *

### Details:

[...]

PR #4282, opened 2026-06-03
Add This PC and Desktop to Nav Top: defer injection to avoid early conflicts

rodboev:
Changed the injection strategy to hook kernelbase!LoadLibraryExW instead of force-loading ExplorerFrame.dll in Wh_ModInit. Processes that don't use ExplorerFrame.dll no longer get it loaded at startup.

A user reported instability on 25H2 in #4280 (browser tabs crashing, Explorer becoming unresponsive after restart). I haven't been able to reproduce it, but the previous approach loaded ExplorerFrame.dll into every injected process at startup, which is unnecessarily aggressive. Deferring to the kernelbase hook (per @m417z's suggestion in #4159) avoids that entirely.

### DIFF
--- a/mods/add-virtual-folders-to-nav-top.wh.cpp
+++ b/mods/add-virtual-folders-to-nav-top.wh.cpp
@@ -2,7 +2,7 @@
 // @id              add-virtual-folders-to-nav-top
 // @name            Add This PC and Desktop to Nav Top
 // @description     Adds This PC and Desktop to the top of Explorer's nav
-// @version         1.2.6
+// @version         1.2.7
 // @author          Rod Boev
 // @github          https://github.com/rodboev
 // @include         *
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

I'm not sure if `ExplorerFrame.dll` is ever unloaded, but the concern makes sense and the fix is easy.

---

**Removing the force-load also removed the implicit pin on `ExplorerFrame.dll` — consider pinning it explicitly when you hook it late.** In the old code, the `LoadLibraryExW(L"ExplorerFrame.dll", ...)` fallback added a reference that was never released, so once hooked, the module stayed resident for the process lifetime. The new code uses the caller's handle in `CheckLateLoadExplorerFrame` without taking a reference, while `g_explorerFrameLoaded` latches `true` permanently. So if a process loads `ExplorerFrame.dll` transiently — e.g. an app that opens a modern file dialog and later dismisses it, and the module's refcount drops to 0 — the inline hooks are left pointing into an unmapped image, and because the flag stays `true`, the *next* dialog's reload is never re-hooked: the feature silently stops working for the rest of that process's life (and a reload at a different base is a stability risk). This won't bite `explorer.exe` (ExplorerFrame stays loaded there), but it's exactly the non-shell-process case `@include *` exists to cover.

The fix keeps the per-process pin you actually need without reintroducing the system-wide preload you're removing — pin it at the point you commit to hooking, in `CheckLateLoadExplorerFrame` (and equally in the already-loaded branch of `Wh_ModInit`):

```cpp
// keep ExplorerFrame.dll resident for as long as our hooks live in it
HMODULE pin = nullptr;
GetModuleHandleExW(GET_MODULE_HANDLE_EX_FLAG_PIN, L"ExplorerFrame.dll", &pin);
```

<p>

- **The `else` fallback to the kernel32-linked `LoadLibraryExW` is effectively dead code.** `kernelbase.dll` is loaded in every Win32 process, so `GetModuleHandleW(L"kernelbase.dll")` + `GetProcAddress` never fail in practice; the fallback would only ever run in an impossible state, and if it did it would hook the wrong copy (the exact thing this PR fixes). Harmless as a belt-and-suspenders guard, but you could drop it (or leave a one-line comment noting it's unreachable) to make the intent clearer.

</p>
</details>

rodboev:
Got it. ExplorerFrame.dll is now pinned via GetModuleHandleExW(GET_MODULE_HANDLE_EX_FLAG_PIN) in both the late-load and already-loaded paths, and the unreachable kernelbase fallback is removed.

Also addressing #4293 by preventing the tree selection from shifting when the mod inserts or removes items. This was causing navigation in Explorer. Genuine user navigation is unaffected. This really speeds things up.

PR #4284, opened 2026-06-04
Add mod: Unlock Office Files for Upload

loerei:
Adds a new mod that transparently unlocks Office files that are locked by Word/Excel/PowerPoint when another process tries to read them for upload or attachment.

## What it does

When a `.docx`, `.xlsx`, `.pptx` (or legacy variants) is open in Word/Excel/PowerPoint, attempting to upload or attach it in Chrome, Edge, Outlook, Teams, Discord, etc. fails with:

> **This file is in use. Enter a new name or close the file that's open in another program.**

This mod hooks `CreateFileW` in every process. When a read-only open on a tracked extension fails with `ERROR_SHARING_VIOLATION`, it:
1. Makes a temp copy of the file in `%TEMP%`
2. Redirects the open to the copy (caller gets a valid handle, no error)
3. Hooks `CloseHandle` to delete the temp copy automatically when done

No user-visible prompts. No need to close Word first.

## Details

- **Hooks:** `CreateFileW`, `CloseHandle` (via `kernel32.dll`)
- **Target:** `@include *` (all processes — needed since the uploader can be any app)
- **Extensions:** Configurable in Windhawk settings; defaults cover `.docx/.docm/.doc`, `.xlsx/.xlsm/.xls`, `.pptx/.pptm/.ppt`
- **Limitation:** The copy reflects the last saved state; unsaved edits in Word are not included
- **Source repo:** https://github.com/loerei/windhawk-unlock-office-file-upload

### DIFF
--- a/mods/unlock-office-file-upload.wh.cpp
+++ b/mods/unlock-office-file-upload.wh.cpp
@@ -1,12 +1,14 @@
 // ==WindhawkMod==
 // @id              unlock-office-file-upload
 // @name            Unlock Office Files for Upload
-// @description     Transparently copies locked Office files (docx, xlsx, pptx, etc.) to a temp location when an upload or send is attempted while they're open in Word/Excel/PowerPoint. The copy is silently deleted after use.
-// @version         0.1
+// @description     Transparently copies locked Office files (docx, xlsx, pptx, etc.) to a temp location when an upload or send is attempted while they're open in Word/Excel/PowerPoint. The OS deletes the copy automatically via FILE_FLAG_DELETE_ON_CLOSE.
+// @version         0.2
 // @author          Loerei
 // @github          https://github.com/loerei
 // @homepage        https://github.com/loerei/windhawk-unlock-office-file-upload
 // @include         *
+// @exclude         SearchIndexer.exe
+// @exclude         explorer.exe
 // @license         MIT
 // ==/WindhawkMod==
 
@@ -22,8 +24,9 @@ variants) while the file is open in Word/Excel/PowerPoint, you get:
 
 This mod silently intercepts that failure in the uploading process (Chrome,
 Edge, Outlook, Teams, Discord, etc.), makes a temporary copy of the file, and
-redirects the open call to the copy. The copy is automatically deleted once the
-handle is closed.
+redirects the open call to the copy. The OS deletes the copy automatically when
+the caller closes the handle (`FILE_FLAG_DELETE_ON_CLOSE`) — even on process
+crash or abnormal exit.
 
 **What you get:** The uploader receives the last-saved version of the file with
 no prompts and no need to close Word first.
@@ -31,6 +34,12 @@ no prompts and no need to close Word first.
 **Limitation:** The copy reflects the last *saved* state. Unsaved in-memory
 edits in Word are not included — this is inherent to how Windows file locking
 works.
+
+**Scope note:** The mod injects into all processes (required since the uploader
+can be any app). `SearchIndexer.exe` and `explorer.exe` are excluded via
+`@exclude`. Session 0 (system service) processes are skipped at init time. The
+`CreateFileW` hook has near-zero overhead on the success path — it returns
+immediately when the original call succeeds.
 */
 // ==/WindhawkModReadme==
 
@@ -52,73 +61,61 @@ works.
 // ==/WindhawkModSettings==
 
 #include <windows.h>
+#include <cwctype>
 #include <vector>
 #include <string>
 #include <algorithm>
-#include <unordered_map>
-#include <mutex>
-
+#include <shared_mutex>
 
 // ---------------------------------------------------------------------------
-// Settings
+// Settings — guarded by a shared_mutex so LoadSettings (unique/writer lock)
+// and IsTrackedExtension (shared/reader lock, called from the CreateFileW
+// hook on arbitrary threads) don't race.
 // ---------------------------------------------------------------------------
 
 std::vector<std::wstring> g_extensions;
+std::shared_mutex g_extensionsMutex;
 
 void LoadSettings() {
-    g_extensions.clear();
+    std::vector<std::wstring> fresh;
 
     int count = 0;
     while (true) {
         auto key = std::wstring(L"extensions[") + std::to_wstring(count) + L"]";
         PCWSTR val = Wh_GetStringSetting(key.c_str());
-        if (!val || val[0] == L'\0') {
+        // Wh_GetStringSetting never returns NULL; it returns L"" when unset.
+        if (val[0] == L'\0') {
             Wh_FreeStringSetting(val);
             break;
         }
         std::wstring ext(val);
         Wh_FreeStringSetting(val);
-        // Normalise to lowercase
-        std::transform(ext.begin(), ext.end(), ext.begin(), ::tolower);
-        g_extensions.push_back(std::move(ext));
+        std::transform(ext.begin(), ext.end(), ext.begin(),
+                       [](wchar_t c) { return static_cast<wchar_t>(towlower(c)); });
+        fresh.push_back(std::move(ext));
         ++count;
     }
 
-    // Fallback defaults if settings are empty (first run / no settings file)
-    if (g_extensions.empty()) {
+    // Fallback defaults if settings are empty (first run / no settings file).
+    if (fresh.empty()) {
         for (auto* e : {L".docx", L".docm", L".doc",
                         L".xlsx", L".xlsm", L".xls",
                         L".pptx", L".pptm", L".ppt"}) {
-            g_extensions.push_back(e);
+            fresh.push_back(e);
         }
     }
+
+    std::unique_lock lock(g_extensionsMutex);
+    g_extensions = std::move(fresh);
 }
 
-// ---------------------------------------------------------------------------
-// Temp handle tracking
-// ---------------------------------------------------------------------------
-
-struct TempEntry {
-    std::wstring tempPath;
-};
-
-std::unordered_map<HANDLE, TempEntry> g_tempHandles;
-std::mutex g_mutex;
-
-void TrackHandle(HANDLE h, std::wstring tempPath) {
-    std::lock_guard<std::mutex> lock(g_mutex);
-    g_tempHandles[h] = {std::move(tempPath)};
-}
-
-// Returns temp path if handle was tracked, empty string otherwise.
-std::wstring UntrackHandle(HANDLE h) {
-    std::lock_guard<std::mutex> lock(g_mutex);
-    auto it = g_tempHandles.find(h);
-    if (it == g_tempHandles.end())
-        return {};
-    std::wstring path = std::move(it->second.tempPath);
-    g_tempHandles.erase(it);
-    return path;
+static bool IsTrackedExtension(const std::wstring& ext) {
+    std::shared_lock lock(g_extensionsMutex);
+    for (const auto& e : g_extensions) {
+        if (e == ext)
+            return true;
+    }
+    return false;
 }
 
 // ---------------------------------------------------------------------------
@@ -132,42 +129,26 @@ static std::wstring GetExtension(LPCWSTR path) {
     LPCWSTR dot = wcsrchr(path, L'.');
     if (!dot)
         return {};
-    // Make sure the dot is after the last separator
+    // Ensure the dot comes after the last path separator.
     LPCWSTR sep = wcsrchr(path, L'\\');
     if (!sep) sep = wcsrchr(path, L'/');
     if (sep && dot < sep)
         return {};
     std::wstring ext(dot);
-    std::transform(ext.begin(), ext.end(), ext.begin(), ::tolower);
+    std::transform(ext.begin(), ext.end(), ext.begin(),
+                   [](wchar_t c) { return static_cast<wchar_t>(towlower(c)); });
     return ext;
 }
 
-static bool IsTrackedExtension(const std::wstring& ext) {
-    for (const auto& e : g_extensions) {
-        if (e == ext)
-            return true;
-    }
-    return false;
-}
-
-// Build a unique temp path in %TEMP%, e.g. C:\Users\...\AppData\Local\Temp\~ul_12345_filename.docx
-static std::wstring MakeTempPath(LPCWSTR originalPath) {
+// Produces a guaranteed-unique temp path via GetTempFileNameW (uUnique=0).
+// GetTempFileNameW creates an empty placeholder file; CopyFileW with
+// bFailIfExists=FALSE overwrites it safely.
+static bool MakeTempPath(wchar_t (&out)[MAX_PATH]) {
     wchar_t tempDir[MAX_PATH];
     if (!GetTempPathW(MAX_PATH, tempDir))
-        return {};
-
-    // Extract just the filename (without path)
-    LPCWSTR sep = wcsrchr(originalPath, L'\\');
-    if (!sep) sep = wcsrchr(originalPath, L'/');
-    LPCWSTR fname = sep ? sep + 1 : originalPath;
-
-    wchar_t result[MAX_PATH];
-    // Use tick count + thread id for uniqueness; good enough for temp files
-    DWORD uid = GetTickCount() ^ (GetCurrentThreadId() << 16);
-    if (swprintf_s(result, MAX_PATH, L"%s~ul_%08X_%s", tempDir, uid, fname) < 0)
-        return {};
-
-    return result;
+        return false;
+    // Prefix is capped at 3 characters; "ul_" fits exactly.
+    return GetTempFileNameW(tempDir, L"ul_", 0, out) != 0;
 }
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

For item 2, perhaps at least exclude the indexer service and other system processes such as explorer.exe. Also consider excluding processes running in session 0 and non-GUI processes.

---

Clever idea, and the core flow (let the open fail, copy on `ERROR_SHARING_VIOLATION`, redirect) is sound. The main concerns are the system-wide blast radius and the temp-file lifecycle — both can be solved at once by dropping the `CloseHandle` hook.

**1. The global `CloseHandle` hook is a system-wide cost — and it's avoidable.** With `@include *`, you've put a `std::mutex` lock + `unordered_map` lookup on *every* `CloseHandle` call in *every* process on the machine. `CloseHandle` is one of the hottest handle APIs (files, events, threads, registry keys, sockets…), so this is a real, always-on overhead for a feature that fires only on a rare upload of a locked Office file. You don't need the hook at all: open the temp copy with `FILE_FLAG_DELETE_ON_CLOSE` and the OS deletes it when the caller closes the handle. That removes the `CloseHandle` hook, the `g_tempHandles` map, the mutex, and `TrackHandle`/`UntrackHandle` entirely:

```cpp
HANDLE hTemp = CreateFileW_Original(
    tempPath.c_str(), dwDesiredAccess,
    FILE_SHARE_READ | FILE_SHARE_WRITE | FILE_SHARE_DELETE,
    lpSecurityAttributes, dwCreationDisposition,
    dwFlagsAndAttributes | FILE_FLAG_DELETE_ON_CLOSE,
    hTemplateFile);
```

**2. Temp files leak on process crash / abnormal exit.** `Wh_ModUninit` runs only on a normal mod *disable*, **not** when the host process terminates (the uploader is closed, crashes, Explorer restarts, sign-out, etc.). Any process that exits with a redirected handle still open leaves a `~ul_*` file in `%TEMP%` permanently. Same story for handles that get closed through a path that bypasses the kernel32 `CloseHandle` export (`NtClose`, some CRT `_close` paths) — your hook never sees them, so the file is never deleted. `FILE_FLAG_DELETE_ON_CLOSE` (item 1) fixes both cases: the OS deletes the temp on the *final* handle close, including the forced close the kernel does at process teardown. With that in place, the cleanup loop in `Wh_ModUninit` becomes unnecessary too.

**3. `@include *` blast radius and unintended redirects.** Beyond the cost of injecting into every process, the redirect logic fires for *any* read-only open that hits a sharing violation on a tracked extension — not just genuine uploads. The Windows Search indexer, antivirus, backup/thumbnail/preview handlers, etc. would silently be handed a throwaway temp copy instead of the real file (and then the copy disappears). That's a surprising side effect to push system-wide. Consider narrowing `@include` to the executables that actually upload/attach files (browsers, mail and chat clients) rather than `*`. If a curated list is genuinely infeasible for your use case, at least call out the tradeoff in the README. (After item 1, the only remaining global hook is `CreateFileW`, which is cheap on the success path since you return immediately when the original call succeeds — so narrowing scope is the remaining win here.)

<p>

- **Data race on `g_extensions` during a settings change.** `IsTrackedExtension` iterates `g_extensions` from the `CreateFileW` hook with no lock, while `Wh_ModSettingsChanged` → `LoadSettings` does `g_extensions.clear()` + `push_back` on another thread. A reallocation mid-iteration can invalidate the iterator and crash — and with `@include *` this runs in every process at once on every settings change. Either guard `g_extensions` with a mutex, or build a fresh `std::vector` locally in `LoadSettings` and swap it in.
- **Type-safe hooking.** `WindhawkUtils::SetFunctionHook(pfnCreateFileW, CreateFileW_Hook, &CreateFileW_Original)` avoids the `void*` casts and checks the prototype at compile time. (The raw `Wh_SetFunctionHook` form you used is accepted and appears in other mods, e.g. [shadowplay-single-folder.wh.cpp](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/shadowplay-single-folder.wh.cpp#L89), so this is just a preference.)
- **Weak temp-name uniqueness.** `GetTickCount() ^ (GetCurrentThreadId() << 16)` collides easily — `GetTickCount` only changes every ~15 ms, so two opens on the same thread in quick succession get the same name. `GetTempFileNameW` gives you a guaranteed-unique name in one call.
- **`::tolower` on `wchar_t` (lines 82, 141)** is the narrow-char function; passing a `wchar_t` > 127 to it is technically out of range and it won't fold non-ASCII correctly. Use `towlower` (from `<cwctype>`). For your ASCII extensions it happens to work, but `towlower` is the correct call.
- **Redundant NULL check.** `Wh_GetStringSetting` never returns NULL — it returns `L""` when unset — so `!val` in `LoadSettings` is dead; the `val[0] == L'\0'` check alone is enough.
- **Access-rights guard only checks `GENERIC_WRITE`.** A write-intent open that uses `FILE_WRITE_DATA`/`FILE_GENERIC_WRITE` without the generic bit would still be redirected to a temp copy that's then deleted — silent write loss. Edge case (uploads are read-only), but cheap to also exclude those bits if you want to be safe.
- `swprintf_s` relies on a transitive include from `<windows.h>`; add `<cstdio>` to be explicit.

</p>
</details>

<p>

- **Silent stale-data substitution.** The uploader receives the last *saved* version with no indication that it differs from what's on screen. You document this in the README and there's no way to read Word's unsaved buffer, so it's inherent — but a user could unknowingly attach an outdated file. Worth considering whether a one-time log/notice is warranted; not a blocker.
- **Whole-file copy is synchronous and inside `CreateFileW`.** Every locked read of a tracked file copies the entire document into `%TEMP%` while the caller blocks in `CreateFileW`. For a large spreadsheet this is a noticeable stall plus a transient disk-space spike (and a momentary plaintext copy of the document in `%TEMP%`). Acceptable given the goal — just an FYI.
- **Hook coverage.** Only `CreateFileW` is intercepted. Uploaders that open via `CreateFile2`, `NtCreateFile`, or structured-storage paths that bypass the kernel32 export won't be helped, and the kernel32 `CreateFileW` export may miss internal kernelbase callers within a process. For typical app-level uploaders this is fine; just know the boundary.
- **Read-deny holders.** The copy only succeeds when the file is actually readable by a permissive opener (Word's usual `FILE_SHARE_READ` lock). If the holder denies reads outright, `CopyFileW` also fails and you correctly fall back to returning the original sharing violation — the mod simply can't help in that case, which is the right behavior.

</p>
</details>

loerei:
All required and optional points addressed in v0.2. Summary of changes:

**1. Dropped the `CloseHandle` hook entirely — replaced with `FILE_FLAG_DELETE_ON_CLOSE`**

The `CloseHandle` hook, the `g_tempHandles` map, the mutex, and `TrackHandle`/`UntrackHandle` are all gone. Instead the temp file is opened with `FILE_FLAG_DELETE_ON_CLOSE` — the OS handles deletion when the last handle closes, including on process crash or abnormal exit. `Wh_ModUninit` no longer needs a cleanup loop.

**2. Narrowed scope**

- Added `@exclude SearchIndexer.exe` and `@exclude explorer.exe` to the metadata.
- Added a session 0 check at the top of `Wh_ModInit` — if `ProcessIdToSessionId` returns session 0, the mod returns `FALSE` immediately and installs no hooks in that process.

**3. Fixed `g_extensions` data race**

`LoadSettings` now builds a fresh `std::vector` locally, then swaps it in under a `std::unique_lock`. `IsTrackedExtension` takes a `std::shared_lock`. Both share a `std::shared_mutex g_extensionsMutex`.

**4. Fixed `::tolower` → `towlower`**

Both `GetExtension` and `LoadSettings` now use a lambda with `towlower` (from `<cwctype>`) via `static_cast<wchar_t>(towlower(c))`.

**5. Fixed temp-name uniqueness — switched to `GetTempFileNameW`**

`MakeTempPath` now calls `GetTempFileNameW(tempDir, L"ul_", 0, out)` which gives a kernel-guaranteed unique name and creates the placeholder file. `CopyFileW` with `bFailIfExists=FALSE` overwrites it.

**6. Removed redundant `!val` check**

`Wh_GetStringSetting` never returns NULL; the `!val` guard is dropped, leaving only `val[0] == L'\0'`.

**7. Widened write-bits guard**

The access check now covers `GENERIC_WRITE | FILE_WRITE_DATA | FILE_WRITE_ATTRIBUTES | FILE_WRITE_EA | FILE_APPEND_DATA` to avoid silently redirecting write-intent opens that use specific bits instead of the generic alias.

m417z:
### Submission review

---

**No reentrancy guard around `CopyFileW` → unbounded recursion is possible.** The hook is installed on `GetProcAddress(kernel32, "CreateFileW")`. On modern Windows `kernel32\!CreateFileW` is a forwarder to `kernelbase\!CreateFileW`, so `GetProcAddress` resolves the forwarder and returns the **kernelbase** address — which is the implementation `CopyFileW` itself calls internally. So when the hook calls `CopyFileW` (line 214), `CopyFileW`'s own open of the source file re-enters `CreateFileW_Hook`. On the success path that's harmless (the inner open succeeds and returns early), but if `CopyFileW`'s source open *also* fails with `ERROR_SHARING_VIOLATION` on a tracked extension — which is exactly the case for files held with a deny-read lock (common for the legacy `.doc`/`.xls`/`.ppt` formats in your default list, where you genuinely *can't* copy the open file) — the hook recurses into itself with no base case: temp copy → `CopyFileW` → sharing violation → temp copy → … → stack overflow and a crash of the host process. Add a `thread_local` reentrancy guard so the hook passes straight through to the original while it's inside its own `CopyFileW`:

```cpp
HANDLE WINAPI CreateFileW_Hook(...) {
    thread_local bool inHook = false;
    HANDLE h = CreateFileW_Original(...);
    if (h != INVALID_HANDLE_VALUE || inHook)
        return h;
    if (GetLastError() != ERROR_SHARING_VIOLATION)
        return h;
    ...
    inHook = true;
    bool copied = CopyFileW(lpFileName, tempPath, FALSE);
    inHook = false;
    if (!copied) { ... }
    ...
}
```

This also makes the underlying assumption explicit: the whole approach only works when the existing lock permits shared read (so `CopyFileW`'s permissive open succeeds where the caller's restrictive open failed). For the deny-read cases, `CopyFileW` fails too and the mod simply re-surfaces the original error — worth stating in the README as a known limitation alongside the "unsaved edits" one.

<p>

- **Use the typed hook helper.** `WindhawkUtils::SetFunctionHook(pfnCreateFileW, CreateFileW_Hook, &CreateFileW_Original)` drops the `(void*)`/`(void**)` casts and gives you a compile-time signature check against the target. You already have the resolved pointer, so it's a drop-in.
- **The "fallback defaults if empty" block (lines 99–106) is redundant and slightly surprising.** The settings block already declares those nine defaults, so `Wh_GetStringSetting` returns them on a fresh install; the only thing the fallback adds is that a user who *intentionally* clears the list to disable tracking gets the defaults silently re-added, with no way to track nothing. Consider dropping it (or honoring an empty list).
- **`SetLastError(ERROR_SUCCESS)` before returning the redirected handle (line 237)** hides the `ERROR_ALREADY_EXISTS` signal that `CreateFileW` sets for `OPEN_ALWAYS`/`CREATE_ALWAYS`. The redirect path is read-only/`OPEN_EXISTING` in practice so it's moot today, but it's a latent inaccuracy if the disposition filter ever loosens — preserving the original last-error from the temp open would be safer.

</p>
</details>

<p>

- **Stability surface.** Any defect in the hook (including the recursion above) doesn't affect one app — it can hang or crash everything from games to compilers to security software. The hook can't tell "Chrome reading a file for upload" from "OneDrive/Dropbox syncing", "a backup tool reading the file", or "Defender scanning it". All of those routinely open files and hit `ERROR_SHARING_VIOLATION` on a locked Office doc, and all of them would now silently receive a *different file* (a temp path that vanishes on close) instead of the real one — e.g. a sync client could upload/track the temp copy, a backup could store the wrong path. The `@exclude SearchIndexer.exe`/`explorer.exe` pair doesn't begin to cover this set.

  Consider narrowing `@include` to the realistic uploader set (browsers, mail/chat clients — `chrome.exe`, `msedge.exe`, `firefox.exe`, `outlook.exe`, `teams.exe`, `discord.exe`, …) rather than `*`. That bounds the blast radius and eliminates the "wrong process gets a temp copy" class of bugs.

- **The uploader sees the temp file's identity, not the original's.** The redirected handle refers to `ul_XXXX.tmp` in `%TEMP%`. Apps that derive the upload name from the *path string* they passed (most browsers, which carry the name from the file picker separately) are unaffected. But any caller that asks the handle for its name — `GetFinalPathNameByHandle`, `GetFileInformationByHandleEx(FileNameInfo)` — will get `ul_XXXX.tmp` and may upload/attach the file under that name. Worth a line in the README, and a reason to keep the prefix recognizable.
- **`CopyFileW` runs synchronously inside the hooked `CreateFileW` call.** For a large file (a media-heavy `.pptx` can be hundreds of MB) the uploader's open call blocks for the full duration of the copy. It's on the rare failure path so it's an acceptable cost, and there's no clean alternative (you need a readable snapshot), so this is just an FYI.
- **Coverage is limited to callers that go through `CreateFileW`.** Apps that open files via `CreateFile2` (some modern/UWP/WinRT paths) bypass this hook entirely, so the unlock silently won't apply there. Not necessarily worth chasing, but good to know for the "why didn't it work in app X" reports.

</p>
</details>

loerei:
Addressed in v0.3.

**1. Reentrancy guard added**

`thread_local bool inHook` is declared at the top of `CreateFileW_Hook`. The guard is set before `CopyFileW` and cleared immediately after, so any `CreateFileW` call that re-enters the hook while a copy is in progress passes straight through to the original and returns, breaking the recursion:

```cpp
thread_local bool inHook = false;

HANDLE h = CreateFileW_Original(...);
if (h != INVALID_HANDLE_VALUE || inHook)
    return h;
...
inHook = true;
bool copied = CopyFileW(lpFileName, tempPath, FALSE);
inHook = false;
```

This also makes the underlying assumption explicit in code: `CopyFileW`'s internal opens are allowed to pass through; only the *outer* caller's sharing-violation open is intercepted.

**Deny-read limitation documented**

Added a "Known Limitations" section to the README that covers both the unsaved-edits case and the deny-read case (legacy `.doc`/`.xls`/`.ppt` held with a deny-read lock — `CopyFileW` fails too and the original error is re-surfaced).

---

Optional items also taken:

- **Fallback defaults removed.** An empty extension list is now honored rather than silently overridden. The settings block defaults cover a fresh install via `Wh_GetStringSetting`.
- **`SetLastError(ERROR_SUCCESS)` removed.** The last-error from the `CreateFileW_Original` call on the temp file now propagates naturally.

---

On `@include *` narrowing (functionality note): keeping `*` for now because narrowing to a fixed list of executables would silently fail for any uploader not on the list (e.g. corporate intranet portals, custom tools, file managers with built-in upload). The reentrancy guard + write-bits guard + extension filter together keep the active-interception surface narrow: the hook only substitutes when a read-only open of a tracked extension hits a sharing violation, which is a fairly specific condition. Happy to revisit if this remains a blocker.

m417z:
Thanks. I merged it since I believe there are no blocking issues. Below is a fresh review in case you'd like to make additional improvements in the future. Specifically, injecting only into Office apps, if possible, could be significantly less intrusive.

---

- **Blast radius.** `CreateFileW` is one of the hottest paths in the OS — every file open in every process now routes through the trampoline. The success path is cheap, but a bug in this hook destabilizes the entire system, not one app. The maintainer is wary of `@include *`; combined with a hook this central, it's a hard sell.
- **It can't tell an "upload" from any other read.** The hook only sees "a read-only open of a `.docx` failed with a sharing violation." That fires for *any* reader that hits Word's lock — backup software, antivirus, **OneDrive/sync clients**, search/preview tooling, etc. — and silently hands them a copy instead of the real failure they're designed to handle. A backup or sync client quietly capturing a stale copy is a real data-correctness hazard. It's telling that you already had to `@exclude explorer.exe`; the same reasoning applies to many processes you can't enumerate, which is the sign that the scope is wrong.

**Strongly consider fixing the root cause instead.** The actual problem is that Word/Excel/PowerPoint open the document with a share mode that denies readers. Rather than injecting into all processes and copying, hook `CreateFileW` (or `CreateFile2`) only in the Office apps — `@include winword.exe excel.exe powerpnt.exe` — and OR `FILE_SHARE_READ` into the document open. Then *any* uploader, in any process, can open the live file directly: no temp copies, no system-wide hook, scope limited to three trusted processes that only run from `Program Files`, and it matches the "root cause over workaround" preference. Adding only `FILE_SHARE_READ` (not write) keeps Word's protection against concurrent *writers* intact, so the risk is low. Worth prototyping to confirm Office uses `CreateFileW`/`CreateFile2` for the main document open and that it doesn't immediately re-lock.

If you keep the redirect approach, at minimum narrow `@include` to the realistic uploader set (browsers, mail, chat clients) so the system-wide hook and the unintended-process redirects go away. Users can add more targets if needed.

<p>

- Prefer the type-safe `WindhawkUtils::SetFunctionHook(CreateFileW, CreateFileW_Hook, &CreateFileW_Original)` over the raw `Wh_SetFunctionHook` + `void*` casts. The imported `CreateFileW` symbol resolves through kernel32's forwarder to the same kernelbase implementation you reach via `GetProcAddress`, so the explicit `GetModuleHandleW`/`GetProcAddress` dance can be dropped.

</p>
</details>

<p>

- **Only `CreateFileW` callers are caught.** Apps that open files via `CreateFile2` or `NtCreateFile`/`NtOpenFile` directly (some modern apps, and parts of Chromium-based ones) won't be helped — the hook never fires for them.
- **`GetFinalPathNameByHandleW` returns the temp path.** A handle to `ul_XXXX.tmp` no longer maps back to the original document. An uploader that re-derives the filename from the handle (rather than using the path it already had) would upload `ul_XXXX.tmp` instead of `report.docx`. Many uploaders use the path they already hold, so this is often fine, but it's a behavioral edge.
- **Retrying readers trigger repeated full-file copies.** An app that polls the locked file (reopen-on-failure loops are common) causes a full `CopyFileW` of a potentially large Office file on every attempt, since each `CreateFileW` failure copies again. Combined with `@include *`, a single misbehaving reader can generate a lot of disk churn.
- **`dwDesiredAccess == 0` (query-only) opens also get redirected** and copied, even though no data is read. Harmless but wasteful; you could require at least a read bit before redirecting.

</p>
</details>

PR #4287, opened 2026-06-04
Fix Monitor Sleep Button tray icon scroll hit testing

### LINKED ISSUE #4148: [Monitor Sleep Button] Scroll wheel hit area too large — triggers over adjacent tray icons
Mod author: @SilverAmd 

  Mod: Monitor Sleep Button

  ---

Monitor Sleep Button responds to scroll wheel events when the cursor is over adjacent tray icons, not just its own.
This causes unintended interactions with other tray mods.

Specifically, when another mod (for example my AudioSwap mod) is installed and its tray icon sits next to Monitor Sleep Button's icon,
scrolling over AudioSwap's icon changes the Monitor Sleep Button countdown and triggers AudioSwap.
both mods fire simultaneously.

  Reproduction Steps

  1. Install Monitor Sleep Button
  2. Install any other mod that adds a tray icon adjacent to Monitor Sleep Button
  3. Scroll the mouse wheel while hovering over the other mod's icon
  4. Observe: Monitor Sleep Button countdown changes even though the cursor is not over it

thank you, and have a great day :)

SilverAmd:
Fixes #4148.

Monitor Sleep Button now uses the actual Shell_NotifyIconGetRect rectangle for mouse wheel hit testing instead of expanding the tray icon rectangle. The rectangle is refreshed before handling mouse wheel events.

Tested with an adjacent tray icon. Monitor Sleep Button now only changes the countdown when scrolling over its own tray icon.

## Changelog

* Changelog item 1...
* Changelog item 2...

## Mod authorship

This mod was created by:

- - [x ] The submitter, with AI assistance
- - [ ] Claude
- - [x ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/monitor-sleep-button.wh.cpp
+++ b/mods/monitor-sleep-button.wh.cpp
@@ -781,7 +781,6 @@ LRESULT CALLBACK LowLevelMouseProc(int nCode, WPARAM wParam, LPARAM lParam) {
         short delta = static_cast<short>(HIWORD(ms->mouseData));
         int direction = (delta > 0) ? 1 : -1;
 
-        RefreshTrayIconRect();
         BOOL inside = IsPointNearTrayIcon(ms->pt);
 
         if (inside && g_hwnd) {
--- a/mods/monitor-sleep-button.wh.cpp
+++ b/mods/monitor-sleep-button.wh.cpp
@@ -18,6 +18,12 @@
 
 Adds a small tray icon for quickly turning off the monitor after a configurable countdown.
 
+## Preview
+
+![Monitor Sleep Button preview 1](https://i.imgur.com/sztL2TK.png)
+
+![Monitor Sleep Button preview 2](https://i.imgur.com/euxiJiQ.png)
+
 ## Features
 
 - Tray icon for quick monitor power-off
@@ -667,9 +673,9 @@ void RefreshTrayIconRect();
 bool IsPointNearTrayIcon(POINT pt) {
     RECT rect = g_trayIconRect;
 
-    // Tray icon rect can be slightly off because of DPI/taskbar scaling.
-    // Expand it a bit so mouse wheel over the visible icon is still detected.
-    InflateRect(&rect, 32, 48);
+    if (IsRectEmpty(&rect)) {
+        return false;
+    }
 
     return PtInRect(&rect, pt) != FALSE;
 }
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

---

Removing the `InflateRect` is the right fix for #4148 — `Shell_NotifyIconGetRect` already returns the exact screen rect and `ms->pt` is in screen coordinates, so the un-inflated `PtInRect` is correct and won't catch neighboring icons. One issue with the other half of the change:

**Don't call `Shell_NotifyIconGetRect` inside the `WH_MOUSE_LL` hook.** The new `RefreshTrayIconRect()` call in `LowLevelMouseProc` runs on *every* `WM_MOUSEWHEEL` event system-wide — i.e. every scroll in any window, anywhere, not just over the tray icon. Low-level mouse hooks run synchronously on the installing thread and block delivery of all input until they return, so the rule is to do as little as possible in them and never call anything that can block. `Shell_NotifyIconGetRect` is not cheap: the notification area is hosted by `explorer.exe`, so the call does cross-process work to fetch the rect. Putting it in the hot path of the global wheel hook adds that latency to every scroll and, if the shell thread is busy, can stall input for the duration of the call.

The good news is the refresh is unnecessary there. `g_trayIconRect` is already refreshed on `WM_MOUSEMOVE` over the icon (`WindowProc`, the `WM_TRAYICON`/`WM_MOUSEMOVE` case), as well as in `AddTrayIcon`/`UpdateTrayIcon`. To scroll over the icon the cursor must be over it, which generates those `WM_MOUSEMOVE` callbacks first, so the cached rect is already current at scroll time. Just drop the in-hook refresh and hit-test against the cached rect:

```cpp
short delta = static_cast<short>(HIWORD(ms->mouseData));
int direction = (delta > 0) ? 1 : -1;

BOOL inside = IsPointNearTrayIcon(ms->pt);  // cached rect, already kept fresh on WM_MOUSEMOVE
```

If you ever need to guard against the icon moving while the cursor sits still (rare), refresh the rect on the UI thread (e.g. on a taskbar layout change) rather than inside the hook.

<p>

- The new `IsRectEmpty(&rect)` early-return is redundant for correctness: `PtInRect` already returns `FALSE` for a zero-area rect, so an empty `g_trayIconRect` can never match. It's harmless and arguably clearer, so feel free to keep it.
- With the inflation gone, `IsPointNearTrayIcon` is now an exact containment test — `IsPointInTrayIcon` would read a touch more accurately, though this is pure naming.

</p>
</details>

SilverAmd:
Thanks, fixed.

I removed the RefreshTrayIconRect call from LowLevelMouseProc. The hook now only tests against the cached tray icon rectangle, which is refreshed from the tray icon mouse move callback and after tray icon add/update.

I kept the IsRectEmpty check since it's harmless and makes the empty-rect case explicit.

All checks have passed again.

PR #4192, opened 2026-05-27
Add Explorer Nav DragOver Fix mod (nav hover-expand jump + wheel during drag)

tonythethompson:
## Summary

Adds Explorer Nav DragOver Fix (explorer-nav-dragover-fix) for explorer.exe.

- Fixes two long-standing File Explorer drag-and-drop annoyances that stock Explorer don’t address together: nav-pane jump on hover-expand and mouse-wheel input swallowed during drag. One mod, both panes, drag stays alive.

- Jump on hover-expand — Dragging over top-level nav items (drives, pinned folders, “This PC”) triggers auto-expand; Explorer often scrolls the pane and the row you were aiming at disappears. The mod pins the first visible item and keeps the drop highlight on the captured parent row until the cursor leaves that row. Deeper folders still expand normally without the anchor.
- Wheel during drag — Restores vertical (and horizontal) wheel scrolling over the left navigation tree and the right file area while a drag is active, so you can reach off-screen drop targets without releasing. Optional setting restores native no-scroll behavior.
- Demo: [explorer-nav-dragover-fix-demo.gif](https://raw.githubusercontent.com/tonythethompson/windhawk-mods/main/mods_assets/explorer-nav-dragover-fix-demo.gif) (submitter fork)

## Notes for reviewers
- No existing Windhawk mod targets this combination. Related discussion: blocking programmatic nav expand ([windhawk-mods#3789](https://github.com/ramensoftware/windhawk-mods/issues/3789)) — this mod instead fixes drag-hover scroll jump and wheel during drag.
- File-pane scrolling deliberately does not post wheel/scroll messages to SHELLDLL_DefView during DoDragDrop — that path scrolled in early testing but ended the drag. Production path uses UI Automation ScrollPattern on the content host (Win11 Details / ItemsView / DirectUI on 25H2), with SendMessageTimeout wheel only as fallback on the host — including classic SysListView32 (Large icons / List) after UIA is tried first.

## Implementation notes

- DoDragDrop hook — drag generation + per-thread WH_MOUSE for the modal drag loop.
- Nav anti-jump — SetWindowSubclass on nav SysTreeView32; intercepts/lies selected TVM_* / timed WM_VSCROLL swallow after suppressed jumps; EVENT_OBJECT_CREATE helper thread for new cabinets.
- Wheel forwarding — WM_MOUSEWHEEL / WM_MOUSEHWHEEL to target under cursor:
Nav: SysTreeView32 when over the tree.
[...]

PR #4294, opened 2026-06-04
explorer-nav-dragover-fix: v1.2 edge-scroll and transient folder collapse

tonythethompson:
## Summary

Follow-up to the merged **Explorer Nav DragOver Fix** mod ([#4192](https://github.com/ramensoftware/windhawk-mods/pull/4192)). Bumps the mod to **v1.2.4** with:

- **Accelerated edge-scroll during drag** — scroll speed scales with cursor distance from the top/bottom of the nav pane (configurable band and max lines).
- **Transient folder collapse** — folders that were collapsed before the drag and auto-expanded only because of drag-hover are collapsed again when the drag ends (staggered animation optional). A successful drop keeps the branch that contained the last nav-pane drop target expanded.
- **Reliable collapse across branch changes** — separates **drag session id** (one per outer drag) from **viewport generation** (bumped on nav wheel during drag). Fixes folders left expanded after scrolling between drives or sibling branches during the same drag.

## New settings

| Setting | Default | Purpose |
|---------|---------|---------|
| `accelerateEdgeScrollDuringDrag` | `true` | Faster edge-scroll vs native one-line crawl |
| `edgeScrollBandPx` | `48` | Pixel band for acceleration (8–128) |
| `edgeScrollMaxLines` | `6` | Max lines per edge-scroll tick (1–24) |
| `collapseTransientExpansions` | `true` | Fold drag-opened folders when drag ends |
| `transientCollapseStaggerMs` | `140` | Delay between collapse steps; `0` = instant (0–800) |

Existing anti-jump and wheel-during-drag settings are unchanged.

## Implementation notes

- Tracks transient expansions with `itemLParam` and re-resolves items after staggered `TVM_EXPAND` collapse if handles go stale.
- Staggered collapse uses `WM_TIMER`, `WM_SETREDRAW`, and a reconcile pass for missed folders.
- Edge-scroll adds extra line steps on top of Explorer’s native edge scroll (with bottom/top outside-slack for clipped cursors).
- Preserve-on-drop uses ancestor check: keep expansion nodes that contain the drop target.

## Testing

- Windows 11, File Explorer navigation pane.
- Drag across **This PC → Drive A → Drive B** with wheel scroll between drives; cancel or drop outside branch — both transient expansions collapse.
- Drag between sibling folders under the same parent without scrolling — both collapse on cancel.
- Successful drop into a nested folder — ancestors on the drop path stay expanded; unrelated drag-opened branches collapse.
- Edge-scroll at top/bottom of nav pane feels faster than native; settings off restore native behavior.
- `transientCollapseStaggerMs` at `0` and at `180–250` — instant vs calmer stepped collapse.

## Changelog

**1.2.4**

- Fix transient collapse skipped after nav wheel scroll (drag session id vs generation).
- Fix preserve-on-drop ancestor check for successful drops.

**1.2.0–1.2.3** (included in this PR)

- Edge-scroll acceleration during drag.
- Transient expansion tracking, staggered collapse, reconcile pass, `lParam` re-resolution.

## Mod authorship

- [ ] Someone else (please credit them in the PR)

### DIFF
--- a/mods/explorer-nav-dragover-fix.wh.cpp
+++ b/mods/explorer-nav-dragover-fix.wh.cpp
@@ -2,7 +2,7 @@
 // @id           explorer-nav-dragover-fix
 // @name         Explorer Nav DragOver Fix
 // @description  Stops File Explorer's nav pane from jumping when a folder auto-expands during a drag, restores mouse-wheel scrolling during drag, accelerates edge-scroll, and optionally collapses drag-opened folders when the drag ends
-// @version      1.2.5
+// @version      1.2.6
 // @author       tonythethompson
 // @github       https://github.com/tonythethompson
 // @include      explorer.exe
@@ -43,10 +43,9 @@ Two fixes for File Explorer's drag-over behavior:
    the edge (instead of Explorer's slow one-line-at-a-time crawl).
 
 4. **Optional transient-folder cleanup.** Folders that auto-expand only
-   because you dragged over them can be collapsed again when the drag ends
-   (cancel, or drop outside that branch). A successful drop keeps the branch
-   that contained the last nav-pane drop target expanded. Stagger delay
-   collapses them one step at a time for a smoother fold-up (configurable).
+   because you dragged over them are collapsed again when the drag ends
+   (cancel, or drop outside that branch). A successful drop keeps ancestors
+   on the path to the last nav-pane drop target expanded.
 
 ## Demo
 
@@ -143,14 +142,6 @@ normal use.
     when the drag ends, unless you dropped into that branch. Cancelled drags
     collapse all such folders. Disable to leave the nav pane as Explorer
     left it.
-- transientCollapseStaggerMs: 140
-  $name: Collapse stagger delay (milliseconds)
-  $description: >-
-    Base delay between each folder collapse when cleaning up drag-opened
-    items. Higher values look calmer (try 180-250 if steps still feel like a
-    blur). When several folders need collapsing, the mod adds a short lead-in
-    and slightly longer gaps for deeper items. 0 collapses everything
-    instantly. (0-800)
 */
 // ==/WindhawkModSettings==
 
@@ -218,7 +209,6 @@ struct Settings {
     std::atomic<int> edgeScrollBandPx{48};
     std::atomic<int> edgeScrollMaxLines{6};
     std::atomic<bool> collapseTransientExpansions{true};
-    std::atomic<int> transientCollapseStaggerMs{140};
 };
 
 Settings g_settings;
@@ -265,12 +255,6 @@ struct TransientCollapseTarget {
     LPARAM itemLParam = 0;
 };
 
-struct StaggeredCollapseSession {
-    std::vector<TransientCollapseTarget> remaining;
-    // Full target set for a final reconcile pass (handles stale HTREEITEM).
-    std::vector<TransientCollapseTarget> reconcile;
-};
-
 struct NavDropTargetRecord {
     HWND hwndTree = nullptr;
     HTREEITEM hItem = nullptr;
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

The main note is about the mod's functionality. I didn't inspect the code's logic manually, please check the review and and address it, or let me know if it's not correct.

---

One correctness issue with the new "preserve on drop" feature:

**Preserve-on-drop collapses the drop target's ancestors instead of preserving them (`IsSameOrDescendant` arguments reversed).** In `FlushDragTreeState`:

```cpp
if (successfulDrop && entry.hwndTree == preserveTree && preserveItem &&
    IsSameOrDescendant(entry.hwndTree, entry.hItem, preserveItem)) {
    continue;  // keep this expansion
}
```

`IsSameOrDescendant(tree, hItem, possibleAncestor)` walks **up** from `hItem` and returns true when `hItem` is `possibleAncestor` *or a descendant of it*. So this skips collapse only for transient expansions that are the drop target **or sit below it** — i.e. it preserves *descendants* of the drop target and collapses its *ancestors*.

But the feature (and your PR notes — "ancestors on the drop path stay expanded") needs the opposite: keep the expansions that **contain** the drop target. Concretely, drag a file, hover `C:` (auto-expands), move into `C:\Projects` (auto-expands), drop on `Projects`. Both `C:` and `Projects` are transient expansions; the drop target is `Projects`. The check preserves `Projects` but collapses `C:` — collapsing `C:` hides `Projects`, so right after a successful nested drop the whole path folds up and you lose sight of where you dropped. (It only looks correct when the ancestor was already expanded before the drag, so it wasn't a transient expansion in the first place — which may be why testing missed it.)

Fix is to swap the two item arguments so you preserve an expansion when it is an ancestor-or-self of the drop target (drop target is same-or-descendant of the expansion):

```cpp
    IsSameOrDescendant(entry.hwndTree, preserveItem, entry.hItem)) {
```

<p>

- **The staggered-collapse machinery is very heavy for a cosmetic cleanup.** The transient-collapse feature pulls in ~700 lines: per-window stagger sessions, a depth-scaled timer schedule (`ComputeInitialCollapseTimerMs` lead-in + `ComputeNextCollapseTimerMs` depth bonus), a 4-pass `ReconcileTransientCollapses`, `lParam` re-resolution via full-tree DFS, and two parallel code paths (instant vs staggered). For "fold the drag-opened folders back when the drag ends," a single collapse pass (deepest-first, which you already do) would deliver almost all of the benefit at a fraction of the surface area. Given the AI-assisted authorship, it's worth asking whether the animated stagger earns its complexity — collapsing instantly is simpler to reason about and to maintain. Not blocking, just a maintainability call.

- **`protectedMaxDepth` struct default disagrees with the settings/README default.** The `Settings` struct initializes `std::atomic<int> protectedMaxDepth{2};` while the settings block and README both document the default as `3`. `LoadSettings()` always overwrites it from the setting, so the `{2}` is never actually used — but aligning it to `{3}` avoids confusion for the next reader.

- **`Wh_ModUninit` performs the collapse animation during teardown via unbounded `SendMessageW`.** `DrainStaggeredCollapseForWindow` → `CollapseTargetsNow`/`ReconcileTransientCollapses` issue `SendInternal` (plain `SendMessageW`, no timeout) to each tree from the uninit thread. Everywhere else this mod is careful to bound teardown (`SMTO_ABORTIFHUNG`, the 100 ms drain wait, leaking rather than blocking), so doing potentially-blocking cross-thread sends here is a bit inconsistent — and doing a visible fold-up *as the mod is being disabled* is questionable UX. Consider just killing the timers and clearing `g_staggeredCollapseSessions` on unload rather than flushing the collapses.

</p>
</details>

<p>

- **A pending staggered collapse isn't cancelled when a new drag begins, so it can fire mid-drag.** `g_staggeredCollapseSessions` is keyed by HWND and is not touched on the `0 -> 1` drag edge in `DoDragDrop_Hook` (only `PurgeStaleDragTreeStateNotInSession` runs, which clears `g_transientExpansions`/`g_lastNavDropTarget` but not the timer sessions). If the user finishes a drag and immediately starts another before the stagger timer drains, the leftover timer keeps collapsing the previously-recorded folders — which can collapse a folder the user is actively hovering in the *new* drag. Draining/cancelling pending staggered collapses on the new drag's `0 -> 1` edge would close this.

</p>
</details>

tonythethompson:
Thanks for the review — addressed in `787e6de` (v1.2.5) and `1218aa3` (v1.2.6):

**Blocking:** Swapped preserve-on-drop to `IsSameOrDescendant(tree, preserveItem, entry.hItem)` so we walk up from the drop target and keep transient expansions that are ancestors of (or equal to) the drop target. You were right that the previous argument order preserved descendants and collapsed ancestors.

**Optional improvements (from review):**
- `protectedMaxDepth` struct default aligned to `3`
- **Stagger removed entirely** (`1218aa3`) — took your maintainability point; dropped `transientCollapseStaggerMs`, timers, and stagger sessions. Drag-end cleanup is instant deepest-first collapse plus the reconcile pass only (~200 lines removed). No collapse work on mod unload anymore since there are no pending timers.

PR #4295, opened 2026-06-04
Fluid Window Engine Pro

kivsak:
# Fluid Window Engine Pro ✨

👉 **PLS GIB MONI FOR 2X SPEED BOOST:** https://www.patreon.com/c/kivsak

Tired of the stiff, boring default animations for your applications? Inspired by macOS, KDE Plasma, and Material Design, this cinema-grade mod completely overhauls how your app windows minimize, restore, and close.

Unlike older, buggy alternatives, this mod features a completely rewritten physics-based engine that guarantees zero visual artifacts, perfectly preserving drop shadows, modern rounded corners, and transparency.

### 9 Animation Styles (Minimize, Restore & Close):
* 🌀 **Genie:** The iconic fluid, organic stretch that pulls app windows perfectly into your taskbar.
* 💧 **Fluid:** Soft-body droplet physics — tunable from rigid rubber to flowing oil.
* 📐 **Scale:** A clean, modern, and snappy shrink-to-icon transition.
* 🔍 **Zoom:** Focal-point scaling with optional inertial overshoot.
* 💨 **Swipe:** Material 3 directional fly-off, Android-smooth.
* 📺 **CRT:** A nostalgic retro cathode-ray tube screen-off sequence.
* ⚫ **Black Hole:** A true particle vortex with motion blur.
* ⏳ **Sand:** Physics grain dissolve — crumble or firework burst.
* 💎 **Glass:** Irregular polygon shatter with rotation and gravity.

### 🛠️ Advanced Engine & Physics:
* **Language-Independent Targeting:** Animations fly exactly into your app's icon on any Windows locale. Matches via exe path, alias maps, and fuzzy title scoring.
* **Multi-Monitor First-Class:** Each animation lives on the correct monitor, fully supporting secondary taskbars.
* **Tunable Physics:** Customize liquidity, particle count, gravity, and speed for each effect directly in the Windhawk settings.
* **Crystal Clear Quality:** Per-effect bounding box rendering ensures smooth 60 FPS even on 4K displays, without full-screen blits.

### v2.0.0
* **Global Overhaul:** Upgraded to "Fluid Window Engine Pro".
* **New Feature:** Added full support for Close window animations.
* **Targeting Rewrite:** Total taskbar targeting rewrite for accurate multi-language support.
* **New Effects:** Introduced Fluid, Zoom, Swipe, Black Hole, Sand, and Glass animations.
* **Optimizations:** Fixed fluid visual seams, ~3x faster Glass inner rasterizer, and adaptive particle resolution for high-DPI (1440p/4K) monitors.
* **Physics Upgrades:** Added true ballistic structure for Sand (explosion mode) and clean particle separation for Black Hole.

## 👤 Mod authorship

This mod was created by:
* [ ] The submitter, without AI assistance
* [ ] The submitter, with AI assistance
* [x] Claude (refactoring)
* [ ] ChatGPT
* [ ] Gemini
* [ ] Another AI (please specify):
* [x] Other (please specify): kivsak (original engine)


Anixx:
Great mod! I have problems with with folder windows under Classic theme. Non-Explorer windows are well.

The folder windows are restored white, the icons appear on mouse hover.

изображение

kivsak:
@Anixx 
Ah, now I completely understand the issue, thanks for the screenshot! 

This happens because the mod currently creates a 32-bit transparent icon using the alpha-channel to hide the arrow. While the modern Desktop view handles this perfectly, legacy Windows components and classic menus (like the one in your screenshot) don't support alpha-channel transparency for shortcut overlays and render the empty space as a solid square/block. 

Restarting `explorer.exe` or clearing the icon cache can sometimes force Windows to redraw it properly, which is likely why the glitch disappeared on my end. 

I'll look into the code and try to fix this within the next few days by using a standard 1-bit hardware mask for transparency, which should completely eliminate the block artifact in legacy menus.

Anixx:
I am not sure it is about what u described.
There is no non-transparent block, the icons appear on hover one-by-one, so it is related to refreshing of the window.
изображение

kivsak:
@Anixx 
Oh, I see.
I'll try to fix that soon.

Have you tried restarting Explorer?

Anixx:
Yes, it does not help. I am sure it is unrelated to icon cache, etc.

Anixx:
The window also fully restores on resize.

Anixx:
I also noticed that during the restoration animation itself the window is fully and correctly drawn, it becones white at the final moment at the end of animation.

kivsak:
@Anixx 
lol, wait, aren't we talking about a mod with window animations?
I must have gotten confused and thought we were talking about a mod that hides icons

kivsak:
@Anixx 
I got the threads mixed up

kivsak:
@Anixx   Is this a bug that occurs with any animations?
I'm actually surprised that this mod works with the Classic theme overall

Do you experience any flickering when you open or close the window?

Anixx:
Your animations work very well in all apps in Classic theme except File Explorer. For instance, works with notepad.exe or whatever. No flickering is experienced.

Anixx:
I found, the issue is the conflict with the mod `Explorerframe fixes for Win11 22H2+`, particularly, its setting `Classic CabinetWClass background color`, the issue occurs only when it's enabled.

Anixx:
Unfortunately, when that option is disabled, its effect does not work, so folder windows appear with black backgrounds first.

kivsak:
@Anixx  I'll see what I can do about it

I'm currently adding some additional animations.
I wonder if this bug will affect all the animations.

Anixx:
The built-in window animations have no issues with that option.

Anixx:
Currently it affects all animations in your mod.

Anixx:
To be more precice, the animations themselves are not affected, the whiting-out happens after the animations.

Anixx:
The mod [`Genie Animation Mod`](https://windhawk.net/mods/genie-minimize-animation) has not this issue, you may try to look into it.

kivsak:
@Anixx 
I have a theory as to why this happens and why it doesn't occur in the Genie Animation Mod.
Essentially, a screenshot of the window is taken, it's superimposed over the actual window, and the actual window closes very quickly—that's how it works when minimizing. When maximizing, the opposite happens.
This enables such complex animations without flickering and with transparency.
Perhaps for some reason the screenshot remains in the internal window and isn’t cleared from the cache, but why exactly on the screen, where my PC is—that’s a mystery.
And what happens if you grab the window and move it back and forth?

Anixx:
Of course, moving window does not change anything, the window needs refresh.

As to your screenshot theory, I doubt it, since your animations are correct during the animation itself (if the window was correct before minimizing), this tells the screenshot taken is correct all the way. The whiting-out happens after the animation.

Anixx:
The issue does not depend on the Classic theme, it only happens due to the incompatibility with the mentioned mod (Explorerframe fixes), so you can test it yourself:

изображение

m417z:
### Submission review

---

This is a capable, ambitious mod, but it overlaps heavily with an existing one and has a few real correctness/stability problems that affect users in normal use.

**1. All rendering is hard-coded to the primary monitor — broken on multi-monitor / mixed-DPI setups.** Every animation path sizes its surfaces with `GetSystemMetrics(SM_CXSCREEN)` / `SM_CYSCREEN` and creates the ghost overlay at origin `(0,0)` (e.g. `RunShrinkToIcon` line 175-176/190-191, `RunCRT` line 291/295-296, `RunGenie` line 368-369/401-402, with `bboxLeft = 0`). Windows use absolute virtual-screen coordinates from `GetWindowRect` / `DWMWA_EXTENDED_FRAME_BOUNDS`, so a window on a secondary monitor (or at negative coordinates left of the primary) gets blitted to coordinates that fall outside the primary-monitor-sized buffer and are clipped — the animation renders wrong or not at all. The dock target also falls back to `SM_CYSCREEN` / primary `Shell_TrayWnd` only. For a mod that draws full-screen overlays, this needs to be driven off the window's actual monitor (`MonitorFromWindow` + `MONITORINFO`, which you already use as one fallback in `StartGenieAnim`) and account for per-monitor DPI.

**2. The minimize path blocks the application's UI thread for up to 150 ms.** In `StartGenieAnim`, the minimize branch busy-waits for the worker thread's first frame:
   ```cpp
   int timeout = 15;
   while (!pReady->load() && timeout > 0) { Sleep(10); timeout--; }   // up to 150 ms
   ```
   This runs synchronously inside `ShowWindow_Hook` / `DefWindowProcW_Hook`, i.e. on the target app's UI thread, so every minimize stalls that app for up to 150 ms before it returns. Combined with `@include *`, that's a perceptible hitch on minimize across every app on the system. Consider hiding the real window from the worker thread once the ghost's first frame is up (post a message / use the same `pReadyFlag` from the worker side) rather than blocking the caller.

**3. Reversibility: windows can be left invisible or permanently `WS_EX_LAYERED`.**
   - **Invisible windows after disabling the mod.** When the mod minimizes a window it sets it `WS_EX_LAYERED` + alpha 0 and leaves it that way while iconic (`RevealRealWindow` keeps alpha 0 for the non-rising case, line 168). If the mod is disabled while a window is in this state, the restore is no longer hooked, so the window comes back `WS_EX_LAYERED` with alpha 0 — fully transparent / invisible. `Wh_ModUninit` should walk any windows still in the mod-minimized state and restore their alpha / ex-style.
   - **`WS_EX_LAYERED` is never removed after a restore.** The restore hooks add `WS_EX_LAYERED` *before* calling `StartGenieAnim(hWnd, TRUE)` (lines 712-713 / 733-734), and `StartGenieAnim` then captures `data->originalExStyle = GetWindowLongPtrW(hWnd, GWL_EXSTYLE)` — which already includes the just-added `WS_EX_LAYERED`. So `RevealRealWindow`'s `if (!(data->originalExStyle & WS_EX_LAYERED))` (line 161) is always false and the style is never reverted, leaving every restored window layered for the rest of the session. Capture the original ex-style *before* adding `WS_EX_LAYERED` and thread it through.

**4. No screenshot/GIF in the README.** This is an inherently visual mod (three distinct animation styles), and the README has no image. A short GIF of each effect would help users a lot — and `i.imgur.com` / `raw.githubusercontent.com` are the allowed image hosts.

<p>

- **Use `WindhawkUtils::SetFunctionHook` instead of raw `Wh_SetFunctionHook` with `void*` casts** (lines 744-745) — it's the type-safe modern form.
- **Drop the unused `-lmsimg32`** from `@compilerOptions`. The mod uses `UpdateLayeredWindow` (user32) with a `BLENDFUNCTION`, but no `msimg32` export (`AlphaBlend`/`TransparentBlt`/`GradientFill`). `-luser32` is also linked by default.
- **Redundant NULL check** on `Wh_GetStringSetting` (line 130) — it never returns NULL (returns `L""` when unset/on error), so `if (mode)` is always true. Harmless, just dead.
- **`THREAD_PRIORITY_HIGHEST`** on the animation thread (line 597) competes with the very UI thread it's animating; normal priority is usually enough.
- **README formatting**: `###Features` / `###  Roadmap` are missing a space after `###`, and the `$name` strings have a leading space (e.g. `" Animation Style"`).
- **Snapshot-cache cleanup relies on `WM_DESTROY` routing through `DefWindowProcW`** (line 721). A window using `DefWindowProcA`, or one whose proc consumes `WM_DESTROY`, can leak its cached bitmap until `Wh_ModUninit`. Bounded and cleaned up on unload, so low impact.

</p>
</details>

<p>

- **`@include *` blast radius.** Animating "all app windows" inherently needs broad injection, so there's no obviously narrower scope — but it does mean these hooks (and any bug in them) run in *every* process on the system, including non-GUI/system ones, which amplifies items 3 and 4 above. Worth being aware of; no clean alternative if the goal is truly all apps.
- **Per-animation cost.** Each animation allocates screen-sized ARGB DIBs and `memset`s them every frame; the genie path additionally allocates an `SS×SS`-larger render buffer (≈9× a screen-sized buffer at `genieQuality 3`) and box-filters it in software per frame, on a high-priority thread. This is the core of the effect so there may be no clean alternative, but it's a heavy CPU/RAM hit, especially at 4K + quality 3 + several concurrent animations.
- **Snapshot fidelity vs. the "zero artifacts" claim.** The minimize snapshot is `BitBlt`'d from the *screen* (line 674), so it captures any occluding windows and the cursor over the target. The restore snapshot uses `PrintWindow(..., PW_CLIENTONLY | 0x2)` (line 670), which omits the non-client frame (title bar/border) — so minimize and restore can capture visibly different content. Worth verifying against the README's "zero visual artifacts" promise.
- **Coverage gaps.** Minimizes/restores that don't go through exported `ShowWindow` or `DefWindowProcW` won't animate — e.g. apps using `DefWindowProcA`, `CloseWindow`, or tray/custom-chrome paths. Conversely, if a window's `DefWindowProc` SC_MINIMIZE handling internally calls the exported `ShowWindow`, both hooks could fire and start two animations for one minimize — worth confirming this doesn't double-trigger in practice.

</p>
</details>

kivsak:
@Anixx 

Today I'm releasing an updated version with lots of new settings and animations.
Let's see if this bug remains

Anixx:
The incompatibility with the Explorerframe Fixes mod still remains, the windows are restored in whited out state.

kivsak:
I have the Explorerframe Fixes mod installed.
And nothing is turning white for me (
In which window exactly does this bug occur for you?

@Anixx

kivsak:
Знімок екрана 2026-06-06 221039

Oh, I found it
This happens when you flip this switch

kivsak:
@Anixx 

I dug into the code and found out why this is happening. Long story short: the flaw is actually on the Explorerframe Fixes side. Unfortunately, I can't fix it from my end.Here is what's happening under the hood:To create a smooth animation, my mod takes a "screenshot" of the window in memory just a split second before it appears, using the PrintWindow system function. I'm simply telling Windows: "draw the window for me." But when the Classic CabinetWClass background color option is enabled in Explorerframe Fixes, it intercepts the background drawing command and just aggressively fills everything with a solid white color. The mod thinks it's drawing the background of a real window on the screen, but in reality, it's painting over my screenshot. Because of this, my mod receives an empty white square and, naturally, animates exactly that.  Why can't I come up with a workaround?For me, the screenshot function works like a "black box". I ask the system to give me an image, but I physically cannot prevent another mod from interfering in this process and drawing its own colors over my buffer. The only way to bypass this would be to rewrite half of my mod's engine using more complex tools (DirectX), which is just unrealistic to do for the sake of one single conflict.  How can this actually be solved?The best option is to drop a bug report to the author of Explorerframe Fixes (Waldemar).He literally just needs to add one line to his code: a check so that it doesn't fill the background with white if the window is currently taking a screenshot (i.e., when the WM_PRINT system message is being sent). This is a very quick fix for his mod, and after that, our mods will work together perfectly!

Anixx:
No, the screenshot captured is correct because the window during the animation looks correctly (if it was looking correctly before the minimization). 

The whitening out happens only after the restore animation is over.

kivsak:
// ==WindhawkMod==
// @id              explorerframe-fixes-for-win11-22h2plus
// @name            Explorerframe fixes for Win11 22H2+
// @description     Fixes three problems with file explorer
// @version         1.0.1
// @author          Waldemar
// @github          https://github.com/CyprinusCarpio
// @architecture    x86-64
// @compilerOptions -lcomctl32 -lgdi32
// ==/WindhawkMod==

// ==WindhawkModReadme==
/*
Фікс для Explorerframe.
*/
// ==/WindhawkModReadme==

// ==WindhawkModSettings==
/*
- DisplayMenuBar: false
  $name: Display menu bar
- ClassicBackgroundColor: true
  $name: Classic CabinetWClass background color
  $description: Use classic background color for explorer windows.
*/
// ==/WindhawkModSettings==

#include <windhawk_utils.h>
#include <Windows.h>
#include <shdeprecated.h>
#include <vector>

bool g_settingDisplayMenuBar;
bool g_settingFixCWCBackground;

std::vector<HWND> g_subclassedRebars;
std::vector<HWND> g_subclassedListviews;

int FindMenuBand(HWND hReBar) {
    int bandCount = (int)SendMessage(hReBar, RB_GETBANDCOUNT, 0, 0);
    for (int i = 0; i < bandCount; i++) 
    {
        REBARBANDINFO rbbi = { sizeof(REBARBANDINFO) };
        rbbi.fMask = RBBIM_CHILD | RBBIM_STYLE;
        SendMessage(hReBar, RB_GETBANDINFO, i, (LPARAM)&rbbi);

        if (rbbi.hwndChild == NULL) continue;

        // Check if the child window is a ToolbarWindow32
        wchar_t className[256];
        GetClassNameW(rbbi.hwndChild, className, sizeof(className));
        if (lstrcmpW(className, L"ToolbarWindow32") != 0) continue;

        // Check for image list. Menu doesn't have one.
        if(SendMessageW(rbbi.hwndChild, TB_GETIMAGELIST, 0, 0) != NULL) continue;

        return i;
    }

    return -1;
}

#define HACKFIX_FLAGS SWP_NOACTIVATE | SWP_NOOWNERZORDER | SWP_NOZORDER | SWP_NOMOVE

LRESULT CALLBACK RebarSubclassProc(_In_ HWND hWnd,
        _In_ UINT uMsg,
        _In_ WPARAM wParam,
        _In_ LPARAM lParam,
        _In_ DWORD_PTR dwRefData){
    if(uMsg == WM_DESTROY)
    {
        g_subclassedRebars.erase(std::remove_if(
                              g_subclassedRebars.begin(), g_subclassedRebars.end(),
                              [hWnd](HWND &h)
        {
            return h == hWnd;
        }));
        return 0;
    }
    if(uMsg == RB_SHOWBAND)
    {
        if((int)wParam == FindMenuBand(hWnd))
        {
            LRESULT toRet = DefSubclassProc(hWnd, uMsg, wParam, g_settingDisplayMenuBar);
            if(g_settingDisplayMenuBar)
            {
                // If the following hack is not done, menu bar items may be invisible.
                HWND stwc = GetParent(GetParent(hWnd));
                RECT rect;
                GetClientRect(stwc, &rect);
                int wx = rect.right - rect.left;
                int wy = rect.bottom - rect.top;
                SetWindowPos(stwc, NULL, 0, 0, wx <= 0 ? 1300 : wx + 1, wy <= 0 ? 900 : wy + 1, HACKFIX_FLAGS);
                if(wx <= 0)
                    SetWindowPos(stwc, NULL, 0, 0, wx, wy, HACKFIX_FLAGS);
                RedrawWindow(stwc, NULL, NULL, RDW_INVALIDATE);
            }
            return toRet;
        }
    }
    
    return DefSubclassProc(hWnd, uMsg, wParam, lParam);
}

typedef long (*__cdecl CBSInitialize_t)(void*, HWND);
CBSInitialize_t CBSInitializeOriginal;

long __cdecl CBSInitializeHook(void *pThis, HWND hWnd){
    // We need not subclass the other ReBarWindow32.
    // At this point in time, it's exstyle is different.
    if(GetWindowLong(hWnd, GWL_EXSTYLE) == 0)
    {
        HWND rb = FindWindowExW(hWnd, NULL, L"ReBarWindow32", NULL);
        if(rb != NULL)
        {
            WindhawkUtils::SetWindowSubclassFromAnyThread(rb, RebarSubclassProc, NULL);
            g_subclassedRebars.push_back(rb);
        }
    }

    return CBSInitializeOriginal(pThis, hWnd);
}

LRESULT CALLBACK ListviewSubclassProc(_In_ HWND hWnd,
        _In_ UINT uMsg,
        _In_ WPARAM wParam,
        _In_ LPARAM lParam,
        _In_ DWORD_PTR dwRefData){
    if(uMsg == WM_DESTROY)
    {
        g_subclassedListviews.erase(std::remove_if(
                              g_subclassedListviews.begin(), g_subclassedListviews.end(),
                              [hWnd](HWND &h)
        {
            return h == hWnd;
        }));
        return 0;
    }
    // This is less than surgical, but it appears to have no adverse effects.
    if(uMsg == WM_SETREDRAW)
    {
        wParam = true;
    }

    return DefSubclassProc(hWnd, uMsg, wParam, lParam);
}

typedef long (*__cdecl CLVHCreateControl_t)(void*, HWND, void*, void*);
CLVHCreateControl_t CLVHCreateControlOriginal;

long __cdecl CLVHCreateControlHook(void* pThis, HWND hWnd, void* a, void* b){
    long ret = CLVHCreateControlOriginal(pThis, hWnd, a, b);
    HWND listview = GetWindow(hWnd, GW_CHILD);
    if(listview != NULL)
    {
        wchar_t name[32];
        GetClassNameW(listview, name, 32);
        if(lstrcmpW(name, L"SysListView32") == 0)
        {
            WindhawkUtils::SetWindowSubclassFromAnyThread(listview, ListviewSubclassProc, NULL);
            g_subclassedListviews.push_back(listview);
        }
    }
    return ret;
}

typedef long(*__cdecl CEFWndProc_t)(void*, HWND, unsigned int, WPARAM, LPARAM);
CEFWndProc_t CEFWndProcOriginal;

long __cdecl CEFWndProcHook(void* pThis, HWND hWnd, unsigned int uMsg, WPARAM wParam, LPARAM lParam){
    if(uMsg == WM_SHOWWINDOW)
    {
        // This message "fixes" the unpredictable appearance of the menu bar,
        // but it sets a internal flag in the CShellBrowser that prevents it
        // from saving it's layout. We stop that in another hook.
        SendMessageW(hWnd, WM_WININICHANGE, 0, 0);
    }
    if(g_settingFixCWCBackground && uMsg == WM_ERASEBKGND)
    {
        HDC hdc = (HDC)wParam;
        
        // --- OUR FIX START ---
        // Skip if it's a memory DC (used by PrintWindow/screenshots)
        if (GetObjectType(hdc) != OBJ_MEMDC) 
        {
            RECT rc;
            GetClientRect(hWnd, &rc);
            HBRUSH hbr = (HBRUSH)(COLOR_WINDOW + 1);
            FillRect(hdc, &rc, hbr);
            return 1;
        }
        // --- OUR FIX END ---
    }
    return CEFWndProcOriginal(pThis, hWnd, uMsg, wParam, lParam);
}

typedef long(*__cdecl CSBSetFlags_t)(void*, unsigned long, unsigned long);
CSBSetFlags_t CSBSetFlagsOriginal;

long __cdecl CSBSetFlagsHook(void* pThis, unsigned long a, unsigned long b){
    // This flag is a deprecated feature that has no business being set
    if(a & BSF_UISETBYAUTOMATION) a &= ~BSF_UISETBYAUTOMATION;
    return CSBSetFlagsOriginal(pThis, a, b);
}

BOOL Wh_ModInit() {
    Wh_Log(L"Explorerframe fixes for Win11 22H2+ Init");

    HMODULE hExplorerFrame = LoadLibraryW(L"explorerframe.dll");

    if (!hExplorerFrame)
    {
        Wh_Log(L"Failed to load explorerframe.dll");
        return FALSE;
    }

    WindhawkUtils::SYMBOL_HOOK explorerframe_dll_hooks[] =
    {
        {   {
                L"private: virtual __int64 __cdecl CExplorerFrame::v_WndProc(struct HWND__ *,unsigned int,unsigned __int64,__int64)"
            },
            (void**)&CEFWndProcOriginal,
            (void*)CEFWndProcHook,
            FALSE
        },
        {   {
                L"protected: virtual long __cdecl CBandSite::_Initialize(struct HWND__ *)"
            },
            (void**)&CBSInitializeOriginal,
            (void*)CBSInitializeHook,
            FALSE
        },
        {   {
                L"public: virtual long __cdecl CShellBrowser::SetFlags(unsigned long,unsigned long)"
            },
            (void**)&CSBSetFlagsOriginal,
            (void*)CSBSetFlagsHook,
            FALSE
        }
    };

    if (!WindhawkUtils::HookSymbols(hExplorerFrame, explorerframe_dll_hooks, 3))
    {
        Wh_Log(L"Failed to install explorerframe hooks");
        return FALSE;
    }

    HMODULE hShell32 = LoadLibraryW(L"shell32.dll");

    if (!hShell32)
    {
        Wh_Log(L"Failed to load shell32.dll");
        return FALSE;
    }

    WindhawkUtils::SYMBOL_HOOK shell32_dll_hooks[] =
    {
        {   {
                L"public: virtual long __cdecl CListViewHost::CreateControl(struct HWND__ *,struct IListControlHost *,struct IViewSettings *)"
            },
            (void**)&CLVHCreateControlOriginal,
            (void*)CLVHCreateControlHook,
            FALSE
        }
        
    };

    if (!WindhawkUtils::HookSymbols(hShell32, shell32_dll_hooks, 1))
    {
        Wh_Log(L"Failed to hook CListViewHost::CreateControl");
        return FALSE;
    }

    g_settingDisplayMenuBar = Wh_GetIntSetting(L"DisplayMenuBar");
    g_settingFixCWCBackground = Wh_GetIntSetting(L"ClassicBackgroundColor");

    return TRUE;
}

void Wh_ModUninit() {
    Wh_Log(L"Explorerframe fixes for Win11 22H2+ Uninit");
    Wh_Log(L"Removing subclasses from %i rebars.", g_subclassedRebars.size());
    for(HWND& h : g_subclassedRebars)
    {
        WindhawkUtils::RemoveWindowSubclassFromAnyThread(h, RebarSubclassProc);
    }
    Wh_Log(L"Removing subclasses from %i listviews.", g_subclassedListviews.size());
    for(HWND& h : g_subclassedListviews)
    {
        WindhawkUtils::RemoveWindowSubclassFromAnyThread(h, ListviewSubclassProc);
    }
}

void Wh_ModSettingsChanged(){
    g_settingDisplayMenuBar = Wh_GetIntSetting(L"DisplayMenuBar");
    g_settingFixCWCBackground = Wh_GetIntSetting(L"ClassicBackgroundColor");
}

kivsak:
@Anixx 

I've updated the Waldemar mod
Explorerframe fixes for Windows 11 22H2+

You can test this locally

kivsak:
[Explorerframe fixes for Win11 22H2+.wh.cpp](https://github.com/user-attachments/files/28671250/Explorerframe.fixes.for.Win11.22H2%2B.wh.cpp)


Here it is, but in a more convenient file format

@Anixx

Anixx:
With your modification the bug is still there.

Anixx:
If it would be due to a bad screenshot, the window would appear wrongly during the animations.

kivsak:
[Explorerframe fixes for Win11 22H2+.wh.cpp](https://github.com/user-attachments/files/28671343/Explorerframe.fixes.for.Win11.22H2%2B.wh.cpp)
@Anixx 
Sorry, I sent the wrong version of the file

kivsak:
@Anixx Yay!!!🎉🎉🎉🎉🎉
Write to the developer of this mod and ask them to update it.

Anixx:
No, with the latest version the issue remains (ignore my previous deleted comment).

There is no change.

Anixx:
I just havent enabled your mod, that's why it looked like there was no issue.

The screenshot is obviously not the culprit.

kivsak:
https://github.com/user-attachments/assets/59b82e3f-7626-420f-80e6-e3c112008626

@Anixx

kivsak:
It still paints the top part of the window white,
but for some reason only on the first load.
When I minimized it again, everything was white.

@Anixx

kivsak:
@Anixx 
But before I applied the fixes to the Explorerframe fixes for Win11 22H2+ mod, everything was turning white

Anixx:
Well, I tried again , and it seems, your fixes work. Unfortunately, the author of the Explorerframe fixes mod is inactive, so I will include these changes into my fork.

But somehow the other mod implementing the Genie animation is not affected.

kivsak:
@Anixx 🎉🎉🎉🎉🎉

Anixx:
There is also another issue that I experoience: if the window is restored from minimized to maximized mode, the animation constructs it without titlebar, and only after animation stops, the titlebar gets added and moves withow contents (such as menubar and even the window's bottom, statusbar) down.

This seemingly happens only in Classic theme. Maybe also in Basic, but I did not test. Standard windows animations have no such issue. The non-maximized windows are restored correctly, instantly with the titlebar.

Anixx:
On some windows (like the classic WinMail), the closing animation is played twice.

kivsak:
Yeah, I've seen this bug before.
I'll fix it in a week.
@Anixx

kivsak:
@Anixx I fixed it.
Unfortunately, the close animation doesn't work in all windows.
That bug with the double animation happened precisely because I tried to apply the animation to all windows.

kivsak:
@m417z 
 All issues addressed and fixed. Ready for a new check!

Anixx:
The Winmail animation still repeats twice, and the restoration to the mazximized state happnas the height of the titlebar upper than should be.The windows are restored without titlebar in the animation.

kivsak:
Does this only happen with Winmail?
@Anixx

Anixx:
For me, yes, but I think it should also happen with Outlook Express because Winmail is built on its code. 

It also does not work with some apps, such as Sysinternals SystemInformer, Windows 7 calculator, classic notepad.

Anixx:
The Zoom effect works with Notepad++ but Glass (shatter) does not...

Anixx:
With File Locator Pro (Agent Ransack) also no effect works. 
There are more apps with which it does not work that those with which works. 
Not working also with Win7 snapping tool, Command prompt (and anny console window), 7-Zip, PascalABC.NET.
OpenOffice (with Glass effect): on Calc there is no visible effect, on Writer the effect it too brief (only starts in the center of the Window and disappears). In LibreOffice Calc works, but on Writer also the effect is too brief.

The properties dialogs are not affected by any effect.

m417z:
Let me know when it's ready for review.

PR #4299, opened 2026-06-04
New mod submission: Taskbar Performance Widget

AngelGarciaODiana:
## Changelog

New mod submission: Taskbar Performance Widget.
A lightweight, standalone performance monitor for Windows 11 taskbar, using PDH metrics and native DWM materials (Acrylic/Mica).

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [x] Gemini

### DIFF
--- a/mods/taskbar-perf-widget-standalone.cpp
+++ b/mods/taskbar-perf-widget-standalone.cpp
@@ -1,0 +1,669 @@
+// ==WindhawkMod==
+// @id              taskbar-perf-widget-standalone
+// @name            Taskbar Performance Widget
+// @description     Módulo avanzado con renderizado Acrílico, centrado perfecto y cursor nativo.
+// @version         9.3
+// @author          Agarciao10
+// @include         explorer.exe
+// @compilerOptions -lpdh -lole32 -ldwmapi -lgdi32 -luser32 -lshcore -lgdiplus -lshell32
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Taskbar Performance Widget
+
+Un widget de monitorización de rendimiento de alta precisión, diseñado para
+integrarse de forma nativa y simétrica en la barra de tareas de Windows 11.
+
+## Características principales
+- **Monitorización en tiempo real:** Visualización instantánea de CPU, Memoria
+(RAM), Red (Bajada/Subida) y GPU.
+- **Arquitectura Aislada:** Utiliza un proceso independiente para garantizar la
+estabilidad total del explorador (`explorer.exe`).
+- **Diseño Nativo:** Integración visual mediante los materiales oficiales de
+Windows 11 (Acrílico y Mica).
+- **Altamente Personalizable:** Ajusta la geometría, materiales de fondo y
+visibilidad de indicadores desde la interfaz de usuario.
+- **Centrado Geométrico:** Algoritmo avanzado de renderizado vectorial que
+garantiza una disposición simétrica y elegante.
+
+## Configuración
+- **Material de Fondo:** Elige entre el efecto transparente para una integración
+invisible, o aplica materiales nativos para un aspecto tipo Fluent Design.
+- **Geometría:** Ajusta la altura y el ancho por columna para adaptar el widget
+a cualquier escala de interfaz o densidad de barra de tareas.
+- **Visibilidad:** Activa solo los indicadores necesarios para optimizar el
+espacio disponible.
+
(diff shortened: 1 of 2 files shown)


m417z:
### Submission review

---

The mod is well structured and it's good that you already reached for the tool-mod pattern — but it should host itself in `windhawk.exe`, not `explorer.exe`, and the overlay needs to stop eating taskbar clicks.

**1. Run as a `windhawk.exe` tool mod, not inside `explorer.exe`.** This is a textbook "mod as a tool": it installs no function hooks and needs no injection — it only creates its own top-level overlay window, reads PDH counters, and registers an out-of-context `SetWinEventHook`, all of which work from any process. You already pasted the tool-mod boilerplate, but it targets `explorer.exe`, which is not according to the suggested pattern (see [the wiki](https://github.com/ramensoftware/windhawk/wiki/Mods-as-tools:-Running-mods-in-a-dedicated-process)).

   Fix: change `@include explorer.exe` → `@include windhawk.exe` and paste the **exact** boilerplate from the wiki. The current copy dropped two pieces that matter once the host is `windhawk.exe`: the session-0 guard (`ProcessIdToSessionId(...) == 0 → return FALSE`, so the launcher doesn't fire in the session-0 service instance) and the full `-service` / `-service-start` / `-service-stop` exclusion.


   ```cpp
   g_hWidgetWnd = CreateWindowEx(
       WS_EX_LAYERED | WS_EX_TRANSPARENT | WS_EX_TOOLWINDOW | WS_EX_TOPMOST, ...);
   ```
**2. Use the exact wiki snippet.** Beyond the two functional omissions in item 1, the boilerplate was reformatted (merged error checks, `EntryPoint_Hook` lost its log). Pasting the snippet verbatim keeps it diffable against the canonical version as it evolves.

**3. Add a screenshot to the README.** The mod has an obvious visual result (a widget on the taskbar) but the README has no image. Please add a screenshot or short GIF so users can see what it looks like — allowed image hosts are `i.imgur.com` and `raw.githubusercontent.com`.

<p>

- **Unused `@compilerOptions` libraries.** `-lshcore` and `-lole32` don't appear to be used (no shcore/COM calls; `CommandLineToArgvW` already pulls in `-lshell32`). Drop the ones you don't need (CI doesn't flag unused `-l` entries).
- **Unused includes.** `<mutex>` and `<string>` aren't used anywhere — remove them.
- **Dead variable.** `g_Running` is set in `WhTool_ModInit`/`WhTool_ModUninit` but never read.
- **Teardown robustness.** `WhTool_ModUninit` relies on `g_hWidgetWnd` existing to post `APP_WM_CLOSE`; if the window was never created (e.g. `CreateWindowEx` failed), nothing wakes the message loop and `join()` blocks forever. Signal the thread directly instead, the way the reference mod does — `PostThreadMessage(threadId, WM_QUIT, 0, 0)` — capturing the widget thread id in `WidgetThread`.
- **Settings data race.** `LoadSettings()` writes `g_Settings` on the settings-changed thread while the widget thread reads it during paint/reposition. The fields are plain `int`/`bool` so torn reads are benign in practice, but it's still an unsynchronized access on every settings change.

</p>
</details>

<p>

- **Per-monitor DPI.** The font sizes (15/13/10 px) and internal layout offsets (`-28`, `+24`) are hard-coded and not scaled by DPI, so on non-100% displays the text and spacing will be mis-sized despite the README's HiDPI claim. `height`/`colWidth` let users compensate for the overall box but not the internal spacing — consider scaling by `GetDpiForWindow`.
- **Multi-monitor.** `FindWindow(L"Shell_TrayWnd")` only finds the primary taskbar, so the widget never appears on secondary-monitor taskbars (`Shell_SecondaryTrayWnd`).
- **Metric semantics.** `% Committed Bytes In Use` (commit charge) differs from the physical-RAM % Task Manager shows; the GPU figure sums `\GPU Engine(*)\Utilization Percentage` across all engines and clamps to 100 (overcounts vs. Task Manager's per-engine max); network sums all interfaces including loopback/virtual adapters. Reasonable approximations, just worth knowing they won't match Task Manager exactly.
- **Glyphs.** The same MDL2 glyph is reused for more than one metric (RAM and GPU both `\xE7F4`; receive and send both `\xEC27`, distinguished only by the trailing ↑/↓) — minor visual polish.

</p>
</details>

m417z:
Please revisit item 2 which asks to "Use the exact wiki snippet" for easier review, thanks.

PR #4301, opened 2026-06-04
 Net-Toggle Major Update v2.0 (previously Network Toggle)

BlackPaw21:
## Changelog

### v2.0.0
- **Complete rebuild.** Mod renamed to Net-Toggle.
- New: **DNS Monitoring** — The icon monitors your connection and changes color if your internet drops out (configurable in Mod Settings).
- New: **WiFi-style tray icon** with 5 color states (Red / Yellow / Blue / Green / Grey).
- New: Middle-click → **Full Network Reset** (flush DNS, release/renew, winsock + IP reset).
- New: Right-click context menu — toggle network, open Network Settings, or exit.
- New: Donate button on the mod page.
- Improved: Tray icon is now independent from the Windhawk app and no longer groups with it in the taskbar.
- Improved: Tray icon persists reliably across Explorer restarts.
- Improved: Icon stays in sync even if you disable your adapter directly in Windows Settings.
- Fixed: Occasional hangs during network toggles or timeouts.
- Fixed: Tray icon could show the wrong state if a toggle failed.

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini


m417z:
### Submission review

---

**1. Middle-click "Full Network Reset" is a destructive, persistent, system-wide change.** `ResetWorkerThreadProc` runs `netsh winsock reset; netsh int ip reset; ...` elevated. These are not adapter toggles — `netsh winsock reset` rebuilds the machine-wide Winsock catalog (removing any third-party LSPs that VPNs/proxies/security tools install) and `netsh int ip reset` rewrites the TCP/IP stack configuration. Both are persistent, can break other software's networking, and require a **reboot** to fully take effect. It runs on a single middle-click with no confirmation. This goes well beyond the mod's "toggle the adapter" premise. Consider at least gating it behind a confirmation dialog that spells out that it resets system-wide network config and needs a reboot.

**2. The "Exit" context-menu item leaves a half-dead, resident tool process.** `case 3` posts `WM_CLOSE`, which deletes the tray icon, destroys the window, and ends `TrayThreadProc` — but the `windhawk.exe` tool process keeps running because `NetWatchThreadProc` is still alive. The result: no tray icon, `g_trayHwnd` is now `nullptr` so the icon can't even come back on an Explorer restart (the thread that handles `TaskbarCreated` is gone), and the only way to recover is to disable + re-enable the mod in Windhawk. "Exit" doesn't really exit. Recommend removing the item (the mod's lifetime is managed by Windhawk's enable/disable), or, if you keep it, have it fully terminate the process so the state is at least consistent:
```cpp
case 3: ExitProcess(0); break;  // fully quits the tool process until the mod is re-enabled
```

**3. No screenshot/GIF in the README.** The entire UX is a colored tray icon with five states (plus the right-click menu). For a mod with a visible effect, a screenshot or short GIF in the README really helps users — recommended.

<p>

- **`WindhawkUtils::StringSetting` (RAII) over manual get/free.** In `WhTool_ModInit` and the `WM_SETTINGS_CHANGED` handler you do `Wh_GetStringSetting(L"dnsServer")` + manual `Wh_FreeStringSetting`. `WindhawkUtils::StringSetting setting = WindhawkUtils::StringSetting::make(L"dnsServer");` frees automatically. (Also note `Wh_GetStringSetting` never returns NULL — it returns `L""` — so the `pDnsIp && pDnsIp[0]` checks can just be `pDnsIp[0]`.)
- **`std::atomic<LONG>` over `volatile` + `Interlocked*`.** The flags (`g_isProcessingClick`, `g_networkIsUp`, `g_dnsIsReachable`, etc.) mix plain `volatile` reads with `Interlocked` writes. `std::atomic<LONG>` expresses the intent and is the modern idiom; `volatile` doesn't actually guarantee atomicity.
- **`g_pingIntervalMs` initializer (30000) doesn't match the setting default (10s).** Harmless since it's re-read in `WhTool_ModInit`, but the dangling 30000 is confusing — initialize it to `10000` or just `0`.
- **Unload-time handle race on `g_activeWorkerThread`.** `WhTool_ModUninit` reads/closes `g_activeWorkerThread` while a tray-thread click could be reassigning it. Benign in practice (the path ends in `ExitProcess(0)`), but worth a note.

</p>
</details>

<p>

- **`CheckActualNetworkState` shells out to `powershell.exe`** at startup and on every `NotifyAddrChange` event just to count physical adapters. It's event-driven (not a tight poll), so it's not abusive, but each spawn costs a few hundred ms and adds latency to state detection. You already link `-liphlpapi`; `GetAdaptersAddresses` could read adapter up/down natively without a process spawn. The catch is that reproducing `Get-NetAdapter -Physical`'s physical-vs-virtual distinction via the API is non-trivial, so there may be no clean drop-in — noting as an FYI rather than a required change.
- **The "DNS ping" is a TCP connect to port 53**, not ICMP — which the code comments explain well (ICMP breaks after `Disable-NetAdapter`). Just be aware the setting names/tooltips say "ping" while it's really a TCP reachability probe; fine functionally.

</p>
</details>

BlackPaw21:
@m417z 
 there are a lot of mods and submissions to go through so no one expects a fully manual review of every single mod, you are doing an amazing job with or without AI and thank you for that.
i will go over everything and apply the needed fixes :)

m417z:
### Submission review

---

**1. The tray icon is blocked at startup by a synchronous `powershell.exe` spawn, and every adapter change re‑spawns PowerShell.** `TrayThreadProc` calls `CheckActualNetworkState()` — which launches `powershell.exe` and waits up to 5 s — *before* it creates the window and adds the tray icon (net-toggle.wh.cpp:963 → :1012). At login, where PowerShell cold‑start is slow, the icon can be missing for a second or more every boot. `NetWatchThreadProc` then runs the same PowerShell spawn on *every* `NotifyAddrChange` event and every 15 s in the poll fallback (:903, :925).

Two improvements, either of which helps:
   - Don't block icon creation on it: create the window + add the icon assuming ON, then let the NetWatch thread (which already posts `WM_UPDATE_TRAY_STATE` asynchronously) correct the color — so the icon appears immediately.
   - A status read doesn't need a process spawn at all. `iphlpapi` is already linked; `GetIfTable2` / `GetAdaptersAddresses` give you per‑adapter `OperStatus` plus the `HardwareInterface`/`ConnectorPresent` flags to filter physical adapters, replacing the `Get-NetAdapter -Physical … .Count` shell‑out. [internet-status-indicator](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/internet-status-indicator.wh.cpp) reads connectivity without launching a process. (The privileged *toggle* still needs elevation — see the functionality note — but the read does not.)

**2. No screenshot/GIF in the README.** The whole UI is a custom‑drawn 5‑color WiFi tray icon plus a context menu, but the README has no image. For a mod whose effect is entirely visual, please add a screenshot or short GIF (only `i.imgur.com` / `raw.githubusercontent.com` are allowed hosts) so users browsing the catalog can see the icon and its color states.

<p>

- **Stale comment:** `ResetWorkerThreadProc` says "Use 12s settle (wParam=1)" (net-toggle.wh.cpp:606) but the `WM_TRIGGER_PING` handler uses `15000` ms for `wParam == 1` (:814).

</p>
</details>

<p>

Non‑critical observations about the feature behavior itself — mostly FYIs with no clean alternative.

- **Icon anti‑aliasing.** `DrawWifiIcon` packs straight `r/g/b` with a sub‑255 alpha into the 32‑bpp DIB (net-toggle.wh.cpp:392), but `CreateIconIndirect` expects **premultiplied** alpha, so the arc's AA edges can look slightly too bright/haloed. Multiply `r/g/b` by `alpha` before packing.

</p>
</details>

mods/net-toggle.wh.cpp
```diff
@@ -1,76 +1,157 @@
...
 */
 // ==/WindhawkModReadme==
 
+#define WIN32_LEAN_AND_MEAN
+// winsock2.h + ws2tcpip.h MUST come before windows.h (pulled in by
+// windhawk_utils.h). iphlpapi.h only declares the netioapi APIs
+// (GetIfTable2 / MIB_IF_TABLE2 / FreeMibTable) when the winsock2 types are
+// already in scope; including ws2tcpip.h after windows.h leaves them hidden
+// and clang falls back to the legacy MIB_IFTABLE.
+#include <winsock2.h>
```
m417z, line 95: This still won't work - `windows.h` is included implicitly via command line that's passed to the compiler, so `#define WIN32_LEAN_AND_MEAN` and the ordering here have no effect.

You can instead add `-DWIN32_LEAN_AND_MEAN` to `@compilerOptions` to fix this. I'm not sure if there's a problem currently, but there's a warning during compilation about the order of the includes.

BlackPaw21: damn, thanks for pointing this out my Claude is having a hard time recently so i need to write things more manual.

m417z:
Is it ready? How about a screenshot? :)

BlackPaw21:
> Is it ready? How about a screenshot? :)

oh you need to manually validate it? i thought the check was just broken, i will bring the `-DWIN32_LEAN_AND_MEAN` back and add some screenshots <3

BlackPaw21:
had some issues with Imgur but made it work
i hope im actually done this time XD
thank you again for your work :)

Issue #4279, opened 2026-06-03
System Settings app Styler (similar to Taskbar styler)


eugenesvk:
Given the richness of styles for the taskbar https://github.com/ramensoftware/windows-11-taskbar-styling-guide due to the cool https://windhawk.net/mods/windows-11-taskbar-styler mod I was wondering whether it would be possible to do something similar to the System Settings app
(which UWPSpy seems to indicate uses the same styling mechanism)

upd: I'm experimenting with it in [this App Styler repo](https://github.com/eugenesvk/windows-11-app-styler)

m417z:
For experimenting, it should already be possible by using one of the existing styler mods (the start menu styler is probably the best fit) and adding `SystemSettings.exe` to the inclusion list in the advanced tab.

eugenesvk:
I've already tried to copy and paste the styles from UWPSpy to the taskbar styler, but it didn't work (added 2 processes
- C:\Windows\ImmersiveControlPanel\SystemSettings.exe
- SystemSettings.exe
(tested
  - `SystemSettings.View.SettingsAutoSuggestCommandSearchBox#CommandSearchBox`
  - `Windows.UI.Xaml.Internal.RootScrollViewer > ScrollContentPresenter > Border > Frame > ContentPresenter > SystemSettings.View.RootPage > Grid#RootPageGrid > Grid#RootCommandSearchGrid > SystemSettings.View.SettingsAutoSuggestCommandSearchBox#CommandSearchBox`
  - `SystemSettings.View.RootPage > Grid#RootPageGrid > Grid#RootCommandSearchGrid > SystemSettings.View.SettingsAutoSuggestCommandSearchBox#CommandSearchBox`
  
variants with `Margin=0,0,0,0`, which should move the search box to the top)


Will check if he menu styler works

eugenesvk:
Interestingly enough, the menu styler mod is better and `SystemSettings.View.RootPage > Grid#RootPageGrid > Grid#RootCommandSearchGrid > SystemSettings.View.SettingsAutoSuggestCommandSearchBox#CommandSearchBox` works, so will experiment with that

Though the effect is not immediate unlike with the taskbar styles, you have to restart the Settings app to see the effect, which is a bit inconvenient to not see live edits

m417z:
> Though the effect is not immediate unlike with the taskbar styles

[...]

PR #4303, opened 2026-06-05
Windows 11 App Styler: Settings

eugenesvk:
Allows styling the Settings App, addresses https://github.com/ramensoftware/windhawk-mods/issues/4279.

Currently only 1 theme is embedded

It's based off the start menu styling mod, more detailed changes commited [here](https://github.com/eugenesvk/windows-11-app-styler/blob/main/src/Settings/windows-11-app-styler-settings.wh.cpp), in this PR all changes on top of the original mod are consolidated

## Changelog

Doesn't update an existing mod

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini


m417z:
Thank you for the submission. Currently, I'm the maintainer of all four styler mods, and many of the fixes and improvements are applied to all of them (e.g. `WindhawkBlur` improvements, target and style syntax and capabilities, and more). Therefore, if there's demand for it, I prefer to be the owner of it to be able to keep all styler mods in sync.

In addition, I could create a styling guide For Settings similar to the existing ones, e.g.:
https://github.com/ramensoftware/windows-11-start-menu-styling-guide

I suggest the following:
* I'll take this submission and release it under my ownership
* I'll create a `windows-11-settings-styling-guide` repo and give you permissions to add the relevant information and the first theme

What do you think?

eugenesvk:
> applied to all of them

That's where the lib/modules would be helpful! I'd also prefer to only import your base styler mod and only change the target exe/style base name/embedde styles (by the way, have you thought about importing those at runtime from text files instead of embedding to avoid having to recompile the mod itself every time and be able to simply update that styling file?)

That's fine if you want to consolidate it, though please use the latest version of the embedded theme as it's fixed a few bugs (there was even a Settings app crashing in one restyled lists which I'd not expect from any proper declarative styling system!)
https://github.com/eugenesvk/windows-11-app-styler/blob/main/src/Settings/windows-11-app-styler-settings.wh.cpp

And maybe for the style guide maybe you could copy that repo since it already has the description of the theme? https://github.com/eugenesvk/windows-11-app-styler/tree/main/Theme/Settings/Densy

m417z:
I created a styling guide repo and invited you as a collaborator:
https://github.com/ramensoftware/windows-11-settings-styling-guide/invitations
Please set it up based on [the existing styling guide repos](https://github.com/orgs/ramensoftware/repositories?q=guide), and add your theme(s). Once it's ready, I'll merge the mod.

> That's where the lib/modules would be helpful! I'd also prefer to only import your base styler mod and only change the target exe/style base name/embedde styles

Modules could be helpful, but it requires some thought and careful design for making it right. Also, the current model of self-contained single-file mods has some upsides despite the limitations. I might think about it in the future, but for now it's not a priority.

> by the way, have you thought about importing those at runtime from text files instead of embedding to avoid having to recompile the mod itself every time and be able to simply update that styling file?

What do you mean? You can update the styles via the settings, there's no need to compile the mod every time. Or did you mean something else?

> That's fine if you want to consolidate it, though please use the latest version of the embedded theme as it's fixed a few bugs

After you add it into the styling guide, I'll take care of it and import it to the mod as is.

eugenesvk:
> You can update the styles via the settings, there's no need to compile the mod every time. Or did you mean something else?

Yes, but then for embedding it into the CPP file you have to move and reformat it. And then if you think some older style in the embedded list misbehaves, you need to delete it from the CPP file and recompile: and you can't easily "delete 50% of the rules" to see which one works.

Currently there are 3 formats:
  - "visual"
  - textual
  - embedded
 
Instead I thought the mod could have a reference to a list of files in it's app settings dir, and then all you had to do to update a style is use your favorite **code editor** to simply edit some `style-name.yaml` (think your textual mode uses it?) and save it. And then also if an embedded style is updated (but the mod code isn't) you wouldn't need to update/recompile the mod to use that style, only the textual style file would be downloaded.
And then patching an embedded style would also be as simple as copy&pasting a file and editing a few values. Or even better: if you had imports for styles, then patching it would be importing the base and copy&pasting just those few extra styles you'd need to change

eugenesvk:
Update the mod and added info to the repo

m417z:
> Yes, but then for embedding it into the CPP file you have to move and reformat it.

I have scripts that help me import the themes from the styling guides into the mods C++ code, and once I'll publish the settings styler, I'll do the same with it. You don't have to do it. As a theme developer, you can just update the settings via the visual editor or the YAML textual editor.

> And then if you think some older style in the embedded list misbehaves, you need to delete it from the CPP file and recompile: and you can't easily "delete 50% of the rules" to see which one works.
> 
> Currently there are 3 formats:
> 
> * "visual"
> * textual
> * embedded

What's wrong with using the textual format to delete 50% of the rules for such tests? You have Ctrl+Z for undo/redo, you have Ctrl+/ for quick comment/uncomment, etc. That's the reason the textual settings mode was added.

> the mod could have a reference to a list of files in it's app settings dir, and then all you had to do to update a style is use your favorite **code editor**

That's similar to the current textual settings. It can be done, but I think that for more users, it will be less convenient and more confusing. If something is missing in the built-in textual settings editor, you can create an issue and I'll consider improving it.

> And then patching an embedded style would also be as simple as copy&pasting a file and editing a few values. Or even better: if you had imports for styles, then patching it would be importing the base and copy&pasting just those few extra styles you'd need to change

Again, already possible with textual settings, isn't it?

eugenesvk:
>As a theme developer, you can just update the settings via the visual editor or the YAML textual editor.

Good to know, might help in the future. Though as a theme modifier, changes in the embedded themes still require a different approach.

> What's wrong with using the textual format to delete 50% of the rules for such tests? 

Because it uses an alien editor with all the different shortcuts and editing conventions. Also, the text area **itself is small** (with an editor you can make it full screen height/width)

For example, 

> You have Ctrl+Z for undo/redo

Yes, I use that

>  you have Ctrl+/ for quick comment/uncomment, etc.

But not that, have a different keybind.

And then there is no 'vim' mode you have in your regular editor.
Also, if I add a comment in there and then tweak the visual styling/change theme etc., the comment is gone because the parser is not roundtripping including comments, I guess?
These are a few of the many issues when you try to use a different text box. 

> Again, already possible with textual settings, isn't it?

Not if you want to delete some styles, they're embedded in the source file, so you need to recompile.
Then there is the issue of trying to patch multiple themes (if you haven't decided yet which one to use permanently) - you can't have one set of changes applied to theme A and another set of changes applied to theme B

> but I think that for more users, it will be less convenient and more confusing.

These users won't see any difference. It's just that all the styling mods will be backed up by actual files instead of being embedded in the code, and users who don't know/don't care can continue to select/edit in the Windhawk app.

m417z:
> Though as a theme modifier, changes in the embedded themes still require a different approach.

For that, you can import the theme from the styling guide manually, and then modify it. Example (Manual installation section):
https://github.com/ramensoftware/windows-11-taskbar-styling-guide/blob/main/Themes/BottomDensy/README.md#manual-installation

Alternatively, you can select the theme and then override specific styles if the change is small and localized.

> Also, the text area **itself is small** (with an editor you can make it full screen height/width)

Agree, I have plans to improve that.

> Also, if I add a comment in there and then tweak the visual styling/change theme etc., the comment is gone because the parser is not roundtripping including comments, I guess?

Yes. Only mod comments (not YAML comments) are preserved:
```yaml
controlStyles:
  - target: Taskbar.TaskbarFrame
    styles:
      # This comment is not preserved
      - // This comment is preserved
      - Fill=Transparent
```
I'll consider improving it in the future too, although it will be tricky in some cases, e.g. if the settings are updated in the visual settings editor. But at least, as long as only the textual editor is used, it should be doable.

> These users won't see any difference. It's just that all the styling mods will be backed up by actual files instead of being embedded in the code, and users who don't know/don't care can continue to select/edit in the Windhawk app.

Depends on the implementation. It's possible to just allow overriding a built-in theme with a file. It's not a trivial change, and not a common request. Currently, I don't plan on adding something like this.

m417z:
In https://github.com/ramensoftware/windows-11-settings-styling-guide:
* Please rename the `ReadMe.md` files to `README.md`.
* For [the Densy theme](https://github.com/ramensoftware/windows-11-settings-styling-guide/blob/main/Themes/Densy/ReadMe.md), move the "Theme selection" section to the bottom, and add a "Manual installation" section with the YAML of the theme. It will be imported to the new mod. You can copy the template from here:
  https://github.com/ramensoftware/windows-11-taskbar-styling-guide/blob/main/Themes/BottomDensy/README.md

eugenesvk:
> For that, you can import the theme from the styling guide manually, and then modify it. Example (Manual installation section):

Only if the author has provided that!!! And given how cumbersome it is to have the styles in different formats, for example, for the Settings app theme I don't have those manual installations styles myself because I was jumping between the app the a local file where I could organize changes properly and comment etc.

https://github.com/ramensoftware/windows-11-settings-styling-guide/blob/main/Themes/Densy/ReadMe.md

> Alternatively, you can select the theme and then override specific styles if the change is small and localized.

Not if you want to delete a single style!

> although it will be tricky in some cases, e.g. if the settings are updated in the visual settings editor.

the visual settings editor could also simply show those comments, that'd be helpful in organizing a set of changes.
So similar to how you can "add a style" you could add a comment field. The main issue here is the parser capability, i.e., whether it can preserve whitespace and comments. Otherwise even if the visual settings editor doesn't show the comments, they still don't need to disappear, just not be editable in the UI.

m417z:
> Only if the author has provided that!!!

Each theme that's part of the mod has a YAML in the styling guide. Both are synchronized. That's also why I asked you to add it to your new settings theme.

> Not if you want to delete a single style!

True, you can't delete a style, but you can override it. You can also use the `Style:=` syntax to clear the value.

eugenesvk:
> Each theme that's part of the mod has a YAML in the styling guide. Both are synchronized.

Yeah, but where did it come from?

> That's also why I asked you to add it to your new settings theme.

That's why I mentioned that I don't have it and would need to manually convert from CPP. Because there is no 1 good place to develop a theme, so I don't have a simple single file that I could comment/live-edit/etc. (see above) and after finishing development simply copy&paste to the guide.

> You can also use the Style:= syntax to clear the value.

Does this restore the default style pre-theme modification or does it simply use a generic default value for this specific property?

m417z:
> > Each theme that's part of the mod has a YAML in the styling guide. Both are synchronized.
> 
> Yeah, but where did it come from?

That's how normally theme authors submit the themes:
* An author creates a theme with a styler mod via the Windhawk settings
* The author switches to textual mode and copies the YAML
* A pull request is created with a readme containing the YAML
* Further updates can be created by creating new pull requests that update the readme and the YAML

Then, before releasing a new mod version, I sync the CPP code to contain the up-to-date style definitions from the YAMLs in the styling guide.

> Does this restore the default style pre-theme modification or does it simply use a generic default value for this specific property?

It unsets the value, similar to removing it in UWPSpy, so it roughly uses a generic default value for this specific property. So if it isn't set by the app itself, this property won't be impacted, but if it's set by the app itself, then this value will be removed.

eugenesvk:
> That's how normally theme authors submit the themes:
> An author creates a theme with a styler mod via the Windhawk settings

This whole conversation with many examples illustrates that this is not as simple a step as you describe. So yes, in the end the submission process gets the cleaned styles, but then the development of those styles is anything but clean!
Like even for the basic visibilitiy: in my editor I can have ~10 styles in the vertical space it takes the Visual editor in Windhawk to display 2 lines (target/style), so there is no way I'm scrolling like mad when I have more than a **hundred of** styles (and the textual mode has other issues, some of which I've mentioned above)!

eugenesvk:
Update the https://github.com/ramensoftware/windows-11-settings-styling-guide. Note that the manual import styles are now out of sync with the CPP file, so don't forget to use your scripts to import the latest changes

m417z:
Thanks. I created a PR: https://github.com/ramensoftware/windhawk-mods/pull/4351
I made some additional changes, such as removing more leftovers from the start menu styler. You can try it to make sure it works correctly.

Also, I imported your theme from the readme yaml, and spotted two issues:
https://github.com/ramensoftware/windows-11-settings-styling-guide/blob/main/Themes/Densy/README.md?plain=1#L192
In the yaml, `\"24\"` shouldn't be escaped.
https://github.com/ramensoftware/windows-11-settings-styling-guide/blob/main/Themes/Densy/README.md?plain=1#L415
`xSystemSettings` should be `SystemSettings`.

Both are easy to fix, but you might want to take a look and make sure that after the fixes, the theme still looks correctly.

eugenesvk:
Thanks, removed the escapes and update the repo.

The updated mod seems to work.

xSystem is intentional "commenting" out a style, think that turned out to be redundant, but don't want to delete it because maybe it'll be needed later, the whole styling system is pretty messy. Though commented it out properly now

m417z:
> xSystem is intentional "commenting" out a style, think that turned out to be redundant, but don't want to delete it because maybe it'll be needed later, the whole styling system is pretty messy. Though commented it out properly now

You could also prefix it with `//`, i.e.:
```yaml
  - target: //SystemSettings.View.RootPage > Grid#RootPageGrid > Microsoft.UI.Xaml.Controls.NavigationView#PermanentNavigationView > Grid#RootGrid > Grid > SplitView#RootSplitView > Grid > Grid#ContentRoot > Border > Grid#ContentGrid > ContentPresenter#ContentPresenter > Frame#PermanentNavRootFrame > ContentPresenter > SystemSettings.View.L2Page#L2Page > Grid > Grid > SystemSettings.View.AlignableContentControl > ContentPresenter > Grid > SystemSettings.View.SettingsPageHost#pageContent > ScrollViewer#SettingsPageHostPanel > Border#Root > Grid > ScrollContentPresenter#ScrollContentPresenter > Grid#RootScrollableGrid > Grid > Grid > ContentControl > ContentPresenter > ItemsControl > ItemsPresenter > SystemSettings.View.SpacingStackPanel > ContentPresenter > SystemSettings.View.AlignableContentControl > ContentPresenter > SystemSettings.View.TwoSegmentsHeroUserControl#OneSegmentHeroEntityItemUserControl > Grid#LayoutRoot > Grid#LeftLayout > ContentPresenter > SystemSettings.View.EntityItem > Grid > ContentPresenter#IconContentPresenter > Grid > ContentControl > Grid > TextBlock
    styles:
      - Padding=0,0,0,0
      - Margin=0,0,10,0
      - Width=48
      - MinWidth=48
      - Height=48
      - MinHeight=48
```
The advantage is that it survives settings roundtrip/export. 

Either way, it's fine, thanks. I'll release the mod soon.

m417z:
Merged.

PR #4307, opened 2026-06-05
Hide insider watermark

ExiledEye:
## Changelog

* Changelog item 1...
* Changelog item 2...

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini


m417z:
### Submission review

---

The biggest problem is that this duplicates an existing mod and does so with a fragile mechanism, when a clean, proven one already exists in the repo.

**1. This overlaps an existing mod — the watermark can already be hidden cleanly.** [`desktop-watermark-tweaks`](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/desktop-watermark-tweaks.wh.cpp) by aubymori already hides the desktop watermark: setting **`display: false`** makes it disappear. It does this by hooking the single decision function shell32 uses, [`CDesktopWatermark::s_WantWatermark`](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/desktop-watermark-tweaks.wh.cpp#L39-L44), and returning `false` — no string filtering, no GDI scanning, build-independent. It's preferable to extend an existing mod over merging a near-duplicate. As submitted, this mod does strictly less than `desktop-watermark-tweaks` and in a less robust way, so it's hard to justify as a separate mod. If you believe there's a real gap (e.g. it doesn't cover some build), please raise it on that mod rather than ship a parallel implementation.

**2. Don't reinvent hooking — use Windhawk's facilities.** The hand-rolled `PatchIAT` plus manual `LoadStringW`/`ExtTextOutW` interception should be replaced by `WindhawkUtils::HookSymbols` / `WindhawkUtils::SetFunctionHook`. IAT patching only intercepts calls that go through *shell32's own import table*, depends on guessing which apiset/DLL the import resolves through (see #2), and silently misses any caller that doesn't route through that table. If you hook `CDesktopWatermark::s_WantWatermark` as in #1, the entire `PatchIAT` + `LoadStringW` + `ExtTextOutW` apparatus disappears.

**3. The `ExtTextOutW` text match is too broad.** `IsWatermarkText` does substring matches on `"Build "`, `"Evaluation copy"`, etc. `"Build "` in particular is common — any text shell32 draws through this path containing that substring (a folder/file name, a label) would be silently suppressed. If you keep this approach at all, match the full watermark format, not loose fragments. (Hooking `s_WantWatermark` makes this moot.)

<p>

Minor polish — none of this affects users, so it's your call. Most of it is moot if you switch to the symbol hook above.

- **Version-string drift in logs.** `@version` is `1.0`, but `Wh_ModInit` logs `v1.4` and `Wh_ModUninit` logs `v1.0 unloaded`. Drop the hardcoded version from the log strings (or use `WH_MOD_VERSION`) so they can't go stale.
- **`Hook_ExtTextOutW` should skip glyph runs.** When `opts & ETO_GLYPH_INDEX` is set, the buffer holds glyph indices, not characters, so scanning it as text is meaningless (and may match garbage). Bail out early in that case.
- **`Hook_LoadStringW` ignores `hInstance`.** It suppresses IDs 62000/62001 for *any* module. That's contained while scoped to shell32's IAT, but if you move to an inline hook (which fires for all callers) you'd also want to check `hInst == hShell32`.
- **Avoid allocating in the GDI hot path.** `IsWatermarkText` constructs a `std::wstring` on every `ExtTextOutW` call. Prefer a non-allocating search (`std::wstring_view` + `find`).

</p>
</details>

<p>

- **The `ExtTextOutW` IAT hook only catches text shell32 draws *directly* through its import table.** If the watermark is rendered via `DrawTextW`/`TextOutW`, or from another module, the call won't pass through shell32's IAT and the hook won't fire — which is likely why the `LoadStringW` backstop was added (and that one doesn't install — see #2). The `s_WantWatermark` hook sidesteps the question of *how* the text is drawn entirely.
- **`"Test Mode"` can't match here.** The Test Mode watermark is the test-signing watermark drawn by win32k/DWM, not by shell32, so it will never reach this hook. It's the wrong layer for that string.
- **The init repaint may not refresh the watermark.** `InvalidateRect(GetShellWindow(), ...)` invalidates Progman, which isn't necessarily where the watermark is painted — the README already notes a restart may be needed.

</p>
</details>

ExiledEye:
> ### Submission review
> 
> 
> The biggest problem is that this duplicates an existing mod and does so with a fragile mechanism, when a clean, proven one already exists in the repo.
> 
> **1. This overlaps an existing mod — the watermark can already be hidden cleanly.** [`desktop-watermark-tweaks`](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/desktop-watermark-tweaks.wh.cpp) by aubymori already hides the desktop watermark: setting **`display: false`** makes it disappear. It does this by hooking the single decision function shell32 uses, [`CDesktopWatermark::s_WantWatermark`](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/desktop-watermark-tweaks.wh.cpp#L39-L44), and returning `false` — no string filtering, no GDI scanning, build-independent. It's preferable to extend an existing mod over merging a near-duplicate. As submitted, this mod does strictly less than `desktop-watermark-tweaks` and in a less robust way, so it's hard to justify as a separate mod. If you believe there's a real gap (e.g. it doesn't cover some build), please raise it on that mod rather than ship a parallel implementation.
> 
> **2. Don't reinvent hooking — use Windhawk's facilities.** The hand-rolled `PatchIAT` plus manual `LoadStringW`/`ExtTextOutW` interception should be replaced by `WindhawkUtils::HookSymbols` / `WindhawkUtils::SetFunctionHook`. IAT patching only intercepts calls that go through _shell32's own import table_, depends on guessing which apiset/DLL the import resolves through (see #2), and silently misses any caller that doesn't route through that table. If you hook `CDesktopWatermark::s_WantWatermark` as in #1, the entire `PatchIAT` + `LoadStringW` + `ExtTextOutW` apparatus disappears.
> 
> **3. The `ExtTextOutW` text match is too broad.** `IsWatermarkText` does substring matches on `"Build "`, `"Evaluation copy"`, etc. `"Build "` in particular is common — any text shell32 draws through this path containing that substring (a folder/file name, a label) would be silently suppressed. If you keep this approach at all, match the full watermark format, not loose fragments. (Hooking `s_WantWatermark` makes this moot.)
> 
> Optional improvements
> 
> Minor polish — none of this affects users, so it's your call. Most of it is moot if you switch to the symbol hook above.
> 
> * **Version-string drift in logs.** `@version` is `1.0`, but `Wh_ModInit` logs `v1.4` and `Wh_ModUninit` logs `v1.0 unloaded`. Drop the hardcoded version from the log strings (or use `WH_MOD_VERSION`) so they can't go stale.
> * **`Hook_ExtTextOutW` should skip glyph runs.** When `opts & ETO_GLYPH_INDEX` is set, the buffer holds glyph indices, not characters, so scanning it as text is meaningless (and may match garbage). Bail out early in that case.
> * **`Hook_LoadStringW` ignores `hInstance`.** It suppresses IDs 62000/62001 for _any_ module. That's contained while scoped to shell32's IAT, but if you move to an inline hook (which fires for all callers) you'd also want to check `hInst == hShell32`.
> * **Avoid allocating in the GDI hot path.** `IsWatermarkText` constructs a `std::wstring` on every `ExtTextOutW` call. Prefer a non-allocating search (`std::wstring_view` + `find`).
> 
> Functionality notes
> 
> 
> * **The `ExtTextOutW` IAT hook only catches text shell32 draws _directly_ through its import table.** If the watermark is rendered via `DrawTextW`/`TextOutW`, or from another module, the call won't pass through shell32's IAT and the hook won't fire — which is likely why the `LoadStringW` backstop was added (and that one doesn't install — see [added brave to chrome-wheel-scroll-tabs #2](https://github.com/ramensoftware/windhawk-mods/pull/2)). The `s_WantWatermark` hook sidesteps the question of _how_ the text is drawn entirely.
> * **`"Test Mode"` can't match here.** The Test Mode watermark is the test-signing watermark drawn by win32k/DWM, not by shell32, so it will never reach this hook. It's the wrong layer for that string.
> * **The init repaint may not refresh the watermark.** `InvalidateRect(GetShellWindow(), ...)` invalidates Progman, which isn't necessarily where the watermark is painted — the README already notes a restart may be needed.

My bad.
I wanted a solution for this and searched for "insider" and found nothing, so I tried to replicate that program from 2015 in windhawk.
I'll abort the PR.

PR #4320, opened 2026-06-06
feat: add taskbar-multi-tray mod

EDM115:
`taskbar-multi-tray`, a mod that allows to have the tray icons & control center displayed on multiple monitors (or hide either of them, as well as choose which monitors get them)  
doesn't duplicate the elements, just display interactive "proxies" on the non-primary monitors

## Mod authorship

This mod was created by:

- - [x] Claude (Fable 5 in Zed, for complex bugfixing)
- - [x] ChatGPT (5.5 in Codex and 5.5 Pro, for all the main code & features)
- - [ ] Gemini


m417z:
Thanks for the submission, that's impressive! From a (very) quick try, I saw the following issues:

An empty space is added between the system tray icons and the clock:

image

On the secondary monitor, in addition to the spacing, there's a bug with the cut bell icon (DPI-related?), and the tray overflow button is missing.

image

Also, the mod doesn't seem to work if it's enabled on Explorer start. It's easy to test it by restarting explorer.exe, but it'd also happen after a reboot.

The CI fails on ARM64 incompatibility. To fix it, you can use this code:
https://github.com/m417z/my-windhawk-mods/blob/544ca828ae800f84183df7439194d2abb5ef7f03/mods/taskbar-notification-icon-spacing.wh.cpp#L801

Below is a preliminary AI-assisted review.

### Submission review

---

This is an ambitious mod, and a lot of careful work clearly went into the teardown/reload handling. The core concern is that it's also a very large, very aggressive piece of code injected into two critical shell processes (`explorer.exe` and `ShellHost.exe`), and the techniques it uses give it a wide blast radius. The items below are about reducing that surface and the self-documented instability.

**1. The mod's own README documents that it can crash the taskbar and is non-deterministic — that's the headline blocker.** The "Known issues" section lists *"On rare occasions, one of the taskbars can crash"*, *"flyouts can open on the wrong monitor"*, and *"risk of not being applied at startup"*. For a mod living inside `explorer.exe`/`ShellHost.exe`, a known crash path is a real problem — a crash there takes down the shell for the user. These need to be root-caused before this is safe to ship broadly, and several of the items below are likely contributors. Treat "the taskbar can crash" as must-fix, not a known-issue checkbox.

**2. `HookSymbols` is called 7 times against `taskbar.dll`.** `HookTaskbarDllSymbols` calls `HookSymbols(module, taskbarDllHooks, …)` once (line 4119) and then the `hookOptional` lambda calls `HookSymbols(module, hook, 1)` six more times (line 4128) for the same module. `HookSymbols` caches resolved symbols per module, and each extra call for the same module invalidates that cache and forces a re-resolution — i.e. you pay the full symbol-resolution cost up to 7× on every load. The `SYMBOL_HOOK` struct already has an `optional` flag (4th constructor arg, see `windhawk_utils.h:88`) precisely for "don't fail if this one is missing". Put every `taskbar.dll` hook in the single `taskbarDllHooks` array, mark the best-effort ones `optional = true`, and resolve them in one `HookSymbols` call.

**3. Remove the diagnostic-only private-symbol hooks — and note that two of them can block the whole mod from loading.** A large set of hooks exist purely to `Wh_Log_safe(...)` and do nothing else functionally:
   - All six "optional" hooks: `TaskbarModel::SetNotificationAreaIconManager2`, `get_NotificationAreaPromotedIcons`, `get_NotificationAreaOverflowIcons`, `NotificationAreaIconManager2::AddIconToVisibleCollection` / `RemoveIconFromVisibleCollection` / `ShellNotifyIcon`. The globals they populate (`g_lastPrimary/SecondaryNotificationAreaIconManager`) are only ever read back inside other log statements.
   - Two hooks in the **required** array that are also log-only: `TaskbarHost::ActivateOverflowFlyout` (line 4071) and `SystemTrayHost::TryLoadNotificationAreaFrameFromXamlExtension` (line 4096). Because they're in the required array, if either symbol drifts on a future Windows build, `HookSymbols` returns `false` → `HookTaskbarDllSymbols` returns `false` → `Wh_ModInit` returns `FALSE` and the entire mod fails to load — all to keep a diagnostic log line working.

   Every private-symbol hook is a maintenance and crash liability across Windows updates. Hooks that only emit verbose-off-by-default logs aren't worth that cost; deleting them also removes most of the extra `HookSymbols` calls from item 2.

**4. `@architecture x86-64` is declared, but the code hard-`#error`s on anything but x64.** `XamlRootFromTaskbarHostSharedPtr` reads the `TaskbarHost::FrameHeight` prologue under `#if defined(_M_X64)` with `#else #error "Unsupported architecture"` (lines 3430-3443). `x86-64` advertises native ARM64 coverage for predefined shell processes like `explorer.exe`, so on an ARM64 device Windhawk will try to build the mod natively as ARM64 and hit the `#error` — the mod simply won't compile there. Implement the ARM64 path.

**5. `OpenProxyFlyout` blocks the taskbar UI thread and drives flyouts with synthetic input.** It runs on the taskbar's own thread (it's invoked from `TaskbarSubclassProc` on the registered proxy message), and it `Sleep(120)`s there before injecting `SendWinHotkey(L'N')` / mouse clicks via `SendInput`. Sleeping on the taskbar thread freezes the taskbar for that window, and synthesizing global Win+N / mouse-down-up is fragile — it races real user input, can fire while the user is typing, and breaks if the shortcut is remapped/disabled. This is a workaround for "open the right flyout on the right monitor" rather than a real fix; if there's any way to drive the flyout through the taskbar host API on the target monitor instead of synthetic input, that would be far more robust. At minimum, don't block the UI thread.

**6. The "primary display" spoof is process-wide in `explorer.exe`, not scoped to the tray.** In `selected` mode, `EnumDisplayDevicesW_Hook` (lines 1752-1793) and the `MonitorFromPoint((0,0))` / `MonitorFromRect((0,0,0,0))` redirects (lines 1337, 1396) report the selected monitor as primary to **every** caller in `explorer.exe`, not just the tray code, and persistently (not gated to an active flyout). Lots of shell code calls `EnumDisplayDevicesW`/`MonitorFromPoint(0,0)` to find the primary monitor for unrelated purposes (wallpaper, default window placement, etc.), and lying to all of them is a plausible source of the "a taskbar can crash"/odd-placement issues. ShellHost is already gated behind `HasActiveKnownFlyoutContext()` — explorer should be scoped at least as tightly (to the specific operations/windows that need it) rather than always-on.

**7. `BroadcastShellHookDisplayChange` posts to `HWND_BROADCAST`.** `PostMessage(HWND_BROADCAST, RegisterWindowMessage(L"SHELLHOOK"), 35, 0)` (line 1797) sends to *every* top-level window on the desktop, in every process — and it's fired on every apply, settings change, proxy open, restore, and deferred retry. Any app that listens for shell-hook notifications gets these synthetic events. Also, `35` isn't a documented `HSHELL_*` code (`HSHELL_MONITORCHANGED` is `0x10`), so it's unclear what this is meant to do. Please scope this to the windows that actually need it (your own taskbars) and document/justify the `35`, or drop it if the `SendMessage(taskbarWnd, 0x5B8, …)` on the line above is already sufficient.

**8. License: GPLv3 code is being relicensed as MIT.** A large amount of the taskbar plumbing is copied (near-verbatim, only reformatted) from m417z's taskbar mods — `XamlRootFromTaskbarHostSharedPtr` including the `TaskbarHost::FrameHeight` byte-pattern offset scan, `RunFromWindowThread`, `GetTaskbarXamlRoot`, `FindCurrentProcessTaskbarWnd`, and the `CTaskBand::GetTaskbarHost` / `std::_Ref_count_base::_Decref` symbol hooks. Those mods state *"Source code is published under The GNU General Public License v3.0"* ([example](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/taskbar-notification-icon-spacing.wh.cpp#L15)). GPLv3 code can't be redistributed under MIT — set `@license` to `GPL-3.0` (or another GPLv3-compatible license).

<p>

- **Dead settings helper / NULL check.** `ReadStringSetting` (lines 738-747) treats `Wh_GetStringSetting` as if it can return `NULL` and falls back to a manual default. `Wh_GetStringSetting` never returns `NULL` (it returns `L""`), and Windhawk already substitutes the YAML default for unset values, so the `value ? value : defaultValue` branch and the `defaultValue` parameter are effectively dead. Prefer `WindhawkUtils::StringSetting` (RAII) over the raw get + `Wh_FreeStringSetting`.
- **Dead `kProxyControlCenter` branch.** `HitTestProxyFlyout` never returns `kProxyControlCenter` (control-center clicks now produce `kNativeControlCenter`), so the `if (flyout == kProxyControlCenter)` path in `OpenProxyFlyout` (line 3234) — the `WaitForLeftButtonRelease(800)` + `SendMouseClickAtCursor()` block — is unreachable. Remove it (and `WaitForLeftButtonRelease`/`SendMouseClickAtCursor` if nothing else uses them) or wire it back up.
- **Trim the README log-filter dump.** The ~90-line "Useful log filters" list (lines 234-312) is debugging scaffolding; it bloats the published README. A short pointer to "enable verbose logging and filter on the mod name" would read better.

</p>
</details>

<p>

Non-critical observations about the feature behavior and robustness.

- **Cached XAML bindings can crash the host at shutdown.** `g_primaryNotificationAreaIconsBinding`, `g_primaryNotifyIconStackBinding`, … are globals holding `winrt::…::IInspectable` (COM) references. You clear them in `Wh_ModBeforeUninit`/`Wh_ModSettingsChanged`, but on a normal process exit (Explorer restart, sign-out, reboot) `Wh_ModUninit` does **not** run — the CRT instead runs these globals' destructors during `DLL_PROCESS_DETACH`, which calls `Release()` on XAML objects potentially after the XAML/COM runtime is already torn down. That's a classic shutdown-crash shape and a candidate for the README's "a taskbar can crash". Consider holding these via a heap pointer that's intentionally leaked at shutdown (trivial destructor), with the real release kept in `Wh_ModUninit`.
- **Hot WinAPI hooks do work before the cheap early-out.** `SetWindowPos_Hook` / `DeferWindowPos_Hook` call `GetWindowRect` and build a `RECT` on every invocation before `TranslateNativeFlyoutRectToActiveMonitor` checks whether any flyout context is even active; `DispatchMessageW_Hook` and the `MonitorFrom*` hooks run on every call too. In `explorer.exe` these are extremely hot paths. Checking the active-context flag first (and bailing out before `GetWindowRect`) would cut the steady-state overhead to near zero.
- **Binding reuse across taskbars is inherently fragile.** Copying `ItemsSource`/`DataContext`/`Content` from the primary tray onto secondary controls depends on internal XAML structure that Microsoft can change at any update; the README already acknowledges content can't always be synthesized. No action needed — just flagging that this is the part most likely to silently break on a Windows update.

</p>
</details>

EDM115:
thanks for the review 😄  
i'm currently trying to fix most of the reported issues, only one stays and that's ensuring the icon-dependent right-click menu on the control center works without crashing taskbars. it either does that or work only on the laptop (non-primary) screen  
i also can only test that at work since i have 3 monitors there but only 1 at home. progress will be slow lol. other than that some code cleanup, perf improvements and bug fixes have been made, and I think i'll reuse logic from https://github.com/ramensoftware/windhawk-mods/blob/main/mods/start-menu-open-location.wh.cpp to properly position the flyouts instead of battling with windows not registering clicks properly and then hacking my way into repositioning said flyouts

EDM115:
okay I finally fixed some crazy edge cases, I had to reach into vtables since XAML didn't exposed some stuff...  
only 1 small bug left to fix and then I'll try to optimize this :) will push v1.1.0 with these changes

EDM115:
heyyy @m417z I believe we should be good now 😄  
the workflow failure is due to the twitter validation but I believe you received my message ?  
also for a few hard bugs I had to resort to Fable 5 Max as any other model absolutely sucked at debugging them and finding what to do

m417z:
Right, I got the Twitter validation, and the workflow will pass in future submissions once this one is merged. Thanks for the update, I'll take a look soon. For now, one quick note is that the readme is full with information that users are unlikely to be interested in, such as the dev versions changelog and fixed bugs during development.

EDM115:
well initially I planned on "hiding" them behind a details tag, like this :
```md
<details><summary>Changelog</summary>

blablala

</details>
```
but then I noticed Windhawk's markdown parser doesn't support them. Pretty logical that we can't just use any HTML tag lmao, but a bummer still. I can remove those sections if you prefer

m417z:
Well, I don't think it has value for users. Windhawk maintains a changelog of released versions, and users are unaware of unreleased versions. Fixed bugs never affected anyone, so its value is also questionable. Remove it, or, if you prefer on keeping it for some reason, move it to a separate comment block in the code.

m417z:
Regarding the empty space, it's still there. If I enable battery percentage, it appears there. Perhaps the mod forces all elements to be visible for some reason?

image

Other issues from a short testing session:
* If all icons are hidden in the overflow area, the secondary tray area doesn't appear.
* Moving from/to the overflow area is buggy, even in the primary taskbar.

Regarding vtable hooks - have you considered using regular hooks instead? vtable hooks may cause crashes if multiple mods hook the same function. Here's an example of hooking a function from a vtable without vtable patching:
https://github.com/ramensoftware/windhawk-mods/blob/625c479f1d6f59a39cfde77301a4895444fdf429/mods/taskbar-multirow.wh.cpp#L605-L612

Another issue I saw is that when a new taskbar is created, it's not being handled. The logs I see:
```
13:10:22.961 6988 explorer.exe  [WH] [local@taskbar-multi-tray] [4619:CSecondaryTray_InitModelAndHost_Hook]: taskbar-multi-tray : CSecondaryTray::InitModelAndHost hook
13:10:23.114 6988 explorer.exe  [WH] [local@taskbar-multi-tray] [1919:LogTaskbarWindow]: taskbar-multi-tray : ApplyStyle hwnd=0x000000000049009A class=Shell_SecondaryTrayWnd monitor=2 device=\\.\DISPLAY2 rect=(-1920,1008)-(0,1080)
13:10:23.116 6988 explorer.exe  [WH] [local@taskbar-multi-tray] [2317:CollectTrayElements]: taskbar-multi-tray : SystemTrayFrameGrid not found (apply)
```

Also, see the AI review below. I haven't verified the collapsed notes, but the functionality notes have some notes that, if correct, are worth addressing.

Generally, it seems to be working and I believe it's ready for the first release. It doesn't have to be perfect right away, but let's address these items.

---

- **Data race on the process-local flyout-context globals.** The `MonitorFrom*`/placement hooks fire on arbitrary Explorer threads and all funnel through `GetActiveProxyFlyoutMonitor()`, which both reads *and writes* the non-atomic `g_proxyFlyout*` globals (it zeroes them on expiry at `taskbar-multi-tray.wh.cpp:928`, and adopts a shared snapshot via `CopySharedProxySnapshotToLocal`). The cross-process shared state is seqlock-protected, but this process-local mirror is not. Word-sized fields won't tear on x64 so it won't crash, but `CopySharedProxySnapshotToLocal` writes multi-field `POINT`/`RECT` values non-atomically, so a concurrent reader can briefly observe a half-updated context and redirect a flyout to the wrong monitor. Since these hooks run during normal operation (not just settings change/unload), consider deriving the context purely from a single seqlock snapshot per call rather than mutating shared globals from inside the read path.

- **Custom logging duplicates Windhawk's built-in toggle.** `enableVerboseLogging` + the `Wh_Log_safe` wrapper re-implement the enable/disable logging that Windhawk already provides per-mod, and every log string is prefixed with `taskbar-multi-tray :` even though Windhawk already prepends the mod name. The convention is to call `Wh_Log` directly and let Windhawk gate it. I understand the motivation here is to avoid flooding logs from the hot hooks, so this one is defensible — but the per-line prefix is pure redundancy and could be dropped. (`enableTreeDump` as a separate noisy-diagnostic toggle is fine.)

<p>

- **`ReadStringSetting` reinvents `WindhawkUtils::StringSetting`** (RAII string setting, no manual `Wh_FreeStringSetting`). Also, the `value ? value : L""` guard is unnecessary — `Wh_GetStringSetting` never returns `NULL`, it returns `L""` on unset/error.
- **Dead branches behind `WantsNotificationCenter()`.** That helper is hardcoded to always return `true` (`taskbar-multi-tray.wh.cpp:855`), so the `else` arms that collapse `NotificationCenterButton` (e.g. `ApplyStyle` at `:2921`) are unreachable. Either wire it to a real setting or drop the dead `else` paths.

</p>
</details>

<p>

- **Click hit-testing is a fixed-layout heuristic.** `HitTestProxyFlyout` reconstructs which surface was clicked from hardcoded DPI-scaled widths measured off the right edge (`[show desktop][notif/clock][control center][icons][chevron]`). Any extra item Windows or another mod injects into that strip (language bar, emoji/pen menu, etc.) shifts the math and breaks the hit-test — which you already document under "Known issues." There isn't a clean alternative short of intercepting the actual XAML click target, so this is just an FYI, not something to fix now.
- **Hot-plugged secondary taskbars don't get the subclass installed in that path.** `CSecondaryTray_InitModelAndHost_Hook` calls `ApplyStyle` but not `InstallTaskbarSubclass`, so a newly-created secondary taskbar's copied surfaces aren't hit-tested for left-click flyout arming until the next full apply pass (deferred retry or settings change). The placement-hook cursor rescue covers most of this, so the visible impact is small.
- **`EnumDisplayDevicesW` primary-device spoofing is process-wide while a context is armed.** For up to a few seconds after a click, every caller of `EnumDisplayDevicesW` in the process sees a non-primary monitor flagged as primary. This is central to steering ShellHost's flyout placement and there's no narrower hook point, so it's noted only as blast-radius awareness.

</p>
</details>

mods/taskbar-multi-tray.wh.cpp
```diff
@@ -0,0 +1,4864 @@
...
+    RemoveWindowSubclass(hWnd, TaskbarSubclassProc, kTaskbarSubclassId);
+}
+
+/// Extracts the XamlRoot from a TaskbarHost std::shared_ptr without knowing the class layout : the offset of the hosted XAML element is read out of TaskbarHost::FrameHeight's prologue bytes (x64 and ARM64 encodings below), then the shared_ptr's control block is released via std::_Ref_count_base::_Decref
+XamlRoot XamlRootFromTaskbarHostSharedPtr(void* taskbarHostSharedPtr[2]) {
+    if (!taskbarHostSharedPtr[0] && !taskbarHostSharedPtr[1]) {
+        return nullptr;
+    }
+
+    size_t taskbarElementIUnknownOffset = 0x48;
```
m417z: Change to 0x10 which is the value recent builds use and is a safer default.

mods/taskbar-multi-tray.wh.cpp
```diff
@@ -0,0 +1,4864 @@
...
+// @include         explorer.exe
+// @include         ShellHost.exe
+// @architecture    x86-64
+// @compilerOptions -lole32 -loleaut32 -lruntimeobject -lcomctl32
+// @license         MIT
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# `taskbar-multi-tray`
```
m417z:
```suggestion
# Taskbar multi-tray
```

mods/taskbar-multi-tray.wh.cpp
```diff
@@ -0,0 +1,4864 @@
...
+## Windows 11 Windhawk mod for controlling taskbar tray and control-center visibility across multi-monitor setups
```
m417z: I think it's too much text for a headline.

```suggestion
Windows 11 Windhawk mod for controlling taskbar tray and control-center visibility across multi-monitor setups.
```

EDM115:
alright I think I have everything implemented and thx for the code to hook without patching :)  
i just have to fix one last bug (moving icons from hidden tray area to visible across monitors) and i'll push

EDM115:
alright here's the changes.  
overall the only issue left is that moving a hidden tray icon from the flyout of a proxied taskbar won't work when dropped in that same proxied taskbar's visible tray icons area. users have to either drag from the primary display's flyout or drag from a proxied one to the primary display's taskbar tray icons area  
also it works when the battery percentage is disabled. and non-proxied (yet) elements such as language selector, emojis button, ... don't crash and are simply left on the primary monitor.  
oh, and no more vtables patching :)

m417z:
Preliminary review is below. Haven't tried it yet.

Also, regarding the `volatile` casts, I suggest to replace them with `std::atomic`, it will make the code more correct and easier to read and review.

### Submission review

---

- **The `ShowAt` function hooks are registered but never applied, so layer 2/3 of the control-center crash fix is inactive.** `EnsureFlyoutShowAtFunctionHooks` calls `Wh_SetFunctionHook` for the four FlyoutBase `ShowAt` targets ([line 3530](https://github.com/ramensoftware/windhawk-mods/blob/37fc5c1b212d06714974cfdd759c3e8c8005ce1f/mods/taskbar-multi-tray.wh.cpp#L3530)), but it runs on the taskbar thread from inside `ApplySettingsFromTaskbarThread` ([line 4574](https://github.com/ramensoftware/windhawk-mods/blob/37fc5c1b212d06714974cfdd759c3e8c8005ce1f/mods/taskbar-multi-tray.wh.cpp#L4574)) — i.e. *after* `Wh_ModInit` has already returned — and there is no `Wh_ApplyHookOperations()` call anywhere in the mod. Per the API, hooks registered with `Wh_SetFunctionHook` only become active on the next `Wh_ApplyHookOperations`, which Windhawk calls automatically **only once, right after `Wh_ModInit`** (see `windhawk_api.h` comments on `Wh_SetFunctionHook`/`Wh_ApplyHookOperations`). So these four hooks are queued but never patched in: `MenuFlyout_ShowAt_FunctionHook` & friends never fire, `PrepareControlCenterFlyoutShowAt` never runs, and the per-island proxy re-root never happens. That's exactly the path your readme describes as load-bearing for *not* crashing Explorer when right-clicking a control-center glyph (network/volume/battery) on a non-owner monitor.

This looks like a regression from the 1.2.0 switch away from vtable patching (1.1.0 patched the vtable in memory directly, which needed no apply step; `Wh_SetFunctionHook` does). Fix: call `Wh_ApplyHookOperations()` once after the four `RequestFlyoutShowAtFunctionHook` calls in `EnsureFlyoutShowAtFunctionHooks`, before setting `g_flyoutShowAtHooksInstalled = 1`. The function is already guarded so it runs once, so the documented "very slow" cost is paid a single time. Please re-verify on a real multi-monitor setup that right-clicking those glyphs on a secondary monitor doesn't crash Explorer with the *current* build — if it doesn't, that itself is worth understanding, because as written the interception isn't installed.

- `ClearCachedXamlBindings` ([line 4024](https://github.com/ramensoftware/windhawk-mods/blob/37fc5c1b212d06714974cfdd759c3e8c8005ce1f/mods/taskbar-multi-tray.wh.cpp#L4024)) releases cached XAML/view-model references, and it runs from `Wh_ModBeforeUninit` / `Wh_ModSettingsChanged`, which execute on an arbitrary thread rather than the taskbar/UI thread. It's very likely benign because the live XAML tree still holds those objects (so the mod's release is never the last ref and no off-thread destructor runs), but releasing UI-thread-affine COM objects off-thread is the kind of thing that bites later — consider clearing these on the taskbar thread (you already marshal teardown there via `RunFromWindowThread`).

- **`g_taskbarTrayMetrics` synchronization.** This `std::vector` is mutated on the taskbar thread (`ApplyStyle`, `WM_NCDESTROY`) and read by the hit-test helpers. Most reads are on the taskbar thread too, but the `DispatchMessageW` -> `PreArmTrayIconDragDropOwnershipForMessage` path resolves the taskbar from the live cursor point (`FindTaskbarWindowFromScreenPoint`) and can read it from a different thread; a concurrent `push_back` reallocation would then be a data race. The window is narrow (drag in another window with the cursor over a taskbar), but if you ever see rare unexplained crashes here, a lock or restricting these reads to the owning thread would close it.

<p>

- `ReadStringSetting` ([line 647](https://github.com/ramensoftware/windhawk-mods/blob/37fc5c1b212d06714974cfdd759c3e8c8005ce1f/mods/taskbar-multi-tray.wh.cpp#L647)) is a raw `Wh_GetStringSetting` + `Wh_FreeStringSetting`. `WindhawkUtils::StringSetting` (RAII) is the idiomatic form. Also note `Wh_GetStringSetting` never returns `NULL` (it returns `L""`), so `std::wstring result = value;` is fine but the helper isn't really needed.
- `enableTreeDump` is a developer/diagnostic toggle exposed in the end-user settings UI. Harmless, but most users will never want it; consider whether it needs to be a user-visible setting.

</p>
</details>

<p>

Non-critical observations about the approach and feature behavior itself.

- **Hit-testing is a right-edge pixel heuristic.** `HitTestProxyFlyout` / `IsControlCenterClientPoint` / `IsTrayClientPoint` reconstruct the layout from DPI-scaled fixed widths plus a measured promoted-icon/control-center width. As you already note in "Known issues", extra items injected into the tray strip (language switcher, emoji/IME, stylus menu, other mods) or non-bottom/vertical taskbars will throw the mapping off. Inherent to the copied-surface approach; noting it for completeness.

</p>
</details>

m417z:
I made some testing. Feels more stable, great work! Things I noticed:
* Switching between the two tray areas is buggy sometimes. Example: Enable the mod, right click on the volume icon, then right click on the other volume icon, and explorer often crashes. 
* I expected the "Components" option to affect only the mirrored taskbars, but it also affects the primary taskbar. Is it intentional?

EDM115:
indeed some bugs slipped through 😅  
the crashes when right clicking control center between taskbars was a regression introduced when we moved away from vtables patching, i should have tested that but forgot  
> I expected the "Components" option to affect only the mirrored taskbars, but it also affects the primary taskbar

honestly I always viewed it like this but now that you say it, it makes sense to let users choose ! added a setting to either apply the Components settings to all taskbars or just the mirrored ones  
also I forgot last time to mirror my README updated known issues list to the mod embedded one, it's done now. and other issues and improvements have been implemented 👍

EDM115:
regarding the other PR (#4413), I also believe there's some nice ideas there and things that might be done better  
ultimately the decision is yours. also can a mod have multiple `@author` ? because if possible, we just take the best of both worlds but still credit both contributors

m417z:
> we just take the best of both worlds but still credit both contributors

I think that'd be ideal, as I also commented in that PR.

> can a mod have multiple `@author` ?

Mods can only have a single author - the person responsible for the mod and for updating it. Additional credits can be added to the mod readme.

m417z:
Optional AI review comments are below, feel free to skip if you prefer. I think it's ready for the first release. If it's ready on your side too, let me know and I'll merge it. Thanks!

---

<p>

- **Route the mod's own monitor queries through the unhooked original.** `IsRealPrimaryTrayTargetWindow` (line ~2552) calls the *hooked* `MonitorFromWindow` rather than `GetActualMonitorFromWindow`. It happens to be safe today because every caller runs under `ActiveFlyoutRedirectionSuppressor` (so the hook passes through), but that's a non-obvious invariant; using `GetActualMonitorFromWindow` like the rest of the mod removes the footgun if a future caller is added outside the suppressor.
- **Dead parameter path.** `SetContextGesturesRecursive`'s `clearContextFlyouts` branch (line ~2734) is never exercised — both `UpdateSharedSurfaceContextGestures` and `EnableSharedSurfaceContextGestures` pass `false`, and the comment already says it's "supported but unused." The parameter, the `ContextGestureCounts::contextFlyouts` field, and the branch can go.
- **Redundant wrappers.** `EnableSharedSurfaceContextGestures` and `UpdateSharedSurfaceContextGestures` are byte-for-byte identical (`UpdateContextGesturesForElement(..., true, false)`); `GetActiveFlyoutTargetMonitor()` just forwards to `GetActiveProxyFlyoutMonitor()`; `ClearProxyFlyoutMonitorState()` just forwards to `ClearSharedProxyFlyoutMonitorState()` discarding its return. These are the kind of one-line indirections that tend to accumulate in AI-assisted code — collapsing them would trim the surface area a little.
- **Subclassing convention.** You use raw `SetWindowSubclass`/`RemoveWindowSubclass` and carefully ensure every call runs on the window's own thread (via `RunFromWindowThread` / on-thread hooks), which is correct. The repo convention is still `WindhawkUtils::SetWindowSubclassFromAnyThread` / `RemoveWindowSubclassFromAnyThread`, which removes the need to hand-marshal and is what other taskbar mods use — worth considering for consistency.

</p>
</details>

<p>

Non-critical observations about the approach itself — mostly "suboptimal but probably no clean alternative," noted as FYIs rather than required changes.

- **Process-wide blast radius of the user32 hooks.** The mod hooks nine user32 functions (incl. the very hot `MonitorFromWindow`, `SetWindowPos`, `DeferWindowPos`, `DispatchMessageW`) in both `explorer.exe` and `ShellHost.exe`. Steady state is gated cheaply (a single atomic read settles the no-context case), and the `MonitorFrom*` redirection is well-filtered (only the clicked taskbar, known popup classes, ambiguous/same-monitor points). The one I'd watch is **`EnumDisplayDevicesW`**: while a flyout context is armed (~0.65–5s) it rewrites the `DISPLAY_DEVICE_PRIMARY_DEVICE` flag for *every* caller in the process, not just flyout-placement code — so any other component enumerating displays in that window briefly sees a falsified primary. It's bounded and there's no obvious per-flyout API to steer ShellHost's display-enumeration-based placement, so this is an FYI, not a request to change.
- **Fragility against Windows updates.** The mod leans on a lot of undocumented internals: vtable-layout walking, function-prologue byte parsing, hardcoded `IFlyoutBase`/`IFlyoutBase5` vtable slot indices (14/12), the private `0x5B8` display-change message, XAML element names (`SystemTrayFrameGrid`, `NotifyIconStack`, `ControlCenterButton`, …), and ShellHost popup class names. You've guarded the load-bearing parts well (refuse to load if `taskbar.dll` symbols drift; degrade gracefully when `twinui.pcshell` symbols are missing; prologue parsing falls back to `0x10`), so a build change is more likely to silently disable a feature than to crash — but expect ongoing maintenance as 24H2+ evolves, and the slot-index/class-name assumptions are the most likely to break quietly.
- **Acknowledged limitations** (already in the README's *Known issues*): dragging icons from the hidden-icons flyout back to the visible tray doesn't work on copied taskbars, and injected extra tray items (language bar, IME/emoji, other mods) aren't replicated. Calling these out up front is the right move; just flagging that they're inherent to the singleton-binding-sharing design rather than quick fixes.

</p>
</details>

EDM115:
> I think that'd be ideal, as I also commented in that PR.

i'll wait for @RYJASM input before reusing any of his code 😄

> Mods can only have a single author

ah damn :(

> Optional AI review comments are below, feel free to skip if you prefer. I think it's ready for the first release. If it's ready on your side too, let me know and I'll merge it.

I did fix a few things and now I think it's all good !  
also should I reset the versionning back to v1.0.0 for the release or not ?

RYJASM:
Take whatever you'd like :)

Mine still has the crash from right clicking the network/volume icons,
though I have been working a bit on trying to prevent that.

Ryan Smith

EDM115:
alright 😁 i'll probably include some of your better logic on a future update ! time to get back to work for now lol

m417z:
Congrats! 🎉 

Small notes:
* The idea behind the `WindhawkUtils::SetWindowSubclassFromAnyThread` suggestion was that you could get rid of the `RunFromWindowThread` calls, such as `RunFromWindowThread(taskbarWnd, RemoveTaskbarSubclassesFromTaskbarThread, nullptr);`, and simplify the code. It's not incorrect now, just redundant.
* A small bug I noticed: Right clicking on a tray icon, and then clicking on the same icon in the other tray doesn't work, there's some jumpy animation instead. Right clicking on it again works. Minor.

PR #4321, opened 2026-06-06
New mod - Taskbar Fluent Media Player 1.2.0

Salyts:
**Taskbar Fluent Media Player —** is a Windhawk mod that integrates a modern media player with Fluent Design directly into the Windows 11 taskbar. It allows you to control music and view track information seamlessly without interrupting your workflow.

img

### Key Features:
* **Deep Integration —** Place the player in the tray area (left or right of the clock), next to system icons, or in the center of the taskbar.
* **Fluent UI Effects —** Full support for native Windows 11 materials including **Acrylic**, **Mica**, and **Mica Alt**.
* **Adaptive Design —** The player can automatically adapt its colors based on the album art or follow the system accent color.
* **Full Media Control —** Buttons for Play/Pause, Skip, 5-second Seek, Shuffle, and Repeat modes.
* **Smart Behavior —** Automatically hides the player when no media is playing, when entering full-screen mode, or after a period of inactivity (idle).

### Advanced Customization:
* **Visuals —** Customize fonts (Segoe UI, Aptos, etc.), text sizes, padding, and element transparency.
* **Album Art —** Adjustable corner radius, source app icon overlay, and pause overlay effects.
* **Mouse Interactions —** Assign custom actions (Volume control, Track seeking, Session switching) to the mouse wheel, single clicks, or double clicks on different parts of the player.

This mod was created by:

- - [x] Claude
- - [ ] ChatGPT
- - [x] Gemini

### DIFF
--- a/mods/taskbar-fluent-media-player.wh.cpp
+++ b/mods/taskbar-fluent-media-player.wh.cpp
@@ -1013,6 +1013,8 @@ static void ApplySettings();
 
 static std::atomic<bool> g_unloading{false};
 static std::atomic<bool> g_applyingSettings{false};
+static std::atomic<bool> g_hookInjectionInProgress{false};
+static std::atomic<int>  g_hookFailCount{0};
 
 static IMMDeviceEnumerator* g_pDeviceEnumerator = nullptr;
 
@@ -3030,10 +3032,6 @@ static void RefreshThemeColors();
 
 static std::atomic<bool> g_themeRefreshPending{false};
 
-static bool ShouldRefreshForSystemThemeChange() {
-    return g_settings.autoTheme;
-}
-
 static DWORD WINAPI TimerThreadProc(void*) {
     static bool lastThemeWasLight = IsSystemLightTheme();
 
@@ -3045,7 +3043,7 @@ static DWORD WINAPI TimerThreadProc(void*) {
 
     while (!g_unloading) {
         HANDLE handles[] = {g_timerStopEvent, hEvent, g_timerUpdateEvent};
-        DWORD wait = WaitForMultipleObjects(3, handles, FALSE, INFINITE);
+        DWORD wait = WaitForMultipleObjects(3, handles, FALSE, 500);
 
         if (wait == WAIT_OBJECT_0) break;
         if (g_applyingSettings) continue;
@@ -3058,16 +3056,11 @@ static DWORD WINAPI TimerThreadProc(void*) {
         }
 
         if (wait == WAIT_OBJECT_0 + 1) {
-            if (ShouldRefreshForSystemThemeChange()) {
+            if (g_settings.autoTheme) {
                 bool currentThemeIsLight = IsSystemLightTheme();
                 if (currentThemeIsLight != lastThemeWasLight) {
                     lastThemeWasLight = currentThemeIsLight;
-                    RunFromWindowThread(hWnd, [](void*) {
-                        if (!g_unloading && !g_applyingSettings && g_playerGrid) {
-                            RefreshThemeColors();
-                            UpdateVisibility();
-                        }
-                    }, nullptr);
+                    g_needsUiUpdate = true;
                 }
             }
             if (hKey) {
@@ -5044,13 +5037,17 @@ static bool InjectPlayerGrid() {
             Wh_Log(L"InjectPlayerGrid: Called UpdateLayout on injectionParent");
         }
 
-        RefreshThemeColors();
+        if (g_playerGrid.ActualWidth() == 0.0 && g_playerGrid.ActualHeight() == 0.0) {
+            Wh_Log(L"InjectPlayerGrid: XAML not laid out yet, scheduling deferred update");
+            g_needsUiUpdate = true;
+            if (g_timerUpdateEvent) SetEvent(g_timerUpdateEvent);
+        }
 
         auto dispatcher = g_playerGrid.Dispatcher();
         if (dispatcher) {
             try {
                 dispatcher.RunAsync(winrt::Windows::UI::Core::CoreDispatcherPriority::Low, [=]() {
-                    RefreshThemeColors();
+                    if (!g_unloading && g_playerGrid) RefreshThemeColors();
                 });
             } catch (...) {
                 Wh_Log(L"InjectPlayerGrid: Failed to dispatch RefreshThemeColors");
@@ -5912,19 +5909,49 @@ static IconView_IconView_t IconView_IconView_Original = nullptr;
 
 static void* WINAPI IconView_IconView_Hook(void* pThis) {
     auto r = IconView_IconView_Original(pThis);
-    if (!g_unloading && !g_playerGrid) {
-        HWND hWnd = FindCurrentProcessTaskbarWnd();
-        if (hWnd) {
-            g_taskbarWnd = hWnd;
-            RunFromWindowThread(hWnd, [](void*) {
-                if (!g_playerGrid && !g_unloading) {
-                    ApplySettings();
-                    if (g_playerGrid) ShowSuccessNotification();
-                    else Wh_Log(L"IconView_IconView_Hook: ApplySettings did not produce a player grid");
-                }
-            }, nullptr);
+
+    if (g_unloading || g_playerGrid || g_hookInjectionInProgress)
+        return r;
+    if (g_hookFailCount >= 3)
+        return r;
+
+    HWND hWnd = FindCurrentProcessTaskbarWnd();
+    if (!hWnd) return r;
+
+    DWORD wndPid = 0;
+    GetWindowThreadProcessId(hWnd, &wndPid);
+    if (wndPid != GetCurrentProcessId()) return r;
+
+    bool expected = false;
+    if (!g_hookInjectionInProgress.compare_exchange_strong(expected, true))
+        return r;
+
+    g_taskbarWnd = hWnd;
+
+    std::thread([hWnd]() {
+        Sleep(400);
+
+        if (g_unloading || g_playerGrid) {
+            g_hookInjectionInProgress = false;
+            return;
         }
-    }
+
+        RunFromWindowThread(hWnd, [](void*) {
+            if (!g_playerGrid && !g_unloading) {
+                ApplySettings();
+                if (g_playerGrid) {
+                    g_hookFailCount = 0;
+                    ShowSuccessNotification();
+                    Wh_Log(L"IconView_IconView_Hook: Injection successful");
+                } else {
+                    g_hookFailCount++;
+                    Wh_Log(L"IconView_IconView_Hook: ApplySettings failed, attempt %d/3", (int)g_hookFailCount);
+                }
+            }
+            g_hookInjectionInProgress = false;
+        }, nullptr);
+    }).detach();
+
     return r;
 }
 
@@ -6001,6 +6028,15 @@ static HMODULE GetTaskbarViewModule() {
 }
 
 BOOL Wh_ModInit() {
+    g_unloading = false;
+    g_applyingSettings = false;
+    g_hookInjectionInProgress = false;
+    g_hookFailCount = 0;
+    g_taskbarViewDllLoaded = false;
+    g_taskbarWnd = nullptr;
+    g_needsUiUpdate = false;
+    g_themeRefreshPending = false;
+
     LoadSettings();
 
     if (!HookTaskbarDllSymbols()) {
@@ -6064,6 +6100,8 @@ void Wh_ModAfterInit() {
 
 void Wh_ModUninit() {
     g_unloading = true;
+    g_hookInjectionInProgress = false;
+    g_hookFailCount = 0;
 
     StopTimerThread();
     StopIdleTimer();
--- a/mods/taskbar-fluent-media-player.wh.cpp
+++ b/mods/taskbar-fluent-media-player.wh.cpp
@@ -6399,14 +6399,39 @@ static HMODULE WINAPI LoadLibraryExW_Hook(LPCWSTR path, HANDLE file, DWORD flags
             }
         }
         if (isCandidate) {
+            Wh_Log(L"LoadLibraryExW_Hook: Found candidate module: %s", base);
             WindhawkUtils::SYMBOL_HOOK systemTrayDllHooks[] = {{
                 {LR"(void __cdecl TrayUI_StartTaskbar(void *))"},
                 &TrayUI_StartTaskbar_Original,
                 TrayUI_StartTaskbar_Hook,
             }};
             if (WindhawkUtils::HookSymbols(h, systemTrayDllHooks, ARRAYSIZE(systemTrayDllHooks))) {
+                Wh_Log(L"LoadLibraryExW_Hook: Successfully hooked TrayUI_StartTaskbar");
                 g_taskbarViewDllLoaded = true;
                 Wh_ApplyHookOperations();
+
+                // TrayUI_StartTaskbar was already called, manually initialize
+                HWND hWnd = FindCurrentProcessTaskbarWnd();
+                if (hWnd && !g_playerGrid) {
+                    Wh_Log(L"LoadLibraryExW_Hook: TrayUI_StartTaskbar already called, manually initializing");
+                    g_taskbarWnd = hWnd;
+                    RunFromWindowThread(hWnd, [](void*) {
+                        ApplySettings();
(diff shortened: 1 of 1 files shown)


mods/taskbar-fluent-media-player.wh.cpp
```diff
@@ -0,0 +1,6189 @@
...
+                    }
+                } else {
+                    Wh_Log(L"UpdateVisibility: g_injectionParent is NULL");
+                }
+
+                if (hide) {
+                    g_playerGrid.MinWidth(0);
+                    g_playerGrid.MaxWidth(0);
+                    g_playerGrid.Width(0);
+                } el
```
m417z: This will cause the symbol to first be searched in `Taskbar.View.dll`, even in newer Windows versions where it was moved to `SystemTray.dll`, needlessly slowing it down. One solution is to check the module version, example:

https://github.com/ramensoftware/windhawk-mods/blob/d212cd2fcc4140d1f06f96e590548ddace6d455a/mods/taskbar-vertical.wh.cpp#L4827-L4850

m417z: This part is really fragile, maybe you should find a better hook that fires when the parts you need are ready.

m417z:
I don't think having an overlay as the default position is a good idea:
image

Personally, I think the previous default of Tray - Far edge - Left was a better default option, but it might be a matter of taste.

Also, below is an AI-assisted review. I went over it and made some adjustments, let me know if there are mistakes.

### Submission review

---

**1. The `Int()` settings helper makes many "0" values impossible to set.** In `LoadSettings`:
```cpp
auto Int = [](const wchar_t* key, int lo, int hi, int def) -> int {
    int v = Wh_GetIntSetting(key);
    if (v == 0 && def != 0) v = def;   // <-- silently rejects an explicit 0
    return std::clamp(v, lo, hi);
};
```
`Wh_GetIntSetting` already returns the default declared in the settings block when a value is unset, so the `def` substitution is redundant — and worse, when a user *intentionally* sets `0` it gets snapped back to the default. This breaks every setting where `0` is meaningful and the default is non-zero, e.g. `cornerRadius`, `albumArtCornerRadius`, `buttonCornerRadius` (0 = square corners), `gradientAngle` (0°), `gradientBalance`, and `solidOpacity`/`acrylicTintOpacity`/`micaOpacity`/`blurOpacity`/`pauseOverlayOpacity` (0 = fully transparent). The user simply cannot select those values. Drop the special-casing and just clamp: `return std::clamp(Wh_GetIntSetting(key), lo, hi);` (the `def` argument can be removed entirely).

**2. Media-control buttons dispatch each command twice.** `MakeControlButton` wires up *both* a `Click` handler and a `PointerReleased` handler, and each one calls `SendMediaCommandAsync(cmd)`:
```cpp
btn.Click([cmd](auto const&, auto const&) { ... SendMediaCommandAsync(cmd); DispatchMediaUpdate(); });
...
btn.PointerReleased([...](auto const& sender, ... e) {
    ...
    if (kind == ...LeftButtonReleased) { SendMediaCommandAsync(cmd); DispatchMediaUpdate(); }
});
```
`Button.Click` is raised by `ButtonBase` on pointer release (its class handler runs before your instance `PointerReleased` handler), so a single left-click on e.g. Next/Previous appears to fire the command twice — skipping two tracks, or toggling play/pause back to where it started. At best the two paths are redundant. Keep exactly one dispatch path (either the `Click` handler, or the manual `PointerReleased` one you need for the custom press visuals) and remove the other.

**3. "Open media app" can focus an unrelated window in another process.** In `ExecuteMediaAction("open_app")`, the code `EnumWindows` over *all* top-level windows on the desktop and activates the first one whose title merely *contains* the current track title (case-insensitive substring), only falling back to the AUMID launch afterwards:
```cpp
auto it = std::search(wTitle.begin(), wTitle.end(),
                      search->targetTitle.begin(), search->targetTitle.end(), ...);
if (it != wTitle.end()) { search->foundHwnd = hwnd; return FALSE; }
...
SetForegroundWindow(search.foundHwnd);
```
A track title like "Yesterday" or "Home" will match a browser tab, a document, a chat window, etc. in a completely different app, and the mod will bring that foreign window to the foreground. The reliable identifier you already have is the AUMID — launch `shell:AppsFolder\<aumid>` first (it activates the existing instance for packaged apps and is what the media session actually points to), and only fall back to fuzzy title matching if that fails.

**4. The idle thread posts an undocumented `WM_APP+44` to a window it doesn't own.** `IdleThreadProc` does `PostMessageW(g_taskbarWnd, WM_APP + 44, 0, 0)` once per second to `Shell_TrayWnd`. The mod never subclasses that window, so nothing handles the message — the actual idle counting happens in `UpdateVisibility` on the timer thread, making this thread dead weight. Posting an app-defined `WM_APP+N` message into a window owned by the shell is also risky: that value can collide with whatever the taskbar (or another mod) uses internally. Please remove the idle thread/message entirely (the idle-hide feature keeps working through the timer thread).

**5. Screen-DC leak in `FetchAppIconBytes`.** (line ~2496)

```cpp
HBITMAP hBmp = CreateCompatibleBitmap(GetDC(nullptr), w, h);
```

`GetDC(nullptr)` returns a screen DC that must be released with `ReleaseDC`, but it's passed inline and never freed — every call leaks one DC. This runs each time the app-icon overlay is (re)built while `showAppIcon` is on, so over a session it slowly drains explorer's GDI handles. Fix:

```cpp
HDC screenDC = GetDC(nullptr);
HBITMAP hBmp = CreateCompatibleBitmap(screenDC, w, h);
ReleaseDC(nullptr, screenDC);
```

**6. Misleading symbol-hook comment.** The `// Taskbar.View.dll, SystemTray.dll, ExplorerExtensions.dll` comment directly above the `hooks[]` array in `HookTaskbarDllSymbols` is wrong — those `CTaskBand`/`TaskbarHost`/`_Ref_count_base` symbols are resolved against `taskbar.dll` / `twinui.pcshell.dll` (the module you `LoadLibraryExW` there). Update it to match. Also remove `twinui.pcshell.dll` since it's not relevant to these symbols.

**7. Cross-thread COM on the audio enumerator.** `g_pDeviceEnumerator` is created via `CoCreateInstance` on whatever thread first hits a volume action (typically the XAML UI thread), but the `app_sound` path calls into it from a detached `std::thread` that never `CoInitialize`d (lines ~3844, ~4320), and `CleanupAudioDeviceEnumerator` releases it from `Wh_ModUninit`'s arbitrary thread. The app-volume feature may silently fail and the release is on the wrong apartment. Consider creating/using/releasing it on one consistent apartment.

**8. Process-shutdown destructors.** Several globals hold strong WinRT/XAML refs with non-trivial destructors (`g_playerGrid`, `g_injectionParent`, `g_trackedElement`, the brushes, `g_sessionMgr`, `g_currentSession`). On a normal unload your `Wh_ModUninit` tears everything down cleanly, but on host termination (Explorer restart / sign-out) `Wh_ModUninit` is **not** called and the CRT runs those destructors during `DLL_PROCESS_DETACH` — releasing WinRT/XAML objects under the loader lock, off their owning thread, while the runtime may already be tearing down. Worker threads are `HANDLE`s (trivial dtors, good — no `std::thread::~thread` `std::terminate` trap), but the WinRT strong refs are the risky part. If you see instability on Explorer restarts, consider holding these as intentionally-leaked heap pointers and keeping the real teardown only in `Wh_ModUninit`.

Remove dead code:

- **`HoverMode` dead branches** — it parses `"true"/"false"/"on"/"1"` etc., but the dropdown `$options` only ever yields `auto`/`black`/`white`/`off`, so those branches can't be reached.
- **Dead theme-watcher.** `ThemeWatchThreadProc` registers the `Windhawk_FMP_ThemeWatch` window class and spins a message-only window whose `WndProc` only calls `DefWindowProc` — it never observes anything (message-only windows don't receive `WM_SETTINGCHANGE`/`WM_THEMECHANGED` broadcasts). Theme changes are already detected via `RegNotifyChangeKeyValue` in `TimerThreadProc`. The whole thread/window/class can be removed.
- **Dead `g_themeRefreshPending`** — only ever set to `false` (three places), never read.
- **Dead `controlsMargin` setting.** `LoadSettings` reads `AppearanceSettings.LayoutSettings.controlsMargin`, but there is no `LayoutSettings` group in the settings block, and `controlsMarginLeft/Right` are never used afterwards. Remove the read and the fields.

<p>

- **Detached worker threads aren't tracked.** Numerous `std::thread(...).detach()` calls (`SendMediaCommandAsync`, `FetchMediaPropertiesAsync`, `FetchPlaybackInfoAsync`, the `stop`/`open_app`/`app_sound` paths) run WinRT `.get()` calls that can take a while. On unload/settings-reload you set `g_unloading` and join the *named* threads, but a detached thread that is mid-flight is never waited on — if it resumes after the DLL is unmapped it crashes Explorer. The `if (g_unloading) return;` checks narrow but don't close the window. Since this only races on unload/reload it's a deliberate-compromise item, but consider funneling this work through a single tracked worker (or at least a counter you wait on in `Wh_ModUninit`).
- **Unused dependencies.** `#include <tlhelp32.h>` (no Toolhelp APIs are used) and `-lversion` in `@compilerOptions` (no version-info APIs are used) can be removed.
- **Opt-in success notification is modal.** `ShowSuccessNotification` calls `MessageBoxW` on the taskbar UI thread (reached via `SendMessage` from a worker), which freezes the taskbar and the calling thread until the user clicks OK. It's off by default, but a non-blocking toast (or just dropping it) would be friendlier.
- **`Wh_ModSettingsChanged`** spawns an extra `CreateThread` just to call `RunFromWindowThread` (already cross-thread-safe) and then waits on it — the helper can be called directly.
- **Redundant defensive code from `Wh_GetStringSetting`.** It never returns `NULL` (it returns `L""` on unset/error), so the `if (!objectStr || …)` NULL checks and the `p ? p : def` in the `Str` lambda are dead — terminate the indexed-array loops on the empty string only.

</p>
</details>

<p>

- **`ChangeSystemVolume` adjusts the volume twice per notch.** It both calls `endpointVolume->SetMasterVolumeLevelScalar(±0.005)` *and* posts `WM_APPCOMMAND` with `APPCOMMAND_VOLUME_UP/DOWN` (which itself steps the system volume ~2% and shows the OSD). So one wheel notch moves the volume by an inconsistent combined amount (0.5% + ~2%). If the goal of the `WM_APPCOMMAND` post is just to show the volume flyout, the manual scalar change is double-counting; pick one mechanism (or set the scalar and trigger the OSD without also requesting a step).
- **Palette extraction / blur run on the UI thread.** `ExtractAlbumPalette` and `MakeAlbumBlurBrush`/`UpdateAlbumBlurBgCache` decode the thumbnail and iterate its pixels synchronously inside `RefreshPlayerContents` (taskbar UI thread), including a blocking `BitmapDecoder...().get()`. On each track/art change this can briefly hitch the taskbar. It's hash-cached, but note the palette is recomputed on every art change even when no color setting actually uses the album-art (`-2`) value — gating that work on whether any active color/background option needs it (and/or moving it to the media worker thread) would avoid the cost.
- **`IsFullscreenActive` resolves `shell32.dll` by ordinal 2573** (`SHQueryUserNotificationState`, line 5861). It works, but undocumented ordinals can shift between Windows builds; resolving by name or documenting the assumption would be more robust.

</p>
</details>

mods/taskbar-fluent-media-player.wh.cpp
Salyts: @m417z

m417z: I meant to comment this here (and not in the other comment):

Maybe this will work better instead of hooking `IconView::IconView` which is too early and requires a thread and multiple attempts:
https://github.com/ramensoftware/windhawk-mods/blob/d212cd2fcc4140d1f06f96e590548ddace6d455a/mods/taskbar-tray-show-on-hover.wh.cpp#L486-L494

m417z: Did you tag me because it's ready? Looks like 6, 7, 8 weren't addressed.

mods/taskbar-fluent-media-player.wh.cpp
```diff
@@ -0,0 +1,6647 @@
...
+                } else {
+                    btn.Visibility(Visibility::Visible);
+                }
+                if (auto ct = btn.Content().try_as<TextBlock>()) {
+                    const wchar_t* glyph = GetGlyph(5);
+                    ct.Text(winrt::hstring(glyph));
+                    ct.Opacity(supported ? 1.0 : 0.35);
+                    ct.Foreground(MakeBrush(ButtonColor()));
+                }
+            }
```
m417z, line 6479: Add `return FALSE;`.

m417z: Add `return FALSE;` if it fails.

Salyts: Where?

m417z:
```cpp
BOOL Wh_ModInit() {
    // ...

    HookTaskbarDllSymbols(); // <- here

    if (HMODULE h = GetTaskbarViewModule()) {
        HookTaskbarViewDllSymbols(h); // <- here
```

Salyts:
```cpp
    if (!HookTaskbarDllSymbols()) {
        return FALSE;
    }

    if (HMODULE h = GetTaskbarViewModule()) {
        if (!HookTaskbarViewDllSymbols(h)) {
            return FALSE;
        }
```
I added it, but the media player won't load.

```
21:20:41.435 25316 explorer.exe  [WH] [local@taskbar-fluent-media-player] [3502:MediaThreadProc]: MediaThreadProc: Shutting down
21:20:41.435 25316 explorer.exe  [WH] [local@taskbar-fluent-media-player] [3509:MediaThreadProc]: MediaThreadProc: Media thread ended
```

m417z: Is it related to this change? Switch to detailed logging from the mod preview advanced tab.

Salyts: Maybe we can do it like this?

Salyts: @m417z

m417z: I'm looking at the code, and it's really messy. The most important thing is that `public: virtual void __cdecl TrayUI::StartTaskbar(void)` is part of Taskbar.dll, not Taskbar.View.dll. There are several other mistakes, I can take another look once this is fixed.

You can look at `taskbar-tray-show-on-hover` for reference.

Salyts:
@m417z

m417z:
This version doesn't load when explorer.exe is restarted.

I tried to implement a localized fix without changing too much of the code. I did it with retries waiting for the tray grid to become available. See if it works for you in all cases.

https://gist.github.com/m417z/7f4ec0e74d97f535cfaad196da823809

Salyts:
> This version doesn't load when explorer.exe is restarted.
> 
> I tried to implement a localized fix without changing too much of the code. I did it with retries waiting for the tray grid to become available. See if it works for you in all cases.
> 
> https://gist.github.com/m417z/7f4ec0e74d97f535cfaad196da823809

works

m417z:
I removed some of the code that I think is no longer necessary, e.g. Taskbar.View.dll hooks. Don't you want to adopt these removals?

Salyts:
> I removed some of the code that I think is no longer necessary, e.g. Taskbar.View.dll hooks. Don't you want to adopt these removals?

I didn't see it, wait.

Salyts:
> I removed some of the code that I think is no longer necessary, e.g. Taskbar.View.dll hooks. Don't you want to adopt these removals?

I can't understand what you're talking about... everything seems correct.

m417z:
To see what I'm talking about, compare your file:
https://github.com/Salyts/windhawk-mods/blob/778d8ababb01d46e89502129143882d0126711fa/mods/taskbar-fluent-media-player.wh.cpp
To my file:
https://gist.github.com/m417z/7f4ec0e74d97f535cfaad196da823809

Salyts:
> To see what I'm talking about, compare your file: https://github.com/Salyts/windhawk-mods/blob/778d8ababb01d46e89502129143882d0126711fa/mods/taskbar-fluent-media-player.wh.cpp To my file: https://gist.github.com/m417z/7f4ec0e74d97f535cfaad196da823809

Thank you

PR #4322, opened 2026-06-06
Update Explorer Ctrl+N to New File to v1.1

lieyanbang:
## Changelog

* Support the active tab in tabbed Explorer windows.
* Use Shell file operation fallback for protected folders, allowing Windows to show the standard elevation prompt when needed.
* Improve rename reliability by using the active Shell view first.
* Limit expensive UI Automation scans and cache the path element per active tab when possible.

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/taskbar-music-lounge.wh.cpp
+++ b/mods/taskbar-music-lounge.wh.cpp
@@ -2,7 +2,7 @@
 // @id              taskbar-music-lounge
 // @name            Taskbar Music Lounge
 // @description     A native-style music ticker with media controls.
-// @version         4.0.1
+// @version         4.0.2
 // @author          Hashah2311
 // @github          https://github.com/Hashah2311
 // @include         explorer.exe
@@ -830,8 +830,13 @@ void WhTool_ModUninit() {
 void WhTool_ModSettingsChanged() {
     LoadSettings();
     if (g_hMediaWindow) {
-         SendMessage(g_hMediaWindow, WM_TIMER, IDT_POLL_MEDIA, 0);
-         SendMessage(g_hMediaWindow, WM_SETTINGCHANGE, 0, 0); 
+         PostMessage(g_hMediaWindow, WM_TIMER, IDT_POLL_MEDIA, 0);
+         PostMessage(g_hMediaWindow, WM_SETTINGCHANGE, 0, 0);
+         // Reposition/resize the window to pick up new size and offsets. This
+         // path is normally only triggered by taskbar movement, so without it
+         // changes to Panel Width/Height/Offset wouldn't apply until the mod is
+         // reloaded.
+         PostMessage(g_hMediaWindow, WM_APP + 10, 0, 0);
     }
 }
 


m417z:
### Submission review

---

This is a big, well-structured rewrite, and the core ideas (worker thread for the heavy work, IFileOperation elevation fallback, caching) are sound. A few things below are worth fixing before merge — mainly that the new heavy work now runs on the keyboard-hook thread, and that path resolution leans on UIA text scraping where a robust shell API already exists.

**1. The new-file action runs on the low-level hook's *own* thread, which blocks the hook and degrades system-wide keyboard input.**

`WH_KEYBOARD_LL` is invoked by the system *on the thread that installed it*, and that thread must keep pumping its message loop or the hook can't fire (and after `LowLevelHooksTimeout`, ~300 ms by default, the event is processed without the hook — and repeated timeouts can get the hook silently removed). But `HookThread` both installs the hook *and* runs `PerformNewFileAction()` synchronously off `WM_APP+1` (`explorer-ctrln-newfile.wh.cpp:1594`). While that action runs — UIA tree scans, `IShellWindows` enumeration, and the `StartRename` retry loop that sleeps up to 900 ms (`:1378`) — the hook thread is *not* pumping, so for that window **every keystroke system-wide stalls up to the timeout**, and the Ctrl+N hook can stop working until reload.

This is exactly the "no blocking work tied to a low-level hook" rule — the work needs to be on a thread *separate from the one that owns the hook*. v1.0 had the same structure but a much lighter action; v1.1 makes it materially worse. Recommend: keep `HookThread` doing nothing but installing the hook and pumping, and spawn a dedicated STA worker thread (its own `CoInitializeEx` + message loop) that receives the `WM_APP+1` request and runs `PerformNewFileAction`. (Your `g_actionRunning` interlock then actually becomes meaningful — today it can never contend since everything is on one thread.)

**2. Resolve the active tab's folder via the shell API, not by scraping the UIA address bar.**

The whole `UIA_GetCurrentTabPath` path (parsing the address-bar element's Name/Value, `ExtractPathCandidate`, `LooksLikePathText`, the per-language prefix stripping at `:234`) is a fragile way to get something the shell hands you directly:

- It only strips `Address:` / `地址:` / `位置:` prefixes, so other UI languages aren't covered; and on the many configurations where the address bar shows a *friendly* breadcrumb ("Documents", "This PC > …") rather than a real path, `LooksLikePathText` rejects it and resolution silently fails — the feature just doesn't create a file.
- It's also expensive (full UIA `FindAllBuildCache` scans), which is what's blocking the hook thread in item 1.

You already have the robust pieces — `GetActiveShellTabHwnd()` (`:511`) finds the active `ShellTabWindowClass`, and `GetPathFromShellView()` (`:976`) reads the real folder via `IFolderView`→`IPersistFolder2::GetCurFolder`. The clean flow is: enumerate `IShellWindows`, for each `QueryActiveShellView`, call `IShellView::GetWindow()` to get its `SHELLDLL_DefView`, walk up to its `ShellTabWindowClass` ancestor, and compare against the active-tab HWND from `GetActiveShellTabHwnd()`. On match, `GetPathFromShellView` gives you the path with no text parsing, no localization issues, and correct handling of special folders. Reference implementations of the shell chain:
- [explorer-tabs-session-saver.wh.cpp#L76-L87](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/explorer-tabs-session-saver.wh.cpp#L76)
- [explorer-single-window-tabs.wh.cpp#L253-L294](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/explorer-single-window-tabs.wh.cpp#L253)

This would also let you drop most of the UIA code and the address-bar caching layer entirely.

**3. Consider converting this to a "mods as tools" mod instead of injecting into `explorer.exe`.**

This mod installs *no function hooks* — it only uses `SetWindowsHookEx(WH_KEYBOARD_LL)`, `EnumWindows`, `IShellWindows`, and UIA, all of which work cross-process from any host. The only thing tying it to `explorer.exe` is the `IsCurrentProcessWindow` guard sprinkled through `GetTargetExplorerHWND`/`IsExplorerForegroundForHotkey` — an artifact of running in-process, not a requirement. Running inside `explorer.exe` means (a) a bug here can destabilize the shell, and (b) with multiple `explorer.exe` instances you install *one global LL keyboard hook per instance*, each adding latency to every keystroke (the per-process foreground check dedups the *action*, not the hook overhead).

This is a strong fit for the [tool-mod pattern](https://github.com/ramensoftware/windhawk/wiki/Mods-as-tools:-Running-mods-in-a-dedicated-process): `@include windhawk.exe`, rename `Wh_ModInit`/`Wh_ModUninit` → `WhTool_*`, paste the launcher snippet, and replace the same-process checks with a plain "is the foreground window an Explorer top-level (or the desktop) window" test (works cross-process). [keep-rainmeter-always-bottom.wh.cpp](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/keep-rainmeter-always-bottom.wh.cpp) is a good reference for a tool mod that manipulates shell windows cross-process. Given how much code it removes and the stability win, it's worth doing as part of this rewrite.

<p>

- **Drop the `[Ctrln]` log prefix.** Windhawk already prefixes log lines with the mod name and gates logging behind the enable/disable toggle, so `[Ctrln]` is redundant in every `Wh_Log` call.
- **README wording.** The English README says the mod "creates an empty `file_new` file" (`:22`), but the code names it `file` / `file (2)` / … and only falls back to `file_new` after 10000 collisions (`MakeUniqueFilename`, `:309`). Update the text to match.
- **`g_tabPathCache` accumulates stale entries.** Closed tabs/windows are never pruned — `UIA_TryTabPathCache` returns early on `!IsWindow(tabHwnd)` *without* releasing the cached `IUIAutomationElement` (`:768`), so those elements stay AddRef'd until thread exit and the vector only grows. Prune entries whose `tabHwnd` is no longer a window (or cap the cache). (Moot if you adopt item 2 and remove the UIA cache.)
- **`#include <ShlGuid.h>` was removed but `SID_STopLevelBrowser` is still used** (`:1062`) — it currently compiles via a transitive include; better to include the header for the symbol you use.

</p>
</details>

<p>

- **Rename can target the wrong tab/window.** `StartRename` resolves the view by *path* (`GetShellViewForFolderPath`, `:1021`), picking the first `IShellWindows` entry whose folder matches. If two tabs or windows show the same folder, the `SVSI_EDIT` rename box can open in a background one. Matching the active tab specifically (item 2's HWND comparison) avoids this.
- **The F2-injection fallback is order-dependent.** `ForceForeground` + `SendInput(VK_F2)` (`:1402`) assumes focus lands on the just-created item; if the foreground changed in between, F2 goes elsewhere. It's a reasonable last resort behind the `IShellView::SelectItem(SVSI_EDIT)` primary path, just noting the fragility.

</p>
</details>

PR #4329, opened 2026-06-06
 Architecture Overhaul and Native Rendering

drgutman:
## Changelog

## 📋 Changelog Summary

**NEW FEATURE**
*   Double-click on empty desktop background to hide or show the overlay

**Added:**
*   Native WIC image decoding (`-lwindowscodecs`).
*   COM integration (`IDesktopWallpaper`) for accurate Fit/Position enums.
*   Registry Watcher thread for instant theme/wallpaper detection.
*   Native Window Subclassing for desktop icon interaction.
*   Support for Solid Colour backgrounds and Centre/Fit letterboxing.

**Changed:**
*   Refactored thread safety: Split `g_stateMutex` (UI) and `g_renderMutex` (Render).
*   Replaced GDI `CaptureWallpaperBitmap` with WIC `LoadWallpaperBitmap`.
*   Optimized render loop sleep states for high-refresh-rate (144Hz+) monitors.
*   Replaced global DispatchMessageW hook with a thread-local WH_GETMESSAGE bootstrap
*   Replaced hardcoded WM_APP macros with dynamic RegisterWindowMessageW routing
*   Replaced per-frame WindowFromPoint Z-order traversal with native TrackMouseEvent (TME_LEAVE) on desktop subclasses.
*   Replaced per-frame WindowFromPoint Z-order traversal with native TrackMouseEvent (TME_LEAVE) on desktop subclasses. This eliminates hundreds of Win32 global lock acquisitions per second, dropping idle CPU usage to absolute zero and preventing micro-stutters on high-refresh-rate monitors.
*   Added bounds checking and path validation to TranscodedImageCache parsing to prevent std::bad_alloc crashes from corrupted registry data, and added ERROR_SUCCESS validation to the Registry Watcher thread to prevent undefined behaviour from null handles.

**Fixed:**
*   Wallpaper "Fill" mode misalignment across all DPI scales.
*   Spotlight/Selection masks bleeding through the wallpaper overlay.
*   Overlay turning solid black/colour after DPI scaling changes.
*   Micro-stutters caused by UI thread mutex contention during D2D rendering.
*   Icon selection not moving from keyboard input and clean-up code

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini
- - [x] Another AI (please specify): Qwen3.7Max


m417z:
Thanks for the update. Below is the Claude review. Nothing serious. `RemoveDesktopSubclasses` is documented by Microsoft to not work across threads, but looks like it works, so there's no actual crash on unload.

### Submission review

---

- **Cross-thread subclass removal on unload -> dangling subclass proc / crash.** The `SysListView32` and `SHELLDLL_DefView` subclasses are installed (correctly) on the desktop UI thread, but `Wh_ModUninit` (which runs on an arbitrary Windhawk thread) calls `RemoveDesktopSubclasses()` (line 1883), which calls raw `RemoveWindowSubclass` (lines 684/688) from that thread -- not the one that owns those windows. Raw `RemoveWindowSubclass` isn't reliable across threads (this is exactly why Windhawk provides the helpers), so the subclass can be left installed with `ListViewSubclassProc`/`ShellViewSubclassProc` pointing into the about-to-be-unloaded mod image -> crash on the next desktop mouse-move/notification. The `if (!g_unloading)` guard inside the procs doesn't help once the DLL is gone. Switch to `WindhawkUtils::SetWindowSubclassFromAnyThread` / `RemoveWindowSubclassFromAnyThread` (they marshal via `SendMessage` to the owning thread) for both `ApplyDesktopSubclasses` and `RemoveDesktopSubclasses` -- see [window-opacity.wh.cpp#L424](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/window-opacity.wh.cpp#L424) and [#L534](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/window-opacity.wh.cpp#L534), and the [Development tips](https://github.com/ramensoftware/windhawk/wiki/Development-tips). (`RemoveDesktopSubclasses` in `Wh_ModUninit` holds no lock, so the `SendMessage`-based helper won't deadlock there.)

<p>

Minor polish -- none of this affects users, so it's your call.

- `WM_APP_INIT` is registered in `Wh_ModInit` (line 1846) but never used anywhere (leftover from the old `DispatchMessageW` bootstrap). Remove the variable and its `RegisterWindowMessageW` call.
- Dead debug scaffolding: the commented-out `DebugLog(...)` block (lines 117-132) and the scattered `//Wh_Log(...)` lines can be dropped -- `Wh_Log` is the right tool and is already cheap/disabled in production, so no need to keep a commented console logger around.
- `RefreshWallpaperAndStyle` calls `CoInitializeEx` twice (the outer one at the top and a nested one around the `IDesktopWallpaper` call). It already runs on Explorer's STA UI thread, so the inner init/uninit pair is redundant -- the outer one suffices.
- `@compilerOptions` / includes look like they carry leftovers from the GDI-capture era: no GDI calls remain (so `-lgdi32`), no direct `shell32` exports are used (`-lshell32`), and `<knownfolders.h>` (no `FOLDERID_*` usage) / `<shlobj.h>` appear unused now that paths come straight from the registry. Trim whatever the toolchain confirms is unneeded.
- `g_settings` is read without the lock in the subclass procs (e.g. `g_settings.toggleOnDoubleClick` at line 734) while `LoadSettings` rewrites the whole struct under `g_renderMutex` on another thread -- a technically-UB data race, benign in practice for a single `bool`. Snapshot the value, read it under the lock, or make that one field `std::atomic`.
- `RegisterOverlayWindowClass`/`RegisterMessageWindowClass` return success on `ERROR_CLASS_ALREADY_EXISTS` (lines 1585/1696). You do `UnregisterClass` in `Wh_ModUninit` (good) and it's needed for the recreate flow, so this is low-risk -- just noting that if a prior instance's cleanup is ever skipped, `CreateWindowEx` would reuse a class whose `lpfnWndProc` points into the old image.

</p>
</details>

<p>

- The "Fill" mode `/3` centring divisor is a reverse-engineered Windows Shell quirk. It may match today's shell across DPI scales as you describe, but it's the kind of undocumented behavior that can shift between Windows builds, so it's worth re-checking after major OS updates.

</p>
</details>

drgutman:
@m417z - I think I'll leave those for the next version 🤷‍♂️
I'll look at them one of these days.
Thank you very much 😊

mods/transparent-desktop-icons-spotlight.wh.cpp
```diff
@@ -150,240 +150,647 @@ struct Settings {
...
         }
     }
 
-    if (mouseMovedPoll || selectionChanged || isEditing != s_wasEditing || g_forceRender) {
-        if (g_renderEvent) SetEvent(g_renderEvent);
+    g_forceRender = true;
+    if (g_renderEvent) SetEvent(g_renderEvent);
+
+    // Clean up COM if we initialized it
+    if (hrCom == S_OK) CoUninitialize();
```
m417z: Why was it changed from `if (SUCCEEDED(hrCom)) CoUninitialize();` to this?

drgutman: "The Explorer UI thread is already initialised as STA. Calling CoInitializeEx on it returns S_FALSE (0x00000001), not S_OK. The SUCCEEDED() macro treats S_FALSE as success, so CoUninitialize() would still run — decrementing the thread's COM reference count. If it hit zero, it would tear down COM for the entire Explorer UI thread, silently breaking drag-and-drop, shell extensions, and context menus until Explorer restarts.

Switching to if (hrCom == S_OK) ensures CoUninitialize() only runs when your call actually initialised COM, not when it was already initialised by Explorer."

Hehe, had to go back to the logs to find out, luckily Deepseek flash was there to find it fast :)

m417z: > To uninitialize the COM library gracefully on a thread, each successful call to [CoInitialize](https://learn.microsoft.com/en-us/windows/desktop/api/objbase/nf-objbase-coinitialize) or CoInitializeEx, including any call that returns S_FALSE, must be balanced by a corresponding call to [CoUninitialize](https://learn.microsoft.com/en-us/windows/desktop/api/combaseapi/nf-combaseapi-couninitialize).

https://learn.microsoft.com/en-us/windows/win32/api/combaseapi/nf-combaseapi-coinitializeex

drgutman: Done. Thank you 🤭

PR #4341, opened 2026-06-08
submitting desktop-icons-toggle mod

ChamPoingYT:
## Changelog

Initial release.

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/desktop-icons-toggle.wh.cpp
+++ b/mods/desktop-icons-toggle.wh.cpp
@@ -2,18 +2,48 @@
 // @id              desktop-icons-toggle
 // @name            Desktop Icons Toggle
 // @description     Taskbar button to toggle desktop icons visibility
-// @version         7.5
+// @version         1.0.0
 // @author          ChamPoing
+// @github          https://github.com/ChamPoingYT
 // @include         explorer.exe
 // @compilerOptions -lgdi32 -ldwmapi
 // ==/WindhawkMod==
 
 // ==WindhawkModReadme==
-/*...*/
+/*
+
+# Desktop Icons Toggle
+Adds a persistent button to your taskbar to show/hide desktop icons with a single click.
+
+If you right click on the icon : 
+
+- show a small menu with all the actions you can do.
+
+If you double-left click on the icon :
+
+- open your file explorer.
+
+## Settings
+- **X Offset / Y Offset**: Fine-tune position in pixels.
+
+## Important
+I dont know if i will provide update for this mod, if u want, or if you want to send feedback/somethgin else, 
+add chaisegamings on discord.
+
+(i accept everyone)
+
+*/
 // ==/WindhawkModReadme==
 
 // ==WindhawkModSettings==
-/*...*/
+/*
+- offsetX: 0
+  $name: Horizontal offset (pixels)
+  $description: Move the button horizontally
+- offsetY: 0
+  $name: Vertical offset (pixels)
+  $description: Move the button vertically
+*/
 // ==/WindhawkModSettings==
 
 #include <windows.h>
--- a/mods/desktop-icons-toggle.wh.cpp
+++ b/mods/desktop-icons-toggle.wh.cpp
@@ -1,0 +1,420 @@
+// ==WindhawkMod==
+// @id              desktop-icons-toggle
+// @name            Desktop Icons Toggle
+// @description     Taskbar button to toggle desktop icons visibility
+// @version         7.5
+// @author          ChamPoing
+// @include         explorer.exe
+// @compilerOptions -lgdi32 -ldwmapi
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*...*/
+// ==/WindhawkModReadme==
+
+// ==WindhawkModSettings==
+/*...*/
+// ==/WindhawkModSettings==
+
+#include <windows.h>
+#include <shellapi.h>
+#include <dwmapi.h>
+#include <strsafe.h>
+#include <thread>
+
+enum ZBID { ZBID_DEFAULT = 0, ZBID_IMMERSIVE_NOTIFICATION = 4 };
+typedef HWND(WINAPI* pCreateWindowInBand)(DWORD, LPCWSTR, LPCWSTR, DWORD, int, int, int, int, HWND, HMENU, HINSTANCE, LPVOID, DWORD);
+
+#define TOGGLE_MSG    0x7402
+#define BTN_W         38
+#define BTN_H         38
+#define WM_REPOS      (WM_APP + 1)
+#define WM_CLOSE      (WM_APP + 2)
+#define WM_RELOAD     (WM_APP + 3)
+#define HOTKEY_ID     100
+
+#define DOT_H         3   
+#define DOT_Y_OFFSET  3   
+#define DOT_R         2   
+
+static HWND          g_hwndBtn           = nullptr;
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

---

**This should be a "tool mod", not an `explorer.exe` injection.** Because there are no hooks, nothing here needs to run inside `explorer.exe`'s address space — and injecting into the shell carries two concrete downsides:
- **Duplicate buttons.** There is no single-instance guard, so every `explorer.exe` instance creates its own button + timer + hotkey. With *Launch folder windows in a separate process* enabled (or any transient extra `explorer.exe`), you get multiple overlapping buttons at the same taskbar position.
- A crash in this code takes down the shell, even though the work is conceptually a standalone widget.

The fix is the [Mods as tools pattern](https://github.com/ramensoftware/windhawk/wiki/Mods-as-tools:-Running-mods-in-a-dedicated-process): `@include windhawk.exe`, rename `Wh_ModInit`/`Wh_ModSettingsChanged`/`Wh_ModUninit` → `WhTool_*`, and paste the launcher snippet from the wiki. The framework then runs a single dedicated process for you (no home-grown single-instance code needed). See [mods/keep-rainmeter-always-bottom.wh.cpp](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/keep-rainmeter-always-bottom.wh.cpp) for a tool mod that uses `SetWinEventHook` + cross-process window manipulation. (`CreateWindowInBand` z-band placement is system-wide, so it works from a dedicated process too.) As a bonus, this also fixes the next item.

Note that using `CreateWindowInBand` with `ZBID_IMMERSIVE_NOTIFICATION` might only work with `explorer.exe`. If that's the case, you can still use the tool mod pattern, but keep `@include explorer.exe` to have the mod running in a separate explorer.exe process.

**`WINEVENT_SKIPOWNPROCESS` defeats the taskbar hook in the shell process.** `RegisterTaskbarHook` scopes the hook to the taskbar's pid/tid and passes `WINEVENT_SKIPOWNPROCESS` (line 269). In the main shell `explorer.exe`, the taskbar *is* the own process, so every `EVENT_OBJECT_LOCATIONCHANGE` is filtered out and `TaskbarEventProc` never fires — the button only ever repositions via the 500 ms polling timer, making the whole `SetWinEventHook` path dead code. Drop `WINEVENT_SKIPOWNPROCESS` (or move to a tool mod, where the taskbar is a different process and the flag becomes harmless).

**Don't redefine the standard `WM_CLOSE` message macro.** Line 62 does `#define WM_CLOSE (WM_APP + 2)` *after* `<windows.h>` has already defined `WM_CLOSE` as `0x0010`. This is a macro redefinition (it warns under Clang) and a genuine footgun — the real system `WM_CLOSE` now silently falls through to `DefWindowProc`, and anyone reading the code will be misled. Rename your private constant, e.g. `WM_DOCLOSE` / `WH_WM_CLOSE`, and leave `WM_CLOSE` alone. (`WM_REPOS`/`WM_RELOAD` are fine since those names aren't taken.)

**Hard-coded positioning breaks on common configurations.** The button is placed at a fixed `barW - BTN_W - 220` from the right "before the clock" (line 248), with `BTN_W = BTN_H = 38` and font height `20` all hard-coded. None of this is DPI-scaled, so on any non-100 % display the button is the wrong size and the `-220` offset lands in the wrong place; the offset also depends on locale/clock format and the Windows version (Win10 vs Win11 systray widths differ). Scale sizes/offsets by the taskbar's DPI (`GetDpiForWindow`/`MonitorFromWindow` + `GetDpiForMonitor`), and consider deriving the position from an actual tray element rather than a magic constant. Also note only the primary `Shell_TrayWnd` is handled — secondary-monitor taskbars get no button.

**The `Vertical offset` setting doesn't move the button.** `RepositionButton` only applies `g_offsetX` to the window's X (line 249); `g_offsetY` is never applied to the window position. Instead `g_offsetY` is folded into the *text rectangle* inside `UpdateButtonLayered` (line 142), so it shifts the emoji within the 38 px button and can push it out of view. That contradicts the setting's `$description` ("Move the button vertically"). Apply `g_offsetY` to `y` in `RepositionButton` and remove it from the text rect.

**Undocumented, hard-coded global hotkey.** `WM_CREATE` registers a system-wide `Ctrl+Alt+D` hotkey (line 281) that isn't mentioned anywhere in the README and can't be changed. It can collide with other apps' hotkeys, and with multiple `explorer.exe` instances only the first registration succeeds. Please at least document it; ideally make it configurable (or removable). A tool mod also avoids the multi-registration problem.

**README.** Has typos ("dont", "somethgin", "u") and is quite informal; since this is a visual mod, please add a screenshot/GIF of the button (only `i.imgur.com` / `raw.githubusercontent.com` are allowed image hosts).

**Overlap with an existing mod.** [zen-desktop-toggle-icons](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/zen-desktop-toggle-icons.wh.cpp) already toggles desktop icons via the same `0x7402` `WM_COMMAND` mechanism. Your UX is different (a taskbar button + menu + hotkey vs. double-clicking the desktop), so this isn't a strict duplicate, but noting for completeness. Worth calling out the distinction in the description.

<p>

- **Unused dependency.** `#include <dwmapi.h>` and `-ldwmapi` (lines 9, 51) — there are no `Dwm*` calls anywhere. Remove both.
- **Stale version strings.** The About box says `v7.5` and the window class is `WhDeskToggle75` (lines 353, 399) while `@version` is `1.0.0`. Make them consistent.
- **`RegisterClassExW` return is unchecked, and the failure path leaks the class.** If `CreateWindow*` fails, `BtnThread` returns at line 415 *before* the `UnregisterClassW` at line 426, leaving the class registered with an `lpfnWndProc` pointing into the (soon-to-be-unloaded) mod — the next load can then hit `ERROR_CLASS_ALREADY_EXISTS` / a dangling proc. Register with the mod's own module handle and ensure the class is unregistered on every exit path.
- **`g_offsetX`/`g_offsetY` data race.** `LoadSettings` writes them from `Wh_ModInit`/`Wh_ModSettingsChanged` (arbitrary thread) while the button thread reads them. It's benign for ints, but you already post `WM_RELOAD` — you could just read the settings on the button thread.
- The 500 ms timer calls `RepositionButton` → `UpdateButtonLayered`, fully re-rendering the layered window (DIB section, font, alpha pass) every tick even when nothing changed. Skip the redraw when position/state are unchanged.

</p>
</details>

<p>

- **`g_iconsVisible` can go stale.** It's initialized to `true` and only flipped by the mod's own toggles. If icons start hidden, or the user toggles them via the desktop's right-click *View → Show desktop icons* menu, the button's state (and the menu label) will be wrong. You can derive the real state from the visibility of the `SHELLDLL_DefView` listview.
- **Double-click both toggles and opens Explorer.** The first `WM_LBUTTONUP` toggles immediately; a second within 300 ms then launches Explorer — so a double-click toggles once *and* opens a window. Consider deferring the single-click action until the double-click window passes.
- The button is drawn over the taskbar at a fixed position and doesn't reserve space, so it can overlap existing tray icons/widgets depending on what's there. This is inherent to the overlay approach.
- `SetForegroundWindow(hwnd)` is called before `TrackPopupMenu`, but the button is `WS_EX_NOACTIVATE`, so it may not actually become foreground — the right-click menu may not always dismiss cleanly on an outside click.

</p>
</details>

PR #4346, opened 2026-06-09
Add Minimalist System Performance HUD mod

agarwaladvit96-coder:
This mod provides a minimalist desktop overlay to monitor system performance metrics such as CPU, RAM, GPU, Disk Activity, and Network Speed. It supports custom Google Fonts and offers various display options.

## Changelog

* Changelog item 1...
* Changelog item 2...

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/desktop-minimalist-performance-hud.wh.cpp
+++ b/mods/desktop-minimalist-performance-hud.wh.cpp
@@ -4,8 +4,8 @@
 // @description  A premium desktop hardware monitor overlay. Supports all local and Google Fonts with optimized ClearType rendering.
 // @version      1.0.0
 // @author       Random Author
-// @github       https://github.com/agarwaladvit96-coder/
-// @homepage     https://github.com/agarwaladvit96-coder/
+// @github       https://github.com/agarwaladvit96-coder
+// @homepage     https://github.com/agarwaladvit96-coder
 // @include      explorer.exe
 // @compilerOptions -lpdh -lgdi32
 // ==/WindhawkMod==
--- a/mods/desktop-minimalist-performance-hud.wh.cpp
+++ b/mods/desktop-minimalist-performance-hud.wh.cpp
@@ -1,0 +1,382 @@
+// ==WindhawkMod==
+// @id           desktop-minimalist-performance-hud
+// @name         Minimalist System Performance HUD
+// @description  A premium desktop hardware monitor overlay. Supports all local and Google Fonts with optimized ClearType rendering.
+// @version      1.0.0
+// @author       Random Author
+// @github       https://github.com/agarwaladvit96-coder/
+// @homepage     https://github.com/agarwaladvit96-coder/
+// @include      explorer.exe
+// @compilerOptions -lpdh -lgdi32
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Minimalist System Performance HUD
+A sleek, elegant desktop overlay displaying real-time hardware metrics. 
+
+### 🌐 Using Google Fonts
+This mod is fully optimized to display any font from the Google Fonts library! To use your favorite web font:
+1. Go to [fonts.google.com](https://fonts.google.com).
+2. Download your favorite font family (e.g., *Inter*, *Roboto*, *Poppins*, or *Montserrat*).
+3. Open the downloaded `.ttf` or `.otf` file and click **Install**.
+4. Open this mod's settings, set the Preset to **Custom**, and type the exact name into the **Custom Font Name Override** box.
+*/
+// ==/WindhawkModReadme==
+
+// ==WindhawkModSettings==
+/*
+- FontPreset: "modern"
+  $name: Font Style Preset
+  $description: Choose a curated design style, or select "Custom" to use a downloaded Google Font.
+  $options:
+    - modern: "Windows 11 Fluent (Segoe UI)"
+    - tech: "Terminal Hacker (Consolas)"
+    - retro: "Classic Bit-Style (Courier New)"
+    - custom: "Custom Google Font / System Font (Type below)"
+- CustomFontName: "Inter"
+  $name: Custom Font Name Override
+  $description: Type the exact name of any installed Google Font or system font (e.g., Roboto, Poppins, Inter, Arial Black).
+- ShowCpu: true
+  $name: Show CPU
+  $description: Toggle visibility of the CPU tracking meter.
+- ShowRam: true
+  $name: Show RAM
+  $description: Toggle visibility of the RAM tracking meter.
+- ShowGpu: true
+  $name: Show GPU
+  $description: Toggle visibility of the GPU tracking meter.
+- ShowStorage: true
+  $name: Show Storage
+  $description: Toggle visibility of the storage capacity metric.
+- ShowDisk: true
+  $name: Show Disk Activity
+  $description: Toggle visibility of the hard drive active response metric.
+- ShowNetworkDown: true
+  $name: Show Download Speed
+  $description: Toggle visibility of network download traffic.
+- ShowNetworkUp: true
+  $name: Show Upload Speed
+  $description: Toggle visibility of network upload traffic.
+*/
+// ==/WindhawkModSettings==
+
+#include <windows.h>
+#include <string>
+#include <pdh.h>
+#include <pdhmsg.h>
+
+// Global Window and Timer variables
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

Please address the following issue.

---

**Substantial overlap with [desktop-live-overlay](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/desktop-live-overlay.wh.cpp).** That mod already does exactly this category: *"Display live, customizable content on the desktop behind icons. Perfect for showing time, date, **system metrics**, weather, and more."* The maintainer's strong preference is to extend an existing mod rather than merge a near-duplicate. This submission is essentially a narrower, fixed-layout subset of that mod. Please describe how it differs in a way that justifies a separate mod, or consider contributing the missing metrics (GPU / disk activity / network speed bars) to desktop-live-overlay instead. As written, the overlap is hard to justify.

agarwaladvit96-coder:
So, after reviewing both the mods, my mod is only meant for system usage metrics, while desktop-live-overlay is meant to show time battery percentage and many more, my mod is meant for the people who want to monitor their system usage without opening task manager

agarwaladvit96-coder:
and also my mod is meant to be used with an animated wallpaper using lively wallpaper unlike desktop-live-overlay which struggles with 3rd-party animated wallpapers

m417z:
OK, please address the preliminary review below.

### Submission review

---

**1. The teardown crashes Explorer on every disable / update / reload.** `Wh_ModInit` spawns `DashboardThreadProc`, which registers a class, creates the overlay window, and runs its own `GetMessage` loop on that thread. `Wh_ModUninit` then calls `KillTimer`/`DestroyWindow` from an *arbitrary* thread (not the one that owns the window):

```cpp
void Wh_ModUninit() { if (g_timerId && g_hWndOverlay) KillTimer(g_hWndOverlay, g_timerId); if (g_hWndOverlay) DestroyWindow(g_hWndOverlay); if (g_hQuery) PdhCloseQuery(g_hQuery); }
```

`DestroyWindow` cannot destroy a window created by another thread, so it's a silent no-op. Nothing ever posts `WM_QUIT`, so the message loop never exits and the thread is never joined. When `Wh_ModUninit` returns, Windhawk unloads the mod DLL while that thread is still executing `GetMessage`/`DashboardWndProc` inside the now-unmapped image → access violation in `explorer.exe`. This is a guaranteed crash on disable, and the mod also fails to actually remove the overlay (it stays until Explorer restarts).

Fix: do the windowing teardown on the owning thread and join it before returning. Signal the thread to quit (e.g. `PostMessage(g_hWndOverlay, WM_CLOSE, 0, 0)` with a `WM_DESTROY` → `PostQuitMessage(0)` handler, or `PostThreadMessage(threadId, WM_QUIT, 0, 0)`), keep the thread handle instead of `CloseHandle`-ing it immediately, and `WaitForSingleObject(hThread, INFINITE)` in `Wh_ModUninit` before unloading. desktop-live-overlay destroys its windows from the owning thread via a posted message: [Wh_ModUninit](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/desktop-live-overlay.wh.cpp#L2958).

**2. The window class is registered under the wrong module and never unregistered.** The class uses `wc.hInstance = GetModuleHandle(NULL)` (Explorer's handle, not the mod's) and there is no `UnregisterClass`. A class registered with `RegisterClass` is **not** auto-removed when the mod DLL unloads, so its `lpfnWndProc` is left dangling into the unmapped (or relocated) mod image. On the next load `RegisterClass` fails with `ERROR_CLASS_ALREADY_EXISTS` (the return value is ignored), and `CreateWindowEx` then reuses the stale class with the dangling `WndProc` → crash/UB. Register with the mod's own module handle and `UnregisterClass(name, hInstance)` in `Wh_ModUninit`, every load registering fresh. desktop-live-overlay uses a `GetCurrentModuleHandle()` helper and unregisters in teardown: [UnregisterClass](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/desktop-live-overlay.wh.cpp#L2626).

**3. Multiple `explorer.exe` instances produce stacked, duplicate HUDs.** With `@include explorer.exe` the mod injects into *every* `explorer.exe` process (e.g. when "Launch folder windows in a separate process" is enabled, or any extra shell instance). Each one spawns `DashboardThreadProc` and creates another overlay parented to the same desktop `WorkerW`, so you get N overlapping HUDs, and there is no single-instance guard.

This also points at a "[mods as tools](https://github.com/ramensoftware/windhawk/wiki/Mods-as-tools:-Running-mods-in-a-dedicated-process)" fit: the mod installs **no function hooks** — it only creates a window and reads PDH counters on a timer — yet a crash in it takes down the shell (see #1). As a tool mod (`@include windhawk.exe`, rename `Wh_ModInit`/`Wh_ModUninit` → `WhTool_*`, paste the launcher snippet) the framework runs a single dedicated instance and a fault can't destabilize Explorer; [keep-rainmeter-always-bottom](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/keep-rainmeter-always-bottom.wh.cpp) is a tool mod that manipulates desktop windows cross-process the same way (Rainmeter itself parents its widgets to the desktop from a separate process). If you instead keep it in-process, follow desktop-live-overlay's model (create the window on the `WorkerW` thread via a `CreateWindowExW` hook rather than a home-grown message-loop thread), and you still need to drop the duplicate-instance problem.

**4. `@author Random Author` is a placeholder, and the authorship disclosure looks off.** Please set `@author` to your real name/handle (it matches `@github`/`@homepage` `agarwaladvit96-coder`). Separately, the PR ticks "without AI assistance", but the code reads as AI-generated (section banners like "Adaptive Rendering Engine" / "System Data Gathering Pipelines", and marketing copy such as "premium" / "fully optimized"). That doesn't affect acceptance, but an accurate disclosure helps the review — and the AI-style framing is worth a pass for plausible-but-wrong Win32 assumptions like the ones above.

**5. README has no screenshot.** This is a purely visual mod, so a screenshot or short GIF of the HUD in the README would help users a lot.

**6. Primary-monitor-only positioning and no DPI handling.** `xPos = GetSystemMetrics(SM_CXSCREEN) - HUD_WIDTH - 30` pins the HUD to the primary monitor, and all sizes/offsets (`310`, `300`, `30`, `45`, paddings, font heights) are hard-coded device pixels. On a non-100% DPI display the HUD is mis-sized, and on multi-monitor setups it only ever appears on the primary screen. Account for the monitor/DPI (e.g. `GetDpiForWindow` and per-monitor work-area metrics).

<p>

- `Wh_GetStringSetting` never returns `NULL` (it returns `L""` when unset/on error), so the `if (presetSetting)` / `if (customSetting)` guards are dead code. Prefer `WindhawkUtils::StringSetting` (RAII) over the raw get + manual `Wh_FreeStringSetting`.
- `WS_POPUP | WS_CHILD` are mutually exclusive window styles — pick one (the window is `SetParent`'d into a child anyway). The explicit `SetParent(g_hWndOverlay, hwndDesktopParent)` is also redundant since the same parent is already passed to `CreateWindowEx`.
- Settings are re-read on every `WM_PAINT` (once per second). It's cheap, but caching them on a `Wh_ModSettingsChanged` (and invalidating the window there) would be cleaner.
- The description ("A premium desktop hardware monitor overlay … optimized ClearType rendering") and README copy lean on marketing language; plain English reads better in the catalog.

</p>
</details>

<p>

- `\Network Interface(*)\Bytes Received/sec` (and `Sent/sec`) sums **every** adapter, including loopback, VPN/virtual, and ISATAP/tunnel pseudo-adapters, so the reported Down/Up figures can be inflated relative to what the user expects from their physical NIC. Consider filtering to the active physical interface(s).
- `\GPU Engine(*)\Utilization Percentage` and `\PhysicalDisk(_Total)\% Disk Time` aren't available or meaningful on every system/driver; they'll quietly read 0 there. The graceful fallback (NULL counter → 0) is fine, just noting the metric can silently be unavailable.
- Parenting to the `WorkerW` that *contains* `SHELLDLL_DefView` and pinning it to `HWND_BOTTOM` places the HUD in the icon layer, where it can be occluded by desktop icons or fail to appear depending on the wallpaper host. desktop-live-overlay instead targets the sibling `WorkerW` *behind* the icons and tracks its size — worth comparing if you see visibility issues.

</p>
</details>

PR #4361, opened 2026-06-10
mac-style-brightness

kivsak:
👉 PLS GIB MONI FOR 2X SPEED BOOST: https://www.patreon.com/c/kivsak

Sick of fumbling with the tiny Windows brightness slider buried in the Action Center? This mod turns your **taskbar into a giant brightness dial** — just hover and scroll. Zero clicks, zero menus, instant response.

Built with a **triple-fallback brightness engine** (direct IOCTL → WMI → PowerScheme) that automatically picks the smoothest, least-jerky method your hardware supports — the same WMI backend used by pro tools like Monitorian and Twinkle Tray, but without installing a separate app.

🌟 **Features**
* **Taskbar Scroll Engine**: Hover the taskbar, scroll up/down. That's it. No modifier keys, no popups, no delay. Brightness changes instantly.
* **Keyboard Shortcuts**: Ctrl+Alt + Arrow Up/Down for precise step-by-step control. Ctrl+Alt+Shift+B for instant panic reset to 100%.
* **Software Dimming Overlay**: A black overlay that goes **below** your hardware minimum brightness. On **OLED laptops** this means deeper blacks, reduced burn-in risk, and significantly better battery life (OLED pixels emit less light = less power draw). On **IPS/VA panels** it kills that eye-searing glow when even 0% is still too bright at 3 AM.
* **Smart Backend Selection**: Tries direct hardware IOCTL first (bypasses Windows brightness pipeline entirely), falls back to WMI with Timeout=0 (no driver animation fighting), and only uses PowerScheme as a last resort with built-in rate limiting to prevent the infamous brightness "jerking" bug.
* **Anti-Jerk Architecture**: Rate-limited writes, poll cooldowns, and input protection ensure the mod never fights with Windows or your BIOS over who controls the backlight.

⚠️ **Note about Fn Brightness Keys**: On most laptops, hardware brightness keys (Fn+F10/F11) are handled by the BIOS firmware via ACPI and completely bypass Windows. No app can intercept them — not even PowerToys or AutoHotkey. Use **taskbar scroll** or **Ctrl+Alt+Arrows** for full control including the software overlay.

## Changelog

* Changelog item 1...
* Changelog item 2...

## Mod authorship

This mod was created by:

- [ ] The submitmitter, without AI assistance
- [ ] The submitmitter, with AI assistance
- [x] Claude
- [ ] ChatGPT
- [x] Gemini
- [ ] Other (please specify):

### DIFF
--- a/mods/mac-style-brightness.wh.cpp
+++ b/mods/mac-style-brightness.wh.cpp
@@ -1,0 +1,891 @@
+// ==WindhawkMod==
+// @id            mac-style-brightness
+// @name          Mac-Style Smooth Brightness
+// @description   Smooth brightness via taskbar scroll, F10/F11, or keyboard. Software overlay below Windows minimum.
+// @version       12.1
+// @author        kivsak
+// @github        https://github.com/kivsak
+// @include       explorer.exe
+// @compilerOptions -lgdi32 -lole32 -loleaut32
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+👉 PLS GIB MONI FOR 2X SPEED BOOST: https://www.patreon.com/c/kivsak
+
+Sick of fumbling with the tiny Windows brightness slider buried in the Action Center? This mod turns your **taskbar into a giant brightness dial** — just hover and scroll. Zero clicks, zero menus, instant response.
+
+Built with a **triple-fallback brightness engine** (direct IOCTL → WMI → PowerScheme) that automatically picks the smoothest, least-jerky method your hardware supports — the same WMI backend used by pro tools like Monitorian and Twinkle Tray, but without installing a separate app.
+
+🌟 **Features**
+* **Taskbar Scroll Engine**: Hover the taskbar, scroll up/down. That's it. No modifier keys, no popups, no delay. Brightness changes instantly.
+* **Keyboard Shortcuts**: Ctrl+Alt + Arrow Up/Down for precise step-by-step control. Ctrl+Alt+Shift+B for instant panic reset to 100%.
+* **Software Dimming Overlay**: A black overlay that goes **below** your hardware minimum brightness. On **OLED laptops** this means deeper blacks, reduced burn-in risk, and significantly better battery life (OLED pixels emit less light = less power draw). On **IPS/VA panels** it kills that eye-searing glow when even 0% is still too bright at 3 AM.
+* **Smart Backend Selection**: Tries direct hardware IOCTL first (bypasses Windows brightness pipeline entirely), falls back to WMI with Timeout=0 (no driver animation fighting), and only uses PowerScheme as a last resort with built-in rate limiting to prevent the infamous brightness "jerking" bug.
+* **Anti-Jerk Architecture**: Rate-limited writes, poll cooldowns, and input protection ensure the mod never fights with Windows or your BIOS over who controls the backlight.
+
+⚠️ **Note about Fn Brightness Keys**: On most laptops, hardware brightness keys (Fn+F10/F11) are handled by the BIOS firmware via ACPI and completely bypass Windows. No app can intercept them — not even PowerToys or AutoHotkey. Use **taskbar scroll** or **Ctrl+Alt+Arrows** for full control including the software overlay.
+*/
+// ==/WindhawkModReadme==
+
+// ==WindhawkModSettings==
+/*
+- taskbar_scroll: true
+  $name: Scroll on Taskbar
+  $description: Just hover your mouse over the taskbar and scroll up/down. No extra keys needed! This is the most reliable method.
+- f10f11_keys: false
+  $name: F10 / F11 Keys (Fn Lock only!)
+  $description: ONLY works if Fn Lock is ON (so F10/F11 send normal key codes). Does NOT work with hardware brightness keys — those bypass Windows entirely and no app can intercept them.
+- ctrlalt_scroll: false
+  $name: Ctrl+Alt + Scroll Wheel
+  $description: Hold Ctrl+Alt and scroll anywhere on screen. Useful if you don't want bare scroll on the taskbar.
+- keyboard_arrows: true
+  $name: Ctrl+Alt + Arrow Keys
+  $description: Ctrl+Alt + Up = brighter, Ctrl+Alt + Down = darker. Works everywhere, always reliable.
+- step_size: 5
+  $name: Brightness Step (%)
+  $description: How much brightness changes per scroll tick or key press. Smaller = more precise, bigger = faster. Range 1-25.
+- software_dimming: true
+  $name: Software Dimming Overlay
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

---

A few things need to be addressed before this can be merged.

**1. This should be a "tool mod", not injected into `explorer.exe`.** The mod has **no function hooks at all** — it only uses APIs that work system-wide from any process: `SetWindowsHookEx(WH_MOUSE_LL/WH_KEYBOARD_LL)`, `RegisterHotKey`, WMI/IOCTL/PowerScheme, plus its own overlay/message windows. It hosts itself in `explorer.exe` purely for a place to run, which is exactly the "[Mods as tools](https://github.com/ramensoftware/windhawk/wiki/Mods-as-tools:-Running-mods-in-a-dedicated-process)" case. Running it inside the shell causes real, user-visible problems:

- **Multiple `explorer.exe` processes run the mod multiple times** (this happens with "Launch folder windows in a separate process", and for other reasons). With two instances you get:
  - **Two `WH_MOUSE_LL` hooks** → every taskbar scroll fires `Adjust()` twice, so brightness jumps **double the step**.
  - **Two full-screen overlay windows** stacked on top of each other.
  - The second instance's `RegisterHotKey(HK_PANIC)` (and the arrow keys) **fails** with `ERROR_HOTKEY_ALREADY_REGISTERED`, silently.
- **Shell stability risk**: any crash in the IOCTL/WMI/COM/overlay code takes down the whole Windows shell.

Convert it to a tool mod: change `@include explorer.exe` → `@include windhawk.exe`, rename `Wh_ModInit`/`Wh_ModSettingsChanged`/`Wh_ModUninit` → `WhTool_ModInit`/`WhTool_ModSettingsChanged`/`WhTool_ModUninit`, and paste the launcher snippet from the [wiki page](https://github.com/ramensoftware/windhawk/wiki/Mods-as-tools:-Running-mods-in-a-dedicated-process). The snippet's mutex (`windhawk-tool-mod_<id>`) gives you single-instance behavior for free, which removes the double-step / duplicate-overlay / failed-hotkey problems above. See [`keep-rainmeter-always-bottom.wh.cpp`](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/keep-rainmeter-always-bottom.wh.cpp) for a merged tool mod that does only `SetWinEventHook`/window work, just like this one.

**2. Remove the donation/marketing copy from the README.** The README (and PR body) opens with `👉 PLS GIB MONI FOR 2X SPEED BOOST: https://www.patreon.com/c/kivsak`. That's a misleading claim — a Patreon pledge does not give a "2X speed boost" to a brightness mod — and the catalog isn't a place for donation begging headers. Please drop that line, and tone down the rest of the marketing copy ("triple-fallback brightness engine", "the same WMI backend used by pro tools like Monitorian", "Anti-Jerk Architecture") into a plain description of what the mod does. A normal, factual README is what gets merged.

For a donation URL, you can use `@donateUrl`.

**3. Add a screenshot or short GIF.** This is a visual mod (taskbar-scroll brightness + a software dimming overlay), but the README has no image. A quick GIF of scrolling on the taskbar and the overlay dimming would help users a lot. Only `i.imgur.com` and `raw.githubusercontent.com` are allowed image hosts.

<p>

Minor polish — none of this is blocking, so it's your call.

- **`LoadLibraryW(L"PowrProf.dll")` by bare name** (`InitPowerApi`) uses the default DLL search order, which includes the loading executable's directory. `PowrProf.dll` is not a KnownDLL. Since the mod runs in `explorer.exe` (a protected directory) the hijack surface is negligible, but it's cheap to harden: `LoadLibraryExW(L"PowrProf.dll", nullptr, LOAD_LIBRARY_SEARCH_SYSTEM32)`.
- **Leftover debug logging.** `KbProc` logs every F1–F24 keypress system-wide (`Wh_Log(L"[KB] vk=...")`) — that looks like debugging that was left in. Also, the custom `[Main]`/`[IOCTL]`/`[WMI]`/`[PWR]` prefixes are redundant: Windhawk already prefixes every `Wh_Log` line with the mod name.
- **Dead/confused code in `InitWmi`.** `bool comOwned = SUCCEEDED(hr);` is computed but never used, and there's no matching `CoUninitialize` to balance the `CoInitializeEx`. The manual `IWbem*` forward-declarations (lines 76–91) are also dead — the very next line does `#include <wbemidl.h>`, which defines all of them; the "manual declarations for MinGW compatibility" comment no longer applies. Drop the unused decls.
- **Window classes registered with the wrong module handle.** `RegisterClassExW`/`CreateWindowExW`/`UnregisterClassW` use `GetModuleHandleW(NULL)` (explorer's base), while the class `WndProc`s live in the mod DLL. It happens to round-trip because you unregister symmetrically, but the idiomatic handle here is the mod's own module (you already have `GetSelfModule()`).
- **Settings-change data race.** `Wh_ModSettingsChanged` → `LoadSettings()` writes the `g_*` setting globals on Windhawk's thread while `MainThread` reads them. Harmless in practice (plain `int`/`bool`), but only correct because of luck; worth a note.
- **Version string mismatch.** `@version 12.1`, but the logs and the readme-less code say `v12.0` (`"=== Brightness v12.0 INIT ==="`, `"[Main] v12.0 starting"`).

</p>
</details>

<p>

- **The F10/F11 option swallows those keys globally.** When enabled, `KbProc` returns `1` for plain F10/F11 in *every* application, so it breaks F11 fullscreen in browsers, F10 menu activation, etc., everywhere — not just on the taskbar. That's inherent to a low-level keyboard hook with no per-app target; at minimum the README should call out that it's a system-wide override (the current `$description` only explains the Fn-Lock caveat).
- **The animation timer runs at ~60 FPS forever.** `SetTimer(..., TM_ANIM, 16, OnAnim)` wakes the thread every 16 ms for the life of the mod, even when `g_alphaSmooth` has already settled and `OnAnim` early-returns. That's a continuous wakeup load that prevents CPU idle — a bit ironic for a mod that advertises battery savings. Consider killing the timer once the alpha reaches its target and restarting it from `Adjust`/`Panic` when a new transition begins.
- **The 2 s poll does a full WMI `ExecQuery` on the WMI backend.** `OnPoll` → `WmiRead` runs `SELECT CurrentBrightness FROM WmiMonitorBrightness` every 2 s once the cooldowns expire. That's a relatively heavy query to run on a steady timer just to detect external slider changes; a longer interval (or a WMI event sink) would be lighter. The IOCTL path is cheap, so this only matters on WMI-backend hardware.
- **`Ctrl+Alt+Arrow` hotkeys collide with Intel graphics' screen-rotation hotkeys** on some systems — `RegisterHotKey` will fail there and the arrow-key feature silently won't work.
- **The full-screen `WS_EX_TOPMOST` overlay can fight other topmost/fullscreen windows.** You re-assert `HWND_TOPMOST` every 3 s (`OnTop`), which mostly works, but exclusive-fullscreen apps/games may still end up above or below it unpredictably.
- **The `\\.\LCD` IOCTL backend only controls the primary integrated panel.** Multi-monitor / external-display setups won't be affected by the IOCTL or WMI paths the way a per-monitor tool (Monitorian/Twinkle Tray, which the README name-drops) would be — worth setting expectations in the README.

</p>
</details>

PR #4362, opened 2026-06-10
Update hide-desktop-icon-text.wh.cpp

kivsak:
👉 PLS GIB MONI FOR 2X SPEED BOOST: https://www.patreon.com/c/kivsak

Tired of the stiff, boring default text labels cluttering your beautiful wallpaper? Inspired by minimalist aesthetic setups and clean UI design, this cinema-grade mod completely overhauls how your desktop icons are displayed.

Unlike older, buggy alternatives, this mod features a completely rewritten memory-based alpha-channel engine that guarantees **zero visual artifacts** — completely destroying the infamous Windows 11 "black square" bug while perfectly preserving drop shadows and modern rounded corners!

🌟 **Features**
* **Triple-Threat Text Engine**: Granular control! Hide text labels independently for **Programs/Games**, **Regular Files**, and **Folders**.
* **Smart Detection**: Dynamically scans your desktop structure to automatically differentiate between a shortcut to an app and a simple document.
* **Flawless Overlays Removal**: Deletes those tiny blue shortcut arrows AND the yellow/blue UAC Admin Shields live, using a multi-layer draw intercept engine for full Win 11 compatibility.

## Changelog

* Added the ability to hide UAC Admin Shields on desktop icons.
* Introduced granular control to hide text labels independently for files, programs/icons, and folders.
* Added the option to hide icon overlays (shortcut arrows and shields) independently.
* Fixed a visual bug in File Explorer where black squares appeared instead of shortcut overlays.

## Mod authorship

This mod was created by:

- [x] Claude
- [ ] ChatGPT
- [x] Gemini
- [ ] Other (please specify):

### DIFF
--- a/mods/hide-desktop-icon-text.wh.cpp
+++ b/mods/hide-desktop-icon-text.wh.cpp
@@ -1,47 +1,57 @@
 // ==WindhawkMod==
-// @id              hide-desktop-icon-text
-// @name            Hide Desktop Icon Text and Shortcut Arrows
-// @description     Provides options to hide icon text labels (keeping folder names) and remove shortcut arrow overlays on the desktop.
-// @version         1.5.0
-// @author          kivsak
-// @github          https://github.com/kivsak
-// @include         explorer.exe
-// @architecture    x86-64
+// @id            hide-desktop-icon-text
+// @name          Hide Desktop Icon Text and Shortcut Arrows
+// @description   Provides advanced options to independently hide text labels for apps, files, and folders, plus flawlessly removes shortcut arrows and the UAC shield.
+// @version       2.5.0
+// @author        kivsak
+// @github        https://github.com/kivsak
+// @include       explorer.exe
+// @architecture  x86-64
 // @compilerOptions -lole32 -lshlwapi -lgdi32 -lcomctl32
 // ==/WindhawkMod==
 
 // ==WindhawkModReadme==
 /*
-Provides options to hide desktop icon labels (keeping folder names) and remove the shortcut arrow overlay.
+👉 PLS GIB MONI FOR 2X SPEED BOOST: https://www.patreon.com/c/kivsak
 
-### Features
-* **Hide Icon Text**: Hides text labels for files and shortcuts on the desktop while preserving folder names.
-* **Hide Shortcut Arrows**: Removes the standard shortcut arrow overlay live without requiring registry edits or administrator rights.
+Tired of the stiff, boring default text labels cluttering your beautiful wallpaper? Inspired by minimalist aesthetic setups and clean UI design, this cinema-grade mod completely overhauls how your desktop icons are displayed.
+
+Unlike older, buggy alternatives, this mod features a completely rewritten memory-based alpha-channel engine that guarantees **zero visual artifacts** — completely destroying the infamous Windows 11 "black square" bug while perfectly preserving drop shadows and modern rounded corners!
+
+🌟 **Features**
+* **Triple-Threat Text Engine**: Granular control! Hide text labels independently for **Programs/Games**, **Regular Files**, and **Folders**.
+* **Smart Detection**: Dynamically scans your desktop structure to automatically differentiate between a shortcut to an app and a simple document.
+* **Flawless Overlays Removal**: Deletes those tiny blue shortcut arrows AND the yellow/blue UAC Admin Shields live, using a multi-layer draw intercept engine for full Win 11 compatibility.
 
 ### Notes
-* The shortcut arrow removal affects the entire running Explorer session because the overlay resides in the shared system image list.
-* Disabling the shortcut arrow feature (or the mod itself) restores the arrows only after Explorer restarts or the icon cache refreshes.
-
-### Known limitation
-Folder detection matches the drawn text against actual desktop folder names. If a *file* has
-exactly the same name as a *folder* on the desktop, that file's label will also be kept.
+* The overlay removal affects the entire running Explorer session. Disabling the feature restores them only after Explorer restarts.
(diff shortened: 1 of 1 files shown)


AromaKitsune:
@kivsak The black squares are still appearing on large icons

image

kivsak:
зображення
зображення
зображенняф

I missed this bug because I use small icons, and everything works fine at that size. The issue only appears when you switch to Large icons (or zoom in with Ctrl + mouse wheel). That's when Windows glitches and renders a black square instead of transparency.

How to fix: for now, you can just make the icons smaller. Meanwhile, I'll update the code so the transparent blank scales properly for larger icon sizes.

kivsak:
@AromaKitsune 

Ah, my bad, I just noticed you specifically mentioned large icons.

m417z:
### Submission review

---

**README is ad copy with a donation solicitation.** The README now opens with `👉 PLS GIB MONI FOR 2X SPEED BOOST: https://www.patreon.com/c/kivsak` and is otherwise marketing hyperbole ("cinema-grade", "Triple-Threat Text Engine", "memory-based alpha-channel engine", "zero visual artifacts", "completely destroying the infamous … bug"). The catalog expects a neutral, factual description — please drop the Patreon line, rewrite the README to plainly state what each setting does, and tone down `@description` (it has the same promotional wording). Also restore the honest "Known limitation" note that was removed (see Functionality notes — it still applies).

You can use `@donateUrl` for a donation URL.

**`SHGetFileInfoW_Hook` strips `SHGFI_ADDOVERLAYS`/`SHGFI_OVERLAYINDEX` for *every* caller in explorer.exe, so it removes all icon overlays, not just shortcut arrows.** `SHGFI_ADDOVERLAYS` composites *every* registered overlay into the returned icon — shortcut arrows, shared-folder, and third‑party `IShellIconOverlayIdentifier` handlers (OneDrive/cloud sync status, Dropbox, TortoiseGit, …). The hook fires whenever `hide_arrows || hide_shield` and isn't restricted to the desktop, so enabling "Hide Shortcut Arrows" silently strips OneDrive sync badges and all other overlays throughout File Explorer too. That's far more than the setting name implies. The targeted image-list / `ImageList_DrawIndirect` path already removes the specific link/shield overlay indices, so the blanket `SHGFI_ADDOVERLAYS` strip should be dropped (or, at minimum, scoped so it doesn't nuke unrelated overlays).

**UAC-shield removal is process-wide and on by default.** Between the `SHGetStockIconInfo` hook blanking `SIID_SHIELD`, `ReplaceIcon(g_shieldSysIdx, …)` overwriting the shield in the shared system image list, and the `SHGFI_*` strip above, the admin shield disappears everywhere in explorer.exe — not only on desktop shortcuts, but on File Explorer's "this needs administrator rights" affordances, property pages, and command-bar buttons. The shield is a security cue, so removing it globally — and `hide_shield: true` makes that the default — is a notable change to hand users unasked. Please default `hide_shield` to `false`, and ideally limit the shield removal to the desktop view rather than the whole process.

**Data race on `g_shieldOverlays` / `g_linkOverlay`.** `ImageList_DrawIndirect_Hook` reads `g_shieldOverlays` (iterates the `std::vector`) from any thread with no gate, while `InitAndPatchImageLists()` does `g_shieldOverlays.clear()` + `push_back(...)`. That function runs in `Wh_ModAfterInit` and again on power-resume (`ReapplyOverlayPatches`), at which point the hook is already live and can fire from another thread. Iterating the vector on one thread while another clears/reallocates it is a data race / potential use-after-free — and it's during normal startup, not just unload. Build the indices into locals and publish them safely (guard the reader/writer with a lock, or swap in an immutable snapshot) instead of mutating the shared vector in place while it may be read.

<p>

Minor polish — none of this affects users directly, so it's your call.

* `ImageList_DrawIndirect_Hook` writes back into the caller-owned struct (`pimldp->fStyle &= ~ILD_OVERLAYMASK`) before calling the original. It works, but mutating a caller's parameter is fragile — if the caller reuses that struct for a later draw it keeps the cleared mask. Safer to copy the struct, clear the mask on the copy, and pass the copy to `ImageList_DrawIndirect_Original`.

</p>
</details>

<p>

* **"Smart Detection" classifies every shortcut as a Program.** In `UpdateFolderCache`, any item with `SFGAO_LINK` becomes `ItemType::Program` (`// Most desktop shortcuts point to apps`) without resolving the link target. So a `.lnk` to a folder or a document is treated as a Program and hidden under "Hide Programs & Games", not its real category. The README's claim that it differentiates "a shortcut to an app and a simple document" overstates what the code does.
* **Name-keyed cache; collisions and unknowns fall back to File.** `g_itemTypes` is keyed by display name, and `ShouldHideLabel` treats any name not in the map as `File`. Two desktop items sharing a display name (e.g. one from the user's Desktop and one from the Public Desktop) collapse to a single type. The previous version's accurate "Known limitation" note — a *file* with the same name as a *folder* keeps its label — still applies and should be restored to the README rather than dropped.
* **Overlay/shield patches persist until Explorer restarts.** The `ReplaceIcon`/`SetOverlayImage` edits and the shield overwrite live in the shared system image list, so disabling the mod (or a feature) doesn't restore arrows/shields until Explorer restarts or the icon cache refreshes. This was already the case for arrows; it now also applies to the shield, so make sure the README's note reflects that. Relatedly, because `Wh_ModSettingsChanged` always reloads, each settings change re-runs `InitAndPatchImageLists` and appends a fresh blank icon via `ReplaceIcon(-1, …)` without removing the prior ones — a slow growth of the shared list across many toggles (cleared on Explorer restart).

</p>
</details>

PR #4372, opened 2026-06-11
[New Mod]: mod to support custom user-defined commands in Explorer

deepak-raven:
## Changelog

* Initial release of the mod.

## Mod authorship

This mod was created by:

- [ ] Claude
- [ ] ChatGPT
- [x] Gemini
- [ ] Other (please specify): 

### DIFF
--- a/mods/explorer-context-menu-custom-items.wh.cpp
+++ b/mods/explorer-context-menu-custom-items.wh.cpp
@@ -158,7 +158,8 @@ std::vector<std::wstring> GetSelectedPaths(HWND hwndTarget) {
     if (!pBrowser) return paths;
 
     IShellView* pShellView = nullptr;
-    if (SUCCEEDED(pBrowser->QueryActiveViewer(&pShellView)) && pShellView) {
+    // Fixed: Changed QueryActiveViewer to QueryActiveShellView
+    if (SUCCEEDED(pBrowser->QueryActiveShellView(&pShellView)) && pShellView) {
         IDataObject* pDataObject = nullptr;
         if (SUCCEEDED(pShellView->GetItemObject(SVGIO_SELECTION, IID_PPV_ARGS(&pDataObject))) && pDataObject) {
             FORMATETC fmt = { CF_HDROP, nullptr, DVASPECT_CONTENT, -1, TYMED_HGLOBAL };
--- a/mods/explorer-context-menu-custom-items.wh.cpp
+++ b/mods/explorer-context-menu-custom-items.wh.cpp
@@ -1,32 +1,27 @@
 // ==WindhawkMod==
 // @id              explorer-context-menu-custom-items
 // @name            Explorer Context Menu Custom Items
-// @description     Adds filtered custom user-defined options to the classic
-// @version         1.0
-// @author          Deepak
-// @github          https://github.com/deepak-raven
+// @description     Adds filtered custom user-defined options to the classic (Show more options) context menu.
+// @version         1.0.1
+// @author          Deepak Santharaju
+// @github          https://github.com/ramensoftware
 // @include         explorer.exe
-// @compilerOptions -lcomctl32 -lshlwapi -lole32 -loleaut32
+// @compilerOptions -lcomctl32 -lole32 -loleaut32 -luuid
 // ==/WindhawkMod==
 
 // ==WindhawkModReadme==
 /*
 # Explorer Context Menu Custom Items
 
-Easily add, filter, and manage custom shortcut actions inside the classic
-context menu (**Show more options**).
+Easily add, filter, and manage custom shortcut actions inside the classic context menu (**Show more options**).
 
 ### Key Features:
-* **Dynamic Commands:** Run any CLI tool or executable (like `whisperx`,
-`ffmpeg`, or `python`) directly on a highlighted file.
-* **Extension Filtering:** Restrict commands to specific formats (e.g., `.mp4,
-.mkv`) so your menu stays clean and relevant.
-* **Folder Toggle:** Choose whether specific shortcuts show up when
-right-clicking folders or folder backdrops.
+* **Dynamic Commands:** Run any CLI tool or executable (like `whisperx`, `ffmpeg`, or `python`) directly on a highlighted file.
+* **Extension Filtering:** Restrict commands to specific formats (e.g., `.mp4, .mkv`) so your menu stays clean and relevant.
+* **Folder Toggle:** Choose whether specific shortcuts show up when right-clicking folders or folder backdrops.
 
 ### How to use variables:
-* Use `%1` in your command string to automatically pass the absolute file path
-of your selected file.
+* Use `%1` in your command string to automatically pass the absolute file path of your selected file. It is recommended to use double quotes around it in your command settings (e.g., whisperx "%1") to handle spaces in file paths properly.
 */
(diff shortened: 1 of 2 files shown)


m417z:
### Submission review

Also fix `@description` which is now truncated.

---

Useful, self-contained idea, and the approach (runtime injection into the classic `TrackPopupMenuEx`, no registry writes) is the right reversible pattern. The closest existing mod, [`open-in-admin-terminal`](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/open-in-admin-terminal.wh.cpp), adds a single hardcoded verb the same way; this generalizes it to user-defined commands, so it doesn't duplicate it — but several details of that mod's injection logic are worth borrowing. The main items:

**1. The hook injects into *every* popup menu in Explorer, not just file/folder context menus.** `HookedTrackPopupMenuEx` appends your items to any menu shown via `TrackPopupMenuEx` in `explorer.exe`, gated only by the extension/folder filter. With the default config of a command (`allowFolders: true`, `extensions: ""`): for those, `GetSelectedPaths` returns nothing, so `targetPath` is empty, `isFolder` becomes `true` (no dot in an empty string), and an all-files/`allowFolders` command is appended. You need to confirm the menu actually belongs to a shell view before touching it. `open-in-admin-terminal` does exactly this — it walks the owner chain looking for `SHELLDLL_DefView` ([`IsShellViewWindow`, L1158-L1174](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/open-in-admin-terminal.wh.cpp#L1158-L1174)) and only injects when that matches ([L1530-L1531](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/open-in-admin-terminal.wh.cpp#L1530-L1531)). Adding the same gate also fixes the fact that you currently run the full `CoCreateInstance(ShellWindows)` + window-enumeration in `GetSelectedPaths` on *every* popup menu in the process, most of which have nothing to do with files.

**2. In the `TPM_RETURNCMD` path you return the custom command ID back to the shell.** The classic "Show more options" menu is invoked with `TPM_RETURNCMD`, so `TrackPopupMenuEx_orig` returns the chosen ID. When the user picks a custom item you run the command but then `return result;` — i.e. you hand `0x8000+i` back to the shell's `CDefFolderMenu`, which will then call `IContextMenu::InvokeCommand` with that out-of-range verb offset. Best case it's a harmless no-op, but you can't rely on that. Return `0` after handling a custom item, the way `open-in-admin-terminal` does ([L1560-L1566](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/open-in-admin-terminal.wh.cpp#L1560-L1566)):
```cpp
if ((uFlags & TPM_RETURNCMD) && result >= MIN_CUSTOM_ID &&
    result < MIN_CUSTOM_ID + (int)activeCommands.size()) {
    ExecuteCustomCommand(result - MIN_CUSTOM_ID, hwnd);
    return 0;   // don't let the shell invoke our ID as a verb
}
```

**3. The folder-background fallback passes the folder's *display name*, not its path.** When nothing is selected, `GetSelectedPaths` falls back to `pFolder->get_Title(&bstrTitle)` and pushes that into `paths`. `Folder.Title` is the *display name* (e.g. `Documents`, `Desktop`, `This PC`), not a filesystem path — so `%1` expands to a non-path string and the command runs against garbage. For the folder-background case, obtain the real path instead (e.g. query the folder's `Self` `FolderItem` and read its `.Path`, or go through `IShellBrowser`/`IPersistFolder2`).

**4. Filtering is computed from fragile string heuristics.** Two related correctness problems in the advertised "extension filtering" / "folder toggle":
   - `isFolder` is inferred by comparing the last `.` and last `\`/`/` positions in the path string. A folder named `Backup.2024` is misclassified as a file with extension `.2024`, and an extensionless file like `Makefile` is treated as a folder. Determine the type from the item itself instead (`GetFileAttributesW(...) & FILE_ATTRIBUTE_DIRECTORY`).
   - Extension matching uses `ContainsStringNoCase(cmd.extensions, targetExt)`, a substring test. So `extensions: .mkv` matches a file `clip.mk` (`".mkv"` contains `".mk"`), and `.mp4` would match unintended names. Split the comma list and compare each token exactly (trimming whitespace), rather than substring-searching.

**5. `%1` is substituted unquoted, so any path with a space breaks the command.** `cmd.replace(pos, 2, path)` inserts the raw path, then the whole thing is wrapped in `cmd.exe /c …`. A perfectly normal path like `C:\Users\John Doe\clip.mp4` gets split into multiple arguments by `cmd`. Either wrap the substitution in quotes when injecting, or — to avoid double-quoting when a user already quoted `%1` — document the convention and update the README example to show `whisperx "%1"`. As written, the common case (spaces in the path) silently fails.

<p>

* **Unused `shlwapi`.** No `Path*`/`Str*` functions are used (you do everything with `std::wstring`), so `#include <shlwapi.h>` (L60) and `-lshlwapi` in `@compilerOptions` (L10) can both be dropped.
* **Use `WindhawkUtils::StringSetting` (RAII)** in `LoadSettings` instead of raw `Wh_GetStringSetting` + manual `Wh_FreeStringSetting`. Also, `Wh_GetStringSetting` never returns `NULL` (it returns `L""` on error/unset), so the `!szLabel` / `!szCommand` / `szExts ? … : L""` NULL checks are unnecessary — just test for an empty string.
* **`MIN_CUSTOM_ID = 0x8000` is a plausible collision range** with shell menu command IDs. `open-in-admin-terminal` uses a single high, unlikely value (`0xBF31`); consider a base less likely to overlap a real verb offset.
* **`dwRefData = (DWORD_PTR)hwnd`** is stored on the subclass but never read in `ExplorerMenuSubclassProc`. Drop it or use it.
* **README has no screenshot.** This mod has a visible effect (new context-menu entries); a screenshot or short GIF of the menu would help users understand it at a glance.
* **`GetSelectedPaths` runs twice per click** — once in `HookedTrackPopupMenuEx` for filtering and again inside `ExecuteCustomCommand`. You could pass the already-resolved paths through to avoid the second COM enumeration.

</p>
</details>

<p>

* **`cmd.exe /c` + `CREATE_NEW_CONSOLE`** means even a GUI executable flashes a console window, and the command is subject to `cmd`'s own parsing/escaping quirks. For CLI tools this is probably what you want; if you intend to also support launching GUI apps cleanly, consider `ShellExecuteExW` (as `open-in-admin-terminal` does) or a direct `CreateProcessW` without the `cmd` wrapper.
* **Multi-selection spawns one process per file.** `ExecuteCustomCommand` loops over every selected path and launches a separate `cmd.exe` for each, so selecting 50 files opens 50 consoles. The README describes acting on "a highlighted file" (singular) — decide whether per-file or a single invocation with all paths is the intended behavior.
* **The `WM_COMMAND` subclass path is effectively dead for the classic menu.** Because the classic context menu uses `TPM_RETURNCMD`, the subclass is never installed for it (you only subclass when `!(uFlags & TPM_RETURNCMD)`), so item clicks are handled purely by the return-value path — same as `open-in-admin-terminal`. The subclass only matters for non-`RETURNCMD` menus, which (after fixing item 1) you won't be injecting into anyway, so you may be able to drop it.
* **Only `TrackPopupMenuEx` is hooked, not `TrackPopupMenu`.** Fine for the classic shell menu (which uses `Ex`), just noting it in case some path you care about uses the non-`Ex` variant.

</p>
</details>

PR #4375, opened 2026-06-11
Create saved-run-commands.wh.cpp

communism420:
## Summary

This pull request adds a new Windhawk mod: **Saved Run Commands**.

The mod adds a compact button to the standard Windows Run dialog (`Win + R`). The button opens a small `Saved Commands` window where users can save, edit, delete, and run frequently used commands.

## Features

- Adds a small unobtrusive button to the Windows Run dialog.
- Opens a `Saved Commands` window with a list of saved commands.
- Saves the current command from the Run dialog.
- Allows typing and saving a new command directly in the mod window.
- Supports editing already saved commands.
- Supports selecting and deleting multiple commands at once.
- Runs saved commands by double-clicking them.
- Supports per-command `Run as administrator` mode.
- Stores commands persistently in the mod storage directory.
- Includes a setting for the maximum number of saved commands.
- Detects the Run dialog structurally instead of relying on localized window titles.

## Implementation Notes

- Target process: `explorer.exe`.
- Uses subclassing to add controls and handle the Run dialog safely.
- Uses WinEvent hooks and window creation hooks for fast detection of the Run dialog.
- Uses `std::vector<std::wstring>` for storing commands in memory.
- Saves commands as UTF-16 text with one command per line, including metadata for elevated launch.
- Existing command files without metadata remain compatible.

## Changelog

* Changelog item 1...
* Changelog item 2...

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini
- - [x] Another AI (please specify): Codex

### DIFF
--- a/mods/saved-run-commands.wh.cpp
+++ b/mods/saved-run-commands.wh.cpp
@@ -4,8 +4,9 @@
 // @description     Adds a compact saved commands list to the Windows Run dialog.
 // @version         1.0.0
 // @author          communism420
+// @github          communism420
 // @include         explorer.exe
-// @compilerOptions -lcomctl32 -lgdi32 -lshell32
+// @compilerOptions -std=c++20 -lcomctl32 -lgdi32 -lshell32
 // ==/WindhawkMod==
 
 // ==WindhawkModReadme==
@@ -21,8 +22,16 @@ Saved commands can be edited with the "Update selected" button.
 Multiple saved commands can be selected and deleted at once.
 Each saved command has a per-row "Run as administrator" checkbox.
 
+## Screenshots
+
+![Windows Run dialog with the Saved Run Commands button](https://raw.githubusercontent.com/communism420/Media-Host-For-My-Windhawk-Mods/refs/heads/main/photo%20saved%20run%20commands%20%28button%29.png)
+
+![Saved Commands window](https://raw.githubusercontent.com/communism420/Media-Host-For-My-Windhawk-Mods/refs/heads/main/photo%20saved%20run%20commands%20%28window%29.png)
+
 Notes:
 - Commands are stored as one line per command in the mod storage folder.
+- Elevated commands expand environment variables before a best-effort
+  ShellExecute runas launch.
 - The UI added by this mod is intentionally English-only.
 */
 // ==/WindhawkModReadme==
@@ -43,7 +52,7 @@ Notes:
 #include <windowsx.h>
 #include <commctrl.h>
 #include <shellapi.h>
-#include <tlhelp32.h>
+#include <windhawk_utils.h>
 
 #include <algorithm>
 #include <cwctype>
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

---

Nice idea, and the mod is reversible (button + subclass removed on unload, commands kept in mod storage). The main problem is that the dialog-detection machinery is far heavier than the feature needs, and it imposes that cost on all of Explorer.

**1. Five overlapping detection layers run continuously across the whole shell — collapse to one event-driven trigger.** To notice a rarely-shown dialog, the mod installs all of:
- a global `CreateWindowExW` function hook (fires for *every* window created in `explorer.exe`);
- `SetWinEventHook(EVENT_OBJECT_CREATE, EVENT_OBJECT_SHOW, …)` — a range that fires for essentially every UI object create/show event in the shell, and each callback runs `IsRunDialog()`, which itself does **three** `EnumChildWindows` passes (`FindRunCommandControl` + `FindBrowseButton` + `HasRunDialogShape`);
- per-thread `WH_CBT` **+** `WH_CALLWNDPROC` **+** `WH_GETMESSAGE` hooks installed on *every* thread in the process — `WH_GETMESSAGE`/`WH_CALLWNDPROC` sit on the hot message path of every Explorer UI thread;
- a background thread (`HookInstallerThreadProc`) that wakes **every 2 seconds for the lifetime of Explorer** and re-enumerates all threads (`CreateToolhelp32Snapshot`) and all top-level windows (`EnumWindows` → `IsRunDialog`);
- `EVENT_SYSTEM_DIALOGSTART` (the one targeted, appropriate trigger).

  This is exactly the "constant polling / global hooks / process-wide enumeration" pattern the maintainer repeatedly pushes back on. The Run dialog appears only on `Win+R`, so all of the above is pure idle overhead on the shell the rest of the time. Recommend keeping **only** `EVENT_SYSTEM_DIALOGSTART` (`WINEVENT_INCONTEXT`, scoped to the current process — which you already register): it fires precisely when a dialog appears and hands you the dialog `HWND` on its own thread, so you can run `IsRunDialog` → subclass → `InstallRunButton` right there. Then delete the `EVENT_OBJECT_CREATE..SHOW` range hook, the three per-thread message hooks (`InstallHooksForThread`, the `ThreadHook` table, `g_hooksLock`, `InstallCbtHooksForProcessThreads`), the polling thread, and — since detection no longer depends on it — the `CreateWindowExW` hook as well.

**2. Use `WindhawkUtils::SetWindowSubclassFromAnyThread` instead of raw `SetWindowSubclass`.** The only reason for the CBT/CallWnd/GetMsg plumbing above is to "get onto" the dialog's thread so the raw `SetWindowSubclass` in `MaybeSubclassDialog` is a same-thread call. Windhawk provides [`WindhawkUtils::SetWindowSubclassFromAnyThread` / `RemoveWindowSubclassFromAnyThread`](https://github.com/ramensoftware/windhawk/wiki/Development-tips) precisely so you can subclass a window owned by another thread from any context. Switching to it makes the consolidation in (1) trivial: detect from the `EVENT_SYSTEM_DIALOGSTART` callback and subclass directly, regardless of whether the callback ran in- or out-of-context.

**3. `RegisterClassExW` swallows `ERROR_CLASS_ALREADY_EXISTS`, which can hand out a dangling `WndProc` after a reload.** `RegisterSavedWindowClass()` returns `true` on `ERROR_CLASS_ALREADY_EXISTS`, and the class is (re)registered lazily on each `ShowSavedCommandsWindow`. You do correctly `UnregisterClassW` in `Wh_ModUninit`, but that call fails if a `WindhawkSavedRunCommandsWindow` window is still alive (e.g. `CloseSavedCommandsWindow`'s `SendMessageTimeoutW(WM_CLOSE, …, 1000)` times out because the owning UI thread is busy). The registration then survives the unload with its `lpfnWndProc` pointing into the now-unmapped mod image; on the next load the `ALREADY_EXISTS` swallow reuses that stale class, and the next `CreateWindowExW` dispatches to a dangling `WndProc` → crash. Fix: register the class once in `Wh_ModInit`, keep the `UnregisterClassW` in `Wh_ModUninit`, and treat `ERROR_CLASS_ALREADY_EXISTS` as a real failure (so a leaked stale class disables the window instead of crashing). See the window-class-lifecycle guidance — register fresh per load with the current `WndProc`.

**4. Add a screenshot or GIF to the README.** The mod adds a visible button and a window, so a short image/GIF of the button in the Run dialog and the Saved Commands window would make the README much clearer (only `i.imgur.com` / `raw.githubusercontent.com` are allowed hosts).

<p>

- **Use `WindhawkUtils::SetFunctionHook()`** instead of raw `Wh_SetFunctionHook` with `void*` casts for the (remaining, if any) function hook — it's type-safe and the modern convention. (Moot if you drop the `CreateWindowExW` hook per item 1.)
- **Cross-window signaling messages.** `kMsgProbeRunDialog` / `kMsgUnsubclassRunDialog` / etc. are fixed `WM_APP + N` constants sent/posted to the Run dialog (`#32770`), a window the mod doesn't own. `WM_APP`-range values can collide with another mod that subclasses the same dialog. For your own inter-window signaling prefer a process-unique id from `RegisterWindowMessage`. (Most of these go away with the simplification in item 1.)
- **`FillCommandsList` holds `g_commandsLock` across `SendMessageW(list, LB_ADDSTRING, …)`.** It happens to be safe (Win32 critical sections are recursive and the owner-draw callbacks run on the same thread), but copying the commands into a local vector under the lock and releasing it before touching the listbox is cleaner and avoids the foot-gun.
- A few globals (`g_savedCommandsWindow`, `g_maxCommands`) are read/written from more than one thread without synchronization. Benign on x86/x64 for aligned pointer/int access, and mostly moot once the worker thread is removed, but worth a glance.

</p>
</details>

<p>

- **"Run as administrator" diverges from the normal Run path.** For non-elevated commands you set the edit text and click OK, so the real Run dialog does its usual processing (environment-variable expansion, MRU, shell verbs). The elevated path instead calls `ShellExecuteW(L"runas", …)` on a string you split yourself in `SplitCommandForShellExecute`. `ShellExecuteW` does **not** expand environment strings, so an entry like `%windir%\system32\cmd.exe` (or other shell-only constructs the Run box accepts) will run fine non-elevated but fail when "Run as administrator" is ticked. Consider expanding env vars (`ExpandEnvironmentStringsW`) before the elevated launch, and/or documenting the limitation.
- **Command splitting heuristic.** `SplitCommandForShellExecute` probes prefixes with `SearchPathW` to find the executable boundary; this is reasonable but will mis-split unusual inputs (e.g. an existing-but-unintended path prefix). Acceptable for the feature, just noting it isn't equivalent to the shell's own parsing.

</p>
</details>

communism420:
Now it's perfect

m417z:
In addition to the review below, I'm not sure about the `FindRunCommandControl`. Specifically this line:
`HWND control = GetDlgItem(runDialog, kRunEditId);`
On my computer, there's a combo box with ID 12298 (0x300A), which then has a child control with ID 1001 (`kRunEditId`=0x3E9`).

First, verify that `FindRunCommandControl`'s first check actually works by adding logs. If not, I suggest checking for a combo box with ID 12298. Then, I'd remove the `FindCommandControlProc` fallback.

### Submission review

---

- **The `WINEVENT_OUTOFCONTEXT` fallback (lines 1942-1945) is effectively dead.** It registers the callback on the `Wh_ModInit` thread, which has no message loop, so out-of-context events would never be dispatched. In practice `WINEVENT_INCONTEXT` (filtered to the current process) succeeds, so the fallback is rarely if ever reached — but it won't work if it ever is. Either drop it or route it through a worker thread with a pump.

- **`IsRunDialog` runs on every reposition.** `RunDialogSubclassProc` calls `InstallRunButton` (which calls the full `IsRunDialog` — 3× `EnumChildWindows`) on every `WM_WINDOWPOSCHANGED`/`WM_SHOWWINDOW`/`WM_SETTEXT`, before checking the installed prop. But only run dialogs are subclassed, so this check is unnecessary.

communism420:
Now it's truly perfect

m417z:
Remove `StartWinEventHook`/`StopWinEventHook` altogether, that was the intention of the previous review.

communism420:
Now it's perfect (I hope)

m417z:
I meant, remove all `SetWinEventHook` and unhook calls, not just move the code around. This code is not correct. You can read the previous review for an explanation.

communism420:
Now it's truly perfect

bats586:
Seems useful, would be good to be able to store "Run as Administrator" option for each command

PR #4386, opened 2026-06-12
Taskbar Fluent Media Player Update 1.3.0

Salyts:
### Added
- **Text scrolling for long track titles and artist names**
  - Enable/disable scrolling for title and artist separately
  - Adjustable scroll speed (1-10)
  - Two scroll modes: bounce (back and forth) and loop (continuous)
  - Configurable pause duration and spacing between repeats
- **Individual corner radius control** for player, buttons, and album art
  - Use single value (e.g., '4') for uniform corners
  - Use four values (e.g., '4 2 4 2') for individual corner customization
 
### Changed
- More compact spacing between title and artist by default
- Improved visual feedback when clicking on the player

## Mod authorship

This mod was created by:

- - [x] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/taskbar-fluent-media-player.wh.cpp
+++ b/mods/taskbar-fluent-media-player.wh.cpp
@@ -299,7 +299,7 @@ If you encounter any issues, bugs, or have suggestions for new features, please 
     - cornerRadius: "4"
       $name: Media player corner radius
       $name:ru-RU: Радиус скругления медиаплеера
-      $description: "Use single value (e.g., '4' for uniform corners, or four space-separated values (e.g., '4 2 4 2') for individual corners."
+      $description: "Use single value (e.g., '4') for uniform corners, or four space-separated values (e.g., '4 2 4 2') for individual corners."
       $description:ru-RU: "Используйте одно значение (например, '4') для одинаковых углов, или четыре значения через пробел (например, '4 2 4 2') для каждого угла отдельно."
     - enablePlayerHoverEffect: "auto"
       $name: Player hover effect
@@ -352,7 +352,7 @@ If you encounter any issues, bugs, or have suggestions for new features, please 
     - buttonCornerRadius: "4"
       $name: Media buttons corner radius
       $name:ru-RU: Скругление кнопок управления
-      $description: "Use single value (e.g., '4' for uniform corners, or four space-separated values (e.g., '4 2 4 2') for individual corners."
+      $description: "Use single value (e.g., '4') for uniform corners, or four space-separated values (e.g., '4 2 4 2') for individual corners."
       $description:ru-RU: "Используйте одно значение (например, '4') для одинаковых углов, или четыре значения через пробел (например, '4 2 4 2') для каждого угла отдельно."
     - buttonColor: "255 255 255"
       $name: Media buttons icons color (RGB)
@@ -616,7 +616,7 @@ If you encounter any issues, bugs, or have suggestions for new features, please 
     - albumArtCornerRadius: "4"
       $name: Album art corner radius
       $name:ru-RU: Радиус скругления обложки альбома
-      $description: "Use single value (e.g., '4' for uniform corners, or four space-separated values (e.g., '4 2 4 2') for individual corners."
+      $description: "Use single value (e.g., '4') for uniform corners, or four space-separated values (e.g., '4 2 4 2') for individual corners."
       $description:ru-RU: "Используйте одно значение (например, '4') для одинаковых углов, или четыре значения через пробел (например, '4 2 4 2') для каждого угла отдельно."
     - showAppIcon: false
       $name: Show media app icon overlay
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

---

A couple of things on the new scrolling feature.

**1. `enableTitleScrolling` is read inverted — the title marquee does the opposite of the toggle.** In `LoadSettings`:

```cpp
g_settings.enableArtistScrolling = Wh_GetIntSetting(L"MainSettings.TextAreaSetting.enableArtistScrolling") != 0;
g_settings.enableTitleScrolling  = Wh_GetIntSetting(L"MainSettings.TextAreaSetting.enableTitleScrolling")  != 1;  // <-- bug
```

The title line uses `!= 1` instead of `!= 0`. Since the setting defaults to `true` (`Wh_GetIntSetting` returns `1`), `1 != 1` evaluates to **`false`**, so title scrolling — the headline feature, on by default — is actually *disabled* out of the box, and enabling the toggle in the UI turns it *off* while disabling it turns it *on*. Change it to `!= 0` to match `enableArtistScrolling`.

**2. The scroll animation runs on a 60 Hz worker thread that `SendMessage`s into the UI thread every frame, and its teardown isn't UI-thread-safe.** Two related concerns with `ScrollTimerThreadProc` / `StopScrollTimer`:

- *Per-frame cross-thread cost:* every 16 ms the thread calls `RunFromWindowThread`, which installs **and removes** a `WH_CALLWNDPROC` hook on the taskbar UI thread and does a synchronous `SendMessageW`. Installing/uninstalling a hook plus a blocking cross-thread send 60×/second is a continuous, non-trivial load on the shell's UI thread for as long as any text is scrolling — far heavier than the existing 500 ms main-timer usage of the same helper.
- *Teardown race:* `StopScrollTimer` runs first inside `StopTimerThread` and waits with a plain `WaitForSingleObject(g_scrollTimerThread, 1000)` — no message pump. The main timer's own stop logic deliberately uses `MsgWaitForMultipleObjects(..., QS_SENDMESSAGE)` and pumps sent messages precisely because `StopTimerThread` can run on the taskbar UI thread. On that path, a scroll-thread `SendMessage` that's in flight to the UI thread can't complete (the UI thread is blocked in `WaitForSingleObject`, not pumping) → up to a 1 s stall, after which the handles are closed while the thread is still blocked, orphaning it; `StartTimerThread` → `StartScrollTimer` then spins up a *second* scroll thread, so you end up with a duplicate hammering the UI thread.

  The cleanest fix for both is to drive the animation from the **UI thread** instead of a worker thread — e.g. a `Windows::UI::Xaml::DispatcherTimer` (or `DispatcherQueueTimer`) that ticks `TickScrollState` + `UpdateScrollTransforms` directly, or a XAML `Storyboard`/composition animation for the offset. That eliminates the per-frame hook+`SendMessage` entirely and removes the teardown race. If you'd rather keep the worker thread, at minimum mirror the main timer's `QS_SENDMESSAGE`-pumping wait in `StopScrollTimer` so it can't deadlock/orphan on the UI-thread stop path.

<p>

- **Unbalanced parenthesis in the corner-radius descriptions.** The three new English `$description` strings read `Use single value (e.g., '4' for uniform corners, or four space-separated values (e.g., '4 2 4 2') for individual corners.` — the `(e.g., '4'` is never closed. (The `ru-RU` versions are fine.)

- **The scroll thread wakes every 16 ms even when nothing is scrolling.** When both toggles are off (or all text fits), `ScrollTimerThreadProc` still loops at 60 Hz and `continue`s. It's cheap, but you could skip `StartScrollTimer()` when neither `enableTitleScrolling` nor `enableArtistScrolling` is enabled and re-evaluate on settings change.

</p>
</details>

<p>

- **Unsynchronized access to `g_titleScroll` / `g_artistScroll`.** These structs are written by the UI thread in `RefreshPlayerContents` (`active`, `offset`, `forward`, `pausing`, `pauseTick`, `textWidth`, `viewWidth`) while the scroll thread reads/writes the same fields in `TickScrollState`. The scalar fields won't tear on x64, so the worst realistic outcome is a momentary jump in the scroll position rather than a crash — but if you keep the worker-thread design, consider confining the mutation to one thread (another reason the UI-thread-timer approach above is nicer) or guarding the few fields.

- **Scroll direction vs. `mirrorLayout`.** The offset always advances the same way (`Canvas::SetLeft(tb, -offset)`), independent of `mirrorLayout`. For the mirrored, right-aligned layout the reveal direction may feel backwards — worth a quick check on that configuration.

</p>
</details>

Salyts:
@m417z

m417z:
Number 2 wasn't addressed, right?

Salyts:
> Number 2 wasn't addressed, right?

Right, wait

Salyts:
@m417z

Salyts:
> Number 2 wasn't addressed, right?

something else is wrong?

PR #4389, opened 2026-06-12
Add Mouse Button Remapper mod

wygodad:
This PR adds a new mod: **Mouse Button Remapper**.

The mod remaps mouse buttons (side back/forward, middle/wheel click, and
optionally left/right) to keyboard keys with optional modifiers (Ctrl,
Shift, Alt, Win) — e.g. the upper side button can send F5 to refresh the
page in a browser. It is aimed at mice whose vendors provide no remapping
software eg. Pulsar X2V3 eS mini.

Mouse Button Remapper

Features:
- Each button has its own on/off switch — disabled buttons keep their
  original (manufacturer) function, so a single button can be remapped
  and the rest left untouched.
- Key selection via dropdown, plus a custom-key field accepting key
  names, hex virtual-key codes, and full combos like `Ctrl+F5`.
- Optional per-button double-click action (a different key on double-click).
- Configurable global hotkey that suspends/resumes all remaps, with an
  on-screen indicator and an optional confirmation sound.
- Game mode: remapping is automatically suspended while a fullscreen
  application is in the foreground.
- Localized settings (English + Polish).

Implementation notes: this is a tool mod (runs in a dedicated process,
not injected into explorer.exe) — a WH_MOUSE_LL hook works from any
process with a message loop. The hook and the RegisterHotKey toggle
hotkey live on a dedicated thread with a message loop, shut down via
WM_QUIT and an untimed wait. Settings are read by the hook thread
through an SRWLOCK-guarded snapshot with a generation counter, so it
never sees partially-updated values.

## Changelog

N/A — new mod.

## Mod authorship

This mod was created by:

- [x] Claude
- [ ] ChatGPT
- [ ] Gemini
- [ ] Other (please specify): 

### DIFF
--- a/mods/mouse-button-remap.wh.cpp
+++ b/mods/mouse-button-remap.wh.cpp
@@ -2,7 +2,7 @@
 // @id              mouse-button-remap
 // @name            Mouse Button Remapper
 // @description     Remap mouse buttons (side, middle, left, right) to keyboard keys with optional modifiers, double-click actions, a master toggle hotkey and game mode
-// @version         1.6.0
+// @version         1.7.2
 // @author          wygodad
 // @github          https://github.com/wygodad
 // @include         windhawk.exe
@@ -32,6 +32,24 @@ simply configure the other one.
 The mod runs as a tool mod in a dedicated process and installs a global
 low-level mouse hook, so it works system-wide (in every application)
 without injecting into the shell or other programs.
+
+## Mouse compatibility
+
+By default, on virtually all mice the two side buttons are reported as
+standard mouse buttons — **Forward** (XButton2) and **Back** (XButton1) —
+and the middle button as the wheel click. Those are exactly the events
+this mod intercepts, so it works out of the box on a normal mouse with no
+vendor software.
+
+However, if the mouse's own software or onboard firmware remaps a button
+to a **multimedia or keyboard function** (e.g. "Refresh", "Play/Pause",
+"Browser Back"), that press is no longer sent as a mouse button — it
+arrives as a media/keyboard command that a mouse hook cannot see, so this
+mod can no longer remap it. If a button is not being intercepted, open
+your mouse software and set it back to the plain **Forward** / **Back** /
+**Wheel click** function, then assign the keyboard key here in the mod.
+In other words: let the mouse send a normal button, and let the mod do
+the remapping — do not remap the same button in both places.
 
 ## Configuration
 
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

---

**Restore the session-0 guard from the wiki tool-mod boilerplate.** The wiki "[Mods as tools](https://github.com/ramensoftware/windhawk/wiki/Mods-as-tools:-Running-mods-in-a-dedicated-process)" snippet begins `Wh_ModInit` with an early-out for session 0, which your copy dropped:

```cpp
BOOL Wh_ModInit() {
    DWORD sessionId;
    if (ProcessIdToSessionId(GetCurrentProcessId(), &sessionId) &&
        sessionId == 0) {
        return FALSE;
    }
    // ... rest of the launcher
```

Without it, the launcher will also spawn a dedicated `windhawk.exe` tool process in session 0, where a low-level mouse hook and `RegisterHotKey` are pointless (no interactive input desktop).

<p>

- **Game-mode fullscreen check inside the LL hook.** `IsForegroundFullscreen()` runs `GetForegroundWindow` / `GetClassName` / `GetWindowRect` / `GetMonitorInfo` synchronously inside `LowLevelMouseProc` on every remapped button-down while game mode is on. It's only on click events (infrequent) and the calls are cheap, so it's fine in practice — but if you want to keep the input-blocking callback minimal, you could cache the fullscreen state and refresh it from a `SetWinEventHook(EVENT_SYSTEM_FOREGROUND, ...)` on the hook thread, then just read the cached `bool` in the hook.
- **Explicit CRT includes.** The code uses `towupper` (`<cwctype>`), `_wtoi` / `wcstoul` (`<cstdlib>`), and `wcstok_s` / `wcsncpy_s` / `swprintf_s` (`<cwchar>`) but only includes `<wchar.h>` (plus `<windows.h>`/`<atomic>`). It compiles today via transitive includes; including the headers for the symbols you actually use is more robust across toolchains.

</p>
</details>

<p>

- **OSD indicator is primary-monitor / fixed-DPI.** `ShowOsd` centers the indicator using `SM_CXSCREEN` at a fixed `280×64` with a `-22` font, so it always appears on the primary monitor and doesn't scale on high-DPI setups. It's a transient cosmetic toast, so this is purely FYI — positioning it on the monitor of the foreground window and scaling by that monitor's DPI would make it look consistent on mixed-DPI / multi-monitor configurations.

</p>
</details>

wygodad:
Thanks for the review! Addressed in the latest push:

* Game-mode fullscreen check — implemented as you suggested, @m417z. The state is now cached and only refreshed via SetWinEventHook (EVENT_SYSTEM_FOREGROUND + EVENT_OBJECT_LOCATIONCHANGE, filtered to the foreground top-level window), and only while game mode is enabled — so nothing extra runs inside the low-level mouse hook. The location-change event keeps F11-style same-window fullscreen detection working.
* Session 0 guard — added to Wh_ModInit, and restored the full tool-mod boilerplate header comment.
* OSD indicator — now scales with the screen and appears on the monitor under the cursor, so it stays readable on large / high-DPI displays.
* Readme (EN/PL) updated to document the above; version bumped to 1.7.2.

Let me know if anything else needs changing.

PR #4390, opened 2026-06-12
New mod submission: TAI [taskbar as an island] mod for windows 11

DarkionAvey:
The github page: https://github.com/DarkionAvey/windhawk-taskbar-centered-condensed

This mod is based on two existing mods: `taskbar-icon-size.wh.cpp` and `taskbar-start-button-position.wh.cpp`, and combines them with a custom `ApplyStyle` function. This mod is not extending the functionality of these two mods but rather just using them as template for potential methods to hook. This allows me to avoid playing catch up with windows updates and signature changes.

The core functionality is implemented in `ApplyStyle`. The rest of the code primarily handles debouncing and other customization features.

TAI is kept current with its upstream dependencies rather than being based on outdated forks of  `taskbar-icon-size.wh.cpp` and `taskbar-start-button-position.wh.cpp`. everytime the mod is assembled (see the GitHub repo), newest versions of these two mods are fetched and patched. 

TAI uses Windhawk brush for custom blur effects since I was not able to hook the default acrylic brush. 

In addition to explorer.exe, TAI hooks the following exe's to be able to move the start menu/search/notification center flyouts in tandem with the taskbar: `StartMenuExperienceHost.exe; ShellHost.exe; ShellExperienceHost.exe`. When hooking these exe's, `g_PartialMode` becomes true, enabling a specific logic graph that only achieves repositioning.

TAI also prevents the taskbar from overflowing by virtually expanding the available space seen by the taskbar, then scaling the taskbar down so that it fits within screen bounds. This effect is subtle and nice.

## Changelog

Initial release

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [x] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/taskbar-dock-like.wh.cpp
+++ b/mods/taskbar-dock-like.wh.cpp
@@ -2,7 +2,7 @@
 // @id              taskbar-dock-like
 // @name            TAI (taskbar as island) for Windows 11
 // @description     Centers and floats the taskbar, moves the system tray next to the task area, and serves as an all-in-one, one-click mod to transform the taskbar into an animated dock.
-// @version         1.5.243
+// @version         1.5.244
 // @author          DarkionAvey
 // @github          https://github.com/DarkionAvey
 // @include         explorer.exe
@@ -2246,7 +2246,7 @@ void ApplySettingsTBIconSize(int taskbarHeight) {
 }
 bool HookSystemTraySymbols(HMODULE module) {
     // SystemTray.dll
-    WindhawkUtils::SYMBOL_HOOK SystemTray_TaskbarViewDll_hooks[] = {
+    WindhawkUtils::SYMBOL_HOOK symbolHooks[] = {
         {
             {LR"(private: double __cdecl winrt::SystemTray::implementation::SystemTrayController::GetFrameSize(enum winrt::WindowsUdk::UI::Shell::TaskbarSize))"},
             &SystemTrayController_GetFrameSize_Original,
@@ -2278,7 +2278,7 @@ bool HookSystemTraySymbols(HMODULE module) {
             true,  // From Windows 11 version 22H2.
         },
     };
-    if (!HookSymbols(module, SystemTray_TaskbarViewDll_hooks, ARRAYSIZE(SystemTray_TaskbarViewDll_hooks))) {
+    if (!HookSymbols(module, symbolHooks, ARRAYSIZE(symbolHooks))) {
         Wh_Log(L"HookSymbols failed");
         return false;
     }
@@ -2294,7 +2294,7 @@ bool HookSystemTraySymbols(HMODULE module) {
 bool HookTaskbarViewDllSymbols(HMODULE module,
                                bool hookSystemTraySymbolsInline) {
     // Taskbar.View.dll, ExplorerExtensions.dll
-    WindhawkUtils::SYMBOL_HOOK TaskbarViewDll_hooks[] =  //
+    WindhawkUtils::SYMBOL_HOOK symbolHooks[] =  //
         {
             {
                 // For Windows 11 version 21H2.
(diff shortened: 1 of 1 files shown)


DarkionAvey:
I renamed the hooks to indicate their targets; fixed the github url; and revised the compiler options

mods/taskbar-dock-like.wh.cpp
```diff
@@ -0,0 +1,9048 @@
...
+// @name            TAI (taskbar as island) for Windows 11
+// @description     Centers and floats the taskbar, moves the system tray next to the task area, and serves as an all-in-one, one-click mod to transform the taskbar into an animated dock.
+// @version         1.5.244
+// @author          DarkionAvey
+// @github          https://github.com/DarkionAvey
+// @include         explorer.exe
+// @include         StartMenuExperienceHost.exe
+// @include         ShellHost.exe
+// @include         ShellExperienceHost.exe
+// @compilerOptions -ldwmapi -lole32 -loleaut32 -lruntimeobject -lshcore -lcomctl32 -Wl,--export-all-symbols -DWINVER=0x0A00 -lversion
```
m417z: Remove `-Wl,--export-all-symbols -DWINVER=0x0A00`, they're only needed for very old Windhawk versions.

m417z:
Thanks for the submission. My biggest concern with it is that it seems to copy the functionality of existing mods, such as "Taskbar height and icon size". It means that any bug fix or improvement in the original mods won't make it into your mod until you sync it manually. Why not have your mod work alongside the existing mods?

Also, from the screenshot, it looks like it could have been a taskbar styler theme, like [WindowGlass](https://github.com/ramensoftware/windows-11-taskbar-styling-guide/blob/main/Themes/WindowGlass/README.md):
image

Why did you decide to create a mod instead?

DarkionAvey:
The dependencies are handled automatically, I dont need to manually merge upstream changes from taskbar and start menu mods. The python script only redirects the ApplyStyle and related functions from these mods so that TAI can override them. 

Regarding overlap, I am not aware that existing mods can dynamically move things around and respond to changes to width of tray and taskbar. The main reason I made this mod is because taskbar themes hardcode the width of the taskbar and that is simply not the right way to go. 

Additionally, TAI does more theming, it moves the start menu along with the taskbar so that on ultrawide screens, the user doesnt have to move the mouse all the way to the middle of the taskbar. TAI also automatically scales down the taskbar so that the user doesnt need to click through the overflow menu - they can see all their running apps at once.

DarkionAvey:
Removed the compilerOptions mentioned in the review

m417z:
> The dependencies are handled automatically, I dont need to manually merge upstream changes from taskbar and start menu mods. The python script only redirects the ApplyStyle and related functions from these mods so that TAI can override them.

You'll still need to do that and update the mod. The idea of Windhawk is having mods and being able to combine them, not having mods which are by themselves combinations of other mods, as that can get out of control quickly. Windhawk can have better support for exporting and sharing a combination of mods and settings, and I'm planning to work on it, but I don't think such merge scripts and combination mods are the right solution.

> Regarding overlap, I am not aware that existing mods can dynamically move things around and respond to changes to width of tray and taskbar. The main reason I made this mod is because taskbar themes hardcode the width of the taskbar and that is simply not the right way to go.

Have you tried the latest version of the WindowGlass theme? I see that your repository is more than one year old, many things were improved since then.

> Additionally, TAI does more theming, it moves the start menu along with the taskbar so that on ultrawide screens, the user doesnt have to move the mouse all the way to the middle of the taskbar.

Some of it is supported by [Shell Flyout Positions](https://windhawk.net/mods/shell-flyout-positions). If features are missing, I'll consider adding them.

> TAI also automatically scales down the taskbar so that the user doesnt need to click through the overflow menu - they can see all their running apps at once.

Sounds good, but what if somebody wants that but not a dock, or not this dock? That's another advantage of Windhawk's modularity that I believe we should use.

DarkionAvey:
👍

PR #4395, opened 2026-06-13
adds minimal Task Manager stats as Taskbar icon

BlackPaw21:
## Changelog

* Changelog item 1...
* Changelog item 2...

## Mod authorship

This mod was created by:

- - [x] Claude
- - [ ] ChatGPT
- - [x] Gemini


m417z:
### Submission review

---

**1. Persistent registry write — reversibility violation.** `SaveIntervalMs` / `LoadIntervalMs` store the refresh interval under `HKEY_CURRENT_USER\Software\BlackPaw\MicroManager\UpdateIntervalMs` via `RegCreateKeyExW` / `RegSetValueExW`. That key survives after the mod is disabled/uninstalled, which breaks Windhawk's core principle that a mod leaves no persistent system changes behind. Persist the mod's own state with Windhawk storage instead — it's removed automatically on uninstall:

```cpp
void SaveIntervalMs(DWORD ms)  { Wh_SetIntValue(L"UpdateIntervalMs", (int)ms); }
DWORD LoadIntervalMs() {
    DWORD ms = (DWORD)Wh_GetIntValue(L"UpdateIntervalMs", 1000);
    if (ms != 300 && ms != 500 && ms != 1000 && ms != 3000) ms = 1000;
    return ms;
}
```

Your own AudioSwap mod already does exactly this with `Wh_SetStringValue` / `Wh_GetStringValue` ([audioswap.wh.cpp#L958](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/audioswap.wh.cpp#L958)). (If you'd rather, the interval could also be a normal Windhawk `$options` setting read in `WhTool_ModSettingsChanged` — but the registry write is the part that needs to go either way.)

**2. Top-CPU detection reads and writes `g_prevProcs` in the same pass.** In `RefreshData` the loop searches `g_prevProcs[0..g_prevProcCount)` for each process's *previous* cumulative time, while simultaneously overwriting `g_prevProcs[count]` with the *current* sample as `count` advances from 0 (lines ~423 and ~448). So by the time you process the process at iteration N, slots `[0, N)` already hold this tick's data. Whenever the process list order shifts between ticks (a process spawns/exits — i.e. almost every tick), any process whose previous-tick slot landed at a lower index than its current position won't find its prior sample → `foundInPrev = FALSE` → delta `0` → it's silently dropped from the "top CPU" comparison. The displayed top CPU consumer can therefore intermittently be wrong or missed. Fix by writing the new sample into a separate buffer and swapping at the end of the pass (double-buffer), so you never read a slot you've already overwritten:

```cpp
static struct { DWORD pid; LONGLONG time; WCHAR name[64]; } g_curProcs[MAX_PROCESSES];
// ... read from g_prevProcs, write into g_curProcs ...
// after the loop:
memcpy(g_prevProcs, g_curProcs, count * sizeof(g_prevProcs[0]));
g_prevProcCount = count;
```

**3. Add a screenshot/GIF to the README.** The mod has a clearly visible UI (tray icon + the CPU/GPU/RAM popup), but the README has no image. A short screenshot or GIF of the popup helps a lot for a visual mod like this. (`i.imgur.com` and `raw.githubusercontent.com` are the allowed hosts.)

**4. Avoid the per-tick `CreateToolhelp32Snapshot` for the top GPU process name.** `CollectGpuStats` takes a full process snapshot every tick just to map the top GPU PID → exe name, even though `RefreshData` already enumerated every process (with PID and name) into `g_prevProcs` a few lines earlier. Reuse that table to resolve the name instead of a second full enumeration.

<p>

- **`g_statsLock` is unnecessary.** `RefreshData` (WM_TIMER / WM_CREATE), the popup `WM_PAINT`, and the right-click status read all run on the single tray thread's message loop, so the cached stats are never touched concurrently. The critical section is harmless but pure overhead — possibly an AI artifact. You can drop it (or keep it if you ever move refresh to its own thread).
- **`ApplyContextMenuTheme` relies on undocumented `uxtheme.dll` ordinals (133/135/136)** and only themes the menu if `uxtheme.dll` happens to already be loaded in the `windhawk.exe` process (you use `GetModuleHandleW`, no load). It's a common pattern and degrades gracefully to a light menu, so it's fine — just noting it's purely cosmetic and on private APIs.

</p>
</details>

<p>

- **GPU total over-counts and can exceed 100%.** `CollectGpuStats` sums `\GPU Engine(*)\Utilization Percentage` across *all* engine instances (3D, Copy, Video Decode/Encode, Compute, …) and all processes into `grandTotal`. Since several engines can each be near 100% at once, the reported total (and the per-PID top value) can run well above 100% and won't match Task Manager, which effectively reflects the busiest engine. Consider taking the max across engines per PID / clamping the total to 100.
- **Left-clicking the tray icon to *close* the popup tends to reopen it.** When the popup is visible and you click the icon, the click deactivates the popup first (`WM_ACTIVATE`/`WA_INACTIVE` → `SW_HIDE`), so by the time `WM_TRAY_CALLBACK`/`WM_LBUTTONUP` runs, `IsWindowVisible(g_popupHwnd)` is already `FALSE` and the toggle falls into the `ShowPopup` branch — so the popup flickers and reopens instead of closing. The close half of the toggle is effectively dead. A short debounce (ignore a left-click that arrives within ~200 ms of the auto-hide) fixes it.
- **Top RAM consumer mixes metrics.** The RAM total row shows `MEMORYSTATUSEX::dwMemoryLoad` (system-wide % in use), while the top consumer is `WorkingSetSize / TotalPhys`. Working set includes shared pages and isn't quite the per-process "memory used" Task Manager shows (private/commit). The numbers are reasonable approximations; just flagging the inconsistency in case the displayed top RAM process ever looks off.

</p>
</details>

PR #4409, opened 2026-06-14
Add Snipping Tool Pre-Focus mod

satoshate:
Adds a new mod that focuses the Windows taskbar immediately before a plain Print Screen key press continues to the native Snipping Tool. This automates the manual workaround of clicking the taskbar first, while keeping the built-in Snipping Tool UI, region/window capture, annotation, and OCR features.

Secure Desktop/UAC prompts are intentionally out of scope.

## Changelog

New mod:

* Added Snipping Tool Pre-Focus.
* Supports configurable focus delay.
* Leaves modifier combinations such as Ctrl/Alt/Shift + Print Screen untouched by default.

## Mod authorship

This mod was created by:

- [ ] Claude
- [x] ChatGPT
- [ ] Gemini
- [ ] Other (please specify):

### DIFF
--- a/mods/snipping-tool-prefocus.wh.cpp
+++ b/mods/snipping-tool-prefocus.wh.cpp
@@ -1,0 +1,208 @@
+// ==WindhawkMod==
+// @id              snipping-tool-prefocus
+// @name            Snipping Tool Pre-Focus
+// @description     Focus the taskbar before Print Screen opens the native Windows Snipping Tool
+// @version         0.1
+// @author          satoshate
+// @github          https://github.com/satoshate
+// @include         explorer.exe
+// @architecture    x86-64
+// @compilerOptions -luser32
+// @license         MIT
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Snipping Tool Pre-Focus
+
+Focuses the Windows taskbar immediately before a plain `Print Screen` key press
+continues to the native Windows Snipping Tool.
+
+This keeps the built-in screenshot UI available, including region capture,
+window capture, full-screen capture, annotation, and OCR/text actions. The goal
+is to automate the manual workaround of clicking the taskbar before taking a
+screenshot when the currently focused elevated window makes Snipping Tool behave
+poorly.
+
+Notes:
+
+- Enable Windows' "Use the Print screen key to open Snipping Tool" setting.
+- Only plain `Print Screen` is handled. Modifier combinations are left alone.
+- Secure Desktop/UAC prompts are not captured and are intentionally out of
+  scope.
+*/
+// ==/WindhawkModReadme==
+
+// ==WindhawkModSettings==
+/*
+- focusDelayMs: 80
+  $name: Focus delay in milliseconds
+  $description: Small delay after focusing the taskbar before Windows handles Print Screen.
+- ignoreIfCtrlAltShiftPressed: true
+  $name: Ignore modifier combinations
+  $description: Leave Ctrl/Alt/Shift + Print Screen combinations untouched.
+*/
+// ==/WindhawkModSettings==
+
+#include <atomic>
+#include <windows.h>
+
+struct Settings {
+    int focusDelayMs;
+    bool ignoreIfCtrlAltShiftPressed;
+};
+
+Settings g_settings;
+HHOOK g_keyboardHook;
+HANDLE g_hookThread;
+DWORD g_hookThreadId;
+std::atomic<DWORD> g_lastFocusTick;
+
+LRESULT CALLBACK LowLevelKeyboardProc(int code, WPARAM wParam, LPARAM lParam);
+
+HMODULE GetCurrentModuleHandle() {
+    HMODULE module = nullptr;
+    if (!GetModuleHandleExW(GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS |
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

---

The core idea is fine, but the implementation blocks inside a low-level keyboard hook and injects into the shell when it doesn't need to. Two structural items and one teardown bug:

**1. `Sleep()` runs inside the `WH_KEYBOARD_LL` callback — freezes all input.** `FocusTaskbar()` is called directly from `LowLevelKeyboardProc` and calls `Sleep(focusDelayMs)` (up to 1000ms). Low-level keyboard/mouse hooks run synchronously in the OS input path: the system waits for the callback to return before delivering the keystroke, so this freezes **all** keyboard and mouse input system-wide for the whole delay. Worse, once the delay exceeds `LowLevelHooksTimeout`, Windows bypasses the hook and delivers the key anyway, so the taskbar may never get focused — the feature silently does nothing. Never do blocking work in an LL hook.

   The whole "focus, wait, then let the key through" sequence inherently requires holding the key, which is exactly what you can't do here. Redesign so the hook returns immediately: swallow the real Print Screen (`return 1;` instead of `CallNextHookEx`), signal the worker thread to focus the taskbar → wait → `SendInput` a synthetic Print Screen, and let that synthetic key fall through (it's marked injected, which your `LLKHF_INJECTED` check already skips). The hook itself then never blocks.

**2. This should be a "mods as tools" mod, not injected into `explorer.exe`.** It installs no function hooks — the only mechanism is `SetWindowsHookExW(WH_KEYBOARD_LL, ...)`, which works from any process — it operates on the desktop as a whole (taskbar) and ignores explorer's own state, and it spins up its own worker thread + message loop. That's the textbook tool-mod profile ([wiki](https://github.com/ramensoftware/windhawk/wiki/Mods-as-tools:-Running-mods-in-a-dedicated-process)). As written, any bug or hang in the hook destabilizes the shell. Convert it: `@include windhawk.exe`, rename `Wh_ModInit`/`Wh_ModSettingsChanged`/`Wh_ModUninit` → `WhTool_*`, and paste the launcher boilerplate verbatim — see the bottom of [explorer-folder-hover-menu.wh.cpp](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/explorer-folder-hover-menu.wh.cpp). Nothing is lost by moving out of explorer: `SetForegroundWindow` behaves the same from a dedicated process (neither explorer nor the tool process is the foreground process at the moment Print Screen is pressed — see the functionality note below).

**3. Teardown race can leave a live hook in an unloaded DLL → crash.** `Wh_ModInit` returns right after `CreateThread` without waiting for the worker to be ready, and `Wh_ModUninit` reads `g_hookThreadId`. If uninit runs before the worker reaches `g_hookThreadId = GetCurrentThreadId()` (and the `PeekMessageW` that creates its queue), the `WM_QUIT` post is skipped or fails; `WaitForSingleObject` then times out after 3s, the handle is closed, and uninit returns — but the worker is still alive with the LL hook installed, pointing into the mod image Windhawk is about to unmap. The next keystroke calls a dangling `LowLevelKeyboardProc` and crashes the host. The window is narrow (unload immediately after enable), but the consequence is a crash. Use a ready-event handshake before returning from init — wait on a `CreateEventW` that the worker sets once its hook/queue are up (exactly what `WhTool_ModInit` in the example mod above does). The tool-mod launcher's mutex/handshake removes this whole class of bug, so item 2 fixes this for free.

**4. `Wh_ModSettingsChanged` forces a full reload for no reason.** It always sets `*reload = TRUE`, so every settings change tears down and recreates the whole thread + hook just to re-read two integers. Since the LL hook reads `g_settings` live, just re-read them: `void Wh_ModSettingsChanged() { LoadSettings(); }` (or post a settings-changed message to the worker, as the tool-mod example does).

<p>

- `AnyModifierDown()` checks `VK_CONTROL` *and* `VK_LCONTROL`/`VK_RCONTROL` (likewise for Alt/Shift) — `VK_CONTROL` already reflects either side, so the L/R checks are redundant. Harmless, just noise.

</p>
</details>

<p>

- **`SetForegroundWindow(taskbar)` may silently fail.** When Print Screen is pressed, the foreground window is the user's (possibly elevated) app, not the mod's thread, so Windows' foreground-lock rules can reject the call. The return value is unchecked, so when it fails the taskbar isn't focused and the mod does nothing — the exact opposite of the intended behavior, with no log. Worth verifying against the real "elevated window" scenario this targets; at minimum, log the return value, and consider whether an `AllowSetForegroundWindow`-style approach is needed.

- **Focus side effect.** After the screenshot, the taskbar keeps focus and the user's window is left deactivated. That's inherent to the approach — worth a line in the README so it's not surprising.

- **Verify Print Screen actually arrives as `WM_KEYDOWN` at the LL hook.** Print Screen has a long-standing quirk of generating only a key-*up* in some message contexts. If that applies here, the `wParam != WM_KEYDOWN && wParam != WM_SYSKEYDOWN` filter would skip it entirely and the mod would never fire. Confirm on a real machine (handle `WM_KEYUP` instead/as well if needed).

- **Fixed-delay timing is inherently racy.** The hardcoded `Sleep` just hopes the focus change has landed; on a busy system the chosen value may be too short (focus not applied yet) or needlessly long. Waiting for the taskbar to actually become foreground (e.g. confirm via `GetForegroundWindow`) would be more robust than a fixed delay.

</p>
</details>

PR #4410, opened 2026-06-14
Add Caps Lock Copy/Paste Alternator mod

satoshate:
Adds a new mod that remaps the physical Caps Lock key to alternate between Ctrl+C and Ctrl+V. The original Caps Lock key event is suppressed, and Caps Lock state is forced off by default.

Secure Desktop/UAC prompts are intentionally out of scope.

## Changelog

New mod:

* Added Caps Lock Copy/Paste Alternator.
* Alternates Caps Lock presses between Ctrl+C and Ctrl+V.
* Suppresses the original Caps Lock key event.
* Adds settings for initial action, Caps Lock state handling, and idle reset.

## Mod authorship

This mod was created by:

- [ ] Claude
- [x] ChatGPT
- [ ] Gemini
- [ ] Other (please specify):

### DIFF
--- a/mods/capslock-copy-paste-alternator.wh.cpp
+++ b/mods/capslock-copy-paste-alternator.wh.cpp
@@ -1,0 +1,257 @@
+// ==WindhawkMod==
+// @id              capslock-copy-paste-alternator
+// @name            Caps Lock Copy/Paste Alternator
+// @description     Remap Caps Lock to alternating Ctrl+C and Ctrl+V
+// @version         0.1
+// @author          satoshate
+// @github          https://github.com/satoshate
+// @include         explorer.exe
+// @architecture    x86-64
+// @compilerOptions -luser32
+// @license         MIT
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Caps Lock Copy/Paste Alternator
+
+Turns the physical Caps Lock key into a stateful clipboard helper:
+
+- first press sends `Ctrl+C`;
+- second press sends `Ctrl+V`;
+- third press sends `Ctrl+C` again;
+- the original Caps Lock key event is suppressed;
+- Caps Lock LED/state is forced off by default.
+
+The mod installs a low-level keyboard hook from Explorer. It works best for
+normal desktop windows. Windows integrity levels can block input into elevated
+applications if Explorer is not elevated, so for elevated/admin windows an
+elevated helper or elevated Windhawk process may still be required.
+
+Secure Desktop/UAC prompts are intentionally out of scope.
+*/
+// ==/WindhawkModReadme==
+
+// ==WindhawkModSettings==
+/*
+- startWithPaste: false
+  $name: Start with Paste
+  $description: If enabled, the next Caps Lock press after loading the mod sends Ctrl+V instead of Ctrl+C.
+- forceCapsLockOff: true
+  $name: Force Caps Lock off
+  $description: Keep the Caps Lock toggle state off while the remap is active.
+- resetToCopyAfterSeconds: 0
+  $name: Reset to Copy after idle seconds
+  $description: Set to 0 to keep strict alternation forever. Set a positive value to reset the next action to Copy after inactivity.
+*/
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

---

The main item is the architecture: this should almost certainly be a **tool mod**, not an `explorer.exe` mod.

**Make this a "mods as tools" mod (`@include windhawk.exe`).** The mod does no function hooking at all — it only installs a system‑wide `WH_KEYBOARD_LL` hook, which works from any process and has nothing to do with Explorer. Hosting it in `explorer.exe` has two concrete downsides:

- **Crash blast radius:** a fault anywhere in the hook path takes down the shell.
- **Multiple `explorer.exe` instances fragment the state.** With *"Launch folder windows in a separate process"* enabled (and transient secondary Explorer processes generally), every `explorer.exe` loads the mod and installs its own global LL hook, each with its **own private `g_nextAction`**. Because returning `1` from an LL hook stops the chain, only one instance handles a given press — but *which* one is nondeterministic and changes as Explorer processes come and go, so the Copy/Paste alternation becomes inconsistent. `forceCapsLockOff` at init also runs once per instance.

A dedicated `windhawk.exe` process via the tool‑mod launcher gives you a single instance and removes both problems. This is exactly what the near‑identical [`caps-ime-switcher`](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/caps-ime-switcher.wh.cpp) does (Caps Lock + `WH_KEYBOARD_LL` on a dedicated worker thread). Concretely:

- `@include windhawk.exe`, add `-lshell32` (the launcher uses `CommandLineToArgvW`).
- Rename `Wh_ModInit` / `Wh_ModSettingsChanged` / `Wh_ModUninit` → `WhTool_ModInit` / `WhTool_ModSettingsChanged` / `WhTool_ModUninit`.
- Paste the launcher boilerplate verbatim from the wiki ([Mods as tools](https://github.com/ramensoftware/windhawk/wiki/Mods-as-tools:-Running-mods-in-a-dedicated-process)) — `caps-ime-switcher` (lines 412‑589) is a verbatim copy you can follow.

<p>

- **`g_hookThreadId` should be atomic.** It's a plain `DWORD` written on the hook thread and read from `Wh_ModUninit` on another thread. Use `std::atomic<DWORD>` (as `caps-ime-switcher` does with `g_workerThreadId`). Relatedly, it's set at the very top of `HookThreadProc`, *before* `PeekMessageW` creates the thread's message queue — a very fast init→uninit could `PostThreadMessageW(WM_QUIT)` before the queue exists, the message is lost, and the 3 s `WaitForSingleObject` then leaks the thread + hook. Set the id after the queue is created (or signal readiness).
- **`Wh_ModSettingsChanged` forces a full reload every time** (`*reload = TRUE`), tearing down and reinstalling the hook on each settings change. Only the initial `g_nextAction` value actually needs recomputing — `LoadSettings()` from a plain `void Wh_ModSettingsChanged()` is enough. (Moot once it's a tool mod: `WhTool_ModSettingsChanged` just calls `LoadSettings()`.)
- `g_capsDown` / `g_nextAction` are `std::atomic` but only ever touched on the single hook thread, so the atomicity isn't needed — harmless, just noise.

</p>
</details>

<p>

- **`forceCapsLockOff` reliability.** `IsCapsLockOn()` calls `GetKeyState(VK_CAPITAL)` on the hook/worker thread, which is not the foreground input thread; the toggle bit it returns can be stale on a thread that isn't pumping normal keyboard input, so `EnsureCapsLockOff()` at init may fail to detect/clear a pre‑existing Caps‑Lock‑on state. (Since you suppress the key, Caps Lock can't toggle on from the user's own presses while the mod is active, so this mainly affects the one‑time startup cleanup.)
- **Modifier desync.** `SendCtrlKey` always emits Ctrl‑down … Ctrl‑up. If the user is physically holding Ctrl (or Shift) when they press Caps Lock, the trailing Ctrl‑up releases their held modifier, and the emitted combo isn't a clean Ctrl+C/Ctrl+V. Edge case, but worth a thought if you want it robust.

</p>
</details>

PR #4427, opened 2026-06-15
Add hide-desktop-icon-labels mod

Akumy-01:
Hides text labels under Windows desktop icons without renaming items. Optionally shows caption on mouse hover.

?? Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/mods/hide-desktop-icon-labels.wh.cpp
+++ b/mods/hide-desktop-icon-labels.wh.cpp
@@ -1,0 +1,667 @@
+// ==WindhawkMod==
+// @id              hide-desktop-icon-labels
+// @name            Hide desktop icon labels
+// @description     Hides text labels under Windows desktop icons.
+// @version         1.5
+// @author          akumy
+// @github          https://github.com/Akumy-01
+// @include         explorer.exe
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

---

**Substantial overlap with an existing mod — `hide-desktop-icon-text`.** There is already a merged mod that hides desktop icon labels in `explorer.exe` by intercepting the desktop `SysListView32` paint: [mods/hide-desktop-icon-text.wh.cpp](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/hide-desktop-icon-text.wh.cpp) ("Hide Desktop Icon Text and Shortcut Arrows", by a different author). The core feature is the same. The differences I can see are:
   - This mod hides **all** labels; the existing one keeps folder names.
   - This mod adds an optional **hover caption**; the existing one adds shortcut-arrow removal instead.

   The project's strong, frequently-stated preference is to *extend* an existing mod (add an option / PR the original author's repo) rather than merge a near-duplicate, because duplicates fragment the catalog and confuse users. Please state explicitly how this is meant to differ, and consider whether "hide all labels" and "hover caption" would be better as options added to `hide-desktop-icon-text`.

PR #4430, opened 2026-06-15
Add Resource-Aware Task Terminator

insane66613:
﻿# Add Resource-Aware Task Terminator

## Summary

Adds **Resource-Aware Task Terminator**, a Windhawk utility mod that lets the user terminate a target process or app group from a sortable, resource-aware candidate chooser.

The mod can target:

- the active foreground window process
- top CPU PID candidates
- top page-fault PID candidates
- top disk I/O PID candidates
- aggregate app groups ranked by a weighted resource score

The chooser includes columns for target, PID count/PID, score, CPU, page faults, disk I/O, memory, and path. Metric cells use green/yellow/red severity shading. The UI also includes manual refresh, optional live refresh, polling interval control, and dark mode styling.

## Screenshots to attach from testing

Use the screenshots from local testing showing:

1. The dark candidate chooser with sortable metric columns and colored severity cells.
2. The settings screen showing sampling window, candidate count, super-score weights, live update settings, and dark UI toggle.
3. The aggregate app-group view showing multiple app groups ranked by resource impact.

## Submission checklist

- [x] Single new mod file only: `mods/resource-aware-task-terminator.wh.cpp`
- [x] `@id` matches file name: `resource-aware-task-terminator`
- [x] `@github` metadata is present
- [x] `@license` metadata is present
- [x] Version reset for initial public submission: `1.0.0`
- [x] Source staged as UTF-8 without BOM

## Notes

The implementation intentionally keeps the termination action behind confirmation by default. Protected process names are configurable and include Windows shell/session-critical processes by default.

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [x] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/resource-aware-task-terminator.wh.cpp
+++ b/mods/resource-aware-task-terminator.wh.cpp
@@ -1,0 +1,1810 @@
+// ==WindhawkMod==
+// @id              resource-aware-task-terminator
+// @name            Resource-Aware Task Terminator
+// @description     Terminate the active window process, a top resource offender, or a weighted resource-heavy app group with confirmation and sortable usage metrics.
+// @version         1.0.0
+// @author          Math Shamenson
+// @github          https://github.com/insane66613
+// @license         MIT
+// @include         windhawk.exe
+// @compilerOptions -lpsapi -lgdi32 -lcomctl32
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+Resource-Aware Task Terminator
+
+- Press the configured hotkey to terminate a target process or app group.
+- Modes:
+  - foreground: Kill the active window process.
+  - top_cpu: Sample CPU over a configurable window and show the top PID candidates.
+  - top_pagefaults: Sample page-fault deltas and show the top PID candidates.
+  - top_disk_io: Sample process read/write byte deltas and show the top PID candidates.
+  - aggregate_super_score: Group eligible processes by executable path/name and rank app groups by CPU, page faults, disk I/O, and memory.
+  - top_gpu_experimental: Disabled in this no-PDH build to avoid loader dependency failures.
+
+Defaults
+- Hotkey: Ctrl + Alt + F4. Pick a different hotkey if another app or Windows reserves it.
+- Protected apps: Conservative defaults including explorer.exe, windhawk.exe, taskmgr.exe, conhost.exe, dwm.exe, and critical system processes.
+
+Notes and Safety
+- You may need to run Windhawk elevated to kill elevated targets; otherwise OpenProcess(PROCESS_TERMINATE) can fail with Access Denied.
+- Do not remove critical system processes from the protected list. Killing them can destabilize Windows.
+- Terminating an app can cause data loss. Save work first.
+- Confirmation mode is enabled by default and shows the selected process/app group, mode, sampling evidence, and protection check before termination.
+- The chooser supports sortable metric columns, severity-colored metric cells, manual refresh, and optional live refresh.
+
+Troubleshooting
+- If the hotkey does not work, pick a different combo. Some Win+ combos or hotkeys taken by other apps will fail to register.
+- GPU mode is disabled in this no-PDH build.
+- This build runs the hotkey in a dedicated Windhawk tool process for a persistent per-session host.
+*/
+// ==/WindhawkModReadme==
+
+// ==WindhawkModSettings==
+/*
+- kill_mode: foreground
+  $name: Target mode
+  $description: "What to target: active window process, top resource offender, or aggregate app group."
+  $options:
+    - foreground: "Active window process"
+    - top_cpu: "Top CPU PID candidates"
+    - top_pagefaults: "Top page-fault PID candidates"
+    - top_disk_io: "Top disk I/O PID candidates"
+    - aggregate_super_score: "Aggregate app-group super score"
+    - top_gpu_experimental: "Disabled in no-PDH build"
+
+- sample_ms: 3000
+  $name: "Sampling window (ms)"
+  $description: "Sampling window for CPU/page-fault/disk/aggregate modes (500-15000). Longer windows are more stable."
+
+- candidate_count: 5
+  $name: "Candidate count"
+  $description: "How many top candidates/app groups to show (1-10). Default 5 shows the top offender plus the next four."
+
+- live_update_default: false
+  $name: "Live update by default"
+  $description: "When true, the chooser starts with live refresh enabled."
+
+- live_poll_ms: 5000
+  $name: "Live refresh interval (ms)"
+  $description: "Polling interval for live chooser refresh (1000-60000). Each refresh re-samples using the sampling window."
+
+- disable_dark_ui: false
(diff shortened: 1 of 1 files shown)


m417z:
In addition to the review below, please fill the "Mod authorship" form.

### Submission review

---

**1. "Terminate selected" terminates the wrong process (kills the top-ranked row, not the highlighted one).** The list manages its highlight manually: every time a row is selected, `LVN_ITEMCHANGED` records `state->visualSelectedRow` and then *clears the native ListView selection*:

```cpp
state->visualSelectedRow = lv->iItem;
ListView_SetItemState(state->hList, -1, 0, LVIS_SELECTED | LVIS_FOCUSED);  // clears all selection
```

But the `IDOK` ("Terminate selected") handler reads the selection back via `ListView_GetNextItem(..., LVNI_SELECTED)`, which therefore always returns `-1` and falls through to index `0`:

```cpp
if (id == IDOK) {
    int idx = state && state->hList ? ListView_GetNextItem(state->hList, -1, LVNI_SELECTED) : -1;
    state->selectedIndex = idx >= 0 ? idx : 0;   // always 0 -> top row
    ...
```

So after the user highlights row N and clicks "Terminate selected", the mod terminates the *top-scored* candidate (row 0) instead. For a process killer this means killing an unintended process. `NM_DBLCLK` already does the right thing by using `visualSelectedRow` — `IDOK` should do the same:

```cpp
if (id == IDOK) {
    int idx = state ? state->visualSelectedRow : -1;
    state->selectedIndex = idx >= 0 ? idx : 0;
    DestroyWindow(hwnd);
    return 0;
}
```

**2. Leftover third-party branding in user-facing text and identifiers.** The kill-confirmation dialog body literally reads:

```cpp
message += L"Sanity Cloud Force Kill is about to terminate a process.\n\n";
```

and every log line is prefixed `[AI Counsel]`, the host mutex is `Local\sanity-cloud-terminate-tool-host`, and the window classes are `AI_Counsel_ForceKill_Class` / `AI_Counsel_ForceKill`. None of these match the mod ("Resource-Aware Task Terminator"), and the user sees "Sanity Cloud Force Kill" right before terminating a process. Please rename all of these to the mod's own name. Separately, the `[AI Counsel]` log prefix should just be dropped — `Wh_Log` already prefixes the mod name, so a custom `[ModName]`-style prefix is redundant.

**3. Tool-mod boilerplate deviates from the wiki snippet.** The launcher is hand-modified relative to the [canonical snippet](https://github.com/ramensoftware/windhawk/wiki/Mods-as-tools:-Running-mods-in-a-dedicated-process): the mutex is a hardcoded `Local\sanity-cloud-terminate-tool-host` instead of `windhawk-tool-mod_" WH_MOD_ID`, the `kernel32.dll` fallback for `CreateProcessInternalW` was dropped, and there's extra logging. Pasting the wiki snippet verbatim (and using the `WH_MOD_ID`-based mutex name) keeps it maintainable. [`explorer-folder-hover-menu.wh.cpp`](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/explorer-folder-hover-menu.wh.cpp) has a verbatim copy at the bottom for reference.

<p>

- **Dead functions.** `PickTopCpuPid` and `PickTopPageFaultsPid` are marked `[[maybe_unused]]` and never called — they were superseded by `BuildResourceCandidateDecisions`. Remove them.
- **`TrySetDarkTheme` reloads `uxtheme.dll` on every call** (it's called once per control during `WM_CREATE`, i.e. ~10x per chooser open) via `LoadLibraryW(L"uxtheme.dll")` and never `FreeLibrary`s it — a small refcount leak. `uxtheme.dll` is effectively always loaded, so `GetModuleHandleW(L"uxtheme.dll")` (no load, no leak) is simpler; if you do need to load it, use `LoadLibraryExW(L"uxtheme.dll", nullptr, LOAD_LIBRARY_SEARCH_SYSTEM32)` once and cache the handle. (DLL-hijack risk is negligible here since the host is `windhawk.exe` in a trusted directory, so this is purely a tidiness/leak nit.)
- **Leftover debug line.** `OutputDebugStringW(L"[AI Counsel] Wh_ModInit entered\n")` at the top of `Wh_ModInit` looks like a debugging leftover — drop it.
- **`state.sortColumn = (state.mode == KillMode::AggregateSuperScore) ? 3 : 3;`** — both branches are `3`; the ternary is a no-op. Just assign `3`.
- **Redundant NULL checks on `Wh_GetStringSetting`.** It never returns NULL (it returns `L""` when unset), so the `if (keyStr && *keyStr)` NULL guard can be simplified to an empty check. Consider `WindhawkUtils::StringSetting` (RAII) to drop the manual `Wh_FreeStringSetting` calls.
- **Critical-process protection relies on the editable list for `windhawk.exe`.** `IsProcessProtected` hardcodes only lsass/csrss/wininit/smss/services/winlogon; `windhawk.exe` is protected only via the default `protected_apps` string, so a user who edits that list could let the tool kill the Windhawk service process. Consider hardcoding `windhawk.exe` (and the host itself) alongside the other criticals.

</p>
</details>

<p>

- **Live refresh can change what "selected" points at.** When live update is on, a re-sample replaces and re-sorts `*state->candidates`, but `visualSelectedRow` is a row *index*, not tied to a PID — so the highlighted row may now refer to a different process than the one the user was looking at when they decide to terminate. Consider keying the selection to the PID (or pausing live refresh once a row is highlighted) so the user can't terminate an unexpected target after a refresh.
- **"CPU %" semantics.** The per-process CPU figure is `process_delta / (kernel+user system delta) * 100`, i.e. share of *total* busy time across all cores — so on a multi-core machine a process pegging one core shows well under 100%. That's fine as a relative ranking, but the bare "CPU %" column header may read oddly to users; a note or a different label could help.
- **Aggregate grouping key.** Groups are keyed by full image path when available, falling back to exe name. That's reasonable; just note that two different programs installed at the same path (uncommon) would merge, and same-named exes at different paths stay separate.

</p>
</details>

insane66613:
Updated in the latest commits (`8f0364a`, `a2cb9de`).

Review items addressed:
- Fixed `Terminate selected` so it uses the manually tracked highlighted row (`visualSelectedRow`) instead of falling back to the top-ranked row.
- Removed leftover Sanity Cloud / AI Counsel branding from the confirmation text, logs, window class/name, and tool-host identifiers.
- Reworked the tool-mod launcher to use the Windhawk `WH_MOD_ID`-based mutex pattern and restored the `kernel32.dll` fallback for `CreateProcessInternalW`.

Additional cleanup included:
- Removed dead CPU/page-fault picker helpers.
- Removed leftover debug output.
- Avoided repeated `LoadLibraryW("uxtheme.dll")` calls.
- Simplified the no-op sort-column ternary.
- Hardcoded `windhawk.exe` as protected.

For the CPU column note, I kept the existing all-logical-processor-normalized calculation because it better represents total system impact, and renamed the display to `CPU total %` / `CPU usage during sample, normalized across all logical processors (%)` to avoid ambiguity.

The mod authorship form is filled in the PR body, and the GitHub checks are currently passing.

PR #4443, opened 2026-06-16
Add Taskbar Clock Spacer v1.1

sb4ssman:
Adds **Taskbar Clock Spacer** (v1.1): a small companion mod for [Taskbar Clock Customization](https://windhawk.net/mods/taskbar-clock-customization) that introduces an elastic `%s%` spacer token for the top/bottom clock line formats. Each `%s%` becomes a gap, and the clock's leftover fixed width is shared out evenly between the gaps — so dense multi-line clocks can justify their items edge-to-edge instead of bunching.

This is deliberately a standalone companion rather than a change to the canonical customizer: the generated-layout approach didn't fit that mod's direction (m417z/my-windhawk-mods#68), so it lives separately, does nothing without Taskbar Clock Customization installed, and leaves that mod completely untouched.

Highlights:

- `%s%` between items in the Top/Bottom line formats: first item hugs the left edge, last hugs the right, gaps share the leftover width evenly; stacked `%s%%s%` weights a gap double.
- `{spacer}` inside the **Weather format** does the same for weather items — it rides through the weather service verbatim (which would consume `%s` as its sunset token) and becomes an identical elastic gap.
- The fixed clock width is taken from Taskbar Clock Customization's own **Max width** setting automatically (read as a setting, never measured — no layout feedback), or from this mod's Max clock width / Line width override.
- The generated rows are rewritten in place on clock ticks, not rebuilt, so the visual tree stays stable; lines without a spacer token are left completely alone.
- Namespace-scope XAML state is `[[clang::no_destroy]]`-guarded and cleanup runs on the taskbar UI thread, so unclean Explorer shutdown can't release XAML objects off-thread.

Tested live on Windows 11: multi-line top/bottom formats with single and stacked spacers, width pinned exactly to the clock mod's Max width across text changes (no width creep), weather `{spacer}` justification, settings reload, and disable/unload restore.

## Changelog

* v1.0 → v1.1: the branch is rebuilt on current main (the previous branch base was long stale). Fixed a width-feedback defect where the generated rows took their width from a live measurement and the clock could ratchet permanently wider; the width now comes only from settings (including reading the clock mod's own Max width), and the generated panel is pinned to it so lines can never exceed the configured width. Stopped clearing the clock mod's MaxWidth on the shared panel. Added the `{spacer}` weather-format token. In-place per-tick updates replace the old rebuild-per-second behavior.

## Mod authorship

This mod was created by:

- - [x] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/taskbar-clock-spacer.wh.cpp
+++ b/mods/taskbar-clock-spacer.wh.cpp
@@ -1,0 +1,1118 @@
+// ==WindhawkMod==
+// @id              taskbar-clock-spacer
+// @name            Taskbar Clock Spacer
+// @description     Companion for Taskbar Clock Customization: adds a %s% elastic spacer token that distributes leftover clock width between items. Windows 11 only.
+// @version         1.1
+// @author          sb4ssman
+// @github          https://github.com/sb4ssman
+// @include         explorer.exe
+// @architecture    x86-64
+// @compilerOptions -lole32 -loleaut32 -lruntimeobject -lversion
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Taskbar Clock Spacer
+
+Adds a `%s%` elastic spacer token to the Windows 11 taskbar clock, so clock items
+can be pushed apart to fill a fixed width instead of bunching together.
+
+## Two requirements — please read before installing
+
+**1. This mod does nothing on its own.** It is a companion for
+[Taskbar Clock Customization](https://windhawk.net/mods/taskbar-clock-customization).
+That mod produces the clock text; this mod only rearranges it. Install and
+configure that mod first.
+
+**2. The clock needs a fixed width.** An elastic spacer distributes *leftover*
+width. If the clock sizes itself to its own text there is no leftover width,
+every gap computes to zero, and the result looks exactly as if the mod were not
+installed. Set a fixed width using either:
+
+- **Max width** in Taskbar Clock Customization's settings, or
+- **Max clock width** in this mod's settings.
+
+Either one works. 120 px is a reasonable starting point.
+
+Windows 11 only. This mod does not work on Windows 10.
+
+## What it does
+
+Put `%s%` between items in the clock's Top Line or Bottom Line format. Each `%s%`
+becomes a gap, and all leftover width is shared out evenly between the gaps.
+
+| Format | Result |
+| --- | --- |
+| `%time%%s%%date%` | time hugs the left edge, date hugs the right, gap fills the middle |
+| `%time%%s%%date%%s%%weekday%` | three items, two equal gaps |
+| `%time%%s%%date%%s%%s%%weekday%` | Double-spacer: more space is weighted between date and weekday |
+
+The first item always hugs the left edge and the last always hugs the right edge,
+so the line stays anchored as the text changes width.
+
+### Spacers inside the weather
+
+The weather service substitutes `%s` as its sunset token, so `%s%` cannot be
+written inside Taskbar Clock Customization's **Weather format**. Write
+`{spacer}` there instead, for example:
+
+```
+%c{spacer}🌡️%t{spacer}🌬️%w
+```
+
+`{spacer}` passes through the weather service verbatim, arrives in the clock
+line, and becomes the same elastic gap as `%s%` — so weather items justify
+with the rest of the clock.
+
+## Setup
+
+1. Install **Taskbar Clock Customization** and set up your clock format.
+2. Set a **Max width** in its settings, for example `120`.
+3. Install this mod.
+4. Edit the clock mod's **Top line** or **Bottom line** to put `%s%` between
+   items, for example `%time%%s%%date%`.
+
+The `%s%` token passes through Taskbar Clock Customization untouched and is
+interpreted here at display time.
+
+## Troubleshooting
+
+**`%s%` disappears and nothing moves.** This is the fixed-width problem in
+requirement 2 above. Set a **Max width** in Taskbar Clock Customization, or a
+**Max clock width** here. The mod also writes a one-line explanation to the
+Windhawk log the first time it detects this.
+
+**Nothing happens at all.** Confirm Taskbar Clock Customization is installed and
+enabled, and that `%s%` is in its **Top line** or **Bottom line** setting — not in
+the tooltip, the middle line, or the weather format.
+
+**The spacer works but the clock is the wrong width.** Adjust the same Max width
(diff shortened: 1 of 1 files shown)


m417z:
I think it makes more sense to have it as part of Taskbar Clock Customization. You can create a PR in the development repo here: https://github.com/m417z/my-windhawk-mods
Let me know what you think.

sb4ssman:
Sounds good, I'll work on integrating it into Taskbar Clock Customization in a fork of your repo.

m417z:
Aside from the notes below, I don't understand why the mod hooks all these functions, when just hooking `DateTimeIconContent::OnApplyTemplate` seems to be enough.

Also, remove `(pre-2604)` from the comment as it confuses the source code parser for the symbol cache.

### Submission review

---

**1. `HookSymbols` is called twice for `taskbar.dll` — collapse it into one call.** `HookTaskbarDllSymbols` resolves the four core symbols in one `HookSymbols` call and then `TrayUI::StartTaskbar` in a second call against the same module (lines 932 and 944). Each additional `HookSymbols` call for the same module re-triggers symbol resolution, which is the slow part; all of a module's hooks should be resolved in a single array/single call. Merge `startTaskbarHooks` into `taskbarDllHooks` (keep the `TrayUI::StartTaskbar` entry `optional = true`) and derive the flag afterward:

```cpp
// after the single HookSymbols(h, taskbarDllHooks, ...) call:
g_trayUiStartTaskbarHooked = (TrayUI_StartTaskbar_Original != nullptr);
```

**2. Add a screenshot or GIF to the README.** The effect is visual (items justified across a fixed clock width vs. bunched), and a before/after image makes the mod much easier to understand at a glance than the ASCII table alone. `i.imgur.com` and `raw.githubusercontent.com` are the allowed hosts.

**3. `[[clang::no_destroy]]` on `g_states` is unnecessary and leaks the vector's buffer on every unload (line 307).** The stated reason — avoiding an off-thread release of the XAML weak refs at shutdown — isn't actually a hazard: `winrt::weak_ref` release is just an in-process control-block decrement and is safe from any thread. Worse, with the attribute in place the automatic destructor never runs, so `g_states.clear()` in `ClearSpacerStates` (which zeroes the size but retains the heap buffer) leaves that buffer permanently leaked on each disable/reload. Since the element destructors are safe at process exit, just drop the attribute — the normal destructor is both correct and leak-free. (The elements are still explicitly released in `ClearSpacerStates`, so nothing else changes.)

<p>

Minor polish — none of this affects users on a standard setup, so it's your call.

- **`Wh_ModUninit`'s fallback runs `ClearSpacerStates()` off the UI thread (line 1099).** When `FindCurrentProcessTaskbarWnd()` returns null, cleanup runs directly on the arbitrary uninit thread, contradicting the "must run on the UI thread" comment two lines above. It's mostly benign (if the taskbar window is gone the weak refs are likely already dead), but you could skip the direct call entirely in that branch since there's no UI thread to marshal to.

- **`StartInitialScan` reads the non-atomic `g_scanThread` without synchronization (line 966).** `Wh_ModAfterInit` (init thread) and `TrayUI_StartTaskbar_Hook` (taskbar thread) can both call it; in the rare overlap two scan threads could be created. In practice the two callers are mutually exclusive, so this is negligible — noting it only for completeness.

</p>
</details>

<p>

- **Collapsing the source text block vs. TCC re-setting `Visibility`.** `CollapseSourceTextBlock` zeroes the original block's size and hides it, but the comment notes TCC re-sets `Visibility` on its own schedule. Zeroing the size guards against width creep, but there could be a brief frame where the original text flashes back before the next tick re-collapses it. Worth a look on slow text-change cadences; not a blocker.

</p>
</details>

sb4ssman:

sb4ssman:
Thanks — and thanks for confirming there's no correctness or stability defect left.

**1. README.** Done, and it was worth doing. There is a new "Try the built-in option first" section right after the requirements, and `@description` now points at the difference too. It names Taskbar Clock Customization 1.8's **Text alignment → Justified** as the zero-install option: ordinary spaces for gaps, a non-breaking space inside items. It then lists what `%s%` adds, kept to the differences that hold up:
- gaps only where you put them. Justified stretches every space and can widen inter-character spacing, which the maintainer himself observed inside `MB/s`;
- weighted gaps (`%s%%s%`);
- a per-gap minimum width;
- `{spacer}` inside the Weather format.

The "Relationship" section now describes the upstream outcome accurately.

One factual correction on the framing, for the record. The review says the maintainer "has said twice … that they'd rather extend TCC than merge a companion". The upstream thread ends the other way. On 2026-06-29, in [m417z/my-windhawk-mods#68](https://github.com/m417z/my-windhawk-mods/pull/68), after testing the `Justify` approach, m417z wrote:

> "The most visible downside that I didn't notice before is that Windows adds spaces between characters, it's visible in "MB/s" for example - it has large gaps between characters. It works better with more than one space, but then it becomes a guessing game, so not ideal. […] So it doesn't work as well as I hoped. I also gave it some thought, and for now I think that creating additional TextBlock controls is out of scope for this mod. **We can merge https://github.com/ramensoftware/windhawk-mods/pull/4443 if you prefer, let me know.**"

So the maintainer offered to merge this PR after weighing the built-in alternative. The README now gives users the facts to choose between the two.

**Optional items, all taken:**
- The stale comments are fixed: the `StartTaskbar` "preferred path" paragraph now describes the `LoadLibraryExW` late-load path as what it is, and the template references and the validator note are gone.
- The scaffolding is flattened. The three single-use namespaces and their aliases are gone, and so are the forwarding wrappers and the `SetExceptionLogger` indirection (the dispatch now calls `Wh_Log` directly). `LoadInt` became `std::clamp` at the two call sites. The dispatch also gained a `ran` flag, so two concurrent dispatches cannot each run a callback twice.
- `ClearSpacerStates` guards each entry on its own and zeroes `textToken` right after unregistering, so the uninit retry is idempotent.
- The init log uses `WH_MOD_VERSION`.

**Functionality notes:**
- **Line height:** taken. `CopyTextStyle` now carries `LineHeight` and `LineStackingStrategy`.
- **`Spacing`:** deliberately *not* applied to the generated panel. An unspaced multi-line block is one TextBlock whose lines are separated only by line height. TCC's *Text spacing* is the shared panel's gap *between* the time and date blocks, not between lines, so copying it between our rows would add a gap the unspaced block doesn't have.
- **Build before record:** fixed. The state is pushed into `g_states` first and built in place, so a throw after the insert still leaves an entry that unload can remove.

PR #4447, opened 2026-06-16
Windows 7 File Open/Save Toggle Bar Icon

Leymonaide:
Restores the bitmap arrow glyph on the "Toggle Folders" button in open/save dialogs.

This glyph was changed to an icon in later versions of Windows. The icon scaling differs, so it wasn't possible to just
make the icon look like the Windows 7 glyph.

Preview image

## Changelog

* This pull request introduces a new mod.

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/win7-file-open-save-toggle-bar-icon.wh.cpp
+++ b/mods/win7-file-open-save-toggle-bar-icon.wh.cpp
@@ -7,8 +7,8 @@
 // @github          https://github.com/Leymonaide
 // @twitter         https://twitter.com/Leym0naide
 // @homepage        https://leymonaide.github.io/
-// @include         notepad.exe
-// @compilerOptions -lcomdlg32 -lgdi32 -lntdll
+// @include         *
+// @compilerOptions -lgdi32
 // @license         MIT
 // ==/WindhawkMod==
 
@@ -89,6 +89,14 @@ void SHLogicalToPhysicalDPI(int *px, int *py)
 thread_local void *g_pCurrentToggleBar = nullptr;
 thread_local HWND g_hwndCurrentToggleBar = nullptr;
 
+// Worth keeping in mind:
+// "Capture relies on the toggle bar being created via CreateWindowExW. If a
+// future build (or a differently-behaving app, once the scope broadens) creates
+// the toolbar through a path your hook doesn't see, the handle stays null and
+// ScaleAndSet… quietly logs 'hwndToggleBar is null' and does nothing — graceful,
+// but the feature silently won't apply. The analogous mod hooks CreateToolbarEx
+// for the same purpose; worth keeping in mind if you broaden @include."
+// https://github.com/ramensoftware/windhawk-mods/pull/4447#issuecomment-4765563181
 using CreateWindowExW_t = decltype(&CreateWindowExW);
 CreateWindowExW_t CreateWindowExW_orig;
 HWND WINAPI CreateWindowExW_hook(DWORD dwExStyle,
@@ -216,12 +224,8 @@ BOOL Wh_ModInit()
 
     LoadSettings();
 
-    RTL_OSVERSIONINFOW osvi;
-    RtlGetVersion(&osvi);
-
-    HMODULE hUser32 = LoadLibraryExW(L"user32.dll", nullptr, LOAD_LIBRARY_SEARCH_SYSTEM32);
-    pfnGetDpiForWindow = (decltype(pfnGetDpiForWindow))GetProcAddress(hUser32, "GetDpiForWindow");
-    FreeLibrary(hUser32);
+    pfnGetDpiForWindow = (decltype(pfnGetDpiForWindow))GetProcAddress(
+        GetModuleHandleW(L"user32.dll"), "GetDpiForWindow");
 
     if (!Wh_SetFunctionHook(
         (void *)CreateWindowExW,
@@ -256,6 +260,18 @@ BOOL Wh_ModInit()
     g_hbmArrowUp = LoadBitmapW(g_hmodComdlg32_7, (LPCWSTR)0x242);
     Wh_Log(L"hbmArrow up: %p", g_hbmArrowUp);
 
+    if (!g_hbmArrowDown || !g_hbmArrowUp)
+    {
+        Wh_Log(
+            L"Failed to load one or more bitmaps from the module \"%s\". "
+            L"Please ensure that bitmap resources 577 (0x241) and 578 (0x242) "
+            L"exist in the target binary. An unmodified binary from Windows 7 "
+            L"will feature these resources.",
+            g_spszComdlg32Path.c_str()
+        );
+        return FALSE;
+    }
+
     return TRUE;
 }
 
@@ -263,6 +279,12 @@ BOOL Wh_ModInit()
 void Wh_ModUninit()
 {
     Wh_Log(L"Uninit");
+
+    if (g_hbmArrowDown)
+        DeleteObject(g_hbmArrowDown);
+
+    if (g_hbmArrowUp)
+        DeleteObject(g_hbmArrowUp);
 
     if (g_hmodComdlg32_7)
         FreeLibrary(g_hmodComdlg32_7);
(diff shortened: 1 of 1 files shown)


mods/win7-file-open-save-toggle-bar-icon.wh.cpp
```diff
@@ -0,0 +1,228 @@
...
+        *py = MulDiv(*py, g_iLPY, 96);
+}
+
+void (__thiscall *CFileOpenSave__ScaleAndSetToggleBarImageListIfNeeded_orig)(class CFileOpenSave *pThis);
+
+class CFileOpenSave
+{
+public:
+    // The current offsets are applicable for 19041.1806, and reportedly works
+    // on other 19041 variants. Support for other builds will be added later.
```
m417z: Will it just crash on other versions? That's unacceptable, especially since this mod targets all processes.

Leymonaide: It probably would, but I don't have the offsets for other versions of Windows. I can block untested versions, but I thought a warning would suffice, since it would only crash once the save dialog is opened.

m417z:
> I don't have the offsets for other versions of Windows

Instead of trying to collect offsets, how about removing the need for it in the first place?

I can think of two options:

1. Hook `CreateWindowExW` or a similar function, and capture the window handle there. Optionally, use another hook to know when to expect it. For example (a guess, untested):

```cpp
thread_local void* g_pToggleBarThis;

void __fastcall SetToggleBar_Hook(void* pThis) {
    g_pToggleBarThis = pThis;
    SetToggleBar_Orig(pThis);
    g_pToggleBarThis = nullptr;
}

HWND WINAPI CreateWindowExW_Hook(/*...*/, LPCWSTR lpClassName, /*...*/) {
    HWND hWnd = CreateWindowExW_Orig(/*...*/);
    if (hWnd && g_pToggleBarThis && IS_INTRESOURCE(lpClassName) == false &&
        _wcsicmp(lpClassName, L"ToolbarWindow32") == 0) {
        // hWnd is the toggle bar for instance g_pToggleBarThis
    }
    return hWnd;
}
```

2. Use the assembly of `ScaleAndSetToggleBarImageListIfNeeded`, it contains the offset at the beginning. For example, on my Win11:

```
    comdlg32!CFileOpenSave::ScaleAndSetToggleBarImageListIfNeeded: CFG
00000001`80036a44 4055             push    rbp
00000001`80036a46 53               push    rbx
00000001`80036a47 56               push    rsi
00000001`80036a48 57               push    rdi
00000001`80036a49 488bec           mov     rbp, rsp
00000001`80036a4c 4883ec48         sub     rsp, 48h
00000001`80036a50 488bf1           mov     rsi, rcx
00000001`80036a53 488b8948020000   mov     rcx, qword ptr [rcx+248h]
```
`0x248` is the offset.

You can look for mods that use `Wh_Disasm` for examples.

This will have to be done for every architecture - x86, x64, ARM64 - so option 1 is easier.

What do you think?

Leymonaide:
I went with the first approach.

m417z:
### Submission review

---

**1. Dead version-check code — and confirming the crash concern is resolved.**
The earlier review thread asked whether the mod would crash on untested Windows builds. The new design already handles that gracefully: the two `CFileOpenSave` symbols are resolved with `WindhawkUtils::HookSymbols`, so on a build where they don't resolve, `HookSymbols` returns `false`, `Wh_ModInit` returns `FALSE`, and the mod simply doesn't load — no crash. And the replacement body only sends standard `TB_*` messages, never touching version-specific offsets. So the concern is addressed by the redesign.

What's left over is the version probe that no longer does anything:
```cpp
RTL_OSVERSIONINFOW osvi;
RtlGetVersion(&osvi);   // osvi.dwOSVersionInfoSize unset, result never read
```
`osvi` is uninitialized and the returned version is never used. Please remove this, the `EXTERN_C ... RtlGetVersion` declaration, and the `-lntdll` it requires. (If you do still want an explicit build check, it needs to actually gate something — but with symbol hooks it's redundant.)

**2. `@include notepad.exe` is very narrow for what the mod describes.**
The readme/title speak of "open/save dialogs" in general, but the common item dialog (`CFileOpenSave` in comdlg32) shows up in a large number of apps, not just Notepad. The closely analogous [`win9x-nt4-open-save-dialog`](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/win9x-nt4-open-save-dialog.wh.cpp#L8) mod uses `@include *` for exactly this reason. I understand the narrowing was partly a reaction to the "targets all processes" worry — but since the symbol-hook redesign is now crash-safe regardless of scope, the `notepad.exe`-only restriction mostly just limits the mod's usefulness. Please reconsider: either broaden to the set of apps where you want the effect (or `*` like the analog), or, if Notepad-only is genuinely intended, make the description say so.

**3. GDI bitmap leak across reloads.**
`g_hbmArrowDown` / `g_hbmArrowUp` are created with `LoadBitmapW` but never freed. Because `Wh_ModSettingsChanged` always sets `*pbReload = TRUE`, every settings change re-runs `Wh_ModInit` and leaks two more GDI bitmaps — and they're never released on a plain disable either. `LoadBitmap` returns a caller-owned object, so `TB_ADDBITMAP` doesn't take ownership. Free them in `Wh_ModUninit`:
```cpp
void Wh_ModUninit()
{
    if (g_hbmArrowDown) DeleteObject(g_hbmArrowDown);
    if (g_hbmArrowUp) DeleteObject(g_hbmArrowUp);
    if (g_hmodComdlg32_7) FreeLibrary(g_hmodComdlg32_7);
}
```

<p>

- **Use the type-safe hook wrapper** for `CreateWindowExW` instead of raw `Wh_SetFunctionHook` with `void*` casts:
  ```cpp
  WindhawkUtils::SetFunctionHook(CreateWindowExW, CreateWindowExW_hook, &CreateWindowExW_orig);
  ```
- **`GetDpiForWindow` lookup**: you `LoadLibraryExW(L"user32.dll", ...)`, take the proc address, then `FreeLibrary` it and use the pointer afterward. It's safe in practice (user32 is a KnownDLL that's always loaded in a GUI process, so the pointer stays valid), but it's tidier to just use the already-loaded handle and not free it:
  ```cpp
  pfnGetDpiForWindow = (decltype(pfnGetDpiForWindow))
      GetProcAddress(GetModuleHandleW(L"user32.dll"), "GetDpiForWindow");
  ```
- **Unused link libraries**: the mod never calls a comdlg32 API directly (it's loaded dynamically and hooked), so `-lcomdlg32` looks unused; `-lntdll` goes away with the dead `RtlGetVersion` above.
- **Dead variable**: `g_fHighDPI` is assigned in `InitDPI` but never read.

</p>
</details>

<p>

- **The captured toolbar handle is a global, re-used loosely.** `g_hwndCurrentToggleBar` is set whenever a `ToolbarWindow32` is created during `_SetToggleBar`, and `ScaleAndSetToggleBarImageListIfNeeded` reads whatever was last stored. This works as long as `ScaleAndSet…` always runs right after a `_SetToggleBar` for the same dialog. If `ScaleAndSet…` is ever invoked without a fresh `_SetToggleBar` (e.g. a re-scale path), it would operate on a stale handle. In the worst case the `SendMessage`s just no-op on a dead/foreign window (no crash), but reading the toolbar from the dialog object/state at the point of use would be more robust. Low priority given you're deliberately avoiding offsets.
- **Capture relies on the toggle bar being created via `CreateWindowExW`.** If a future build (or a differently-behaving app, once the scope broadens) creates the toolbar through a path your hook doesn't see, the handle stays null and `ScaleAndSet…` quietly logs "hwndToggleBar is null" and does nothing — graceful, but the feature silently won't apply. The analogous mod hooks `CreateToolbarEx` for the same purpose; worth keeping in mind if you broaden `@include`.
- **`LoadBitmapW` result isn't checked before use.** If the user points the setting at a DLL without resources `0x241`/`0x242`, the bitmaps come back null and get added as empty — the button just shows nothing, with no indication why. A null check / log (or failing init) would make a misconfigured path more obvious.
- **High-DPI**: as your own comment notes, the Win7 bitmap has no high-DPI variants, so it renders low-res when scaled up. Just an inherent limitation to keep documented.

</p>
</details>

Leymonaide:
The latest commit addresses most of these concerns.

PR #4457, opened 2026-06-16
No Properties Icon v1.0.3

tetawaves:
## Changelog

* Fix property sheet window showing on the taskbar when focusing on child windows

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/no-properties-icon.wh.cpp
+++ b/mods/no-properties-icon.wh.cpp
@@ -2,7 +2,7 @@
 // @id              no-properties-icon
 // @name            No Properties Icon
 // @description     Removes the window icon from property sheets like previous Windows versions
-// @version         1.0.2
+// @version         1.0.3
 // @author          xalejandro
 // @github          https://github.com/tetawaves
 // @include         *
@@ -202,6 +202,17 @@ BOOL Wh_ModInit(void)
             _SetPropertySheetIcon_hook,
             false
         },
+          {
+            {
+            #ifdef _WIN64
+            L"__int64 __cdecl StubWndProc(struct HWND__ *,unsigned int,unsigned __int64,__int64)"
+            #else
+            L"long __stdcall StubWndProc(struct HWND__ *,unsigned int,unsigned int,long)"
+            #endif
+            },
+            &WndProc_orig,
+            WndProc_hook
+        }
     };
 
     WindhawkUtils::SYMBOL_HOOK comCtl32Dllhook = {
--- a/mods/no-properties-icon.wh.cpp
+++ b/mods/no-properties-icon.wh.cpp
@@ -154,12 +154,29 @@ INT_PTR CALLBACK PropSheetDlgProc_hook(
     return PropSheetDlgProc_orig(hwnd, uMsg, wParam, lParam);
 }
 
-WNDPROC WndProc_orig;
-LRESULT CALLBACK WndProc_hook(HWND hwnd, UINT uMsg, WPARAM wParam, LPARAM lParam)
+WNDPROC Shell32_WndProc_orig;
+WNDPROC Sysdm_WndProc_orig;
+WNDPROC RunDLL32_WndProc_orig;
+
+LRESULT CALLBACK Shell32_WndProc_hook(HWND hwnd, UINT uMsg, WPARAM wParam, LPARAM lParam)
 {
     if (uMsg == WM_ACTIVATEAPP)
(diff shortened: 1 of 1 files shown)


m417z:
Hooking `WndProc` using the same orig pointer more than once like this is not supported. If it works, it's probably a lucky coincidence. Please have a different orig pointer and hook function per hook.

m417z:
Below is an AI review with a more detailed explanation and other notes.

### Submission review

---

**1. The new `shell32` `StubWndProc` hook is required — if the symbol ever fails to resolve, the entire mod stops loading in every process.**

The new entry omits the 4th (`optional`) argument, so it defaults to `false` (required):

```cpp
{
    {
    #ifdef _WIN64
    L"__int64 __cdecl StubWndProc(struct HWND__ *,unsigned int,unsigned __int64,__int64)"
    #else
    L"long __stdcall StubWndProc(struct HWND__ *,unsigned int,unsigned int,long)"
    #endif
    },
    &WndProc_orig,
    WndProc_hook   // <-- no optional flag → required
}
```

Because it's in `shell32DllHooks`, which is hooked via the fatal path:

```cpp
if (!WindhawkUtils::HookSymbols(hShell32, shell32DllHooks, ARRAYSIZE(shell32DllHooks))) {
    Wh_Log(L"Failed to hook shell32.dll");
    return FALSE;
}
```

…a single missing/renamed `StubWndProc` symbol on any Windows build makes `Wh_ModInit` return `FALSE` — and with `@include *`, that's *every* process. The icon-removal feature (the mod's primary purpose, driven by `_SetPropertySheetIcon`) would stop working too, even though it's unrelated. Note the existing `sysdmCplHook`/`rundll32ExeHook` for the same `WndProc_hook` are deliberately non-fatal (their `HookSymbols` failures are only logged), so this new entry is the odd one out. Mark it optional so its absence degrades gracefully:

```cpp
{
    { ... StubWndProc ... },
    &WndProc_orig,
    WndProc_hook,
    true   // optional — don't fail the whole mod if shell32 lacks this symbol
}
```

**2. `WndProc_orig` is a single global shared by three distinct hooks across three modules — the later hook overwrites the earlier original.**

`&WndProc_orig` is now used as the original-function pointer for:

- `shell32DllHooks[1]` → shell32's `StubWndProc` (new in this PR, applied in every process)
- `sysdmCplHook` → sysdm.cpl's `StubWndProc`
- `rundll32ExeHook` → rundll32.exe's `WndProc`

These are three different compiled functions. Each `HookSymbols` call writes into the same `WndProc_orig`, so in any process where more than one resolves, all but the last-hooked end up calling the *wrong* original — wrong behavior or a crash. This is reachable: enable the mod while **System Properties** is open (rundll32 hosting sysdm.cpl, with shell32 loaded). All three hooks install; `WndProc_orig` retains only the last write, so the others dispatch through a foreign window procedure.

This was already latent between `sysdmCplHook` and `rundll32ExeHook` in v1.0.2, but the PR adds a third consumer, so it's worth fixing while you're here. Give each hooked function its own `*_orig` pointer (and a thin per-module hook that forwards to it), e.g.:

```cpp
WNDPROC StubWndProc_shell32_orig;
WNDPROC StubWndProc_sysdm_orig;
WNDPROC WndProc_rundll32_orig;
```

Each hook body is identical (swallow `WM_ACTIVATEAPP`, else forward), so they can be trivial wrappers — or a single templated helper parameterized on the orig pointer — but the original they call must match the function they were installed on.

<p>

- **`WndProc_hook` swallows `WM_ACTIVATEAPP` for *every* window backed by `shell32`'s `StubWndProc`, not just property sheets.** `StubWndProc` is a generic shell helper window procedure, so the hook fires for any shell window that uses it — and it unconditionally returns `DefWindowProcW` for `WM_ACTIVATEAPP` without checking the window class or whether it's actually a property sheet. If other shell windows rely on `StubWndProc`'s `WM_ACTIVATEAPP` handling, this could have side effects beyond the intended property-sheet fix. Worth confirming the blast radius is limited to property sheets; if not, consider gating the suppression (e.g. by class name or by tracking the property-sheet HWNDs you care about).

</p>
</details>

tetawaves:
> Hooking `WndProc` using the same orig pointer more than once like this is not supported. If it works, it's probably a lucky coincidence. Please have a different orig pointer and hook function per hook.

done

PR #4464, opened 2026-06-17
Add VS Code dark non-client frame mod

v1b3s0:
Adds a new mod: VS Code Dark Non-client Frame.

This mod forces Visual Studio Code’s Windows non-client frame into immersive dark mode. It fixes a visible 2px white bottom edge that can appear when VS Code is maximized with the custom title bar in Windows light mode, especially with a hidden or transparent taskbar.

## Changelog

* N/A - this pull request introduces a new mod.

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [x] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/vscode-dark-nonclient-frame.wh.cpp
+++ b/mods/vscode-dark-nonclient-frame.wh.cpp
@@ -20,13 +20,15 @@ This fixes the 2px light gray bottom edge that can appear when VS Code is maximi
 
 Before:
 
-![Before](https://i.imgur.com/ldT3xJt.png)
+![Before](https://i.imgur.com/DVrOPag.png)
 
 After:
 
-![After](https://i.imgur.com/DVrOPag.png)
+![After](https://i.imgur.com/ldT3xJt.png)
 
-The mod targets only `Code.exe`.
+The mod targets only `Code.exe` by default. Users of VS Code Insiders, VSCodium, or renamed/portable builds can add the relevant executable name to the mod's custom include list.
+
+When the mod is disabled, it stops overriding VS Code's frame mode. Already affected windows may keep their current frame mode until VS Code applies a new value, for example after a theme or window state change.
 */
 // ==/WindhawkModReadme==
 
@@ -36,9 +38,6 @@ The mod targets only `Code.exe`.
 using DwmSetWindowAttribute_t = decltype(&DwmSetWindowAttribute);
 DwmSetWindowAttribute_t DwmSetWindowAttribute_orig;
 
-HWINEVENTHOOK g_objectCreateHook = nullptr;
-HWINEVENTHOOK g_objectShowHook = nullptr;
-
 BOOL IsCurrentProcessWindow(HWND hWnd) {
     DWORD windowPid = 0;
     GetWindowThreadProcessId(hWnd, &windowPid);
@@ -47,14 +46,14 @@ BOOL IsCurrentProcessWindow(HWND hWnd) {
 }
 
 BOOL IsValidWindow(HWND hWnd) {
-    LONG_PTR dwStyle = GetWindowLongPtr(hWnd, GWL_STYLE);
+    LONG_PTR style = GetWindowLongPtr(hWnd, GWL_STYLE);
 
-    return (dwStyle & WS_THICKFRAME) == WS_THICKFRAME ||
-           (dwStyle & WS_CAPTION) == WS_CAPTION;
+    return (style & WS_THICKFRAME) == WS_THICKFRAME ||
+           (style & WS_CAPTION) == WS_CAPTION;
 }
 
 void ApplyDarkFrame(HWND hWnd) {
-    if (!IsCurrentProcessWindow(hWnd) || !IsValidWindow(hWnd)) {
+    if (!IsValidWindow(hWnd)) {
         return;
     }
 
@@ -96,32 +95,16 @@ HRESULT WINAPI DwmSetWindowAttribute_hook(
 }
 
 BOOL CALLBACK EnableEnumWindowsCallback(HWND hWnd, LPARAM lParam) {
-    DWORD pid = lParam;
+    DWORD targetPid = static_cast<DWORD>(lParam);
 
     DWORD windowPid = 0;
     GetWindowThreadProcessId(hWnd, &windowPid);
 
-    if (pid == windowPid) {
+    if (targetPid == windowPid) {
         ApplyDarkFrame(hWnd);
     }
 
     return TRUE;
-}
-
-void CALLBACK WinEventProc(
-    HWINEVENTHOOK,
-    DWORD,
-    HWND hWnd,
-    LONG idObject,
-    LONG idChild,
-    DWORD,
-    DWORD
-) {
-    if (idObject != OBJID_WINDOW || idChild != CHILDID_SELF) {
-        return;
-    }
-
-    ApplyDarkFrame(hWnd);
 }
 
 BOOL Wh_ModInit() {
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

---

- **Replace the 1-second polling thread with an event-driven approach.** The worker thread re-runs `EnumWindows` + `DwmSetWindowAttribute(TRUE)` every `REAPPLY_INTERVAL_MS` (1000 ms) for the whole lifetime of the process. Two problems with that:

- The DWM immersive-dark-mode attribute is **persistent** — once set on a window it stays set, so re-applying it to the *same* window every second is pure redundancy. The only thing the timer actually buys you is catching *new* windows (and re-maximize) — and it catches them up to a full second late.
- That latency is user-visible: the mod injects at process start, so when VS Code creates its main window the timer applies the fix only on the next tick → the light 2px edge (exactly the artifact this mod exists to remove) flashes for up to ~1 s on every launch, and again after each event that resets the frame ("remaximized", per your comment). A polling fix can't beat the thing it's racing.

The established pattern for this is to hook `DwmSetWindowAttribute` and react to the window setting its own attributes, plus a one-time `EnumWindows` for already-open windows at init and for restoring at uninit — no timer, no flash. See [`win11-accent-border.wh.cpp`](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/win11-accent-border.wh.cpp#L70-L78) (the `DwmSetWindowAttribute_hook`, plus `EnumWindows` in [`Wh_ModAfterInit`/`Wh_ModBeforeUninit`](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/win11-accent-border.wh.cpp#L231-L240)). Concretely:

- If VS Code/Electron itself calls `DwmSetWindowAttribute(..., DWMWA_USE_IMMERSIVE_DARK_MODE, ...)` (the "remaximized" behavior strongly suggests it re-applies a light value on certain events), hooking it and forcing the value is the **root-cause fix**: it fires exactly when needed, with zero flash and zero polling, and it makes un-installing trivial (stop overriding → VS Code's own value applies again on the next call).
- If some windows never get an explicit call, also catch creation via `SetWinEventHook(EVENT_OBJECT_SHOW/CREATE)` filtered to `GetCurrentProcessId()`, and apply once.

Either way you drop the worker thread, the stop event, and the per-second `EnumWindows` entirely.

- **Add a before/after screenshot to the README.** This is a small but visible cosmetic fix; a maximized-window screenshot showing the light edge gone makes the mod much easier to evaluate (`i.imgur.com` or `raw.githubusercontent.com` are the allowed image hosts).

<p>

- **Use the SDK constant instead of a hand-rolled one.** Modern `dwmapi.h` (Clang 20 / mingw-w64) defines `DWMWA_USE_IMMERSIVE_DARK_MODE` (value 20); you can drop `DWMWA_USE_IMMERSIVE_DARK_MODE_VALUE` and use the real enum, which documents intent and tracks the header.
- **Narrow which windows you touch.** `EnumWindowsProc` applies to every visible top-level window in the process; restricting to windows that actually have a frame (`WS_CAPTION` / `WS_THICKFRAME`, as `win11-accent-border`'s `IsValidWindow` does) avoids no-op calls on transient/popup windows. Harmless either way, just tidier.

</p>
</details>

<p>

- **Restore-on-unload forces light, it doesn't truly "unset".** `ApplyDarkFrameToCurrentProcessWindows(FALSE)` in `Wh_ModBeforeUninit` forces `DWMWA_USE_IMMERSIVE_DARK_MODE = FALSE` (light frame). There's no API to un-set the attribute, so for a user running VS Code in a *dark* theme (where VS Code itself would want a dark frame), disabling the mod can leave the frame light until VS Code next re-applies its own value. The `DwmSetWindowAttribute`-hook approach above sidesteps this — once the hook stops overriding, VS Code's intended value naturally takes effect.
- **The mod always forces dark regardless of theme.** That matches the stated purpose, but if you wanted to make it general you could expose a boolean setting (default on) so users on a light VS Code theme who don't want a dark frame can opt out. Not required.

</p>
</details>

m417z:
### Submission review

---

The previous round of feedback (drop the polling thread, hook `DwmSetWindowAttribute`, add screenshots) has been addressed well — the `DwmSetWindowAttribute` hook is the right root-cause fix and the README now has before/after images. One real problem remains in the new code.

- **The `EVENT_OBJECT_CREATE` / `EVENT_OBJECT_SHOW` hooks never fire, so `WinEventProc` is dead code.** `WINEVENT_OUTOFCONTEXT` events are delivered to the message queue of the *thread that called `SetWinEventHook`*, and that thread has to pump messages for the callback to run. You install both hooks in `Wh_ModInit`, which runs on Windhawk's loader/init thread — that thread does **not** run a message loop, so the queued events are never dispatched and `ApplyDarkFrame` is never invoked from this path. (`UnhookWinEvent` in `Wh_ModBeforeUninit` has the same thread-affinity caveat — it should be called from the same thread that registered the hook.) This is exactly the situation called out verbatim in [`explorer-nav-dragover-fix.wh.cpp`](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/explorer-nav-dragover-fix.wh.cpp#L1921-L1925): *"WINEVENT_OUTOFCONTEXT delivers callbacks to the thread that installed the hook; that thread must keep pumping messages. … Windhawk's loader/init thread does not, so we host the hook here on a dedicated thread with its own message loop."*

  Two ways to resolve it, in order of preference:
  1. **Just delete the WinEvent code.** Your `DwmSetWindowAttribute` hook already catches every window for which VS Code/Electron sets the dark-mode attribute itself (which is how Electron drives the immersive titlebar), and the one-time `EnumWindows` in `Wh_ModAfterInit` covers windows that already exist when the mod loads. For VS Code that combination almost certainly covers everything, which would make the WinEvent hooks redundant even if they *did* fire. Test with the function hook + `EnumWindows` only; if frames still come up light on some window, go to option 2.
  2. **Move the hooks onto a dedicated thread with a message loop** (and unhook on that same thread), following [`taskbar-thumbnail-hover-fix.wh.cpp`](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/taskbar-thumbnail-hover-fix.wh.cpp#L151-L190) or `explorer-nav-dragover-fix`. If you keep this path, `EVENT_OBJECT_SHOW` alone is enough — registering both `CREATE` and `SHOW` just makes `ApplyDarkFrame` run twice per window for no benefit.

<p>

- **Redundant `IsCurrentProcessWindow` checks.** The `EnumWindows` callback already filters by PID before calling `ApplyDarkFrame`, and a PID-filtered `SetWinEventHook` would too, so the `IsCurrentProcessWindow` call inside `ApplyDarkFrame` is only meaningful on the `DwmSetWindowAttribute_hook` path. Harmless, just slightly redundant.

</p>
</details>

<p>

- **Scope is `Code.exe` only.** VS Code Insiders (`Code - Insiders.exe`), VSCodium (`VSCodium.exe`), and portable builds with renamed executables won't be covered. That's a reasonable default; you might mention in the README that users of those can add the executable to the mod's custom include list.
- **Disabling doesn't instantly revert already-affected windows.** As your comment notes, the mod deliberately stops overriding rather than forcing the frame light on unload — which is the right call. Just be aware the practical effect is that a window already pushed to a dark frame stays dark until VS Code next re-applies its own value (e.g. on a theme change), since there's no API to "unset" the attribute.
- **`DWMWA_USE_IMMERSIVE_DARK_MODE` value differs on old builds.** The header constant is `20`; on Windows 10 builds older than 18985 the attribute was `19`. On those (now end-of-life) builds the proactive `EnumWindows` apply would be a no-op, though VS Code's own calls — intercepted by the function hook — would still carry whatever value it uses. Not worth special-casing.

</p>
</details>

v1b3s0:
Thanks, updated. I removed the WinEvent hook code entirely and kept the implementation to the DwmSetWindowAttribute hook plus the one-time EnumWindows pass in Wh_ModAfterInit. I also kept the README note about existing windows not instantly reverting on disable.

PR #4470, opened 2026-06-17
Common Controls Hook v1.0.2

m417z:
* Common Controls v6 activation is now skipped if already activated, improving stability.


mohsinhasanpc:
Hello? Can you view my PR on taskbar guide? 
I've updated the liquid glass theme & I'm the new author now 
Took permission from previous author

PR #4472, opened 2026-06-17
Update Hover Text Magnifier to v1.3.3

insane66613:
## Summary
- bump Hover Text Magnifier to v1.3.3
- add shell32 linkage for CommandLineToArgvW
- pass immutable UIA request snapshots instead of reading live settings from the UIA thread
- reject stale UIA results by sequence, point, settings generation, and age
- remove the unload-time worker detach path and join the worker to avoid unloading the DLL under a live thread
- reduce repeated magnifier filter setup and non-erasing repaint invalidation

## Validation
- \git diff --check\
- \py -3 scripts\\compile_mod.py -w 'C:\\Program Files\\Windhawk' -f mods\\hover-text-magnifier.wh.cpp -o32 out\\hover-text-magnifier-x86.dll -o64 out\\hover-text-magnifier-x64.dll -oarm64 out\\hover-text-magnifier-arm64.dll\
- compiled successfully for x86, x86-64, and arm64

## Notes
Local PR validation script could not fully run because the metadata fetch from mods.windhawk.net returned HTTP 403 in this environment.

### DIFF
--- a/changelog_override/hover-text-magnifier/1.3.4.md
+++ b/changelog_override/hover-text-magnifier/1.3.4.md
@@ -1,3 +1,0 @@
-- Bound unload shutdown by detaching the worker if the tool process is exiting and UIA is wedged.
-- Hardened System32 DLL loading for DPI and magnification helpers.
-- Removed ineffective C++ exception wrapper around UIA calls.
--- a/changelog_override/hover-text-magnifier/1.3.4.md
+++ b/changelog_override/hover-text-magnifier/1.3.4.md
@@ -1,0 +1,3 @@
+- Bound unload shutdown by detaching the worker if the tool process is exiting and UIA is wedged.
+- Hardened System32 DLL loading for DPI and magnification helpers.
+- Removed ineffective C++ exception wrapper around UIA calls.
--- a/mods/hover-text-magnifier.wh.cpp
+++ b/mods/hover-text-magnifier.wh.cpp
@@ -2,7 +2,7 @@
 // @id              hover-text-magnifier
 // @name            Hover Text Magnifier (macOS-style)
 // @description     On-cursor hover bubble with large text via UI Automation; optional pixel magnifier fallback.
-// @version         1.3.3
+// @version         1.3.4
 // @author          Math Shamenson
 // @github          https://github.com/insane66613
 // @license         MIT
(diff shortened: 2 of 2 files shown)


m417z:
### Submission review

---

**The unconditional `worker.join()` can hang `ExitProcess(0)` indefinitely on unload.** The new `WhTool_ModUninit` waits up to 5 s for `workerFinished`, but then calls `g.worker.join()` *regardless* of whether the wait timed out:

```cpp
while (!g.workerFinished.load(std::memory_order_acquire) &&
       GetTickCount64() - start < 5000) {
    Sleep(10);
}
if (!g.workerFinished.load(std::memory_order_acquire)) {
    Wh_Log(L"WorkerThread did not report clean shutdown within timeout; joining anyway ...");
}
g.worker.join();   // <-- blocks here if the worker is wedged
```

The worker thread only sets `workerFinished` after its own `g.uiaThread.join()` returns, and the UIA thread can be parked inside a cross-process UIA call (`ElementFromPoint` / `RangeFromPoint` / `GetText`) against a hung target app. In that case the worker never finishes, the 5 s loop expires, and the subsequent `join()` blocks for as long as the UIA call is stuck — which delays the `ExitProcess(0)` that `Wh_ModUninit` issues right after. So the timeout you added doesn't actually bound anything; it just precedes an unbounded wait.

This is exactly the case the v1.3.2 detach-on-timeout path was guarding against. The commit message justifies removing it as "avoid unloading the DLL under a live thread," but that risk doesn't apply here: this is a tool mod, and `Wh_ModUninit` calls `ExitProcess(0)` immediately after `WhTool_ModUninit` returns. `ExitProcess` terminates the leftover thread before the image is unmapped, so a detached worker can't touch unmapped code/state. Recommend keeping the longer grace period but detaching (or just returning) on timeout, so a wedged UIA call can't block process exit:

```cpp
if (g.worker.joinable()) {
    const ULONGLONG start = GetTickCount64();
    while (!g.workerFinished.load(std::memory_order_acquire) &&
           GetTickCount64() - start < 5000) {
        Sleep(10);
    }

    if (g.workerFinished.load(std::memory_order_acquire)) {
        g.worker.join();
    } else {
        Wh_Log(L"WorkerThread did not report clean shutdown; detaching (process is exiting).");
        g.worker.detach();
    }
}
```

<p>

- **`try { ... } catch (...)` around `TryExtractTextAtPoint` does little.** UIA reports failures via `HRESULT`, not C++ exceptions, so the catch almost never fires; and under the default exception model on Clang/mingw it won't intercept an SEH access violation from a misbehaving UIA provider either. It's harmless, but if the intent was to harden against a UIA-side crash, this doesn't accomplish it.

- **Harden the bare-name `LoadLibraryW` calls** (pre-existing, not in this diff): `LoadLibraryW(L"Shcore.dll")` (line ~524) and `LoadLibraryW(L"Magnification.dll")` (line ~813). Neither is a KnownDLL, so the default search order checks the executable's own directory first. The risk is low here because the host is `windhawk.exe` (normally in a protected directory), but since both are System32 DLLs you can close the gap cheaply with `LoadLibraryExW(L"...", nullptr, LOAD_LIBRARY_SEARCH_SYSTEM32)`.

</p>
</details>

<p>

- **The stricter freshness gate may delay first appearance / fast-cursor text.** A result is now accepted only if `result.seq == g.lastUiaRequestSeq` **and** within 250 ms **and** within 12 px of the current cursor. Since a new request is dispatched every `uiaQueryMinIntervalMs` (60 ms) and tied to the newest seq, on the very first hover (before `showingText` is set, so the 800 ms hysteresis doesn't apply yet) and while the cursor is moving faster than ~12 px/query, the matching result can be rejected, so text may take an extra UIA cycle to surface. This is the intended staleness rejection and is fine — just flagging that the 12 px / 250 ms thresholds are hard-coded, so if testing shows the bubble feels laggier than 1.3.2 when reading while moving, those are the knobs to relax.

</p>
</details>

PR #4485, opened 2026-06-18
Add Taskbar Folder Menus v0.7

sb4ssman:
## Summary

Adds **Taskbar Folder Menus v0.7** (`taskbar-folder-menus`) for Windows 11.
It adds compact system-tray buttons that open configured folders, drives, and
Shell namespace targets as native cascading menus, recreating the useful part
of classic taskbar toolbars without requiring users to minimize their windows.

## Features

- Configurable `folders` array with a short label and target for each button.
- Normal paths, drive roots, environment variables, and Shell namespace roots
  such as `shell:Desktop` and `shell:ControlPanelFolder`.
- Native Shell popup menus with icons, lazy subfolder loading, and an
  **Open in Explorer** command for the configured root and every subfolder.
- Full classic Shell context menus on right-click, including owner-drawn and
  dynamic extensions through `IContextMenu2`/`IContextMenu3`.
- Canonical Smart Grid layout with smart automatic, single row, single column,
  fixed rows, fixed columns, and fixed-grid modes; row/column fill order;
  first/last short groups; and short-group alignment.
- Five tray placement anchors, configurable button geometry, native or custom
  button colors, Windows accent tokens, borders, opacity, shine, and group
  padding/offsets.
- Optional limits for menu size, subfolder depth, and hidden/system items.

## Review and reliability fixes

- Uses documented `WM_MENURBUTTONUP` handling for right-clicked menu items.
- Suppresses unsafe process-shutdown destruction of namespace-scope XAML
  holders and releases delegates, tooltips, and boxed content before removing
  the injected subtree.
- Returns `FALSE` when required taskbar symbol hooks fail and clearly documents
  that the mod supports Windows 11 only.
- Keeps the requested `TrayUI::StartTaskbar` trigger. After live testing found
  that the simpler startup path could still miss the tray after a restart,
  v0.7 adds the standard three-module `IconView.Loaded` trigger and a bounded
  slow-sign-in retry fallback for deterministic reapplication.
- Replaces the old mixed-unit taskbar-window row calculation with the shared
  Smart Grid template and the live `SystemTrayFrameGrid` height in DIPs,
  including correct inter-row spacing.
- Repairs native tray-column collisions after recreation and resets stale XAML
  pointer-over state after a modal native menu closes.

## Validation

- [x] Default Desktop and Control Panel buttons load and open correctly.
- [x] Folder, drive-root, environment-variable, Desktop, and Control Panel
      targets work with Shell icons and lazy submenus.
- [x] Root and nested **Open in Explorer** commands work.
- [x] Classic right-click context menus, ordinary verbs, and modal Properties
      work; click-away dismissal releases the taskbar-button visual state.
- [x] Smart Grid layout and automatic row capacity look correct in live use.
- [x] Settings reload works and the mod survives a taskbar/Explorer restart.
- [x] Disable/unload cleanup completes without the prior Explorer crash.
- [x] Windhawk Clang syntax compilation passes.
- [x] Folder README and embedded README match, and all six screenshot links
      point to current assets.

## Mod authorship

This mod was created by:

- [ ] Claude
- [ ] ChatGPT
- [ ] Gemini
- [ ] Other (please specify):

Please select the options that best apply. Your selection does not affect the
acceptance criteria, but it helps reviewers understand the context of the code
and provide relevant feedback.

### DIFF
--- a/mods/taskbar-folder-menus.wh.cpp
+++ b/mods/taskbar-folder-menus.wh.cpp
@@ -1,0 +1,2529 @@
+// ==WindhawkMod==
+// @id              taskbar-folder-menus
+// @name            Taskbar Folder Menus
+// @description     Adds compact Windows 11 taskbar buttons that open configured Shell targets as popup menus, similar to classic taskbar toolbars.
+// @version         0.7
+// @author          sb4ssman
+// @github          https://github.com/sb4ssman
+// @include         explorer.exe
+// @architecture    x86-64
+// @compilerOptions -lole32 -loleaut32 -lruntimeobject -lshell32 -luuid -lgdi32 -lcomctl32 -lversion
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Taskbar Folder Menus
+
+A Windows 11-only mod that adds compact taskbar buttons which open Shell targets
+as native popup menus, recreating the most useful part of the classic taskbar
+toolbar workflow. Windows 10 is not supported.
+
+![Two folder buttons in the system tray](https://raw.githubusercontent.com/sb4ssman/Windhawk-Mod-Lab/main/taskbar-folder-menus/assets/desktop-controlpanel.png)
+*A minimal two-button setup for Desktop and Control Panel.*
+
+![Folder button destination tooltip](https://raw.githubusercontent.com/sb4ssman/Windhawk-Mod-Lab/main/taskbar-folder-menus/assets/tooltip-shows-destination.png)
+*Hovering a compact button shows its configured Shell target.*
+
+![Four folder buttons on a standard taskbar](https://raw.githubusercontent.com/sb4ssman/Windhawk-Mod-Lab/main/taskbar-folder-menus/assets/c-github-desktop-controlpanel.png)
+*Drive, GitHub, Desktop, and Control Panel shortcuts arranged in a grid.*
+
+![Four folder buttons on a taller taskbar](https://raw.githubusercontent.com/sb4ssman/Windhawk-Mod-Lab/main/taskbar-folder-menus/assets/c-github-desktop-controlpanel-v.png)
+*The same four shortcuts arranged vertically by the grid layout on a taller taskbar.*
+
+![Control Panel opened as a native Shell menu](https://raw.githubusercontent.com/sb4ssman/Windhawk-Mod-Lab/main/taskbar-folder-menus/assets/controlpanel-menu-open.png)
+*The Control Panel namespace opens directly as a native Shell menu with full icons.*
+
+![Whole drive opened from a taskbar folder button](https://raw.githubusercontent.com/sb4ssman/Windhawk-Mod-Lab/main/taskbar-folder-menus/assets/whole-drive-on-taskbar.png)
+*A drive-root button opens the whole drive as a native cascading Shell menu.*
+
+Click a small button to browse a folder, drive, Desktop, or Control Panel
+directly from the taskbar — no minimizing required. Subfolders expand on
+hover. Right-click any item for the full classic Windows Shell context menu.
+Every folder popup includes an "Open in Explorer" shortcut at the top,
+including the configured root folder.
+
+## Example folder entries
+
+```text
+Label: 🖥    Target: shell:Desktop
+Label: ⚙     Target: shell:ControlPanelFolder
+Label: 📥    Target: %USERPROFILE%\Downloads
+Label: C:    Target: C:\
+```
+
+Add one record per button in the Folders setting. Each record has a short button
+label and a target. Targets can be normal paths or Shell namespace roots like
+`shell:Desktop` and `shell:ControlPanelFolder`. Environment variables such as
+`%USERPROFILE%` are expanded automatically. The full label and target appear in
+the tooltip.
+
+Emoji labels are a natural fit for narrow buttons. Label ideas: 📁 folder,
+🖥 desktop, 💻 laptop, 🪟 windows, 📥 downloads, 🌐 network, 🗄 drive,
+📄 documents, 🔧 tools, ⚙ settings, ⭐ favorites.
+
+## Reordering folder buttons
+
+For several existing records, open the mod's Settings page and switch to
+**Textual mode**, then move each complete `label` + `target` record as a block.
+The regular form currently provides add/remove controls but no direct
+drag-to-reorder control.
+
+## Settings
+
+| Setting | Default | Description |
+|---------|---------|-------------|
+| Position | Before notification icons | Where to inject the button group in the tray |
+| Folders | Desktop and Control Panel | Add/remove records in the form; reorder complete records in Textual mode |
+| Grid mode | Smart automatic | Smart automatic, single row/column, fixed rows/columns, or fixed grid |
+| Smart layout | Balanced | Balanced, pack vertical, or pack horizontal in smart mode |
+| Grid columns | 0 (auto) | Exact columns in fixed-column modes; maximum columns in smart mode (0 = automatic) |
+| Grid rows | 0 (auto) | Exact rows in fixed-row modes; maximum rows in smart mode (0 = tray height) |
+| Fill order | Row-first | Row-first or column-first |
+| Short row/column position | Last | Put an incomplete row or column first or last |
+| Short row/column alignment | Start | Align an incomplete final row or column |
+| Button width | 24 px | Width of each folder button |
+| Button height | 22 px | Height of each folder button |
+| Button spacing | 4 px | Gap between buttons |
+| Default button text | 📁 | Fallback icon for entries with long labels |
+| Text/icon size | 10 pt | Button label font size |
+| Text color | *(system)* | Button label color; empty keeps the system default |
+| Background color | *(system)* | Button background; empty keeps the system default |
+| Hover background color | `accent` | Hover background; empty keeps the native hover color |
+| Click background color | *(system)* | Pressed background; empty keeps the native pressed color |
+| Border color | *(system)* | Button border; empty keeps the system default |
+| Border thickness | -1 (system) | -1 = system default, 0 = no border |
+| Corner rounding | -1 (system) | -1 = system default, 0 = square |
+| Opacity | 100% | Button transparency |
+| Shine effect | Off | Gradient highlight; activates when a custom background color is set |
+| Group padding left | 0 px | Extra space to the left of the button group |
+| Group padding right | 0 px | Extra space to the right of the button group |
+| Group X offset | 0 px | Shift the entire group left or right |
+| Group Y offset | 0 px | Shift the entire group up or down |
+| Max menu items | 0 (unlimited) | Limit items shown per folder |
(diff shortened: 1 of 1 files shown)


m417z:
* I think it makes more sense for the `folders` option to be an array.
* If `HookTaskbarDllSymbols` fails, return FALSE at `Wh_ModInit`.
* Instead of handling `Taskbar.View.dll` and hooking `LoadLibraryExW`, you can hook `TrayUI::StartTaskbar` and simplify the implementation. There are several mods you can use as reference, such as `taskbar-ai-quota`.
* You might want to clarify that this is a Win11-only mod, and that it won't work on Win10.

sb4ssman:
Updated to v0.6 with all review points addressed:

- `folders` is now an array setting (per-record `label` + `target`)
- `Wh_ModInit` returns FALSE when `HookTaskbarDllSymbols` fails
- Replaced the Taskbar.View.dll handling and `LoadLibraryExW` hook with a `TrayUI::StartTaskbar` hook, following taskbar-ai-quota
- Readme and description now state this is a Windows 11-only mod

Also in v0.6:
- Fixed an Explorer crash on mod disable: Click delegates, tooltips, and boxed Content are now released before the injected subtree is removed, since XAML tears removed subtrees down on a later UI tick that can land after the mod DLL has unloaded
- Color settings now accept `accent` / `accentLight` / `accentDark` and `transparent` in addition to hex; the hover default changed from a hardcoded `#4488FF` to `accent`

m417z:
### Submission review

I haven't verified 1, let me know if it's incorrect.

---

**1. The right-click context-menu feature never fires — wrong window message.** The subclass handles a self-defined `WM_MENURBUTTONDOWN = 0x012E`:

```cpp
#ifndef WM_MENURBUTTONDOWN
#define WM_MENURBUTTONDOWN 0x012E
#endif
...
    if (msg == WM_MENURBUTTONDOWN) {
        UINT index = (UINT)wParam;
        HMENU hMenu = (HMENU)lParam;
```

`0x012E` is not a real Win32 message, and there is no `WM_MENURBUTTONDOWN`. When the user right-clicks a menu item, Windows sends **`WM_MENURBUTTONUP` (`0x0122`)** to the menu owner — and its parameters are exactly what your handler already reads (`wParam` = item index, `lParam` = `HMENU`). So the constant is just wrong; the message you want is already defined in `winuser.h`. As written, `IContextMenu` on right-click (a headline feature, and the corresponding test-plan box is still unchecked) can never trigger. Fix:

```cpp
    if (msg == WM_MENURBUTTONUP) {   // 0x0122, sent to the menu owner on menu-item right-click
        UINT index = (UINT)wParam;
        HMENU hMenu = (HMENU)lParam;
```

and drop the `#define`.

**2. Global XAML strong references are destroyed at process shutdown, off the UI thread.** `g_buttonGrid` (`Grid`), `g_injectionParent` (`FrameworkElement`), and `g_buttonEventStates` (a `vector` of `Button`) hold strong XAML references at namespace scope:

```cpp
static Grid              g_buttonGrid = nullptr;
static FrameworkElement  g_injectionParent = nullptr;
...
static std::vector<ButtonEventState> g_buttonEventStates;
```

`Wh_ModUninit` is *not* called when the host process terminates (Explorer restart, sign-out, reboot — extremely common for a taskbar mod). In that case the OS runs these globals' destructors on the shutdown thread after the XAML core is already torn down, releasing strong `FrameworkElement`s off their UI thread — the exact thread-affinity hazard described in the Windhawk cross-thread notes, and a potential crash on every Explorer restart. Your controlled-unload cleanup (`RemoveButtonGrid` / `ClearButtonEventState`) is already correct; you just need to suppress the *automatic* destructor:

```cpp
[[clang::no_destroy]] static Grid g_buttonGrid = nullptr;
[[clang::no_destroy]] static std::vector<ButtonEventState> g_buttonEventStates;
```

`g_injectionParent` is only ever assigned, never read (see lines 1581/1660) — it holds a strong ref for no functional reason, so the cleanest fix there is to just delete the variable rather than annotate it.

<p>

- **Unused dependency**: `#include <winver.h>` and `-lversion` in `@compilerOptions` — no version API (`GetFileVersionInfo`, `VerQueryValue`, …) is used in the file. Both can be removed.
- **Log-category prefixes** like `[Init]`, `[Apply]`, `[Menu]` in `Wh_Log` strings are redundant — Windhawk already prefixes each line with the mod name. Harmless, but they can go.
- **Fallback owner**: in `ShowFolderMenu`, if the taskbar window isn't found you fall back to `GetForegroundWindow()` and then `SetWindowSubclass` it. Plain `SetWindowSubclass` can't subclass a window owned by another thread, so on that (rare) path the lazy-load / right-click handlers silently won't work. Not worth much, since in `explorer.exe` the `Shell_TrayWnd` lookup essentially always succeeds — just noting it.

</p>
</details>

<p>

- **DPI units mismatch in `GetAvailableFolderRows`**: the available-row count divides `GetWindowRect(taskbar)` (physical pixels) by `buttonHeight + buttonSpacing`, but XAML sizes (`buttonHeight`, etc.) are DIPs, not physical pixels. On a non-100% display scale the two are different units, so smart/auto grid row counts come out wrong (too many rows at >100% scale). Convert the rect height to DIPs (divide by the monitor scale factor) before dividing by the button pitch.
- **Synchronous Shell enumeration on the UI thread**: `WM_INITMENUPOPUP` runs `EnumerateShellFolder` (Shell binding, `EnumObjects`, per-item `SHGetFileInfo` for icons) inline on the taskbar UI thread while the menu is up. For a very large folder this can briefly freeze the taskbar the first time a submenu opens. There's no clean alternative for a synchronous popup menu, so this is just an FYI — the lazy per-submenu loading already keeps it bounded in practice.

</p>
</details>

sb4ssman:
Updated the PR to v0.7 on a clean rebase of current `main`.

The two required review items are addressed:

- Right-click now uses the documented `WM_MENURBUTTONUP` path; the full classic
  Shell context-menu flow was live-tested, including modal Properties and
  click-away dismissal.
- Namespace-scope XAML holders use `[[clang::no_destroy]]`, the unused strong
  parent reference is gone, and delegates/tooltips/boxed content are released
  before subtree removal.

The DPI/layout note is also addressed by replacing the inline layout code with
the shared Smart Grid template and calculating row capacity from the live
`SystemTrayFrameGrid` height in DIPs.

One earlier simplification was adjusted based on live regression evidence:
`TrayUI::StartTaskbar` remains, but after the mod failed to reappear following a
restart, v0.7 also uses the standard three-module `IconView.Loaded` trigger plus
a bounded slow-sign-in retry fallback. The updated build now survives restart
in live testing.

The PR description and completed test plan now reflect the current array
settings, six grid modes, accent default, reliability fixes, and screenshot set.

m417z:
1. Replace these two usages of `clang::no_destroy`:

```cpp
[[clang::no_destroy]] static std::vector<ButtonEventState> g_buttonEventStates;

[[clang::no_destroy]] static std::list<FrameworkElement::Loaded_revoker>
    g_loadedRevokers;
```

With:

```cpp
[[clang::no_destroy]] static std::optional<std::vector<ButtonEventState>> g_buttonEventStates;

[[clang::no_destroy]] static std::optional<std::list<FrameworkElement::Loaded_revoker>>
    g_loadedRevokers;
```

Then call `reset` at `Wh_ModUninit`.

The reason is that otherwise, memory will likely be leaked even if the vector/list is cleared.

I know you followed the instructions of the previous review, I'm working on improving it.

2. Regarding process startup, you handle it in 2 places:

* `TrayUI_StartTaskbar_Hook` - `StartRetryThread`
* `IconView_IconView_Hook` - `ApplyAllSettings`

The second place requires handling `SystemTray.dll`/`Taskbar.View.dll`/`ExplorerExtensions.dll` (which are also not marked correctly in `SYMBOL_HOOK`). As I noted in the other review, if this code can be removed to simplify the mod and remove code that can break in future updates, let's do it.

3. In `Wh_ModAfterInit`, don't run `StartRetryThread`, just init once for the case the mod is loaded while the taskbar already exists. In the other case, it's handled by `TrayUI_StartTaskbar_Hook`, so retrying is unnecessary. It also avoids possible races, see below.

---

**Data race on the retry-thread handles.** `StartRetryThread` / `StopRetryThread` mutate the plain (non-atomic) globals `g_retryThread` and `g_retryStopEvent`, and they are invoked from two different threads with no synchronization:

- the **taskbar UI thread**, via the `TrayUI::StartTaskbar` hook → `StartRetryThread` (line 2354), and
- the **Windhawk callback thread**, via `Wh_ModAfterInit` (line 2490) and `Wh_ModSettingsChanged` (lines 2512/2518/2528).

At boot with the mod already installed, `Wh_ModAfterInit` and `StartTaskbar` can run concurrently, so two `StartRetryThread`→`StopRetryThread` sequences race on the same two handles. Possible outcomes: `CloseHandle` called twice on the same handle (which can later close an unrelated handle), a leaked thread/event handle, or one thread reading `g_retryStopEvent` while the other closes it → crash. The same overlap can recur whenever Explorer restarts (`StartTaskbar`) around a settings change.

sb4ssman:
All three review items are addressed, plus the data race.

**1. `clang::no_destroy` containers.** `g_buttonEventStates` is now
`[[clang::no_destroy]] static std::optional<std::vector<ButtonEventState>>`,
constructed with `std::in_place` and released with `reset()` on the controlled
unload path after the UI-thread cleanup has run. `g_loadedRevokers` is gone
entirely — the remaining handlers no longer need a namespace-scope revoker
list, so there was nothing left to annotate. The only other namespace-scope
annotation left is the nullable `Grid g_buttonGrid`, which is the form you said
is fine as-is.

**2. The `IconView` startup path is removed.** The `IconView::IconView` hook is
gone, and with it the `SystemTray.dll` / `Taskbar.View.dll` /
`ExplorerExtensions.dll` handling and the `LoadLibraryExW` hook. The mod now
has exactly one `SYMBOL_HOOK` array, `taskbarDllHooks`, against `taskbar.dll`.
That also removes the incorrectly-marked symbol hooks you noted.

**3. `Wh_ModAfterInit` is a single apply.** It calls
`ApplyAllSettingsOnWindowThread()` once, covering the case where the mod loads
into an already-running taskbar. Taskbar startup and rebuilds are handled by
the `TrayUI::StartTaskbar` hook, which owns the bounded retry.

**Data race on the retry-thread handles.** Fixed, and thank you for the precise
diagnosis. `g_retryThread` and `g_retryStopEvent` are now only mutated while
holding an SRW lock. `StopRetryThread` detaches both handles under the lock and
then performs the wait and `CloseHandle` outside it, so two concurrent stops
cannot double-close, and a worker synchronously dispatching to the taskbar
thread cannot deadlock against a stop. With `Wh_ModAfterInit` no longer
starting a retry thread, the boot-time overlap you described is gone at the
source as well.

This build was live-tested on Windows 11 before pushing and works as described.
Locally it also passes a Windhawk compile-and-link check with the declared
`@compilerOptions`, an exit-time destructor audit, README parity between the
folder README and the embedded `==WindhawkModReadme==`, and the upstream PR
validator.

PR #4507, opened 2026-06-20
Update Taskbar Fluent Media Player 1.4.0

Salyts:
### Added
- **Real-time audio visualizer** (WASAPI loopback + FFT)
  - 5 bar shapes: Stereo, Mountain, Mirror, Wave, Breathe
  - 5 color modes: Solid, Dynamic Album, Dynamic Gradient, Custom Gradient, Acrylic
  - EQ presets: Balanced, Bass, Rock, Pop, Jazz, Electronic
  - Configurable position (left/right of player), vertical anchor (top/middle/bottom), bar count/width/gap, idle bar size, padding, and sensitivity
- **Right-click context menu on the player**: Toggle Repeat/Shuffle, Rewind/Forward 5s, Next/Previous Track, Switch Sessions, Open media app, optional Restart Player, Open Windhawk
- **"Switch Sessions" button/action** — new media button type, also available as a mouse/context-menu action
- **Custom empty album art icon** — replaces the fixed "question mark"/"music note" presets with a configurable glyph (hex code), font, size, color, and opacity
- **Per-theme color support** — almost every color field now accepts two values separated by `$` (e.g. `0 0 0$255 255 255`) for light/dark theme: background, button icons, title, artist, empty-art icon, visualizer
- Separate pause-overlay icon size control
- Dedicated second color for the gradient background (`gradientColor2`), independent from the two-tone solid color
- Debug setting to toggle the "Restart Player" entry in the context menu
- Customizable placeholder texts for the title/artist fields — separate, editable strings for "no title on the current track", "no artist on the current track", "no media session at all" (title and artist), each can be left empty to hide that field entirely in that state

### Changed
- "Spacing between buttons" and "Button icon size" moved from Main Settings to Appearance → Media Buttons Style
- Reworked player hover/press visuals — new gradient "elevation" border instead of manual brush swapping per state
- Buttons renamed for clarity: Previous → Previous Track, Next → Next Track, Shuffle → Toggle Shuffle, Repeat → Toggle Repeat
- Removed the standalone "Auto theme" toggle — replaced by the universal light$dark color-pair support across all elements
- The title/artist text container now collapses entirely (no reserved space) when both fields end up empty, instead of just hiding the individual text blocks

### DIFF
--- a/mods/taskbar-fluent-media-player.wh.cpp
+++ b/mods/taskbar-fluent-media-player.wh.cpp
@@ -3,31 +3,29 @@
 // @name            Taskbar Fluent Media Player
 // @description     Taskbar Fluent Media Player — is a Windhawk mod that integrates a modern media player with Fluent Design directly into the Windows 11 taskbar. It allows you to control music and view track information seamlessly without interrupting your workflow.
 // @description:ru-RU Taskbar Fluent Media Player — это мод Windhawk, который интегрирует современный медиаплеер в стиле Fluent Design прямо в панель задач Windows 11. Он позволяет управлять музыкой и просматривать информацию о треке без прерывания работы.
-// @version         1.3.0
+// @version         1.4.0
 // @author          Salyts
 // @github          https://github.com/Salyts
 // @include         explorer.exe
-// @compilerOptions -lole32 -loleaut32 -lruntimeobject -lversion -luuid -luser32 -lwindowsapp -lshell32 -lgdi32 -lshlwapi -lwindowscodecs -ldwmapi -lshcore
+// @compilerOptions -lole32 -loleaut32 -lruntimeobject -lversion -luuid -luser32 -lwindowsapp -lshell32 -lgdi32 -lshlwapi -lwindowscodecs -ldwmapi -lshcore -lksuser
 // ==/WindhawkMod==
 
 // ==WindhawkModReadme==
 /*
-# Taskbar Fluent Media Player 1.3.0
+# Taskbar Fluent Media Player 1.4.0
 
 **Taskbar Fluent Media Player —** is a Windhawk mod that integrates a modern media player with Fluent Design directly into the Windows 11 taskbar. It allows you to control music and view track information seamlessly without interrupting your workflow.
 
 ### [> Russian documentation <](https://github.com/Salyts/Taskbar-Fluent-Media-Player/blob/main/README_RU.md)
 
-![img](https://i.imgur.com/9xXlVyq.png)
-
-![img](https://i.imgur.com/Md5CvZB.png)
-
+![video](https://i.imgur.com/pGFnmiG.gif)
 ### Key Features:
 * **Deep Integration —** Place the player in the tray area (left or right of the clock), next to system icons, or in the center of the taskbar.
 * **Fluent UI Effects —** Full support for native Windows 11 materials including **Acrylic**, **Mica**, and **Mica Alt**.
 * **Adaptive Design —** The player can automatically adapt its colors based on the album art or follow the system accent color.
 * **Full Media Control —** Buttons for Play/Pause, Skip, 5-second Seek, Shuffle, and Repeat modes.
 * **Smart Behavior —** Automatically hides the player when no media is playing, when entering full-screen mode, or after a period of inactivity (idle).
+* **Audio Visualizer —** A real-time spectrum visualizer (WASAPI loopback + FFT) with multiple bar shapes (Stereo, Mountain, Mirror, Wave, Breathe), color modes (Solid, Dynamic Album, Gradient, Acrylic, Accent), EQ presets, vertical anchoring, and per-side placement.
 
 ### Advanced Customization:
 * **Visuals —** Customize fonts (Segoe UI, Aptos, etc.), text sizes, padding, and element transparency.
@@ -35,6 +33,10 @@
 * **Mouse Interactions —** Assign custom actions (Volume control, Track seeking, Session switching) to the mouse wheel, single clicks, or double clicks on different parts of the player.
 
 ---
+
+### Credits
+* **[Salyts](https://github.com/Salyts) —** Author of Taskbar Fluent Media Player.
+* **[GR0UD](https://github.com/GR0UD) —** Audio visualizer (capture + FFT engine) ported from his Taskbar Media Player.
 
 ### Report a Bug
 If you encounter any issues, bugs, or have suggestions for new features, please report them on the project's GitHub page:
@@ -145,9 +147,6 @@ If you encounter any issues, bugs, or have suggestions for new features, please 
     - showTrackArtist: true
       $name: Show artist name
       $name:ru-RU: Отображать имя исполнителя
-    - swapTitleArtist: false
-      $name: Swap artist name and track title
-      $name:ru-RU: Поменять местами название трека и имя исполнителя
     - textAreaWidth: "0 120"
       $name: Text area width (min max)
       $name:ru-RU: Ширина текстовой области (min max)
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

---

Nice update — the new audio visualizer, context menu, and per-theme color pairs are well integrated, and the credit for the ported visualizer engine is correctly placed in the README (single `@author`, GR0UD credited in the "Credits" section), which is the right approach. A couple of things to address, mostly around the new capture thread.

**1. The new global `std::thread g_CaptureThread` will call `std::terminate()` on host shutdown.** ([L4246](https://github.com/ramensoftware/windhawk-mods/blob/7b1ee0a223c467b7785e0869d2e37075f41dc343/mods/taskbar-fluent-media-player.wh.cpp#L4246)) The capture thread is only joined in `StopVizCaptureThread()`, which runs via `Wh_ModUninit`/`Wh_ModSettingsChanged`. But Windhawk does **not** call `Wh_ModUninit` when the host process itself terminates (explorer restart, sign-out, reboot) — the OS just detaches the DLL and the CRT runs global destructors during `DLL_PROCESS_DETACH`. At that point the OS has already killed the capture thread, but the `std::thread` object is still `joinable()` (never joined/detached), so `~thread()` calls `std::terminate()` → the host aborts instead of exiting cleanly. This happens on every clean explorer exit whenever the visualizer is enabled. Fix: use a `std::thread*` (heap-allocated, intentionally leaked at shutdown — a trivial pointer destructor), keeping the real join in `Wh_ModUninit`:

```cpp
static std::thread* g_CaptureThread = nullptr;
// start:  g_CaptureThread = new std::thread(VizCaptureThreadProc);
// stop:   if (g_CaptureThread) { if (g_CaptureThread->joinable()) g_CaptureThread->join();
//                                delete g_CaptureThread; g_CaptureThread = nullptr; }
```

**2. The visualizer never pauses while the player is hidden.** `StartVizCaptureThread()` / `StartVizTimer()` are tied to `StartTimerThread`/`StopTimerThread` and the inject hook only — nothing in `UpdateVisibility()` ([L7983](https://github.com/ramensoftware/windhawk-mods/blob/7b1ee0a223c467b7785e0869d2e37075f41dc343/mods/taskbar-fluent-media-player.wh.cpp#L7983)) stops them. So when the player is collapsed by `hideFullscreen`, idle-hide, or `hideWhenNoMedia`, the WASAPI loopback FFT thread and the 60 fps `DispatcherTimer` keep running and keep updating offscreen XAML elements. The fullscreen case is the most costly one — the visualizer competes for CPU with a game/video at exactly the moment the player is deliberately hidden. Since you already compute the hide decision in `UpdateVisibility`, gate the capture thread and viz timer on it (start when shown, stop when hidden).

<p>

- **Per-frame color parsing at 60 fps includes a registry read.** `VizApplyFrame()` calls `ParseColorWithThemeSupport(g_settings.vizColor, 255)` every frame ([L4666](https://github.com/ramensoftware/windhawk-mods/blob/7b1ee0a223c467b7785e0869d2e37075f41dc343/mods/taskbar-fluent-media-player.wh.cpp#L4666)), and the default `vizColor` (`"0 0 0$255 255 255"`) contains a `$`, so each frame hits `IsSystemLightTheme()` → `RegGetValueW`. That's ~60 registry reads/sec on the taskbar UI thread for a value that only changes on theme/settings change. You already detect theme changes in `TimerThreadProc` (via `RegNotifyChangeKeyValue`); cache the resolved base color and refresh it on theme/settings change instead of recomputing it per frame.
- **"Restart Player" (debug, default off) blocks the UI thread.** The `MenuFlyoutItem` click handler runs on the taskbar UI thread and does `RemovePlayerGrid(); std::this_thread::sleep_for(200ms); ...` ([L5398](https://github.com/ramensoftware/windhawk-mods/blob/7b1ee0a223c467b7785e0869d2e37075f41dc343/mods/taskbar-fluent-media-player.wh.cpp#L5398)) — a 200 ms freeze of the taskbar. It also doesn't stop the viz timer or clear `g_vizBars` across the rebuild, so the timer briefly drives detached rectangles (harmless thanks to the `try/catch`, but untidy). Posting the rebuild via the dispatcher instead of sleeping would avoid the stall.
- **`ParseColorWithThemeSupport`: `size_t dashPos = colorStr.find(L'$');`** ([L2473](https://github.com/ramensoftware/windhawk-mods/blob/7b1ee0a223c467b7785e0869d2e37075f41dc343/mods/taskbar-fluent-media-player.wh.cpp#L2473)) — the variable name says "dash" but it searches for `$`; rename for clarity.

</p>
</details>

<p>

Non-critical observations about behavior changes for existing users.

- **`albumArtEmptyBehavior` option rename.** The old `show_question` / `show_note` values were replaced by a single `show_icon` + glyph. Existing users who had selected one of the old values will fall through both the `hide` and `show_icon` branches in `RefreshPlayerContents`, so the empty-art area shows nothing until they re-pick "Show area with icon". Default (`show`) is unaffected, so this only hits users who customized it.
- **Gradient color remap.** `solidColor2` changed meaning from "Secondary color for gradient" to "Gradient color 1", with the new `gradientColor2` as the second stop, and its default flipped from `128 128 128` to `35 35 35`. Existing users with a tuned gradient will see their colors shift after the update. Intentional per the changelog, just worth being aware it's a visible migration.

</p>
</details>

Salyts:
> ### Submission review
> 
> 
> 
> Nice update — the new audio visualizer, context menu, and per-theme color pairs are well integrated, and the credit for the ported visualizer engine is correctly placed in the README (single `@author`, GR0UD credited in the "Credits" section), which is the right approach. A couple of things to address, mostly around the new capture thread.
> 
> **1. The new global `std::thread g_CaptureThread` will call `std::terminate()` on host shutdown.** ([L4246](https://github.com/ramensoftware/windhawk-mods/blob/7b1ee0a223c467b7785e0869d2e37075f41dc343/mods/taskbar-fluent-media-player.wh.cpp#L4246)) The capture thread is only joined in `StopVizCaptureThread()`, which runs via `Wh_ModUninit`/`Wh_ModSettingsChanged`. But Windhawk does **not** call `Wh_ModUninit` when the host process itself terminates (explorer restart, sign-out, reboot) — the OS just detaches the DLL and the CRT runs global destructors during `DLL_PROCESS_DETACH`. At that point the OS has already killed the capture thread, but the `std::thread` object is still `joinable()` (never joined/detached), so `~thread()` calls `std::terminate()` → the host aborts instead of exiting cleanly. This happens on every clean explorer exit whenever the visualizer is enabled. Fix: use a `std::thread*` (heap-allocated, intentionally leaked at shutdown — a trivial pointer destructor), keeping the real join in `Wh_ModUninit`:
> 
> ```c++
> static std::thread* g_CaptureThread = nullptr;
> // start:  g_CaptureThread = new std::thread(VizCaptureThreadProc);
> // stop:   if (g_CaptureThread) { if (g_CaptureThread->joinable()) g_CaptureThread->join();
> //                                delete g_CaptureThread; g_CaptureThread = nullptr; }
> ```
> 
> **2. The visualizer never pauses while the player is hidden.** `StartVizCaptureThread()` / `StartVizTimer()` are tied to `StartTimerThread`/`StopTimerThread` and the inject hook only — nothing in `UpdateVisibility()` ([L7983](https://github.com/ramensoftware/windhawk-mods/blob/7b1ee0a223c467b7785e0869d2e37075f41dc343/mods/taskbar-fluent-media-player.wh.cpp#L7983)) stops them. So when the player is collapsed by `hideFullscreen`, idle-hide, or `hideWhenNoMedia`, the WASAPI loopback FFT thread and the 60 fps `DispatcherTimer` keep running and keep updating offscreen XAML elements. The fullscreen case is the most costly one — the visualizer competes for CPU with a game/video at exactly the moment the player is deliberately hidden. Since you already compute the hide decision in `UpdateVisibility`, gate the capture thread and viz timer on it (start when shown, stop when hidden).
> Optional improvements
> 
> 
> Functionality notes

@m417z

m417z:
### Submission review

---

This is a large, well-structured update — the visualizer threading uses a `std::thread*` (trivial destructor, joined in `Wh_ModUninit`), the capture thread is stopped before `LoadSettings()` rewrites `g_settings`, and the credit for the ported FFT engine is correctly in the README rather than the `@author` field. A few issues worth addressing:

**1. Acrylic visualizer color mode ignores the `$` light/dark theme syntax (and its own default), so the bars are black.**

In `VizApplyFrame`, the Acrylic branch uses `acrCol`, which is parsed with the non-theme-aware function:

```cpp
auto acrCol = ParseColorWithSpecialValues(g_settings.vizColor, 255);
...
} else if (g_settings.vizColorMode == VizColorMode::Acrylic) {
    BYTE aa = ...;
    c = winrt::Windows::UI::Color{aa, acrCol.R, acrCol.G, acrCol.B};
}
```

The default `vizColor` is `"0 0 0$255 255 255"`. `ParseColorWithSpecialValues` splits on the first two spaces, so it reads `r=0`, `g=0`, then `b = std::stoi("0$255 255 255")` which stops at `$` and yields `0` — i.e. opaque black, regardless of theme. On a dark taskbar the Acrylic bars are invisible, and the `$` theme split documented in the setting description ("Used by Solid and Acrylic modes … two colors separated by a `$`") never takes effect for Acrylic. Solid mode is fine because it uses `baseCol` (parsed via `ParseColorWithThemeSupport`).

Fix: parse with theme support (or just reuse `baseCol`, which already holds the correct value for the non-DynamicAlbum modes):

```cpp
auto acrCol = ParseColorWithThemeSupport(g_settings.vizColor, 255);
```

**2. The visualizer capture thread and 60 FPS timer keep running while the player is hidden.**

`StartTimerThread()` starts the visualizer unconditionally when `vizEnabled`, but `g_vizCurrentlyVisible` is `false` at that point:

```cpp
if (g_settings.vizEnabled) {
    StartVizCaptureThread();
    StartVizTimer();
}
```

`UpdateVisibility` only ever *stops* the visualizer when `g_vizCurrentlyVisible` is already `true`:

```cpp
} else if (!nowVisible && g_vizCurrentlyVisible) {
    g_vizCurrentlyVisible = false;
    StopVizTimer();
    StopVizCaptureThread();
}
```

So if the player starts hidden (no media at init, or hidden after a settings change — `Wh_ModSettingsChanged` also resets `g_vizCurrentlyVisible = false` and then calls `StartTimerThread`), the WASAPI loopback capture thread and the ~60 FPS `DispatcherTimer` run indefinitely against a collapsed player (`g_vizBars` is already populated, so `VizApplyFrame` does full per-bar work every 16 ms). It only self-corrects after the first visible→hidden cycle. For a user who enables the visualizer but isn't currently playing media, that's continuous CPU/GPU and audio-capture work in `explorer.exe` for no visible output.

Fix: make visibility the single source of truth — don't auto-start the visualizer from `StartTimerThread`, and let `UpdateVisibility` start/stop it based on `nowVisible` (or, equivalently, initialize `g_vizCurrentlyVisible` so the first `UpdateVisibility` reconciles a hidden start to "stopped"). The scroll timer in the same block has the same "start unconditionally" shape, but it's cheap; the visualizer is the costly one.

<p>

- **Per-frame color re-parsing in `VizApplyFrame`.** Every tick (~60/s) parses `cg0`, `cg1`, and `acrCol` from strings even for color modes that don't use them, e.g.:
  ```cpp
  auto cg0    = ParseColorWithSpecialValues(g_settings.vizColor1, 255);
  auto cg1    = ParseColorWithSpecialValues(g_settings.vizColor2, 255);
  auto acrCol = ParseColorWithSpecialValues(g_settings.vizColor, 255);
  ```
  You already cache the Solid base color behind `g_vizBaseColorDirty`; the gradient/acrylic colors could be cached the same way (invalidated in `RefreshThemeColors`/`Wh_ModSettingsChanged`) and computed only for the active mode. It's a tiny cost, but it runs continuously.

</p>
</details>

<p>

- **Default `vizSensitivity: 300`** is the maximum of the documented `0–300` range. That makes the bars saturate/clip easily out of the box; a mid value might give a better first impression. Your call.

</p>
</details>

Salyts:
> Submission review

@m417z

PR #4515, opened 2026-06-21
Auto Theme Switcher v.1.3.2

tinodin:
## Changelog

* Fix applying lockscreen wallpaper by injecting into explorer.exe again and clearing lock screen cache

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/auto-theme-switcher.wh.cpp
+++ b/mods/auto-theme-switcher.wh.cpp
@@ -1126,7 +1126,6 @@ void ClearLockScreenCache() {
             RegDeleteValueW(hKey, (L"Details_" + suffix).c_str());
             RegDeleteValueW(hKey, (L"ImageId_" + suffix).c_str());
             RegDeleteValueW(hKey, (L"OriginalFile_" + suffix).c_str());
-            Wh_Log(L"Cleared Lock Screen Cache");
         }
     }
 
--- a/mods/auto-theme-switcher.wh.cpp
+++ b/mods/auto-theme-switcher.wh.cpp
@@ -2,10 +2,10 @@
 // @id              auto-theme-switcher
 // @name            Auto Theme Switcher
 // @description     Automatically switch between light and dark appearance/wallpapers/themes based on a schedule with hotkey and custom script support
-// @version         1.3.1
+// @version         1.3.2
 // @author          tinodin
 // @github          https://github.com/tinodin
-// @include         windhawk.exe
+// @include         explorer.exe
 // @license         GPL-3.0
 // @compilerOptions -lole32 -loleaut32 -lwindowsapp -lruntimeobject -lkernel32 -luser32 -lshell32
 // ==/WindhawkMod==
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

For the first item, I'm not sure about the over-deletion, but the path concern makes sense.

---

- **`ClearLockScreenCache` deletes lock-screen cache values it doesn't own, and over-broadly.** The new function reaches into `HKCU\…\Lock Screen` and, the moment *any one* of `Details_A`…`Details_G` matches the hardcoded string `IMAGENAME:C:\Program Files\Windhawk\windhawk.exe`, it deletes **all** seven slots' `Details_/ImageId_/OriginalFile_` values. Two problems:

  - **Over-deletion.** If the user has Spotlight or any manually-set lock-screen image cached in the other slots, those get wiped too, forcing Windows to rebuild the cache. The deletion should be scoped to only the slots that actually reference the mod's own stale entry.
  - **Hardcoded install path.** Windhawk can be installed outside `C:\Program Files\Windhawk` (custom/portable installs). For those users the exact-match fails, so the stale-entry cleanup this PR is meant to perform silently never runs — defeating the fix for exactly the users migrating from the v1.3.1 windhawk.exe host. Match on the `windhawk.exe` file name instead of the full path.

A version that fixes both — match by file name, delete only matching slots:

```cpp
auto isWindhawkEntry = [](PCWSTR s) -> bool {
    constexpr PCWSTR kPrefix = L"IMAGENAME:";
    if (_wcsnicmp(s, kPrefix, 10) != 0) return false;
    PCWSTR path = s + 10;
    PCWSTR name = wcsrchr(path, L'\\');
    return _wcsicmp(name ? name + 1 : path, L"windhawk.exe") == 0;
};

for (wchar_t c = L'A'; c <= L'G'; ++c) {
    std::wstring suffix(1, c);
    wchar_t data[512];
    DWORD dataSize = sizeof(data);
    if (RegGetValueW(hKey, NULL, (L"Details_" + suffix).c_str(), RRF_RT_REG_SZ, NULL, data, &dataSize) == ERROR_SUCCESS
        && isWindhawkEntry(data)) {
        RegDeleteValueW(hKey, (L"Details_" + suffix).c_str());
        RegDeleteValueW(hKey, (L"ImageId_" + suffix).c_str());
        RegDeleteValueW(hKey, (L"OriginalFile_" + suffix).c_str());
        Wh_Log(L"Cleared Lock Screen cache slot %c", c);
    }
}
```

If the slots are genuinely interdependent and Windows needs *all* of them cleared together for the rebuild to take effect, that's fine — but then say so in a comment, because as written it reads like an unintentional "match one, nuke all." You've tested the behavior, so you're best placed to judge; the install-path-agnostic match is the part that's clearly worth keeping regardless.

This is also a persistent registry deletion that isn't undone when the mod is disabled. It's defensible as a one-time migration cleanup of the mod's *own* footprint (the entries it created), which is why it's not a blocker — but keeping it scoped to only the mod's entries is what keeps it within Windhawk's "no persistent system changes" principle.

- **Add a comment explaining why `explorer.exe` is targeted.** The hollow-`explorer.exe` host does preserve the crash-isolation benefit of a tool mod (a scheduler crash kills only the phantom process, not the shell), which is a point in favor of this approach over running the scheduler directly inside the real shell `explorer.exe`. If the explorer identity really is required for the lock-screen API, this is a reasonable way to keep both that identity and crash isolation — worth a short code comment noting the intent, since the tool-mod boilerplate comment still says "dedicated windhawk.exe process," which no longer matches reality.

tinodin:
@m417z 
I **removed the hardcoded path** to support portable installations of Windhawk. 

**Clearing all of the lock screen cache is required**, otherwise the wallpapers set by the user via the Windows Settings app remain broken (shows a different wallpaper than selected). This does not remove the history of previously selected wallpapers from the Windows Settings app, though selecting them from the recent images picker still shows a different wallpaper than selected until the wallpaper is readded using the `Browse photos` button. This of course assumes the user added a new lock screen wallpaper while it was in a broken state. Windows Spotlight as lock screen wallpaper is unaffected.

I **replaced** the old `windhawk.exe` **comment** with `explorer.exe`.

PR #4516, opened 2026-06-21
Update Taskbar Virtual Desktop Switcher to v1.7

sb4ssman:
Updates `taskbar-vd-switcher` from v1.5 to v1.7.

## Changelog

* Fixed the active-desktop highlight sticking to the startup desktop: every desktop notification now fully rebuilds the button grid (lightweight-styling resources resolve once at template application, so in-place swaps never took effect). This may also resolve the hover/hit-test report in #4784
* Fixed inactive buttons becoming partially unclickable after switching desktops
* Windows accent color support: color settings accept `accent`, `accentLight`, `accentDark`, and `transparent` in addition to hex; the active desktop now uses the Windows accent color by default (empty = plain native surface)
* Experimental "Show on all taskbars" option: also injects the switcher into secondary monitors' taskbars (tray positions only; requested in a comment on #4785)
* Renamed "Master button" to "Task View button" in settings and readme
* Settings page reordered so color/state options follow a consistent order; expanded readme with screenshot gallery, feature summary, and full settings table

All published setting keys are unchanged; existing configurations keep working.

### DIFF
--- a/mods/taskbar-vd-switcher.wh.cpp
+++ b/mods/taskbar-vd-switcher.wh.cpp
@@ -2,7 +2,7 @@
 // @id              taskbar-vd-switcher
 // @name            Taskbar Virtual Desktop Switcher
 // @description     Injects clickable buttons into the taskbar — one per virtual desktop — with configurable grid arrangement for direct switching.
-// @version         1.6
+// @version         1.7
 // @author          sb4ssman
 // @github          https://github.com/sb4ssman
 // @include         explorer.exe
@@ -58,6 +58,7 @@ A [Windhawk](https://windhawk.net) mod for Windows 11 that injects clickable but
 - Per-state text color, font size, corner radius, bold, and border
 - Tooltip on each button shows the desktop's display name
 - Option to hide the bar entirely when only one desktop exists
+- Experimental option to also show the switcher on secondary monitors' taskbars
 
 ## Settings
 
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

---

The backward-compatibility shim for the renamed settings — the headline feature of this update ("old setting names still work") — does not actually work for the integer settings, and works for only one of the two string settings.

The cause is the `Wh_GetIntSetting` / `Wh_GetStringSetting` contract. Their variadic arguments are **printf-style format specifiers for the setting *name*** (used for indexed settings like `key[%d]`), **not a default value** — for a setting that has no value the functions return `0` / `L""`, full stop ([wiki](https://github.com/ramensoftware/windhawk/wiki/Creating-a-new-mod), `Wh_GetIntSetting`). The `default` that the rest of the file passes as a second argument (e.g. `Wh_GetIntSetting(L"buttonWidth", 20)`) only *appears* to work because the real default comes from the settings block (`Each setting is defined by a name and a default value. The default value is set when the mod is installed`), and authors happen to repeat the same number in both places — the `20` is silently consumed as a format arg and ignored.

That breaks the alias detection:

```cpp
auto AliasedInt = [](const wchar_t* canonical, const wchar_t* legacy, int fallback) {
    constexpr int kUnset = -2147483647;
    int value = Wh_GetIntSetting(canonical, kUnset);  // kUnset is a format arg → ignored
    if (value == kUnset) {                            // never true
        value = Wh_GetIntSetting(legacy, fallback);   // dead code
    }
    return value;
};
```

`gridRows`, `gridColumns`, `opacity`, `groupPaddingLeft`, `groupPaddingRight`, and `groupOffsetY` are all declared in the settings block, so each always returns its block default (or the user's new value) — never `kUnset`. The `value == kUnset` branch is unreachable, so the legacy names (`buttonRows`, `buttonColumns`, `buttonOpacity`, `paddingLeft`, `paddingRight`, `gridVerticalOffset`) are never read.

`AliasedStr` has the same flaw via a different path: it falls back to the legacy name only when the canonical value is empty. `inactiveBackgroundColor` has an empty block default, so its alias to `inactiveColor` does work — but `activeBackgroundColor` defaults to `#4488FF` (non-empty), so the legacy `activeColor` is never consulted.

**User impact:** an existing user who had customized rows/columns, opacity, padding, vertical offset, or active color in v1.5 will silently lose those values on update — they reset to the new defaults. This is the opposite of what the PR claims.

There's no clean way to fix the shim, because (a) mods can't write user settings (there is no `Wh_SetIntSetting`), and (b) a declared setting at its default is indistinguishable from a setting the user explicitly set to that default, so you can't reliably detect "the user hasn't touched the canonical name." Given that, the realistic options are:

- **Recommended: keep the legacy key names.** Revert the YAML key renames and the corresponding code reads back to the v1.5 names (`buttonRows`, `buttonColumns`, `activeColor`, `inactiveColor`, `buttonOpacity`, `paddingLeft`, `paddingRight`, `gridVerticalOffset`) and drop `AliasedInt`/`AliasedStr` entirely. The cosmetic benefit of "canonical rubric names" isn't worth silently resetting existing users' configuration. You can still relabel them in the UI freely via `$name` — only the *key* needs to stay stable.
- If the rename really must happen, then accept that updating resets those settings to defaults, remove the non-functional shim, and say so plainly in the README / PR description (the current "old setting names still work" wording is inaccurate). Human note: I don't think there's a reason renames "really must happen". I think it's best to keep them unchanged to avoid confusion, as the migration can have edge cases one way or another.

Since this is AI-assisted (commit co-authored by Claude Sonnet 4.6), this is worth calling out specifically: the sentinel-default pattern is a plausible-but-wrong assumption about the `Wh_Get*Setting` API that compiles and runs cleanly but never does what it looks like it does.

<p>

- The `ModSettings` struct still initializes `int buttonOpacity = 70;` (line 355), while the settings block, the README table, and the `LoadSettings` call all use `100`. The struct initializer is dead (it's overwritten by `LoadSettings` before first use), but it's confusing — make it `100` for consistency.
- In `AliasedStr`, the legacy branch is `std::wstring r = p ? p : fallback;`. Since `Wh_GetStringSetting` never returns `NULL` (it returns `L""`), `p` is always truthy, so `fallback` is dead code — `r` becomes `p` (the legacy value, or `L""` if unset) regardless. It happens not to matter today (the only path that reaches it is `inactiveBackgroundColor`, whose `fallback` is `L""` anyway), but if you keep this helper, the empty-string check would need to mirror the canonical one. (Moot if you take the "keep legacy names" route above and delete the helper.)
- The README's H1 is now `# Virtual Desktop Switcher`, but the mod's `@name` is still `Taskbar Virtual Desktop Switcher`. Harmless, but the two reading differently is a small inconsistency.

</p>
</details>

sb4ssman:
Updated v1.6 with the requested review fixes and recent issue reports:

- Retained the published setting keys so existing configurations continue working.
- Removed the broken alias helpers.
- Fixed inactive buttons becoming partially unclickable after switching desktops.
- Added Windows accent-color support by entering accent in the applicable color settings.
- Clarified the accent-color instructions.
- Aligned the opacity default, mod title, and v1.6 initialization log.

The updated source compiles successfully, and repeated desktop switching was tested locally with the buttons remaining clickable.

m417z:
Please update the changelog too at the top.

m417z:
Looks like you updated the changelog and then restored it. Please clarify so that I'll be able to merge the update.

sb4ssman:
Sorry for the confusion — the first revision renamed setting keys with alias fallbacks; I reverted that so the published keys stay unchanged. The PR description now carries the changelog for the final diff.

sb4ssman:
Pushed v1.7, superseding the v1.6 revision — it folds in the fixes and features from the last few weeks of local testing (full-rebuild highlight fix, accent-by-default active color with the new color tokens, and the experimental multi-monitor option requested in #4785). The changelog above reflects the full v1.5 → v1.7 diff. Everything has been live-tested locally.

PR #4519, opened 2026-06-22
Update taskbar-ai-quota: built-in OAuth sign-in

Cleroth:
Update taskbar-ai-quota: built-in OAuth sign-in

- Built-in OAuth: the mod now signs you in to Anthropic and OpenAI through your browser instead of reading CLI credential files. Tokens are stored encrypted and refreshed automatically. Sign in / Sign out from a column's right-click menu.
- Show/hide individual accounts from the right-click menu.
- Fix: bars no longer overlap the clock/tray on a cold start.
- More robust sign-in and token-refresh handling.

Note: accounts are now just provider + label, so re-check your account settings after updating.

### DIFF
--- a/mods/taskbar-ai-quota.wh.cpp
+++ b/mods/taskbar-ai-quota.wh.cpp
@@ -2,13 +2,13 @@
 // @id              taskbar-ai-quota
 // @name            Taskbar AI Quota Bars
 // @description     Shows compact 5-hour and weekly AI agent/LLM subscription quota bars for Anthropic and OpenAI on the Windows 11 taskbar
-// @version         0.9.2
+// @version         0.10.2
 // @author          Cleroth
 // @github          https://github.com/Cleroth
 // @include         explorer.exe
 // @architecture    x86-64
 // @license         MIT
-// @compilerOptions -lole32 -loleaut32 -lruntimeobject -lwindowsapp -lwinhttp -luser32 -lshell32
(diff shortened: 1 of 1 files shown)


mods/taskbar-ai-quota.wh.cpp
```diff
@@ -157,11 +157,18 @@ own auth files if requests start returning `401`.
 */
 // ==/WindhawkModSettings==
 
+// winsock2.h must precede windows.h to avoid the legacy winsock.h being pulled in.
```
m417z: This doesn't work - Windhawk implicitly includes `windhawk_api.h` which includes `windows.h`. See the alternative solution here: https://github.com/ramensoftware/windhawk-mods/pull/4301#discussion_r3363719328.
tl;dr add `-DWIN32_LEAN_AND_MEAN` to `@compilerOptions`.

Cleroth: Good catch, thanks. Added -DWIN32_LEAN_AND_MEAN to @compilerOptions so the implicitly-included windows.h no longer pulls in legacy winsock.h, and updated the comment to reflect that. Verified the original include order triggers clang's Please include winsock2.h before windows.h warning and that the define clears it. Force-pushed the fix.

Cleroth: You're absolutely right!

PR #4530, opened 2026-06-23
[codex] Add Dynamic Taskbar Transparency mod

r1file:
﻿## What changed

Adds `dynamic-taskbar-transparency`, a Windows 11 taskbar mod that changes the taskbar background based on shell state.

The mod supports configurable appearances for desktop/no-maximized-window, maximized windows, Start, search, task view, tray flyouts, and optional unclassified shell interactions. Supported appearances include existing style/other mods, Windows native/default, clear, blur, and acrylic-like blur.

## Demo

Dynamic Taskbar Transparency demo

## Difference from taskbar-background-helper

`taskbar-background-helper` is a lower-level background helper for the Windows 11 taskbar. It mainly applies blur/acrylic/color either always or when a maximized window exists, and it is designed to pair with Taskbar Styler.

This mod is intentionally focused on the TranslucentTB-style shell state machine: desktop/no-maximized-window, maximized windows, Start, search, task view, tray/notification/quick-settings flyouts, and an optional catch-all for other shell-host surfaces can each resolve to a separate configured appearance. It also owns the fade animation between those resolved states. The overlap is the XAML taskbar background surface; the additional behavior is the per-shell-state resolution and transition layer.

## Notes

The mod targets `explorer.exe` and edits the Windows 11 taskbar XAML background rectangles. The blur implementation follows the WindhawkBlur-style approach used by Windows 11 Taskbar Styler / TranslucentTaskbar instead of relying on `AcrylicBrush` directly.

Shell-state detection is heuristic and based on shell/window events plus class/process-name matching for Start, search, task view, tray flyouts, and other shell-host surfaces.

Review follow-up changes:

- License changed to `GPL-3.0` because the blur stack and taskbar symbol hook approach reuse GPL-compatible Windhawk mod code.
- Worker lifetime now uses a heap `std::thread*`, joined and deleted in `Wh_ModUninit`.
- Captured XAML brush references are released via each element's dispatcher during unload instead of by a global vector destructor/clear path.
- `g_scanPending` is reset on all non-dispatched scan paths.
- State detection is now driven by `SetWinEventHook` instead of a fixed polling loop; the worker only re-resolves shell/window state when an event or settings change requests it.
- Steady-state duplicate UI dispatch is skipped when the appearance did not change.
- English-only localized title matching was removed from shell-state detection.
- Worker startup now happens after fallible init hook setup succeeds.
- Minor cleanup: removed the dead dispatch parameter, the unused `normal` style alias, redundant teardown restore, and the configurable polling interval.

## Validation

- `.github/pr_validation.py` passed locally with the PR changed-file environment.
- `scripts/compile_mod.py` passed locally with Windhawk from `C:\Program Files\Windhawk` and `-ldwmapi -lole32 -loleaut32 -lruntimeobject`.
- PR scope is a single file: `mods/dynamic-taskbar-transparency.wh.cpp`.

### DIFF
--- a/mods/dynamic-taskbar-transparency.wh.cpp
+++ b/mods/dynamic-taskbar-transparency.wh.cpp
@@ -2,13 +2,13 @@
 // @id              dynamic-taskbar-transparency
 // @name            Dynamic Taskbar Transparency
 // @description     Dynamically changes the Windows 11 taskbar XAML background transparency for desktop, Start, search, tray flyouts, task view, and maximized windows.
-// @version         0.3.5
+// @version         0.3.6
 // @author          11581
 // @github          https://github.com/r1file
 // @include         explorer.exe
 // @architecture    x86-64
 // @compilerOptions -ldwmapi -lole32 -loleaut32 -lruntimeobject
-// @license         MIT
+// @license         GPL-3.0
 // ==/WindhawkMod==
 
 // ==WindhawkModReadme==
@@ -29,6 +29,15 @@ blur-backed acrylic.
 
 Do not run this together with TranslucentTB or another tool that continuously
 writes the same taskbar background.
+
+Compared with taskbar-background-helper, this mod focuses on a TranslucentTB-
+style state machine: desktop/no-maximized-window, maximized windows, Start,
+search, task view, tray flyouts, notification/quick settings flyouts, and an
+optional catch-all for other shell surfaces can each choose their own
+appearance and fade duration. taskbar-background-helper is a lower-level
+background helper for always/maximized behavior and pairs well with Taskbar
+Styler; this mod adds per-shell-state switching and animation on top of the
+same general taskbar background use case.
 */
 // ==/WindhawkModReadme==
 
@@ -262,7 +271,10 @@ std::mutex g_settingsMutex;
 Settings g_settings;
 
 std::mutex g_elementsMutex;
-std::vector<TrackedBackgroundElement> g_backgroundElements;
+std::vector<TrackedBackgroundElement>& BackgroundElements() {
+    static auto* elements = new std::vector<TrackedBackgroundElement>();
+    return *elements;
+}
 
 struct ActiveAppearance {
     bool hasValue = false;
@@ -274,7 +286,7 @@ std::mutex g_activeAppearanceMutex;
 ActiveAppearance g_activeAppearance;
 
 std::atomic<bool> g_stopWorker = false;
-std::thread g_workerThread;
+std::thread* g_workerThread = nullptr;
 std::atomic<bool> g_taskbarViewDllLoaded = false;
 std::atomic<bool> g_scanPending = false;
 
@@ -299,10 +311,8 @@ int ClampInt(int value, int minValue, int maxValue) {
 }
 
 std::wstring GetStringSetting(PCWSTR name) {
-    PCWSTR value = Wh_GetStringSetting(name);
-    std::wstring result = value ? value : L"";
-    Wh_FreeStringSetting(value);
-    return result;
+    auto value = WindhawkUtils::StringSetting::make(name);
+    return value.get();
 }
 
 AppearanceStyle ParseStyle(const std::wstring& value) {
@@ -1134,7 +1144,7 @@ void ApplyAppearanceToElement(Shapes::Rectangle rectangle,
 
 TrackedBackgroundElement GetOriginalElementInfo(void* abi) {
     std::lock_guard<std::mutex> lock(g_elementsMutex);
-    for (const auto& tracked : g_backgroundElements) {
+    for (const auto& tracked : BackgroundElements()) {
         if (tracked.abi == abi) {
             return tracked;
         }
@@ -1147,7 +1157,7 @@ void StoreAppliedAppearance(void* abi,
                             AppearanceStyle style,
                             bool restoreOriginal) {
     std::lock_guard<std::mutex> lock(g_elementsMutex);
-    for (auto& tracked : g_backgroundElements) {
+    for (auto& tracked : BackgroundElements()) {
         if (tracked.abi == abi) {
             if (restoreOriginal) {
                 tracked.hasAppliedAppearance = false;
@@ -1158,23 +1168,6 @@ void StoreAppliedAppearance(void* abi,
             return;
         }
     }
-}
-
-Media::Brush CaptureNativeFill(Shapes::Rectangle rectangle) {
-    Media::Brush originalFill = rectangle.Fill();
-
-    if (IsTransparentBrush(originalFill)) {
-        try {
-            rectangle.ClearValue(Shapes::Shape::FillProperty());
-            Media::Brush restoredFill = rectangle.Fill();
-            if (restoredFill && !IsTransparentBrush(restoredFill)) {
-                originalFill = restoredFill;
-            }
-        } catch (...) {
-        }
-    }
-
-    return originalFill;
 }
 
 struct BackgroundElementSnapshot {
@@ -1202,11 +1195,12 @@ void DispatchApplyAppearance(const Appearance& appearance,
     std::vector<BackgroundElementSnapshot> snapshot;
     {
         std::lock_guard<std::mutex> lock(g_elementsMutex);
-        std::erase_if(g_backgroundElements, [](const auto& tracked) {
+        auto& elements = BackgroundElements();
+        std::erase_if(elements, [](const auto& tracked) {
             return !tracked.element.get();
         });
-        snapshot.reserve(g_backgroundElements.size());
-        for (const auto& tracked : g_backgroundElements) {
+        snapshot.reserve(elements.size());
+        for (const auto& tracked : elements) {
             snapshot.push_back({tracked.element, tracked.abi});
         }
     }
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

---

This is a well-structured mod and the core architecture (hook `TaskListButton`/`ExperienceToggleButton` to discover the XAML background rectangles, manipulate them on the UI thread via the dispatcher, restore on unload) is sound and follows established taskbar-mod patterns. A few things need attention before merge.

**1. `@license MIT` is incompatible with the borrowed GPLv3 code.** The whole blur stack — `IGraphicsEffectD2D1Interop`, the `guid_v` specialization, `CompositeEffect` / `FloodEffect` / `GaussianBlurEffect`, and the `XamlCompositionBrushBase` blur brush — is taken almost verbatim from [Windows 11 Taskbar Styler](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/windows-11-taskbar-styler.wh.cpp#L6486-L6979), which is **GPLv3** (your readme even credits "the WindhawkBlur-style approach used by Windows 11 Taskbar Styler"). `GetFrameworkElementFromNative` and the three symbol hooks are likewise copied from [taskbar-content-presenter-injector](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/taskbar-content-presenter-injector.wh.cpp#L90-L98). GPLv3 code cannot be relicensed under MIT. Change `@license` to `GPL-3.0` (compatible) or replace the borrowed code with your own implementation.

**2. The global `std::thread g_workerThread` will call `std::terminate()` on process shutdown.** It's only joined in `Wh_ModUninit`, but `Wh_ModUninit` does **not** run when the host process terminates (Explorer restart, sign-out, reboot) — there the CRT runs the global's destructor during `DLL_PROCESS_DETACH`, the thread is still `joinable()`, so `~thread()` calls `std::terminate()` and aborts explorer. Use a heap pointer with a trivial destructor instead, and join + delete it in `Wh_ModUninit`:
```cpp
std::thread* g_workerThread = nullptr;
// init:    g_workerThread = new std::thread(WorkerLoop);
// uninit:  g_stopWorker = true; g_workerThread->join(); delete g_workerThread;
```
Relatedly, `g_backgroundElements` is a global `std::vector` of `Media::Brush` (XAML) strong refs; its destructor (at shutdown, and `clear()` in `Wh_ModUninit`) releases XAML objects from an arbitrary/shutdown thread rather than their UI thread. Prefer releasing the captured brushes on the dispatcher thread during the restore step, and keep globals trivially-destructible.

**3. The worker re-runs full window enumeration on every animation frame.** `WorkerLoop` calls `FindTaskbars()` + `DetectShellActivity()` + `HasMaximizedWindow()` — three separate `EnumWindows` passes, each doing per-window `OpenProcess` + `QueryFullProcessImageNameW` — once per loop iteration, and the loop iterates every `frameTimeMs` (16 ms) while an animation is in progress. That's ~180 `EnumWindows` calls/second plus a process-image lookup per top-level window, just to drive a fade. The animation only needs to interpolate opacity between frames; state detection only needs to run at the polling cadence. At minimum, decouple them: detect at `pollingIntervalMs`, and only interpolate/apply on the in-between frames. Better still, drive detection from events rather than polling — [taskbar-background-helper](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/taskbar-background-helper.wh.cpp#L1068-L1092) (which does the same "maximized window?" detection) uses `SetWinEventHook` (`EVENT_OBJECT_*`, `EVENT_SYSTEM_PEEK*`) instead of a poll loop.

**4. Significant overlap with [taskbar-background-helper](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/taskbar-background-helper.wh.cpp).** That mod (by m417z) already sets the Win11 taskbar background to blur / acrylic / color, "always or only when there's a maximized window," and is designed to pair with the Taskbar Styler — which covers a large part of what this mod does. Your addition is the per-shell-state machine (desktop / Start / search / task view / tray flyout) with fades, which is genuinely more than the existing mod, so this isn't a pure duplicate. But the maintainer strongly prefers extending an existing mod over shipping an overlapping one. Please state plainly in the PR how this differs, and consider whether the state-machine behavior could be contributed to `taskbar-background-helper` (or whether that mod's maximized-detection could just be reused) rather than reimplemented.

**5. No screenshot/GIF in the readme.** This is a visual mod, so a short GIF showing the taskbar changing across the states (desktop → maximized → Start, etc.) would help users a lot. `taskbar-background-helper` includes one. Only `i.imgur.com` and `raw.githubusercontent.com` are allowed image hosts.

**6. Some of the state detection matches on localized window titles** — `title.find(L"task view")`, `L"notification"`, `L"quick settings"` in `EnumShellActivityProc`. These are English-only and won't match on other UI languages. The class-name (`MultitaskingViewFrame`, `NotifyIconOverflowWindow`, …) and process-name checks are language-independent and already cover the main cases, so the title checks are mostly redundant — you could drop them or document the limitation.

<p>

- `GetStringSetting` reinvents `WindhawkUtils::StringSetting` (RAII), and its `value ? value : L""` guard is dead code — `Wh_GetStringSetting` never returns `NULL`, only `L""`. You can drop the helper and read strings via `WindhawkUtils::StringSetting` directly.

- In steady state (animation finished, state unchanged) the worker still calls `DispatchApplyAppearance(currentAppearance, false)` every `pollingIntervalMs`, dispatching a redundant `Opacity` set to the UI thread ~12×/second forever. Skip the dispatch when `currentAppearance` is unchanged since the last applied frame.

- `Wh_ModInit` starts the worker thread *before* the fallible `HookTaskbarViewDllSymbols` call, then can `return FALSE`. On init failure `Wh_ModUninit` is not called, so the worker thread is left running (and `g_stopWorker` never set). Start the thread only after all fallible setup has succeeded.

</p>
</details>

<p>

- The cross-style transition barely fades. In the `visibleStyleChange` path the opacity floor is `min(opacity, 245)` while the styles are both opaque at 255, so a blur→acrylic (or similar) change dips only ~4% before swapping the brush at the midpoint — effectively an abrupt brush swap rather than a visible crossfade. If a smoother handoff is intended, the floor needs to drop much lower (or fade the outgoing brush out fully before fading the incoming one in).

- `CaptureNativeAppearance` temporarily clears the rectangle's fill/opacity and the parent grid background on the live element to read the "native" defaults, then restores them. This runs on the UI thread during scanning, but it does briefly mutate the visible element, which could show up as a one-frame flicker when a new background rectangle is first discovered.

</p>
</details>

m417z:
Some of the items weren't addressed. If you decided to skip them deliberately, please let me know why. Below is an updated review.

### Submission review

---

**1. `g_scanPending` can latch to `true` permanently and silently stop element discovery.** In `QueueScanFromElement`, the flag is set with `exchange(true)` up front but the two early-exit paths don't reset it:

```cpp
void QueueScanFromElement(void* pThis) {
    if (g_scanPending.exchange(true)) {
        return;
    }
    try {
        FrameworkElement element = GetFrameworkElementFromNative(pThis);
        if (!element) {
            return;            // g_scanPending stays true forever
        }
        auto dispatcher = element.Dispatcher();
        if (!dispatcher) {
            return;            // g_scanPending stays true forever
        }
        ...
```

`GetFrameworkElementFromNative` swallows exceptions and returns `nullptr` on any failure, so a single null result (e.g. an `ExperienceToggleButton` whose projection doesn't resolve, or a transient `Dispatcher()` failure) wedges the flag and **no further scans ever run** — newly created taskbar backgrounds (secondary monitor connected, taskbar recreated) are never registered and the mod silently stops working for them. Reset the flag on every non-dispatched path (set `g_scanPending = false;` before each `return`, or use a scope guard that clears it unless ownership is handed to the dispatched lambda).

**2. The detection loop is pure polling with heavy per-poll work.** `WorkerLoop` runs `FindTaskbars()` + `DetectShellActivity()` + `HasMaximizedWindow()` every `pollingIntervalMs` (default 80 ms ≈ 12×/s). That's **three full `EnumWindows` sweeps**, and two of them call `GetProcessFileName` → `OpenProcess` + `QueryFullProcessImageNameW` for *every* visible top-level window each pass — on a busy desktop that's well over a thousand `OpenProcess` calls per second, forever, inside `explorer.exe`, even when nothing changes. Your sibling mod [taskbar-background-helper](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/taskbar-background-helper.wh.cpp#L1066) solves the same maximized/fullscreen detection event-driven via `SetWinEventHook` (`EVENT_OBJECT_*`, `EVENT_SYSTEM_PEEK*`) and keeps per-monitor state incrementally, with no polling at all. Please move at least the window-state detection to `SetWinEventHook` and react on change; if the Start/Search/TaskView/flyout surfaces are harder to do purely event-driven, at minimum cache process names and avoid re-`OpenProcess`-ing every window on every tick. The current cost in the shell process is the main concern.

**3. No screenshot/GIF in the README.** This is a purely visual mod, so a short GIF showing the desktop→maximized / Start-open transitions would help a lot (taskbar-background-helper includes one). Only `i.imgur.com` and `raw.githubusercontent.com` are allowed image hosts.

<p>

- **Event-handle leak on timeout.** In `DispatchApplyAppearance` (and identically in `ReleaseTrackedBackgroundElements`), the completion event is only closed on `WAIT_OBJECT_0`:
  ```cpp
  if (WaitForSingleObject(completionEvent, 500) == WAIT_OBJECT_0) {
      CloseHandle(completionEvent);
  }
  ```
  On timeout the handle leaks (and the still-pending lambda will `SetEvent` it later). Close it on all paths (e.g. `WaitForSingleObject(...); CloseHandle(completionEvent);`).

- **Dead parameter.** `DispatchApplyAppearance`'s 4th argument `unusedForceClearNative` is always passed and immediately `(void)`-discarded — drop it (and the always-`false` call-site arguments).

- **Dead alias.** `ParseStyle` accepts `L"normal"` as a synonym for `captured`, but no `$options` value emits `normal` — remove it.

- **Redundant teardown restore.** `WorkerLoop` already calls `DispatchApplyAppearance(currentAppearance, /*restoreOriginal=*/true, ...)` before exiting, and `Wh_ModUninit` does the same restore again after joining the worker. The second call is redundant (restore is idempotent).

- **Over-plumbed `frameTimeMs`.** `Settings::frameTimeMs` is a struct field that's only ever set to the constant `1000 / 60` — it isn't a setting and never varies, so a local constant would be clearer.

</p>
</details>

<p>

- **Maximized state is global, not per-monitor.** `HasMaximizedWindow` returns a single bool and the resolved appearance is applied to *all* tracked taskbar backgrounds. So maximizing a window on monitor 1 also switches monitor 2's (empty-desktop) taskbar into the maximized appearance. taskbar-background-helper tracks this per-monitor; if you want parity you'd need to resolve/apply per taskbar monitor.

- **Shell-activity detection is heuristic.** Classification leans on class-name substring matching (`multitasking`, `notifyiconoverflow`, `clockflyout`, `shellflyout`) and a process-name allowlist, with an `otherInteraction` catch-all (off by default). This can misclassify or miss surfaces across Windows 11 builds — e.g. a `textinputhost.exe` IME popup falls into `otherInteraction`. Worth noting the heuristic nature in the README.

- **State-change latency.** Because detection only runs on the polling tick, a Start/flyout open or a window maximize lags by up to `pollingIntervalMs` before the fade begins; the default 80 ms is fine, but the larger values the setting allows (up to 1000 ms) will feel sluggish.

</p>
</details>

PR #4531, opened 2026-06-23
Add mod: lock-keys-notifier

havrlisan:
## Changelog

A new mod that shows a highly customizable toast when a lock key (Caps/Num/Scroll/Insert) is toggled, displaying its new state.

## Mod authorship

This mod was created by:

- - [X] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/lock-keys-notifier.wh.cpp
+++ b/mods/lock-keys-notifier.wh.cpp
@@ -6,8 +6,7 @@
 // @author          Havrlisan
 // @github          https://github.com/havrlisan
 // @homepage        https://github.com/havrlisan/lock-keys-notifier
-// @include         explorer.exe
-// @architecture    x86-64
+// @include         windhawk.exe
 // @license         MIT
 // @compilerOptions -lgdiplus -ldwmapi -lwinmm -lgdi32 -luser32 -lshcore -lshell32
 // ==/WindhawkMod==
@@ -18,6 +17,18 @@
 
 Displays a small, customizable toast whenever a lock key (Caps Lock, Num Lock,
 Scroll Lock, or Insert) is toggled, showing its new ON/OFF state.
+
+## Preview
+
+Pill, Tile, and Minimal layouts (following the system theme and accent):
+
+![Pill layout](https://raw.githubusercontent.com/havrlisan/lock-keys-notifier/master/images/caps-lock-on-pill.png)
+![Tile layout](https://raw.githubusercontent.com/havrlisan/lock-keys-notifier/master/images/scroll-lock-on-tile.png)
+![Minimal layout](https://raw.githubusercontent.com/havrlisan/lock-keys-notifier/master/images/caps-lock-on-minimal.png)
+
+Insert can show a single fixed label in a neutral color instead of ON/OFF:
+
+![Insert single-value](https://raw.githubusercontent.com/havrlisan/lock-keys-notifier/master/images/insert-pressed-tile.png)
 
 ## Features
 - Per-key enable/disable.
@@ -35,11 +46,12 @@ Scroll Lock, or Insert) is toggled, showing its new ON/OFF state.
   size on mixed-DPI multi-monitor setups.
 
 ## Notes
-- Runs in explorer.exe; notifications pause if Explorer is not running.
+- Runs as a tool mod in its own windhawk.exe process, so notifications keep
+  working even when Explorer is restarting or not running.
 - Toggles made while an elevated (administrator) app has focus are not detected.
-  Explorer runs at medium integrity and Windows (UIPI) blocks it from observing
-  input to higher-integrity windows; this can't be worked around from a mod. The
-  next toggle in a normal app shows the correct state.
+  The mod process runs at medium integrity and Windows (UIPI) blocks it from
+  observing input to higher-integrity windows; this can't be worked around from a
+  mod. The next toggle in a normal app shows the correct state.
 - Fullscreen exclusive apps may cover the toast (or enable the option above to skip it).
 - Insert reports the OS toggle bit, not an app's overtype mode (off by default).
 */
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

---

**This should be a tool mod, not an `explorer.exe` mod.** The mod installs **no function/symbol hooks** — its only system integration is `SetWindowsHookExW(WH_KEYBOARD_LL, ...)`, which is a global hook that works from any process. Everything else (its own worker thread, toast windows, `EnumDisplayMonitors`, `GetForegroundWindow`, theme/accent reads, `PlaySound`) operates on the desktop as a whole and ignores Explorer's state entirely. It hits exactly the problems the "mods as tools" pattern exists to fix:

- **Duplicate toasts**: with "launch folder windows in a separate process" enabled (or any time more than one `explorer.exe` exists), each instance loads the mod and installs its own `WH_KEYBOARD_LL` hook — so one Caps Lock press shows two (or more) overlapping toasts.
- **Shell stability**: any fault in the GDI+ rendering path takes down the whole shell, for something that's conceptually a standalone helper.

Please convert it to a tool mod: change `@include explorer.exe` → `@include windhawk.exe`, rename `Wh_ModInit`/`Wh_ModSettingsChanged`/`Wh_ModUninit` → `WhTool_*`, and paste the launcher snippet verbatim. See the wiki ([Mods as tools](https://github.com/ramensoftware/windhawk/wiki/Mods-as-tools:-Running-mods-in-a-dedicated-process)) and a working example with the boilerplate at the bottom of the file: [mods/theme-toggler-tray.wh.cpp](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/theme-toggler-tray.wh.cpp) (which also uses `SetWindowsHookExW`) and [mods/explorer-folder-hover-menu.wh.cpp](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/explorer-folder-hover-menu.wh.cpp).

**Add a screenshot or GIF to the README.** The mod is heavily visual (three layouts, theming, positions, fade) but the README has no image. A short GIF of a toast on a key toggle would help a lot — only `i.imgur.com` and `raw.githubusercontent.com` are allowed image hosts.

<p>

Minor polish — none of this affects users in the default configuration, so it's your call.

- **Don't run `IsFullscreenActive()` inside the low-level keyboard hook.** `WH_KEYBOARD_LL` callbacks run synchronously and block *all* system input until they return, so `SHQueryUserNotificationState` + the window/monitor queries shouldn't sit on that path. The blast radius here is small (it only runs on lock-key *releases*, and only when `suppressFullscreen` is enabled — off by default), but it's trivial to move: have the hook always `PostMessage` and perform the suppress check inside `DoShow`, which already runs on the worker thread *after* the hook returns (so it no longer blocks input). `GetKeyState` is fine to keep in the hook — it's cheap.

  ```cpp
  // in the hook: just marshal, don't decide
  if (enabled) RequestToast(i, isOn);
  // in DoShow(): drop the toast if suppressed
  if (s.suppressFullscreen && IsFullscreenActive()) return;
  ```

- `GetStr()`'s `p ? p : L""` null check is unnecessary — `Wh_GetStringSetting` never returns `NULL` (it returns `L""` when unset/on error). You could also replace the raw get/free pair with `WindhawkUtils::StringSetting` (RAII), though since `Settings` stores owning `std::wstring`s, the current helper is reasonable as-is.

- In `StopWorker()` (which runs on the arbitrary `Wh_ModUninit` thread), `g_toasts[0].hwnd` is read while the worker thread can still be mutating `g_toasts` (the hotplug pool-grow in `DoShow`). It's benign in practice (the grow path is rare and only on monitor hotplug), but it's technically a data race on unload — caching the worker's HWND in a separate atomic, or always using the `PostThreadMessageW(g_workerThreadId, WM_QUIT)` fallback, would sidestep it.

- `#include <shlobj.h>` looks unused — `SHQueryUserNotificationState`/`QUERY_USER_NOTIFICATION_STATE` come from `shellapi.h`. Minor; doesn't affect compilation.

</p>
</details>

<p>

- `SystemAccentArgb()` uses `DwmGetColorizationColor`, which returns the DWM colorization color — that can differ from the "accent color" the user picked in Settings (the value other UI surfaces use). For an accent tint on a toast it's a fine approximation; just noting it isn't strictly the same value. The reliable per-user accent lives under `HKCU\...\Explorer\Accent` / the immersive color APIs if you ever want an exact match.

</p>
</details>

havrlisan:
Submission review issues addressed and fixed. Below is a summarized list of changes written by Claude.

---

Converted to a tool mod (required)
- @include explorer.exe → windhawk.exe, renamed Wh_ModInit/Wh_ModSettingsChanged/Wh_ModUninit → WhTool_*, and pasted the launcher boilerplate verbatim.
- One gotcha worth flagging: I also had to drop @architecture x86-64. windhawk.exe is 32-bit, so an x86-64-only mod never loads into the tool host and fails completely silently — no log, no toast. Removing the restriction (builds both, like theme-toggler-tray) fixed it. Might be worth a note on the wiki page.

Screenshot (required)
- Added a Preview section showing the Pill / Tile / Minimal layouts and Insert's single-value mode.

Optional improvements
- IsFullscreenActive() in the LL hook — moved out. The hook now only marshals (RequestToast); the suppressFullscreen && IsFullscreenActive() check runs in DoShow on the worker thread, off the input-blocking path.
- StopWorker data race — fixed. The primary toast HWND is now cached once at worker init, and StopWorker reads that instead of g_toasts[0].hwnd, so it can't race the hotplug pool-grow's vector reallocation. Kept the reliable window-message shutdown.
- #include <shlobj.h> — you were right it's wrong, with a twist: on the bundled MinGW toolchain SHQueryUserNotificationState / QUERY_USER_NOTIFICATION_STATE actually resolve from shellapi.h (verified), not shlobj.h. Replaced the include with <shellapi.h>.
- GetStr() null check — left as-is per your note; the guard is harmless defensive code since Settings owns its std::wstrings.

Functionality note
- SystemAccentArgb() / DwmGetColorizationColor — kept intentionally; the DWM colorization color is a fine approximation for a translucent toast tint. Added a comment noting it isn't the exact per-user accent.

PR #4540, opened 2026-06-24
Add One Key Keyboard Layout Switcher mod

Anixx:
This mod allows switching keyboard layouts with a single key press instead of a two-key combination. It includes settings for various keys and methods for layout switching.

## Changelog

* Changelog item 1...
* Changelog item 2...

## Mod authorship

This mod was created by:

- - [x] Claude
- - [ ] ChatGPT
- - [ ] Gemini


Anixx:
I re-submit the request because in the previous one I messed up with files (changed the wrong one).

The automated review was addressed, but switching the method from sending keys to pure API does not work in all programs. So, I keep the sending the sey combination method.

m417z:
### Submission review

---

- **The `Sleep(100)` runs on the hook-owning thread.** Because the `WH_KEYBOARD_LL` callback and the `WM_SWITCH_LAYOUT` handler share the same thread's message loop, while `SendLayoutSwitch` sleeps the low-level hook isn't being serviced, so keystrokes during that ~100 ms window may be delayed or pass through (the system applies `LowLevelHooksTimeout`). This also disappears with the `WM_INPUTLANGCHANGEREQUEST` approach.

<p>

- **Redundant unhook in `Wh_ModUninit`.** The worker thread already calls `UnhookWindowsHookEx` when it receives `WM_QUIT` and exits its loop, so the `UnhookWindowsHookEx(g_hook)` at the top of `Wh_ModUninit` is redundant and races with the worker thread on the non-atomic `g_hook` (the `NULL` guard makes it benign, but it's unnecessary). You can just post `WM_QUIT` and `WaitForSingleObject` for the thread to do the cleanup. If you keep an explicit teardown, prefer a bounded wait (e.g. `WaitForSingleObject(g_thread, 1000)`) over `INFINITE`, as `caps-ime-switcher` does.
- **RAII for the string setting.** `WindhawkUtils::StringSetting` (from `windhawk_utils.h`) replaces the manual `Wh_GetStringSetting` + `Wh_FreeStringSetting` pair in `LoadSettings`.
- **`memset(s_keyDown, 0, ...)` in `Wh_ModUninit` is unnecessary** — the array is a zero-initialized global and is re-created fresh each time the mod is loaded.

</p>
</details>

<p>

- **The `enableRightAlt: true` default consumes AltGr.** On many international layouts (German, Polish, Portuguese, etc.) the Right Alt key *is* AltGr, used to type `@`, `€`, `ł`, `@`, accented characters, and so on. Because an enabled key is fully blocked on both down and up, defaulting `enableRightAlt` to `true` breaks AltGr text entry out of the box for exactly the multi-layout users this mod targets. Consider defaulting it to `false` (Menu / Right Win are safer always-on defaults).
- **`Both` mode performs two different operations.** Ctrl+Shift and Alt+Shift are bound to two distinct Windows actions (switch keyboard *layout* vs. switch input *language*); if a user has both hotkeys enabled, `Both` triggers two switches in a row (with the `Sleep(100)` between), which can land back on the original layout or skip one. Switching via `WM_INPUTLANGCHANGEREQUEST` (item 2) sidesteps this entirely.
- **An enabled key loses all its other functions.** This is inherent to "single key = switch", but worth documenting: enabling Left Alt disables Alt+Tab / Alt+F4 / menu mnemonics on that key, and enabling Menu disables the context-menu key. Users may not anticipate it.

</p>
</details>

Anixx:
I fixed the issues.

Issue #2629, opened 2025-10-29
New version of Control Panel Color Fix does not work as intended.


Anixx:
The mod is by @chip33.

The previous version of the mod, 1.0 worked for me perfectly, but after the update to version 1.0.1, the CP sidebar is white:

<img width="744" height="493" alt="Image" src="https://github.com/user-attachments/assets/6d062aac-f79c-4824-9cbc-cb945e28bb95" />

Under version 1.0:

<img width="744" height="493" alt="Image" src="https://github.com/user-attachments/assets/6c0a7011-b549-473c-a201-d65a34404fea" />

chip33:
Hi, thanks for the report.

Version 1.0.1 introduced changes to address an incompatibility with StartAllBack. 

From your screenshot it doesn’t appear to be stock Windows 11, so if you’re running other customizations that interaction is likely what’s causing the difference compared to v1.0.

As [v1.0](https://github.com/ramensoftware/windhawk-mods/commit/b78d9f04997ec0b281fd2cdf88224bfbe50fb737) is working correctly for you, I’d recommend rolling back to that version for now. 

In the future, when I have some time, I may be able to prepare a test build with more detailed logging to pinpoint the failure point, but for that to be useful I’d need to know exactly which customizations you’re running.

Anixx:
It s stock Win11 wth Classc theme. You can enable Classic theme with the "Classc theme" mod.

Anixx:
Version 1.0:
<img width="744" height="493" alt="Image" src="https://github.com/user-attachments/assets/d187566f-c50a-4a04-90bc-3a0a8deb211b" />

Version 1.0.1 has no effect (the appearance is th same as with no mod):

<img width="744" height="493" alt="Image" src="https://github.com/user-attachments/assets/9e89d83d-c156-426d-8d3c-b7b6961a0109" />

m417z:
The reason I changed it to an enhancement is that, per the mod's author, classic theme is unsupported, and worked previously by accident. Therefore, adding proper support to it is an enhancement request.

chip33:
@Annix, 
the Control Panel Color Fix mod is primarily designed to support the many custom themes that use the custom .msstyles method.

Version 1.0 works nicely for your own [Classic Theme Enable](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/classic-theme-enable.wh.cpp)  mod because it just generally repaints the control panel header.

[...]

PR #4554, opened 2026-06-24
Merge with the mod Control Panel Color Fix version 1.0

Anixx:
Add code from the mod Control Panel Color Fix version 1.0 because the version 1.0.1 breaks the compatibility with Classic theme and the author is not going to fix it.

## Changelog

* Changelog item 1...
* Changelog item 2...

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/restore-folder-menubar-25h2.wh.cpp
+++ b/mods/restore-folder-menubar-25h2.wh.cpp
@@ -1,8 +1,8 @@
 // ==WindhawkMod==
 // @id              restore-folder-menubar-25h2
 // @name            ExplorerFrame and menubar fixes fork
-// @description     Fixes explorer problems: menu bar, listview redraw, classic background color
-// @version         2.0.1
+// @description     Fixes explorer problems: menu bar, listview redraw, classic background color, Control Panel header/sidebar color for dark themes
+// @version         3.0.0
 // @author          Anixx
 // @github          https://github.com/Anixx
 // @include         explorer.exe
@@ -14,11 +14,12 @@
 /*
 
 This is a fork of the mod `Explorerframe fixes for Win11 22H2+` by Waldemar made due to the author's apparent inactivity.
-It addresses several issues with the original mod.
+It addresses several issues with the original mod. Additionally, it incorporates the `Control Panel Color Fix` mod version 1.0 by chip33
+because the version 1.0.1 does not support Classic Theme.
 
 **The original mod description**
 
-This mod fixes a couple of issues with File Explorer windows appearance in Windows 11 releases 22H2, 23H2 and 24H2.
+This mod fixes a couple of issues with File Explorer windows appearance in Windows 11 releases 22H2 and later.
 
 Unpredictable menu bar appearance: On Windows 11 22H2 and 23H2, the menu bar may appear unpredictably. This mod allows the user to disable the menu bar entirely, or have it be shown at all times.
 You can force show/hide the menu bar in the mod settings. Forcing the menu bar to be shown is intended for the Classic theme.
@@ -29,7 +30,7 @@ Black artifacts briefly visible in new file explorer windows under Classic theme
 
 **Changes from the original**
 
-* This mod restores the menubar in recent builds of 24H2 and in 25H2 with which the original mod was not compatible. 
+* This mod restores the menubar in recent builds of 24H2 and in 25H2 with which the original mod was not compatible.
 **!Important!** On newer releases of Windows 11 24H2 and later, you may need to use the [ViVeTool](https://github.com/thebookisclosed/ViVe/releases) utility
 with the following command so to enable the menu:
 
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

---

This PR bundles the `Control Panel Color Fix` mod's `DirectUI::Element::PaintBackground` hook into the existing ExplorerFrame fork. The main concern is that the code copied in is the *older, less safe* `1.0` version, and the design of merging a separately-maintained mod into an unrelated one.

**1. Reintroduced null-deref crashes in explorer.exe (the copied code is missing the v1.0.1 safety guards).** The existing `control-panel-color-fix` mod is at version **1.0.1**, and its current code guards `value` and `pRect` before dereferencing them — see [`mods/control-panel-color-fix.wh.cpp#L172-L179`](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/control-panel-color-fix.wh.cpp#L172-L179). Those guards were almost certainly added because the author hit real null cases. The code copied here (the `1.0` version) drops both:

```cpp
if ((int)(*(DWORD*)value << 26) >> 26 != 9)   // <-- dereferences `value` with no null check
{
    ...
    FillRect(hdc, pRect, brush);               // <-- uses `pRect` with no null check
```

`PaintBackground` is a hot, core DirectUI entry point that fires for *every* element paint in explorer.exe, so a single `value == NULL` (or `pRect == NULL`) call faults the shell. This hook runs whether or not the user is in Control Panel, so the exposure is process-wide. Add the guards back:

```cpp
if (!value || !pRect) {
    Element_PaintBg(This, hdc, value, pRect, pClipRect, pExcludeRect, pTargetRect);
    return;
}
```

**2. Design: this duplicates a separately-maintained mod, and forks it backwards.** `control-panel-color-fix` is an active mod in the catalog (currently 1.0.1, author chip33). Copying its hook into an unrelated "ExplorerFrame and menubar fixes" mod creates two problems:

- **Duplication / conflict.** Both mods now hook `DirectUI::Element::PaintBackground` in `dui70.dll`. A user who enables both gets two hooks fighting over the same paint (yours returns early and suppresses the original, so the result is order-dependent). The maintainer's standing preference is to *not* fragment functionality across near-duplicate mods.
- **Regressing instead of fixing.** The stated rationale is that 1.0.1 broke Classic theme compatibility, so you reverted to 1.0. But 1.0.1 also contains the null guards above *and* the `ThemeAltersCPBody` check that avoids StartAllBack conflicts and redundant repaints ([`#L194-L195`](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/control-panel-color-fix.wh.cpp#L194-L195)). Reverting to 1.0 throws those away to regain Classic-theme support. The cleaner path is to take the *current* 1.0.1 logic and add Classic-theme handling on top (e.g. a setting), and ideally pursue that as an issue/PR on `control-panel-color-fix` itself rather than copying it in here.

If the maintainer is fine with the bundle as a personal fork, at minimum keep the v1.0.1 safety/StartAllBack logic rather than the v1.0 code — there's no reason to lose the crash fixes along with the behavior change you actually want.

Human comment: I know it was discussed at #2629, I'm OK with merging it, but it makes sense to take the actual fixes and improvements from the newer version.

<p>

- **Redundant include.** `#include <windhawk_api.h>` is already pulled in by `windhawk_utils.h`; the explicit include is unnecessary.
- **Hook is installed even when the feature is disabled.** When `ControlPanelColorFix` is off you still `LoadLibraryExW(L"dui70.dll", ...)` and install the `PaintBackground` hook, then bail at runtime via `g_settingControlPanelColorFix`. Since `Wh_ModSettingsChanged` doesn't request a reload, the runtime flag check is needed for live toggling, so this is a reasonable trade-off — just noting the dui70 load + hook are paid unconditionally.
- **`LoadLibraryW` consistency.** The pre-existing `LoadLibraryW(L"explorerframe.dll")` / `LoadLibraryW(L"shell32.dll")` calls use the default search order, while the new dui70 load correctly uses `LOAD_LIBRARY_SEARCH_SYSTEM32`. Since this mod is scoped to `explorer.exe` (only ever runs from System32), the hijack surface is negligible, but for consistency these could use `GetModuleHandleW` (both are already loaded in explorer at init) or the same `LOAD_LIBRARY_SEARCH_SYSTEM32` flag.
- **README has no screenshot for the new effect.** The Control Panel color change is a visible effect; consider adding a before/after image (chip33's mod has good examples).

</p>
</details>

<p>

- **`ThemeAltersCPBody` guard removed.** The copied 1.0 code always repaints when the element matches (`v45 == 4`), whereas 1.0.1 only repaints when the theme actually alters the Control Panel body color. Always repainting is presumably intentional (it's what restores Classic-theme behavior), but it can conflict with StartAllBack and repaint on default/light themes where it isn't wanted.
- **Unpainted region on theme-query failure.** When `v45 == 4`, the hook `return`s (suppressing the original) even if `OpenThemeData`/`GetThemeColor` fails, so nothing gets painted in that case. chip33's 1.0.1 falls back to `RGB(32,32,32)` and always fills. Consider falling through to the original (or a fallback fill) when the theme query fails.
- **Narrower scope than the source.** `control-panel-color-fix` includes `systemsettings.exe` in addition to `explorer.exe`; this fork is `explorer.exe`-only, so any Control Panel/Recovery pages that render in `systemsettings.exe` won't be recolored. If that's intended, fine — just flagging the coverage difference.
- **Per-paint theme/brush allocation.** `OpenThemeData` + `CreateSolidBrush` + `CloseThemeData`/`DeleteObject` run on every matching paint. It's gated behind cheap bitfield checks so the common path is fast, but the matched path could cache the resolved color/brush and invalidate on `WM_THEMECHANGED` if you want to trim it.
- **Reverse-engineered field offsets.** The `(*(DWORD*)value << 26) >> 26` and `*((unsigned __int64*)value + 1)` decoding is undocumented dui70 layout (inherited from the source mod). It's x64-specific, which matches `@architecture x86-64`, but it may need revisiting on future Windows builds if the layout shifts.

</p>
</details>

Anixx:
I am not interested in StartAllBack logic, I am interested in Classic Theme compatibility and the author ignored the bug report half a year ago. In this mod the fix is optional, so anyone can disable the feature here and use that mod of version 1.0.1 in parallel.

m417z:
No problem, you can skip StartAllBack. The main concern is the stability improvements.

Anixx:
Added the check for those two pointers if they are null.

PR #4558, opened 2026-06-25
Lock Keys Notifier v1.0.1

havrlisan:
## Changelog

* Pre-warm GDI+ text pipeline at startup to fix cold-boot first-toast delay

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/lock-keys-notifier.wh.cpp
+++ b/mods/lock-keys-notifier.wh.cpp
@@ -1184,6 +1184,27 @@ static LRESULT CALLBACK LowLevelKeyboardProc(int code, WPARAM wParam, LPARAM lPa
     return CallNextHookEx(nullptr, code, wParam, lParam);
 }
 
+// Pre-warm the GDI+ text/font pipeline on a throwaway thread. The first
+// RenderToast pays a one-time lazy cost — font enumeration, loading the
+// configured family, spinning up the GDI+ text subsystem and pulling in its
+// DLLs — which on a cold boot (cold disk) can be several seconds. GDI+ is
+// initialized per-process and the caches it warms are process-wide, so warming
+// on any thread benefits the worker's first real RenderToast. Crucially this
+// runs OFF the worker thread: that thread owns the WH_KEYBOARD_LL hook and must
+// keep pumping messages, or every system-wide keystroke stalls on the
+// LowLevelHooksTimeout until the hook returns. Renders into a scratch DIB and
+// discards it; nothing is presented.
+static DWORD WINAPI WarmupThreadProc(LPVOID) {
+    Settings s;
+    EnterCriticalSection(&g_settingsCs);
+    s = g_settings;
+    LeaveCriticalSection(&g_settingsCs);
+    ToastWindow warm;
+    RenderToast(warm, s, KI_Caps, true, 96);
+    if (warm.dib) DeleteObject(warm.dib);
+    return 0;
+}
+
 static DWORD WINAPI WorkerThreadProc(LPVOID) {
     GdiplusStartupInput gsi;
     GdiplusStartup(&g_gdiplusToken, &gsi, nullptr);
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

---

The change is small and the goal is sound — moving the one-time GDI+/font init off the first key press is a real improvement, and the scratch `RenderToast` is clean (it never touches `tw.hwnd`, and `DeleteObject(warm.dib)` is the right cleanup). One thing is worth reconsidering:

**The warm-up blocks the thread that owns the system-wide low-level keyboard hook.** The hook is installed at `WorkerThreadProc` line 1207 and `g_workerReady` is signaled at line 1213, but the warm-up runs at lines 1223–1231 — *before* the message pump at line 1234. `WH_KEYBOARD_LL` callbacks are dispatched on the installing thread and require it to be retrieving messages; while it's blocked inside `RenderToast`, the hook can't be serviced. By your own comment that cost can be "several seconds" on a cold boot — which is exactly when this process starts (right after sign-in). During that window:

- Lock-key toggles produce no toast (minor — consistent with the mod's existing "missed toggle is fine, the next one is correct" design).
- More importantly, **all** keyboard input on the system passes through this hook. With the owning thread unresponsive, every keystroke incurs the `LowLevelHooksTimeout` (~300 ms by default) before the OS proceeds without the hook, so there's a brief window of degraded/laggy keyboard input system-wide, and Windows can decide a persistently unresponsive LL hook should stop being called.

Note this cost is now paid **unconditionally on every startup** (even when the user has disabled every `notify*` key, or never touches a lock key), whereas before it was paid lazily and only when a toast was actually needed. So while the change improves the first-toast latency, it trades that for an every-boot window of degraded global input.

The clean fix is to keep the hook installed early (so the ready handshake still reflects hook status) but run the warm-up on a **separate, short-lived thread** so the worker's pump — and thus the LL hook — becomes responsive immediately. GDI+ is initialized per-process and the font/DLL caches it warms are process-wide, so warming on any thread benefits the worker's first real `RenderToast`. Spawn it after `GdiplusStartup`, and join its handle before `GdiplusShutdown` in the worker's teardown so it can't touch GDI+ after shutdown. Sketch:

```cpp
static DWORD WINAPI WarmupThreadProc(LPVOID) {
    Settings s;
    EnterCriticalSection(&g_settingsCs);
    s = g_settings;
    LeaveCriticalSection(&g_settingsCs);
    ToastWindow warm;
    RenderToast(warm, s, KI_Caps, true, 96);
    if (warm.dib) DeleteObject(warm.dib);
    return 0;
}
// ... after SetEvent(g_workerReady), before the pump:
HANDLE warmThread = CreateThread(nullptr, 0, WarmupThreadProc, nullptr, 0, nullptr);
// ... after the GetMessage loop exits, before GdiplusShutdown:
if (warmThread) { WaitForSingleObject(warmThread, INFINITE); CloseHandle(warmThread); }
```

<p>

- The comment says "Render into a scratch window and discard it" but nothing is windowed here — `RenderToast` only builds a scratch DIB (it doesn't use `tw.hwnd`). Reword to "scratch DIB" to avoid implying a window is created/destroyed.
- The warm-up always runs even when every `notify*` setting is off. If you keep it inline (or even on a separate thread), you could skip it when no key is enabled — e.g. `if (s.notifyCaps || s.notifyNum || s.notifyScroll || s.notifyInsert)` — so a user who has disabled all notifications doesn't pay the startup cost (and global-input blast, if kept inline) for work that will never render.

</p>
</details>

<p>

- The warm-up renders at a fixed 96 DPI. That's fine for warming the expensive one-time work (DLL loads, the GDI+ text subsystem, loading the configured font family — all size/DPI-independent), but on a high-DPI monitor the first *real* toast still builds `Font` objects at the scaled em size, so a small incremental cost remains on the first toggle. Not worth changing — the bulk of the cold-boot cost is already covered — just noting it isn't a 100% warm-up.

</p>
</details>

havrlisan:
Good catch on the LL-hook starvation — that's a real every-boot regression, not just a missed toggle. Fixed in c8a5839.

The warm-up now runs on a throwaway thread spawned right after the worker signals ready, so the worker's message pump (and thus WH_KEYBOARD_LL servicing) goes live immediately. As you noted, GDI+ is per-process initialized and the font/DLL caches are process-wide, so warming from another thread still benefits the worker's first real RenderToast. The worker joins the warm-up thread before GdiplusShutdown in teardown so it can't touch GDI+ after shutdown.

Also took both optional improvements:
- Reworded the comment to say "scratch DIB" (no window is created).
- Skip the warm-up entirely when no notify* key is enabled — a user with all notifications off no longer pays the startup cost for work that would never render.

Left the 96-DPI warm-up as-is per your functionality note; the bulk of the cold-boot cost (DLL loads, text subsystem, font-family load) is DPI-independent and covered.

PR #4559, opened 2026-06-25
Resource-Aware Task Terminator

insane66613:
Adds Resource-Aware Task Terminator v1.0.1.

Validation:
- Windhawk PR validator passed
- git diff --check passed
- Compiled successfully for x86, x64, and ARM64

Changes:
- PID reuse hardening with creation-time/path revalidation
- Safer protected-process handling
- Sortable chooser and result summary
- Compatibility-preserving settings keys

### DIFF
--- a/mods/resource-aware-task-terminator.wh.cpp
+++ b/mods/resource-aware-task-terminator.wh.cpp
@@ -1,817 +1,154 @@
 // ==WindhawkMod==
 // @id              resource-aware-task-terminator
 // @name            Resource-Aware Task Terminator
-// @description     Terminate the active window process, a top resource offender, or a weighted resource-heavy app group with confirmation and sortable usage metrics.
-// @version         1.0.0
+// @description     Terminate the active window process, a top resource offender, or a weighted resource-heavy app group with verified process identity and sortable usage metrics.
+// @version         1.0.1
 // @author          Math Shamenson
 // @github          https://github.com/insane66613
 // @license         MIT
 // @include         windhawk.exe
-// @compilerOptions -lpsapi -lgdi32 -lcomctl32
+// @compilerOptions -std=c++20 -lpsapi -lgdi32 -lcomctl32 -lshell32
 // ==/WindhawkMod==
 
 // ==WindhawkModReadme==
 /*
-Resource-Aware Task Terminator
+# Resource-Aware Task Terminator
 
-- Press the configured hotkey to terminate a target process or app group.
-- Modes:
-  - foreground: Kill the active window process.
-  - top_cpu: Sample CPU over a configurable window and show the top PID candidates.
-  - top_pagefaults: Sample page-fault deltas and show the top PID candidates.
-  - top_disk_io: Sample process read/write byte deltas and show the top PID candidates.
-  - aggregate_super_score: Group eligible processes by executable path/name and rank app groups by CPU, page faults, disk I/O, and memory.
-  - top_gpu_experimental: Disabled in this no-PDH build to avoid loader dependency failures.
+A global-hotkey process terminator that runs in a dedicated Windhawk tool process.
+It can target:
 
-Defaults
-- Hotkey: Ctrl + Alt + F4. Pick a different hotkey if another app or Windows reserves it.
-- Protected apps: Conservative defaults including explorer.exe, windhawk.exe, taskmgr.exe, conhost.exe, dwm.exe, and critical system processes.
+- The process that owns the foreground window.
+- The highest CPU, page-fault, or read/write I/O offender measured over a sample window.
+- An executable group ranked by a weighted aggregate of CPU, page faults, I/O, and working-set memory.
 
-Notes and Safety
-- You may need to run Windhawk elevated to kill elevated targets; otherwise OpenProcess(PROCESS_TERMINATE) can fail with Access Denied.
-- Do not remove critical system processes from the protected list. Killing them can destabilize Windows.
-- Terminating an app can cause data loss. Save work first.
-- Confirmation mode is enabled by default and shows the selected process/app group, mode, sampling evidence, and protection check before termination.
-- The chooser supports sortable metric columns, severity-colored metric cells, manual refresh, and optional live refresh.
+## Safety model
 
-Troubleshooting
-- If the hotkey does not work, pick a different combo. Some Win+ combos or hotkeys taken by other apps will fail to register.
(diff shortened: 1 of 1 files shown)


insane66613:
﻿## Implementation and safety details

The current head is `50c94c8609e59b0bfb8f29a5eeff4aa6ba7892c4`.

### Process identity and termination safety

- Captures PID, process creation time, and normalized full executable path for every target.
- Requires matching creation time and path between sampling snapshots before calculating deltas.
- Revalidates creation time and path on the same handle used for `TerminateProcess`.
- Treats unqueryable or identity-mismatched processes as protected/skipped rather than killable.
- Checks the return value of `TerminateProcess` and reports real success/failure counts and error codes.
- Terminates only the PIDs displayed and confirmed; newly spawned processes are not silently added to aggregate groups.

### Sampling correctness

- Adds per-metric validity flags so failed API reads are not interpreted as zero counters.
- Prevents lifetime counters, unsigned underflow, or PID reuse from fabricating top offenders.
- Handles the cumulative 32-bit page-fault counter with modulo subtraction only after process-instance verification.
- Removes unnecessary `PROCESS_VM_READ`, improving candidate visibility without requesting excess access.

### Chooser and threading

- Moves sampling/live refresh off the chooser and hotkey message threads.
- Preserves selection by stable process identity across refresh and re-sort.
- Invalidates stale candidates after empty/failed refreshes.
- Makes chooser creation failure cancel safely instead of selecting row zero.
- Reposts `WM_QUIT` from the nested chooser loop.
- Keeps hotkey re-registration and settings reload responsive while the chooser is open.

### UI, settings, and documentation

- Populates the full Windhawk settings schema and README.
- Preserves existing public setting names/defaults.
- Fixes protected-app parsing so only surrounding whitespace is trimmed; spaces inside executable names remain intact.
- Renames the display from physical “Disk” activity to process read/write I/O transfer counts.
- Populates sortable metric columns and corrects rank-column sorting.
- Documents CPU percentage semantics, aggregate-group behavior, elevated-process limitations, confirmation behavior, and the disabled GPU mode.

### Build verification

The repository compiler script completed successfully for:

- `i686-w64-mingw32`
- `x86_64-w64-mingw32`
- `aarch64-w64-mingw32`

Compiler options include explicit C++20, Psapi, GDI32, Comctl32, and Shell32 linkage.

m417z:
It can't be merged, please resolve the branch conflicts.

insane66613:
Resolved the merge conflicts and updated this branch against the current upstream `main`.

Changes made:

- Kept the hardened Resource-Aware Task Terminator **v1.0.1** implementation when resolving the add/add conflict.
- Preserved the PR as a single-file update to `mods/resource-aware-task-terminator.wh.cpp`.
- Removed the unnecessary explicit `-std=c++20` compiler option, which was being reported as a manual-verification warning and caused the PR validation workflow to fail.
- Retained the required link options, including `-lshell32`.
- Revalidated the source and rebuilt it successfully for **x86, x64, and ARM64**.

Current results:

- PR mod validation: **passed**
- Windhawk compatibility 1.6.1: **passed**
- Windhawk compatibility 1.7.3: **passed**
- GitHub merge state: **CLEAN / MERGEABLE**

Current head: `0a876be7cd73e3d02c62cb66cc038058b8e9ebb8`

m417z:
### Submission review

---

- **Mojibake in the chooser "no data" placeholder.** The em-dash used for missing metrics is double-encoded in the source — the bytes are `C3 A2 E2 82 AC E2 80 9D`, i.e. a UTF-8 `—` that was round-tripped through Windows-1252 and re-saved, so the literal is actually three characters `â€"`. It appears in four places:

```cpp
// lines ~1255, 1259, 1261, 1263
return decision.cpuValid ? FormatDouble(decision.cpuPercent) : L"â€”";
...
: L"â€”";
return decision.ioValid ? FormatMb(decision.ioBytes, true) : L"â€”";
return decision.memoryValid ? FormatDouble(decision.memoryMb) : L"â€”";
```

The file is still valid UTF-8 (so CI's encoding check passes), but every cell without a valid metric will render garbled `â€"` in the list — and missing metrics are common (many processes show zero I/O delta over a sample, etc.), so users will see this regularly. Replace these with a proper `—` (or a plain `-` / `N/A`) and make sure your editor saves as UTF-8 without re-encoding.

- **Tool-mod launcher block** is a reworded version of the wiki snippet (renamed locals, `reinterpret_cast`, `swprintf_s`/`GetModuleFileNameW` instead of the snippet's `switch`-on-`GetModuleFileName`). It's functionally equivalent, but for easier review, paste the snippet verbatim:

<p>

- **Dead GPU scaffolding.** `KillMode::TopGpuExperimental`, its `ParseKillMode` / `KillModeName` / `FormatPrimaryMetric` cases, and the "GPU ranking is disabled in this no-PDH build" message are all leftovers from a removed PDH build. The value isn't even offered in the `kill_mode` `$options`. Unless you plan to restore it, removing the whole enum value and its handling would simplify the file.
- **Use `WindhawkUtils::StringSetting` (RAII)** instead of raw `Wh_GetStringSetting` + manual `Wh_FreeStringSetting` for `kill_key`, `kill_mode`, and `protected_apps`. While there, note `Wh_GetStringSetting` never returns `NULL` (it returns `L""`), so the `if (keyStr)` / `if (modeStr)` / `if (appsStr)` guards are dead — `ParseKeyCode`/`ParseKillMode` already treat empty input as the default.

</p>
</details>

<p>

- **Bare-key hotkey footgun.** If a user sets both `disable_ctrl` and `disable_alt` (and no Shift/Win), `g_modifiers` ends up as just `MOD_NOREPEAT`, so `RegisterHotKey` binds a bare key (e.g. F4) system-wide and swallows it in every application. The defaults are safe (Ctrl+Alt+F4); consider refusing to register, or warning, when no real modifier remains.
- **Per-monitor DPI.** The chooser is sized/scaled to the system DPI at creation (`EffectiveDpi()` falls back to the desktop DC) and doesn't handle `WM_DPICHANGED`, so it can look slightly off if moved to a monitor with a different scale factor. Low impact for a transient dialog.
- **Sampling cost.** Each sample opens every process twice (before/after the interval) plus `GetProcessMemoryInfo`/`GetProcessIoCounters`. That's inherent to the toolhelp + delta approach you've chosen and there's no clean alternative without PDH/ETW, so this is just an FYI — the work is bounded and cancellable, which is good.

</p>
</details>

PR #4580, opened 2026-06-26
Add Windows 11 Minimal Start Menu mod

tchack:
Adds a new Windhawk mod: Windows 11 Minimal Start Menu.

The mod customizes the redesigned Windows 11 Start menu on Windows 11 build 26200.8655 and newer by hiding the All Apps block while keeping search and pinned apps available.

Validation:
- Local mod functionality and metadata were checked before submission.
- The PR contains a single new file: `mods/windows-11-minimal-start-menu.wh.cpp`.
- The mod metadata includes `@github https://github.com/tchack` as required by the Windhawk mods repository.

### DIFF
--- a/mods/windows-11-minimal-start-menu.wh.cpp
+++ b/mods/windows-11-minimal-start-menu.wh.cpp
@@ -1,0 +1,2024 @@
+// ==WindhawkMod==
+// @id              windows-11-minimal-start-menu
+// @name            Windows 11 Minimal Start Menu
+// @description     Hides the All Apps section in the redesigned Windows 11 Start menu on Windows 11 build 26200.8655 and newer while keeping search and pinned apps.
+// @version         1.5
+// @author          Gaika
+// @github          https://github.com/tchack
+// @include         StartMenuExperienceHost.exe
+// @architecture    x86-64
+// @compilerOptions -lole32 -loleaut32 -lruntimeobject
+// @license         MIT
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Windows 11 Minimal Start Menu
+
+This mod customizes the redesigned Windows 11 Start menu in
+`StartMenuExperienceHost.exe`. It targets the new Start menu in Windows 11
+build 26200.8655 and newer.
+
+Features:
+- Hides the All Apps block in the new UWP/XAML Start menu.
+- Keeps the search box and pinned apps visible.
+- Hides the "All" heading and the view selector row.
+- Renames the pinned apps heading to a custom text.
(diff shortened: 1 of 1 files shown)


m417z:
Thanks for the submission. How is it different from https://windhawk.net/mods/pinned-only-on-start-menu? And why can't it be a theme for Windows 11 Start Menu Styler?

tchack:
Thanks for checking.

The main difference from `pinned-only-on-start-menu` is the approach and the target Start menu version.

`pinned-only-on-start-menu` uses registry policy values (`NoStartMenuMorePrograms` and `HideRecommendedSection`) by hooking `RegQueryValueExW`. My mod targets the redesigned Windows 11 Start menu in build 26200.8655+ and works directly with the XAML tree/layout. It hides the new All Apps area and related controls, keeps search and pinned apps usable, and also forces the compact `SmallStart` visual state so the pinned grid width stays consistent across monitors. It also has settings for allowing the native large layout and for changing the pinned heading text.

Regarding Windows 11 Start Menu Styler: part of this can probably be expressed as a theme, especially the simple XAML element hiding/styling rules. The reason I submitted it as a separate mod is that it also contains behavior which is not just styling:

- hooks `VisualStateManager::GoToState` to force `SmallStart` instead of `LargeStart`;
- detects/handles the new Start menu feature state via `RtlQueryFeatureConfiguration`;
- contains layout-specific logic for the redesigned Start menu, not only style/resource overrides;
- includes settings that affect behavior, not only visual appearance.

That said, if you think the preferred direction is to make this a Windows 11 Start Menu Styler theme, I can rework it that way and keep only the parts that are possible as a theme. The main thing I wanted to solve is reliable behavior on the new redesigned Start menu, where the policy-based approach doesn't cover the layout issues.

m417z:
> That said, if you think the preferred direction is to make this a Windows 11 Start Menu Styler theme, I can rework it that way and keep only the parts that are possible as a theme.

That's preferable, both for maintenance, and for user experience. Let's go that route and see what capabilities can't be achieved with styling alone.

tchack:
Sounds good, let's do that.

I'll try to convert the visual/layout part into a Windows 11 Start Menu Styler theme first, and keep the separate mod only as a fallback if some behavior can't be achieved with styling alone.

The main things I'll check are:

- hiding the redesigned All Apps area and related controls;
- keeping search and pinned apps usable;
- moving/adjusting the remaining layout so it doesn't leave empty space;
- whether the compact `SmallStart` behavior can be approximated without hooking `VisualStateManager`;
- whether the pinned header text can be handled as a theme/style change or should be dropped.

I'll prepare the theme version and report back with what works and what doesn't.

PR #4505, opened 2026-06-20
Add MacOS Minimize Animation

Abdullah-Masood-05:
Adds macOS-style genie minimize and restore (open) animations to every window. The mod hooks ShowWindow and DefWindowProcW, snapshots the window, and renders a per-frame warped layered "ghost" synced to the DWM compose cycle (DwmFlush) for vsync-aligned smoothness. Includes an optional reverse genie on restore and an opt-in **experimental** app-launch animation that is off by default. Setting: animation duration. A demo GIF is included in the readme.

PR #4581, opened 2026-06-26
Add MacOS Minimize Animation

Abdullah-Masood-05:
Adds macOS-style genie minimize and restore (open) animations to every window. The mod hooks ShowWindow and DefWindowProcW, snapshots the window, and renders a per-frame warped layered "ghost" synced to the DWM compose cycle (DwmFlush) for vsync-aligned smoothness. Includes an optional reverse genie on restore and an opt-in experimental app-launch animation that is off by default. Setting: animation duration. A demo GIF is included in the readme.

### DIFF
--- a/mods/macos-minimize-animation.wh.cpp
+++ b/mods/macos-minimize-animation.wh.cpp
@@ -1,0 +1,686 @@
+// ==WindhawkMod==
+// @id              macos-minimize-animation
+// @name            MacOS Minimize Animation
+// @description     Smooth macOS-style genie minimize and restore (open) animations for every window.
+// @version         2.0.0
+// @author          Abdullah Masood
+// @github          https://github.com/Abdullah-Masood-05
+// @include         *
+// @compilerOptions -ldwmapi -lgdi32
+// @license         MIT
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# MacOS Minimize Animation
+
+![Demo](https://raw.githubusercontent.com/Abdullah-Masood-05/my-windhawk-mods-media/main/macos-minimize-animation.gif)
+
+Brings the classic macOS **genie** effect to Windows. When you minimize a window
+it warps and flows down into the taskbar; when you restore it, it flows back out.
+
+Eye candy for something you do a hundred times a day.
+
+## See it in action
+- Compile the mod with the button on the left or with Ctrl+B.
+- Enable the mod and open any normal window (Explorer, a browser, Notepad...).
+- Minimize it with the title bar `[-]` button or `Win`+`Down` and watch it warp
+  and pour into the taskbar.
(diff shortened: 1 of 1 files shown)


Abdullah-Masood-05:
Hi @m417z  this one's been quiet so just checking in. Checks are green, no conflicts, and there's nothing outstanding that I can see. It replaces #4505 (which I closed). Happy to make any changes if something's needed, thanks for all the reviewing you do here.

m417z:
Hi, 

Recently, several Genie and other minimize/restore animation mods were submitted:
* Genie Animation Mod v2.0.0 (abandoned): https://github.com/ramensoftware/windhawk-mods/pull/4243
* Fluid Window Engine Pro by @kivsak: https://github.com/ramensoftware/windhawk-mods/pull/4295
* Animation & Wobbly Mod by @shoaibhassan2: the PR and the GitHub user were removed, but here's an archived version: https://gist.github.com/m417z/1d5f68e81d7e3096d6c36f2a91af4db7

Your mod looks like an improvement to Genie Animation Mod, but I'm worried about a situation of many similar submissions, each doing a slight change to the original mod. Something similar happened to Taskbar Music Lounge:
https://gist.github.com/m417z/334ee4371449592faa4575aa2a1bbfc4

* How does the mod compare to the [Genie Animation Mod](https://windhawk.net/mods/genie-minimize-animation)?
* Are you planning on maintaining it? 
* Did you plan on adding additional animations?

Abdullah-Masood-05:
Hi @m417z, thanks for looking at this and for the context on the other submissions, the fragmentation concern makes sense, and the Taskbar Music Lounge example is a fair worry.

Answering your three questions directly:

**How it compares to the Genie Animation Mod:**
The core difference is the transition itself. Instead of a fade or simple effect, mine does a true frame-captured genie: on mouse-down it captures the current frame and shows it in a persistent overlay so there's no visible instant jump, the window state changes behind the overlay, then it animates from the old frame to the new one. The result is a smoother, more macOS-accurate genie rather than jump-then-animate, and it's fully configurable (duration, restore animation, etc.). I'm happy to record a side-by-side against the existing mod if that helps you judge the difference.

**Beyond minimize/restore:**
Yes it also has an app-launch animation that plays the genie when a window first opens, working across most apps. It's opt-in and off by default since it's experimental, so the default behavior stays clean.

**Maintenance and additional animations:**
Yes to both, and I can point to a track record rather than just saying it. I actively maintain my other published mod, [Task Manager Tab Slide Animation](https://github.com/ramensoftware/windhawk-mods/pull/4594) that PR is the v0.3.3 update you merged, which reworked the tab transition into a frame-captured crossfade, and I'm pushing another update to it right now. I intend to maintain this genie mod the same way, including adding more animations over time.

Thanks again glad to make any changes you think would improve it.

PR #4584, opened 2026-06-26
Add Disk Usage Bar Customizer mod

Valer100:
This mod allows customization of the disk usage bar in the File Explorer, including theme-aware colors, height, and border settings.

## Changelog

Not the case here, this is a new mod.

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini


m417z:
### Submission review

---

The mod is well-written and clean, with good screenshot coverage. A few things to look at:

**The hook does expensive work on every `DrawThemeBackground` call before the cheap discriminator.** `DrawThemeBackground` is an extremely hot path in `explorer.exe` — it fires for every themed element paint. On *each* call your hook unconditionally calls `GetThemeClass(hTheme, ...)` (a call into uxtheme) and `GetModuleHandleEx(..._FROM_ADDRESS, ...)` (a `RtlPcToFileHeader`-style lookup), and only afterward checks the part id. The vast majority of calls are not part 5/11, so this overhead is pure waste on all of Explorer's theme rendering. Gate on the cheapest check (part id) first and only then do the expensive lookups:

```cpp
HRESULT WINAPI HookedDrawThemeBackground(
    HTHEME hTheme, HDC hdc, INT iPartId, INT iStateId, LPCRECT pRect, LPCRECT pClipRect
) {
    // Cheapest discriminators first — bail out for everything that can't be a drive bar.
    if ((iPartId != 5 && iPartId != 11) || !pRect || pRect->left <= 0)
        return DrawThemeBackground_orig(hTheme, hdc, iPartId, iStateId, pRect, pClipRect);

    HMODULE callerModule = nullptr;
    BOOL isCallerShell32 = GetModuleHandleEx(
        GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS | GET_MODULE_HANDLE_EX_FLAG_UNCHANGED_REFCOUNT,
        reinterpret_cast<LPCWSTR>(__builtin_return_address(0)), &callerModule) && callerModule == g_shell32;

    WCHAR themeClass[256] = {};
    BOOL isThemeClassValid = GetThemeClass &&
        SUCCEEDED(GetThemeClass(hTheme, themeClass, 256)) && wcscmp(themeClass, L"Progress") == 0;

    if (isCallerShell32 && isThemeClassValid) {
        ...
    }
    return DrawThemeBackground_orig(hTheme, hdc, iPartId, iStateId, pRect, pClipRect);
}
```

This also fixes a latent uninitialized-read: as written, `themeClass` is never initialized and `GetThemeClass`'s return value is ignored, so if the call ever fails the later `wcscmp(themeClass, ...)` reads garbage.

**`useSystemAccentColor` description references a setting that doesn't exist.** Its `$description` says *"Not appliable when "Adapt to system's theme" is disabled"*, but there is no "Adapt to system's theme" option — the mod always follows the system theme. Remove that sentence to avoid confusing users.

<p>

- **Remove `-lmsimg32` from `@compilerOptions`.** It's carried over from the original mod (which used `AlphaBlend`); this version draws only with `FillRect`/`CreateSolidBrush` (gdi32), so msimg32 is unused. `-luxtheme` and `-lgdi32` are both needed.
- **Dead NULL check / prefer RAII in `LoadColorSetting`.** `Wh_GetStringSetting` never returns `NULL` (it returns `L""` on unset/error), so `if (!hexString) return fallback;` is dead — the empty string is already handled by the `wcslen != 6` check below. Consider `WindhawkUtils::StringSetting` (RAII) instead of the manual `Wh_GetStringSetting` + `Wh_FreeStringSetting` pairs.
- **Cache the dark-mode flag.** `AreAppsUsingDarkTheme()` does a fresh `RegGetValueW` (open+query+close) on every qualifying bar paint. Reading it once into a global and refreshing it in `Wh_ModSettingsChanged` / on a theme-change message would avoid the per-paint registry hit. Alternatively, throttle the reads.
- **Limit the `hTheme = g_darkHTheme` swap to the visual-styles path.** The swap at the top of the block runs unconditionally, but the dark theme handle is only ever used when `g_renderUsingVisualStyles` is true. In custom-rendering mode the swap can leak into the part-5 fallback (`return DrawThemeBackground_orig(hTheme, ...)` for states other than 2/4), drawing with the wrong theme. Moving the swap inside the `if (g_renderUsingVisualStyles)` branches makes the intent clearer and avoids that edge case.

</p>
</details>

<p>

- **The `caller == shell32` + `pRect->left > 0` heuristics are build-dependent.** As your own comment acknowledges, these are fragile. On builds where the drive-list bar is drawn from a different module (e.g. `windows.storage.dll` rather than `shell32.dll`), `isCallerShell32` would be false and the mod would silently do nothing. Worth keeping in mind for future Windows updates; nothing to change now if it's tested working on current builds.
- **Warning-threshold percentage is derived from pixel widths.** `usedPercentage = progressWidth * 100 / g_barWidth` relies on (a) part 11 always being drawn before part 5 so `g_barWidth` is current, and (b) the full bar being redrawn rather than a clipped sub-rect — `clipRect` is `pRect ∩ pClipRect`, so a partial invalidation would shrink `progressWidth` and skew the percentage (and the remaining-space calc). In practice these small bars usually repaint whole, so it mostly holds, but it's less robust than the original mod's reliance on `iStateId` alone. The trade-off is understandable since the OS only exposes a binary full/partial state and you need a custom threshold.

</p>
</details>

Valer100:
Done! I've addressed the issues you pointed out (except the dark mode check on every bar draw).

m417z:
It looks like you didn't address the first note too. Any reason for that?

Also see the two notes below. Any reason you removed the `IsDriveListDC` check from the original mod?

---

**The bar-detection heuristic is fragile where a cleaner option exists.** You flagged this yourself in the comment. The `Progress` theme-class check and the `__builtin_return_address` → shell32 check are reasonable, but the load-bearing exclusion for **copy/move progress dialogs** is `pRect->left > 0`, which is the weak link — those dialogs' progress bars are also drawn by shell32 with the `Progress` class, so if their rect ever starts at `left > 0` (layout/DPI/version differences) they'll get recolored. The mod you forked solves exactly this with a window-ancestry filter instead:

```cpp
static bool IsDriveListDC(HDC hdc) {
    HWND hwnd = WindowFromDC(hdc);
    if (!hwnd) return true;               // memory DC — fall through to your other checks
    WCHAR cls[256];
    for (HWND cur = hwnd; cur; cur = GetAncestor(cur, GA_PARENT))
        if (GetClassNameW(cur, cls, 256) && wcscmp(cls, L"CabinetWClass") == 0)
            return true;
    return false;
}
```

([source](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/disk-usage-bar-color.wh.cpp#L155)). The drive list lives under `CabinetWClass` while the copy dialog does not, so this distinguishes them directly. The one caveat is the memory-DC case (`WindowFromDC` returns `NULL`) — which is likely why you didn't use it — but you can combine the two: require the `CabinetWClass` ancestor when a window is found, and keep the shell32-caller check as the fallback for the `NULL` case. That removes the fragile `left > 0` guess.

- **`EnumWindows` doesn't filter by process.** `RefreshExplorerCallback` `SendMessage`s every `CabinetWClass` window on the desktop, including ones owned by *other* `explorer.exe` instances (each of which already runs this mod and its own refresh). Add a `GetWindowThreadProcessId(hwnd, &pid); pid == GetCurrentProcessId()` check to keep it in-process.

<p>

- **`WindhawkUtils::StringSetting` (RAII)** in `LoadColorSetting` instead of the manual `Wh_GetStringSetting` + `Wh_FreeStringSetting` on every return path.
- **Magic numbers** — `iPartId` `5`/`11` and `iStateId` `2`/`4` are `PP_FILL`/`PP_TRANSPARENTBAR` and `PBFS_ERROR`/`PBFS_PARTIAL` from `vsstyle.h`; using the named constants would make the part/state logic readable.
- **`AreAppsUsingDarkTheme()` does a `RegGetValueW` on every qualifying paint.** Consider caching the value and refreshing it on `WM_SETTINGCHANGE`/`WM_THEMECHANGED` rather than hitting the registry per bar repaint.
- **`warningPercentageThreshold` isn't clamped** the way `heightFactor` is — a value `> 100` never warns and `<= 0` always warns. Probably worth clamping to `1..100`.

</p>
</details>

<p>

- **`g_barWidth` is captured during the `PP_TRANSPARENTBAR` (11) draw and reused in the `PP_FILL` (5) draw**, so the used-percentage math depends on the background always painting first and at full width in the same repaint. On a partial-region repaint (only the fill invalidated) `g_barWidth` is stale, and on the very first fill paint before any background it's the initializer `1`, which makes `usedPercentage` huge and forces the warning color. Harmless cosmetically and self-corrects on the next full paint, but worth being aware of.
- **The "drawn by shell32" assumption is version-dependent.** The This-PC drive tiles are currently drawn from shell32, but the rendering module has moved around across Windows 11 builds; if a future build draws the bar from a different module the hook will silently stop matching. Nothing to do now — just a heads-up for future breakage reports.

</p>
</details>

Valer100:
> It looks like you didn't address the first note too. Any reason for that?

I did. I just moved the shell32 caller check inside the `if (isRectValid && isThemeClassValid) {` block. Is calling `GetThemeClass` expensive that I really need to check for the requested parts to be drawn before calling it?

<br>

> The bar-detection heuristic is fragile where a cleaner option exists. You flagged this yourself in the comment. The Progress theme-class check and the __builtin_return_address → shell32 check are reasonable, but the load-bearing exclusion for copy/move progress dialogs is pRect->left > 0, which is the weak link — those dialogs' progress bars are also drawn by shell32 with the Progress class, so if their rect ever starts at left > 0 (layout/DPI/version differences) they'll get recolored. The mod you forked solves exactly this with a window-ancestry filter instead

I've tried it, but it isn't working. The bars from the file move/copy/delete dialogs also get affected by the mod. I tried logging both cases, but both disk usage bars and file operation dialog's progress bar fall under the memory DC case. That's why I dropped that function and I improvised the check. The `pRect->left > 0` is still a valid concern though, but, as I stated in my comment, I have no idea for a better check.

image
image
image

Perhaps the original mod author didn't notice this, because their mod was just checking for the part 5 in the 2nd and the 4th state (you don't see the 2nd (error) state inside the file operations dialog very often and when everything is ok it's using the first (normal) state, so maybe they assumed that their check was ok). The moment I added more customization options I found out that the checks from the original mod were not enough.

<br>

> EnumWindows doesn't filter by process. RefreshExplorerCallback SendMessages every CabinetWClass window on the desktop, including ones owned by other explorer.exe instances (each of which already runs this mod and its own refresh). Add a GetWindowThreadProcessId(hwnd, &pid); pid == GetCurrentProcessId() check to keep it in-process.

Will do that, thank you for pointing this out.

m417z:
> Is calling `GetThemeClass` expensive that I really need to check for the requested parts to be drawn before calling it?

It's surely more expensive to call it than not, and skipping it most of the time with an integer check is easy. It's not a deal-breaker, but that's what the review point mentioned, including a snippet demonstrating the cheap check.

Valer100:
Done

PR #4589, opened 2026-06-27
Add JonaOS Glass Widgets (Draggable) mod

SeniorMajor7:
## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini
- - [x] Other (Codex): 

### DIFF
--- a/mods/jonaos-glass-widgets-draggable.wh.cpp
+++ b/mods/jonaos-glass-widgets-draggable.wh.cpp
@@ -1,28 +1,37 @@
 // ==WindhawkMod==
-// @id              jonaos-glass-widgets-graggable
-// @name            JonaOS Glass Widgets (Draggable)
-// @description     Achieve that Premium Glass morph widgets and elevate your Windows Display to look Clean, Professional, High-end and Futuristic.
-// @version         3.4
+// @id              jonaos-glass-widgets-draggable
+// @name            JonaOS Glass Widgets
+// @description     Adds draggable glass-style desktop widgets for calendar, clock, media, volume, battery, search, personalization, network, Bluetooth, and accessibility shortcuts.
+// @version         3.5
 // @author          Jona like It, Code It
 // @github          https://github.com/SeniorMajor7
-// @include         explorer.exe
-// @compilerOptions -lcomctl32 -lgdi32 -luxtheme -ldwmapi -lgdiplus -lole32 -luuid -lshell32 -lshlwapi
+// @include         windhawk.exe
+// @compilerOptions -lcomctl32 -lgdi32 -luxtheme -ldwmapi -lgdiplus -lole32 -luuid -lshell32
 // @license         MIT
 // @architecture    x86-64
 // ==/WindhawkMod==
 // ==WindhawkModReadme==
 /*
-// JonaOS Liquid Glass Widgets
-// "Achieve that Premium Glass morph widgets and elevate your Windows Display to look Clean, Professional, High-end and Futuristic".
-// The light mode works best on DARK WALLPAPERS only.
-//"<img width="478" height="376" alt="Image" src="https://github.com/user-attachments/assets/afa6c355-4a19-40d8-b055-a3a5a7d7b2a5" /> "Light mode"
-//<img width="479" height="370" alt="Image" src="https://github.com/user-attachments/assets/f0fdf3c9-e95d-4701-92ee-165c981f4eda" /> "Dark mode"
-//<img width="475" height="362" alt="Image" src="https://github.com/user-attachments/assets/a90c3cf5-ee5a-4d24-90ab-1e53ffa7ef61" /> "Acrylic/Translucent"
-//<img width="475" height="369" alt="Image" src="https://github.com/user-attachments/assets/497b525c-ea6f-4d2e-b113-e97c1bd3c5d8" /> "Mica"
-//<img width="484" height="363" alt="Image" src="https://github.com/user-attachments/assets/358af692-960e-4269-91a5-bf1acc3a8e0b" /> "Tabbed Mica"
-//<img width="476" height="371" alt="Image" src="https://github.com/user-attachments/assets/38353d6d-321c-4331-b4c5-224b68cf18f7" /> "No DWM Background"
-//" Change the Appearance mode. Corner roundness. Switch off any widget"
-// "It is better to leave all settings in default since they are designed perfectly except for the appearance mode which the user can change according to their tastes and preferance
+JonaOS Glass Widgets adds a group of draggable desktop widgets to Explorer's
+desktop layer. The widgets include:
+
+- Calendar
+- Now Playing media controls
+- Clock
+- Volume
+- Battery
+- Windows Search and Personalization shortcuts
+- WiFi, Hotspot, Bluetooth, and Accessibility shortcuts
+
+Each widget can be enabled or disabled individually. You can also adjust the
+position, size, appearance mode, opacity, corner roundness, font, and colors.
+
+The automatic layout places the widgets on the selected desktop side when a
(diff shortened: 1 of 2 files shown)


m417z:
### Submission review

---

**1. No message loop — the widgets don't actually work when enabled normally (blocking).** The mod creates its windows on whatever thread runs `Wh_ModInit`, then returns. There is no `GetMessage`/`DispatchMessage` loop and no dedicated UI thread anywhere in the mod (confirmed: no `CreateThread`/`std::thread`/`GetMessage`). When a mod is enabled on an **already‑running** `explorer.exe` (the normal case — explorer is always running when you install/toggle a mod), `Wh_ModInit` runs on the *Windhawk Engine thread* (see the [Mod lifetime](https://github.com/ramensoftware/windhawk/wiki/Mod-lifetime) flow chart), which goes back to waiting for events and does not pump your windows' messages. Consequences:
   - `SetTimer(hwnd, TIMER_ID, …)` never delivers `WM_TIMER`, so the clock/calendar/volume/battery never refresh after the first paint.
   - No `WM_MOUSEMOVE`/`WM_LBUTTONDOWN`/`WM_LBUTTONUP` are dispatched, so **dragging** (the headline feature), the volume slider, and the media/utility buttons do nothing.
   - On the less common startup‑injection path, the windows are created on explorer's *main* thread but then `Wh_ModUninit` runs on the engine thread, so `DestroyWindow` (which must be called by the creating thread) fails and the windows leak on unload.

   The correct pattern for a windowed mod is a dedicated thread that owns the windows and runs its own message loop. [mods/neko-cat.wh.cpp](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/neko-cat.wh.cpp#L1389) — a desktop‑pet mod almost identical in spirit — does exactly this: it spawns a thread, creates its windows there, and runs `while (... GetMessage(&msg, ...)) { TranslateMessage(...); DispatchMessage(...); }`. This is best solved together with item 2.

**2. This should be a "mod as a tool", not an `explorer.exe` injection.** The mod installs **no function hooks and no symbol hooks** — it only creates windows, sets timers, calls `EnumWindows`/`FindWindow`, and uses COM (audio) and `SendInput`, all of which work from any process. It also "ignores its host process's specific state and just operates on the desktop as a whole." Those are precisely the signals listed in [Mods as tools: Running mods in a dedicated process](https://github.com/ramensoftware/windhawk/wiki/Mods-as-tools:-Running-mods-in-a-dedicated-process), whose own example list even includes *Taskbar Music Lounge* (a media‑controls widget). Running this inside `explorer.exe` means a GDI+/COM mistake destabilizes the whole shell, and you get **duplicate widget sets** when there is more than one `explorer.exe` process (the mod has no single‑instance guard). Recommended fix: convert to a tool mod — `@include windhawk.exe`, rename `Wh_ModInit`/`Wh_ModSettingsChanged`/`Wh_ModUninit` → `WhTool_*`, paste the launcher snippet from the wiki verbatim (see [mods/explorer-folder-hover-menu.wh.cpp](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/explorer-folder-hover-menu.wh.cpp#L3778) for a working copy), and run the windows + message loop on a dedicated thread as in `neko-cat`. This subsumes item 1.

**3. README images are on a disallowed host.** All six screenshots use `https://github.com/user-attachments/assets/...`. Only `i.imgur.com` and `raw.githubusercontent.com` are accepted image hosts (these are mirrored by Windhawk after merge); `github.com/user-attachments` is not, so the images won't display in the catalog. Please re‑host the screenshots on imgur or in a repo under `raw.githubusercontent.com`.

**4. Dead settings — declared, read, but with no effect.**
   - `LayoutGap` is loaded into `g_layoutGap` (line 758) but never used anywhere in the layout math (the offsets use hard‑coded `+30`/`+26`/`ACTION_WIDGET_GAP`). Either wire it into the automatic layout or drop the setting.
   - Every `*LabelColor` setting (`CalendarLabelColor`, `ClockLabelColor`, …) is parsed in `LoadWidgetConfig` (lines 733‑738) and then **unconditionally overwritten** by `ApplyPerWidgetDefaultLabelColors` at the end of the same function, so the user's value is always discarded. Either honor the parsed color (only fall back to the theme/black default when blank) or remove these settings.

**5. "0/1 means default" overrides make valid values unreachable.** In `LoadSettings`, `WidgetOpacity == 1` is forced back to `255` and `WidgetRoundness == 0` is forced back to `26` (and similarly `LayoutTop`/`LayoutGap` 0→default). So a user can't actually set a near‑transparent opacity (1) or square corners (roundness 0) — the very edges of the documented ranges silently snap to the default. Use a sentinel that isn't a legal value (e.g. treat `< 0` as "use default", or add an explicit "use default" toggle) instead of overriding a valid in‑range value.

- **The mod id/filename is misspelled: `graggable` → `draggable`.** The `@id` (and therefore the mod's permanent URL) is `jonaos-glass-widgets-graggable`. The id can't be changed after publishing, so it's worth fixing now.

- **README formatting.** Inside the `==WindhawkModReadme==` block every line is prefixed with `//` and several captions sit outside the `` tags with stray quotes, so the rendered markdown shows literal `//` and broken captions. Drop the `//` prefixes and use normal markdown image syntax.

<p>

- **Unused includes and libraries.** `<wbemidl.h>` and `<comdef.h>` are included but no WMI/`_com_error` code is used; `<uxtheme.h>` / `-luxtheme` and `-lcomctl32` have no corresponding API calls. Remove them (CI doesn't catch *unused* deps).
- **`ApplyPerWidgetDefaultLabelColors` is over‑built.** Its `if` enumerates all 11 widget types, but since that covers every possible type the whole branch is always true — it reduces to `labelColor = isLight ? RGB(0,0,0) : ThemeLabelColor(appearance)`. (This is also the function clobbering the `*LabelColor` settings in item 4.)
- **`GetEndpointVolume` returns `TRUE` even on failure.** It only returns `false` when `CoCreateInstance` fails; if the later `GetMasterVolumeLevelScalar`/`GetMute` calls fail it still returns `TRUE` with a 0 value, so callers can't tell a real 0% from an error.

</p>
</details>

<p>

Non‑critical observations and ideas about the feature behavior itself.

- **"Now Playing" doesn't show what's playing.** It draws the static strings "Media controls" / "Track metadata is not available in this Windhawk context.", and Play just launches the *first audio file found in the Music library* via `ShellExecute` and then sends a global `VK_MEDIA_PLAY_PAUSE`. That's surprising behavior for a "Now Playing" widget. The Windows `GlobalSystemMediaTransportControls` (WinRT media session) API gives the real current track/title/thumbnail and proper play/pause/next state if you want this widget to do what its name implies.
- **The Battery widget renders as a slider** (`DrawSlider`) that visually looks draggable, but only the Volume widget actually handles slider input — battery isn't (and shouldn't be) adjustable. Consider a non‑slider representation (bar/ring) so it doesn't look interactive.
- **Per‑second repaint + per‑move COM.** Every widget gets a 1‑second timer that triggers a full repaint (the analog clock has no second hand, so a 1s tick is more than it needs), and the Volume widget re‑creates an `IMMDeviceEnumerator` via `CoCreateInstance` on every paint *and* on every `WM_MOUSEMOVE` while dragging the slider. Caching the `IAudioEndpointVolume` endpoint and using a coarser refresh would cut a lot of churn — minor, but easy.

</p>
</details>

m417z:
The mod doesn't work.
* Remove `// @architecture    x86-64`
* Test it to make sure it actually works

Also:
* On my computer, the widget edges are black and not transparent.
* I suggest organizing the settings into groups, right now it's difficult to navigate and use.

m417z:
The corners still aren't transparent for me. I assume it's not supposed to be this way, right?

image

SeniorMajor7:
close all pull requests

PR #4590, opened 2026-06-27
New mod: borderless-fullscreen

TomberWolf:
## Changelog

* Changelog item 1...
* Changelog item 2...

## Mod authorship

This mod was created by:

- - [x] Claude
- - [ ] ChatGPT
- - [ ] Gemini
- - [x] Other (please specify): Code by me; Claude handled debugging and optimization.


m417z:
### Submission review

---

A clean, self-contained mod with a clear purpose, but the architecture and reversibility need rework before merge.

**1. This should be a tool mod, not `@include *`.** The mod injects its DLL into *every* process on the system just so a few target processes can manage their own windows — a real stability and performance cost (it loads into AV, system services, every background process). The tell-tale signs that it belongs in a dedicated process are all present: it installs **no function hooks**, and every API it uses (`EnumWindows`, `GetWindowThreadProcessId`, `GetWindowLongPtr`/`SetWindowLongPtr`, `SetWindowPos`, `GetSystemMetrics`) works **cross-process** — you don't need to be inside the game to restyle/resize its window; the window manager handles those calls across process boundaries (this is exactly how stand-alone tools like Borderless Gaming operate). Rewrite it as a tool mod:
- `@include windhawk.exe`, rename `Wh_ModInit`/`Wh_ModSettingsChanged`/`Wh_ModUninit` → `WhTool_*`, and paste the launcher boilerplate verbatim from the wiki.
- In the single dedicated process, `EnumWindows` over the **whole desktop**, resolve each window's owning exe via `GetWindowThreadProcessId` → `OpenProcess(PROCESS_QUERY_LIMITED_INFORMATION)` → `QueryFullProcessImageNameW`, and apply borderless to the ones whose exe is in the list.
- The exe-name-from-window pattern is already used in [`force-kill-active-window.wh.cpp`](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/force-kill-active-window.wh.cpp#L143). Reference: [Mods as tools](https://github.com/ramensoftware/windhawk/wiki/Mods-as-tools:-Running-mods-in-a-dedicated-process) and the launcher at the bottom of [`explorer-folder-hover-menu.wh.cpp`](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/explorer-folder-hover-menu.wh.cpp).

This also lets you drop the per-process worker thread and the `IsCurrentProcessATarget` / `GetCurrentExeName` plumbing entirely.

**2. The effect isn't reversible.** A core Windhawk principle is that a mod's effects disappear when it's disabled — here they don't. `ApplyBorderless` strips the window styles and resizes, but nothing ever restores them:
- Disabling the mod leaves the window permanently borderless and resized; the user has to restart the app.
- Worse, because `Wh_ModSettingsChanged()` is the no-reload `void` variant, removing an exe from the target list while it's running does **not** stop the worker — it has no `IsCurrentProcessATarget` re-check, so it keeps re-applying borderless to a window the user just untargeted.

Save each window's original `GWL_STYLE`/`GWL_EXSTYLE` and its `RECT` (keyed by `HWND`) the first time you touch it, then restore them in `Wh_ModUninit` and whenever a window/target stops matching. NoMoreBorder restores the border when you toggle it off; this mod should too.

**3. The 5-second re-apply is unconditional.** Each pass calls `SetWindowLongPtr` (twice) + `SetWindowPos(..., SWP_FRAMECHANGED)` on every matched window even when it's *already* borderless and correctly sized, forcing a non-client frame recalc/repaint every 5 seconds. This is in direct tension with the README's own argument: you avoid hooking window creation precisely because resizing a running game's window can disturb its swap chain — yet you re-issue `SetWindowPos` on that same running window every 5 seconds regardless. Guard the work so it only fires when the window isn't already in the desired state, e.g.:

```cpp
LONG_PTR style = GetWindowLongPtrW(hWnd, GWL_STYLE);
if (!(style & (WS_CAPTION | WS_THICKFRAME))) {
    // already borderless — also verify the rect before resizing, then skip
    return;
}
```

(An event-driven approach removes the need for the periodic sweep almost entirely — see Functionality notes.)

<p>

- **`WaitForSingleObject(g_workerThread, 2000)`** in `Wh_ModUninit` uses a 2-second timeout, after which `CloseHandle` runs while the thread may still be executing inside the (about-to-be-unloaded) mod image — a small crash window if the wait is ever exceeded. Since the worker polls `g_stopWorker` every 100 ms it normally exits fast; just wait `INFINITE` to be safe.
- **`g_customWidth` / `g_customHeight`** are plain `int`s written by `LoadSettings` (settings-change thread) and read by the worker thread without synchronization. Benign in practice but technically a data race — make them `std::atomic<int>` (or fold the size logic into the worker). Only matters on a settings change.
- **Stripping `WS_MINIMIZE | WS_MAXIMIZE`** is unusual — those are window *state* bits, not decoration. If the intent is to un-maximize before resizing, `ShowWindow(hWnd, SW_RESTORE)` is the clearer way to do it; the decoration bits you actually want gone are already covered by `WS_CAPTION | WS_THICKFRAME`.

</p>
</details>

<p>

- **Always the primary monitor.** `GetSystemMetrics(SM_CXSCREEN/SM_CYSCREEN)` and the `0,0` origin mean a game running on a secondary monitor gets yanked onto the primary one. The description does say "primary monitor", so this may be intended, but using the monitor the window is currently on (`MonitorFromWindow` + `GetMonitorInfo`, `rcMonitor`) would be friendlier on multi-monitor setups — and avoids DPI surprises, since `SM_CXSCREEN` is reported in the host process's DPI context.
- **`IsEligibleMainWindow` can match more than the main window.** The heuristic (visible + has title + ≥200×150) will also catch a launcher window, a secondary tool window, etc., and fullscreen-stack them all at `0,0`. Consider targeting only the foreground/largest window per process, or letting the heuristic be tightened.
- **Polling vs. events.** Instead of a 5-second sweep, `SetWinEventHook` (e.g. `EVENT_OBJECT_SHOW` for new windows, `EVENT_OBJECT_LOCATIONCHANGE`/`EVENT_OBJECT_STATECHANGE` to catch a game resetting its own style) reacts immediately and does no work when nothing changes. It works from any process, so it pairs naturally with the tool-mod design in item 1. You can keep your deliberate "let the window settle" delay by arming a short one-shot timer from the event rather than acting on it instantly.

</p>
</details>

PR #4603, opened 2026-06-28
New mod: vmware-audio-fix

steveji2021:
## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/vmware-audio-fix.wh.cpp
+++ b/mods/vmware-audio-fix.wh.cpp
@@ -1,0 +1,174 @@
+// ==WindhawkMod==
+// @id              vmware-audio-fix
+// @name            VMware Audio Distortion Fix
+// @name:zh-CN      VMware 音频失真修复
+// @name:fr-FR      Correction de distorsion audio VMware
+// @description     Fixes distorted audio in VMware virtual machines by setting high-resolution timers.
+// @description:zh-CN 通过设置高分辨率定时器修复 VMware 虚拟机中的失真音频。
+// @description:fr-FR Corrige la distorsion audio dans les machines virtuelles VMware en utilisant des temporiseurs à haute résolution.
+// @version         1.0
+// @author          Starfield
+// @github          https://github.com/steveji2021
+// @include         vmware-vmx.exe
+// @architecture    x86-64
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# VMware Audio Distortion Fix
+
+### This project is inspired by [VMAudioBackHost](https://github.com/Raymai97/VMAudioBackHost). Thanks to Raymai97
+
+Fixes distorted audio in VMware virtual machines by setting a high-resolution
+timer in the vmware-vmx.exe process.
+
+## Background
+
+Windows 10 build 2004+ made timer resolution per-process. VMware virtual
+machines may experience distorted audio because the vmware-vmx.exe process
+uses the default low timer resolution (15.625 ms). This mod injects into
+vmware-vmx.exe and calls `NtSetTimerResolution` to set a high-resolution
+timer, fixing the audio stuttering/distortion.
+
+## Usage
+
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

---

- **The timer-resolution change is never reverted — the mod isn't reversible.** A core Windhawk principle is that a mod's effect must disappear when it's disabled. `NtSetTimerResolution(..., TRUE, ...)` registers a *process-wide* resolution request that is tied to the process, not to the mod DLL — unloading the mod does **not** release it. So after the user disables the mod, `vmware-vmx.exe` keeps running at the high resolution (and the elevated power cost) until the VM process restarts. `Wh_ModUninit` currently only logs. Release the request there:

```cpp
void Wh_ModUninit(void)
{
    // restore the process default by cancelling our request
    HMODULE ntdll = GetModuleHandleW(L"ntdll.dll");
    if (ntdll) {
        auto pNtSetTimerResolution = (NTSTATUS (WINAPI*)(ULONG, BOOLEAN, ULONG*))
            GetProcAddress(ntdll, "NtSetTimerResolution");
        if (pNtSetTimerResolution) {
            ULONG cur;
            pNtSetTimerResolution(g_desiredTimerRes100ns, FALSE, &cur);
        }
    }
}
```

- **Use-after-free in `LoadSettings`.** `resStr` is freed at the `Wh_FreeStringSetting(resStr)` call, then dereferenced again by the `Wh_Log(..., %s ..., resStr, ...)` call two lines later — the `%s` reads freed memory. It only triggers when logging is enabled (logging is off by default in production), but it's still a genuine UAF. Move the log above the free, or — cleaner — switch to the RAII `WindhawkUtils::StringSetting` (see optional improvements), which removes the manual free entirely.

- **Target `vmware-vmx-debug.exe` as well**. See https://github.com/Raymai97/VMAudioBackHost/issues/9.

<p>

- **Use `WindhawkUtils::StringSetting` (RAII)** instead of the manual `Wh_GetStringSetting` + `Wh_FreeStringSetting` pair. It frees automatically and naturally eliminates the use-after-free above:
  ```cpp
  WindhawkUtils::StringSetting resStr = WindhawkUtils::StringSetting::make(L"timerResolution");
  double resMs = ParseMsString(resStr.get());
  ```
- **`ParseMsString` reinvents float parsing.** The values all come from a constrained `$options` dropdown, so a hand-rolled parser is unnecessary — `wcstod(resStr, nullptr)` does the same thing.
- **Unused `#include <stdlib.h>`** — nothing from it is used; remove it.
- **Log strings reference the wrong name.** The logs say `"VMAudioBackHost initializing"` / `"... settings changed"` / `"... uninitializing"` — that's the *inspiration* project, not this mod. Windhawk also already prefixes every log line with this mod's name, so the name is redundant in the string regardless; drop it (`Wh_Log(L"Initializing")` etc.).
- **`NtSetTimerResolution` second parameter should be `BOOLEAN`, not `BOOL`.** The native signature is `NTSTATUS NtSetTimerResolution(ULONG, BOOLEAN, PULONG)` (1-byte `BOOLEAN`). Passing `TRUE` happens to work, but match the canonical type as done in [timer-resolution-control.wh.cpp#L82](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/timer-resolution-control.wh.cpp#L82).

</p>
</details>

<p>

- **Lowering the resolution on a settings change may not take effect at runtime.** Within a single process, `NtSetTimerResolution(..., TRUE, ...)` effectively keeps the *finest* resolution requested — re-calling it with a *coarser* value (e.g. switching from 0.5 ms to 5.0 ms) often won't relax the timer until the request is released. If you adopt the release-on-uninit fix above, you can apply the same idea on settings change (release with `FALSE`, then re-request with `TRUE`) so coarser values take effect immediately rather than only after the VM process restarts.

</p>
</details>

steveji2021:
All mentioned issues are solved. Please check

PR #4615, opened 2026-06-29
Compact Start Menu

Asteski:
# Compact Start Menu

Cleans up the Windows 11 Start menu by showing only a compact, ungrouped All apps view. No header, no grouping, no gaps, just apps and folders, like it always should be.

## Changelog

* Changelog item 1...
* Changelog item 2...

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [x] ChatGPT
- - [ ] Gemini


m417z:
Thanks for the submission. Why can't it be a theme for Windows 11 Start Menu Styler?

Asteski:
@m417z 
From what I've discovered it required a custom logic that I don't think is possible with just a theme. 
Styler theme can handle the cosmetic parts, like hiding elements, but the actual compact/ungrouped grid requires code, because the gaps come from the grouped All apps ItemsSource / ICollectionView, not just the visible group header elements.
When only the headers are hidden via styles, the group layout still reserves the spacing. This mod removes the gaps by flattening the grouped collection into a flat observable vector through direct XAML hooks.
So a Styler theme can approximate the look, but a fully compact ungrouped grid would need custom logic in Start Menu Styler itself, not just YAML styles.

m417z:
If you look at the existing themes, you can see that many things are possible with just styles. Removing gaps should be one of those things. You're welcome to join [the Windhawk Discord server](https://discord.com/servers/windhawk-923944342991818753) to discuss this with the community.

I'm reluctant to merge it as a standalone mod that uses XAML Diagnostics, as it will conflict with the Windows 11 Start Menu Styler mod, and may cause confusion and bad experience.

Asteski:
Okay will try to do a theme out of it.

m417z:
### Submission review

Regarding `Wh_ModUninit`, if there's no easy way to restore the changes, you might want to use `ExitProcess(0);` like in `Wh_ModSettingsChanged`.

---

**1. Registered XAML callbacks are never removed → the host crashes after the mod is disabled/updated.**
The mod installs eight `RegisterPropertyChangedCallback` handlers (lines 284, 293, 302, 364, 396, 593, 603, 613) and one `ScrollViewer.ViewChanged` handler (line 773), and discards every returned token. `Wh_ModUninit` → `UninstallXamlTraversal` only stops the timers — it never unregisters any of these. The delegates' invoke thunks live in the mod DLL, but the XAML elements (and the WinRT delegate objects) outlive the mod, because Windhawk only removes its own function/symbol hooks on unload, not your WinRT registrations. So after the mod is disabled/updated/reloaded while `StartMenuExperienceHost.exe` keeps running, the very next `Visibility`/`Width`/`Height`/`Margin`/view change — or simply the eventual teardown of the element, which releases the delegate — calls into the now-unmapped DLL and access-violates the Start menu process. This is essentially guaranteed to fire at some point.

Fix: capture the token from each registration, store it next to the `weak_ref` you already keep (`g_keepCollapsedElements`, `g_headerTextElements`, `g_gapAdjustedItemsWrapGrids`, `g_hookedSemanticZooms`, `g_scrollBarModeScrollViewers`), and unregister them on the window thread inside `UninstallXamlTraversal` (which already runs via `RunFromWindowThread`). Use `element.UnregisterPropertyChangedCallback(prop, token)` and an `auto_revoke`/token-based revoke for `ViewChanged`. See how [windows-11-start-menu-styler.wh.cpp](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/windows-11-start-menu-styler.wh.cpp#L11282) stores and tears down its property-changed tokens.

**2. The mod doesn't revert when disabled (reversibility).**
`CollapseElement` (lines 236–273) overwrites `Visibility`/`Width`/`Height`/`Margin`/`Opacity`/… without saving the originals, and `FlattenAllAppsGridSource` swaps `ItemsSource` and clears `GroupStyle` (lines 555–559). `Wh_ModUninit` restores none of it, so disabling the mod leaves the Start menu collapsed and the app list flattened until the user manually restarts the process. A core Windhawk principle is that a mod's effects disappear when it's disabled. You already keep `entry.originalSource`, so restoring the grouped source is feasible; the collapsed elements would need their pre-change values saved to be restored. At minimum, restore what you can on the window-thread cleanup path.

**3. Consider the event-driven approach the other Start-menu XAML mods use instead of a 250 ms polling timer.**
`g_xamlTraversalTimer` walks the entire window content subtree plus every open popup every 250 ms, forever, even while the Start menu is closed (`ProcessCurrentXamlTree` → `ProcessXamlSubtree`). The established shell XAML mods subscribe to the XAML-diagnostics visual-tree service (`InitializeXamlDiagnosticsEx` → `IVisualTreeService3::AdviseVisualTreeChange`) and get a callback per element as it's added, apply a *reversible* customization, and unregister cleanly — see the `VisualTreeWatcher` in [windows-11-start-menu-styler.wh.cpp](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/windows-11-start-menu-styler.wh.cpp#L6688). Adopting it would address #1, #2, and the constant full-tree walk together (it's the root-cause fix rather than three separate patches). It's a substantial rewrite, so I'm flagging it as the recommended direction rather than a mechanical change — but the crash in #1 must be fixed regardless of which approach you keep.

**4. Add a screenshot/GIF, and fix the stale README title.**
This is a visual mod with no image in the README — please add a before/after screenshot or GIF (only `i.imgur.com` and `raw.githubusercontent.com` are allowed hosts). Also the README H1 still reads "Compact Start Menu Grid" (line 53) while `@name` is now "Compact Start Menu" — leftover from the file rename.

<p>

* **Dead `hideScrollBar` read (line 168):** `hideScrollBar` isn't declared in the settings block, and `scrollBarMode` always has a non-empty default, so `(!scrollBarMode || !*scrollBarMode)` is never true — this whole branch is unreachable. Remove it (looks like a leftover from an earlier settings layout).
* **`GetCurrentModuleHandle()` (line 176)** is defined but never called — dead code, remove it.
* **The `PCWSTR type` parameter** threaded through `HideStartMenuElement`, `FlattenAllAppsGridSource`, `IsGroupHeaderElement`, `IsAllAppsGridView`, `IsPinnedSectionElement` is always passed `nullptr` in the live path, so every `type && lstrcmpiW(...)` branch is dead. Looks like an artifact of an earlier type-string-based design; dropping the parameter and those branches would simplify a lot of code.
* **`Wh_GetStringSetting` never returns NULL** (it returns `L""` on unset/error), so the `headerText && *headerText` and `scrollBarMode && ...` NULL checks are unnecessary — just test for empty. You could also switch to `WindhawkUtils::StringSetting` (RAII) instead of the manual `Wh_GetStringSetting` + `Wh_FreeStringSetting`.
* **Default mismatch:** `Settings::scrollBarMode` defaults to `Show` in the struct but the setting default is `showWhileScrolling`. Harmless since `LoadSettings` always runs, but inconsistent.

</p>
</details>

<p>

* **Hide-category relies on an English string match.** `IsCategoryText` compares against the literal `L"Category"` (line 314), but that label is localized in the real UI, so `FindCategoryViewOptionTarget` silently won't hide the option on non-English Windows. The `DisableCategorySemanticZoom` path (disabling `CanChangeViews`) is language-independent and does most of the work, but if you rely on the text match too, document the limitation or match more robustly.
* **Hand-rolled scroll-bar fade.** "Show while scrolling" fades opacity in −0.2 steps on a 50 ms timer (`FadeAutoScrollBars`), which can look steppy compared to a XAML fade storyboard or just leaving `ScrollBarVisibility::Auto`. Works, just FYI.

</p>
</details>

m417z:
### Submission review

---

The mod works by polling the Start menu's XAML tree on a timer and force-collapsing / rebinding elements. The core idea is fine, but a few things affect stability and users of the mod:

**1. `Wh_ModSettingsChanged` force-terminates the whole process (`ExitProcess(0)`).** Every settings tweak kills `StartMenuExperienceHost.exe` (and tears down every other Windhawk mod loaded into it), then relies on Windows to restart it. The `LoadSettings()` call right before `ExitProcess(0)` is dead code — the process dies before it's used. The mod already has full apply (`InstallXamlTraversal`) and restore (`UninstallXamlTraversal`) machinery, so it should re-apply live instead of killing the process, e.g. marshal to the UI thread and do `UninstallXamlTraversal(); LoadSettings(); InstallXamlTraversal();`. `windows-11-start-menu-styler.wh.cpp` handles the same class of tree mutation cleanly in `Wh_ModSettingsChanged` by reverting/re-applying rather than exiting.

```cpp
    HWND coreWnd = GetCoreWnd();
    if (coreWnd) {
        RunFromWindowThread(coreWnd, [](PVOID) {
            UninstallXamlTraversal();
            LoadSettings();
            InstallXamlTraversal();
        }, nullptr);
    }
}
```

**2. Continuous 250 ms full visual-tree polling.** `g_xamlTraversalTimer` walks the entire visual tree (plus open popups) every 250 ms forever, calling `winrt::get_class_name` / name comparisons on every `FrameworkElement`, even while the Start menu is closed — a persistent CPU cost in a shell-adjacent process. It also produces a visible flash of the un-modified menu (headers, grouping, gaps) on each open, because collapsing lags the menu appearing by up to 250 ms.

Would it be possible to be more event-driven? e.g. process the XAML tree on window open, perhaps after mouse click, etc.

**3. `hideCategoryViewOption` matches the localized text "Category".** `IsCategoryText` compares the flyout item text against `L"Category"`, so on any non-English Windows UI language the menu item is never found and stays visible (the `SemanticZoom` disabling still works, so the behavior is partial/inconsistent). User-facing text is localized — match by element name / type rather than the visible string, or otherwise handle localization.

**4. Process-shutdown crash potential from global winrt objects with non-trivial destructors.** `g_xamlTraversalTimer`, `g_scrollBarHideTimer`, `g_scrollBarFadeTimer` (all `wux::DispatcherTimer`) and the global `std::vector`s that hold winrt refs (`g_flattenedSources`, `g_keepCollapsedElements`, …) run their destructors at `DLL_PROCESS_DETACH`. On a normal Start-menu restart (sign-out, update, crash) — and on the `ExitProcess(0)` path above, which you hit on *every* settings change — `Wh_ModUninit` is **not** called, so the CRT releases these XAML/COM objects during process shutdown from an arbitrary thread, which can crash the host. Prefer holding this state behind a heap pointer that's intentionally leaked at shutdown (trivial destructor) and keep the real teardown in `Wh_ModUninit`. See the note in the Windhawk docs about globals with non-trivial destructors at `DLL_PROCESS_DETACH`.

<p>

- **Use `WindhawkUtils::StringSetting`** (RAII) instead of the raw `Wh_GetStringSetting` + `Wh_FreeStringSetting` pairs in `LoadSettings` (`headerText`, `scrollBarMode`).
- **`CollapseElement` is heavier than it needs to be.** Setting `Visibility(Collapsed)` removes an element from layout; additionally zeroing `Width`/`Height`/`MinWidth`/`MaxWidth`/`Margin`/alignments/`Opacity`/`IsHitTestVisible` **and** registering three `RegisterPropertyChangedCallback`s per element to keep re-forcing them is a lot of machinery to maintain and revert. If `Collapsed` alone is insufficient for a specific container (e.g. a sticky `GridViewHeaderItem`), scope the extra work to that case rather than applying it to everything.
- **`headerText` has no visible effect under the defaults**: `hideTopLevelHeader` defaults to `true`, which collapses `AllListHeadingText`, so the header text you set is applied to an element that's hidden. Consider only applying/showing the setting when the header is visible.
- `g_xamlTraversalInstalled = false;` at the end of `Wh_ModUninit` is redundant — `UninstallXamlTraversal` already resets it.
- `$name: Change Header Text` uses title case; the rest of the settings use sentence case (e.g. "Hide top header").

</p>
</details>

<p>

- **The scroll-bar "show while scrolling" fade is a hand-rolled reimplementation** (a 200 ms hide timer + a 50 ms timer decrementing opacity by 0.2) of behavior XAML's `ScrollViewer` already provides via its auto-hide scrollbars. If the native `Auto` visibility gives an acceptable result, it would be simpler and avoid the extra timers.

</p>
</details>

PR #4620, opened 2026-06-30
v2.8.6: Added dark theme, Added German and Portuguese languages, and …

babamohammed2022:
This is an update for the mod.

## Changelog
- Added custom dark theme for the entire flyout UI
- Added support for German and Portuguese languages
- Implemented dark theme for password dialog and context menus
- Improved the UI layout with dynamic refresh button positioning
- Enhanced visual consistency with Windows 7 original design

## Mod authorship

This mod was created by:

- - [X] Claude
- - [ ] ChatGPT
- - [X] Gemini

### DIFF
--- a/mods/win7-network-flyout-recreation.wh.cpp
+++ b/mods/win7-network-flyout-recreation.wh.cpp
@@ -72,7 +72,6 @@ The mod has been tested on Windows 10 21H2, Windows 10 1809, Windows 11 23H2, Wi
 
 If you encounter issues, please report them on the author of the mod.
 */
-// ==/WindhawkModReadme==
 // ==WindhawkModReadme==
 // ==WindhawkModSettings==
 /*
--- a/mods/win7-network-flyout-recreation.wh.cpp
+++ b/mods/win7-network-flyout-recreation.wh.cpp
@@ -72,7 +72,7 @@ The mod has been tested on Windows 10 21H2, Windows 10 1809, Windows 11 23H2, Wi
 
 If you encounter issues, please report them on the author of the mod.
 */
-// ==WindhawkModReadme==
+// ==/WindhawkModReadme==
 // ==WindhawkModSettings==
 /*
 - language: auto
--- a/mods/win7-network-flyout-recreation.wh.cpp
+++ b/mods/win7-network-flyout-recreation.wh.cpp
@@ -2,12 +2,12 @@
 // @id             win7-network-flyout-recreation
 // @name           Windows 7 Network Flyout Recreation
 // @description    This mod recreates the Windows 7 network flyout for Windows 10 and 11
-// @version        2.8.6
+// @version        2.8.7
 // @author         babamohammed
 // @github         https://github.com/babamohammed2022
 // @include        explorer.exe
 // @architecture   x86-64
-// @compilerOptions -lgdi32 -ldwmapi -luxtheme -lole32 -lshell32 -luser32 -lcomctl32 -liphlpapi -lwlanapi -luuid -lpsapi -lmsimg32
+// @compilerOptions -lgdi32 -ldwmapi -luxtheme -lole32 -lshell32 -luser32 -lcomctl32 -liphlpapi -lwlanapi -luuid -lpsapi
 // ==/WindhawkMod==
 // ==WindhawkModReadme==
 /*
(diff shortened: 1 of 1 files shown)


babamohammed2022:
Hi, a new update has been released. If any fixes are required, let me know

m417z:
### Submission review

---

**`SetPreferredAppMode(ForceDark)` is applied process-wide and darkens *all* Explorer menus, not just the flyout.** `DarkContextMenu::OnSettingsChanged()` calls `Apply(g_Settings.theme == 1)`, which calls `pSetPreferredAppMode(AppMode::ForceDark)` — a process-global uxtheme preference. Once the dark theme is selected and settings change, every context menu in `explorer.exe` (right-click on the desktop, the taskbar, File Explorer windows) renders dark, even though the setting is described as "Select the network flyout's theme." It's also inconsistent: `DarkContextMenu::Init()` (called from `Wh_ModInit`) does *not* call `Apply`, so on a fresh load the global state isn't set, and the broad darkening only kicks in after the first settings change. `ShowContextMenu` already brackets the mod's own `TrackPopupMenu` with a transient `Apply(TRUE)` / `Apply(FALSE)` pair, which is the correct, scoped way to do this. Recommend dropping the persistent `Apply` in `OnSettingsChanged` (and keeping only the transient toggle) so the global preference is active only while the mod's own menu is up.

**The global `SetMenuInfo` hook never themes the mod's own menu and only affects unrelated Explorer menus.** The mod builds its context menu with `CreatePopupMenu` + `TrackPopupMenu` and never calls `SetMenuInfo` on it, so `SetMenuInfo_Hook` / `g_hMenuBrush` (the `0x2C2C2C` brush) are never exercised for the flyout's menu — its dark appearance comes entirely from `SetPreferredAppMode`. The only `SetMenuInfo` calls that flow through the hook are Explorer's own, whose `MIM_BACKGROUND` brush then gets overridden with the mod's dark brush whenever the dark theme is enabled. So this hook is dead for its intended purpose and is a process-wide side effect on menus the mod doesn't own. Recommend removing the `SetMenuInfo` hook (and `g_hMenuBrush`) entirely. If you specifically want the menu background to be `0x2C2C2C` rather than the default immersive-dark color, set `MIM_BACKGROUND` on the mod's *own* `hMenu` in `ShowContextMenu` instead of hooking the API globally.

<p>

Minor polish / dead-code cleanup — none of this affects users, so it's your call.

- **Dead localized-text matching in the flyout's `WM_CTLCOLORSTATIC`.** The new block reads the control's window text and matches substrings `L"nascondi"`, `L"Hide"`, `L"Ocultar"`, `L"Masquer"`, `L"Verstecken"`, `L"Скрыть"` to choose colors. The flyout has no "hide characters" static — that control lives in the password dialog, which has its own `WM_CTLCOLORSTATIC` handler. In the flyout this message only fires for the auto-connect checkbox, whose window text is empty (`""`), so the match never succeeds. It's dead code, and matching on localized UI strings is exactly the kind of fragile workaround to avoid — please drop it.
- **`WM_INITMENUPOPUP` → `FindWindowW(L"#32768", NULL)`.** This grabs the first menu-class window on the *whole desktop* (which can belong to another process) and applies `DarkMode_Explorer` to it. It's also redundant — `SetPreferredAppMode(ForceDark)` already makes the popup dark. Recommend removing it.
- **`ApplyNativeControlsTheme`'s connect-button recreate path is unreachable.** A theme change already triggers a full flyout recreation in `Wh_ModSettingsChanged` (`needRecreate` → `WM_SAFE_CLOSE` + `ToggleFlyoutWindow`), and the button is created with the correct `BS_OWNERDRAW`/`BS_PUSHBUTTON` style up front, so `g_ButtonConnectIsOwnerDraw != wantOwnerDraw` is never true when this runs. The whole `DestroyWindow` + `CreateWindowExW` block can be removed.
- **`-lmsimg32` is unnecessary.** `AlphaBlend` is the only msimg32 API used, and it's resolved dynamically via `LoadLibraryW(L"msimg32.dll")` + `GetProcAddress` (pre-existing code), so it's never linked directly. The newly-added `-lmsimg32` in `@compilerOptions` has no effect.
- **README typo:** "Portoguese" → "Portuguese" in the language-support feature line.
- Several new `static HBRUSH … = CreateSolidBrush(...)` brushes (e.g. `hBrushHideDark`, `hBrPwdStatic`) are created once and never freed — negligible one-time leaks, but the `lastBg`-tracked variants elsewhere already show the cleaner pattern if you want consistency.

</p>
</details>

<p>

- **Percentage-based refresh-button / checkbox repositioning.** The new `InitRefreshButtonRect` and the `WM_PAINT` refresh-button block move controls by hard-coded fractions of the window size (`+4%`, `-1.3%`, `-1.9%`, `-0.5%`, `-1.3%`, etc.) depending on whether the scrollbar is visible. These magic offsets are fragile and hard to reason about; if the goal is to avoid overlapping the scrollbar, deriving the position from the actual scrollbar width (`ScaleDpi(13)` / `GetSystemMetrics(SM_CXVSCROLL)`) would be sturdier. Behavior-only, so it's your call.

</p>
</details>

babamohammed2022:
@m417z I've updated the code, let me know if additional fixes are required

PR #4624, opened 2026-06-30
Update click-on-empty-explorer to v2.3.0

LiHua81:
## Summary

v2.3.0 update for Click on Empty Explorer mod.

**New features:**
- **Paste action** (Ctrl+V) — pastes into the current folder view
- **Custom Hotkey action** — each trigger (double/middle/double-middle click) gets an independent key combo field supporting multiple modifiers (Ctrl, Shift, Alt, Win) and all common keys
- Detailed format instructions per custom hotkey field

## Test Plan

- Set each trigger to Paste, Custom Hotkey, and existing actions — verify all work
- Multiple Explorer windows and tabs open — verify no crashes
- Change settings in Windhawk editor — verify immediate effect
- Use multi-modifier combos (Win+Shift+S, Ctrl+Shift+N)
- Test on Windows 11 with both SysListView32 and DirectUIHWND views

### DIFF
--- a/mods/click-on-empty-explorer.wh.cpp
+++ b/mods/click-on-empty-explorer.wh.cpp
@@ -2,7 +2,7 @@
 // @id              click-on-empty-explorer
 // @name            Click on Empty Explorer
 // @description     Configure double click, middle click and double middle click actions on empty space in File Explorer
-// @version         2.2.0
+// @version         2.3.0
 // @author          LiHua81
 // @github          https://github.com/LiHua81
 // @include         explorer.exe
@@ -62,7 +62,7 @@ no file or folder is located) and performs the action you've configured.
 
 When you choose "Custom Hotkey" for a trigger, a text field appears where you define the shortcut.
 
-**Format:** `Modifier + Modifier + ... + Key`
+**Format:** `Modifier+Modifier+...+Key` — spaces around `+` are optional and auto-trimmed
 
 - **Modifiers:** `Ctrl`, `Shift`, `Alt`, `Win` — can combine multiple, e.g. `Ctrl+Shift+N`, `Win+Shift+S`
 - **Keys:** letters (A-Z), digits (0-9), function keys (F1-F24), named keys (Tab, Enter, Space, Backspace, Delete, Escape, Left, Right, Up, Down, Home, End, PageUp, PageDown, Insert)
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

---

One blocking issue with the new custom-hotkey feature.

**The hotkey parser doesn't trim whitespace, so combos written with spaces silently do nothing.** `SendParsedHotkey` splits the string on `'+'` but never trims the resulting tokens, while the README's own **Format** line documents the syntax *with* spaces:

```
**Format:** `Modifier + Modifier + ... + Key`
```

A user who follows that line literally and types `Ctrl + V` produces the tokens `"Ctrl "` and `" V"`. `StrICmp("Ctrl ", L"Ctrl")` returns non-zero (the trailing space doesn't match), `" V"` isn't a recognized single char, so both tokens parse to `0`, `keys` ends up empty, and the function returns without sending anything — the hotkey just quietly fails with no feedback. The README contradicts itself too (the examples use `Ctrl+Shift+N` with no spaces), so users will hit this either way.

Fix: trim each token before parsing (more robust than relying on users matching an exact spacing):

```cpp
static void SendParsedHotkey(const std::wstring& combo) {
    auto trim = [](std::wstring s) {
        size_t a = s.find_first_not_of(L" \t");
        if (a == std::wstring::npos) return std::wstring();
        size_t b = s.find_last_not_of(L" \t");
        return s.substr(a, b - a + 1);
    };

    std::vector<WORD> keys;
    size_t start = 0;
    while (start < combo.length()) {
        size_t end = combo.find(L'+', start);
        std::wstring token = trim((end == std::wstring::npos)
            ? combo.substr(start) : combo.substr(start, end - start));
        WORD vk = ParseHotkeyToken(token.c_str());
        if (vk) keys.push_back(vk);
        if (end == std::wstring::npos) break;
        start = end + 1;
    }
    ...
}
```

<p>

- **Version mismatch.** The PR title and body say v2.3.0, but `@version` is set to `2.2.0` (bumped from `2.1.1` on main). Reconcile the two so the changelog/version is accurate.
- **Stale action count in the README.** The intro still says "Supports 12 different actions" (line 26), but this PR adds **Paste** and **Custom Hotkey**, bringing it to 14 options. Update the count.
- **`StrICmp` reinvents `_wcsicmp`.** The hand-rolled case-insensitive compare in `ParseHotkeyToken` can be replaced with the standard `_wcsicmp` (or `lstrcmpiW`), which is simpler and already available.
- **Silent 4-key cap.** `if (keys.empty() || keys.size() > 4) return;` drops any combo with 5+ keys (e.g. `Ctrl+Alt+Shift+Win+S`) with no indication. Unlikely to bite anyone, but worth either bumping the cap (the `inputs[8]` buffer would need enlarging too) or documenting the limit.
- **Custom `StringSetting` class** (pre-existing) duplicates `WindhawkUtils::StringSetting`. Since this PR adds three more instances of it, it might be worth switching to the built-in RAII helper. Not required.

</p>
</details>

<p>

- **Extended-key flag for navigation keys.** The README advertises arrows, `Home`/`End`/`PageUp`/`PageDown`/`Insert`/`Delete` as supported, but `SendParsedHotkey` sends them via `ki.wVk` only, without `KEYEVENTF_EXTENDEDKEY`. These are extended keys; most apps interpret the VK correctly anyway, but some (and certain numpad/NumLock configurations) can misinterpret them. If you see odd behavior with those keys, set `KEYEVENTF_EXTENDEDKEY` for the relevant VKs.
- **Custom hotkeys are injected via `SendInput`**, i.e. against whatever window currently has focus — consistent with the existing `SendKeyCombo`-based actions (New Tab, Paste, etc.). For the *double*-middle-click path, the combo fires ~500ms after the click (timer-based detection), so if focus moves elsewhere in that window the keystrokes land on the new foreground window. This is inherent to the synthetic-input approach and matches the pre-existing canned actions, so it's just an FYI.

</p>
</details>

PR #4635, opened 2026-07-01
Neko cat v1.4.0

ciizerr:
## 🚀 v1.4.0 Update

**feat:** Added Neko Dog, context menus, and multi‑monitor improvements.

### Added
- **Neko Dog:** New desktop companion, selectable in settings or mixed with existing litters.  
- **Right‑Click Menu:** Quickly change behaviors (Chase Mouse, Pace, Play With Window).  
- **Save Last Behavior:** Companions now remember their active behavior after restart.  

### Improved
- **Multi‑Monitor Bounds:** Behaviors now align correctly to the monitor edges.  
- **Run Around Physics:** Natural rest cycles (~45s play, ~15s nap) and realistic ball physics with angles/reflections.  
- **Leg Animations:** Walking/running animations scale to actual distance traveled (improved overall animation).
- **Drag & Drop:** Dropping a companion resumes their previous behavior (except Play With Window mode).  
- **Behavior Logic:** Smoother movement and state transitions.  
- **Assets:** Downloads now point to latest commit including Neko Dog assets.  

### Fixed
- **Multi‑Monitor Bugs:** Companions no longer go off‑screen or get stuck.  
- **Window Maximizing:** Double‑clicking companions no longer maximizes underlying windows.  
- **Theme Switcher Leak:** Cleaned up dynamic theme loading to prevent memory leaks.

### DIFF
--- a/mods/neko-cat.wh.cpp
+++ b/mods/neko-cat.wh.cpp
@@ -2,7 +2,7 @@
 // @id              neko-cat
 // @name            Desktop Companions
 // @description     Spawn multiple animated characters (neko cat, sakura, and tomoyo) on your screen that interact with your windows and follow your cursor.
-// @version         1.3.0
+// @version         1.4.0
 // @author          ciizerr
 // @github          https://github.com/ciizerr
 // @include         windhawk.exe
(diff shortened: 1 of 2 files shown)


m417z:
### Submission review

---

**1\. `neko-dog` has no audio files.** The new `neko-dog` theme ships only `spritesheet.png` --- the `sounds/` folder doesn't exist in the repo. On first init the mod issues 5 requests that return HTTP 404, and `EnsureThemeDownloaded` returns `false` for it. The dog will be permanently silent. (The 404 bodies are briefly written then deleted by the status check, and the spritesheet-existence early-return prevents re-requesting on later runs, so it's cosmetic --- but the author should either add the dog sounds or skip sound downloads for that theme.)

<p>

- **`behaviors.txt` persistence could use `Wh_SetIntValue`/`Wh_GetIntValue`.** `SaveBehavior`/`LoadBehavior` read/write a hand-rolled text file in the mod storage folder. This works and is consistent with how you already cache downloaded assets there, so it's not a correctness problem — but the idiomatic Windhawk way to persist per-mod state is `Wh_SetIntValue(L"behavior_0", mode)` / `Wh_GetIntValue(...)`, which avoids the file I/O and the re-read-whole-file-on-every-change pattern. Purely a preference.

- **Tighten the `LoadBehavior` range check.** `if (mode >= CHASE_MOUSE && mode < MAX_BEHAVIOR)` accepts `FORCED_SLEEP`/`EXHAUSTED_SLEEP` (6/7). In practice only user-chosen modes (0–5) are ever saved, so this can't happen today — but a corrupted/edited file could load a pet permanently asleep. Bounding it with `mode <= PLAY_WITH_WINDOW` makes it robust.

- **Include hygiene.** The new `behaviors.txt` code uses `FILE*`, `_wfopen_s`, `fscanf_s`, `fprintf`, `fclose` (`<cstdio>`), and `std::min`/`std::max`/`std::max({...})` are used throughout (`<algorithm>`), but neither header is `#include`d — they resolve transitively today. Adding `#include <cstdio>` and `#include <algorithm>` makes it toolchain-independent.

- **Context menu check style.** `AppendMenuW(hMenu, cond ? MF_CHECKED : MF_STRING, ...)` works because `MF_STRING == 0`, but it reads as "checked OR string." Passing `MF_STRING | (cond ? MF_CHECKED : 0)` is clearer, and since these are mutually-exclusive behaviors a radio mark (`SetMenuItemInfo` with `MFT_RADIOCHECK`) conveys "current selection" better than a checkmark. Cosmetic only.

</p>
</details>

<p>

- **`WM_COMMAND` (menu) doesn't reset `actionCount`, but `CycleBehavior` does.** `actionCount` is shared across `PlayWithWindow`, `RunRandomly`, and `RunAround`. When you pick a mode from the right-click menu, the stale counter carries over — e.g. switching to Run Around while a previous mode left `actionCount` high can land the pet mid-nap immediately. Resetting `actionCount = 0` in the menu handler (as `CycleBehavior` already does) would make the two entry points behave the same.

- **Run Around ball state isn't reset on (re)entry.** `ballX`/`ballY` are only re-seeded when both equal the `-9999` sentinel, so re-entering Run Around resumes the last ball position/velocity rather than picking a fresh one. Minor; just noting the intent may have been a fresh ball each time.

- **Dropping a pet that's already force-sleeping.** On drop, the `else` branch runs `SetState(AWAKE)` but leaves `behaviorMode` unchanged. If the pet was already in `FORCED_SLEEP`/`EXHAUSTED_SLEEP` (e.g. dropped once in Play With Window, then picked up and dropped again), it wakes for a frame and then falls back asleep before the auto-restore kicks in. It self-recovers to `prevBehaviorMode`, but the drop feels slightly inconsistent with dropping an active pet.

</p>
</details>

ciizerr:
Thanks for the detailed review!  
I’ve committed the missing sound resources for `neko-dog` and updated the download URL to point to the latest commit.  

I also addressed the other feedback items:
- Switched `behaviors.txt` persistence to use `Wh_SetIntValue` / `Wh_GetIntValue`.  
- Tightened the `LoadBehavior` range check.  
- Added missing `#include <cstdio>` and `#include <algorithm>` headers.  
- Updated context menu check style for clarity.  
- Reset `actionCount` in the menu handler for consistency.  
- Re‑seeded ball state on re‑entry to Run Around.  
- Fixed drop behavior when pets are force‑sleeping.  

Everything should now be in good shape — let me know if it looks ready to merge!

PR #4650, opened 2026-07-02
Improve icons and fix a bug when two apps are opened

Anixx:
Improve the icon appearance and fix a bug when the icons were confused when two UWP apps were opened at the same time.

## Changelog

* Changelog item 1...
* Changelog item 2...

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/native-titlebars-uwp-lite.wh.cpp
+++ b/mods/native-titlebars-uwp-lite.wh.cpp
@@ -2,7 +2,7 @@
 // @id              native-titlebars-uwp-lite
 // @name            Remove UWP titlebars Lite
 // @description     Enables native titlebars in UWP apps
-// @version         1.2.0
+// @version         1.2
 // @author          Anixx
 // @github          https://github.com/Anixx
 // @include         ApplicationFrameHost.exe
@@ -18,7 +18,7 @@ Replaces the UWP titlebars with native Win32 titlebars.
 
 This mod is focused on the Classic theme, so may produce sub-optimal results in other cases.
 
-![Screenshot](https://i.imgur.com/oeCYWJY.png)
+![Screenshot](https://i.imgur.com/gqw78kB.png)
 
 */
 // ==/WindhawkModReadme==
@@ -43,7 +43,7 @@ static const GUID FOLDERID_AppsFolder_local =
 static const GUID IID_IExtractIconW_local =
     {0x000214fa, 0x0000, 0x0000, {0xc0, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x46}};
 
-static const PROPERTYKEY PKEY_AppUserModel_ID_local = 
+static const PROPERTYKEY PKEY_AppUserModel_ID_local =
     {{0x9F4C2855, 0x9F79, 0x4B39, {0xA8, 0xD0, 0xE1, 0xD4, 0x2D, 0xE1, 0xD5, 0xF3}}, 5};
 
 typedef INT64 (*CTitleBar__CreateTitleBarWindow_t)(void *);
@@ -75,7 +75,7 @@ struct FindAppFrameData
 BOOL CALLBACK FindAppFrameProc(HWND hWnd, LPARAM lParam)
 {
     FindAppFrameData *pData = (FindAppFrameData *)lParam;
-    
+
     DWORD pid = 0;
     GetWindowThreadProcessId(hWnd, &pid);
     if (pid != pData->processId)
@@ -100,10 +100,6 @@ HWND FindApplicationFrameWindow()
     return data.hAppFrame;
 }
 
-// Tries to find the ApplicationFrameWindow that actually owns the given
-// titlebar window, by walking the owner/parent chain, instead of blindly
-// grabbing the first ApplicationFrameWindow found in the process (which
-// breaks when the host process hosts more than one app window).
 HWND FindOwningApplicationFrameWindow(HWND hWnd)
 {
     HWND hCandidate = hWnd;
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

A note regarding 1: in practice it might work, but it's documented as not being supported. Using `RemoveWindowSubclassFromAnyThread` is preferable.

---

The two-app-confusion fix (walking the owner chain) and the sharper icon selection are good improvements. The main problem is the new teardown code: it doesn't actually remove the subclasses, which turns "add missing uninit" into a crash on unload rather than a fix.

**1. `Wh_ModUninit` removes subclasses cross-thread with raw `RemoveWindowSubclass` — it silently fails, leaving dangling subclass procs, which crashes on unload.** (`Wh_ModUninit`, line 498)

`Wh_ModUninit` runs on an arbitrary Windhawk thread, but the subclassed titlebar windows live on their own UI thread(s) in `ApplicationFrameHost.exe`. `RemoveWindowSubclass` (like `SetWindowSubclass`) can only be called from the window's own thread — the wiki spells this out: *"can't be used to subclass a window across threads"* ([Development-tips](https://github.com/ramensoftware/windhawk/wiki/Development-tips)). So the cleanup loop no-ops on every foreign-thread window, the `SubclassProc` stays installed, and after the mod DLL is unmapped the next message to any of those windows dispatches into freed code, causing an access violation in the host process. This is exactly the crash the added `Wh_ModUninit` was meant to prevent.

The fix is to install and remove the subclass with the from-any-thread helpers (they marshal the call to the window's thread via `SendMessage`). Note both sides must change together: `RemoveWindowSubclassFromAnyThread` works by sending a registered message that only the wrapper installed by `SetWindowSubclassFromAnyThread` handles, so keeping the raw `SetWindowSubclass` at install would leave removal broken too. `WH_SUBCLASSPROC` also has no `uIdSubclass` parameter, and the wrapper removes the subclass for you on `WM_NCDESTROY`:

```cpp
// 5-param signature (no uIdSubclass); wrapper handles RemoveWindowSubclass on WM_NCDESTROY
LRESULT CALLBACK SubclassProc(HWND hWnd, UINT uMsg, WPARAM wParam,
                              LPARAM lParam, DWORD_PTR dwRefData) {
    if (uMsg == WM_NCDESTROY) {
        ForgetSubclassedWindow(hWnd);
        RemovePropW(hWnd, L"NativeTitlebarIconSet");
        // no manual RemoveWindowSubclass here
    }
    ...
}

// install (CreateWindowInBandEx_hook):
WindhawkUtils::SetWindowSubclassFromAnyThread(res, SubclassProc, 0);

// Wh_ModUninit:
for (HWND hWnd : windowsCopy)
    WindhawkUtils::RemoveWindowSubclassFromAnyThread(hWnd, SubclassProc);
```

The copy-then-release pattern you already use in `Wh_ModUninit` is correct — keep calling the removal outside the critical section (`RemoveWindowSubclassFromAnyThread` uses `SendMessage` and would deadlock if held under a lock the subclass proc also takes). See [window-opacity.wh.cpp#L424](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/window-opacity.wh.cpp#L424) and [#L534](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/window-opacity.wh.cpp#L534) for the pattern.

**2. The `HICON` set via `WM_SETICON` is never destroyed, so one icon handle leaks per subclassed window.** (`GetIconForWindow` / `CreateWindowInBandEx_hook` / `SubclassProc`)

`WM_SETICON` does not take ownership of the icon — the app owns it and must `DestroyIcon` it. The mod creates a fresh `HICON` (via `GetHICON`, `LoadImageW`, `ExtractIconExW`, or `SHGetFileInfoW`), sends it with `WM_SETICON`, and never frees it, so every UWP titlebar window leaks one GDI icon for the life of `ApplicationFrameHost.exe`. Store the handle instead of the `(HANDLE)1` sentinel and destroy it when the window goes away:

```cpp
// on set:
SendMessageW(hWnd, WM_SETICON, ICON_SMALL, (LPARAM)hIcon);
SetPropW(hWnd, L"NativeTitlebarIcon", (HANDLE)hIcon);

// WM_NCDESTROY (and the Wh_ModUninit loop):
if (HICON h = (HICON)GetPropW(hWnd, L"NativeTitlebarIcon")) DestroyIcon(h);
RemovePropW(hWnd, L"NativeTitlebarIcon");
```

(This is still an improvement over 1.1.0, which re-set — and leaked — an icon on *every* `WM_NCACTIVATE`; the new prop-guard already limits it to once per window, so finishing it with a `DestroyIcon` closes the leak entirely.)

<p>

- **`EnsureGdiplus()` isn't thread-safe.** UWP views can run on separate threads inside `ApplicationFrameHost.exe`, and the `if (g_gdiplusToken == 0)` check followed by `GdiplusStartup` isn't atomic, so two threads loading an icon at the same time can both call `GdiplusStartup` (one token leaks, and `GdiplusShutdown` then mismatches). Guard it with the critical section you already have, or a `std::once_flag`.
- **Dead null-checks after `new`.** `if (!bmp ...)` / `if (!scaled ...)` in `LoadPngAsIconSized` — the throwing `new Gdiplus::Bitmap(...)` never returns `nullptr`, so those `!bmp`/`!scaled` branches are unreachable. Harmless, just noise; the `GetLastStatus()` check is the one that matters.

</p>
</details>

<p>

- **Only `ICON_SMALL` is set now (1.1.0 set both `ICON_BIG` and `ICON_SMALL`).** The 16x16 icon looks crisper in the titlebar, but with no `ICON_BIG`, `WM_GETICON`/`ICON_BIG` consumers (Alt+Tab, the taskbar thumbnail, some window-list UIs) fall back to the class icon or an upscaled small icon, which can look worse there. If that's intentional, ignore; otherwise consider also rendering a larger variant (e.g. 32x32) and setting `ICON_BIG`.
- **`FindOwningApplicationFrameWindow` falls back to `FindApplicationFrameWindow()`** (first `ApplicationFrameWindow` in the process) when the owner/parent walk doesn't reach a frame. In the exact multi-app scenario this PR targets, that fallback can still attach the wrong app's icon. It's a reasonable best-effort default, but worth being aware the confusion isn't fully eliminated if the chain walk ever misses.

</p>
</details>

Anixx:
I do not understand, how it keeps beeing installed into other threads if installing cross-thread is impossible with the current approach? It installs and cLeanups only from the current thread! And I do not need to make it cross-thread, I do not need it installed in any thread in the process.

m417z:
`SetWindowSubclass` is called from `CreateWindowInBandEx_hook` from the window thread (that's fine).

`RemoveWindowSubclass` is called from `Wh_ModUninit` from a different thread (that's what the review flags).

Anixx:
Fixed the issues.

PR #4652, opened 2026-07-02
network-speed-tray-icon

kar2ner:
Simple mod that shows current download speed in system tray.
[network-speed-tray-icon.wh.cpp](https://github.com/kar2ner/windhawk-mods/blob/2c4c3fae07bf4ae4c9b7aa3b6cfd7849ce2e2aee/mods/network-speed-tray-icon.wh.cpp)

## Mod authorship

This mod was created by:

- - [ ] Claude Code

### DIFF
--- a/mods/network-speed-tray-icon.wh.cpp
+++ b/mods/network-speed-tray-icon.wh.cpp
@@ -2,7 +2,7 @@
 // @id              network-speed-tray-icon
 // @name            Network Speed Tray Icon
 // @description     Shows current download speed icon in system tray
-// @version         1.0
+// @version         1.1
 // @author          kar2ner
 // @github          https://github.com/kar2ner
 // @include         explorer.exe
@@ -11,32 +11,24 @@
 // ==/WindhawkMod==
 
 // ==WindhawkModReadme==
-/*
-# Network Speed Tray Icon
-
-Показывает динамическую иконку в системном трее с текущей скоростью
-**входящего** трафика (суммарно по всем активным сетевым интерфейсам,
-кроме loopback). Иконка и всплывающая подсказка обновляются раз в секунду.
-
-Скорость считается через IP Helper API (GetIfTable2), поэтому мод не
-зависит от диспетчера задач и работает даже если тот закрыт.
-*/
+/*...*/
 // ==/WindhawkModReadme==
 
 // ==WindhawkModSettings==
-/*
-- updateIntervalMs: 1000
-  $name: Интервал обновления (мс)
-  $description: Как часто пересчитывать и перерисовывать иконку
-*/
+/*...*/
 // ==/WindhawkModSettings==
 
-// MIB_IF_TABLE2 / GetIfTable2 доступны начиная с Windows Vista.
+// RU: MIB_IF_TABLE2 / GetIfTable2 доступны начиная с Windows Vista.
 // Без явного указания версии некоторые тулчейны берут более старый
 // _WIN32_WINNT по умолчанию, и тогда эти типы/функции "не находятся" —
 // это самая частая причина сразу пачки ошибок компиляции в таком коде.
+//
+// EN: MIB_IF_TABLE2 / GetIfTable2 are available starting with Windows
+// Vista. Without an explicit version, some toolchains default to an
+// older _WIN32_WINNT, so these types/functions become "not found" —
+// this is the most common cause of a whole batch of compile errors here.
 #ifndef _WIN32_WINNT
-#define _WIN32_WINNT 0x0602  // Windows 8, с запасом
+#define _WIN32_WINNT 0x0602  // Windows 8, с запасом / Windows 8, with headroom
 #endif
 #define WIN32_LEAN_AND_MEAN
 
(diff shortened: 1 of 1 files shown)


Anixx:
Показывает 0.00. И не видно цифр на светлом фоне.

kar2ner:
> Показывает 0.00. И не видно цифр на светлом фоне.

Попробую чуть позже посмотреть. Обновление показателей сделал с шагом в 0.5 МБ/сек, чтобы текст умещался (более-менее только 2 цифры влезет). Касаемо фона - не учел, у меня таскбар темно-серый. Постараюсь прикрутить выбор цвета.

Anixx:
Если 0.5 вместится, то и 0.1 вместится.

Anixx:
И что показывается-то? Ширина канала или текущая закачка?

m417z:
### Submission review

---

Thanks for the submission! The core idea is fine, but there are a few things to address before this can be merged.

**1. This should be a "tool mod", not injected into `explorer.exe`.** The mod installs **no function hooks** — it only uses APIs that work from any process (a message-only window, a `SetTimer` loop, `Shell_NotifyIcon`, and the IP Helper API). That's the textbook signal for the [Mods as tools](https://github.com/ramensoftware/windhawk/wiki/Mods-as-tools:-Running-mods-in-a-dedicated-process) pattern (the wiki even lists a "network indicator" as an example). Running it inside `explorer.exe` has two real downsides for users:
   - A bug/crash in the icon-rendering or IP-Helper code destabilizes the whole shell.
   - `explorer.exe` is not a single process — with "launch folder windows in a separate process" enabled (or other extra `explorer.exe` instances) the mod is injected into each one and you get **duplicate tray icons**, since there's no single-instance dedup.

   Converting it to a tool mod fixes both and is straightforward: set `@include windhawk.exe`, rename `Wh_ModInit`/`Wh_ModUninit` → `WhTool_ModInit`/`WhTool_ModUninit` (add `WhTool_ModSettingsChanged` if you want live setting updates), and paste the launcher boilerplate verbatim from the wiki. The dedicated process gives you single-instance behavior for free, so you can drop the home-grown threading if you like. A near-identical tray-icon tool mod to copy from is [theme-toggler-tray.wh.cpp](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/theme-toggler-tray.wh.cpp) (hidden window + `Shell_NotifyIcon` + `WhTool_*` + launcher), and [internet-status-indicator.wh.cpp](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/internet-status-indicator.wh.cpp) is another network-tray tool mod. The launcher snippet lives at the bottom of [explorer-folder-hover-menu.wh.cpp](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/explorer-folder-hover-menu.wh.cpp#L3778).

**2. Handle `TaskbarCreated` to survive Explorer restarts.** Once the mod lives in a dedicated process (item 1), it no longer dies and re-injects when `explorer.exe` restarts — so its tray icon disappears permanently after an Explorer crash/restart and never comes back. Register the standard message and re-add the icon when it arrives:
   ```cpp
   g_uMsgTaskbarCreated = RegisterWindowMessageW(L"TaskbarCreated");
   // in WndProc:
   if (msg == g_uMsgTaskbarCreated) { Shell_NotifyIconW(NIM_ADD, &g_nid); return 0; }
   ```
   See [theme-toggler-tray.wh.cpp#L81](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/theme-toggler-tray.wh.cpp#L81) and [theme-toggler-tray.wh.cpp#L102](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/theme-toggler-tray.wh.cpp#L102).

**3. User-facing strings must default to English.** The README, the setting's `$name`/`$description`, and the tooltip text (`L"Входящая скорость: %.1f МБ/с"`, `L"Скорость сети: --"`) are all Russian. Per convention, the primary language must be English, with other languages provided via the localization syntax. For the metadata/README/settings use the `$name:ru-RU` / `$description:ru-RU` form for the Russian translation; for the runtime tooltip strings you'll need to key them off `Wh_GetIntSetting`-driven logic or just ship them in English. (The Russian code comments are fine to keep — that's not user-facing — see the optional section.)

<p>

- **`g_running` is dead code.** It's declared `volatile bool`, set to `true` in `Wh_ModInit` and `false` in `Wh_ModUninit`, but never read anywhere (the worker thread exits via `GetMessageW` returning 0, not by checking this flag). Remove it.
- **Uninitialized mask bitmap.** `CreateBitmap(size, size, 1, 1, nullptr)` leaves the monochrome mask bits undefined. Since `hbmColor` is a 32-bpp DIB with an alpha channel this usually works out, but it's safer to allocate a zeroed mask buffer (or reuse a known-good icon-creation helper) to avoid depending on undefined contents.
- **`RegisterClassW` return value is unchecked** — minor, since `CreateWindowExW` will fail loudly if the class isn't registered, but worth checking for clarity.
- Code comments being in Russian is fine (not user-facing), just calling it out so it's clearly distinct from item 3 above.

</p>
</details>

<p>

- **README/comments claim `GetIfTable2`, but the code uses `GetIfEntry`.** Beyond the doc mismatch, `MIB_IFROW::dwInOctets` is a **32-bit** counter that wraps at ~4 GB (on a gigabit link that's every ~34 s). Your delta code clamps a negative delta to 0, so a wrap just drops one sample rather than crashing — but you'll get an occasional bogus 0 reading. Prefer `GetIfEntry2` + `MIB_IF_ROW2::InOctets` (a `ULONG64`), which matches what the README already advertises.
- **README says "sum over all interfaces", code measures only the best-route interface.** The comment block explains *why* you switched to `GetBestInterface` (avoiding double-counting from virtual adapters), which is reasonable — but the README text still describes the old summing behavior. Update the README to match.
- **Coarse display granularity.** Rounding to the nearest 0.5 MB/s means anything under ~0.25 MB/s shows as `0.0` and typical browsing hovers around `0.0`/`0.5`. Consider showing KB/s for sub-MB speeds (or a finer step), which would make the icon far more informative for most users. Also note it's download-only by design — worth stating clearly in the README.
- **Text rendering via black-color-key transparency.** `CLEARTYPE_QUALITY` produces sub-pixel colored fringes, and the "make black pixels transparent" step gives hard edges everywhere else — so anti-aliased text can look rough or fringed. `ANTIALIASED_QUALITY` (or drawing white text and using the luminance as alpha) generally looks cleaner for a tiny tray glyph.
- **Fixed 32×32 icon regardless of DPI.** On high-DPI displays the tray expects a larger icon; consider sizing from `GetSystemMetrics(SM_CXSMICON)`/`SM_CXICON` (and re-rendering on DPI change) so the glyph stays crisp.

</p>
</details>

PR #4636, opened 2026-07-01
Add Monitor Rounded Edges

crazyboyybs:
Applies smooth, anti-aliased rounded corners to all connected monitors using Direct2D.

## Features

- Two corner styles -- Circle (classic arc) and Squircle (smooth superellipse).
- Inner accent border -- optional outline with configurable thickness and alpha color (#AARRGGBB).
- Fullscreen awareness -- uses SHQueryUserNotificationState to reliably detect fullscreen games.
- Minimal CPU overhead -- a continuous polling timer (750 ms) is only active when "Hide in
  Fullscreen" is enabled. "Dynamic Inner Border" uses a brief transition timer (5 checks over
  1.5 s) after each foreground change to catch delayed fullscreen detection, then stops
  automatically. With neither feature active, only a lightweight WinEvent hook runs.
- Multi-monitor support -- each monitor gets its own independent set of corner overlays.
- Single tool process -- the rendering process is only created once; reopening Windhawk from the
  system tray does not spawn additional instances.

## Mod authorship

If this pull request introduces a new mod, please complete the section below.

This mod was created by:

- - [ ] The submitter, without AI assistance
- - [X] The submitter, with AI assistance
- - [X] Claude
- - [ ] ChatGPT
- - [ ] Gemini
- - [ ] Another AI (please specify): 
- - [ ] Other (please specify): 

Please select the options that best apply. Your selection does not affect the acceptance criteria, but it helps reviewers understand the context of the code and provide relevant feedback.

PR #4653, opened 2026-07-02
True Rounded Corners

mursman:
Rounded Corners Overlay (Screen-anchored)
Simpler alternative to per-window tracking: since a maximized window is always flush with its monitor's work area, this pins four small click-through, topmost, layered "corner mask" windows to the work-area corners of every monitor -- once, at startup, plus whenever display configuration changes (resolution, monitor added/removed, taskbar moved/resized).

Whatever window happens to be maximized underneath a corner automatically appears to have that corner clipped round. Works instantly for any app, any process, with no window enumeration loop at all.

Trade-off vs. per-window tracking: this only rounds corners that touch a monitor's work-area edge. A maximized window on a single monitor gets all four corners rounded (since it fills the whole work area). If you ever use a window that's large but not technically maximized, or an edge-snapped half-screen window, only the corners that land on the monitor's own corners will be rounded -- snapped windows won't get their inner corners touched by this mod, which is usually what you want anyway (only the screen-facing outer corners are visually squared off by DWM to begin with).

Same limitation as before on fill color: the corner cutout is a solid color, not a live capture of what's underneath. Tune fill_color to match your wallpaper/taskbar area.

## Changelog

* Changelog item 1...
* Changelog item 2...

## Mod authorship

This mod was created by: mursman

- - [ yes ] The submitter, with AI assistance
- - [ yes ] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/round-screen-corners.wh.cpp
+++ b/mods/round-screen-corners.wh.cpp
@@ -1,0 +1,380 @@
+// ==WindhawkMod==
+// @id              round-screen-corners
+// @name            True Rounded Corners
+// @description     Pins rounded-corner masks to each monitor's work-area corners, so any maximized window underneath appears rounded -- no per-window tracking needed
+// @version         1.0
+// @author          mursman
+// @github          https://github.com/mursman
+// @twitter         https://twitter.com/murs_man
+// @homepage        https://behance.net/mursman
+// @include         explorer.exe
+// @compilerOptions -lgdi32 -luser32 -lkernel32
+// @license         MIT
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Rounded Corners Overlay (Screen-anchored)
+
+Simpler alternative to per-window tracking: since a maximized window is
+always flush with its monitor's work area, this pins four small
+click-through, topmost, layered "corner mask" windows to the work-area
+corners of every monitor -- once, at startup, plus whenever display
+configuration changes (resolution, monitor added/removed, taskbar
+moved/resized).
+
+Whatever window happens to be maximized underneath a corner automatically
+appears to have that corner clipped round. Works instantly for any app,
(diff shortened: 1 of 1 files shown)


m417z:
How does it compare to this mod?
https://github.com/ramensoftware/windhawk-mods/pull/4636

mursman:
> How does it compare to this mod? #4636

Is there anything wrong with the procedure? The mod is working on my machine. Please tell me a bit in details, this my first time

m417z:
The linked mod seems to be similar. My question is whether they're truly the same, or if your mod is different. There's no reason to have two mods doing the exact same thing.

mursman:
After looking at its architecture and feature set, I'd say the #4636  implementation isn't primarily about rounded corners. Rounded corners are just one feature built on top of a much more general rendering framework.

Its real purpose is to provide a high-quality desktop overlay and composition engine.

m417z:
Can you explain what your mod provides to users that #4636 doesn't? I didn't understand that from your messages, and you didn't post a screenshot.

mursman:
In One Line
My Mod is the more efficient and simpler design built for only rounded monitor corners.
✅ Lightweight
✅ Minimal resources
✅ Simple
✅ Does one job very well

mursman:
image  rounds 4 corners of the monitor edges.

mursman:
image
full shot

m417z:
So it seems that my understanding is correct and your mod doesn't add additional functionality compared to the other mod. I prefer not having multiple mods which achieve the same thing without a good reason.

mursman:
> So it seems that my understanding is correct and your mod doesn't add additional functionality compared to the other mod. I prefer not having multiple mods which achieve the same thing without a good reason.

okay

PR #4654, opened 2026-07-02
Files 2 Folders v2.5

triatomic:
## Summary
- Adds an **Operation** segmented control (Move - fast / Move - safe / Copy) to the dialog, above the mode list. Picking one changes the operation for just that run without touching the **Operation** setting.
- Adds two optional aids for building nested subfolder paths in **Fixed name** mode (each with its own settings toggle, on by default):
  - A live breadcrumb preview line under the box (e.g. typing `archive/incoming` shows `archive › incoming`).
  - An **Add level** (`+`) button that appends a `\` separator and refocuses the box.
- Two new settings: `showNestedPreview`, `showAddLevelButton`.

## Mod authorship
- [x] Claude

## Test plan
- [x] Verified in Windhawk: dialog compiles and opens with the new Operation control and both nested-path aids.
- [x] Verified switching the Operation control relabels Move/Copy text live and the chosen operation is used on OK.
- [x] Verified the nested-folder preview updates live while typing, including multi-segment and sanitized names.
- [x] Verified the Add level button appends `\` and keeps focus/caret in the edit box.
- [x] Verified both new settings toggles hide/show their respective controls.

### DIFF
--- a/mods/files-2-folders.wh.cpp
+++ b/mods/files-2-folders.wh.cpp
@@ -2,7 +2,7 @@
 // @id              files-2-folders
 // @name            Files 2 Folders
 // @description     Move or copy one or more selected files in Explorer into a subfolder (named — nested paths with "/" supported, by extension, by name, or by date), with a workaround hotkey for other file managers
-// @version         2.2
+// @version         2.5
 // @author          tria
 // @github          https://github.com/triatomic
 // @include         explorer.exe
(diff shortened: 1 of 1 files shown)


mods/files-2-folders.wh.cpp
```diff
@@ -244,6 +272,13 @@ Forbidden characters in folder names (`* : ? " < > | / \`) are replaced with
 #include <vector>
 #include <algorithm>
 
+// Not pulled in by this toolchain's <windows.h>/<winuser.h> headers even
+// though it's a long-standing documented static-control style; define it
+// directly rather than bumping WINVER (which the PR validator disallows).
+#ifndef SS_END_ELLIPSIS
+#define SS_END_ELLIPSIS 0x00004000L
```
m417z, line 279: The constant name is `SS_ENDELLIPSIS`.

PR #4656, opened 2026-07-02
Add Island Media Controls mod

usho-lear:
## Changelog

Not applicable; this pull request introduces a new mod.

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [x] ChatGPT
- - [ ] Gemini
- - [ ] Another AI (please specify): Primarily AI-generated; exact model not specified

### DIFF
--- a/mods/island-media-controls.wh.cpp
+++ b/mods/island-media-controls.wh.cpp
@@ -1,0 +1,6624 @@
+// ==WindhawkMod==
+// @id              island-media-controls
+// @name            Island Media Controls
+// @description     Dynamic island-like media controls for the Windows 11 taskbar.
+// @version         0.9.27
+// @author          usho
+// @github          https://github.com/usho-lear
+// @license         MIT
+// @include         explorer.exe
+// @architecture    x86-64
+// @compilerOptions -lole32 -loleaut32 -lruntimeobject -lwindowsapp -lversion -luuid -luser32 -lshell32 -lgdi32 -lmsimg32 -lshlwapi -lwindowscodecs -ldwmapi -lshcore
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Island Media Controls
+
+This version is rebuilt around the Windows 11 taskbar XAML tree. The compact
+media island is inserted as a XAML Grid into the taskbar layout, so it occupies
+real taskbar space instead of being a Win32 window layered over the taskbar.
+
+The island follows the Windows app theme and includes album art and smooth
+hover feedback. Clicking it replaces the compact control with a desktop XAML
+surface that uses native Acrylic, rounded XAML geometry, and display-synced
+CompositionTarget animation above the taskbar.
+*/
+// ==/WindhawkModReadme==
+
+// ==WindhawkModSettings==
+/*
+- Main:
+  - Position: "tray_left"
+    $name: Taskbar position
+    $options:
+    - "tray_left": "Tray - far left"
+    - "tray_right": "Tray - far right"
+    - "tray_before_clock": "Tray - left of clock"
+    - "tray_after_clock": "Tray - right of clock"
+    - "taskbar_left_edge": "Taskbar - left overlay"
+    - "taskbar_center_edge": "Taskbar - center overlay"
+    - "taskbar_right_edge": "Taskbar - right overlay"
+  - CompactWidth: 168
+    $name: Compact width
+  - ExpandedWidth: 360
+    $name: Expanded overlay width
+  - ExpandedHeight: 500
+    $name: Expanded overlay height
+  - PopupSpacing: 8
+    $name: Expanded outer spacing
+  - PopupCardGap: 8
+    $name: Expanded cover-to-controls gap
+  - PopupShadowDepth: 48
+    $name: Expanded shadow depth
+  - PopupShadowOpacity: 70
+    $name: Expanded shadow opacity (%)
+  - PopupButtonStyle: "minimal_transport"
+    $name: Expanded button style
+    $options:
+    - "minimal_transport": "Minimal transport"
+    - "fluent_bold": "Fluent bold"
+  - PopupBackdropCoverEffect: "dark_only"
+    $name: Expanded background album wash
+    $options:
+    - "off": "Off"
+    - "dark_only": "Dark mode only"
+    - "on": "Always on"
+  - ArtworkAbstractMode: "mesh_gradient"
+    $name: Low-res video artwork mode
+    $options:
+    - "off": "Off"
+    - "mesh_gradient": "Mesh gradient"
+    - "energy_flame": "Flame"
+  - Height: 40
+    $name: Component height
+  - MarginLeft: 4
+    $name: Left margin
+  - MarginRight: 4
+    $name: Right margin
+  - HideWhenNoMedia: false
+    $name: Hide when no media
+  - PollIntervalMs: 1000
+    $name: Media refresh interval
+  - HoverScale: 106
+    $name: Hover scale (%)
+  - HoverLerpSpeed: 28
+    $name: Hover smoothing
+  - Material: "mica_like"
+    $name: Background material
+    $options:
+    - "mica_like": "Mica-like content layer"
+    - "solid": "Solid"
+    - "acrylic": "Acrylic / glass"
+*/
+// ==/WindhawkModSettings==
+
+#include <windows.h>
+#include <windowsx.h>
+#include <dwmapi.h>
+#include <shlwapi.h>
+#include <wincodec.h>
+#include <windows.ui.xaml.hosting.desktopwindowxamlsource.h>
+#include <windhawk_utils.h>
+
+#ifdef GetCurrentTime
+#undef GetCurrentTime
+#endif
+
+#include <winrt/base.h>
+#include <winrt/Windows.Foundation.h>
+#include <winrt/Windows.Foundation.Collections.h>
+#include <winrt/Windows.Foundation.Numerics.h>
+#include <winrt/Windows.Media.Control.h>
+#include <winrt/Windows.Storage.Streams.h>
+#include <winrt/Windows.UI.h>
+#include <winrt/Windows.UI.Composition.h>
+#include <winrt/Windows.UI.Input.h>
+#include <winrt/Windows.UI.Text.h>
+#include <winrt/Windows.UI.Xaml.h>
+#include <winrt/Windows.UI.Xaml.Hosting.h>
+#include <winrt/Windows.UI.Xaml.Controls.Primitives.h>
+#include <winrt/Windows.UI.Xaml.Controls.h>
+#include <winrt/Windows.UI.Xaml.Input.h>
+#include <winrt/Windows.UI.Xaml.Media.h>
+#include <winrt/Windows.UI.Xaml.Media.Imaging.h>
+
+#include <algorithm>
+#include <array>
+#include <functional>
+#include <atomic>
+#include <chrono>
+#include <cstdint>
+#include <cmath>
+#include <climits>
+#include <mutex>
+#include <string>
+#include <thread>
+#include <vector>
+
+#ifndef WS_EX_NOREDIRECTIONBITMAP
+#define WS_EX_NOREDIRECTIONBITMAP 0x00200000L
+#endif
+
+namespace xaml = winrt::Windows::UI::Xaml;
+namespace composition = winrt::Windows::UI::Composition;
+namespace controls = winrt::Windows::UI::Xaml::Controls;
+namespace input = winrt::Windows::UI::Xaml::Input;
+namespace mediax = winrt::Windows::UI::Xaml::Media;
+namespace imaging = winrt::Windows::UI::Xaml::Media::Imaging;
+namespace hosting = winrt::Windows::UI::Xaml::Hosting;
+namespace streams = winrt::Windows::Storage::Streams;
+namespace gsm = winrt::Windows::Media::Control;
+
+using xaml::FrameworkElement;
+using xaml::GridLength;
+using xaml::GridUnitType;
+using xaml::HorizontalAlignment;
+using xaml::Thickness;
+using xaml::UIElement;
+using xaml::VerticalAlignment;
+using xaml::Visibility;
+using controls::Border;
+using controls::Button;
+using controls::ColumnDefinition;
+using controls::Grid;
+using controls::Image;
+using controls::ProgressBar;
+using controls::StackPanel;
+using controls::TextBlock;
+using mediax::SolidColorBrush;
+using mediax::ScaleTransform;
+using mediax::TransformGroup;
+using mediax::TranslateTransform;
+using mediax::VisualTreeHelper;
+
+namespace {
+
+struct Settings {
+    std::wstring position = L"tray_left";
+    int compactWidth = 168;
+    int expandedWidth = 360;
+    int expandedHeight = 430;
+    int popupSpacing = 8;
+    int popupCardGap = 8;
+    int popupShadowDepth = 48;
+    int popupShadowOpacity = 70;
+    std::wstring popupButtonStyle = L"minimal_transport";
+    std::wstring popupBackdropCoverEffect = L"dark_only";
+    std::wstring artworkAbstractMode = L"mesh_gradient";
+    int height = 40;
+    int marginLeft = 4;
+    int marginRight = 4;
+    bool hideWhenNoMedia = false;
+    int pollIntervalMs = 1000;
+    double hoverScale = 1.06;
+    double hoverLerpSpeed = 28.0;
+    std::wstring material = L"mica_like";
+};
+
+struct MediaState {
+    std::wstring title = L"Not Playing";
+    std::wstring artist;
+    bool hasSession = false;
+    bool isPlaying = false;
+    int64_t positionTicks = 0;
+    int64_t durationTicks = 0;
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

Some issues I noticed in a quick manual test:
* The widget doesn't show up after restarting explorer. Most likely, that means that the mod won't work after a reboot until it's disabled and re-enabled manually.
* The thumbnail doesn't show up for me. I tested with Edge and YouTube.
* Seeking doesn't work. I don't know if it's supposed to work, but it'd be nice to have it working.
* It's often stuck in expanded mode.

---

**1. Remove the unused code.** `CreateAbstractAlbumCoverBytes`, `CreateRibbonWaveAlbumCoverBytes`, `CreateCalmMeshAlbumCoverBytes`, `coolB`, `CreatePinkRibbonAlbumCoverBytes`, `CreateOrangeVolumeAlbumCoverBytes`, `CreatePurpleBeamAlbumCoverBytes`, `ContrastTextColorFor`, `PopupHorizontalProgress`, `PopupControlStrokeBrush`, `finalHeight`, `kDwmCornerRound`, `kDwmBackdropTransient`, `showProgressGlow`, `glowCenterX`, `glowCenterY`, `OnExpandRendering`.

**2. Data race on the pending-media-switch globals → heap corruption in explorer.exe.** `SetMedia` (media worker thread) mutates `g_pendingMediaSwitchKey` / `g_pendingMediaSwitchCount` / `g_pendingMediaSwitchSince` **while holding `g_mediaMutex`** (lines ~890–946), but `FinishCloseExpandedPopup` runs on the taskbar UI thread and clears the same globals **without the lock**:

```cpp
// FinishCloseExpandedPopup, lines 5100-5102 — no g_mediaMutex held
g_pendingMediaSwitchKey.clear();
g_pendingMediaSwitchCount = 0;
g_pendingMediaSwitchSince = {};
```

If the user closes the popup at the same instant a poll fires (polls run every ~1 s by default, so this is a normal-use collision), one thread runs `std::wstring::clear()` (freeing the buffer) while the other assigns/compares that same `std::wstring` → data race → use-after-free / heap corruption / crash. Take `g_mediaMutex` around that reset (or move it into a mutex-guarded helper).

**3. WinRT/XAML exceptions can escape Win32 callbacks → `std::terminate` crashes Explorer.** You correctly wrap many XAML calls in `try/catch` (e.g. `GetElementScreenRect`, the bitmap sections, `OnHoverRendering`, `OnCompactTextRendering`), but three foreign-callback entry points are unguarded, and any `winrt::hresult_error` thrown there unwinds through user32/hook C frames (undefined behavior → typically terminates explorer.exe):
- `ExpandedPopupWndProc` — `WM_TIMER` (calls `RenderExpandedPopupLayer` / `UpdatePopupLiveProgressFromSnapshot` / `BeginCloseExpandedPopup`) and `WM_SETTINGCHANGE`/`WM_THEMECHANGED` (calls `ApplyPopupXamlTheme(true)`), lines ~5202–5252. A raw WndProc is called directly by `DispatchMessageW`, so unlike a C++/WinRT delegate it has no ABI thunk to convert the exception.
- `UpdatePlayerContents` (line 6011) — invoked via the `WH_CALLWNDPROC` hook in `RunFromWindowThread` (the `payload->proc(...)` call at line 664 has no guard). Only the two bitmap sections are wrapped; the theme/text/accent setters (6182–6234) are not.
- `OnPopupXamlRendering` (line 5107) — this one *is* a WinRT delegate so the exception is caught at the thunk rather than crashing, but it never clears `g_popupRenderingHooked`/the token or stops the loop, so the same throwing call fires every frame and the popup can get stuck (contrast `OnCompactTextRendering`, which stops + resets on failure). Add the same self-healing `catch`.

Wrap each of these in a top-level `try { … } catch (...) {}`.

**4. Worker-thread lifetime — `std::terminate` on logoff, and detached threads outliving unload.**
- `g_mediaThread` is a global `std::thread` (line 250) joined only in `Wh_ModUninit` via `StopMediaThread`. But `Wh_ModUninit` is **not** called when the host process exits (Explorer restart / sign-out / shutdown) — the CRT instead runs the global's destructor during `DLL_PROCESS_DETACH`, and `~thread()` on a still-`joinable()` thread calls `std::terminate()`, aborting the host. Use a heap pointer with a trivial destructor (`std::thread* g_mediaThread;`) so nothing runs at shutdown, and keep the real join in `Wh_ModUninit`.
- `RunMediaCommand` (line 991) spawns a **detached** thread per transport command that runs mod code (`CurrentSession`, `RefreshMediaState` → `SetMedia`, touching `g_media`/`g_mediaMutex`). If the user hits play/next and disables the mod immediately after, that thread can still be executing inside the DLL after it's unloaded → use-after-unload crash. Consider routing commands through the existing media thread, or otherwise tracking/guarding these so none can run past unload.

**5. `InitializePopupXamlHost` failure leaves inconsistent state → broken popup.** On the `catch` (lines 4134–4138) only `g_popupXamlChild` and `g_popupXamlSource` are reset, but `g_popupXamlRoot` and all the element globals were already assigned at 4094–4128, before `InitializePopupCompositionShadow()` / `ApplyPopupXamlTheme(true)` (which can throw). If init fails there, `EnsureExpandedPopup` falls back to the GDI/layered path, yet `g_popupXamlRoot` is still non-null so `RenderExpandedPopupLayer` / `PositionExpandedPopup` keep taking the XAML branch against an un-hosted tree → the popup renders blank / never shows. Reset the full element-global set on the failure path (mirroring `DestroyExpandedPopup`).

<p>

- **`g_settings` is reassigned on the settings thread while other threads read it.** `LoadSettings` (line 576) runs on the `Wh_ModSettingsChanged` thread and overwrites `g_settings`' `std::wstring` members while the media thread (`pollIntervalMs`, line 5638) and UI thread read the same struct. This only races on a settings change, so it's a low-priority robustness item — but a torn read of a non-SSO `std::wstring` there could crash. A lock or a `std::atomic` snapshot/swap would close it.
- **Use `WindhawkUtils::StringSetting` (RAII)** instead of the hand-rolled `GetStringSetting` wrapper (line 569) that does a raw `Wh_GetStringSetting` + `Wh_FreeStringSetting`. Functionally equivalent, just less to maintain.
- **Cache the dark-mode value.** `IsDarkModeApprox()` (line 493) does a full `RegOpenKeyEx`/`RegQueryValueEx`/`RegCloseKey` on every call, and it's called from the GDI render path. Cache it and invalidate on `WM_SETTINGCHANGE`/`WM_THEMECHANGED` (which you already handle).
- **Possibly-unused `@compilerOptions` libraries** (line 11): I couldn't find any Version-API (`-lversion`) or shcore-specific usage — `GetDpiForWindow` lives in `user32`, not shcore (`-lshcore`). Drop them if they're not needed (CI doesn't flag unused `-l` libs).
- **Popup window class registration** (lines 5272–5275) uses `GetModuleHandleW(nullptr)` (the host exe) as `hInstance` and continues on `ERROR_CLASS_ALREADY_EXISTS`. Since you `UnregisterClassW` in `DestroyExpandedPopup` this is mostly fine, but the continue-on-exists path would reuse a stale class (with a `WndProc` pointing into a previous, possibly-unmapped image) if a prior instance ever failed to unregister. Registering/unregistering under the mod's own module handle is a touch safer.

</p>
</details>

<p>

- **The compact-island expand animation is dead code.** `OnExpandRendering` (line 5567), the `g_expandScale`/`g_expandTranslate` transforms, and the `g_currentExpandProgress`/`g_targetExpandProgress` machinery are never driven — there is no `CompositionTarget::Rendering(OnExpandRendering)` registration anywhere (line 5560 is the *revoke* inside `StopExpandRenderLoop`), and `g_expandRenderingHooked` is never set. `StartExpandRenderLoop` just calls `ApplyExpandedState()` once and shows/closes the popup, so the island's own expand transition is inert (the popup still animates via its own render loop). Either wire the loop up or delete the unused path.
- **Taskbar-host element offset is derived by reading raw function bytes** of `TaskbarHost::FrameHeight` (lines 707–720). It's inherently fragile against future Windows builds, but there's a sane `0x10` fallback and it's the established taskbar-modding technique, so this is just an FYI.
- **Outside-click dismissal polls** via `WM_TIMER` (15 ms non-XAML / 80 ms XAML) with `GetAsyncKeyState` + `GetCursorPos` (lines 5202–5219). There isn't a clean event-driven alternative for a `WS_EX_NOACTIVATE` tool window that never takes focus, so this is acceptable — just noting the 15 ms rate is quite tight.
- **`UpdatePlayerContents` blocks the taskbar UI thread on `.get()`** (e.g. `writer.StoreAsync().get()`, line 6047) every refresh. It's fast for in-memory streams, but it's a synchronous wait on the UI thread — worth keeping an eye on if it ever grows.

</p>
</details>

usho-lear:
Thanks for the detailed review. I addressed the required items and the manual-test issues in commit `9a681a8`:

- Added a retry path for taskbar injection when Explorer/XAML is not ready yet, so the widget can recover after an Explorer restart or reboot.
- Fixed thumbnail loading by consuming the `IBuffer` returned by `ReadAsync` instead of the original zero-length buffer.
- Added seeking from the expanded progress bar in both the XAML and fallback popup paths.
- Added self-healing popup cleanup when a render/update callback throws, preventing the expanded state from getting stuck.
- Removed all of the listed unused functions, variables, and the dead compact expand-animation machinery.
- Protected the pending-media-switch reset with `g_mediaMutex`.
- Added top-level exception guards for the taskbar-thread callback, `ExpandedPopupWndProc`, `UpdatePlayerContents`, and `OnPopupXamlRendering`; popup render failures now stop the render loop and restore the compact island.
- Replaced the global `std::thread` object with a heap-owned thread and routed transport/seek commands through that tracked worker. Shutdown now wakes, joins, and deletes the worker, with no detached mod-code threads.
- Added `ResetPopupXamlElementState` and use it on XAML host initialization failure after closing the partial host.
- Also removed the unused `-lversion` and `-lshcore` options.

Validation completed successfully with the repository PR validator (including symbol extraction), and the mod compiles with zero warnings for both x64 and ARM64.

m417z:
### Submission review

---

**1. `RegisterClassExW` reuses a possibly-stale class with a dangling window proc.** In `EnsureExpandedPopup` (`mods/island-media-controls.wh.cpp:4863`):

```cpp
windowClass.hInstance = GetModuleHandleW(nullptr);           // explorer's base, not the mod
windowClass.lpfnWndProc = ExpandedPopupWndProc;              // points into the mod DLL
if (!RegisterClassExW(&windowClass) && GetLastError() != ERROR_CLASS_ALREADY_EXISTS) {
    return false;
}
```

Continuing on `ERROR_CLASS_ALREADY_EXISTS` can reuse a class that a *previous* mod instance registered, whose `lpfnWndProc` now points into an unmapped/relocated copy of the mod → crash. You do unregister in `DestroyExpandedPopup` (`:5024`), which covers the normal path, but `Wh_ModUninit` (`:6190`) only runs `RemoveIslandGrid` (→ `DestroyExpandedPopup` → `UnregisterClassW`) *if the taskbar window is found*; if it isn't, the class survives the unload with a dangling proc and the next load hits exactly this case. Fix: register the class with the **mod's own** module handle (not `GetModuleHandleW(nullptr)`), and treat a registration failure as fatal rather than reusing an existing class. Using the mod DLL's `HINSTANCE` also lets `RegisterClass`/`UnregisterClass` pair up correctly across reloads.

**2. Purely poll-based; consider GSMTC change events.** The media worker wakes every `PollIntervalMs` (default 1000 ms, floor 500 ms) *forever* in every `explorer.exe` — even with no media playing and the popup closed — and each wake does a cross-thread `RunFromWindowThread` (a `SetWindowsHookExW(WH_CALLWNDPROC)` + synchronous `SendMessage` to the taskbar thread) to refresh the island. `GlobalSystemMediaTransportControlsSessionManager` exposes `SessionsChanged` / `CurrentSessionChanged`, and each session exposes `MediaPropertiesChanged` / `PlaybackInfoChanged` / `TimelinePropertiesChanged`. Subscribing to those would make the refresh event-driven and idle when nothing changes, which is the approach the maintainer prefers over constant polling. You already extrapolate the playback position locally in the render loop, so a steady timer isn't required for smooth progress. This is the most impactful design change short of item 1.

<p>

Minor polish — none of this meaningfully affects users, so it's your call.

- **Seek handler is the one XAML event handler without a `try/catch`.** Every other handler in the file wraps its body defensively, but the progress-bar `canvas.PointerPressed` lambda (`:3623`) calls `GetCurrentPoint(...).Position()` / `ActualWidth()` / `ActualHeight()` bare. An exception escaping a XAML input handler in an in-process island can tear down Explorer; wrap it in `try { … } catch (...) {}` to match the rest of the file.
- **`g_settings` data race on settings change.** `LoadSettings` (called from `Wh_ModSettingsChanged` on an arbitrary thread) reassigns the `std::wstring` members of `g_settings` while the taskbar/media threads read them (e.g. `g_settings.material` in `IslandBackgroundBrush`, `g_settings.pollIntervalMs` in the worker's `wait_for`). It only races during a settings change and is very unlikely to bite in practice, but a `std::wstring` read torn against a reassignment is genuine UB — a small lock or an apply-on-the-taskbar-thread snapshot would close it.
- **`GetStringSetting` helper vs. `WindhawkUtils::StringSetting`.** The custom `Wh_GetStringSetting` + `Wh_FreeStringSetting` wrapper works, but `WindhawkUtils::StringSetting` is the idiomatic RAII form.
- **`MediaThreadProc` calls `winrt::init_apartment(multi_threaded)` but never `uninit_apartment`.** Harmless (the thread is joined at teardown), but pairing them is cleaner.

</p>
</details>

<p>

- **Album-art color extraction reads premultiplied pixels as straight color.** The accent/gradient samplers at `:1624`, `:1661`, `:1920`, `:1953` build colors straight from `32bppPBGRA` (premultiplied) pixels without dividing by alpha — unlike `ExtractAlbumAccentColor` (`:2125`), which un-premultiplies correctly. For semi-transparent thumbnail pixels this biases the derived color toward black. Benign in practice (GSMTC album art is opaque JPEG/PNG), but it's an inconsistency worth aligning.
- **Compact progress bar steps at the poll rate.** The compact island's progress value is only updated per poll (up to ~1 s granularity), so it advances in visible steps while playing, in contrast to the smoothly-extrapolated expanded progress. Switching to events (item 4) plus a lightweight local extrapolation for the compact bar would smooth this out.
- **Media-switch debounce is heavily provider-tuned.** The multi-poll "stable across N samples / M ms" heuristics in `SetMedia` (`:860`) are tuned around specific browser GSMTC quirks; they may under/over-debounce for some players. Not a bug, just a fragility to keep in mind as media sources vary.

</p>
</details>

usho-lear:
### Changelog since the latest review

- Fixed progress-bar pointer handling, including pointer capture, cancellation, and release behavior, so dragging previews update in real time and the seek is committed only when released.
- Increased the interactive area around the popup progress bar, making it easier and more reliable to drag without changing its visual size.
- Added support for GSMTC timelines with non-zero start offsets and a relative/absolute timestamp fallback for broader provider compatibility.
- Added local interpolation for the compact progress indicator to remove visible step-by-step movement between media timeline updates.
- Improved active-session selection by monitoring all available media sessions and preferring the session that is currently playing. This allows the island to switch correctly from paused music to browser playback.
- Refreshed the target media session before executing playback commands, preventing controls from being sent to a stale session after the active source changes.
- Restored Apple Music seeking with a narrowly scoped UI Automation fallback. It targets Apple Music's writable `LCDScrubber` only when Apple Music reports GSMTC seeking as unavailable; other players continue to use the standard GSMTC path.

The updated progress and seeking behavior has been tested locally with both browser media and Apple Music.

m417z:
I tried the mod again with Edge and YouTube. The thumbnail image isn't shown. It works with Taskbar Fluent Media Player. Do you want to fix it, or do you prefer to address it in a future version?

usho-lear:
> I tried the mod again with Edge and YouTube. The thumbnail image isn't shown. It works with Taskbar Fluent Media Player. Do you want to fix it, or do you prefer to address it in a future version?

If what is displayed is an abstract gradient image based on the cover’s theme color, then there may be an issue with the default setting for the artwork type. If a blue placeholder cover is displayed, then it may be caused by something else.

When I set **Low-res video artwork mod** to **browser thumbnail**, the cover displays correctly when playing YouTube videos in the Edge browser. Could you please describe what exactly is being displayed on your end?

m417z:
Here's what I see, using a fairly standard Win11 setup.

image

m417z:
Changing "Low-res video artwork mode" to "Browser thumbnails" works, but I don't think it's an optimal default:
* The taskbar thumbnail is small and I don't see a reason to use a gradient instead.
* The popup thumbnail is larger, and the scale could be large. Still, I think it's better than just a gradient. As a middle ground, you could show the thumbnail in its original size, and have the blur as the container background.

image

usho-lear:
> Here's what I see, using a fairly standard Win11 setup.
> 
> image

I think I roughly understand what the issue is. I tried setting **Low-res video artwork mode** to **mesh gradient**, and I got the same result. You could try setting this option to **browser thumbnail** and see whether it shows the original cover before post-processing.

This was originally a fallback solution I added because the cover artwork from browser content had too low a resolution. After the first uploaded version, I changed the default to **browser thumbnail** to address the issue you initially reported. It’s possible that the settings file from the earliest version was left over, which caused it to not display correctly.

（i somehow wrote this before your next replay jez

m417z:
> It’s possible that the settings file from the earliest version was left over, which caused it to not display correctly.

That's probably it. Looks good, thanks.

PR #4659, opened 2026-07-02
MacOS Minimize Animation v2.1.0

Abdullah-Masood-05:
Updates **MacOS Minimize Animation** to **v2.1.0**.

Fixes the genie animation not playing for apps whose windows bypass the classic `ShowWindow` / `WM_SYSCOMMAND` paths:

- **New minimize hooks**: `ShowWindowAsync`, `SetWindowPlacement`, and `CloseWindow` (which minimizes, despite its name). Apps with custom title bars - e.g. the Zed editor - minimize through these, so their own minimize button previously showed no animation while taskbar minimize worked. Also verified to fix minimize for Store/UWP apps such as the modern Media Player.
- **New launch path**: `SetWindowPos` + `SWP_SHOWWINDOW` for frameworks that show their first window without `ShowWindow`, plus a bounded wait for DWM-cloaked (Store/UWP) windows to uncloak before snapshotting, so the launch genie doesn't warp a black frame.
- **Robustness**: a one-genie-per-window guard - with several minimize entry points hooked, a single gesture could otherwise start two overlapping ghost animations. New hook paths also skip hidden/already-minimized windows (e.g. apps restoring a saved minimized session at startup).

Existing behavior for classic Win32 windows is unchanged.

### DIFF
--- a/mods/macos-minimize-animation.wh.cpp
+++ b/mods/macos-minimize-animation.wh.cpp
@@ -2,7 +2,7 @@
 // @id              macos-minimize-animation
 // @name            MacOS Minimize Animation
 // @description     Smooth macOS-style genie minimize and restore (open) animations for every window.
-// @version         2.0.0
+// @version         2.1.0
 // @author          Abdullah Masood
 // @github          https://github.com/Abdullah-Masood-05
 // @include         *
@@ -14,12 +14,10 @@
 /*
 # MacOS Minimize Animation
 
-![Demo](https://raw.githubusercontent.com/Abdullah-Masood-05/my-windhawk-mods-media/main/macos-minimize-animation.gif)
-
 Brings the classic macOS **genie** effect to Windows. When you minimize a window
 it warps and flows down into the taskbar; when you restore it, it flows back out.
 
-Eye candy for something you do a hundred times a day.
+Definitely not inspired from MacOS.
 
 ## See it in action
 - Compile the mod with the button on the left or with Ctrl+B.
@@ -41,8 +39,11 @@ Eye candy for something you do a hundred times a day.
   your cursor was when you minimized) and aims the genie at that spot next time.
 
 ## How it works
-The mod hooks `ShowWindow` and `DefWindowProcW` to catch minimize / restore
-requests. It snapshots the window, then on a dedicated high-priority thread it
+The mod hooks `ShowWindow`, `ShowWindowAsync`, `SetWindowPlacement`, `CloseWindow`,
+`SetWindowPos` and `DefWindowProcW` to catch minimize / restore / first-show
+requests - including apps with custom title bars (e.g. Zed) and Store/UWP apps,
+whose minimize buttons bypass the classic paths. It snapshots the window, then on
+a dedicated high-priority thread it
 draws a transparent layered "ghost" window that warps the snapshot frame by frame
 into the taskbar, leaving the real window to minimize behind it without the
 system's own animation getting in the way.
@@ -50,11 +51,6 @@ system's own animation getting in the way.
 ## Settings
 - **Animation duration** - how long the effect lasts (50-2000 ms).
 - **Animate window restore (open)** - also play the reverse genie when restoring.
-- **Animate app launch (experimental)** - also play the genie when an application
-  window first opens. **Experimental and off by default**: it may briefly flash the
-  window before animating, may not catch borderless apps, and can occasionally fire
-  on splash / dialog windows. Leave it off and the mod behaves exactly like plain
-  minimize / restore.
 
 ## Notes
 - Works on all top-level windows; child / tiny / hidden windows are skipped.
@@ -99,6 +95,19 @@ DefWindowProcW_t DefWindowProcW_Original;
 typedef BOOL (WINAPI *ShowWindow_t)(HWND hWnd, int nCmdShow);
 ShowWindow_t ShowWindow_Original;
 
+typedef BOOL (WINAPI *ShowWindowAsync_t)(HWND hWnd, int nCmdShow);
+ShowWindowAsync_t ShowWindowAsync_Original;
+
+typedef BOOL (WINAPI *SetWindowPlacement_t)(HWND hWnd, const WINDOWPLACEMENT* lpwndpl);
+SetWindowPlacement_t SetWindowPlacement_Original;
+
+typedef BOOL (WINAPI *CloseWindow_t)(HWND hWnd);
+CloseWindow_t CloseWindow_Original;
+
+typedef BOOL (WINAPI *SetWindowPos_t)(HWND hWnd, HWND hWndInsertAfter, int X, int Y,
+                                      int cx, int cy, UINT uFlags);
+SetWindowPos_t SetWindowPos_Original;
+
 struct MacGenieAnimData {
     HWND hRealWnd;
     HBITMAP hBitmap;      // snapshot of the window (a DDB at original size)
@@ -114,6 +123,7 @@ struct MacGenieAnimData {
 std::unordered_map<HWND, HBITMAP> g_SnapshotCache;
 std::unordered_map<HWND, int> g_IconPositions; // Remembers where icons live
 std::unordered_set<HWND> g_LaunchSeen;         // windows we've already shown/animated once
+std::unordered_set<HWND> g_AnimActive;         // windows with a genie currently in flight
 std::mutex g_CacheMutex;
 
 // --- SETTINGS ---
@@ -420,6 +430,10 @@ DWORD WINAPI MacGenieAnimThread(LPVOID lpParam) {
     DeleteDC(hSrcDC);
     ReleaseDC(NULL, hScreenDC);
     DestroyWindow(hGhost);
+    {
+        std::lock_guard<std::mutex> lock(g_CacheMutex);
+        g_AnimActive.erase(data->hRealWnd);
+    }
     delete data;
     return 0;
 }
@@ -435,10 +449,13 @@ void StartMacGenieAnim(HWND hWnd, BOOL rising) {
 
     if (w <= 0 || h <= 0) return;
 
-    // Any window we animate (minimize/restore/launch) counts as "seen", so the
-    // launch WinEvent hook won't also fire on it (e.g. on a later restore).
+    // One genie per window at a time: with several minimize entry points hooked
+    // (ShowWindow, ShowWindowAsync, SetWindowPlacement, WM_SYSCOMMAND, ...), a single
+    // gesture could otherwise start two overlapping ghosts. Any window we animate
+    // also counts as "seen", so the launch path won't fire on it later (e.g. restore).
     {
         std::lock_guard<std::mutex> lock(g_CacheMutex);
+        if (!g_AnimActive.insert(hWnd).second) return;
         g_LaunchSeen.insert(hWnd);
     }
 
@@ -532,6 +549,8 @@ void StartMacGenieAnim(HWND hWnd, BOOL rising) {
         MacGenieSetDwmTransitions(hWnd, TRUE);
         DeleteObject(data->hBitmap);
         delete data;
+        std::lock_guard<std::mutex> lock(g_CacheMutex);
+        g_AnimActive.erase(hWnd);
     }
 }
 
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

---

Nice update — the new minimize entry points and the one-genie guard are well thought out. A few things to fix, the first one being a real regression the guard introduces.

**1. The new one-genie guard can leave a window stuck fully invisible.** Every *rising* caller hides the window (`WS_EX_LAYERED` + alpha 0) **before** calling `StartMacGenieAnim(hWnd, TRUE)`:
- `ShowWindow_Hook` restore branch (lines 608-612)
- `DefWindowProcW_Hook` `SC_RESTORE` branch (lines 658-662)
- launch path via `MacGenieLaunchPrepare` (line 579)

Only `MacGenieAnimThread` restores alpha back to 255 and removes the layered style, at the very end of the animation. But the new guard bails out early *before* spawning that thread:

```cpp
if (!g_AnimActive.insert(hWnd).second) return;   // line 458 — nothing undoes the alpha-0 hide
```

So whenever the guard blocks a rising animation, the caller has already set alpha 0 and nobody ever sets it back. Repro: minimize a window and, while its ~450 ms minimize genie is still playing, click the taskbar to restore it. The minimize genie still holds `g_AnimActive[hWnd]`, so the restore's `StartMacGenieAnim(hWnd, TRUE)` returns immediately → the window is restored but stays completely transparent. Because it's no longer `IsIconic`, a second taskbar click won't re-enter the restore branch, so the window can remain invisible indefinitely. (v2.0.0 didn't have this because the rising thread always ran and restored alpha.)

Fix: reserve the `g_AnimActive` slot **before** the caller hides the window, and only hide/animate if the reservation succeeded — or, if the guard blocks a `rising` animation, undo the hide before returning:

```cpp
if (!g_AnimActive.insert(hWnd).second) {
    if (rising) {   // caller already set alpha 0; don't leave it invisible
        SetLayeredWindowAttributes(hWnd, 0, 255, LWA_ALPHA);
        MacGenieSetDwmTransitions(hWnd, TRUE);
        // (and drop WS_EX_LAYERED if the window wasn't originally layered)
    }
    return;
}
```

**2. The demo GIF was removed from the README.** This PR deletes the only visual (`Demo`), yet the README still has a "See it in action" section that now shows nothing. This mod is pure eye candy — a genie animation is exactly the kind of effect the catalog wants demonstrated — so please keep a GIF/screenshot in the README (the `raw.githubusercontent.com` host is fine as-is).

**3. Detached animation/launch threads are never awaited on unload (pre-existing, but this PR widens the exposure).** `MacGenieAnimThread` and `MacGenieLaunchThread` are started with `CreateThread` + `CloseHandle` and run for up to ~2 s (animation) / ~1.6 s (the new bounded DWM-cloak wait in the launch thread). `Wh_ModUninit` doesn't wait for them. Since Windhawk unmaps the mod DLL after `Wh_ModUninit`, a thread still executing mod code when the user disables/updates the mod risks crashing that process — and with `@include *` that's *any* process currently mid-animation. Disabling an animation mod while an animation is on screen is a pretty natural action, so this is reachable. Consider tracking outstanding worker threads and waiting for them (or signaling them to abort and restoring any touched windows) in `Wh_ModBeforeUninit`/`Wh_ModUninit`. The new launch thread's up-to-1.6 s wait — during which the window is held invisible — makes the window notably larger than in 2.0.0.

<p>

- **README/settings drift.** The README "## Settings" section no longer lists `launch_animation`, but the setting is still present in the `==WindhawkModSettings==` block and this PR actually *expands* its plumbing (`SetWindowPos` + `ShowWindowAsync` launch parity). It's a bit contradictory to grow the launch feature while dropping its README docs — please keep the README and the settings UI in sync (either document it again, or if you're deprecating it, remove the code paths too).
- The new hooks use the raw `Wh_SetFunctionHook((void*)…, (void**)&…)` form. `WindhawkUtils::SetFunctionHook()` is the type-safe equivalent and would catch signature mismatches at compile time. The file already uses the raw form throughout, so this is whole-file consistency rather than a blocker.

</p>
</details>

<p>

- **`SetWindowPos` is hooked in every process but only ever serves the experimental, off-by-default launch animation.** When `launch_animation` is disabled (the default), the hook does nothing useful yet sits on a very hot, global API. The fast path is just a `uFlags & SWP_SHOWWINDOW` bit test, so per-call cost is negligible — but if you want to trim it, you could gate the launch-only hooks (`SetWindowPos`, and the launch branch of `ShowWindowAsync`) so they're only installed when launch animation is actually enabled.
- **Launch reveal holds the window invisible for up to ~1.6 s.** `MacGenieLaunchPrepare` sets alpha 0, then `MacGenieLaunchThread` waits `Sleep(60)` + up to 30×50 ms for the DWM cloak to lift. If a Store/UWP app takes longer than that to render, the user sees nothing for over a second and then possibly a black genie (`PrintWindow` on a still-cloaked window). Inherent to the experimental approach — noting it as an FYI.
- **The new minimize paths have no restore counterpart.** `ShowWindowAsync`/`SetWindowPlacement`/`CloseWindow` deliberately only handle minimize, so custom-titlebar apps (e.g. Zed) that *also* restore through those APIs won't get the reverse genie. You've documented the reasoning in the comment; noting it as expected behavior, not a bug.

</p>
</details>

Abdullah-Masood-05:
Thank you for the thorough review - all three issues addressed in the latest commit:

1. **Stuck-invisible guard path - fixed.** When the one-genie guard (or mod unload) blocks a *rising* animation, the window's alpha-0 hide is now undone on the spot (alpha restored to 255, DWM transitions re-enabled) before returning, so restoring a window while its minimize genie is still in flight can no longer leave it invisible. `WS_EX_LAYERED` is left in place since the original style isn't known at that point; layered + alpha 255 renders normally.

2. **Demo GIF restored** at the top of the readme, and the readme Settings section documents `launch_animation` again so it matches the settings UI.

3. **Worker threads drained on unload.** Animation/launch workers now register in an atomic counter (incremented before `CreateThread`, decremented on every exit path including `CreateThread` failure). `Wh_ModBeforeUninit` sets an unloading flag and waits for the counter to drain (capped at ~3 s, normally well under 100 ms) - the animation loop checks the flag every composed frame and the launch worker every 50 ms, and both restore the real window's visibility/transitions when aborting. New animations are refused while unloading.

On the optional notes: kept the raw `Wh_SetFunctionHook` form for whole-file consistency, and left the launch-only hooks always installed for now given the fast-path cost is a single flag test - happy to gate them on the setting if you'd prefer.

m417z:
### Submission review

---

**`WS_EX_LAYERED` is never removed after a rising (restore / launch) animation — the cleanup is dead code.** Every rising caller ORs `WS_EX_LAYERED` into the window *before* calling `StartMacGenieAnim(hWnd, TRUE)`:
- `ShowWindow_Hook` restore branch (line 648)
- `DefWindowProcW_Hook` `SC_RESTORE` branch (line 698)
- `MacGenieLaunchPrepare` (line 615)

But `StartMacGenieAnim` captures the "original" style *after* the caller already added the flag:

```cpp
data->originalExStyle = GetWindowLongPtrW(hWnd, GWL_EXSTYLE);   // line 527 — already has WS_EX_LAYERED
```

so the restore guard in `MacGenieAnimThread` can never fire:

```cpp
if (!(data->originalExStyle & WS_EX_LAYERED)) {                 // line 429 — always false for rising
    SetWindowLongPtrW(data->hRealWnd, GWL_EXSTYLE, data->originalExStyle);
}
```

The result: every window that gets restored (and restore is on by default) keeps `WS_EX_LAYERED` permanently — even after the mod is disabled or uninstalled. It is mostly invisible (layered + alpha 255 renders normally), but it is a genuine state leak that works against Windhawk's reversibility principle, and it can interfere with apps that manage their own layered attributes / DirectComposition surfaces. This is the same gap you consciously left in the blocked-guard path last round (`SetLayeredWindowAttributes(hWnd, 0, 255, LWA_ALPHA)` without dropping the style) — it just turns out the *main* path never drops it either.

Fix: capture the true pre-layered ex-style at the hook site (where you read `exStyle` before ORing the flag) and thread it into `MacGenieAnimData` instead of re-reading it inside `StartMacGenieAnim`. For example, add a parameter:

```cpp
// hook, before hiding the window:
LONG_PTR origEx = GetWindowLongPtrW(hWnd, GWL_EXSTYLE);
SetWindowLongPtrW(hWnd, GWL_EXSTYLE, origEx | WS_EX_LAYERED);
...
StartMacGenieAnim(hWnd, TRUE, origEx);   // pass the real original down
```

and set `data->originalExStyle = origEx;` from the argument (falling paths can pass the current style, since they never touch layering). Then the existing `if (!(originalExStyle & WS_EX_LAYERED))` cleanup starts working, and you can apply the same restore in the blocked-guard path (line 490) too.

<p>

Minor polish — none of this affects users in the default configuration, so it is your call.

- **The new unload drain does not cover the launch path.** `StartMacGenieAnim` refuses to start a worker once `g_unloading` is set (line 477), so the minimize path is safe during teardown. But `MacGenieLaunchPrepare` / `MacGenieLaunchCommit` never check `g_unloading` — a launch-eligible window shown during the drain will still `fetch_add(1)` and spawn `MacGenieLaunchThread`. If that increment lands *after* the drain loop already observed `g_workerCount == 0`, `Wh_ModBeforeUninit` returns, the DLL is unmapped, and the freshly-spawned worker crashes its host. The window is extremely narrow (needs `launch_animation` on, which is experimental and off by default, plus a launch in the sub-millisecond gap after the drain), but it is the one hole left in the otherwise-solid drain. Mirror the minimize path by bailing early:
  ```cpp
  static bool MacGenieLaunchPrepare(HWND hWnd, int nCmdShow) {
      if (g_unloading.load(std::memory_order_relaxed)) return false;
      if (!g_launchAnimation.load(std::memory_order_relaxed)) return false;
      ...
  ```

</p>
</details>

<p>

- **Overlapping minimize to restore shows the real window without a rising genie.** If the user restores a window while its ~450 ms minimize genie is still in flight, the one-genie guard blocks the rising animation and (correctly) just reveals the real window — but the still-running *falling* ghost keeps animating downward for a moment, so you briefly see the restored window plus a descending ghost. Harmless and rare; noting it since it is a side effect of the new guard.
- **DWM transitions can be left disabled on a window if a minimize gesture lands during unload.** `MacGenieTryMinimizeAnim` calls `MacGenieSetDwmTransitions(hWnd, FALSE)` before `StartMacGenieAnim`, which then refuses (unloading) and returns without re-enabling them — and there is no in-flight worker left to restore them. Transient and cosmetic (until the window closes), and only during the brief unload window; FYI, not something to block on.

</p>
</details>

Abdullah-Masood-05:
Fixed - thank you for catching that the cleanup was dead code.

**WS_EX_LAYERED leak**: `StartMacGenieAnim` now takes the window's true pre-hide extended style as a parameter instead of re-reading it after the caller already ORed in `WS_EX_LAYERED`. All three rising call sites pass the style they captured before hiding (the launch path threads it through a small `MacGenieLaunchData` struct into the worker, since the thread previously only received the HWND). With the real original style in `data->originalExStyle`, the existing `if (!(originalExStyle & WS_EX_LAYERED))` cleanup in the animation thread now actually removes the bit after every restore/launch - and the same restore was added to every rollback path: the blocked-guard branch, the launch worker's early exit, and both `CreateThread`-failure branches. Windows that were genuinely layered to begin with keep their bit. Falling paths never touch layering and just pass the current style.

Also took both optional/FYI items:
- `MacGenieLaunchPrepare` now bails first thing when `g_unloading` is set, closing the drain race where a launch worker could be spawned after `Wh_ModBeforeUninit` already observed a zero worker count.
- A minimize gesture landing during unload no longer leaves DWM transitions force-disabled: when the guard blocks a falling animation and no in-flight worker owns the window, transitions are re-enabled on the spot.

Abdullah-Masood-05:
One more commit on top of the review fixes - user-visible polish found while testing under CPU throttling (battery saver), where two timing artifacts that are nearly invisible on fast hardware become obvious:

- **Restores are now hidden via `DWMWA_CLOAK` instead of `WS_EX_LAYERED` + alpha 0.** Freshly adding the layered style forces DWM to rebuild the window's redirection surface, and until that lands the bare frame is composed for a few frames - a visible flash on every restore (it was previously masked by the very leak this PR fixes, since windows stayed layered forever). A cloaked window is not rendered at all (frame included), has no rebuild lag, and the window's styles are never touched on the restore path - which also structurally removes the `WS_EX_LAYERED` leak class there. The launch path keeps the layered mechanism (cloaking would collide with its wait-for-uncloak logic for Store apps) with the true pre-hide style threaded through as before.
- **The real minimize is now gated on the genie's first frame being composed.** Previously the hook let the actual minimize run immediately while the animation thread was still starting up, so the window vanished a few frames before the ghost appeared - a background flash in the gap, clearly visible when throttled. The hook now waits (capped at 200 ms; two duplicated event handles so neither thread can touch a freed handle) until the animation thread has composed its first frame via `DwmFlush`, then performs the minimize invisibly underneath the ghost. Applies to every minimize entry point; worst case on timeout is the old behavior.

Both verified on a throttled machine: no frame flash on restore, no gap flash on minimize.

PR #4685, opened 2026-07-05
Added Taskbar Elastic WinUI Pill mod

Lockframe:
Contains fixes and more animation modes compared to the Discord release.

## Mod authorship

This mod was created by:

- - [X] Claude
- - [ ] ChatGPT
- - [X] Gemini

### DIFF
--- a/mods/taskbar-elastic-pill.wh.cpp
+++ b/mods/taskbar-elastic-pill.wh.cpp
@@ -1376,7 +1376,7 @@ HMODULE GetTaskbarViewModuleHandle() {
 
 bool HookTaskbarViewDllSymbols(HMODULE module) {
     // Taskbar.View.dll, ExplorerExtensions.dll
-    WindhawkUtils::SYMBOL_HOOK taskbarViewDllHooks1[] = {
+    WindhawkUtils::SYMBOL_HOOK taskbarViewHooks1[] = {
         {
             {LR"(private: void __cdecl winrt::Taskbar::implementation::TaskListButton::UpdateVisualStates(void))"},
             &TaskListButton_UpdateVisualStates_Original,
@@ -1384,13 +1384,13 @@ bool HookTaskbarViewDllSymbols(HMODULE module) {
         }
     };
     bool success = true;
-    if (!WindhawkUtils::HookSymbols(module, taskbarViewDllHooks1, ARRAYSIZE(taskbarViewDllHooks1))) {
+    if (!WindhawkUtils::HookSymbols(module, taskbarViewHooks1, ARRAYSIZE(taskbarViewHooks1))) {
         Wh_Log(L"Failed to hook Taskbar.View.dll (TaskListButton)");
         success = false;
     }
 
     // Taskbar.View.dll, ExplorerExtensions.dll
-    WindhawkUtils::SYMBOL_HOOK taskbarViewDllHooks2[] = {
+    WindhawkUtils::SYMBOL_HOOK taskbarViewHooks2[] = {
         {
             {
                 L"public: void __cdecl winrt::Taskbar::implementation::TaskbarResources::OnExperienceToggleButtonVisualStateChanged(struct winrt::Taskbar::ITaskbarButton const &,struct winrt::Taskbar::TaskbarButtonVisualStateChangedEventArgs const &)",
@@ -1400,7 +1400,7 @@ bool HookTaskbarViewDllSymbols(HMODULE module) {
             OnExperienceToggleButtonVisualStateChanged_Hook,
         }
     };
-    if (!WindhawkUtils::HookSymbols(module, taskbarViewDllHooks2, ARRAYSIZE(taskbarViewDllHooks2))) {
+    if (!WindhawkUtils::HookSymbols(module, taskbarViewHooks2, ARRAYSIZE(taskbarViewHooks2))) {
         Wh_Log(L"Failed to hook Taskbar.View.dll (OnExperienceToggleButtonVisualStateChanged)");
     }
 
@@ -1413,7 +1413,7 @@ HMODULE GetSearchUxModuleHandle() {
 
 bool HookSearchUxDllSymbols(HMODULE module) {
     // SearchUx.UI.dll
-    WindhawkUtils::SYMBOL_HOOK searchUxDllHooks[] = {
+    WindhawkUtils::SYMBOL_HOOK searchUxHooks[] = {
         {
             {LR"(private: void __cdecl winrt::SearchUx::SearchUI::implementation::SearchIconButton::UpdateVisualStates(void))"},
             &SearchIconButton_UpdateVisualStates_Original,
@@ -1421,10 +1421,10 @@ bool HookSearchUxDllSymbols(HMODULE module) {
         }
     };
 
-    if (!WindhawkUtils::HookSymbols(module, searchUxDllHooks, ARRAYSIZE(searchUxDllHooks))) {
+    if (!WindhawkUtils::HookSymbols(module, searchUxHooks, ARRAYSIZE(searchUxHooks))) {
         Wh_Log(L"Failed to hook SearchUx.UI.dll (SearchIconButton::UpdateVisualStates). Trying PlayStateChange...");
         // SearchUx.UI.dll
-        WindhawkUtils::SYMBOL_HOOK searchUxDllHooks2[] = {
+        WindhawkUtils::SYMBOL_HOOK searchUxHooks2[] = {
             {
                 {
                     L"protected: void __cdecl winrt::SearchUx::SearchUI::implementation::SearchIconButton::PlayStateChange(void)",
@@ -1434,7 +1434,7 @@ bool HookSearchUxDllSymbols(HMODULE module) {
                 SearchIconButton_UpdateVisualStates_Hook,
             }
         };
-        if (!WindhawkUtils::HookSymbols(module, searchUxDllHooks2, ARRAYSIZE(searchUxDllHooks2))) {
+        if (!WindhawkUtils::HookSymbols(module, searchUxHooks2, ARRAYSIZE(searchUxHooks2))) {
             Wh_Log(L"Failed to hook SearchUx.UI.dll symbols entirely.");
             return false;
         }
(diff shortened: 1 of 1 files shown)


mods/taskbar-elastic-pill.wh.cpp
```diff
@@ -0,0 +1,794 @@
...
+  $options:
+  - stretch: Stretch (WinUI accurate)
+  - squish: Squish
+  - stretchsquish: Stretch-Squish
+  - bounce: Bounce
+  - linear: Linear
+  - easein: Ease-In
+  - easeout: Ease-Out
+  - easeinout: Ease-In-Out
+  - stretchsquish: Stretch-Squish
```
m417z: Remove duplicate option.

m417z:
### Submission review

---

**1. `SearchUx.UI.dll` late-load is missed when the mod is enabled mid-session, so the "Track System buttons" search pill silently never works.** In `Wh_ModInit`, the `LoadLibraryExW` hook is only installed in the `else` branch — i.e. only when `Taskbar.View.dll` is *not* yet loaded:

```cpp
if (HMODULE m = GetTaskbarViewModuleHandle()) {
    ... hook directly ...          // no LoadLibraryExW hook installed
} else {
    ... hook LoadLibraryExW ...    // catches BOTH taskbar and searchux later
}
if (HMODULE sm = GetSearchUxModuleHandle()) {
    ... hook directly ...
}
// taskbar present + searchux absent  ->  searchux late load is never caught
```

When the mod is toggled on while Explorer is already running, `Taskbar.View.dll` is loaded, so the `LoadLibraryExW` hook is skipped. If `SearchUx.UI.dll` then loads later (search opened for the first time), `SearchIconButton::UpdateVisualStates` is never hooked and the search-button pill doesn't appear. Install the `LoadLibraryExW` hook whenever *either* target module is missing at init (or just always), so any later load is caught regardless of init state.

**2. `HookSymbols` is called twice against `Taskbar.View.dll` in a single load.** `HookTaskbarViewDllSymbols` resolves `taskbarViewHooks1` and then `taskbarViewHooks2` in two separate `HookSymbols(module, ...)` calls. Each call for the same module invalidates the symbol cache and forces a full re-resolution (slow — a PDB fetch in the worst case, adding to taskbar load time). Use a single array, and mark some symbols as optional if needed.

Also, remove the mangled variant:
`?OnExperienceToggleButtonVisualStateChanged@...`

**3. Replace the `IsBadReadPtr`-based vtable probing with the standard ABI cast.** `GetFrameworkElementFromNative` / `IsSafeToQueryInterface` scan offsets 0–9 of `pThis`, validate each with `IsBadReadPtr`, and check the vtable module — a fragile heuristic, and `IsBadReadPtr` is a [known-broken API](https://devblogs.microsoft.com/oldnewthing/20060927-07/?p=29563) (it can hide real bugs and touch guard pages). Every other taskbar mod converts `pThis` to the XAML element at the known fixed offset:

```cpp
void* iunknownPtr = (void**)pThis + 3;
winrt::Windows::Foundation::IUnknown iunknown;
winrt::copy_from_abi(iunknown, iunknownPtr);
auto element = iunknown.as<FrameworkElement>();
```

See [taskbar-labels.wh.cpp#L1346](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/taskbar-labels.wh.cpp#L1346). This removes the probing loop, the `IsBadReadPtr` calls, and `IsSafeToQueryInterface` entirely.

**4. Global objects with non-trivial destructors run at process shutdown, when `Wh_ModUninit` does not.** On an Explorer restart / sign-out / reboot, Windhawk does **not** call `Wh_ModUninit`; the CRT just runs the destructors of your global `std::map`/`std::vector` globals during `DLL_PROCESS_DETACH`. Several of these hold strong WinRT/COM references — `g_easingCaches` (each `EasingCache` owns a `Compositor` + 7 `CompositionEasingFunction`s) and `g_attachedEvents` (each holds a `CoreDispatcher`). Releasing those into `Windows.UI.Xaml`/`Windows.UI.Composition` during process teardown can crash or hang the host. Prefer keeping the real teardown in `Wh_ModUninit` (which you already do) and making the globals themselves trivially-destructible — e.g. heap-allocate the caches behind a raw pointer that's intentionally leaked at shutdown, so nothing calls `Release` on `DLL_PROCESS_DETACH`.

<p>

- **Use `WindhawkUtils::StringSetting` (RAII) instead of raw `Wh_GetStringSetting` + `Wh_FreeStringSetting`** in `LoadSettings` — it frees automatically and is harder to leak. Also, `Wh_GetStringSetting` never returns `NULL` (it returns `L""` on unset/error), so the `if (dimStr)` / `if (animStr)` / etc. null-checks are unnecessary.
- **`g_easingCaches` per-thread dead-thread reaping is over-engineered.** The `OpenThread` + `WaitForSingleObject` scan in `UpdatePillPosition` only triggers when `g_easingCaches.size() > 30`, which effectively never happens — the taskbar UI runs on a single thread. This looks like an AI artifact; a single cache (or a plain per-thread map cleared in `Wh_ModUninit`) would be simpler.
- **`Wh_ModSettingsChanged` only refreshes one pill.** The loop `break`s after dispatching to the first context (the code comment already flags this), so on a multi-monitor setup only one taskbar's pill updates immediately on a settings change; the others catch up on the next hover/activation. Dispatch to each distinct context's dispatcher instead.
- **Dead default:** `Settings::PillMarginBottom = 4` is never used — `LoadSettings` resets margins to `0, 0` before parsing and the settings-block default is `"0, 0"`. Harmless, but the `4` is misleading.

</p>
</details>

<p>

- **Icon-color mode renders each app icon via `RenderTargetBitmap` + `GetPixelsAsync`** to pick a dominant color. It's cached (with LRU eviction) and gated behind opt-in, so cost is amortized, but the first activation of each app does a full off-thread render — expected, just noting the trade-off.

</p>
</details>

m417z:
Missed the `searchUxHooks` - same issue.

---

- **`HookSymbols` is called twice on the same module in the SearchUx fallback.** In `HookSearchUxDllSymbols` (lines 1358-1387), if `SearchIconButton::UpdateVisualStates` doesn't resolve, you call `HookSymbols(module, ...)` a second time on the same `SearchUx.UI.dll` for `PlayStateChange`. Each additional `HookSymbols` call for a module invalidates its symbol cache and forces a re-resolution. Since the two symbols are alternative entry points for the same handler, put them in a **single** `SYMBOL_HOOK` array (both marked optional) and resolve them in one call:

```cpp
WindhawkUtils::SYMBOL_HOOK searchUxHooks[] = {
    {
        {LR"(private: void __cdecl winrt::SearchUx::SearchUI::implementation::SearchIconButton::UpdateVisualStates(void))"},
        &SearchIconButton_UpdateVisualStates_Original,
        SearchIconButton_UpdateVisualStates_Hook,
        true  // optional
    },
    {
        {L"protected: void __cdecl winrt::SearchUx::SearchUI::implementation::SearchIconButton::PlayStateChange(void)"},
        &SearchIconButton_UpdateVisualStates_Original,
        SearchIconButton_UpdateVisualStates_Hook,
        true  // optional
    },
};
```

If both happen to resolve on some build, the handler just runs twice per state change, which is harmless. Alternatively, proceed in `SearchIconButton_UpdateVisualStates_Hook` only if `SearchIconButton_UpdateVisualStates_Original` is null.

- Remove the mangled variant: `"?PlayStateChange@...`.

<p>

- **Unload path is fragile — prefer synchronous cleanup in `Wh_ModBeforeUninit`.** All teardown happens in `Wh_ModUninit` (lines 1439-1556) via fire-and-forget `dispatcher.RunAsync` cleanup guarded by 2-second timeouts. Two concerns:

  - The many `RunAsync` callbacks the *hooks* post (`TaskListButton_UpdateVisualStates_Hook` line 914, the state-changed handlers, etc.) are independent queued delegates — revoking the event tokens during uninit does **not** cancel one that's already queued. Each only no-ops via the `if (g_unloading) return;` check *if the mod DLL is still mapped when it runs*. Windhawk removes the function hooks before `Wh_ModUninit` but does not drain the taskbar's dispatcher queue, so a callback that runs after the DLL is unmapped executes into freed code → intermittent explorer crash on disable, most likely on a busy taskbar.
  - The `for (...) { Sleep(10); }` busy-wait for `g_asyncExtractCount` (line 1442) and the `WaitForSingleObject(..., 2000)` (line 1552) mean unload can stall up to ~2s per stage, and if uninit ever runs on the same thread that would service those completions, the wait can never satisfy and always times out.

The established pattern (see [`taskbar-labels.wh.cpp` `Wh_ModBeforeUninit`](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/taskbar-labels.wh.cpp#L2199-L2211)) is to set `g_unloading` and do the restore **synchronously in `Wh_ModBeforeUninit`** — hooks are still active, the UI thread is guaranteed live, and a short `Sleep` lets the layout settle — rather than posting async work from `Wh_ModUninit`. Reworking teardown that way removes the reliance on timeouts and the after-unmap race.

- **Redundant NULL checks on string settings.** `WindhawkUtils::StringSetting::get()` never returns `NULL` (it's an empty string at worst), so `if (radiusStr.get())` (line 197), `if (animStr.get())` (205), `if (cmStr.get())` (219), `if (colorStr.get())` (228) are always true. You can drop the guards; the empty-string path is already handled (e.g. `std::stod(L"")` throws and is caught). It's clearer to check `radiusStr.get()[0]` / compare against `L""` when you actually need "is it set".

</p>
</details>

<p>

- **Icon-color mode can briefly show the previous app's color.** Extraction is async (`ExtractIconColor`), and the single shared `ElasticPill` is repositioned to whichever button is active. If the active app switches while an extraction is in flight, the completion at line 618 updates the shared pill's fill even though the pill may already be tracking a different button (the state re-check is per-`button`, not per-current-pill-target). It self-corrects on the next update, so it's cosmetic — just noting it.

- **`LayoutUpdated`-driven repositioning fires very frequently.** The per-grid `LayoutUpdated` handler (line 1052) runs on every layout pass of the whole RootGrid subtree and recomputes `TransformToVisual`. The `> 0.1f` guard keeps the steady-state cost to a transform computation, and there isn't an obviously cheaper way to keep the pill glued to a moving button, so this is just an FYI on the cost.

- **Icon-color mode is heavy machinery.** Rendering each taskbar icon to a `RenderTargetBitmap`, reading back pixels, bucketing, and HSL-adapting per app is a lot of work for a color pick (gated behind a non-default setting, LRU-capped at 100, so bounded). It works, but if you're trimming complexity it's the first candidate — some of it reads like AI-generated over-engineering.

</p>
</details>

PR #4694, opened 2026-07-05
Explorer Open Folder Icon 1.0.0

aubymori:
In Windows XP and before, File Explorer showed an open folder icon in the title bar and for selected folders in the Folders tree. This mod restores that behavior to Windows 10.

**Before**:
Before

**After**:
After

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/explorer-open-folder-icon.wh.cpp
+++ b/mods/explorer-open-folder-icon.wh.cpp
@@ -1,0 +1,164 @@
+// ==WindhawkMod==
+// @id              explorer-open-folder-icon
+// @name            Explorer Open Folder Icon
+// @description     Restores the open folder icon in Explorer's icon and tree view
+// @version         1.0.0
+// @author          aubymori
+// @github          https://github.com/aubymori
+// @include         explorer.exe
+// @compilerOptions -lshell32 -lcomctl32
+// @architecture    x86-64
+// @license         GPL-3.0
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Explorer Open Folder Icon
+In Windows XP and before, File Explorer showed an open folder icon in the title bar and
+for selected folders in the Folders tree. This mod restores that behavior to Windows 10.
+
+**Before**:
+
+![Before](https://raw.githubusercontent.com/aubymori/images/refs/heads/main/explorer-open-folder-icon/before.png)
+
+**After**:
+
+![After](https://raw.githubusercontent.com/aubymori/images/refs/heads/main/explorer-open-folder-icon/after.png)
+*/
+// ==/WindhawkModReadme==
+
+// ==WindhawkModSettings==
+/*
+- replace_downloads: true
+  $name: Replace Downloads Icon
+  $description: Also replaces the Downloads folder's icon with an open folder icon.
+*/
+// ==/WindhawkModSettings==
+
+#include <initguid.h>
+#include <windhawk_utils.h>
+#include <shlobj.h>
+#include <commctrl.h>
+#include <commoncontrols.h>
+
+bool g_fReplaceDownloads = false;
+
+/* Replace the window icon with the open folder. */
+int (*CShellBrowser__GetIconIndex_orig)(class CShellBrowser *);
+int CShellBrowser__GetIconIndex_hook(class CShellBrowser *pThis)
+{
+    int index = CShellBrowser__GetIconIndex_orig(pThis);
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

I haven't verified it, but if `pnm->hdr.hwndFrom` is the same value, it's indeed a better option.

---

Small, focused mod and the overall approach is sound (force-load `ExplorerFrame.dll` from System32, one `HookSymbols` call, no self-installed subclasses to clean up). One robustness item is worth addressing:

**Replace the hardcoded `+ 47` member offset with `pnm->hdr.hwndFrom`.** In `CNscTree__OnGetDisplayInfo_hook` you reach the tree HWND via `CNscTree__hwndTree(pThis)` -> `*((HWND *)pThis + 47)`, a magic field offset that will silently change across Windows builds. If it drifts, you pass a wrong/garbage HWND to `TreeView_GetImageList` (best case it returns NULL and the feature quietly stops; worst case `TVM_GETIMAGELIST` hits an unrelated same-thread window and you `QueryInterface` a bogus handle). The notification already carries the source window: for `TVN_GETDISPINFO` the tree control fills `NMHDR::hwndFrom` with its own handle, so you can drop the offset entirely:

```cpp
void CNscTree__OnGetDisplayInfo_hook(class CNscTree *pThis, LPNMTVDISPINFOEXW pnm)
{
    CNscTree__OnGetDisplayInfo_orig(pThis, pnm);
    SetOpenFolderIcon(pnm->hdr.hwndFrom, (LPTVITEMW)&pnm->item);
}
```

You can then delete the `CNscTree__hwndTree` macro. See [add-virtual-folders-to-nav-top.wh.cpp#L1499](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/add-virtual-folders-to-nav-top.wh.cpp#L1499) for the same `hdr->hwndFrom == hTree` idiom on the nav tree.

<p>

Minor polish -- none of this affects users, so it's your call.

- **`TVM_SETITEMW` path mutates the caller's `TVITEM`.** In `CNscTree_s_SubClassTreeWndProc_hook` you pass `(LPTVITEMW)lParam` straight to `SetOpenFolderIcon`, which flips `mask |= TVIF_SELECTEDIMAGE` and writes `iSelectedImage` on a struct the shell code still owns. In practice the shell builds a fresh `TVITEM` per call so this is harmless, but strictly the convention is to not mutate caller-owned parameters -- copy first and forward the copy:
  ```cpp
  case TVM_SETITEMW: {
      TVITEMW item = *(LPTVITEMW)lParam;
      SetOpenFolderIcon(hwnd, &item);
      return CNscTree_s_SubClassTreeWndProc_orig(hwnd, uMsg, wParam, (LPARAM)&item, uIdSubclass, dwRefData);
  }
  ```
  (The `_OnGetDisplayInfo` path is *not* affected -- there `&pnm->item` is the callback fill-in buffer you're meant to write.)

- **`-lshell32` and `#include <initguid.h>` look unused.** The mod only calls `HIMAGELIST_QueryInterface` (comctl32), the `TreeView_*` macros, and `LoadLibraryExW`; `IID_PPV_ARGS(&pil2)` resolves the IID via `__uuidof`, so no `DEFINE_GUID` instantiation (initguid) is needed. Nothing appears to touch shell32. Both are harmless but can likely be dropped. (Low confidence -- please double-check against your build.)

</p>
</details>

<p>

- **Hardcoded system-image-list indices (`3`, `4`, `243`).** The closed->open folder mapping (`3`->`4`) relies on well-known shell icon indices and is stable, but `243` for the Downloads folder is build-specific and will likely shift on other/future Windows builds -- when it does, `replace_downloads` silently either misses the Downloads icon or replaces an unrelated one. There's no clean dynamic alternative, so this is just a heads-up on maintenance cost, not a change to make.

- **Toggling `replace_downloads` doesn't refresh already-drawn icons.** `Wh_ModSettingsChanged` updates `g_fReplaceDownloads`, but existing title-bar / tree icons only pick up the change on the next redraw (navigating, refreshing the view). Expected given the hook-on-demand design; worth a mention only if users report the setting "not taking effect."

</p>
</details>

aubymori:
@m417z Applied a few of the recommended changes. As for the system image list indices, Downloads seems to use the same 243 index both before and after applying a registry hack that makes it use the default folder icon, so I'm guessing there's a special index for it somehow. As well, trying to get the downloads icon index dynamically using `Shell_GetCachedImageIndexW` doesn't return the same index, so that's not viable either.

PR #4696, opened 2026-07-06
Update Island Media Controls to 0.9.146

usho-lear:
## Summary

Updates Island Media Controls to version 0.9.146. This revision expands the original Acrylic and lifecycle work with a GPU-based Liquid Glass material, refined media transitions, recording-friendly static previews, improved browser media handling, and the latest maintainer-requested threading, DPI, migration, and performance fixes.

### Materials and defaults

- Changed the default material for new installs from Mica-like to Acrylic.
- Added a one-time migration for users who still had the old Mica-like default.
- Restored Mica-like as a normal explicit choice after migration.
- Fixed the light-mode Mica-like expanded background becoming almost pure white from a duplicated tint layer.
- Kept compact and expanded material colors consistent across light and dark themes.

### Liquid Glass and Acrylic rendering

- Added the Liquid Glass material option.
- Added GPU displacement-map refraction with a rounded-rectangle SDF so edge and corner refraction follows the component shape.
- Added a narrow high-index rim and a wider, smooth interior falloff into the center.
- Increased sampling quality and removed the earlier tiled/nine-slice appearance.
- Added theme-aware edge highlights, reduced dark-mode tint, and independently tuned blur for the main surface and playback card.
- Applied matching material treatment to the expanded playback controls.
- Retained the live WGC/D3D/D2D Acrylic backdrop and fallback blur handoff.

### Recording support

- Added an optional screen-capture setting for the expanded popup.
- Added a capturable static blurred backdrop for Acrylic and Liquid Glass recording.
- Avoided recursive self-capture while keeping the blur layer visible to recorders.
- Added dark/light Mica-like, Acrylic, and Liquid Glass screenshots to the mod details page.

### Media transitions and presentation

- Added smoother directional title and artist transitions.
- Added edge-aware fading that appears as text moves beneath the artwork or popup boundary.
- Reversed expanded transition direction when navigating to the previous track.
- Removed transition-end flicker, persistent fade overlays, black blocks, and abrupt clipping.
- Improved compact and expanded text clipping consistency.

### Media and browser session handling

- Improved browser/live-media artwork identity tracking.
- Prevented stale thumbnails from leaking between browser sessions.
- Added bounded waits around media properties, thumbnail reads, transport commands, and manager requests.
- Kept event-driven GSMTC refreshes with a low-frequency provider compatibility fallback.
- Preserved Apple Music seeking and transport-control compatibility.

### Lifecycle and unload safety

- Released GSMTC manager references before uninitializing the WinRT apartment.
- Explicitly released WGC D3D/DXGI/D2D and WinRT capture resources before DLL unload.
- Added lifecycle guards around injection, callbacks, settings changes, and teardown.
- Avoided indefinite unload joins and lost media-thread wakeups.
- Improved recompilation and reload behavior so Explorer no longer needs to be restarted after each build.

### Threading, performance, and DPI review fixes

- Removed the dead WGC availability macro and unreachable stub implementation.
- Moved overlay window presentation to its owning UI thread through an asynchronous latest-frame queue.
- Kept cross-thread WGC rendering under its resource mutex without holding that lock across `UpdateLayeredWindow`, `ShowWindow`, or capture-affinity calls.
- Replaced the permanent per-frame taskbar layout monitor with a 600 ms Win32 timer.
- Created the popup on the taskbar monitor before initializing XAML and added `WM_DPICHANGED` handling for mixed-DPI multi-monitor setups.
- Switched the D2D factory to multi-threaded mode.
- Allocated the readback buffer before mapping the staging texture and explicitly closed consumed WGC frames.
- Removed the duplicate Windows.Foundation include and the dead popup controls scale state.

### Details page

- Added a top-level What's new section.
- Added a three-material dark/light comparison table.
- Preserved the original overview, animated previews, and feature list.

## Validation

- `D:\Windhawk\Compiler\bin\clang++.exe @D:\Windhawk\Compiler\compile_flags.txt -ID:\Windhawk\Compiler\include -fsyntax-only mods\island-media-controls.wh.cpp`
- `git diff --check -- mods/island-media-controls.wh.cpp`
- Unicode format-control character scan: 0 matches.
- All six public material-preview URLs return `200 image/png` with the expected content lengths.

### DIFF
--- a/mods/island-media-controls.wh.cpp
+++ b/mods/island-media-controls.wh.cpp
@@ -2,7 +2,7 @@
 // @id              island-media-controls
 // @name            Island Media Controls
 // @description     Dynamic island-like media controls for the Windows 11 taskbar.
-// @version         0.9.27
+// @version         0.9.39
 // @author          usho
 // @github          https://github.com/usho-lear
 // @license         MIT
@@ -63,6 +63,8 @@ it fits naturally alongside your taskbar and system tray items.
     - "taskbar_right_edge": "Taskbar - right overlay"
   - CompactWidth: 168
     $name: Compact width
+  - AutoSizeToTaskbar: true
+    $name: Auto size to taskbar
   - ExpandedWidth: 360
     $name: Expanded overlay width
   - ExpandedHeight: 500
@@ -202,6 +204,7 @@ namespace {
 struct Settings {
     std::wstring position = L"tray_left";
     int compactWidth = 168;
+    bool autoSizeToTaskbar = true;
     int expandedWidth = 360;
     int expandedHeight = 430;
     int popupSpacing = 8;
@@ -220,6 +223,25 @@ struct Settings {
     std::wstring material = L"mica_like";
 };
 
+struct RuntimeLayout {
+    double taskbarHeightDip = 40.0;
+    double compactHeight = 40.0;
+    double compactWidth = 168.0;
+    double artSize = 28.0;
+    double artImageSize = 30.0;
+    double artCornerRadius = 8.0;
+    double cornerRadius = 20.0;
+    double contentMarginX = 10.0;
+    double contentMarginY = 4.0;
+    double artColumnWidth = 30.0;
+    double textMarginX = 8.0;
+    double titleFontSize = 12.0;
+    double artistFontSize = 10.0;
+    double progressWidth = 74.0;
+    double progressHeight = 4.0;
+    double progressMarginLeft = 8.0;
+};
+
 struct MediaState {
     std::wstring title = L"Not Playing";
     std::wstring artist;
@@ -229,6 +251,7 @@ struct MediaState {
     int64_t timelineStartTicks = 0;
     int64_t positionTicks = 0;
     int64_t durationTicks = 0;
+    std::wstring sourceAppUserModelId;
     std::vector<uint8_t> thumbnailBytes;
 };
 
@@ -241,6 +264,7 @@ using MediaCommand =
     std::function<void(gsm::GlobalSystemMediaTransportControlsSession const&)>;
 
 Settings g_settings;
+RuntimeLayout g_layout;
 std::mutex g_pendingSettingsMutex;
 std::optional<Settings> g_pendingSettings;
 std::mutex g_mediaMutex;
@@ -337,7 +361,7 @@ mediax::RectangleGeometry g_popupXamlProgressGlowClip = nullptr;
 std::vector<Border> g_popupXamlProgressGlowLayers;
 std::vector<Border> g_popupXamlProgressCoreBlurLayers;
 Border g_popupXamlProgressGlowCore = nullptr;
-StackPanel g_popupXamlControls = nullptr;
+controls::Canvas g_popupXamlControls = nullptr;
 ScaleTransform g_popupXamlControlsScale = nullptr;
 uint64_t g_popupXamlThumbnailHash = UINT64_MAX;
 bool g_popupXamlBackdropCoverEnabled = false;
@@ -401,6 +425,11 @@ winrt::event_token g_compactTextRenderingToken{};
 bool g_compactTextRenderingHooked = false;
 double g_compactTextProgress = 1.0;
 std::chrono::steady_clock::time_point g_lastCompactTextFrameTime{};
+winrt::event_token g_taskbarLayoutRenderingToken{};
+bool g_taskbarLayoutRenderingHooked = false;
+std::chrono::steady_clock::time_point g_lastTaskbarLayoutCheckTime{};
+FrameworkElement g_taskbarLayoutWatchRoot = nullptr;
+FrameworkElement g_taskbarLayoutWatchTarget = nullptr;
 bool g_compactTextInitialized = false;
 std::wstring g_compactLastTitle;
 std::wstring g_compactLastArtist;
@@ -542,7 +571,6 @@ bool TrySeekAppleMusicWithUiAutomation(double ratio) {
 SolidColorBrush Brush(winrt::Windows::UI::Color color) {
     return SolidColorBrush(color);
 }
-
 
 void AttachGpuFriendlyTransform(FrameworkElement const& element,
                                 ScaleTransform& scale,
@@ -712,6 +740,7 @@ Settings ReadSettings() {
     Settings settings;
     settings.position = GetStringSetting(L"Main.Position", L"tray_left");
     settings.compactWidth = Clamp(Wh_GetIntSetting(L"Main.CompactWidth"), 96, 320);
+    settings.autoSizeToTaskbar = Wh_GetIntSetting(L"Main.AutoSizeToTaskbar") != 0;
     settings.expandedWidth = Clamp(Wh_GetIntSetting(L"Main.ExpandedWidth"), 240, 640);
     settings.expandedHeight = Clamp(Wh_GetIntSetting(L"Main.ExpandedHeight"), 430, 760);
     settings.popupSpacing = Clamp(Wh_GetIntSetting(L"Main.PopupSpacing"), 2, 24);
@@ -974,12 +1003,38 @@ uint64_t MediaThumbFingerprint(std::vector<uint8_t> const& bytes) {
     return hash;
 }
 
+std::wstring MediaIdentityKey(MediaState const& state) {
+    if (!state.hasSession) {
+        return L"";
+    }
+
+    // Deliberately excludes artwork bytes. Browsers can emit metadata before a
+    // live-room/video thumbnail is available, and some providers temporarily
+    // repeat the previous thumbnail. This key represents the actual media item
+    // for deciding whether old artwork can be reused safely.
+    return state.sourceAppUserModelId + L"\n" +
+           state.title + L"\n" +
+           state.artist + L"\n" +
+           std::to_wstring(state.durationTicks);
+}
+
+bool LooksLikeBrowserMediaSource(std::wstring const& source) {
+    return source.find(L"Chrome") != std::wstring::npos ||
+           source.find(L"MSEdge") != std::wstring::npos ||
+           source.find(L"MicrosoftEdge") != std::wstring::npos ||
+           source.find(L"Firefox") != std::wstring::npos ||
+           source.find(L"Brave") != std::wstring::npos ||
+           source.find(L"Opera") != std::wstring::npos ||
+           source.find(L"Vivaldi") != std::wstring::npos ||
+           source.find(L"Arc") != std::wstring::npos ||
+           source.find(L"Browser") != std::wstring::npos;
+}
+
 std::wstring MediaContentKey(MediaState const& state) {
     if (!state.hasSession) {
         return L"";
     }
-    return state.title + L"\n" + state.artist + L"\n" +
-           std::to_wstring(state.durationTicks) + L"\n" +
+    return MediaIdentityKey(state) + L"\n" +
            std::to_wstring(MediaThumbFingerprint(state.thumbnailBytes));
 }
 
@@ -1135,6 +1190,11 @@ void RefreshMediaState(
         auto session = CurrentSessionFromManager(manager);
         if (session) {
             state.hasSession = true;
+            try {
+                state.sourceAppUserModelId = std::wstring(session.SourceAppUserModelId());
+            } catch (...) {
+                state.sourceAppUserModelId.clear();
+            }
             auto playback = session.GetPlaybackInfo();
             state.isPlaying =
                 playback.PlaybackStatus() ==
@@ -1314,7 +1374,6 @@ HBITMAP DecodeAlbumBitmap(std::vector<uint8_t> const& bytes, UINT size) {
     if (stream) stream->Release();
     return bitmap;
 }
-
 
 std::vector<uint8_t> CreateLowDetailAlbumCoverBytes(std::vector<uint8_t> const& bytes, bool bottomFadeToMiddle = false) {
     std::vector<uint8_t> output;
@@ -1507,7 +1566,6 @@ winrt::Windows::UI::Color DefaultPopupAccentColor() {
     return Color(0xFF, 0x4F, 0x7D, 0xE8);
 }
 
-
 std::vector<uint8_t> EncodePbgraPngBytes(UINT width, UINT height, std::vector<BYTE> const& pixels) {
     std::vector<uint8_t> output;
     if (width == 0 || height == 0 || pixels.size() < static_cast<size_t>(width) * height * 4) {
@@ -1681,8 +1739,6 @@ bool ShouldUseAbstractArtworkForDisplay(std::vector<uint8_t> const& bytes) {
     // covers stay untouched.
     return videoLikeAspect || lowResolution;
 }
-
-
 
 winrt::Windows::UI::Color BoostArtworkColor(winrt::Windows::UI::Color color,
                                             double saturationScale,
@@ -2296,7 +2352,6 @@ std::vector<uint8_t> CreateDisplayAlbumCoverBytes(std::vector<uint8_t> const& by
     return abstractBytes.empty() ? bytes : abstractBytes;
 }
 
-
 winrt::Windows::UI::Color ExtractAlbumAccentColor(std::vector<uint8_t> const& bytes) {
     if (bytes.empty()) {
         return DefaultPopupAccentColor();
@@ -2848,7 +2903,6 @@ int PopupFinalHeightFromArtSize(int artSize) {
            PopupSurfaceGap() * 2 + kPopupHostShadowMargin * 2;
 }
 
-
 mediax::Brush PopupControlCardBrush() {
     bool dark = IsDarkModeApprox();
     // Keep light and dark tint strengths independent. Light mode keeps the
@@ -2963,23 +3017,14 @@ RECT UnionPopupRects(RECT const& a, RECT const& b) {
 }
 
 RECT CurrentPopupSurfaceScreenRect() {
-    RECT targetArt{};
-    RECT targetCard{};
-    RECT targetTitle{};
-    RECT targetArtist{};
-    RECT targetProgress{};
-    RECT targetElapsed{};
-    RECT targetDuration{};
-    RECT targetControls{};
-    CalculatePopupFinalLayout(g_popupFinalRect, targetArt, targetCard,
-                              targetTitle, targetArtist, targetProgress,
-                              targetElapsed, targetDuration, targetControls);
-    RECT targetBackdrop = PopupBackdropRectFromParts(targetArt, targetCard);
+    // Used only for outside-click hit testing. Keep this in physical screen
+    // pixels and use the full animated host bounds; the precise XAML surface is
+    // calculated in DIP coordinates in UpdatePopupXamlVisuals().
     double progress = PopupProgress();
-    return {LerpInt(g_popupSourceRect.left, targetBackdrop.left, progress),
-            LerpInt(g_popupSourceRect.top, targetBackdrop.top, progress),
-            LerpInt(g_popupSourceRect.right, targetBackdrop.right, progress),
-            LerpInt(g_popupSourceRect.bottom, targetBackdrop.bottom, progress)};
+    return {LerpInt(g_popupSourceRect.left, g_popupFinalRect.left, progress),
+            LerpInt(g_popupSourceRect.top, g_popupFinalRect.top, progress),
+            LerpInt(g_popupSourceRect.right, g_popupFinalRect.right, progress),
+            LerpInt(g_popupSourceRect.bottom, g_popupFinalRect.bottom, progress)};
 }
 
 bool PopupShouldExpandRight(MONITORINFO const& monitorInfo) {
@@ -3232,7 +3277,6 @@ void PaintExpandedPopup(HWND hwnd, HDC dc) {
     DeleteObject(titleFont);
 }
 
-
 bool PopupButtonStyleIs(std::wstring_view style) {
     return std::wstring_view(g_settings.popupButtonStyle) == style;
 }
@@ -3258,13 +3302,35 @@ winrt::hstring PopupTransportGlyph(Border const& surface, bool playing) {
 }
 
 void UpdatePopupTransportButtonGlyph(Border const& surface, bool playing) {
-    if (auto icon = surface.Child().try_as<TextBlock>()) {
-        icon.Text(PopupTransportGlyph(surface, playing));
+    if (auto icon = surface.Child().try_as<controls::FontIcon>()) {
+        icon.Glyph(PopupTransportGlyph(surface, playing));
+    } else if (auto text = surface.Child().try_as<TextBlock>()) {
+        text.Text(PopupTransportGlyph(surface, playing));
     }
 }
 
 double PopupControlsWidth() {
     return 126.0;
+}
+
+double PopupControlsHeight() {
+    return 38.0;
+}
+
+double PopupButtonWidth() {
+    return 38.0;
+}
+
+double PopupButtonHeight() {
+    return 38.0;
+}
+
+double PopupButtonTop() {
+    return (PopupControlsHeight() - PopupButtonHeight()) * 0.5;
+}
+
+double PopupButtonLeft(int index) {
+    return 2.0 + 42.0 * static_cast<double>(index);
 }
 
 mediax::Brush PopupButtonHoverBrush(bool dark) {
@@ -3278,49 +3344,70 @@ void ApplyPopupButtonVisual(Border const& surface, bool hovered = false) {
     auto primaryText = dark ? Color(0xFF, 0xFF, 0xFF, 0xFF)
                             : Color(0xFF, 0x00, 0x00, 0x00);
 
-    double width = 36.0;
-    double height = 34.0;
-    double corner = 17.0;
-    double fontSize = primary ? 22.0 : 21.0;
+    double width = PopupButtonWidth();
+    double height = PopupButtonHeight();
+    double corner = height * 0.5;
+    double fontSize = primary ? 21.0 : 20.0;
     mediax::Brush background = hovered ? PopupButtonHoverBrush(dark)
                                        : Brush(Color(0x00, 0x00, 0x00, 0x00));
     mediax::Brush foreground = Brush(primaryText);
 
     if (PopupButtonStyleIs(L"fluent_bold")) {
-        fontSize = primary ? 26.0 : 24.0;
+        fontSize = primary ? 23.0 : 22.0;
     }
 
     surface.Width(width);
     surface.Height(height);
+    surface.Visibility(Visibility::Visible);
     surface.CornerRadius({corner, corner, corner, corner});
     surface.Background(background);
     surface.BorderThickness({0, 0, 0, 0});
     surface.BorderBrush(Brush(Color(0x00, 0x00, 0x00, 0x00)));
 
-    if (auto icon = surface.Child().try_as<TextBlock>()) {
-        bool playing = SnapshotMedia().isPlaying;
-        icon.Text(PopupTransportGlyph(surface, playing));
+    bool playing = SnapshotMedia().isPlaying;
+    if (auto icon = surface.Child().try_as<controls::FontIcon>()) {
+        icon.Visibility(Visibility::Visible);
+        icon.Opacity(1.0);
+        icon.Width(width);
+        icon.Height(height);
+        icon.Margin({0, 0, 0, 0});
+        icon.HorizontalAlignment(HorizontalAlignment::Center);
+        icon.VerticalAlignment(VerticalAlignment::Center);
+        icon.Glyph(PopupTransportGlyph(surface, playing));
         icon.Foreground(foreground);
         icon.FontFamily(mediax::FontFamily(L"Segoe Fluent Icons"));
         icon.FontSize(fontSize);
-        icon.FontWeight(winrt::Windows::UI::Text::FontWeights::Bold());
+        icon.FontWeight(winrt::Windows::UI::Text::FontWeights::SemiBold());
+    } else if (auto text = surface.Child().try_as<TextBlock>()) {
+        text.Visibility(Visibility::Visible);
+        text.Opacity(1.0);
+        text.Width(width);
+        text.Height(height);
+        text.Margin({0, 0, 0, 0});
+        text.HorizontalAlignment(HorizontalAlignment::Stretch);
+        text.VerticalAlignment(VerticalAlignment::Center);
+        text.TextAlignment(xaml::TextAlignment::Center);
+        text.Text(PopupTransportGlyph(surface, playing));
+        text.Foreground(foreground);
+        text.FontFamily(mediax::FontFamily(L"Segoe Fluent Icons"));
+        text.FontSize(fontSize);
+        text.FontWeight(winrt::Windows::UI::Text::FontWeights::SemiBold());
     }
 }
 
 Border MakePopupXamlButton(const wchar_t* name, void (*onClick)(), bool primary = false) {
     Border surface;
     surface.Name(name);
-    surface.Margin({3, 0, 3, 0});
+    surface.Margin({0, 0, 0, 0});
     surface.Padding({0, 0, 0, 0});
     surface.Tag(winrt::box_value(winrt::hstring(primary ? L"primary" : L"secondary")));
     surface.IsHitTestVisible(true);
 
-    TextBlock icon;
+    controls::FontIcon icon;
     icon.FontFamily(mediax::FontFamily(L"Segoe Fluent Icons"));
-    icon.Text(L"\uE768");
+    icon.Glyph(L"\uE768");
     icon.HorizontalAlignment(HorizontalAlignment::Center);
     icon.VerticalAlignment(VerticalAlignment::Center);
-    icon.TextAlignment(xaml::TextAlignment::Center);
     surface.Child(icon);
     ApplyPopupButtonVisual(surface, false);
 
@@ -3469,7 +3556,6 @@ void ApplyPopupXamlTheme(bool force = false) {
     }
     ApplyPopupDynamicAccentVisuals();
 }
-
 
 void ApplyPopupDynamicAccentVisuals() {
     auto accent = PopupAccentColor();
@@ -3638,7 +3724,6 @@ void TickPopupMediaTransition(double dt) {
         ApplyPopupMediaTransitionVisuals();
     }
 }
-
 
 void InitializePopupCompositionShadow() {
     // UWP XAML Border in Windhawk does not expose GetAlphaMask(), so an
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

---

**1. `winrt::uninit_apartment()` releases the session manager after the apartment is already torn down.** In `MediaThreadProc`, the teardown is:

```cpp
    clearSessionEvents();
    try {
        if (manager) {
            manager.SessionsChanged(sessionsChangedToken);
            manager.CurrentSessionChanged(currentSessionChangedToken);
        }
    } catch (...) {
    }
    winrt::uninit_apartment();
}   // <-- `manager` is destructed *here*, after CoUninitialize
```

`manager` still holds a COM reference when `winrt::uninit_apartment()` (i.e. `CoUninitialize`) runs; its `Release()` then happens during stack unwind, *after* the apartment has been uninitialized. Releasing a WinRT/COM object after its apartment is gone is undefined — C++/WinRT specifically requires all WinRT objects to be destroyed before `uninit_apartment()`, and for a system-served object like the GSMTC manager this can crash or hang the media thread on unload (which, since this is `explorer.exe`, means the shell). The `winrt::uninit_apartment()` call is new in this PR, so this is a freshly introduced teardown hazard. Fix is one line — drop the reference first:

```cpp
    clearSessionEvents();
    try {
        if (manager) {
            manager.SessionsChanged(sessionsChangedToken);
            manager.CurrentSessionChanged(currentSessionChangedToken);
        }
    } catch (...) {
    }
    manager = nullptr;            // release all WinRT objects before uninit
    winrt::uninit_apartment();
```

(`sessionEventSubscriptions` is already emptied by `clearSessionEvents()`, so `manager` is the only outstanding WinRT object.)

<p>

- **Dead `g_popupXamlControlsScale`.** Now that the transport controls are a `Canvas` with fixed child slots instead of a `StackPanel` + `ScaleTransform`, `g_popupXamlControlsScale` is only ever assigned `nullptr` (lines 365, 3814, 4210, 5935). The global and its `if (g_popupXamlControlsScale) { ... }` branch in `UpdatePopupXamlVisuals()` are unreachable — safe to remove both.

</p>
</details>

<p>

- **`LooksLikeBrowserMediaSource` is a broad substring match.** It keys on tokens like `L"Opera"`, `L"Arc"`, and `L"Browser"` inside the source AppUserModelId, which can both false-positive (a non-browser app whose AUMID happens to contain one of these) and false-negative (a Chromium browser packaged under a custom AUMID). The impact is limited — it only gates the short "hide suspicious repeated thumbnail" grace window for browser live media — so this is just an FYI that the heuristic is inherently fuzzy, not a correctness problem.

</p>
</details>

usho-lear:
### Summary

This updates `island-media-controls` to version `0.9.87`.

The update focuses on Fluent Acrylic visuals, safer update behavior for existing users, startup/unload stability, browser/live-media artwork correctness, themed-taskbar compatibility, high-DPI popup fixes, and smoother media/session handling.

### Changelog

#### Material defaults and migration

- Changed the default background material from `mica_like` to `acrylic`.
- Added a one-time settings migration so existing users who still have the old default `mica_like` material are moved to Acrylic behavior after updating.
- Preserved user choice: if a user explicitly changes the material to `solid` or `acrylic`, the migration override stops applying.
- Bumped the mod version to `0.9.87` so existing installations receive the update.

#### Acrylic popup backdrop

- Added a native Acrylic/glass popup backdrop path backed by a separate overlay window.
- Added WGC/D3D/D2D resources for live blurred backdrop rendering when available.
- Added fallback blurred backdrop rendering while WGC is not ready.
- Added handoff controls for the fallback-to-live backdrop transition.
- Kept the compact island visually stable with a tinted brush instead of relying on XAML Acrylic fallback behavior.
- Reduced noisy WGC diagnostics during normal unload/startup transitions.

#### Startup, unload, and lifecycle safety

- Added an explicit active lifecycle flag so startup, taskbar rebuild, settings changes, media callbacks, and unload paths do not race each other.
- Guarded taskbar-thread injection and settings application so stale callbacks cannot re-inject UI while the mod is unloading.
- Made unload idempotent with an atomic unload transition.
- Stopped media refreshes and queued media commands when the mod is no longer active.
- Added timeout-based GSMTC manager requests so startup or shutdown is less likely to get stuck waiting on Windows media APIs.
- Added safer media thread startup failure handling.
- Continued stopping render/layout hooks during island removal and unload.

#### Media/session handling

- Switched the media worker toward GSMTC event-triggered refreshes:
  - `SessionsChanged`
  - `CurrentSessionChanged`
  - `MediaPropertiesChanged`
  - `PlaybackInfoChanged`
  - `TimelinePropertiesChanged`
- Kept a low-frequency fallback refresh for provider compatibility.
- Added safer media command handling around popup seek interactions.
- Refreshed media/session state before issuing playback commands to reduce stale-session control failures.
- Released GSMTC objects before uninitializing the media thread WinRT apartment.

#### Apple Music and seeking

- Improved seek handling for providers that report timeline offsets differently.
- Added Apple Music UI Automation fallback seek support.
- Restored compatibility for Apple Music progress seeking.
- Fixed compact and popup progress seek pointer handling.
- Smoothed compact playback progress updates.

#### Browser/live artwork correctness

- Added source-app/media identity tracking for artwork.
- Prevented stale thumbnails from being reused across different browser live sessions.
- Limited same-media empty-thumbnail reuse to a short grace period.
- Added a guard for repeated browser live thumbnails so suspicious old artwork is temporarily hidden instead of immediately reused.
- Kept the default low-resolution video artwork mode as browser thumbnail/original artwork rather than generated abstract artwork.
- Added G2-shaped popup cover generation for placeholder and real artwork.

#### Compact taskbar island layout

- Added `AutoSizeToTaskbar`, enabled by default.
- Introduced runtime layout calculation for compact island size, album art size, corner radius, text sizes, margins, and compact progress dimensions.
- Improved visual taskbar-height detection by considering the injection target, root height, `SystemTrayFrameGrid`, `TaskbarFrame`, positive top/bottom margins used by Taskbar Styler themes, and Win32 taskbar window fallback.
- Prevented auto sizing from enlarging the compact island on visually smaller themed taskbars.
- Added live taskbar layout monitoring so switching Taskbar Styler themes can update the island without reloading the mod.
- Improved compact text clipping after settings/theme changes by replacing a fixed-width fallback with geometry-derived fallback sizing.

#### Expanded popup

- Fixed high-DPI scaling issues when Windows display scaling is above 100%.
- The popup host HWND now uses physical pixels, while XAML layout calculations are converted back to popup-local DIP coordinates.
- Fixed popup fallback source geometry to account for DPI scaling.
- Stabilized popup transport controls:
  - Replaced `StackPanel + RenderTransform` with fixed local `Canvas` placement.
  - Switched popup transport glyphs from `TextBlock` to `FontIcon`.
  - Removed baseline/line-height compensation that caused vertical drift.
  - Reduced hover-highlight/glyph mismatch across high DPI scales.
- Improved popup shadow, cover, panel, and progress rendering consistency.

#### Popup/window class and code quality

- Avoided reusing a stale popup window class.
- Registered/unregistered popup classes with the current mod module instance.
- Centralized repeated "apply pending settings and inject island" logic.
- Added clearer lifecycle guards around UI injection, media refresh, and settings changes.
- Kept unrelated metadata unchanged except for the version bump.

### Notes

- The experimental `translucent_taskbar` and `window_glass` material options were tested but intentionally not included because the visual result was not good enough yet.
- Acrylic is now the default material for new users, and old-default users are migrated without overriding later explicit user choices.

### Test plan

- Build in Windhawk without compiler errors.
- Install/update the mod and verify the compact island appears.
- Verify Acrylic is the default material for fresh settings.
- Verify a user with the old `mica_like` default is migrated to Acrylic behavior after updating.
- Verify users can still manually choose `mica_like`, `solid`, or `acrylic` after migration.
- Verify `Auto size to taskbar` is enabled by default for fresh settings.
- Verify compact island layout adapts to smaller and larger taskbar visual heights.
- Switch Taskbar Styler themes and verify compact island dimensions update without reloading the mod.
- Change Windows display scaling to 125%, 150%, and 200%:
  - expanded popup should no longer be clipped
  - popup controls should remain visible
  - play/pause glyph should align with hover/highlight region
- Verify compact title/artist text is not clipped after changing mod settings.
- Verify browser/live media without artwork does not reuse stale artwork from a previous session.
- Verify regular media artwork still displays correctly.
- Verify seek/transport controls still work in the expanded popup.
- Reload/unload the mod and verify Windhawk does not remain stuck in `uninitializing`.

mods/island-media-controls.wh.cpp
```diff
@@ -121,6 +141,59 @@ it fits naturally alongside your taskbar and system tray items.
 #include <shlwapi.h>
 #include <wincodec.h>
 #include <windows.ui.xaml.hosting.desktopwindowxamlsource.h>
+
+#if __has_include(<winrt/Windows.Graphics.Capture.h>) && \
+    __has_include(<winrt/Windows.Graphics.DirectX.Direct3D11.h>)
+#define IMC_WGC_OVERLAY_AVAILABLE 1
```
m417z: Why is this condition used? It might mean that the mod will behave differently in different Windhawk versions. Remove it.

m417z:
### Submission review

---

**1. The WGC D3D/DXGI/D2D device objects are never released during teardown — only at `DLL_PROCESS_DETACH`.** The D3D11/DXGI/D2D device `com_ptr`s and the WinRT capture globals (declarations at lines 473–486) are reset to `nullptr` in exactly one place: `EnsurePopupOverlayWgcDeviceResources` when it *recreates* them (lines 6721–6739). Nothing in the unload path touches them — `StopPopupOverlayWgcBackdrop` (7160) closes only the session/frame-pool, and `DestroyPopupBackdropOverlayWindow` (7909) → `Wh_ModUninit` leave the whole device set alive. So once the acrylic popup has been opened even once (and acrylic is now the default), these objects get released by their global destructors when the mod DLL is unloaded — i.e. during `DLL_PROCESS_DETACH`, under the loader lock. Releasing D3D11/DXGI/D2D COM objects from that context is a well-known deadlock/crash risk, and it happens on **every** disable/update/reload after the first popup. Add a small `ReleasePopupOverlayWgcDeviceResources()` that reuses the reset block at 6721–6739 and call it from the teardown that `Wh_ModUninit` drives (alongside `StopPopupOverlayWgcBackdrop`), so the device set is gone before the DLL unloads.

**2. Unbounded `.get()` on the media thread can make unload hang ("stuck `uninitializing`").** `StopMediaThread` unconditionally `join()`s the worker (line 8814), but the worker routinely blocks in WinRT `.get()` calls with no timeout: `session.TryGetMediaPropertiesAsync().get()` (1449), the thumbnail `OpenReadAsync()`/`ReadAsync().get()` (891/900), and every queued transport command — `TryChangePlaybackPositionAsync`/`TrySkipPreviousAsync`/`TryTogglePlayPauseAsync`/`TrySkipNextAsync().get()` (1513/1518, 6308–6316, 9269–9283, 4603–4615). Flipping `g_mediaThreadRunning = false` does not interrupt an in-flight `.get()`, so if a GSMTC provider (Apple Music, a wedged browser tab) is slow or hung when the user disables the mod, the `join()` blocks the uninit thread — exactly the "remain stuck in `uninitializing`" symptom your test plan is trying to rule out. You already solved this for the manager request with `RequestMediaManagerWithTimeout` (1340); apply the same bounded `wait_for` to the per-session property/thumbnail/command `.get()`s, or bound the `join()` and `detach()` on timeout.

<p>

- **Lost-wakeup on the media command CV.** The wait predicate (8767–8771) reads `g_mediaThreadRunning` and `g_mediaRefreshRequested`, but `StopMediaThread` (8808–8810) and `RequestMediaRefresh` (1332–1336) both flip those atomics and `notify` *without* holding `g_mediaCommandMutex`. If the flip+notify lands between the worker evaluating the predicate and actually blocking, the notification is dropped and the worker sleeps until the 30 s fallback fires — so an unload or a settings-driven refresh can stall for up to ~30 s. The command-enqueue path (1478–1482) already does this correctly; mutate the two flags under `g_mediaCommandMutex` before notifying the same way.
- **D2D factory threading model.** `D2D1CreateFactory(D2D1_FACTORY_TYPE_SINGLE_THREADED, …)` (6782) is created on the UI thread, but every D2D draw runs on the free-threaded WGC callback thread. It's safe today only because `g_popupOverlayWgcMutex` serializes access; Microsoft's WGC samples pair a free-threaded frame pool with a `MULTI_THREADED` factory, which is more robust if the locking ever changes.
- **`Direct3D11CaptureFrame` not explicitly closed.** The frame from `TryGetNextFrame()` (7338) is released only by its local destructor. It works with the 2-buffer pool, but `frame.Close()` is the documented way to return the surface to the pool promptly.
- **`g_taskbarWnd` data race.** It's a plain `HWND`, read on the media thread (8743) and written on the UI thread. Benign on x64, but `std::atomic<HWND>` would make it clean.
- **Duplicate include.** `#include <winrt/Windows.Foundation.h>` appears at both line 204 and line 209.

</p>
</details>

<p>

- **Live-blur backdrop rebuilds GPU objects every frame.** `RenderPopupOverlayWgcFrame` runs on every `FrameArrived` (≈monitor refresh rate for as long as the popup is open) and recreates a `CLSID_D2D1GaussianBlur` effect (6948), the intermediate/target D2D bitmaps (6884/6894/6914), and a full readback buffer each frame. The blur effect and size-dependent bitmaps only change when the popup resizes, so caching them across frames would cut a lot of per-frame CPU/GPU churn. Related: the `FrameArrived` handler holds `g_popupOverlayWgcMutex` (locked at 7334) across the synchronous GPU readback (`CopyResource` + `Map`), the per-pixel loop, and `UpdateLayeredWindow`, while the popup UI timer contends for the same mutex on its fast path — that can add taskbar jank during the open/morph animation.
- **`mica_like` material migration.** `ApplySettingsMigrations` (923) sets a persistent flag when `material == "mica_like"` at migration time and thereafter rewrites any `mica_like` read to `acrylic`. Two consequences worth checking against the stated intent: (a) it can't tell a user who *deliberately* chose `mica_like` from one who merely still had the old default, so it silently overrides an explicit choice; and (b) for a migrated user the `mica_like` dropdown option is a no-op — selecting it again keeps mapping to `acrylic`, and the only way to make `mica_like` work again is to first pick `solid`/`acrylic` (which clears the flag) and then switch back. If `mica_like` is meant to stay selectable, a one-shot rewrite of the stored value would be cleaner than a permanent read-time override.
- **Taskbar-layout monitor is a persistent per-frame callback.** `StartTaskbarLayoutMonitor` (9387) hooks `CompositionTarget::Rendering` for the island's whole lifetime and early-returns on a 600 ms throttle (9345–9349). It works, but it runs ~60×/s just to poll for theme/size changes; subscribing to the watched element's `SizeChanged` would be event-driven. Minor.

</p>
</details>

usho-lear:
Updated this PR branch to `0.9.136` in commit `3068fd4`.

Review feedback addressed:
- `MediaThreadProc` now releases the GSMTC `manager` before `winrt::uninit_apartment()`.
- Removed the dead `g_popupXamlControlsScale` global and no-op assignments.
- Checked for Unicode format-control/bidi characters; none found.

Validation run locally:
- `git diff --check -- mods/island-media-controls.wh.cpp`
- `D:\Windhawk\Compiler\bin\clang++.exe @D:\Windhawk\Compiler\compile_flags.txt -ID:\Windhawk\Compiler\include -fsyntax-only mods\island-media-controls.wh.cpp`

The update also includes the later `liquid_glass` material/refraction work, smoother compact/expanded text transitions, and the unload/recompile stability fixes from the local/web iterations.

m417z:
### Submission review

---

This is a large, well-structured update to your own mod, so most of the review is about the new WGC/acrylic backdrop and lifecycle code. One blocking deadlock, plus a few smaller correctness/perf items.

**1. Deadlock: the WGC `FrameArrived` callback holds `g_popupOverlayWgcMutex` across cross‑thread `ShowWindow`/`UpdateLayeredWindow` on the overlay window.**
The frame pool is created with `CreateFreeThreaded`, so the callback runs on a WinRT threadpool thread, and it holds `g_popupOverlayWgcMutex` for the whole of `RenderPopupOverlayWgcFrame`:

```cpp
g_popupOverlayWgcFrameArrivedToken = g_popupOverlayWgcFramePool.FrameArrived(
    [](capture::Direct3D11CaptureFramePool const& sender, ...) {
        std::lock_guard callbackLock(g_popupOverlayWgcMutex);   // held across everything below
        if (!g_popupOverlayWgcRunning) return;
        auto frame = sender.TryGetNextFrame();
        RenderPopupOverlayWgcFrame(frame);   // -> UpdateLayeredWindow / ShowWindow / SetWindowDisplayAffinity
    });
```

`g_popupBackdropOverlay` is created on the taskbar UI thread, but `RenderPopupOverlayWgcFrame` calls `UpdateLayeredWindow(g_popupBackdropOverlay, …)` ([L8127](https://github.com/ramensoftware/windhawk-mods/blob/e260edba3ca056f9e2a8ca31ebd6048ea153fcee/mods/island-media-controls.wh.cpp#L8127)), `SetPopupWindowCaptureExclusion(g_popupBackdropOverlay, …)` and `ShowWindow(g_popupBackdropOverlay, SW_SHOWNOACTIVATE)` ([L8143‑L8144](https://github.com/ramensoftware/windhawk-mods/blob/e260edba3ca056f9e2a8ca31ebd6048ea153fcee/mods/island-media-controls.wh.cpp#L8143)) every visible frame. `ShowWindow` (and the other USER32 calls) on a window owned by another thread send messages synchronously to that thread and block until it pumps.

Meanwhile the taskbar UI thread takes the *same* mutex in `StartPopupOverlayWgcBackdrop` ([L8359](https://github.com/ramensoftware/windhawk-mods/blob/e260edba3ca056f9e2a8ca31ebd6048ea153fcee/mods/island-media-controls.wh.cpp#L8359)), `StopPopupOverlayWgcBackdrop` ([L8211](https://github.com/ramensoftware/windhawk-mods/blob/e260edba3ca056f9e2a8ca31ebd6048ea153fcee/mods/island-media-controls.wh.cpp#L8211)) and `UpdatePopupOverlayWgcCaptureRect` ([L8438](https://github.com/ramensoftware/windhawk-mods/blob/e260edba3ca056f9e2a8ca31ebd6048ea153fcee/mods/island-media-controls.wh.cpp#L8438)) — and while blocked on the mutex it does not pump messages. So the render thread holds the lock and blocks in a cross‑thread `SendMessage`, while the UI thread waits for the lock: a classic lock‑and‑`SendMessage` deadlock that hangs Explorer's taskbar. This is very likely to hit during the popup open/close morph, where `UpdatePopupOverlayWgcCaptureRect` (or `StartPopupOverlayWgcBackdrop`) runs on the UI thread on every animation frame while WGC frames are arriving at ~60 Hz.

Fix: don't hold `g_popupOverlayWgcMutex` across the window‑output phase. Under the lock do only the D2D/D3D render + `Map`/copy‑to‑local‑buffer/`Unmap`; release the lock, then do the `CreateDIBSection`/`UpdateLayeredWindow`/`ShowWindow`/`SetWindowDisplayAffinity` block — ideally marshalled to the UI thread that owns those windows (`RunFromWindowThread`/`PostMessage`), since they have thread affinity. See the lock‑then‑release‑then‑call pattern in the [wiki dev tips](https://github.com/ramensoftware/windhawk/wiki/Development-tips).

**2. Permanent per‑frame `CompositionTarget::Rendering` subscription used only as a 600 ms poll.**
`StartTaskbarLayoutMonitor` hooks `CompositionTarget::Rendering` for the entire lifetime of the injected island ([L10551‑L10555](https://github.com/ramensoftware/windhawk-mods/blob/e260edba3ca056f9e2a8ca31ebd6048ea153fcee/mods/island-media-controls.wh.cpp#L10551)), and `OnTaskbarLayoutRendering` only *throttles the work* to 600 ms ([L10499‑L10502](https://github.com/ramensoftware/windhawk-mods/blob/e260edba3ca056f9e2a8ca31ebd6048ea153fcee/mods/island-media-controls.wh.cpp#L10499)) while still waking on every composition frame. Unlike your other render loops (hover/compact‑text/compact‑progress/popup), which start and stop on demand, this one runs unconditionally and keeps the UI thread + DWM from ever idling (~60–144 wakeups/sec on an otherwise idle taskbar → battery/perf cost). A clean alternative exists, so it's worth changing: drive re‑detection from a `SetTimer`(≈500 ms, killed in teardown) / `DispatcherQueueTimer`, or from `SizeChanged`/`LayoutUpdated` on the watched elements, instead of a per‑frame tick.

**3. Popup XAML host is created at `(0,0)` (primary monitor) and there's no `WM_DPICHANGED` handling — worth verifying on mixed‑DPI multi‑monitor.**
`EnsureExpandedPopup` creates `g_expandedPopup` at `0,0` and then calls `InitializePopupXamlHost` while the window is still on the primary monitor ([L9345‑L9355](https://github.com/ramensoftware/windhawk-mods/blob/e260edba3ca056f9e2a8ca31ebd6048ea153fcee/mods/island-media-controls.wh.cpp#L9345)), and `ExpandedPopupWndProc` has no `WM_DPICHANGED` case. Your native px math self‑corrects each frame via `GetDpiForWindow`, but the `XamlRoot`'s rasterization scale is fixed when the host is initialized (primary‑monitor DPI) and isn't refreshed when `PositionExpandedPopup` later moves the window to the taskbar's monitor. On a config where the taskbar is on a non‑primary monitor with a different DPI (e.g. primary 100%, taskbar monitor 150%), the popup HWND is sized in 150% physical px but the XAML content can still rasterize at 100% → contents ~33% too small / misaligned. Since this PR's headline is the DPI fixes, please verify this specific case; if it reproduces, create the popup already positioned on the taskbar's monitor before `InitializePopupXamlHost`, and/or handle `WM_DPICHANGED` to re‑run `PositionExpandedPopup`/`UpdatePopupXamlVisuals`.

**4. Material migration locks users out of `mica_like`.**
`ApplySettingsMigrations` ([L1104‑L1120](https://github.com/ramensoftware/windhawk-mods/blob/e260edba3ca056f9e2a8ca31ebd6048ea153fcee/mods/island-media-controls.wh.cpp#L1104)) consults the persistent `MigrateMicaLikeMaterialToAcrylic` flag on *every* read and overrides `mica_like` → `acrylic`, only clearing the flag when the material is something *other* than `mica_like`:

```cpp
if (Wh_GetIntValue(kMigrateMicaLikeMaterialValue, 0)) {
    if (settings->material == L"mica_like") {
        settings->material = L"acrylic";     // always overrides, never clears
    } else {
        Wh_SetIntValue(kMigrateMicaLikeMaterialValue, 0);
    }
}
```

So a migrated user who genuinely wants `mica_like` back and selects it in settings is silently forced to `acrylic` — they first have to pick `solid`/`acrylic`/`liquid_glass` (which clears the flag) and only *then* can they select `mica_like`. That's a confusing dead setting for the affected users. Consider treating an explicit selection of `mica_like` after migration as a real user choice too (e.g. clear the flag whenever `Wh_ModSettingsChanged` observes a material selection), so the migration override only applies to the untouched old default.

<p>

Minor polish and edge‑case hardening — none of this affects users in normal operation, so it's your call.

- **Leftover `IMC_WGC_OVERLAY_AVAILABLE` conditional (follow‑up to the maintainer's comment).** The macro is now hardcoded to `1` ([L149](https://github.com/ramensoftware/windhawk-mods/blob/e260edba3ca056f9e2a8ca31ebd6048ea153fcee/mods/island-media-controls.wh.cpp#L149)), which does resolve the "behaves differently across Windhawk versions" concern, but it leaves the `#if IMC_WGC_OVERLAY_AVAILABLE` guards and the whole `#else` stub branch ([L8450‑L8459](https://github.com/ramensoftware/windhawk-mods/blob/e260edba3ca056f9e2a8ca31ebd6048ea153fcee/mods/island-media-controls.wh.cpp#L8450)) as dead code. Cleanest is to drop the macro and the `#if/#else/#endif` entirely and keep only the real implementation.
- **`sessionListChanged` is a stack‑local `std::atomic_bool` captured by reference in the `SessionsChanged` handler** ([L9734](https://github.com/ramensoftware/windhawk-mods/blob/e260edba3ca056f9e2a8ca31ebd6048ea153fcee/mods/island-media-controls.wh.cpp#L9734) / [L9815‑L9819](https://github.com/ramensoftware/windhawk-mods/blob/e260edba3ca056f9e2a8ca31ebd6048ea153fcee/mods/island-media-controls.wh.cpp#L9815)). Removing the event token doesn't fence out an invocation already in flight on a threadpool thread, so a session appearing/disappearing exactly at teardown could write to the (now‑gone) stack frame. Make it a file‑scope atomic like the other `g_media*` flags and drop the `[&]` capture. (The other four handlers capture `[]` and are fine.)
- **Staging texture can be left mapped if an allocation throws between `Map` and `Unmap`.** `Map` is at [L8012](https://github.com/ramensoftware/windhawk-mods/blob/e260edba3ca056f9e2a8ca31ebd6048ea153fcee/mods/island-media-controls.wh.cpp#L8012), the `layeredPixels` allocation is *after* it, and `Unmap` is only at [L8090](https://github.com/ramensoftware/windhawk-mods/blob/e260edba3ca056f9e2a8ca31ebd6048ea153fcee/mods/island-media-controls.wh.cpp#L8090); the `catch` doesn't unmap. A `std::bad_alloc` would leave the staging texture mapped and break every subsequent frame until Stop/Start recreates it. Wrap the map in an RAII unmap guard (or allocate `layeredPixels` before the `Map`).
- **`StartPopupOverlayWgcBackdrop` exception paths** ([L8415‑L8433](https://github.com/ramensoftware/windhawk-mods/blob/e260edba3ca056f9e2a8ca31ebd6048ea153fcee/mods/island-media-controls.wh.cpp#L8415)) null out the frame pool/session via `ResetPopupOverlayWgcDeviceResourcesLocked()` without revoking `FrameArrived` or calling `Close()` (as `StopPopupOverlayWgcBackdrop` does). The `running=false` guard prevents a use‑after‑free, so this is just lazy teardown — mirror the Stop path for consistency.
- **A few WGC status flags written under the mutex in the callback are read on the UI thread without it** (e.g. `g_popupOverlayWgcReadbackHadVisibleFrame`, `g_popupOverlayWgcDiagnosticState`, `g_popupOverlayWgcFallbackPainted`, `g_popupBackdropOverlayLastRadius`). Torn/stale reads are technically UB; in practice they only cause a one‑frame‑late fallback/handoff decision. Make them `std::atomic` or read under the mutex.
- **`EnsureExpandedPopup` passes DIP values as physical‑px `CreateWindowExW` dimensions** ([L9349‑L9350](https://github.com/ramensoftware/windhawk-mods/blob/e260edba3ca056f9e2a8ca31ebd6048ea153fcee/mods/island-media-controls.wh.cpp#L9349)). Harmless because `PositionExpandedPopup` sets the real px size before `ShowWindow`, but it's a DIP‑as‑px mismatch — use `PopupDipToPx(...)` for consistency.
- **`PopupPxToDip`** ([L5953](https://github.com/ramensoftware/windhawk-mods/blob/e260edba3ca056f9e2a8ca31ebd6048ea153fcee/mods/island-media-controls.wh.cpp#L5953)) is defined but never used — dead code.
- **The five fine‑grained backdrop tuning settings** (`BackdropHandoffDelayMs`, `BackdropHandoffFadeMs`, `BackdropInitialFrameSkip`, `BackdropFallbackBlurPasses`, `BackdropFallbackCaptureScale`, `BackdropWgcBlurStdDev`) are fairly obscure knobs that most users won't understand or tune. Consider hardcoding sensible defaults and only exposing them if there's real demand — fewer settings is usually better.

</p>
</details>

<p>

- **Live WGC monitor capture runs continuously while the popup is open**, which is a real GPU cost inherent to a live blurred backdrop — no cleaner alternative given the effect, just noting it. The fallback CPU blur path mitigates the startup window.
- **`StopMediaThread`'s detach‑on‑timeout** ([L9942‑L9958](https://github.com/ramensoftware/windhawk-mods/blob/e260edba3ca056f9e2a8ca31ebd6048ea153fcee/mods/island-media-controls.wh.cpp#L9942)) is a reasonable safety valve to avoid hanging on unload; in practice it's very unlikely to trigger, since the media thread's only blocking cross‑thread call (`RunFromWindowThread` → `SendMessage` to the taskbar thread) completes fine while Explorer keeps pumping. No change needed; it's a deliberate tradeoff over an unconditional `join()`.
- **The browser/live artwork identity logic** ([L10572‑L10643](https://github.com/ramensoftware/windhawk-mods/blob/e260edba3ca056f9e2a8ca31ebd6048ea153fcee/mods/island-media-controls.wh.cpp#L10572)) reads cleanly: the `MediaIdentityKey`‑based `differentMedia`/`repeatedPreviousArtwork`/`browserLiveLike` gating suppresses stale repeats for the 3.5 s grace window without dropping legitimately shared album art, and it runs entirely on the UI thread (function‑local statics), so there's no threading concern there.

</p>
</details>

usho-lear:
Updated the existing PR branch to `0.9.146` in commit `6d27adb`.

Latest maintainer feedback addressed:

1. WGC frame rendering no longer performs cross-thread `UpdateLayeredWindow`, `ShowWindow`, or display-affinity calls while holding `g_popupOverlayWgcMutex`. Frames are now copied into a latest-frame queue and presented by the overlay window on its owning UI thread via `PostMessage`.
2. The permanent `CompositionTarget::Rendering` taskbar-layout poll was replaced with a 600 ms Win32 timer.
3. The expanded popup is created on the taskbar monitor before XAML host initialization and now handles `WM_DPICHANGED`.
4. The old-default migration is cleared when settings explicitly retain Mica-like, so migrated users can select Mica-like normally.
5. Removed the hardcoded WGC availability macro and unreachable stub branch requested in the inline review.

Additional review polish:
- D2D factory now uses multi-threaded mode.
- WGC frames are explicitly closed.
- Readback storage is allocated before `Map`.
- Removed the duplicate Windows.Foundation include.
- The previously requested GSMTC manager-before-uninit and explicit WGC device teardown fixes remain in place.

The README now includes the full changelog and hosted dark/light previews for Mica-like, Acrylic, and Liquid Glass.

Validation:
- Windhawk Clang `-fsyntax-only`: passed
- `git diff --check`: passed
- Unicode format-control scan: 0 matches
- Six raw preview links: all `200 image/png`

m417z:
### Submission review

---

**1. WGC/D3D/D2D globals run `Release()` at process exit, not just on `Wh_ModUninit` — can hang or crash Explorer.** The capture-device globals are value types with non-trivial destructors, not intentionally-leaked pointers:

```cpp
winrt::com_ptr<ID3D11Device>        g_popupOverlayWgcD3dDevice;       // ~L556
winrt::com_ptr<ID3D11DeviceContext> g_popupOverlayWgcD3dContext;
winrt::com_ptr<IDXGIDevice>         g_popupOverlayWgcDxgiDevice;
winrt::com_ptr<ID3D11Texture2D>     g_popupOverlayWgcRenderTexture;
winrt::com_ptr<ID3D11Texture2D>     g_popupOverlayWgcStagingTexture;
winrt::com_ptr<ID2D1Factory1>       g_popupOverlayWgcD2dFactory;
winrt::com_ptr<ID2D1Device>         g_popupOverlayWgcD2dDevice;
winrt::com_ptr<ID2D1DeviceContext>  g_popupOverlayWgcD2dContext;
winrt::com_ptr<ID2D1Bitmap1>        g_popupOverlayWgcLensDisplacementMap;
direct3d11::IDirect3DDevice         g_popupOverlayWgcGraphicsDevice{nullptr};  // ~L569
capture::GraphicsCaptureItem        g_popupOverlayWgcItem{nullptr};
capture::Direct3D11CaptureFramePool g_popupOverlayWgcFramePool{nullptr};
capture::GraphicsCaptureSession     g_popupOverlayWgcSession{nullptr};
```

Windhawk does **not** call `Wh_ModUninit` when the host process itself terminates (Explorer restart, sign-out, reboot) — the OS just detaches the DLL and the CRT runs these globals' destructors during `DLL_PROCESS_DETACH`, after every other thread has already been terminated and while other modules (`combase`, `dwmcore`, `d3d11`/`dxgi`) are mid-detach. Tearing down a live `GraphicsCaptureSession`/`FramePool` against the capture service, and `Release()`ing the D3D/D2D objects, on that path is the classic "release WGC/D3D at raw process exit" hazard → hang or AV in the host during shutdown. The trigger is realistic: an expanded popup with the Acrylic or Liquid Glass material is open and capturing at the moment Explorer restarts (on a normal popup close these globals are already reset to null, so the destructors are no-ops — the risk is specifically the process-termination-while-open case).

You already apply the correct pattern to the media thread (`std::thread* g_mediaThread` at ~L399 is a raw pointer that's intentionally leaked at shutdown, so `~thread()`/`std::terminate` can't fire). Do the same for the capture resources: hold them behind heap pointers that are only released inside `Wh_ModUninit` (and on popup close) and intentionally left alone otherwise, so nothing runs from a static destructor at process termination.

Human note: This caused crashes with other mods in the past, so worth fixing. An easier solution is to simply add `[[clang::no_destroy]]` to the global variables.

**2. `PopupBackdropOverlayWndProc` has no top-level `try/catch`, so a C++ exception from the timer-driven WGC init can escape through `DispatchMessage` and crash Explorer.** `EnsurePopupOverlayWgcDeviceResources` uses throwing `winrt::com_ptr::as<>()`:

```cpp
g_popupOverlayWgcDxgiDevice = g_popupOverlayWgcD3dDevice.as<IDXGIDevice>();          // L7817
...
g_popupOverlayWgcGraphicsDevice = graphicsDeviceInspectable.as<direct3d11::IDirect3DDevice>();  // L7855
```

That function is called from `StartPopupOverlayWgcBackdrop` at L8568 — **before** Start's own `try` (which begins at L8612) — and Start is reached from `UpdatePopupBackdropOverlayWindow`, which is called directly from the `WM_TIMER` case of `PopupBackdropOverlayWndProc` (L9164-9166). Unlike `ExpandedPopupWndProc`, this window proc does **not** wrap its body in `try/catch`, so a throw would propagate as an unhandled exception out of the message pump. The likelihood of these particular QIs actually failing is essentially nil (a D3D11 device always implements `IDXGIDevice`; the WGC inspectable always implements `IDirect3DDevice`), so this is low-probability — but the fix is trivial and matches what you already do elsewhere: wrap the body of `PopupBackdropOverlayWndProc` in the same `try { ... } catch (...) { }` that `ExpandedPopupWndProc` uses (or extend Start's `try` to cover the `EnsurePopupOverlayWgcDeviceResources` call).

<p>

Minor cleanup — none of this affects users, so it's your call.

- **Dead code:** `RenderPopupOverlayWgcDiagnosticFrame()` (L7871) is an empty function with no callers — remove it.
- **Unused constant:** `kSettingsMigrationVersion` (L1136) is never referenced; `ApplySettingsMigrations` uses the literal `1` (`< 1`) instead. Either use the constant or drop it.
- **Diagnostic instrumentation:** there's a large amount of `g_popupOverlayWgc*Diagnostic*` state (the `PopupOverlayWgcDiagnosticState` enum, `g_popupOverlayWgcDiagnosticHr`, `...RenderFailCount`, etc.) that exists almost entirely to feed `Wh_Log` — only one assignment (the `NotStarted` check at L9542) drives real behavior. Now that the WGC path is stable, consider trimming this development scaffolding; it's a fair bit of state and branching to carry for logging.

</p>
</details>

<p>

Non-critical observation about the feature itself.

- The live-blur materials capture the **whole monitor** via WGC and run a per-frame D2D Gaussian blur (plus the SDF lens warp for Liquid Glass) while the expanded popup is open. That's inherently heavy, but it's the core mechanism of the feature and you already gate it well (only while the popup is open, a static fallback frame during the morph, initial-frame skip, and a 3-second retry throttle after a `CreateForMonitor` failure), so there's no obviously cleaner alternative — noting it only as an FYI.

</p>
</details>

usho-lear:
Addressed the latest review feedback in commit 60bfd0a:

- Marked the WGC/D3D/D2D/WinRT capture globals with [[clang::no_destroy]] so they are not released from static destructors during Explorer process shutdown.
- Wrapped PopupBackdropOverlayWndProc in a top-level try/catch, matching the expanded popup wndproc exception boundary.
- Also removed the unused RenderPopupOverlayWgcDiagnosticFrame() stub and changed the settings migration check to use kSettingsMigrationVersion.

Validation:
- D:\Windhawk\Compiler\bin\clang++.exe '@D:\Windhawk\Compiler\compile_flags.txt' -ID:\Windhawk\Compiler\include -fsyntax-only mods\island-media-controls.wh.cpp
- git diff --check -- mods/island-media-controls.wh.cpp

RajStartsHere:
_Greetings_,
@usho-lear

A Feature/Compatibility Request, I Think This Will Be Easy To Implement; (Specifically I've Been Unavailable/Busy From Sometime, Which Indeed Keeping Me Away From The GitHub, Lacking The Time To Do Things... _Enough Yapping_)

The Mod Seems To Be Incompatible With The Mod [Taskbar On Top For Windows 11](https://www.windhawk.net/mods/taskbar-on-top)

This Can Be Fixed Directly By Either Providing The Option To Toggle Its Position Changing The Animations, Motion, And The Orientation Relevant To The Options Named: Left, Right, Down, And Default (Up)

As There Are Or This Mod Specifically Can Be Tweak To Change The Position Of The Taskbar To Either Left Or Right Configurably

_Hope This Helps!_

usho-lear:
> _Greetings_, @usho-lear
> 
> A Feature/Compatibility Request, I Think This Will Be Easy To Implement; (Specifically I've Been Unavailable/Busy From Sometime, Which Indeed Keeping Me Away From The GitHub, Lacking The Time To Do Things... _Enough Yapping_)
> 
> The Mod Seems To Be Incompatible With The Mod [Taskbar On Top For Windows 11](https://www.windhawk.net/mods/taskbar-on-top)
> 
> This Can Be Fixed Directly By Either Providing The Option To Toggle Its Position Changing The Animations, Motion, And The Orientation Relevant To The Options Named: Left, Right, Down, And Default (Up)
> 
> As There Are Or This Mod Specifically Can Be Tweak To Change The Position Of The Taskbar To Either Left Or Right Configurably
> 
> _Hope This Helps!_

get it, for now I couldn't find a mod that can tweak the taskbar's position to the side though so i am only making it compatible to the mod you just mentioned. It'll change it's position accordingly and automatically which I think would feels more seamlessly though,

usho-lear:
> _Greetings_, @usho-lear
> 
> A Feature/Compatibility Request, I Think This Will Be Easy To Implement; (Specifically I've Been Unavailable/Busy From Sometime, Which Indeed Keeping Me Away From The GitHub, Lacking The Time To Do Things... _Enough Yapping_)
> 
> The Mod Seems To Be Incompatible With The Mod [Taskbar On Top For Windows 11](https://www.windhawk.net/mods/taskbar-on-top)
> 
> This Can Be Fixed Directly By Either Providing The Option To Toggle Its Position Changing The Animations, Motion, And The Orientation Relevant To The Options Named: Left, Right, Down, And Default (Up)
> 
> As There Are Or This Mod Specifically Can Be Tweak To Change The Position Of The Taskbar To Either Left Or Right Configurably
> 
> _Hope This Helps!_

image
image
image
so i get it work for now but still polishing it's performance, and a also added a compact mode that user can toggle on or off for i thing might look somehow better this way,

RajStartsHere:
> > _Greetings_, @usho-lear
> > A Feature/Compatibility Request, I Think This Will Be Easy To Implement; (Specifically I've Been Unavailable/Busy From Sometime, Which Indeed Keeping Me Away From The GitHub, Lacking The Time To Do Things... _Enough Yapping_)
> > The Mod Seems To Be Incompatible With The Mod [Taskbar On Top For Windows 11](https://www.windhawk.net/mods/taskbar-on-top)
> > This Can Be Fixed Directly By Either Providing The Option To Toggle Its Position Changing The Animations, Motion, And The Orientation Relevant To The Options Named: Left, Right, Down, And Default (Up)
> > As There Are Or This Mod Specifically Can Be Tweak To Change The Position Of The Taskbar To Either Left Or Right Configurably
> > _Hope This Helps!_
> 
> image image image so i get it work for now but still polishing it's performance, and a also added a compact mode that user can toggle on or off for i thing might look somehow better this way,

I Believe So, For Me... It Is Still Looking Insanely Good Even Without The Little Tweaks

PR #4699, opened 2026-07-06
PivotLink: Browser Router 1.0

gauthumj:
## Changelog

## Mod authorship

This mod was created by:

- - [X] Claude
- - [ ] ChatGPT
- - [X] Gemini


m417z:
### Submission review

---

A few of these are functional blockers — the mod currently fails for its own advertised use case (Discord/Slack), and one hook reports failure to the caller even after a successful redirect.

**`(HINSTANCE)32` / `hInstApp = 32` signals *failure*, not success.** `ShellExecute`/`ShellExecuteEx` treat any value **≤ 32** as an error (`32` == `SE_ERR_DLLNOTFOUND`); success is strictly **> 32**. So after `ShellExecuteW_Hook` redirects a link, it returns `(HINSTANCE)32` (`pivotlink-browser-router.cpp:288`) and `ShellExecuteExW_Hook` sets `hInstApp = (HINSTANCE)32` (line 276) — both read as *"the launch failed"*. Callers that check the return (`if ((INT_PTR)ShellExecuteW(...) <= 32) { error }` is the standard idiom — see [`mods/auto-theme-switcher.wh.cpp#L363`](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/auto-theme-switcher.wh.cpp#L363)) will show an error dialog or fall back to another launch path, even though the browser did open. Return a value **> 32** (33 is the convention — [`mods/f1-blocker.wh.cpp#L82`](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/f1-blocker.wh.cpp#L82), [`mods/classic-taskbar-properties.wh.cpp#L96`](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/classic-taskbar-properties.wh.cpp#L96)).

**The D3D "game detection" disables the mod in the exact apps it targets.** `IsCurrentProcessGame()` (line 156) returns `true` for any process with `d3d9/11/12.dll` or `vulkan-1.dll` loaded, and the only exclusions are the configured browsers. But Electron/Chromium apps — including **Discord and Slack, the README's own motivating examples** — routinely load D3D (ANGLE), as do WPF/UWP apps, media players, Office, etc. In all of those, `RouteLinkIfNecessary` bails out at line 234 and the link is never redirected, so the mod silently does nothing in its primary use case. Recommend dropping the game-detection feature entirely (simplest, and removes the false positives) — disabling redirection specifically in games is a niche need that isn't worth this cost. If you keep it, it needs a far more reliable signal than "loaded a graphics DLL."

**The `CreateProcessW` hook is hardcoded to `zen.exe`.** The entire hook body only does anything when the command line or application name contains the literal string `zen.exe` (lines 306–307). This looks like leftover personal configuration (or an AI artifact) — it doesn't match the generic, browser-agnostic mod described in the README, and no user other than a Zen Browser user gets any `CreateProcessW` handling at all. On top of that, when it does fire, the synthesized `PROCESS_INFORMATION` is broken: `hThread = NULL`, `dwThreadId = 0`, and `hProcess` points at a *different* browser than the caller asked to launch (lines 340–345). Any caller that waits on the returned process, resumes the thread (`CREATE_SUSPENDED` → `ResumeThread(NULL)`), or reads the PID will misbehave. Either generalize this to detect *any* configured browser being launched with a URL, or drop the `CreateProcessW` hook — which would also meaningfully shrink the mod's blast radius (see below).

**`@author You` is a placeholder** (line 6). It should be your real name/handle; your `@github` is already `gauthumj`.

<p>

- **Settings-change data race.** `LoadSettings()` does `g_priorityBrowsers.clear()` and repopulates it (lines 207–223) on the settings-changed thread while other threads iterate the same vector in `GetHighestPriorityRunningBrowser()` / `IsCurrentProcessGame()`. That's a data race on a `std::vector` that can crash on a settings change. Guard the globals with a mutex (copy out a local snapshot under the lock before use) or swap in a freshly-built vector atomically.
- **Drop the custom `[PivotLink]` log prefix** (lines 253, 327, 366). Windhawk already prefixes every `Wh_Log` line with the mod name, so the bracketed prefix is redundant.
- **Dead / redundant default handling.** `Wh_GetStringSetting` never returns `NULL` (it returns `L""` on miss), so the `if (rawBrowser)`/`if (rawSetting …)` NULL checks, the `hardcodedDefaults[]` fallback branch (lines 210, 220–222), and the `if (g_priorityBrowsers.empty())` fallback (lines 225–227) are all unreachable — the defaults already live in the `==WindhawkModSettings==` block. This can be simplified a lot.
- **Prefer the type-safe utilities.** Use `WindhawkUtils::SetFunctionHook()` instead of raw `Wh_SetFunctionHook` with `(void*)` casts (lines 370–372), and `WindhawkUtils::StringSetting` (RAII) instead of `Wh_GetStringSetting` + manual `Wh_FreeStringSetting`.
- **Call the original directly.** `RouteLinkIfNecessary` invokes the hooked `ShellExecuteExW` (line 262) and relies on `t_inHook` to avoid recursing; calling `ShellExecuteExW_Original` directly is clearer and avoids the extra hook round-trip.

</p>
</details>

<p>

- **`@include *` blast radius.** Catching link opens from arbitrary apps inherently requires broad injection, so `*` is defensible for the `ShellExecute*` hooks — there's no cleaner alternative. But be aware it injects and installs three hooks into *every* user-session process; dropping the `CreateProcessW` hook (see above) would at least remove one of them from every process.
- **Per-call snapshot cost.** `GetHighestPriorityRunningBrowser()` takes a full `CreateToolhelp32Snapshot` and runs `EnumWindows` per candidate PID on every intercepted link/launch. Fine given link opens are infrequent, just noting it.
- **Verb is ignored.** The `ShellExecute*` hooks redirect any `http(s)` `lpFile` to a browser regardless of the operation/verb, so a non-`open` verb (e.g. `edit`, `print`) on a URL would still get rerouted. Edge case, but worth a guard if you want to be precise.

</p>
</details>

gauthumj:
Hi @m417z , thank you for the recommendations. I have tried my best to address them and tested them out for a day to make sure the mod doesn't get in the way(unless there's a link to open). Here's the summary:

## Changes addressing review feedback

### 1. `(HINSTANCE)32` signals failure, not success
Fixed. Both `ShellExecuteW_Hook` and `ShellExecuteExW_Hook` now return `(HINSTANCE)33` / set `hInstApp = (HINSTANCE)33`, which is the standard success convention for hooked ShellExecute calls.

### 2. D3D "game detection" removed entirely
Dropped `IsCurrentProcessGame()` completely. As noted, Electron/Chromium apps (Discord, Slack, VSCodium) load D3D via ANGLE, which caused the mod to silently disable itself in its primary use case. No replacement heuristic was needed — the mod's early-exit guards (Session 0 bypass, verb check, URL check, circular routing prevention) are sufficient without a blanket graphics-DLL filter.

### 3. `CreateProcessW` hook generalized (not removed)
The hook is no longer hardcoded to `zen.exe`. It now:
- Skips if the **calling process** is any configured browser (prevents cascading redirects and internal URL noise like `cr.brave.com`).
- Extracts the **target executable** name from `lpApplicationName` or the command line.
- Only intercepts if the target exe matches a browser in the priority list (prevents false positives from `git.exe`, `curl.exe`, etc. that have URLs in their args).
- Rewrites `lpApplicationName` + `lpCommandLine` in-place to the highest-priority running browser via `CreateProcessW_Original`, so the returned `PROCESS_INFORMATION` is valid and correct.

**Why kept**: Apps like WhatsApp (and other UWP/packaged apps) call `CreateProcessW` directly with the default browser path + URL, bypassing ShellExecute entirely. Without this hook, links from those apps always open in the OS default browser regardless of what's running. The false-positive and blast-radius concerns are now addressed by the target-exe check — the hook exits immediately (single string comparison) for any non-browser `CreateProcessW` call.

### 4. `@author` fixed
Changed from placeholder `You` to `gauthumj`.

---

### Optional improvements (all addressed)

- **Settings-change data race**: Added `std::mutex g_settingsMutex`. `LoadSettings()` builds a new vector and swaps it in under the lock. All readers copy the vector under the lock before iterating.
- **`[PivotLink]` log prefix**: Removed. Only two minimal `Wh_Log` calls remain (one for ShellExecute redirects, one for CreateProcessW rewrites).
- **Dead/redundant default handling**: Removed `hardcodedDefaults[]`, NULL checks, and the `if (g_priorityBrowsers.empty())` fallback. Settings come exclusively from the `==WindhawkModSettings==` defaults.
- **Type-safe utilities**: Now using `WindhawkUtils::SetFunctionHook()` (no `(void*)` casts) and `WindhawkUtils::StringSetting::make()` (RAII, no manual free).
- **Call original directly**: `RouteLinkIfNecessary` now calls `ShellExecuteExW_Original` directly instead of going through the hooked entry point.

---

### Functionality notes (addressed)

- **`@include *` blast radius**: Kept — required for the ShellExecute hooks. The CreateProcessW hook's cost is mitigated by exiting on the first string comparison when the target isn't a browser.
- **Per-call snapshot cost**: Acknowledged as acceptable given link opens are infrequent.
- **Verb guard added**: `RouteLinkIfNecessary` now only redirects when the verb is `NULL` (default) or `"open"`. Non-open verbs pass through unmodified.

m417z:
### Submission review

---

**`ShellExecuteEx` fabricated success doesn't populate `hProcess`.** When routing succeeds you set `pExecInfo->hInstApp = (HINSTANCE)33` and return `TRUE` (lines 222-224), but if the caller passed `SEE_MASK_NOCLOSEPROCESS` and then waits on / closes `pExecInfo->hProcess`, that field is left untouched. At the very least, initialize it to `nullptr`.

<p>

Minor polish — none of this affects correctness, so it's your call.

- **Cache the current process name.** `GetCurrentProcessName()` (line 148) calls `GetModuleFileNameW` every time, and it's invoked in `RouteLinkIfNecessary` and on *every* `CreateProcessW` in every process (the `CreateProcessW` hot path). The executable name never changes for the life of a process — compute it once into a `static` and reuse it.

</p>
</details>

<p>

- **Only `ShellExecuteW`/`ShellExecuteExW` are hooked, not the ANSI variants.** Apps that open URLs via `ShellExecuteA` won't be caught. [url-blocker](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/url-blocker.wh.cpp) hooks both `ShellExecuteW` and `ShellExecuteA` for exactly this reason — you may want the same for completeness.

</p>
</details>

gauthumj:
Hello! I have implemented the suggested changes. Here's the changelog:

## Changes addressing review feedback (Round 2)

### 1. `ShellExecuteEx` fabricated success doesn't populate `hProcess`
Fixed. When routing succeeds and the caller passed `SEE_MASK_NOCLOSEPROCESS`, `pExecInfo->hProcess` is now explicitly set to `NULL`. This prevents callers from waiting on or closing a stale/uninitialized handle. Applied to both the wide (`ShellExecuteExW_Hook`) and ANSI (`ShellExecuteExA_Hook`) hooks.

### Optional improvements (addressed)

- **Cached current process name**: `GetCurrentProcessName()` previously called `GetModuleFileNameW` on every invocation — including the `CreateProcessW` hot path in every process. Since the executable name never changes for the life of a process, it's now computed once into a `static` local (via an immediately-invoked lambda) and returned by `const&`.

### Functionality notes (addressed)

- **ANSI `ShellExecute` variants**: Added hooks for `ShellExecuteA` and `ShellExecuteExA`. Apps that open URLs via the ANSI API are now caught — strings are widened via `MultiByteToWideChar` and delegated to the existing `RouteLinkIfNecessary` logic. This matches the approach used by [url-blocker](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/url-blocker.wh.cpp). The mod now hooks all five entry points: `ShellExecuteW`, `ShellExecuteExW`, `ShellExecuteA`, `ShellExecuteExA`, and `CreateProcessW`.

PR #4706, opened 2026-07-07
Restore Explorer "Exploring" Mode 1.4.0

aubymori:
## Changelog

- Introduce Vista tree initial state behavior
- Fix bug where setting "Explorer icon mode" to the Vista option didn't work

### DIFF
--- a/mods/restore-explorer-exploring-mode.wh.cpp
+++ b/mods/restore-explorer-exploring-mode.wh.cpp
@@ -2,7 +2,7 @@
 // @id              restore-explorer-exploring-mode
 // @name            Restore Explorer "Exploring" Mode
 // @description     Reintroduces File Explorer's "Exploring" mode from Windows XP and before
-// @version         1.3.0
+// @version         1.4.0
 // @author          aubymori
 // @github          https://github.com/aubymori
 // @include         *
@@ -69,11 +69,15 @@ Windows Registry Editor Version 5.00
   $description:
     When enabled, taking the "Open" action on a folder in an open/save dialog will open
     the folder in a new Explorer window (Windows XP and before).
-- dont_store_tree_state: true
-  $name: Don't store tree view state
+- dont_store_tree_state: "1"
+  $name: Tree view initial state mode
   $description:
-    When enabled, the tree view's initial open state will be based on whether the window was opened in
-    explore mode and will not be carried between windows (Windows XP and before).
+    Changes the behavior of the tree view's initial open state to match the selected
+    OS.
+  $options:
+  - 1: Windows 95 - XP
+  - vista: Windows Vista
+  - 0: Windows 7+
 */
 // ==/WindhawkModSettings==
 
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

---

Most of the concerns are in the one new block — the `TVM_VISTA` branch added to `CMemPropStore_Write_hook` (`restore-explorer-exploring-mode.wh.cpp:374-402`). The icon-mode `else` fix and the settings-to-code wiring for the new dropdown look correct.

- **The Vista mode writes a persistent registry value, which breaks reversibility.** The new branch opens `HKCU\...\Explorer\Modules\GlobalSettings\Sizer`, reads `PageSpaceControlSizer`, flips `lpData[4] = 1`, and writes it back. That's a persistent change to a system-owned value — it isn't undone when the mod is disabled, so the tree stays "sticky-expanded" afterward until Explorer next overwrites it. A core Windhawk principle is that a mod's effects disappear on disable and it must not make persistent registry changes; the guidance is to *hook the read* instead of writing. Recommended fix: hook the read of `PageSpaceControlSizer` (`RegQueryValueExW`/`RegGetValueW`) and OR the bit into the returned buffer, or carry the "tree expanded" state in the mod's own storage (`Wh_GetIntValue`/`Wh_SetIntValue`) and apply it in-memory — either way nothing persists after disable. See mods that hook registry reads for this pattern: [spoof-light-dark-theme.wh.cpp](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/spoof-light-dark-theme.wh.cpp) and [app-theme-crash-fixer.wh.cpp](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/app-theme-crash-fixer.wh.cpp).

- **`hkey` is leaked** (`restore-explorer-exploring-mode.wh.cpp:378-384`). `RegOpenKeyExW` succeeds but `RegCloseKey(hkey)` is never called, so this leaks a registry-key handle every time an explore-mode window writes `ExpandInitialNav` while in Vista mode — it accumulates over a session. The file already pulls in WIL, so `wil::unique_hkey` is the cleanest fix (or just add `RegCloseKey`).

- **`pszTreeMode` is leaked on every settings change** (`restore-explorer-exploring-mode.wh.cpp:754`). `Wh_GetStringSetting(L"dont_store_tree_state")` returns a string that must be freed, and unlike the sibling `pszIconMode` (freed on line 753) this one never is. Add `Wh_FreeStringSetting(pszTreeMode)`, or read it via `WindhawkUtils::StringSetting`.

- **Unchecked write to `lpData[4]`** (`restore-explorer-exploring-mode.wh.cpp:394`). Offset 4 is written without verifying the value is at least 5 bytes. `PageSpaceControlSizer` is normally a large blob so this won't bite in practice, but if the value is ever missing/short/empty, `LocalAlloc(LPTR, cbData)` yields a buffer smaller than 5 and `lpData[4] = 1` overflows it. Guard with `if (cbData > 4)` before the write. (This concern goes away entirely if you switch to hooking the read per the first item.)

<p>

- Leftover blank lines at `restore-explorer-exploring-mode.wh.cpp:136-137` where `g_fDontStoreTreeState` was removed.
- If you keep the `LocalAlloc` path, `wil::unique_hlocal` would match the RAII style already used elsewhere in the file (`wil::unique_cotaskmem_string`, `wil::com_ptr`).

</p>
</details>

aubymori:
Applied most of the recommended changes. As for the Vista mode writing a persistent registry value, all it does is makes opening Explorer in explore mode enable the navigation pane for any future Explorer windows instead of just the one, as it was in Vista. After disabling the mod, hiding the navigation pane through the GUI is trivial.

PR #4716, opened 2026-07-08
Update 1.1.0

Lockframe:
## Changelog

* Added 'Elastic' animation mode, which is a hybrid of stretch and bounce, inspired by Apple's Dynamic Island animations.
* Added spring physics toggle for smoother animations, which is especially noticeable paired with bounce or elastic animation modes.
* Added a speed multiplier setting.
* Added a simple color-crossfade animation when using app icon color mode.
* Fixed initialization bug where the pill renders at the far left of the taskbar.
* Fixed an issue where the pill retains a closed app's position before animating to the currently focused app's.

### DIFF
--- a/mods/taskbar-elastic-pill.wh.cpp
+++ b/mods/taskbar-elastic-pill.wh.cpp
@@ -1163,6 +1163,7 @@ void EnsurePillAndPosition(winrt::Windows::UI::Xaml::FrameworkElement const& but
     }
 
     std::wstring stableKey = L"";
+    bool isExtracting = false;
 
     if (anyActive) {
         std::wstring appName = winrt::Windows::UI::Xaml::Automation::AutomationProperties::GetName(targetButton).c_str();
@@ -1183,6 +1184,9 @@ void EnsurePillAndPosition(winrt::Windows::UI::Xaml::FrameworkElement const& but
                     if (!g_iconColorCache->count(stableKey) && (it == g_iconColorExtracting.end() || !it->second)) {
                         needsExtract = true;
                         g_iconColorExtracting[stableKey] = true;
+                        isExtracting = true;
+                    } else {
+                        isExtracting = (it != g_iconColorExtracting.end() && it->second);
                     }
                 }
                 if (needsExtract) {
@@ -1196,10 +1200,12 @@ void EnsurePillAndPosition(winrt::Windows::UI::Xaml::FrameworkElement const& but
                         } else {
                             std::lock_guard<std::mutex> lock(g_iconColorMutex);
                             g_iconColorExtracting[stableKey] = false;
+                            isExtracting = false;
                         }
                     } else {
                         std::lock_guard<std::mutex> lock(g_iconColorMutex);
                         g_iconColorExtracting[stableKey] = false;
+                        isExtracting = false;
                     }
                 }
             }
@@ -1219,7 +1225,7 @@ void EnsurePillAndPosition(winrt::Windows::UI::Xaml::FrameworkElement const& but
                 std::lock_guard<std::mutex> lock(g_iconColorMutex);
                 hasColor = g_iconColorCache->count(stableKey) > 0;
             }
-            if (settingsToUse.ColorMode != 2 || hasColor) {
+            if (settingsToUse.ColorMode != 2 || !isExtracting || hasColor) {
                 winrt::Windows::UI::Color newColor = GetPillColor(settingsToUse, stableKey);
                 winrt::Windows::UI::Color oldColor = brush.Color();
                 if (oldColor.A != newColor.A || oldColor.R != newColor.R || oldColor.G != newColor.G || oldColor.B != newColor.B) {
--- a/mods/taskbar-elastic-pill.wh.cpp
+++ b/mods/taskbar-elastic-pill.wh.cpp
@@ -407,48 +407,16 @@ using namespace winrt::Windows::UI::Xaml::Media;
 using namespace winrt::Windows::UI::Xaml::Hosting;
 using namespace winrt::Windows::UI::Composition;
 
-int g_offsetTaskListButton = -1;
-int g_offsetSearchIconButton = -1;
-
-FrameworkElement GetFrameworkElementFromNative(void* pThis, int& cachedOffset) {
+FrameworkElement GetFrameworkElementFromNative(void* pThis) {
     if (!pThis) return nullptr;
-    void** ptrs = (void**)pThis;
-
-    if (cachedOffset != -1) {
-        try {
-            ::IUnknown* possibleUnk = (::IUnknown*)ptrs[cachedOffset];
-            if (possibleUnk) {
-                winrt::Windows::UI::Xaml::FrameworkElement result{nullptr};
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

---

**1. `GetFrameworkElementFromNative` now uses `IsBadReadPtr` and a brute-force QI scan — please drop `IsBadReadPtr` and reconsider the scan.** (lines 412-431)

`IsBadReadPtr` is on the "never use" list — it's unreliable (a readable pointer isn't a valid COM object) and can actually *cause* crashes by tripping the guard page used for stack growth, and it races under concurrent access. See [The Old New Thing](https://devblogs.microsoft.com/oldnewthing/20060927-07/?p=29563). On top of that, the new approach walks slots 1-15 of the object, treats each stored pointer as an `IUnknown*`, and calls `QueryInterface` on the first one whose first few vtable bytes happen to be readable. That's semantically different from the previous `copy_from_abi` at a fixed sub-object offset: you're now dereferencing arbitrary *member* pointers and invoking a vtable slot on whatever they point at, and this runs on **every** `UpdateVisualStates` call (hover, click, app switch), so it's also a per-event cost regression over the old O(1) path.

If the reason for the rewrite was that the fixed offset (`pThis + 3`) worked for `TaskListButton` but not for `SearchIconButton`, prefer a correct per-class offset (or two dedicated helpers) rather than scanning 15 slots and probing memory each event.

**2. `g_attachedEvents` was changed from a leaked heap pointer to a plain global with a non-trivial destructor.** (line 296)

```cpp
-std::vector<AttachedEvent>* g_attachedEvents = new std::vector<AttachedEvent>();
+std::vector<AttachedEvent> g_attachedEvents;
```

`Wh_ModUninit`/`Wh_ModBeforeUninit` do **not** run when the host process terminates (Explorer restart, sign-out, reboot) — the OS just detaches the DLL and the CRT runs global destructors during `DLL_PROCESS_DETACH`. `AttachedEvent` holds a `winrt::weak_ref` and a `CoreDispatcher`, so `~vector` now releases WinRT/COM references at process-detach time, when the XAML apartment may already be gone. The previous leaked-pointer form deliberately avoided this. Recommend keeping it a pointer (`new std::vector<...>()`). (The same latent concern applies to the other WinRT-holding globals here — `g_pillContexts`, `g_iconColorCache`, `g_attachedGroups` — which were already plain globals before this PR; worth considering the leaked-pointer pattern for those too, but this one is a regression introduced by this diff.)

**3. `HookSymbols` is called twice for `SearchUx.UI.dll` on the fallback path** (lines 1479-1484). Resolving symbols for a module a second time invalidates the per-module symbol cache and re-resolves. It only happens when `UpdateVisualStates` fails to resolve (uncommon), and it's init-time, so the impact is small — but if you want a single call, put both symbols in one array (both optional) and resolve once. That does change the semantics to "hook both if both exist" though (see the functionality note below), so only do it if that's acceptable.

<p>

- **Clamp `SpeedMultiplier` to a sane upper bound.** You correctly reject `<= 0` and NaN, but there's no ceiling — a large value (e.g. `>60`) drives the spring `Period` / animation durations to `0`, which can make `CreateSpringScalarAnimation().Period(0)` throw (swallowed by the surrounding `try/catch`, so the pill just stops animating). A clamp like `std::clamp(x, 0.1, 10.0)` keeps every derived duration valid.

</p>
</details>

<p>

- **Search-button tracking now hooks `UpdateVisualStates` *or* `PlayStateChange`, not both** (lines 1459-1486). v1.0.0 hooked both (though both wrote into the same `_Original`, which was a latent bug you've now fixed by splitting them). The new fallback only reaches `PlayStateChange` if `UpdateVisualStates` fails to resolve. If, on a given build, a search-button active-state transition ever fires *only* through `PlayStateChange` while `UpdateVisualStates` also exists, the pill won't update for it. Worth confirming that one entry point covers all the transitions you care about.
- **The `point.X == 0 && point.Y == 0` guard** (line 693) treats a transform of exactly (0,0) as "not laid out yet" and skips positioning. That's a reasonable readiness heuristic, but a button legitimately at the grid origin would never get a pill. In practice the left edge is occupied by Start/Task View, so this is unlikely to bite — just flagging the assumption.
- **Icon-color cache eviction changed from LRU-10 to a full `clear()`** (lines 623-626). Simpler, but on a taskbar with >100 distinct apps it periodically wipes and re-extracts every color instead of dropping the 10 least-recently-used. Edge case, so probably fine — just noting the behavior change.

</p>
</details>

m417z:
Both issues weren't really solved.
1. Now the code uses `ReadProcessMemory`. It's slightly better, but it'd preferable to not try to guess the offsets, which may cause instability and crashes, and instead use a known offset like other mods do. I don't know what's your level of understanding here and how much of it is AI, but I'm reluctant to keep the current solution.
2. `HookSymbols` is called twice for `SearchUx.UI.dll` - it's still the case and it will screw up the caching. Please use a single `HookSymbols` call.

mods/taskbar-elastic-pill.wh.cpp
```diff
@@ -1357,27 +1376,22 @@ HMODULE GetSearchUxModuleHandle() {
...
-        },
+    WindhawkUtils::SYMBOL_HOOK hooks[] = {
         {
-            {L"protected: void __cdecl winrt::SearchUx::SearchUI::implementation::SearchIconButton::PlayStateChange(void)"},
-            &SearchIconButton_UpdateVisualStates_Original,
-            SearchIconButton_UpdateVisualStates_Hook,
-            true
+            {
+                LR"(private: void __cdecl winrt::SearchUx::SearchUI::implementation::SearchIconButton::UpdateVisualStates(void))",
+                L"protected: void __cdecl winrt::SearchUx::SearchUI::implementation::SearchIconButton::PlayStateChange(void)"
```
m417z: If both symbols exist, an arbitrary function will be hooked, which can make the mod inconsistent.

Why not use the solution from here? https://github.com/ramensoftware/windhawk-mods/pull/4685#issuecomment-4887722793

mods/taskbar-elastic-pill.wh.cpp
```diff
@@ -1536,7 +1563,11 @@ void Wh_ModBeforeUninit() {
     }
 
     if (pending->load() > 0 && eventLifetime.get()) {
-        WaitForSingleObject(eventLifetime.get(), INFINITE);
+        if (WaitForSingleObject(eventLifetime.get(), 2000) == WAIT_TIMEOUT) {
+            Wh_Log(L"Cleanup timed out. Pinning module to prevent crash.");
+            HMODULE hMod;
+            GetModuleHandleExW(GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS | GET_MODULE_HANDLE_EX_FLAG_PIN, (LPCWSTR)&Wh_ModBeforeUninit, &hMod);
```
m417z: Was it added because of an actual issue, or just in case? Windhawk expects the mod to be unloaded when `Wh_ModUninit()` returns, pinning the mod is unusual and violates that assumption.

If there's an issue, it's preferable to fix it properly. If it's "just in case", I prefer to have it crash - remove the two lines and just leave the log.

PR #4722, opened 2026-07-08
MicroManager small 1.1.0 patch

BlackPaw21:
## Changelog

- **Fixed:** Tooltip now displays correctly when hovering the tray icon.
- **Fixed:** Ghost window prevention — popup no longer flickers on rapid open/close.
- **Fixed:** Data delay recovery — processes that take time to report usage are retried with exponential backoff.
- **Fixed:** Safe mod reload — icon and window clean up properly without crashing.
- **Improved:** Tray tooltip updates only when values change, reducing unnecessary CPU work.
- **Fixed:** Popup window reference cleaned up on destroy.

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/micromanager.wh.cpp
+++ b/mods/micromanager.wh.cpp
@@ -38,7 +38,7 @@ Right-click the tray icon to change the refresh rate (0.3s / 0.5s / 1s / 3s).
 # 1.1.0
 - **Fixed:** Tooltip now displays correctly when hovering the tray icon.
 - **Fixed:** Ghost window prevention — popup no longer flickers on rapid open/close.
-- **Fixed:** Data delays no longer cause missing stats — monitor recovers automatically.
+- **Fixed:** Removed stale exponential backoff in process enumeration retry — STATUS_INFO_LENGTH_MISMATCH only needs a larger buffer.
 - **Fixed:** Safe mod reload — icon and window clean up properly without crashing.
 - **Improved:** Tray tooltip updates only when values change, reducing unnecessary CPU work.
 - **Fixed:** Popup no longer leaves a ghost window behind.
@@ -251,7 +251,6 @@ static int CollectProcessInfo(MY_SYSTEM_PROCESS_INFO** outBuf) {
 
     *outBuf = nullptr;
 
-    DWORD retryDelay = 50;
     for (int retry = 0; retry < 5; retry++) {
         ULONG bufSize = PROCESS_BUF_SIZE * (retry + 2);
         MY_SYSTEM_PROCESS_INFO* newBuf = (MY_SYSTEM_PROCESS_INFO*)realloc(*outBuf, bufSize);
@@ -266,8 +265,6 @@ static int CollectProcessInfo(MY_SYSTEM_PROCESS_INFO** outBuf) {
             }
             return 1;
         }
-        Sleep(retryDelay);
-        retryDelay *= 2;
     }
     // All 5 attempts failed with STATUS_INFO_LENGTH_MISMATCH
     free(*outBuf);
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

---

**The `Sleep`/exponential backoff in `CollectProcessInfo` is misapplied to a buffer-size error.** `NtQuerySystemInformation(SystemProcessInformation, ...)` returns `STATUS_INFO_LENGTH_MISMATCH` only when the supplied buffer is too small — it's a size problem, not a timing problem. Growing the buffer (which the loop already does via `PROCESS_BUF_SIZE * (retry + 2)`) is what resolves it; sleeping does nothing but add latency. Concretely, on a heavily-threaded system where the first 1 MB buffer is insufficient, the loop now sleeps 50 ms (then 100, 200, …) *on the tray thread* before retrying with a larger buffer — and this repeats every refresh tick. Since the tray thread also dispatches the icon's click/hover callbacks and the popup's `WM_ACTIVATE` auto-hide, that's a per-tick stall in normal UI handling. Drop the `Sleep`/`retryDelay` and just retry immediately with the bigger buffer:

```cpp
for (int retry = 0; retry < 5; retry++) {
    ULONG bufSize = PROCESS_BUF_SIZE * (retry + 2);
    MY_SYSTEM_PROCESS_INFO* newBuf = (MY_SYSTEM_PROCESS_INFO*)realloc(*outBuf, bufSize);
    if (!newBuf) { free(*outBuf); *outBuf = nullptr; return 0; }
    *outBuf = newBuf;
    NTSTATUS status = NtQuery(SystemProcessInformation, *outBuf, bufSize, nullptr);
    if (status != STATUS_INFO_LENGTH_MISMATCH) {
        if (status < 0) { free(*outBuf); *outBuf = nullptr; return 0; }
        return 1;
    }
    // buffer was too small — loop and grow, no sleep needed
}
```

Relatedly, the changelog line _"Data delay recovery — processes that take time to report usage are retried with exponential backoff"_ describes a behavior the code doesn't implement (there's no "usage reporting delay" behind `STATUS_INFO_LENGTH_MISMATCH`) — worth correcting so the changelog matches reality. If this loop was AI-generated, this is exactly the kind of plausible-but-wrong rationale to double-check.

<p>

- **Changelog heading level.** The new entries use `# 1.1.0` / `# 1.0.0` (H1), which render larger than the `## Changelog` (H2) section they sit under. The original `### v1.0.0` (H3) reads better; consider keeping `###`.
- **`InterlockedExchangePointer` on `g_popupHwnd`** (in the popup's `WM_DESTROY`) is unnecessary — `g_popupHwnd` is only ever read/written on the tray thread (`ShowPopup`, `WM_LBUTTONUP`, `WM_CLOSE`, `RefreshData`), so a plain `g_popupHwnd = nullptr;` is equivalent. Harmless as-is.

</p>
</details>

RajStartsHere:
_Greetings_,
@BlackPaw21

As I've Been Going Busy And Unavailable From Sometimes And Will Be Further, I'm Finding It Comfortable To Make The Request Here...

The Compatibility With The Mod [Taskbar On Top For Windows 11](https://www.windhawk.net/mods/taskbar-on-top) (Or Similar, Depends On The Position) Can Be Implement By Providing The Option To Toggle Its Position, Changing The Animations, Motion, And The Orientation Relevant To The Options Named: Left, Right, Up, And Default (Down)

As Of Now, The Only Visible Part Is The Shadow Of The Container, The Container Itself Is Being Rendered Out Of The Canvas (Desktop Area) When The Incompatible Mod Is Toggled On

_Hope This Helps!_

BlackPaw21:
@RajStartsHere 
wow, I actually did not even think about the taskbar position at all, thank you a LOT for bringing these things to my attention so they can be patched🤘

PR #4723, opened 2026-07-08
Micswitch v2.0.1

BlackPaw21:
## Changelog

- **New:** Persistent Mute toggle in dashboard settings. When checked, mute state survives device cycles (USB replug, sleep/wake, driver resets).
- **Fixed:** Persistent mute now survives device cycling — mute is re-applied to the newly defaulted device after switching.
- **Fixed:** Tray icon red-dot overlay updates after persistent re-muting.
- **Fixed:** "Persistent Mute" checkbox text now renders correctly on the dark dashboard.

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini


m417z:
### Submission review

Note: I haven't verified the review. Address it or let me know if it's not correct.

---

The main new feature (Persistent Mute) has a state-tracking bug that leaves a microphone muted behind the user's back — please fix that before merge.

**Persistent-mute re-apply never moves the mute tracking to the new device — the active mic can stay muted after the mod is disabled.** Both re-apply paths mute the *new* default endpoint with `ApplyMute(nullptr, TRUE)` but never update `g_mutedDeviceId` (or the persisted `MutedDeviceId`) and never unmute the previously-muted device:

- `CycleAudioDevice` — [`microswap.wh.cpp:2019-2024`](https://github.com/ramensoftware/windhawk-mods/blob/77fec31d056d3ac2109706fbd3d4fa911ba75a9a/mods/microswap.wh.cpp#L2019-L2024) (the `keepMute` branch also skips `RestoreMute()`, so the old device stays muted).
- `TrayWndProc` / `WM_REBIND_VOLUME_CALLBACK` — [`microswap.wh.cpp:2675-2679`](https://github.com/ramensoftware/windhawk-mods/blob/77fec31d056d3ac2109706fbd3d4fa911ba75a9a/mods/microswap.wh.cpp#L2675-L2679) (this is the USB-replug / sleep-wake path — the headline scenario).

  Concrete failure: mute device A (default), then cycle/replug to device B. Now **both** A and B are muted, but `g_mutedDeviceId` still points at A. When the mod is disabled, `RestoreMuteExternal()` unmutes A and leaves **B — the user's current microphone — muted**, with no indication the mod did it. Because `MutedDeviceId` in storage is also stale, the next mod launch unmutes A again and B stays muted across restarts. During normal use, every cycled device also accumulates a stuck mute that only the *current* default ever gets cleared. This breaks Windhawk's reversibility principle (a mod's effects must disappear when it's disabled).

  Fix: whenever you re-apply the mute to the new default, unmute the previously-tracked device and record the new one, so exactly one device is ever muted-by-us. Roughly:
  ```cpp
  if (keepMute) {
      WCHAR prevId[512];
      { EnterCriticalSection(&g_stateLock); lstrcpynW(prevId, g_mutedDeviceId, 512); LeaveCriticalSection(&g_stateLock); }
      if (prevId[0]) ApplyMute(prevId, FALSE);   // release the old device

      IMMDevice* pDef = nullptr;
      if (SUCCEEDED(pEnum->GetDefaultAudioEndpoint(eCapture, eMultimedia, &pDef))) {
          LPWSTR pId = nullptr;
          if (SUCCEEDED(pDef->GetId(&pId))) {
              ApplyMute(pId, TRUE);
              EnterCriticalSection(&g_stateLock);
              lstrcpynW(g_mutedDeviceId, pId, 512); g_isMutedByUs = true;
              LeaveCriticalSection(&g_stateLock);
              Wh_SetStringValue(L"MutedDeviceId", pId);
              CoTaskMemFree(pId);
          }
          pDef->Release();
      }
  }
  ```
  Apply the equivalent in the `WM_REBIND_VOLUME_CALLBACK` handler.

<p>

- **`g_initComplete` reload guard looks inert in the tool-mod pattern.** [`WhTool_ModInit`](https://github.com/ramensoftware/windhawk-mods/blob/77fec31d056d3ac2109706fbd3d4fa911ba75a9a/mods/microswap.wh.cpp#L2806-L2808) / [`WhTool_ModUninit`](https://github.com/ramensoftware/windhawk-mods/blob/77fec31d056d3ac2109706fbd3d4fa911ba75a9a/mods/microswap.wh.cpp#L2779-L2781) each run exactly once per dedicated process (`Wh_ModUninit` calls `WhTool_ModUninit()` then `ExitProcess(0)`), so the "prevent double-init/double-uninit on DLL reload" guard can never actually fire. Harmless, but it's dead complexity you can drop.

- **`g_persistentMute` / `g_isMutedByUs` read without the lock** in `WM_REBIND_VOLUME_CALLBACK` ([`microswap.wh.cpp:2676`](https://github.com/ramensoftware/windhawk-mods/blob/77fec31d056d3ac2109706fbd3d4fa911ba75a9a/mods/microswap.wh.cpp#L2676)). `g_isMutedByUs` is written under `g_stateLock` from the worker thread, so this is a technical data race — benign for a bool, but for consistency either take the lock (as `CycleAudioDevice` does) or make these `std::atomic<bool>`.

</p>
</details>

<p>

- **Double mute is applied on a manual cycle.** `CycleAudioDevice` calls `ApplyMute(nullptr, TRUE)` after `SetDefaultEndpoint`, and the resulting `OnDefaultDeviceChanged` → `WM_REBIND_VOLUME_CALLBACK` re-applies it again. Idempotent, so no harm, but the re-mute logic effectively runs twice for the same switch. If you centralize the "move mute to new default" logic (per the main fix), consider driving it from a single path to avoid the redundant COM round-trip.

</p>
</details>

m417z:
### Submission review

There's one functionality note, let me know if it's worth addressing.

---

**Persistent‑mute re‑application is duplicated across two threads and races.** When the mod cycles a device, `CycleAudioDevice` (worker thread) calls `pPolicyConfig->SetDefaultEndpoint(...)` for `eConsole`/`eMultimedia`/`eCommunications` (`microswap.wh.cpp:2013`), which fires `OnDefaultDeviceChanged(eCapture, eMultimedia, …)` → posts `WM_REBIND_VOLUME_CALLBACK` to the tray thread. So the *exact same* "unmute the old tracked device, then mute the new default and update `g_mutedDeviceId`" sequence now runs in **both** places concurrently:

- worker thread: `microswap.wh.cpp:2019`–`2046`
- tray thread (`WM_REBIND_VOLUME_CALLBACK`): `microswap.wh.cpp:2698`–`2724`

Because each path independently reads `g_mutedDeviceId`, unmutes it, then re‑mutes the current default, an unlucky interleaving unmutes the *newly‑active* mic and re‑mutes it a moment later (e.g. worker sets `g_mutedDeviceId = B` and mutes B; the tray handler then reads `old = B`, unmutes B, and re‑mutes B). The final state always converges to "muted" so it doesn't get stuck, but for a *mute* feature a transient live‑mic window on every device cycle isn't ideal, and it's redundant work plus a data race on `g_mutedDeviceId`.

Since `SetDefaultEndpoint` already drives `WM_REBIND_VOLUME_CALLBACK`, make that handler the single owner of the mute re‑application and drop the duplicated block from `CycleAudioDevice` (lines 2019–2046). `WM_REBIND` already handles the external‑default‑change case too (unplug / user changes default in Sound settings), so one path covers both triggers.

---

<p>

- **`g_initComplete` double‑init/uninit guard looks unnecessary.** In the tool‑mod pattern the launcher calls `WhTool_ModInit` exactly once for the dedicated process and `WhTool_ModUninit` once (immediately followed by `ExitProcess(0)`); there's no in‑place DLL reload that would re‑enter these. The guard (`microswap.wh.cpp:2854`, `2946`) is harmless but guards a scenario that doesn't occur here — worth removing unless you've actually observed a double call.

- **`RIDEV_REMOVE` before registering raw input** (`microswap.wh.cpp:2818`) clears a prior mouse raw‑input registration — but each tool‑mod instance is a fresh `windhawk.exe` process with no prior registration to clear, so this is a no‑op. Fine to keep, but it reads as defensive code for a non‑existent state.

- **Unlocked read of `g_persistentMute`/`g_isMutedByUs`** at `microswap.wh.cpp:2709` (and the `g_persistentMute` write at `2731`), whereas `CycleAudioDevice` reads them under `g_stateLock` (`microswap.wh.cpp:1966`). It's a benign `bool` race today, and consolidating the mute logic per the item above makes it moot.

- **Changelog headings changed from `### v2.0.1` to `# 2.0.1`** (H1). H1 renders very large on the mod page; `###` (as before) reads better. Purely cosmetic.

</p>
</details>

BlackPaw21:
my AI decided he is running the PR for me. hoping for the best on that one ngl.
fingers crossed he did not mess anything up.

m417z:
### Submission review

Looks good. I haven't verified the main note below. LMK if you're interested in addressing it.

---

**`WM_REBIND_VOLUME_CALLBACK` unmutes the tracked device even when Persistent Mute is off, leaving a stale "muted" indicator.** In `TrayWndProc` (around lines 2669-2695) the "unmute the previously-tracked device" step runs unconditionally, *before* the `if (g_persistentMute && g_isMutedByUs)` gate:

```cpp
// runs on EVERY default-device change, regardless of Persistent Mute
EnterCriticalSection(&g_stateLock);
WCHAR oldMutedId[512] = {};
lstrcpynW(oldMutedId, g_mutedDeviceId, 512);
LeaveCriticalSection(&g_stateLock);
if (oldMutedId[0]) {
    ApplyMute(oldMutedId, FALSE);
}

if (g_persistentMute && g_isMutedByUs) { /* mute new default ... */ }
```

Repro: Scroll-to-Swap mode, Persistent Mute **off**, left-click to mute the current mic (`g_isMutedByUs = true`, `g_mutedDeviceId = A`). Now the default capture device changes **externally** (e.g. you plug in a headset that becomes the default mic). `OnDefaultDeviceChanged` posts `WM_REBIND_VOLUME_CALLBACK`, which unmutes device A - but since Persistent Mute is off it does **not** clear `g_isMutedByUs`/`g_mutedDeviceId`. The tray icon keeps showing the red "muted" dot (the overlay is driven by `g_isMutedByUs`, see `UpdateTrayTip` line 1761) while the mic is actually live. For a mute feature, "shows muted while hot" is the dangerous direction.

Also note `g_persistentMute` and `g_isMutedByUs` are read here without holding `g_stateLock` (line 2680), while `CycleAudioDevice` reads them under the lock.

Fix: gate the whole transfer on a single locked snapshot, so nothing touches the mute state when Persistent Mute is off:

```cpp
bool persistent, muted;
WCHAR oldMutedId[512] = {};
EnterCriticalSection(&g_stateLock);
persistent = g_persistentMute;
muted      = g_isMutedByUs;
lstrcpynW(oldMutedId, g_mutedDeviceId, 512);
LeaveCriticalSection(&g_stateLock);

if (persistent && muted) {
    if (oldMutedId[0]) ApplyMute(oldMutedId, FALSE);
    // ... get new default, ApplyMute(newDevId, TRUE), update g_mutedDeviceId ...
    PostMessageW(hWnd, WM_UPDATE_TRAY_STATE, 0, 0);
}
```

<p>

Minor polish - none of this affects users, so it's your call.

- **`g_persistentMute` is written without the lock but read under it.** The writes at line 2702 (`WM_RELOAD_ALL`) and line 2894 (`WhTool_ModInit`) are unlocked, while `CycleAudioDevice` (line 1966) reads it inside `g_stateLock`, so the lock doesn't actually synchronize the worker thread against the tray thread. Either write it under `g_stateLock` too, or make it a `std::atomic<bool>`.

- **The `g_initComplete` double-init/uninit guard looks unnecessary.** In the tool-mod model `WhTool_ModInit` runs exactly once per dedicated `windhawk.exe` process, and `Wh_ModUninit` calls `ExitProcess(0)` right after `WhTool_ModUninit`, so there is no in-process reload that could double-init. The `InterlockedCompareExchange` guards are harmless but never fire - was this intentional, or leftover?

- **Magic `6` vs `MAX_DEVICE_SLOTS` in `WM_REFRESH_DEVICE_LIST`.** The slot loops use a literal `6` (`std::vector<int> savedSlotSel(6, -1)`, `for (int i = 0; i < 6; i++)`) while `slots[6]` is really `MAX_DEVICE_SLOTS`. Using the constant keeps them in sync if the slot count ever changes.

</p>
</details>

<p>

Non-critical observation about the feature behavior itself.

- The red-dot overlay tracks the global `g_isMutedByUs` flag rather than the mute state of the *current* default device, so after any external default-device switch the indicator can drift out of sync with reality even aside from the bug above. A more robust design would derive the overlay from a live `IAudioEndpointVolume::GetMute()` query on the current default endpoint (which you already bind in `BindEndpointVolume`) instead of a cached bool - that would also make the fix above self-correcting.

</p>
</details>

BlackPaw21:
i think that should do it, @m417z thank you once again for helping and maintaining the app :)

PR #4726, opened 2026-07-08
Add Power Options Button v1.0.0

SilverAmd:
Adds a tray icon for switching between Windows power plans.

Features:
- Shows the active power plan in the system tray.
- Allows switching plans with the mouse wheel over the tray icon.
- Provides a right-click menu for manual power plan selection.
- Supports Power Saver, Balanced, High Performance, and Ultimate Performance.
- Can create the Ultimate Performance plan with administrator approval.
- Restores the tray icon after Explorer/taskbar restart.

The mod runs as a Windhawk tool mod in a dedicated windhawk.exe process.

## Changelog

* Changelog item 1...
* Changelog item 2...

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [x ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/power-options-button.wh.cpp
+++ b/mods/power-options-button.wh.cpp
@@ -1,0 +1,1711 @@
+// ==WindhawkMod==
+// @id              power-options-button
+// @name            Power Options Button
+// @description     Tray icon to switch between Windows power plans with the mouse wheel.
+// @version         1.0.0
+// @author          SilverAmd
+// @github          https://github.com/SilverAmd
+// @license         MIT
+// @include         windhawk.exe
+// @compilerOptions -luser32 -lshell32 -lgdi32 -lpowrprof
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Power Options Button
+
+Adds a tray icon for switching between Windows power plans.
+
+## Preview
+
+![Power Options Button preview](https://i.imgur.com/26A3Ilv.gif)
+
+*Shows the tray icon, tooltip, right-click menu, and mouse-wheel switching between power plans.*
+
+## Features
+
+- Tray icon shows the currently active Windows power plan.
+- Tooltip shows the active plan.
+- Mouse wheel over the tray icon cycles between:
+  - Power Saver
+  - Balanced
+  - High Performance
+  - Ultimate Performance
+- Right-click opens a context menu.
+- Icon color changes depending on the active plan:
+  - Green = Power Saver
+  - Yellow = Balanced
+  - Red = High Performance
+  - Violet = Ultimate Performance
+- If Ultimate Performance is not available, it is shown as disabled and can be created from the tray menu with administrator approval.
+- Automatically restores the tray icon after Explorer/taskbar restart.
+
+## Usage
+
+- Scroll up over the tray icon: next power plan.
+- Scroll down over the tray icon: previous power plan.
+- Right-click the tray icon: choose a power plan manually.
+- Left-click the tray icon: refresh current status.
+- Use "Create Ultimate Performance..." from the tray menu to create the Ultimate Performance plan if it is not available yet.
+*/
+// ==/WindhawkModReadme==
+
+// Important:
+// After Windows startup, the low-level mouse hook can appear installed but not
+// receive wheel events over the tray icon until it is refreshed.
+// Refreshing the hook once on the first tray WM_MOUSEMOVE makes mouse-wheel
+// switching work immediately after reboot.
+
+#include <windows.h>
+#include <shellapi.h>
+#include <powrprof.h>
+#include <strsafe.h>
+
+#ifndef NIF_SHOWTIP
+#define NIF_SHOWTIP 0x00000080
+#endif
+
+#ifndef NIN_SELECT
+#define NIN_SELECT (WM_USER + 0)
+#endif
+
+#ifndef NIN_KEYSELECT
+#define NIN_KEYSELECT (NIN_SELECT | 0x1)
+#endif
+
+#define WM_TRAYICON        (WM_USER + 1)
+#define WM_TRAY_SCROLL     (WM_APP + 1)
+#define WM_APP_EXIT        (WM_APP + 2)
+
+#define ID_TRAYICON               24001
+#define ID_MENU_POWER_SAVER       3001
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

---

**1. Multi-second blocking calls run on the same thread as the `WH_MOUSE_LL` hook, so they stall system-wide input.** The low-level mouse hook is installed by the UI thread (`InstallMouseWheelHook` is called from `UiThreadProc`/the window proc), and a low-level hook callback is dispatched on the thread that installed it — the same thread that runs `HiddenWndProc`. Several paths on that thread block for seconds while inside a message handler:
- `CreateUltimatePerformancePlan` → `WaitForSingleObject(sei.hProcess, 10000)` + `Sleep(500)` (line 541, 553)
- `SetPowerPlanWithPowercfg` → `WaitForSingleObject(pi.hProcess, 5000)` (line 414)

While the UI thread is blocked in one of these it isn't pumping messages, so the mouse hook can't service events. Windows then applies the `LowLevelHooksTimeout` (~300 ms) to *every* wheel event during that window — and can silently drop the hook from the chain if it keeps timing out — so all mouse-wheel scrolling across the system lags for the duration. The triggers are infrequent (the elevated create, and the rare `powercfg` fallback when `PowerSetActiveScheme` fails), but the blast radius is the whole desktop. Run the child-process launch + wait on a dedicated worker thread and post the result back to the UI thread, so the hook thread never blocks.

**2. `Wh_ModAfterInit` deviates from the verbatim wiki snippet:** it uses `wsprintfW` where the [wiki boilerplate](https://github.com/ramensoftware/windhawk/wiki/Mods-as-tools:-Running-mods-in-a-dedicated-process) uses `swprintf_s`. Keeping the snippet verbatim makes future maintenance easier.

<p>

- **The mouse-hook "refresh" machinery looks like a workaround for the wrong cause and can likely be deleted.** `RefreshMouseWheelHook`, the "first tray icon hover" refresh, the post-start-sync timer's re-hook, and the globals `g_lastMouseHookRefresh` / `g_mouseHookRefreshedAfterFirstHover` / `g_mouseHookInstalledAt` all exist to fix the "hook installed but no wheel events after boot" symptom described in the header comment. But reinstalling the hook doesn't change whether the *tray rect* is valid — the actual reason wheel events are ignored early after boot is that `Shell_NotifyIconGetRect` returns an empty/stale rect until the tray settles, which you already handle via the rect-refresh timer and the `g_lastTrayMouseMoveTime` fallback. Installing the hook once (as [mic-tray-control](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/mic-tray-control.wh.cpp#L598) does) and dropping the refresh dance should behave identically with much less code.
- **Debug scaffolding can be removed:** `DEBUG_MOUSE_WHEEL` and the `#if`-guarded blocks, plus `g_lastWheelDebugLog` / `g_lastTrayMouseMoveLog` (and `g_mouseHookInstalledAt`, which is set but never read) are dead in production.
- **Use the cached tray rect inside the hook instead of a synchronous shell round-trip per event.** `IsPointNearTrayIcon` calls `RefreshTrayIconRect` → `Shell_NotifyIconGetRect` on every system-wide wheel event. You already keep `g_trayIconRect` fresh from the timer and `WM_MOUSEMOVE`, so the hook can just read it — avoiding a cross-process call on the input path. (This is the same pattern as the previous point.)
- **`powercfg.exe` / `control.exe` are launched by bare name.** `CreateProcessW(nullptr, L"powercfg.exe …")` (line 392) with a NULL application name searches the current directory before System32, and the `ShellExecute*` calls are similar. The mod only runs inside the trusted `windhawk.exe`, so the hijack surface is small — but passing the full `%SystemRoot%\System32\powercfg.exe` path is a cheap hardening win.
- Minor: the fallback paths in `CreatePowerTrayIcon` return `LoadIconW(nullptr, IDI_APPLICATION)` (a shared icon), which later gets `DestroyIcon`'d in `UpdateTrayIcon`/`RemoveTrayIcon` — shared icons shouldn't be destroyed. Harmless in practice, and only on the rare icon-creation-failure path.
- `ShowContextMenu` sets `menuText = L"Ultimate Performance"` for `i == 3` (line 1137), which is already `g_plans[3].name` — redundant.

</p>
</details>

<p>

- **Ultimate Performance detection by friendly name is English/German-only.** A user-duplicated Ultimate plan gets a new random GUID, so `RefreshUltimatePerformanceAvailability` falls back to matching the localized name against `L"Ultimate"` / `L"Ultimativ"` (line 313-315). On other UI languages (French, Spanish, …) a created Ultimate plan won't be detected, and the menu will keep offering "Create…". Consider capturing the new GUID from `powercfg -duplicatescheme` output, or enumerating schemes before/after creation to find the added one, rather than hardcoding two languages.
- **The tray shows "Balanced" whenever the active plan isn't one of the four built-ins.** `GetCurrentPowerPlanIndex` returns the last known `g_currentPlanIndex` (default 1 = Balanced) when the active scheme is a custom/unknown plan, so the icon/tooltip can misreport the real state. Not fixable in general (the mod only knows four plans), just noting the limitation.

</p>
</details>

SilverAmd:
Updated the mod to v1.0.13.

Addressed the required review items:
- Moved multi-second powercfg / Ultimate Performance operations to worker threads and post results back to the UI thread.
- Changed Wh_ModAfterInit to use swprintf_s.

Also addressed several optional items:
- Removed debug scaffolding.
- Use cached tray rect in the mouse hook.
- Removed post-start hook refresh, kept only first-hover refresh because it was needed in testing after reboot.
- Use full System32/Sysnative paths for powercfg.exe and control.exe.
- Fixed shared icon fallback by copying the icon before destroying.
- Removed redundant Ultimate Performance menu text handling.
- Improved Ultimate Performance detection by capturing newly created GUIDs.
- Custom/unknown power plans now show a ? tray icon and the real active custom plan name in the tooltip.
- Added automatic refresh for externally changed power plans.
- Added minor safety/polish fixes for RegisterClassW, UI-thread uninit logging, and fallback result bounds checking.

Tested:
- Mouse wheel switching works after reboot.
- Power Saver / Balanced / High Performance / Ultimate Performance switching works.
- Custom power plan detection works.
- Tooltip correctly shows the custom power plan name.
- Classic Power Options, Windows power settings, Windhawk menu items and Refresh work.

PR #4740, opened 2026-07-10
Add Desktop Audio Visualizer 1.0.0

Salyts:
A real-time audio spectrum visualizer that displays on your Windows desktop. Captures system audio output and renders animated frequency bars directly on the desktop background.

## Mod authorship

This mod was created by:

- - [x] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/desktop-audio-visualizer.wh.cpp
+++ b/mods/desktop-audio-visualizer.wh.cpp
@@ -1,5 +1,5 @@
 // ==WindhawkMod==
-// @id                 desktop-audio-visualizer
+// @id                 desktop-audio-visualizer1
 // @name               Desktop Audio Visualizer
 // @description        Real-time audio visualizer on your Windows desktop with customizable appearance
 // @description:ru-RU  Аудиовизуализатор в реальном времени на рабочем столе Windows с настраиваемым внешним видом
@@ -56,7 +56,7 @@ A real-time audio spectrum visualizer that displays on your Windows desktop. Cap
 * **Adjustable Sensitivity**: 0–300 range
 
 ### Performance
-* **Adjustable Frame Rate**: Higher = smoother. 60 is good for most displays, 120 for high refresh rate monitors.
+* **Adjustable Frame Rate**: Set any FPS value for maximum smoothness or power-saving rendering
 * **Fullscreen Pause**: Detects both exclusive fullscreen (DX9/11) and borderless fullscreen windows; pauses automatically and resumes when the app closes
 * **Silence Detection**: Reduces refresh rate after a configurable idle period to save CPU/GPU
 
@@ -64,7 +64,7 @@ A real-time audio spectrum visualizer that displays on your Windows desktop. Cap
 
 ### Credits
 * **[Salyts](https://github.com/Salyts)** — Author of Desktop Audio Visualizer
-* **[GR0UD](https://github.com/GR0UD)** — Audio capture and FFT engine
+* Borrowed the audio visualizer code from **[GR0UD](https://github.com/GR0UD)**.
 
 */
 // ==/WindhawkModReadme==
@@ -266,7 +266,6 @@ A real-time audio spectrum visualizer that displays on your Windows desktop. Cap
 #include <audioclient.h>
 #include <wrl/client.h>
 #include <wincodec.h>
-#include <mutex>
 
 #include <algorithm>
 #include <atomic>
@@ -283,8 +282,6 @@ using namespace winrt::Windows::Storage::Streams;
 
 using Microsoft::WRL::ComPtr;
 
-#define TIMER_ID_OVERLAY_REFRESH 1
-
 #define TIMER_ID_MSG_DISPLAY_CHANGE 1
 #define TIMER_ID_MSG_RECREATE_OVERLAY 2
 #define TIMER_ID_MSG_WALLPAPER_REFRESH 3
@@ -292,6 +289,7 @@ using Microsoft::WRL::ComPtr;
 
 #define WM_APP_CLEANUP (WM_APP + 1)
 #define WM_APP_SETTINGS_CHANGED (WM_APP + 2)
+#define WM_APP_RENDER_TICK (WM_APP + 3)
 #define OVERLAY_WINDOW_CLASS (L"DesktopAudioVisOverlay_" WH_MOD_ID)
 #define MESSAGE_WINDOW_CLASS (L"DesktopAudioVisMessage_" WH_MOD_ID)
 
@@ -354,7 +352,7 @@ ComPtr<ID2D1Device> g_d2dDevice;
 
 HWND g_messageWnd;
 
-HWND g_overlayWnd;
+std::atomic<HWND> g_overlayWnd{nullptr};
 ComPtr<IDXGISwapChain1> g_swapChain;
 ComPtr<ID2D1DeviceContext> g_dc;
 ComPtr<IDCompositionDevice> g_compositionDevice;
@@ -378,6 +376,10 @@ std::thread* g_captureThread = nullptr;
 HANDLE g_captureEvent = nullptr;
 std::atomic<bool> g_deviceChanged{false};
 std::atomic<float> g_bands[VIZ_NUM_BANDS] = {};
+
+std::thread* g_renderThread = nullptr;
+std::atomic<bool> g_renderThreadRunning{false};
+std::atomic<bool> g_renderTickPending{false};
 
 float g_hannWindow[VIZ_FFT_SIZE] = {};
 float g_twiddleRe[VIZ_FFT_SIZE / 2] = {};
@@ -410,8 +412,8 @@ static std::atomic<bool>  g_albumArtFetchPending{false};
 static std::atomic<DWORD> g_accentColorCache{0xFF0078D4};
 
 static HANDLE g_gsmtcStopEvent = nullptr;
-static std::thread g_gsmtcThread;
-static HANDLE g_albumArtDoneEvent = nullptr;
+[[clang::no_destroy]] static std::thread g_gsmtcThread;
+static std::thread* g_albumArtThread = nullptr;
 
 using RunFromWindowThreadProc_t = void(WINAPI*)(void* parameter);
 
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

---

**Detached background threads are never joined on unload → crash on disable / update / reload.** `InitGsmtcListener` and `FetchAlbumArtColorAsync` both spawn `std::thread(...).detach()` threads, and nothing ever waits for them:

- The GSMTC listener thread spends essentially all its time parked in `Sleep(200)` inside `while (!g_unloading.load()) Sleep(200);` (line 828), with a return address that points into the mod's DLL image.
- `Wh_ModUninit` sets `g_unloading = true` and returns immediately (line 2387) — it does not wait for either thread.

When Windhawk unloads the mod, it frees the DLL as soon as `Wh_ModUninit` returns. `Wh_ModUninit` is fast, so the GSMTC thread is almost always still inside `Sleep(200)` (and afterwards runs its winrt-cleanup code, which also lives in the DLL) when the image is unmapped → the thread returns into freed memory and crashes explorer. `FetchAlbumArtColorAsync`'s thread has the same problem if a fetch is in flight (WIC decode, `winrt` calls, etc.). Because a mod update or an occasional settings-triggered reload both unload the DLL, this will bite users during ordinary operation, not just on manual disable.

There's a second, related hazard: the GSMTC event tokens on `g_gsmtcMgr` / `g_gsmtcSession` are only revoked from inside that detached thread after it notices `g_unloading`. If a media-session change fires the registered lambda after the DLL is gone, that's also a crash.

Fix: make these threads joinable and join them in `Wh_ModUninit` before returning, the same way the base mod handles its weather thread — see [`WeatherUpdateThreadUninit`](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/desktop-live-overlay.wh.cpp#L882-L890), which signals a stop event and `WaitForSingleObject(..., INFINITE)` before the DLL can unload. Keep a `std::thread`/`HANDLE` global for the GSMTC listener (signal via an event instead of the `Sleep(200)` poll, then join), revoke the winrt tokens, and either track the in-flight album-art thread and join it too, or gate it so it cannot outlive uninit.

- **`InitGsmtcListener` runs unconditionally in every `explorer.exe`.** It's started from `Wh_ModInit` (line 2363) in every injected `explorer.exe` process and runs for the whole session, but it's only needed for the `album_art` color mode and only in the shell process (the one that actually owns the overlay). Consider starting it lazily when album-art mode is selected / the overlay exists, so secondary explorer processes and non-album-art users don't pay for a permanent thread + winrt apartment + media-session subscription.

---

<p>

- **`g_settings` is read from the capture thread without synchronization.** The capture thread reads `g_settings.sensitivity` and `g_settings.eq` (lines 1123, 1126) while `LoadSettings()` rewrites the struct on a settings change. It's a benign data race (a torn `int`/`enum` read at worst produces one glitchy frame of gain), but you could snapshot those two values into atomics if you want it clean.
- **Prefer `WindhawkUtils::StringSetting`** over the manual `Wh_GetStringSetting` + `Wh_FreeStringSetting` pairs in `LoadSettings` — it's the RAII form used by most modern mods.
- **Dead variable**: `bool wasSlow = g_slowMode;` … `(void)wasSlow;` in `ScheduleNextUpdate` (lines 1943, 1948) is computed and discarded — drop it.

</p>
</details>

<p>

- **Target FPS above ~60–64 likely won't be delivered.** Rendering is driven by `WM_TIMER` (`SetTimer(g_overlayWnd, TIMER_ID_OVERLAY_REFRESH, 1000/fps, ...)`, line 1953). `WM_TIMER` is clamped to the system timer tick (~15.6 ms ≈ 64 Hz) unless something in the process has raised the timer resolution, so the "120 for high refresh rate monitors" description in the settings is optimistic. Not a bug — just note the practical ceiling. A dedicated render loop (e.g. a waitable timer / present-driven loop off the UI thread) would be needed for genuinely high frame rates.
- **Rendering + vsync `Present(1, 0)` happen on explorer's desktop UI thread** (the `WorkerW`-owning thread, via the timer). This matches the base mod's architecture, but that mod repaints at a low rate while this one runs at 60 FPS, so the per-frame vsync wait is more likely to add latency to desktop-icon interactions on that thread. There's no clean drop-in alternative within this design, so this is just an FYI; if it proves noticeable, moving the swap-chain present off the UI thread would help.
- **`gradientColor1` default is `#00FFFFFF`** (fully transparent), so the first bar/edge of a static gradient starts invisible. That's only visible if the user switches to gradient mode (default is `solid`), but a fully-opaque default (e.g. `#FF...`) might be less surprising.

</p>
</details>

Salyts:
> ### Submission review
> 
> 
> 
> **Detached background threads are never joined on unload → crash on disable / update / reload.** `InitGsmtcListener` and `FetchAlbumArtColorAsync` both spawn `std::thread(...).detach()` threads, and nothing ever waits for them:
> 
>     * The GSMTC listener thread spends essentially all its time parked in `Sleep(200)` inside `while (!g_unloading.load()) Sleep(200);` (line 828), with a return address that points into the mod's DLL image.
> 
>     * `Wh_ModUninit` sets `g_unloading = true` and returns immediately (line 2387) — it does not wait for either thread.
> 
> 
> When Windhawk unloads the mod, it frees the DLL as soon as `Wh_ModUninit` returns. `Wh_ModUninit` is fast, so the GSMTC thread is almost always still inside `Sleep(200)` (and afterwards runs its winrt-cleanup code, which also lives in the DLL) when the image is unmapped → the thread returns into freed memory and crashes explorer. `FetchAlbumArtColorAsync`'s thread has the same problem if a fetch is in flight (WIC decode, `winrt` calls, etc.). Because a mod update or an occasional settings-triggered reload both unload the DLL, this will bite users during ordinary operation, not just on manual disable.
> 
> There's a second, related hazard: the GSMTC event tokens on `g_gsmtcMgr` / `g_gsmtcSession` are only revoked from inside that detached thread after it notices `g_unloading`. If a media-session change fires the registered lambda after the DLL is gone, that's also a crash.
> 
> Fix: make these threads joinable and join them in `Wh_ModUninit` before returning, the same way the base mod handles its weather thread — see [`WeatherUpdateThreadUninit`](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/desktop-live-overlay.wh.cpp#L882-L890), which signals a stop event and `WaitForSingleObject(..., INFINITE)` before the DLL can unload. Keep a `std::thread`/`HANDLE` global for the GSMTC listener (signal via an event instead of the `Sleep(200)` poll, then join), revoke the winrt tokens, and either track the in-flight album-art thread and join it too, or gate it so it cannot outlive uninit.
> 
>     * **`InitGsmtcListener` runs unconditionally in every `explorer.exe`.** It's started from `Wh_ModInit` (line 2363) in every injected `explorer.exe` process and runs for the whole session, but it's only needed for the `album_art` color mode and only in the shell process (the one that actually owns the overlay). Consider starting it lazily when album-art mode is selected / the overlay exists, so secondary explorer processes and non-album-art users don't pay for a permanent thread + winrt apartment + media-session subscription.
> 
> 
> Optional improvements
> 
> 
> Functionality notes

@m417z

m417z:
### Submission review

---

**1. Global `std::thread g_gsmtcThread` will `std::terminate()` (crash Explorer) on every process shutdown.** `g_gsmtcThread` (line 413) is a full `std::thread` object that is joined **only** in `Wh_ModUninit` (line 2490). But `Wh_ModUninit` is *not* called when Explorer terminates (Explorer restart, sign-out, reboot) — there the OS just detaches the DLL and the CRT runs the destructors of global objects during `DLL_PROCESS_DETACH`. At that point the thread is still `joinable()`, so `~thread()` calls `std::terminate()` and aborts the host. This fires on essentially every graceful Explorer restart / shutdown while the mod is active. Note you already use the correct pattern for the capture thread — `std::thread* g_captureThread` (line 377), a trivial destructor. Do the same for the GSMTC thread (make it a `std::thread*`, or `detach()` it), keeping the real teardown in `Wh_ModUninit`. The maintainer's own [desktop-live-overlay](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/desktop-live-overlay.wh.cpp) — from which this mod's overlay infrastructure is derived — uses a raw `HANDLE g_weatherUpdateThread` for exactly this reason. (The WinRT globals `g_gsmtcMgr` / `g_gsmtcSession` at lines 831–832 are a milder version of the same issue — they hold COM references that get released under the loader lock at `DLL_PROCESS_DETACH`; consider leaving them to be leaked at shutdown too, with real cleanup only in `Wh_ModUninit`.)

Human note: This caused crashes with other mods in the past, so worth fixing. An easier solution is to simply add `[[clang::no_destroy]]` to the global variables.

**2. The album-art fetch thread is fire-and-forget and can outlive the DLL on unload.** `FetchAlbumArtColorAsync` spawns a `std::thread([...]).detach()` (line 826) that reads/writes mod globals and touches mod code. On unload, `Wh_ModUninit` waits at most 3000 ms on `g_albumArtDoneEvent` (line 2499); if the WinRT calls inside it (`RequestAsync().get()`, `OpenReadAsync().get()`, WIC decode) stall past that window — which WinRT calls occasionally do — `Wh_ModUninit` returns, the DLL unmaps, and the still-running detached thread faults when it resumes into freed code. The 3 s wait mitigates the common case but doesn't guarantee it. Track the thread and join it (bounded is fine), or otherwise ensure it cannot touch mod memory after `Wh_ModUninit` returns, rather than relying on a timed best-effort wait on a detached thread.

<p>

- **Use `WindhawkUtils::SetFunctionHook`** (type-safe) instead of the raw `Wh_SetFunctionHook((void*)CreateWindowExW, (void*)CreateWindowExW_Hook, (void**)&CreateWindowExW_Original)` at line 2451.
- **Use `WindhawkUtils::StringSetting`** (RAII) instead of the many raw `Wh_GetStringSetting` + `Wh_FreeStringSetting` pairs in `LoadSettings` — it removes the manual free at each call site.
- **The `Settings` struct member defaults (lines 304–336) diverge from the YAML defaults** (e.g. `barCount` 32 vs 20, `verticalPosition` 88 vs 90, `barWidth` 6 vs 10, `barIdleSize` 4 vs 10). Harmless because `LoadSettings()` always runs first and overwrites them, but the struct defaults are dead/misleading — align them with the settings block or drop them.
- **`CreateWindowExW_Hook`'s `static UINT_PTR s_timer`** (line 2308) is shared across all Explorer threads that create windows, so concurrent FolderView creations race on it. Extremely unlikely to matter in practice, but worth an `std::atomic` or a comment.
- **`RenderVisualizer` calls `GetMonitorById` (→ `EnumDisplayMonitors`) + `GetMonitorInfo` every frame** (lines 1725–1729), i.e. up to 120×/s. Cheap, but the monitor rect only changes on display change — you could cache it and refresh it from `HandleDisplayChange`.

</p>
</details>

<p>

- **The `WM_TIMER`-driven render loop caps smoothness.** `SetTimer` has ~10–16 ms granularity, so "120 FPS" (interval 8 ms) can't actually be reached, and `Present(1, 0)` (vsync) already throttles to the refresh rate, so the timer double-gates the frame pace. For a high-FPS audio visualizer a DXGI waitable swap chain or a `DwmFlush`-paced render thread would give smoother, more predictable output than a message-timer loop (the reference desktop-live-overlay is text at low FPS, so a timer is fine there — less so here).
- **Fullscreen pause is global across monitors.** `IsWindowFullscreen` tests the *foreground* window against *its own* monitor (`MonitorFromWindow`, line 629), so a borderless-fullscreen app/game on any monitor pauses the visualizer even when it's shown on a different monitor. May be intended (save resources whenever anything is fullscreen), just flagging.

</p>
</details>

Salyts:
> Submission review

@m417z

PR #4745, opened 2026-07-11
Add Flexible Explorer Toolbars Deluxe mod

Anixx:
This mod allows the Search Bar, Breadcrumb Bar, and Up Button in Windows Explorer to be movable toolbars, enhancing user customization.

## Changelog

* Changelog item 1...
* Changelog item 2...

## Mod authorship

This mod was created by:

- - [x] Claude
- - [ ] ChatGPT
- - [ ] Gemini
- - [x] Another AI (please specify): Muse Spark

### DIFF
--- a/mods/flexible-explorer-toolbars-deluxe.wh.cpp
+++ b/mods/flexible-explorer-toolbars-deluxe.wh.cpp
@@ -26,10 +26,13 @@
 # Flexible Explorer Toolbars Deluxe
 
 **!Important!** This mod curently only supports Windows 10 or Windows 11 versions up to 23H2 and 24H2/25H2 builds up to 8037. 
-On 24H2 and 25H2 you may have to use vivetool to enable menus: `vivetool /disable /id:55063786`.
+On 24H2 and 25H2 you may have to use vivetool to enable toolbars in Explorer: `vivetool /disable /id:55063786`.
 
 **!Important!** To use this mod, you shoud disable any other mods that hide the clasic Navigation bar, such as the `Disable Navigation Bar` mod by ItsProfessional.
-This mod hides the Navication Bar by itself. It is recommended to install the `Windows 7 Comand Bar` mod.
+This mod hides the Navigation Bar by itself. 
+
+For this mod to work you should enable a mod that restores the Navigation bar, it is recommended to install the `Windows 7 Comand Bar` mod, altough, `Classic Explorer navigation bar`
+also would work if you want to retain elements of Windows 11 fluent interface, as well or any modification that restores ribbon.
 
 This mod hides the Navigation Bar and instead creates the following optional toolbars, which could be freely moved and ordered together with the Menu Bar, if it is enabled:
 
--- a/mods/flexible-explorer-toolbars-deluxe.wh.cpp
+++ b/mods/flexible-explorer-toolbars-deluxe.wh.cpp
@@ -1,12 +1,12 @@
 // ==WindhawkMod==
 // @id              flexible-explorer-toolbars-deluxe
 // @name            Flexible Explorer Toolbars Deluxe
-// @description     Makes Search Bar, Broeadcrumb Bar and others into movable toolbars
-// @version         1.0
+// @description     Makes Search Bar, Breadcrumb Bar and others into movable toolbars
+// @version         1.1
 // @author          Anixx
 // @github          https://github.com/Anixx
 // @include         explorer.exe
-// @compilerOptions -luxtheme -lgdi32 -ladvapi32 -lcomctl32
+// @compilerOptions -luxtheme -lgdi32 -lcomctl32
 // ==/WindhawkMod==
 
 // ==WindhawkModSettings==
@@ -70,7 +70,6 @@ The toolbars can be locked and unlocked.
 
 constexpr int UP_BUTTON_ICON_SIZE  = 16;
 constexpr UINT WM_APP_DO_MOVE      = WM_APP + 0x501;
-constexpr LPCWSTR REG_KEY          = L"Software\\WindhawkMods\\FlexibleExplorer";
 
 struct Settings {
     bool moveSearchBand    = true;
@@ -101,7 +100,6 @@ std::unordered_map<HWND,int> g_childFlags;
 struct ToolbarGuard{int x=0,y=0,cx=0,cy=0;bool hasGood=false;};
 std::unordered_map<HWND,ToolbarGuard> g_toolbarGuards;
 
-// Храним пары HWND → subclass proc для удаления
 std::unordered_map<HWND, WindhawkUtils::WH_SUBCLASSPROC> g_subclassedWindows;
 
 thread_local bool g_insideApply = false;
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

---

**1. The init enumeration isn't filtered to the current process.** In `Wh_ModInit` the top-level scan walks every window on the desktop and matches only by class name:

```cpp
for(HWND w=GetTopWindow(NULL);w;w=GetNextWindow(w,GW_HWNDNEXT)){
    WCHAR cls[64];
    if(GetClassName(w,cls,ARRAYSIZE(cls))&&!wcscmp(cls,L"CabinetWClass")){
        ProcessWindow(w);EnumChildWindows(w,EnumChildHook,0);PostMessage(w,WM_APP_DO_MOVE,0,0);
    }
}
```

`GetTopWindow`/`GetNextWindow` enumerate all top-level windows across **all processes**. With "Launch folder windows in a separate process" enabled (or any second `explorer.exe` hosting a folder window), this process's instance will `ProcessWindow()` a `CabinetWClass` window that belongs to a *different* `explorer.exe`. `SetWindowSubclassFromAnyThread` can't subclass a foreign process (its hook proc / subclass-proc pointers live in this image), so subclassing fails — but the non-subclass side effects still fire cross-process (e.g. `ProcessWindow` calls `SetWindowPos(..., SWP_HIDEWINDOW)` on the other process's nav `WorkerW`, hiding another window's nav bar). Filter by PID:

```cpp
DWORD pid=0; GetWindowThreadProcessId(w,&pid);
if(pid==GetCurrentProcessId() && GetClassName(...)==L"CabinetWClass") { ... }
```

(The `CreateWindowExW` hook path is fine — it only sees windows this process creates.)

**2. Disabling the mod leaves Explorer in a broken state.** `Wh_ModUninit` only removes the subclasses:

```cpp
void Wh_ModUninit(){
    ...
    for(auto& pair : windowsToClean)
        if(IsWindow(pair.first))
            WindhawkUtils::RemoveWindowSubclassFromAnyThread(pair.first, pair.second);
    ...
}
```

It does **not** move the search/breadcrumb/up bands back to the nav rebar, un-hide the force-hidden nav `WorkerW`, un-neuter the Address Band Root, or restore the expanded `ShellTabWindowClass`. So after the user disables the mod, they're left with a hidden navigation bar and no address bar until they restart Explorer. Reversibility ("effects disappear when the mod is disabled") is a core Windhawk principle. A full restore is admittedly involved, but at minimum the teardown should un-hide the nav `WorkerW` and move the bands back so the window is usable; if a complete restore genuinely isn't feasible, state the "restart Explorer to revert" requirement prominently in the README.

**3. `WM_APP_DO_MOVE` is a fixed `WM_APP`-relative constant sent to windows the mod doesn't own.** `WM_APP_DO_MOVE = WM_APP + 0x501` is posted to / handled on the `CabinetWClass` window (which belongs to Explorer, not the mod), and the handler swallows it (`return 0;` without chaining to `DefSubclassProc`):

```cpp
if(msg==WM_APP_DO_MOVE){DoMoveSearchBandToMenuBar(hwnd);return 0;}
```

`WM_APP+N` values are defined by the application (Explorer) for its own windows, so this can collide two ways: an Explorer-internal `WM_APP+0x501` sent to the cabinet would be hijacked into a spurious move, and your posts could be misread by other subclassers. Register a process-unique message instead:

```cpp
UINT g_msgDoMove = RegisterWindowMessage(L"FlexibleExplorerToolbars_DoMove");
```

and compare against that. `RegisterWindowMessage` returns a value in the `0xC000–0xFFFF` range that can't collide with a class's own `WM_APP`/`WM_USER` messages.

**4. `NtSetValueKey_Hook` drops writes with a NULL value name.**

```cpp
if (!ValueName||UnicodeStringEqualsIgnoreCase(ValueName, L"ITBar7Layout")) {return 0;}
```

The `!ValueName` branch returns `STATUS_SUCCESS` **without calling the original** — i.e. it silently discards the write (setting a key's default/unnamed value) and reports success. `UnicodeStringEqualsIgnoreCase` already null-checks its argument, so the guard is both redundant and wrong. It's low-probability, but this is a process-wide syscall hook, so a stray dropped write is undebuggable. Drop the `!ValueName ||` clause so a null name falls through to the original.

<p>

- **Unused libraries.** `@compilerOptions -luxtheme -lgdi32 -lcomctl32` — there's no uxtheme or GDI usage in the mod; `-luxtheme` and `-lgdi32` can be dropped. (`-lcomctl32` is needed for `DefSubclassProc`.)
- **Unbounded `wsprintf` into fixed buffers.** `wsprintf(ck,L"Cx_%s",cls)` / `Break_%s` write into `WCHAR[128]` from a `cls` buffer of up to 256 chars. Real class names here are short, so it won't overflow in practice, but `wsprintf` is unbounded — prefer `swprintf_s`/`StringCchPrintf` with the buffer size.
- **Settings changes only apply to newly-opened windows.** Once a cabinet is marked moved, `DoMoveSearchBandToMenuBar` short-circuits and there's no un-move path, so toggling e.g. `MoveBreadcrumb` on an already-open window has no effect until it's reopened. Worth documenting, or re-applying on settings change.
- **Struct default vs. settings default mismatch.** `Settings::moveBreadcrumb = true` (line 82) contradicts the settings-block default `MoveBreadcrumb: false` (line 16). It's dead (overwritten by `LoadSettings`), but align them to avoid confusion.
- **`$name` wording.** The `$name`s say "Show Search Bar" / "Show Up Button", but the mod *moves* these into toolbars rather than showing/hiding them — "Move ..." would read truer.

</p>
</details>

<p>

- **The `NtSetValueKey` block is process-wide.** Suppressing every `ITBar7Layout` write is a reasonable way to keep Explorer from persisting the mangled rebar layout (good for reversibility), but it also means Explorer can't persist *any* toolbar layout (band sizes for the command/menu bars, etc.) while the mod is active. At the `NtSetValueKey` level you only see the value name, not which rebar, so there's no obvious narrower alternative — just noting the side effect.
- **Layout is stored under global keys.** `Cx_<class>`, `Order_%d`, and `BandCount` are per-process globals, so all Explorer windows (and future ones) share a single saved layout and concurrent saves race (last writer wins). Fine if a single global layout is intended.
- **DPI.** `GetDesiredBandHeight()` derives the band height from system-DPI `GetSystemMetrics`; on a mixed-DPI multi-monitor setup it may be slightly off on secondary monitors. Classic Explorer's rebar is system-DPI anyway, so this is minor.

</p>
</details>

Anixx:
"Reversibility ("effects disappear when the mod is disabled") is a core Windhawk principle." - I thought this applies to the long-term system state, not the current session or process. Otherwise, the Classic Theme mod is not reversible, as many others. Also, restoring the exact state is simply impossible.

m417z:
Yeah, the wording is too strong. You can treat it as a recommendation.

Anixx:
I can add Explorer's restart at uninit or something.

m417z:
I think adding a note about it in the readme is good enough.

Anixx:
Also, according my tests, Explorer's look gets fully restored on mod's unload, even the already opened windows. So, it is better not to do anything, I think. It self-repairs.

Anixx:
fixed the issues

Issue #4686, opened 2026-07-05
[Dynamic Taskbar Transparency] Disabling states aren't disabling actions | Multi monitor support


Deen-0x:
Mod author: @r1file 

This mod is life saver. I appreciate the work!
I'd like to report a bug where disabling a state, for example "Start opened" state, it still functions, ultimately breaking the purpose. In other words, when a state is disabled, I expect it not to function at all, so triggering it shouldn't perform actions.

Another thing is a feature request: I noticed the behavior isn't consistent across multiple monitors. On my second monitor, the taskbar shows despite no maximized windows.

Deen-0x:
Thanks, the first bug was addressed in 0.3.9. I'll open the other as a mod-enhancement/feature request.

Issue #4677, opened 2026-07-03
[Dynamic Taskbar Transparency] windows and borders


littlemot:
Mod author: @r1file 

hi, thanks for the very cool mod!

i noticed something that differentiates this mod from ttb regarding the start menu option

i noticed a behavior that differentiates this mod from ttb regarding the start menu. in ttb, when you disable the start menu option, the taskbar behaves according to the desktop and maximized window options no matter what, even when the start menu is opened.

although in this mod, for example, if i have the desktop set to clear, maximized window set to default, and the start menu option disabled, the taskbar stays clear when i open the start menu, regardless of whether a window is maximized or not. it doesnt seem to respect the maximized window option once the start menu is opened. i would like if the mod behaved exactly like ttb in this case

another thing, i use [this](https://github.com/bydevara/Glowing-Taskbar-for-Windows-11/blob/main/Glowing%20Taskbar%20byDevara.yaml) theme in taskbar styler that adds a glowing border to the taskbar. unfortunately, when setting the taskbar to clear, the glowing border remains visible. is there any way for me to make the border disappear along with the rest of the taskbar, if not, then something the mod could be able to do in a future update

thanks again, and lmk got confusing in the text

PR #4746, opened 2026-07-11
Fix disabled Dynamic Taskbar Transparency states

r1file:
## Summary

- Make disabled shell states fall through instead of overriding the current taskbar appearance
- Bump Dynamic Taskbar Transparency to 0.3.9

## Details

When Start/Search/Task View/tray/other shell states were disabled, `ResolveState` still selected those states. `GetAppearanceForState` then fell back to the desktop appearance, so opening Start could ignore the maximized-window appearance. Disabled states now behave as inactive and allow later state checks, such as maximized windows, to apply.

This addresses the disabled-state behavior reported in #4677 and #4686. The multi-monitor maximized behavior from #4686 is intentionally left as a separate larger change.

## Validation

- `python .github/pr_validation.py` with one changed mod file
- `python -u scripts/compile_mod.py -w "C:\Program Files\Windhawk" -f mods/dynamic-taskbar-transparency.wh.cpp -o32 "$env:TEMP\dynamic-taskbar-transparency_32.dll" -o64 "$env:TEMP\dynamic-taskbar-transparency_64.dll" -oarm64 "$env:TEMP\dynamic-taskbar-transparency_arm64.dll"`

### DIFF
--- a/mods/dynamic-taskbar-transparency.wh.cpp
+++ b/mods/dynamic-taskbar-transparency.wh.cpp
@@ -2,7 +2,7 @@
 // @id              dynamic-taskbar-transparency
 // @name            Dynamic Taskbar Transparency
 // @description     Dynamically changes the Windows 11 taskbar XAML background transparency for desktop, Start, search, tray flyouts, task view, and maximized windows.
-// @version         0.3.8
+// @version         0.3.9
 // @author          11581
 // @github          https://github.com/r1file
 // @include         explorer.exe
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

I haven't verified the review, address it or let me know if it's incorrect.

---

The fix is correct and does what the summary describes — disabled shell states now fall through to lower-priority checks instead of overriding the appearance. One follow-up: the same fix wasn't applied to the `maximized` branch, which leaves it inconsistent with the states you just fixed.

**`ResolveState` still selects `maximized` unconditionally, reproducing the same bug for that state** (`mods/dynamic-taskbar-transparency.wh.cpp:1952`). Every other branch now guards on `settings.<state>.enabled`, but the maximized branch is still:

```cpp
if (hasMaximizedWindow) {
    return {TaskbarDynamicState::maximized, L"maximizedWindow",
            hasMaximizedWindow};
}
```

`maximized` does have an enable toggle (`settings.maximized.enabled`, read at line 378 and guarded in `GetAppearanceForState` at line 1873). So when the user disables the maximized state but enables `otherInteraction`, a maximized window still makes `ResolveState` return `maximized` → `GetAppearanceForState` falls back to the desktop appearance and `otherInteraction` never gets a chance to apply — exactly the class of bug this PR fixes for Start/Search/Task View/tray. For consistency:

```cpp
if (hasMaximizedWindow && settings.maximized.enabled) {
    return {TaskbarDynamicState::maximized, L"maximizedWindow",
            hasMaximizedWindow};
}
```

`hasMaximizedWindow` is still computed and carried in the resolution's third field for logging, so this is safe. If leaving `maximized` as an always-on floor is intentional (different from the others), a short comment explaining why would help; otherwise it looks like the fix was just applied to five of the six states.

PR #4749, opened 2026-07-11
Add lid-sleep-delay v1.0: Configurable delay before sleep on lid close

remixansh:
## New Mod: Lid Close Sleep Delay

Adds a configurable delay (30s, 1min, 5min, 10min) before the system sleeps when the laptop lid is closed.

### Features
- Dynamic system detection: auto-detects lid switch via power capabilities, battery status, and WMI chassis type
- Multi-architecture: x86, x86-64, ARM64
- Gracefully disables on desktops without a lid switch
- Live settings reload without restart
- Windows 7/8/8.1/10/11 compatible

### Prerequisites
User must set Windows lid-close action to **Do Nothing** in Power Options.

### Source
https://github.com/remixansh/lid-sleep-delay

## Changelog

* v1.0 - Initial release with dynamic lid detection, multi-architecture support, and configurable delay.

## Mod authorship

This mod was created by:

- [x] The submitter, without AI assistance

### DIFF
--- a/mods/lid-sleep-delay.wh.cpp
+++ b/mods/lid-sleep-delay.wh.cpp
@@ -1,0 +1,652 @@
+// ==WindhawkMod==
+// @id              lid-sleep-delay
+// @name            Lid Close Sleep Delay
+// @description     Delays system sleep after laptop lid is closed by a configurable time (30s, 1min, 5min, 10min)
+// @version         1.0
+// @author          Ansh Raj
+// @github          https://github.com/remixansh
+// @include         explorer.exe
+// @compilerOptions -lpowrprof -luser32 -ladvapi32 -lole32 -loleaut32 -lwbemuuid
+// @license         MIT
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Lid Close Sleep Delay
+
+This mod adds a configurable delay before your laptop goes to sleep when the
+lid is closed. It works dynamically on **any Windows system** — laptops,
+convertibles, and tablets with a lid sensor. On desktops or systems without
+a lid switch, the mod detects this at startup and gracefully disables itself.
+
+## Compatibility
+
+- **Architectures**: x86 (32-bit), x86-64 (64-bit), and ARM64
+- **Windows versions**: Windows 7, 8, 8.1, 10, and 11
+- **System types**: Automatically detects laptops vs desktops
+
+## How it works
+
+1. On startup, the mod checks if the system has a **lid switch** (battery
+   presence, chassis type, and power capabilities are all checked).
+2. If a lid switch is detected, it listens for **lid state change** power
+   notifications.
+3. When the lid is closed, it keeps the system awake and starts a countdown.
+4. If you reopen the lid before the timer expires, sleep is cancelled.
+5. If the timer expires with the lid still closed, the system goes to sleep.
+6. On desktops without a lid, the mod logs a message and exits cleanly.
+
+## Prerequisites
+
+**Set your Windows lid-close action to "Do Nothing":**
+
+1. Open **Control Panel** → **Power Options** → **Choose what closing the lid
+   does**
+2. Set **"When I close the lid"** to **"Do nothing"** for both **On battery**
+   and **Plugged in**
+3. Click **Save changes**
+
+## Options
+
+| Option      | Delay           |
+|-------------|-----------------|
+| **30sec**   | 30 seconds      |
+| **1min**    | 1 minute (default) |
+| **5min**    | 5 minutes       |
+| **10min**   | 10 minutes      |
+*/
+// ==/WindhawkModReadme==
+
+// ==WindhawkModSettings==
+/*
+- SleepDelay: 1min
+  $name: Sleep delay after lid close
+  $description: Select how long to wait before sleeping after the lid is closed. Changes apply immediately on save.
+  $options:
+    - 30sec: 30 seconds
+    - 1min: 1 minute (default)
+    - 5min: 5 minutes
+    - 10min: 10 minutes
+*/
+// ==/WindhawkModSettings==
+
+#include <windows.h>
+#include <powrprof.h>
+#include <batclass.h>
+#include <comdef.h>
+#include <wbemidl.h>
+
+#pragma comment(lib, "wbemuuid.lib")
+
+// ─── GUIDs ───────────────────────────────────────────────────────────────────
+
+// {BA3E0F4D-B817-4094-A2D1-D56379E6A0F3}
+static const GUID GUID_LIDSWITCH_STATE_CHANGE_LOCAL = {
+    0xBA3E0F4D, 0xB817, 0x4094,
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

I haven't verified 3, but it seems to make sense. Let me know if you verified it and it works as is.

---

**1. This should be a "mods as tools" dedicated-process mod, not an `explorer.exe` injection.** Every signal for the tool-mod pattern is present here (see the [wiki page](https://github.com/ramensoftware/windhawk/wiki/Mods-as-tools:-Running-mods-in-a-dedicated-process)):

- **No function hooks.** The mod only uses system-wide APIs (`RegisterPowerSettingNotification`, `SetSuspendState`, `SetThreadExecutionState`, WMI) that work from any process. There is no reason to be inside the shell.
- **You hand-rolled a single-instance mutex** (`WindhawkLidSleepDelay_SingleInstance`) precisely to dedupe across multiple `explorer.exe` instances — the tool-mod launcher does this for you.
- **The work is desktop-global**, not tied to explorer's state — you just watch the lid and suspend the machine.
- **Blast radius:** a bug here forcibly *suspends the whole system* from within the shell process, and a crash destabilizes explorer.

Rewrite it as a tool mod: set `@include windhawk.exe`, rename `Wh_ModInit`/`Wh_ModSettingsChanged`/`Wh_ModUninit` → `WhTool_*`, paste the launcher boilerplate verbatim from the wiki, and delete the home-grown mutex. The bottom of [mods/explorer-folder-hover-menu.wh.cpp](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/explorer-folder-hover-menu.wh.cpp#L3778-L3855) is a working reference for both the `WhTool_*` callbacks and the launcher snippet.

**2. `TerminateThread` on unload can corrupt the process and leaks a dangling window class.** In `Wh_ModUninit`, if the thread doesn't exit within 3s you call `TerminateThread(g_hThread, 0)`:

```cpp
if (waitResult == WAIT_TIMEOUT) {
    Wh_Log(L"Thread did not exit in time — terminating");
    TerminateThread(g_hThread, 0);
}
```

`TerminateThread` is [documented as unsafe](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-terminatethread) — it can leave loader/heap locks held and corrupt the host. Worse, killing the thread skips the cleanup at the end of `LidMonitorThread` (`UnregisterPowerSettingNotification`, `DestroyWindow`, `UnregisterClass`), so the process is left with the class `WindhawkLidSleepDelay_v14` still registered and its `lpfnWndProc` pointing into the now-unmapped mod image. On the **next load** that either fails (`ERROR_CLASS_ALREADY_EXISTS`) or — since your code swallows that error and proceeds — dispatches messages to a dangling `WndProc` → crash. The `if (!atom && GetLastError() != ERROR_CLASS_ALREADY_EXISTS)` swallow and the `_v14` suffix on the class name are both symptoms of having hit this during development. Drop the `TerminateThread` fallback and make the normal `WM_QUIT_THREAD` → cleanup path reliable (with the tool-mod launcher, `WhTool_ModUninit` handles this — it `ExitProcess`es the dedicated process instead of force-killing a thread inside the shell).

**3. The primary lid-detection path (`HasLidSwitchCapability`) is broken — it never actually reads `LidPresent`.** You define a truncated 8-byte `SYSTEM_POWER_CAPABILITIES_FULL` and pass `sizeof(caps)` to `CallNtPowerInformation(SystemPowerCapabilities, ...)`:

```cpp
SYSTEM_POWER_CAPABILITIES_FULL caps = {};   // only 8 bytes
NTSTATUS status = CallNtPowerInformation(SystemPowerCapabilities, NULL, 0, &caps, sizeof(caps));
```

`SystemPowerCapabilities` writes a full `SYSTEM_POWER_CAPABILITIES` (~90 bytes); if the output buffer is smaller the call returns `STATUS_BUFFER_TOO_SMALL` and writes nothing, so `status != 0` and the function returns `FALSE` on **every** machine. Detection then silently falls through to the `HasBattery() && IsPortableChassis()` heuristic — meaning the most reliable signal (the actual `LidPresent` capability) is never consulted, and you unnecessarily spin up WMI on real laptops. `SYSTEM_POWER_CAPABILITIES` is already declared in `<windows.h>` — just use the real type and its `sizeof` instead of the partial reimplementation:

```cpp
SYSTEM_POWER_CAPABILITIES caps = {};
NTSTATUS status = CallNtPowerInformation(SystemPowerCapabilities, NULL, 0, &caps, sizeof(caps));
```

<p>

- **`volatile` is not a synchronization primitive.** `g_sleepDelaySeconds`, `g_bRunning`, `g_bLidClosed`, `g_bTimerActive` are shared across the Windhawk thread and the monitor thread. `volatile` doesn't give you atomicity or ordering under the C++ memory model — use `std::atomic<int>` / `std::atomic<bool>` instead. (The flags touched only from the monitor thread's own message loop are fine, but `g_sleepDelaySeconds` and `g_bRunning` genuinely cross threads.)
- **Error-path leaks in `Wh_ModInit`.** When it returns `FALSE` after the thread fails to come up (the `!g_bRunning` check) — or when `RegisterPowerSettingNotification` fails inside the thread — `g_hThread` and `g_hMutex` are never closed. Close them on those paths. (Moot once you adopt the tool-mod launcher and drop the mutex.)
- **`Sleep(200)` to wait for thread startup** is racy. Use an event the thread signals once it's ready, and `WaitForSingleObject` on it — see `g_readyEvent` in the [reference tool mod](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/explorer-folder-hover-menu.wh.cpp#L3787-L3806).
- **Redundant lib directive.** `#pragma comment(lib, "wbemuuid.lib")` duplicates `-lwbemuuid` in `@compilerOptions`; keep one.
- **Leftover version/dev artifacts.** `@version` is `1.0` but the init log says `v1.5`, and the window class is named `WindhawkLidSleepDelay_v14`. Pick a stable class name and make the strings consistent.

</p>
</details>

<p>

- **The "Do Nothing" prerequisite isn't reversible with the mod.** Because the feature works by requiring the user to set the Windows lid action to *Do Nothing* and then suspending manually, disabling the mod leaves the machine in a state where closing the lid does *nothing* until the user reverts that setting by hand — i.e. the effect doesn't fully disappear on disable. There's no clean API to inject a delay into the native lid action, so this is an inherent trade-off rather than a required change; just make sure the README makes the "you must manually revert this later" part explicit (it currently only documents the setup direction).
- **Settings change during an active countdown restarts the full delay from "now."** If the lid is closed with a 60s delay and you switch to 30s 50s later, `WM_RELOAD_SETTINGS` kills and re-arms a fresh 30s timer, so the machine sleeps at ~80s rather than adjusting the remaining time. Probably fine, but worth being aware of.

</p>
</details>

m417z:
Does it work for you? How was the mod tested?

remixansh:
I commit new files and it is tested and working properly

PR #4752, opened 2026-07-11
Add Low Battery Alarm v1.0.0

communism420:
Adds Low Battery Alarm, a new Windhawk mod for laptops that plays configurable repeating alerts when the battery reaches low or critical charge levels.

Key features:

- Separate configurable low-battery and critical-battery thresholds.
- Independent polling and repeat intervals with explicit time units.
- Immediate playback cancellation when external power is connected.
- Custom WAV selection through either the standard Windows file picker or manual full-path entry.
- Mutually exclusive sound-selection modes with disabled inactive controls.
- User-selectable alarm output device, including a specific speaker endpoint or the default Windows output.
- Automatic fallback to the Windows default output if the selected device becomes unavailable.
- Configurable repetitions and pauses between repetitions.
- Optional Windows critical-system-sound fallback.
- Temporary volume increase and unmuting of the selected output endpoint, with restoration of the previous state.
- Test playback after settings changes.
- Safe handling of systems without a battery.
- A single active controller per Windows session to avoid duplicate alarms.

Implementation details:

- Targets 64-bit `explorer.exe`.
- Uses Windows power notifications for low-latency AC/DC state changes.
- Uses Core Audio to enumerate and manage output endpoints.
- Uses endpoint-specific WinMM playback for routing WAV audio to the selected device.
- Uses `waveOutReset` for immediate and safe playback interruption.
- Does not create external player processes, persistent tray icons, or registry-based configuration.
- Falls back safely if a WAV file or selected output device is unavailable.

Validation performed:

- Successfully compiled with the Windhawk Clang toolchain.
- Compiled with strict warning flags without warnings.
- Passed Clang static analysis without findings.
- Verified metadata, settings schema, and cleanup paths.

## Changelog

* Not applicable — this pull request introduces a new mod.

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [x] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/low-battery-alarm.wh.cpp
+++ b/mods/low-battery-alarm.wh.cpp
@@ -6,9 +6,8 @@
 // @author          communism420
 // @github          https://github.com/communism420
 // @homepage        https://github.com/communism420
-// @include         explorer.exe
-// @architecture    x86-64
-// @compilerOptions -lwinmm -lole32 -luuid -lcomdlg32 -lgdi32 -ladvapi32
+// @include         windhawk.exe
+// @compilerOptions -lwinmm -lole32 -luuid -ladvapi32 -lshell32 -lgdi32 -lcomdlg32
 // @license         MIT
 // ==/WindhawkMod==
 
@@ -30,18 +29,16 @@ stopped immediately instead of waiting for the next polling interval.
 
 ## Custom sound and fallback
 
-Enable **Configure sound source** and save the settings to open the sound-source
-window. Choose a method with its radio button. In **Windows file picker** mode,
-use **Browse...** to select a WAV file. In **Manual full path** mode, type or paste
-the complete path into the editable field. Both paths remain visible, but the
-controls belonging to the inactive method are disabled. The paths are independent,
-and only the selected method supplies the alarm sound path.
-
-The same window contains **Alarm output device**. Choose **Windows default
-output** to follow the normal Windows route, or choose a specific active output
-such as the laptop speakers to route the alarm directly to that device. If the
-saved device is temporarily unavailable, the mod logs the condition and uses the
-Windows default output for that alert so that the alarm is not silently lost.
+Enter the complete WAV path manually in **WAV file path**, or use **Browse...**
+in the custom configuration window to select it with Windows Explorer. A manual
+path change saved in Windhawk replaces a path previously selected in the custom
+window. To open the window, enable **Open alarm configuration** and save the
+settings. The topmost window also lists the Windows default output roles and
+every active Core Audio output device. It stores the selected device by its
+stable endpoint ID. To reopen the window, disable the option and save, then
+enable it and save again. An enabled request is handled only once, so saving
+unrelated settings does not reopen the window. It also opens automatically on
+the first run, before an output selection has been stored.
 
 WAV is the guaranteed supported custom format. If no file has been selected, or
 if the selected file is later removed, becomes inaccessible, is not a WAV file,
@@ -51,7 +48,7 @@ audible, and recognizable WAV file is recommended.
 
 ## Volume and testing
 
-The mod can temporarily change the default output device's master volume and can
+The mod can temporarily change the selected output device's master volume and can
 temporarily unmute it while an alert is playing. The previous volume and mute
 state are normally restored afterward. If the user changes the volume during
 playback, the new value is preserved when the change can be detected by comparing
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

---

**1. This should be a "mods as tools" mod, not an `explorer.exe` injection.**

The mod injects into `explorer.exe` but hits every signal for the [Mods-as-tools pattern](https://github.com/ramensoftware/windhawk/wiki/Mods-as-tools:-Running-mods-in-a-dedicated-process):

- **It hooks nothing.** It only calls system-wide APIs that work from any process: `GetSystemPowerStatus`, `RegisterPowerSettingNotification`, Core Audio (`IMMDeviceEnumerator`/`IAudioEndpointVolume`), and WinMM (`waveOut*`/`mmio*`/`PlaySound`). None of that needs to run inside Explorer.
- **You hand-rolled a per-session single-instance mutex** (`CreateSessionInstanceMutex`, `g_instanceMutex`, `g_isPrimaryInstance`, and the whole "secondary instance" path) precisely to dedupe across the multiple `explorer.exe` instances a session can have. The tool-mod launcher does exactly this for you — one dedicated `windhawk.exe` process per mod.
- **The work is a standalone helper** that ignores Explorer's own state and just watches the machine's battery. A bug or crash in ~3300 lines of audio/COM/window code currently destabilizes the shell for no benefit.
- **Robustness bonus:** as written, if the primary Explorer instance crashes, the mutex is released but no secondary instance ever promotes itself to primary (secondaries only decide at init), so the alarm silently stops until a reload. A dedicated tool-mod process is relaunched by the framework, avoiding this entirely.

Recommended change: set `@include windhawk.exe`, rename `Wh_ModInit`/`Wh_ModSettingsChanged`/`Wh_ModUninit` → `WhTool_ModInit`/`WhTool_ModSettingsChanged`/`WhTool_ModUninit`, paste the launcher boilerplate **verbatim** from the wiki, and delete all the home-grown single-instance code (`CreateSessionInstanceMutex`, `g_instanceMutex`, `g_isPrimaryInstance`, the secondary-instance branches). A complete, up-to-date copy of that boilerplate lives at the bottom of [mods/explorer-folder-hover-menu.wh.cpp](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/explorer-folder-hover-menu.wh.cpp#L3835-L4013). As a side effect, blocking in `Wh_ModInit` waiting for worker readiness also stops mattering, since it no longer stalls Explorer startup.

**2. Consider whether the custom configuration window is worth its complexity.**

A large fraction of the mod is a hand-built Win32 dialog (`SoundConfigurationWindowProcedure`, its window class registration/lifecycle, `SoundFilePickerHookProcedure`, `UpdateSoundConfigurationControls`, etc.), driven by the awkward "toggle `ConfigureSoundSource` off, save, on, save to open a window" UX. Most of this can be expressed as plain Windhawk settings, which removes a lot of surface area (a whole window class, a file-picker hook, and the `ERROR_CLASS_ALREADY_EXISTS` workarounds):

- The **manual WAV path** is just a string setting read via `Wh_GetStringSetting`/`WindhawkUtils::StringSetting` — no dialog needed.
- The **output device** can be stored as a string setting (device name/ID) that the code resolves at playback time, as it already does in `ResolveWaveOutDeviceId`.

The file browse button is the only piece that genuinely benefits from a dialog, and it's not clearly worth the machinery. Since the code is AI-assisted, it's worth asking whether the whole custom-window subsystem is intentional or just accreted complexity — a plain string setting for the path would be simpler for both users and maintenance. (Note: the `RegisterClassExW` → "reuse the existing class if `lpfnWndProc` matches" fallback in `RunSoundConfiguration` is the kind of stale-class workaround that's best avoided; dropping the custom window removes the concern.)

<p>

- **Setting descriptions say "default output device"** for the volume/unmute options, but `AudioStateGuard` actually applies them to the *selected* endpoint when a specific device is chosen. Consider wording it as "the selected output device" for accuracy.

</p>
</details>

<p>

- **Fuzzy device matching may route to the wrong endpoint.** `ResolveWaveOutDeviceId` and `FindAudioEndpointByFriendlyNamePrefix` fall back to prefix matching on friendly names when the exact ID isn't found. They do guard against ambiguous matches, but two devices sharing a name prefix (e.g. identical USB headsets) could still resolve to an unexpected one. Storing/matching on the stable endpoint ID alone would be more predictable.

</p>
</details>

PR #4759, opened 2026-07-12
Add Proxifier drag to rule mod 0.1

zxhzxhz:
Adds proxifier-drag-to-rule version 0.1.

- Opens a new Proxification Rule when executable files or folders are dropped on the Proxifier main window.
- Prefills Applications and Name; passes .ppx drops through to Proxifier's default handling.
- Verified against Proxifier 4.11 (x86).
- Compiled successfully with the repository's scripts/compile_mod.py using Windhawk's x86 toolchain.

### DIFF
--- a/mods/proxifier-drag-to-rule.wh.cpp
+++ b/mods/proxifier-drag-to-rule.wh.cpp
@@ -1,0 +1,428 @@
+// ==WindhawkMod==
+// @id              proxifier-drag-to-rule
+// @name            Proxifier drag and drop to rule
+// @description     Open a new Proxification Rule prefilled with dropped applications.
+// @version         0.1
+// @author          zxhzxhz
+// @github          https://github.com/zxhzxhz
+// @include         Proxifier.exe
+// @architecture    x86
+// @compilerOptions -lcomctl32 -lshell32
+// @license         MIT
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Proxifier drag and drop to rule
+
+Drop executable files or folders on the Proxifier main window. The mod opens
+the Add Proxification Rule dialog and fills its Applications field, but never
+saves a rule automatically.
+
+The defaults were verified against Proxifier 4.11. Restart Proxifier after
+enabling or updating the mod so its main window can be subclassed.
+*/
+// ==/WindhawkModReadme==
+
+// ==WindhawkModSettings==
+/*
+- enabled: true
+  $name: Enable drag and drop
+  $description: Accept executable files and folders dropped on the main window.
+- folderMode: 0
+  $name: Folder handling mode
+  $description: 0 adds folder\\*.exe to the rule. 1 ignores dropped folders.
+- rulesCommand: 32798
+  $name: Proxification Rules command ID
+  $description: Proxifier 4.11 uses 32798 (0x801E).
+- rulesAddButtonId: 1001
+  $name: Add button control ID
+  $description: Proxifier 4.11 uses 1001.
+- applicationsControlId: 1018
+  $name: Applications control ID
+  $description: Proxifier 4.11 uses 1018.
+- nameControlId: 1017
+  $name: Name control ID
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

In addition, please fix the CI error by restoring the removed PR template and filling it.

---

**1. Cross-thread subclassing — this is why you need the "restart Proxifier" instruction.** The mod attaches with raw `SetWindowSubclass`/`RemoveWindowSubclass` (lines 359, 420). These only work when called on the window's *own* thread, but two of the three call sites run on the wrong thread:

- `Wh_ModInit` → `EnumWindows(AttachExistingMainWindow)` → `AttachToMainWindow` → `SetWindowSubclass` runs on the Windhawk injection thread, not Proxifier's UI thread. So when the mod is enabled while Proxifier is already running, the subclass silently fails to attach — which is almost certainly why the README has to tell users to restart Proxifier.
- `Wh_ModUninit` → `RemoveWindowSubclass` also runs on an arbitrary thread.

Use `WindhawkUtils::SetWindowSubclassFromAnyThread` / `RemoveWindowSubclassFromAnyThread` (see [`mods/explorer-name-windows.wh.cpp#L285`](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/explorer-name-windows.wh.cpp#L285)). Note the subclass proc signature must change: `WH_SUBCLASSPROC` is `(HWND, UINT, WPARAM, LPARAM, DWORD_PTR)` — drop the `UINT_PTR uIdSubclass` parameter, and removal takes no subclass id. `DragAcceptFiles` likewise touches the window's extended style and has the same thread affinity, so for the already-running path it should also be applied on the window's UI thread (the `CreateWindowExW` hook path already runs there; the `EnumWindows` path does not). Once this is fixed the restart instruction can be dropped.

**2. The automation timer never times out, and a stuck sequence blocks all further drops.** In `AdvanceAutomation`, both stages just `return` when the expected dialog isn't found yet (lines 253, 274), and `StopAutomation` is only reached on the success path or when a control ID lookup fails. If the Rules dialog never appears — wrong `rulesCommand`/control ID for the user's Proxifier version, or the user closes a dialog mid-sequence — the 50 ms timer polls forever, and because `g_automationStage` stays non-idle, `StartRuleCreation` rejects every subsequent drop with "Ignoring drop while a rule dialog is being opened" (line 221) until Proxifier is restarted. Add a bounded attempt/elapsed-time counter that calls `StopAutomation()` after N ticks so the mod recovers instead of wedging silently.

**3. `enabled` setting is redundant, and its "off" behavior is surprising.** The subclass and `DragAcceptFiles` are installed regardless of this setting, so it only gates `StartRuleCreation` — meaning users can already disable the whole mod from Windhawk itself. Worse, when `enabled` is `false`, an exe/folder drop still gets `DragFinish`'d and swallowed (`return 0`, lines 320-327) rather than passed through to Proxifier. Drop the setting entirely.

**4. `folderMode` as an int `0`/`1`** would read more clearly as a boolean (`ignoreFolders: false`), since it's a binary toggle.

**5. `@architecture x86` only covers 32-bit Proxifier.** The window class (`Proxifier32Cls`) and your verified build are both 32-bit, so this may be intentional — but if 64-bit Proxifier builds exist and use a different class name, they won't be covered. Worth confirming whether 64-bit support is wanted.

<p>

- **`KillTimer` on unload has thread affinity.** `Wh_ModUninit` → `StopAutomation` → `KillTimer(g_mainWindow, ...)` runs on an arbitrary thread, but `KillTimer` only works from the thread that set the timer. If the mod is unloaded mid-automation the timer keeps firing (harmless stray timer until the window is destroyed). Same root cause as item 1 — resolved by marshalling teardown to the UI thread.
- **Use `WindhawkUtils::SetFunctionHook`** instead of raw `Wh_SetFunctionHook` with `void*` casts (lines 409-411) for type safety.
- A short GIF/screenshot of the drag-to-rule flow in the README would help, since the mod has a visible UI effect.

</p>
</details>

PR #4765, opened 2026-07-12
SnapSentry 0.16.0

mario0318:
## Changelog

Initial release.

## Mod authorship

This mod was created by:

- - [x] Claude
- - [x] ChatGPT
- - [ ] Gemini

## Relationship to screenshot-path-to-clipboard

figured I should mention this up front since near-duplicates tend to get held up here. screenshot-path-to-clipboard also watches the Screenshots folder and copies the new file's path, which is basically what SnapSentry's Path mode does, so there's real overlap on that one setting. the rest of it goes further though. it can put the actual image on the clipboard instead of just the path, and then there's the delete side, either after a delay or into the recycle bin. the notification with Keep and Delete and Copy on it is probably the main difference, the cleanup ends up being something you pick rather than something that just happens to you. so more of a superset than a duplicate, but I didn't want to leave that unsaid.

## What it registers while enabled

worth being upfront about this too. for the notification buttons to actually work an unpackaged process like this needs an AUMID-tagged Start Menu shortcut and an activator CLSID, so while the popup is on it writes a SnapSentry.lnk under the Start Menu and a CLSID key under HKCU. both get removed when you turn the popup off, disable the mod, or unload it, and if an earlier run got killed and left them behind they get reclaimed on the next start. the one gap is if the process is killed or Windows shuts down without the unload running and the mod is then disabled before it next starts, those two can hang around. the dialog fallback needs none of it, so if that footprint isn't acceptable the toast path can be dropped for the dialog instead.

### DIFF
--- a/mods/snap-sentry.wh.cpp
+++ b/mods/snap-sentry.wh.cpp
@@ -2,7 +2,7 @@
 // @id              snap-sentry
 // @name            SnapSentry
 // @description     Copy saved screenshots, delete them automatically, or choose what to do from a notification.
-// @version         0.13.0
+// @version         0.14.0
 // @author          mario0318
 // @github          https://github.com/mario0318
 // @include         windhawk.exe
@@ -944,6 +944,8 @@ public:
                 action = ACTION_COPY_DELETE;
             } else if (command == L"keep") {
                 action = ACTION_KEEP;
+            } else if (command == L"body") {
+                action = ACTION_COPY_ONLY;  // Toast clicked away, not answered.
             }
         }
         EnterCriticalSection(&g_toastLock);
@@ -1086,7 +1088,8 @@ static bool ShowToast(const std::wstring& path, const Settings& s, int& action) 
     ULONGLONG toastId = GetTickCount64();
     std::wstring id = std::to_wstring(toastId);
     std::wstring xml =
-        L"<toast scenario=\"reminder\" duration=\"long\">"
+        L"<toast scenario=\"reminder\" duration=\"long\" "
+        L"activationType=\"background\" launch=\"body|" + id + L"\">"
         L"<visual><binding template=\"ToastGeneric\">"
         L"<text>Screenshot saved</text><text>" +
         XmlEscape(name) +
@@ -1391,7 +1394,11 @@ static int AskAction(const std::wstring& path, const Settings& s) {
         case ACTION_KEEP:
             return result;
         default:
-            return ACTION_KEEP;
+            // Esc / the X close the dialog with IDCANCEL. Teardown clicks the real
+            // Keep button, and entry re-checks the stop event, so a cancel here can
+            // only be a live user dismissing it: copy, never delete, matching the
+            // toast-swipe path.
+            return ACTION_COPY_ONLY;
     }
 }
 
--- a/mods/snap-sentry.wh.cpp
+++ b/mods/snap-sentry.wh.cpp
@@ -2,7 +2,7 @@
 // @id              snap-sentry
 // @name            SnapSentry
 // @description     Copy saved screenshots, delete them automatically, or choose what to do from a notification.
-// @version         0.14.0
+// @version         0.15.0
 // @author          mario0318
 // @github          https://github.com/mario0318
 // @include         windhawk.exe
@@ -433,17 +433,23 @@ static std::wstring RenameByActiveWindow(const std::wstring& path) {
 // Path text formatting (Path clipboard mode)
 // ============================================================================
 
-// Minimal file URI: forward slashes and %20 for spaces, enough for a link or a
-// Markdown image to paste cleanly.
+// File URI with proper percent-encoding, so a name with #, %, ( or ) (which window
+// titles are full of, and which SanitizeForFilename keeps) still makes a link and a
+// Markdown image that paste cleanly rather than a broken one.
 static std::wstring PathToFileUri(const std::wstring& path) {
     std::wstring uri = L"file:///";
     for (wchar_t c : path) {
-        if (c == L'\\') {
+        if (c == L'\\' || c == L'/') {
             uri += L'/';
-        } else if (c == L' ') {
-            uri += L"%20";
+        } else if ((c >= L'A' && c <= L'Z') || (c >= L'a' && c <= L'z') ||
+                   (c >= L'0' && c <= L'9') || wcschr(L"-._~:", c)) {
+            uri += c;  // unreserved, plus the drive colon
+        } else if (c < 0x80) {
+            wchar_t esc[4];
+            swprintf_s(esc, L"%%%02X", (unsigned)c);
+            uri += esc;  // ' ' -> %20, '#' -> %23, '(' -> %28, '%' -> %25
         } else {
-            uri += c;
+            uri += c;  // non-ASCII passes through; file: consumers accept it
         }
     }
     return uri;
--- a/mods/snap-sentry.wh.cpp
+++ b/mods/snap-sentry.wh.cpp
@@ -1,0 +1,1547 @@
+// ==WindhawkMod==
+// @id              snap-sentry
+// @name            SnapSentry
+// @description     Copy saved screenshots, delete them automatically, or choose what to do from a notification.
+// @version         0.4.9
+// @author          mario0318
+// @github          https://github.com/mario0318
+// @include         windhawk.exe
+// @license         GPL-3.0
+// @compilerOptions -lole32 -lshell32 -lcomctl32 -lwindowscodecs -lruntimeobject -ladvapi32 -luuid
+// ==/WindhawkMod==
+
+// Source code is published under the GNU General Public License v3.0.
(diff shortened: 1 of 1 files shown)


m417z:

mario0318:
thanks, this was a useful one to get.

reversibility first since that was the big one. the mod pulls the start menu shortcut and the clsid key back out when it unloads now, both wow64 views, so disabling it doesn't leave stuff sitting around, and they get recreated next time it's enabled. I ended up doing the removal in the worker thread's teardown rather than WhTool_ModUninit though, because reading the shortcut's props to confirm it's actually ours needs a COM apartment and the uninit thread doesn't have one. spelled it out in the readme as well, that enabling registers a start menu entry and a bit of registry and disabling takes them back off.

folder resolution goes through FOLDERID_Screenshots by guid now, same knownfolders.h workaround screenshot-path-to-clipboard uses, instead of assuming Pictures\Screenshots. so a moved Screenshots folder actually gets followed.

the unload hang I did the way you suggested, stop-aware at both ends. AskAction bails to Keep if stop is already set so it won't put UI up during teardown, and TDN_CREATED does the same if stop fired before the handle got stored. put a check on the timer tick too, otherwise delaySeconds=0 / INFINITE could still just sit there.

the destructive default is the one I thought about most. there's a recycle bin option now and it's on by default, so deletes go through IFileOperation with FOF_ALLOWUNDO and you can pull the file back out. if the recycle fails it keeps the file rather than falling through to a hard delete. left deleteFile on by default since the auto cleanup is kind of the point of the whole thing, but missing the notification isn't destructive anymore, which was really the concern.

stale toast was just removeToast in the timeout branch, and the overlap note is in the description now.

on the mutex you were right, it's back to the stock snippet verbatim. the duplicates I saw were coming from an in-between version that waited on the mutex and carried on at WAIT_ABANDONED, not from the stock ERROR_ALREADY_EXISTS check, so that was me chasing my own bug. flagging the race you pointed out seemed better than carrying a private fork of the snippet.

grabbed a few of the optional ones as well. the ReadDirectoryChangesW buffer is aligned now, g_activeToastId gets cleared if Show fails, and there's an upper clamp on delaySeconds. also took the ARCHITECTURE.md mention out of that comment. the bigger refactors I've left alone, dropping logDetails and the StringSetting RAII swap and the idle loop timeout, mostly so this didn't turn into a rewrite, but they're noted.

mario0318:
the escape one's a real bug. cancel comes back as IDCANCEL through pnButton and the call still succeeds, so my FAILED check never saw it and it fell through to copy and delete. maps to keep now, and ProcessOne only acts on the ids it knows.

recycle takes FOFX_RECYCLEONDELETE (defined locally, not in the bundled sdk) so it recycles or fails, plus the aborted check. the keep-on-failure path was already there.

toast checks get_Setting first and drops to the dialog otherwise, dismissals count as keep, and delay 0 has a backstop now instead of waiting forever. no wrl/event.h in the sdk so that handler is a static class implementing the generated interface.

registration follows the popup setting, on and off live rather than written at startup.

took most of the small ones. left -luuid since I'd rather link without it than guess, and I've not moved the clipboard copy ahead of the popup or changed where the window title gets sampled, both too behavioral to fold into a review round.

mario0318:
yeah the readme was overselling that one. it goes off creation time now, so anything that wasn't made in the last 30s or so just gets ignored, which covers the sync case and stuff dragged in. did the check at the event rather than at processing time though, the worker can be sat on an earlier countdown for a while and real screenshots would've aged out by then. reworded the readme and the folder override bit.

registration looks at Enable processing now and not only the popup. first pass runs regardless so anything a shutdown left behind gets picked up.

dialog had the same forever wait, capped. whether the toast is worth the persistent state at all is up to you really, the dialog does the same three buttons without any of it.

mario0318:
good catches, all three are in.

the notifications settings entry was the one I'd missed. windows writes that per-aumid key the first time a toast goes out and nothing was taking it back, so SnapSentry stuck around in the Settings > Notifications list even after everything else was gone. it gets removed now in the same teardown pass as the shortcut and the clsid key, so turning the popup off or disabling the mod clears all three.

the copied-in file case is real too. a move keeps the creation time so that part was already covered, but a copy gets a fresh one, which is exactly why an old screenshot dragged in was slipping through. it checks the last write time as well now, a real capture has both within the window and a copy of an older image doesn't, so those get left alone. reworded the privacy paragraph to say what's actually true, a fresh download saved straight into the folder still can't be told apart from a capture by timestamps so I called that out rather than pretend otherwise.

took the optionals as well. the dismiss handler matches the toast it was raised for now so a stale dismissal can't answer the next screenshot's popup, the dialog default and any unrecognized result are keep instead of copy and delete, and a file it can't stat is left alone like every other unknown. left the bigger deferred stuff (StringSetting, dropping logDetails, the redundant CF_DIB, the 500ms poll) since those don't affect users and you'd said they're fine to leave.

mario0318:
all three are in, that settings key one was a dumb miss on my end.

mario0318:
deletion's off by default now, safer call. my notif-key fix last round missed the case where the popup was already off at unload so it was still getting left behind, sorted properly now. the system-wide-off path keeps instead of auto-deleting too, and i put the shortcut/clsid tradeoff in the description. cleared the optional ones you'd flagged as well.

mario0318:
good catches, both from me overreaching with that keep change. added an ACTION_COPY_ONLY so the "couldn't actually ask" cases still do the configured copy and just never delete, instead of doing nothing. and DisabledForApplication is treated the same as the rest now, copy only, no dialog, so silencing the app doesn't get you a modal on every shot. readme updated to match.

mario0318:
biggest one, the popup defaults off now, so out of the box it's just a clipboard tool with nothing written outside the mod's own settings, the start menu shortcut and clsid only appear if you turn the popup on. that should settle the reversibility side for the shipped config.

the two copy-vs-keep paths i'd missed are fixed, swiping the toast away and the dialog's no-answer backstop both copy now instead of doing nothing, and the dismiss handler checks the reason so a timed-out toast isn't read as an answer. went through the rest of the optional list too, and relabeled Keep to "Keep, don't copy" since the auto action is just the copy now.

left a couple on purpose, the dismiss sender check is gated against the live toast so nothing stale gets through, and WaitForStableFile stays off the COM pump to avoid re-entrancy on the path that must never delete early. clearing the notifications key on unload resets a pref on a reload, but it's the reversible call and moot now the popup's off.

mario0318:
both fixed, same shape as last round. the toast carries a launch id now so clicking the body maps to copy-only instead of falling through to the countdown, and cancelling the dialog with esc or the x returns copy-only too so it matches the swipe. the copy-vs-delete paths should all agree now.

mario0318:
ok. PathToFileUri percent-encodes everything outside the unreserved set now, so a name with #, %, or parens (which the rename produces) makes a valid file link and markdown image instead of a broken one.

mario0318:
handing this back to you, the three things left in the last review are yours to weigh rather than stuff i can keep iterating on.

the footprint (start menu shortcut + clsid key) is the main one. it's off by default now, only written if you opt into the popup, removed on toggle-off and on unload, and reclaimed on the next start. the one gap is a hard kill or shutdown before an unload and then disabling before it next runs, which can strand the clsid key, and that key points at windhawk.exe so a stale notification button would just start windhawk normally rather than do anything of ours. if that footprint isn't acceptable for a mod at all, the dialog fallback needs none of it and i'll drop the toast path.

on the notifications settings key the guidance went both ways, remove it for reversibility vs leave it so a user's own notification pref isn't reset on an update. i can go either way, just say which.

the serialized delete delay is real but only with deletion on and two shots inside the countdown, the second's copy waits on the first's timer. happy to add the pending-deletion queue for it, i just didn't want to rework the worker loop right at handoff.

and screenshot-path-to-clipboard is the path-copy case with no settings, this is a superset with image/file/delete/rename/prompt on top, your call whether that's fine as its own mod.

m417z:
How about addressing the "enabled" option removal?

> enabled duplicates Windhawk's own mod toggle. This is a tool mod: disabling it in Windhawk unloads it and Wh_ModUninit calls ExitProcess(0), so the process is gone. enabled: false keeps the whole process resident with two idle threads to achieve the same visible result. Dropping the setting (and the s.enabled && in SyncToastRegistration / the !s.enabled branch in WatchThread) removes a chunk of state for no loss.

mario0318:
good call, dropped the enabled setting and the two checks that used it. disabling in windhawk already unloads the whole thing, so it wasn't buying anything. also took a pass over the settings descriptions to make them clearer and grouped them while i was in there.

mario0318:
enabled's gone as you suggested. the four items still in the review are either your call (the shortcut/clsid footprint, and the overlap with screenshot-path-to-clipboard) or ones i offered above and will do on your word (the copy-then-delete serialization, and whether to stop clearing the notifications key). back to you.

m417z:
I think the rest is acceptable, thanks.

PR #4766, opened 2026-07-12
Create desktop transparent widgets

Stunning-dev:
## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini
- - [x] Other (Devin): 

### DIFF
--- a/mods/desktop-transparent-widgets.wh.cpp
+++ b/mods/desktop-transparent-widgets.wh.cpp
@@ -1,0 +1,820 @@
+// ==WindhawkMod==
+// @id              desktop-transparent-widgets
+// @name            JonaOS Draggable Transparent Widgets
+// @description     Transparent draggable Windows desktop widgets for calendar, clock, music, volume, battery, search, personalization, WiFi, Bluetooth, hotspot, and accessibility.
+// @version         1.0
+// @author          Jona like it, code it
+// @github          https://github.com/Stunning-dev
+// @include         windhawk.exe
+// @compilerOptions -ldwmapi -lgdiplus -lgdi32 -lole32 -lshell32
+// @license         MIT
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# JonaOS Draggable Transparent Widgets
+![Widgets preview](https://imgur.com/gallery/widgets-pggFD5f#mGAWwJP.png)
+*/
+// ==/WindhawkModReadme==
+
+#include <windhawk_api.h>
+
+#include <windows.h>
+#include <windowsx.h>
+#include <dwmapi.h>
+#include <gdiplus.h>
+#include <endpointvolume.h>
+#include <mmdeviceapi.h>
+#include <shlobj.h>
+#include <shellapi.h>
+
+#include <algorithm>
+#include <cmath>
+#include <cwchar>
+
+using namespace Gdiplus;
+
+namespace {
+
+constexpr int kDesignW = 400;
+constexpr int kDesignH = 680;
+constexpr UINT_PTR kTickTimer = 1;
+constexpr DWORD kTimerMs = 1000;
+constexpr wchar_t kClassName[] = L"WindhawkDesktopGlassWidgets";
+
+enum WidgetId {
+    WidgetCalendar,
+    WidgetClock,
+    WidgetMusic,
+    WidgetVolume,
+    WidgetBattery,
+    WidgetSearch,
+    WidgetPersonalization,
+    WidgetWifi,
+    WidgetHotspot,
+    WidgetBluetooth,
+    WidgetAccessibility,
+    WidgetCount,
+};
+
+struct Widget {
+    int id;
+    RectF rc;
+    bool round;
+};
+
+Widget g_widgets[WidgetCount] = {
+    {WidgetCalendar,        RectF(25, 15, 350, 170), false},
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

---

**1. `@include windhawk.exe` without the "mods as tools" launcher — the mod is loaded into Windhawk's own processes, not a dedicated one.** The mod has no function/symbol hooks at all — it only creates a window and calls system-wide APIs (`CoCreateInstance`, `GetSystemPowerStatus`, `ShellExecute`, `SendInput`). That's exactly the ["Mods as tools"](https://github.com/ramensoftware/windhawk/wiki/Mods-as-tools:-Running-mods-in-a-dedicated-process) case. But setting `@include windhawk.exe` while keeping the plain `Wh_ModInit`/`Wh_ModUninit` callbacks is **not** how tool mods run — it makes Windhawk inject the mod directly into every existing `windhawk.exe` process, including the session-0 service process (where a desktop window can't be shown and audio COM won't behave), with no dedicated persistent process and no single-instance guard, so you can get zero, duplicate, or short-lived widgets depending on which `windhawk.exe` instances happen to be running.

The tool-mod pattern solves all of this: it spawns one dedicated `windhawk.exe -tool-mod "desktop-transparent-widgets"` process that stays alive independently. To adopt it:
- Rename `Wh_ModInit` → `WhTool_ModInit` and `Wh_ModUninit` → `WhTool_ModUninit`.
- Paste the launcher boilerplate **verbatim** from the [wiki](https://github.com/ramensoftware/windhawk/wiki/Mods-as-tools:-Running-mods-in-a-dedicated-process) at the end of the file (keep `@include windhawk.exe`).

Every `windhawk.exe`-targeting mod in the catalog uses this pattern — see [`explorer-folder-hover-menu.wh.cpp`](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/explorer-folder-hover-menu.wh.cpp) (the launcher block at the bottom is a verbatim copy of the wiki snippet), or [`taskbar-music-lounge.wh.cpp`](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/taskbar-music-lounge.wh.cpp) for a similar always-on-desktop helper. As a bonus, the pattern's `Wh_ModUninit` ends in `ExitProcess(0)`, which removes the fragility of the current teardown (the `WaitForSingleObject(g_uiThread, 3000)` timeout will `CloseHandle` and let the DLL unload even if the UI thread is still executing mod code — a crash window, however unlikely).

**2. README image URL is not a valid/allowed image.** `https://imgur.com/gallery/widgets-pggFD5f#mGAWwJP.png` is an Imgur **gallery page**, not a direct image, and `imgur.com` is not one of the allowed image hosts — only `i.imgur.com` and `raw.githubusercontent.com` are. As written, the image won't render in the README at all. Upload the preview and use the direct link, e.g. `https://i.imgur.com/<id>.png` (or a `raw.githubusercontent.com` URL). Since this is a purely visual mod, a good screenshot/GIF really is worth getting right.

**3. Light mode is not supported.** Either add support for light mode, or, less preferable, add a note that the mod is only for dark wallpapers. https://i.imgur.com/RNsH24s.png

<p>

- **Don't recreate the audio endpoint every second.** The 1 s `WM_TIMER` calls `RefreshVolume()` → `GetEndpointVolumeInterface()`, which does a fresh `CoCreateInstance(MMDeviceEnumerator)` + `GetDefaultAudioEndpoint` + `Activate` on every tick just to read one float. Cache the `IAudioEndpointVolume` (recreating it only when the default device changes, e.g. via `IMMNotificationClient`), or subscribe to `IAudioEndpointVolumeCallback` and drop the polling entirely. `GetSystemPowerStatus` is cheap, so the battery poll is fine.
- `volatile bool g_unloading` is used for cross-thread signalling — prefer `std::atomic<bool>`. (With the tool-mod rewrite the flag largely goes away anyway.)
- `#include <shlobj.h>` appears unused (you only use `ShellExecuteW` from `<shellapi.h>`). Remove it if nothing else needs it.
- `OpenMusicFolder()` can pass `L"shell:My Music"` directly to `ShellExecuteW` as the file, without launching `explorer.exe` explicitly.

</p>
</details>

<p>

- **Single-monitor / DPI assumptions.** The widget panel is a fixed 400×680 canvas pinned to the bottom-right of the **primary** monitor's work area, with `UpdateScale()` deriving a scale factor from `GetSystemMetrics(SM_CXSCREEN/CYSCREEN)` against a hard-coded 1920×1080. On multi-monitor or per-monitor-DPI setups this can land on the wrong monitor or scale oddly. There's no trivial fix, so this is just an FYI — but it's the main thing users on non-standard displays will notice.
- **Widget positions aren't persisted.** Dragging updates `g_widgets[i].rc` in memory only, so every reload/toggle resets the layout to the defaults. Consider persisting positions with `Wh_SetIntValue`/`Wh_GetIntValue`.
- **`OpenWindowsSearch()` synthesizes Win+S via `SendInput`.** It works, but injecting global keystrokes is fragile (races with modifier state, can be swallowed if focus shifts). There's no clean public API for the search flyout, so this may be the pragmatic choice — noting it as a known trade-off.
- Consider exposing a settings block (opacity/colors, which widgets are shown, anchor corner) — the mod currently has no `==WindhawkModSettings==` at all, so everything is hard-coded.

</p>
</details>

Stunning-dev:
close pull request

PR #4775, opened 2026-07-13
Add Magic Mouse: Custom Mouse Gestures Mod

iMAboud:
Description: Introduces the "Magic Mouse" mod, enabling users to draw custom shapes with their mouse to trigger powerful system actions.

Key Features:

Built-in Gesture Recorder: An on-screen canvas to easily draw, capture, and save custom gesture shapes.
30+ Supported Actions: Quickly launch apps, manage windows, control media, snap windows, and more.
Special Utilities: Features unique tools like a drawable Sticky Note, an on-screen Spotlight, and text Search Selection.
Flexible Activation: Arm the gesture mode via configurable modifier keys or a physical "mouse wiggle" trigger.

vivaldi_0xFhFgZrF8 
vivaldi_vfWlb6luIZ 
vivaldi_sIKjlmazih 
vivaldi_DhmHXl1fec

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [x] Gemini


m417z:
### Submission review

---

Impressive amount of work, and the feature set is fun. The main things to fix are architectural: this is a self-contained desktop tool that shouldn't be injected into `explorer.exe`, plus a fragile window-lookup that breaks teardown, and a persistent file that isn't cleaned up.

**1. This should be a "tool mod", not injected into `explorer.exe`.** The mod installs no function hooks at all — it uses only `SetWindowsHookEx(WH_MOUSE_LL / WH_KEYBOARD_LL)` (line 4566‑4567), `SendInput`, `FindWindow`/`SendMessage`, and cross‑process COM (`IShellWindows`) — every one of which works from any process. It ignores explorer's own state and just operates on the desktop as a whole, and it even ships a home‑grown single‑instance guard (`FindWindow(MAIN_MSG_WND_CLASS)` at line 1673). Those are exactly the signals for the [Mods as tools](https://github.com/ramensoftware/windhawk/wiki/Mods-as-tools:-Running-mods-in-a-dedicated-process) pattern.

Injecting into `explorer.exe` has two concrete downsides here:
- **Multiple `explorer.exe` instances each install their own global LL hooks and hook thread** (only the notes thread is deduped). With a second explorer process (e.g. "Launch folder windows in a separate process", or the separate desktop/taskbar process on some setups), every gesture gets processed twice and each action fires twice.
- Any crash in ~4800 lines of gesture/GDI+/overlay code takes down the shell.

Recommended: convert to a tool mod — `@include windhawk.exe`, rename `Wh_ModInit`/`Wh_ModSettingsChanged`/`Wh_ModUninit` → `WhTool_*`, paste the launcher snippet verbatim, and delete the `FindWindow`‑based single‑instance code (the framework's launcher handles single‑instancing). The launcher boilerplate at the bottom of [mods/explorer-folder-hover-menu.wh.cpp](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/explorer-folder-hover-menu.wh.cpp) is a verbatim copy of the wiki snippet and shows all three `WhTool_*` callbacks. All the explorer‑touching actions (desktop‑icon toggle via `FindWindow`+`ShowWindow`, current‑folder path via `IShellWindows`, back/forward via `SendInput`) already work cross‑process, so nothing needs to actually run inside explorer.

**2. Teardown relies on `FindWindow` to locate a message-only window, which doesn't work.** The notes message window `g_mainMsgWnd` is created with `HWND_MESSAGE` as parent (line 1639), i.e. it's a *message-only window*. Those are not children of the desktop and are [not found by `FindWindow`](https://learn.microsoft.com/en-us/windows/win32/winmsg/window-features#message-only-windows) — you must pass `HWND_MESSAGE` to `FindWindowEx`. But `Wh_ModUninit` locates it with `FindWindow(MAIN_MSG_WND_CLASS, NULL)` (line 4778) to post `WM_QUIT`:

```cpp
HWND hwndNotesMsg = FindWindow(MAIN_MSG_WND_CLASS, NULL); // returns NULL for a message-only window
if (hwndNotesMsg) {
    PostMessage(hwndNotesMsg, WM_QUIT, 0, 0);
}
if (g_notesThread) {
    WaitForSingleObject(g_notesThread, 3000);   // times out — thread never got WM_QUIT
    CloseHandle(g_notesThread);                 // closes handle of a still-running thread
```

The notes thread's loop (`while (GetMessage(...))`, line 1649) has no other exit, so it never terminates: on disable the 3s wait times out, the still‑open sticky‑note windows stay on screen, and the orphaned thread lives until explorer exits. Re‑enabling then spawns a second notes thread and duplicates the notes. The same lookup is why note *creation* needs a 1‑second retry loop (line 1937‑1945) and can spawn duplicate threads. Fix: don't use `FindWindow` here at all — you already store the handle and thread id in globals. Use `g_mainMsgWnd` / `g_notesThreadId` directly (they're readable from any thread), or `FindWindowEx(HWND_MESSAGE, NULL, MAIN_MSG_WND_CLASS, NULL)`.

**3. The mod leaves a persistent file behind (reversibility).** Notes are written to `%APPDATA%\WindhawkMagicNotes.dat` (line 1273 / 1556). A mod's effects should disappear when it's disabled/removed — this file survives uninstall. Windhawk gives you a per‑mod value store that *is* cleaned up when the mod is removed; the notes blob maps naturally onto `Wh_SetBinaryValue` / `Wh_GetBinaryValue` (or serialize to a string and use `Wh_SetStringValue`). Please persist via the Windhawk store rather than an ad‑hoc file in `%APPDATA%`.

<p>

- **Unused dependencies.** `-lwinhttp` (line 9) and `#include <winhttp.h>` (line 322) are never used — no `WinHttp*` calls anywhere. `#include <tlhelp32.h>` (line 319) is also unused (no Toolhelp APIs). Remove them.
- **README documents a setting that doesn't exist.** The "Additional Settings" section lists **Enable Particle Effects**, but there's no `EnableParticles` setting — `g_settings.enableParticles` is hard‑coded to `TRUE` (line 4654). Either expose it as a real setting or drop it from the README. (`blockOtherClicks`, also hard‑coded at line 4660, is internal and fine.)
- **`LoadLibrary(L"Msftedit.dll")`** (lines 1579, 1641) loads by bare name via the default search order. Low risk from the current `explorer.exe` (protected directory) and from `windhawk.exe` if you convert, but as hardening prefer `LoadLibraryExW(L"Msftedit.dll", nullptr, LOAD_LIBRARY_SEARCH_SYSTEM32)`.
- **Redundant unregister.** `Wh_ModUninit` calls `UnregisterClass(NOTE_CLASS, ...)` (line 4799), but the notes thread already unregisters `NOTE_CLASS` and `MAIN_MSG_WND_CLASS` on exit (lines 1664‑1665), and `Wh_ModUninit` waits for that thread first — so the call is a harmless no‑op you can drop.
- **`g_settings` is rebuilt without synchronization.** `Wh_ModSettingsChanged` → `LoadSettings` runs on an arbitrary thread and rewrites the whole `g_settings` struct (resetting `gestureCount` to 0 and repopulating) while the hook thread reads it. It's guarded by the `g_gestureActive` early‑out, so this only matters for a gesture that fires exactly during a settings change — an edge case, but worth being aware of.

</p>
</details>

<p>

- **"Search Selection" destroys the clipboard.** `GetSelectedText` (line 818) synthesizes Ctrl+C and then reads `CF_UNICODETEXT`, overwriting whatever the user had on the clipboard with no restore, and waits a fixed `Sleep(50)` that can be too short for slow apps (empty result) — the classic tradeoff of the simulate‑copy approach. Consider saving and restoring the prior clipboard contents around the copy so the feature is non‑destructive.
- **Search engine is hard‑coded to Google** (`https://www.google.com/search?q=`, line 2218). A settings entry for the search URL template would make this friendlier to users of other engines.

</p>
</details>

m417z:
### Submission review

---

**1. Actions and note-creation run on the input-hook thread, which stalls system-wide input.**

The low-level mouse and keyboard hooks are installed on the `HookThreadProc` thread (lines 4603-4604), and that same thread owns `g_msgWnd` and runs the message loop that dispatches `ExecuteAction`. A low-level hook callback is only delivered while its thread is pumping messages, so any blocking work on this thread stalls all mouse/keyboard input: the OS silently drops hook events that take longer than `LowLevelHooksTimeout` (~300 ms), and shorter blocks show up as input stutter. Blocking sites:

- `HandleNoteCreationMouse` runs **directly inside** `LowLevelMouseProc` (line 3486), and on button-up it can `Sleep(50)` up to 20 times — **up to 1 second inside the hook** (lines 1978-1982).
- `ExecuteAction` is deferred via `PostMessage(g_msgWnd, …)`, but `g_msgWnd` lives on the *same* hook thread, so blocking actions still stall the hook: `GetSelectedText`'s `Sleep(100)` (line 851, hit on every *Search Selection*), the cross-process `CoCreateInstance(CLSID_ShellWindows, …, CLSCTX_LOCAL_SERVER)` in `GetExplorerPath`, and `SendMessage(target, WM_SYSCOMMAND, …)` to arbitrary foreground windows (lines 2131-2143, 2182) — `SendMessage` has no timeout, so a hung target hangs the hook thread indefinitely.

Fix: run `ExecuteAction` (and the note-creation wait) on a **dedicated worker thread with its own message loop**, separate from the hook thread; never `Sleep` inside the LL hook; and use `SendMessageTimeout` (or `PostMessage`) for the window commands so an unresponsive window can't wedge input handling.

**2. GDI pen leak in the recorder-canvas paint path.**

In `CanvasProc`'s `WM_PAINT`, `startPen` is selected into `memDC` without saving the previously-selected pen (line 2962), and `DeleteObject(startPen)` at line 2990 runs while the pen is *still* selected. `DeleteObject` fails on a selected object, so the pen handle leaks. The canvas invalidates on every mouse-move while drawing (many repaints per recording), and it runs in the long-lived `windhawk.exe` tool process, so repeated recordings steadily leak GDI handles. Fix:

```cpp
HPEN oldPen = (HPEN)SelectObject(memDC, startPen);
// … Ellipse / Polygon …
SelectObject(memDC, oldPen);
DeleteObject(startPen);
```

<p>

- **Dead code:** `SpawnSavedNotes` (line 1616) is never called — `RunNotesProcess` duplicates its logic inline. Remove it.
- **Unused dependency:** nothing calls any `Dwm*` function, so `#include <dwmapi.h>` (line 312), the `DWMWA_*` defines (lines 325-328), and `-ldwmapi` in `@compilerOptions` are all unused — drop them.
- **Dead settings fields:** `enableParticles` and `blockOtherClicks` are hardcoded to `TRUE` in `LoadSettings` (lines 4691, 4697) and never configurable — either expose them as settings or just inline the constant where they're read.
- **Missing failure handling:** the `CreateDIBSection` results in `DrawModeProc`/`ColorPickerProc` (≈lines 3967, 4205) are `memset` without a null check → crash if allocation fails; and `CreateWindowEx` isn't checked in `ToggleDrawMode`/`StartColorPicker`, so on failure `g_drawModeActive`/`g_pickerActive` stay `TRUE` with a null window (mode stuck "on", invisible).
- **Unsaved note text on unload:** the notes thread's teardown (end of `RunNotesProcess`) destroys the note windows without a final `SaveNotes()`, so text typed since the last save/move/menu action is lost when the mod is disabled. A `SaveNotes()` before the loop exits would preserve it.
- **Threading primitive:** `g_running` is a `volatile BOOL` shared across threads (line 464); `std::atomic<bool>` expresses the intent correctly (`volatile` isn't a synchronization primitive).
- `WM_CTLCOLOREDIT`'s `static HBRUSH s_br` (≈line 3103) is never freed — a trivial one-time leak.

</p>
</details>

<p>

- **Preset shapes are mis-centered in the recorder:** the layout math uses `drawArea.bottom = rc.bottom - 20` (line 2733), while the actual draw pane renders/hit-tests at `rc.bottom - 60`, so generated preset shapes land ~40 px too low relative to where you draw or paste. Cosmetic.
- **Search Selection clobbers non-text clipboard content:** `GetSelectedText` (lines 823-871) saves and restores only `CF_UNICODETEXT`. If the clipboard held an image or files, that content is replaced by the copied text and not restored.
- **Spotlight is a full-screen software fill:** `SpotlightProc` writes every pixel of the virtual screen on each 16 ms tick (lines 1804-1829). It's CPU-heavy, but there's no trivially cleaner way to do the soft punch-hole gradient, so just an FYI.
- **Fullscreen/F11 handling is window-class based:** `ToggleFullscreen` keys off a fixed set of class names (lines 1098-1104) and falls back to maximize for anything else — expected, just note that unknown apps won't get the F11 path.

</p>
</details>

iMAboud:
I have addressed all critical and optional issues. 
However, I have added 7 new actions, a fix for drawing and color picker tools. 
I did not want to apply the fixes to the older version, hopefully this does not interduce new issues.

m417z:
### Submission review

---

**1. The mod silently uploads local files to an external server.** `SearchImage` (lines 983–1005), reached from `ACTION_SEARCH_SELECTION` when the selection is an image path (lines 3204–3205), spawns hidden PowerShell that `curl`s the user's local image to `https://tmpfiles.org/api/v1/upload` and then opens the resulting **public** URL in Google Lens:

```cpp
L"$res = curl.exe -s -F \"file=@$filePath\" https://tmpfiles.org/api/v1/upload; "
...
L"Start-Process ('https://lens.google.com/uploadbyurl?url=' + [uri]::EscapeDataString($url)); "
```

Silently uploading a user's local file to an anonymous public file host is a real privacy/security problem (the file becomes reachable by anyone with the URL). Make the image-upload feature optional, disabled by default, and make it clear that images are uploaded to tmpfiles.org. The feature also depends on `curl.exe`/PowerShell, which is sub-optimal.

**2. The low-level input hooks share their thread with heavy/slow work → risk of system-wide input stalls.** Both `WH_MOUSE_LL` and `WH_KEYBOARD_LL` are installed on `HookThreadProc`'s thread (lines 5742–5743), and that same thread also:
- runs the overlay/canvas/aura rendering (the `WM_TIMER` at ~16 ms drives GDI+ `UpdateLayeredWindow` painting during a gesture), and
- handles the feature toggles posted to it (`WM_HOOK_TOGGLE_FLASH`, etc.), including `ToggleFlash` → `SaveMonitorBrightness`/`RestoreMonitorBrightness`, which call DDC/CI `GetMonitorBrightness`/`SetMonitorBrightness` **per monitor** — DDC/CI over I²C routinely takes tens to hundreds of ms per call.

A low-level hook callback must return quickly: all system input is serialized behind it, and Windows silently removes hooks that exceed `LowLevelHooksTimeout`. When this thread is busy rendering a dense trail or doing DDC/CI brightness I/O, the OS can't service the input hooks, so input across the whole system lags (and the hooks can get dropped, breaking the mod until reload). Additionally, several button-event paths do window create/destroy and a full-screen-DC `GetPixel` *inline* in `LowLevelMouseProc` (`ShowAura`/`StartGesture`/`ToggleDrawMode`/`StopColorPicker`, the color-picker `GetPixel` at line 4660). Recommend dedicating the LL-hook thread to hook servicing only (return from the callback as fast as possible, `PostThreadMessage` the real work to a separate worker/UI thread) and moving the overlay rendering and the DDC/CI brightness work off that thread.

<p>

- **Settings-reload data race.** `WhTool_ModSettingsChanged` → `LoadSettings` runs on a Windhawk-internal thread and rewrites the whole `g_settings` (it even resets `gestureCount = 0` at line 5899 before rebuilding) while the hook thread reads `g_settings` in `LowLevelMouseProc`/`MatchGesture`. The `if (g_gestureActive) return;` guard only covers an in-progress draw, not `MatchGesture` (which runs *after* `g_gestureActive` is cleared) or general reads. In practice it only matters on a settings change, but it can read torn state. Consider building into a fresh struct and publishing it with a single atomic pointer swap, or guarding with a lock.
- **`Wh_GetStringSetting` never returns NULL** — it returns `L""` on unset/error. The `if (colorStr)` / `if (fsList)` / `!name` NULL checks in `LoadSettings` are dead code; check for an empty string instead. You could also use `WindhawkUtils::StringSetting` (RAII) rather than raw `Wh_GetStringSetting` + `Wh_FreeStringSetting`.
- **Redundant second keyboard hook.** `ToggleFlash` installs a *second* `WH_KEYBOARD_LL` hook (`g_flashKeyboardHook`, line 2910) just to dismiss the flash on any key, but the global `LowLevelKeyboardProc` is already active and `FlashProc` already handles `WM_KEYDOWN`. You can dismiss flash from the existing keyboard hook instead of installing another one.

</p>
</details>

<p>

- **Flashlight brightness isn't restored on a hard exit.** DDC/CI brightness is restored on toggle-off and in `HookThreadProc` teardown, but if the dedicated tool process is killed while the flash is active (crash, `TerminateProcess`), the monitor stays at max brightness until manually toggled again. Since it's a physical-monitor change there's no clean way to guarantee restore on abnormal exit — noting it as an FYI.
- **Fullscreen detection is a rect-equality heuristic.** `IsFullscreenAppActive` treats "foreground window rect == monitor rect" as fullscreen. That catches borderless-windowed apps but also matches ordinary maximized-without-taskbar states, and won't reliably detect exclusive-fullscreen (DirectX) games, whose window bounds may not equal the monitor rect. Fine as a best-effort gate, just imperfect.
- **Wiggle activation only measures X-axis motion** (`ms->pt.x` reversals). A vertical shake won't arm gesture mode. Probably intentional, but worth documenting.

</p>
</details>

m417z:
### Submission review

---

**1. The aura window is created and destroyed across two different threads.** `UpdateModifierToggle()` runs *inside* the low-level mouse/keyboard hooks (i.e. on the hook thread) and calls `ShowAura()` / `HideAura()` directly (magic-mouse.wh.cpp:4690, :4693). Everywhere else these run on the **worker** thread via `WM_HOOK_SHOW_AURA` / `WM_HOOK_HIDE_AURA` → `MsgWndProc` (:3721, :3728). A window created on one thread can't be `DestroyWindow`'d from another — the call silently fails — and its `SetTimer` animation only pumps on the creating thread. Concrete repro: with `Modifier Behavior = Toggle`, a modifier key set, and `Wiggle = Disabled`, toggling on creates the aura on the *hook* thread (:4690); when the arm timeout fires it posts `WM_HOOK_HIDE_AURA` to the *worker* thread (:4713), whose `DestroyWindow` fails → the aura ring stays stuck on screen. It also means `CreateWindowEx` is being called synchronously inside a low-level hook callback, which is exactly the blocking UI work LL hooks should avoid (they hold up all system input until they return). Fix: don't touch windows from the hook — post `WM_HOOK_SHOW_AURA` / `WM_HOOK_HIDE_AURA` to `g_msgWnd` from `UpdateModifierToggle` just like the rest of the code already does.

**2. Physical-monitor handle leak in the Flash feature.** In `SaveMonitorBrightnessCallback` (magic-mouse.wh.cpp:2736), when `GetPhysicalMonitorsFromHMONITOR` succeeds (opening `physMon[0].hPhysicalMonitor`) but `GetMonitorBrightness` fails, `info.valid` stays `FALSE`. `RestoreMonitorBrightness` (:2757) only calls `DestroyPhysicalMonitor` for entries where `valid == TRUE`, so the opened handle is never released. Monitors that don't expose DDC/CI brightness (many laptop internal panels, some external displays) hit this on **every** Flash toggle, leaking a handle each time. Track "handle was opened" separately from "brightness was read", and always `DestroyPhysicalMonitor` an opened handle:
```cpp
if (GetPhysicalMonitorsFromHMONITOR(hMon, 1, physMon)) {
    info.hPhysical = physMon[0].hPhysicalMonitor;
    info.opened = TRUE;                       // always destroy this later
    DWORD minB = 0, curB = 0, maxB = 0;
    if (GetMonitorBrightness(info.hPhysical, &minB, &curB, &maxB)) {
        info.originalBrightness = curB; info.valid = TRUE;
        SetMonitorBrightness(info.hPhysical, maxB);
    }
}
// in RestoreMonitorBrightness: restore only if valid, but Destroy if opened
```

<p>

- **Dead/duplicate code:** `HandleNoteCreationMouse` (magic-mouse.wh.cpp:2137, the `MSLLHOOKSTRUCT*` variant) is never called — it was superseded by `HandleNoteCreationMouseWorker` (:3500), which is the one dispatched from `MsgWndProc` (:3747). The dead copy also carries its own inline `NoteSelWndProc` registration (:2143–2156). Remove it.
- **Two window classes are never unregistered:** `"MagicMouseSpotlight"` (registered :2065) and `NOTE_SEL_CLASS` (registered :2147 and again :3511) are missing from the `UnregisterClass` block at :5864–5874. This is harmless *here* because the tool-mod process calls `ExitProcess(0)` on unload (:6327), so the process dies and takes its class registrations with it — but for consistency with the other 12 classes you may as well add them.
- **Setting descriptions don't match the code clamps:** `WiggleStrength`'s `$description` says "10-200" but the code clamps to 2–200 (:5970); `GestureSensitivity`'s says "8-20" but the value isn't clamped at all (:5973). Align the text (or the clamps).
- The `ArmTimeout` setting is labelled "Gesture Mode Timeout (ms)" in the settings block but "Armed Timeout" in the README — pick one name.

</p>
</details>

<p>

- **Some work runs in the worker thread's message loop and briefly freezes its animations.** `GetSelectedText()` does `Sleep(100)` (:925), `TakeScreenshot()` does `Sleep(150)` plus up to 10×`Sleep(50)` clipboard retries (:994, :1071), and `ACTION_SLEEP_MONITOR` does `Sleep(500)` (:3028) — all on the worker thread that also drives the overlay/aura/toast/stopwatch/timer redraws. Those animations will stutter while an action with a sleep runs. Not a correctness bug, just an FYI; if it bothers you, those actions could be offloaded to a short-lived worker thread.
- **`IsFullscreenAppActive()` is called inside the LL mouse hook** (:4789), and when the fullscreen include/exclude lists are non-empty it does `OpenProcess` + `QueryFullProcessImageNameW`. It only runs on a *completed* wiggle (`g_wiggleCount >= 6`), not per mouse-move, so it's acceptable — noting it only because doing that kind of work in a low-level hook is generally something to keep an eye on.
- **Flash + crash:** the Flash feature lowers/raises physical monitor brightness via DDC/CI and restores it on normal unload (:5857). If the tool process is killed while Flash is active, the OS won't auto-restore brightness, so it stays maxed until the next Flash toggle or a manual change. Inherent to the feature; just noting it.

</p>
</details>

Issue #4711, opened 2026-07-07
[Simple Window Switcher] Switch between windows of the same program


ctlaltdefeat:
Would be cool to have an option to set a hotkey (such as Alt + `) that switches only between windows of the same program.

Mod author: @Louis047

Louis047:
@ctlaltdefeat @gevosuro27 I haven't touched this mod in a while now as I've moved to linux but have windows as dual-boot to work on the mod once I'm free. 

Also, I think you can already switch between the windows of the same program. I have already added it as a feature in v2. Please do check it out. 

Here's the setting, you just have to enable it. I chose to keep it disabled by default.

<img width="1149" height="769" alt="Image" src="https://github.com/user-attachments/assets/e997173e-3176-4f35-a595-f817fd5e6b10" />

ctlaltdefeat:
I saw that setting, but it's not quite the workflow I intended (and @gevosuro27 too I guess).
We're envisioning Alt+Tab maintaining the same behavior, but Alt + ` just cycling through the same program.
(I actually got used to this in GNOME on Linux so maybe you're using that too now haha)

Louis047:
@ctlaltdefeat I think I get you. Is it like when you are focused on a process that has multiple same windows open, Alt+` would display the window switcher only for that particular process with multiple windows directly. Am I right ?

> I actually got used to this in GNOME on Linux so maybe you're using that too now haha

Cool. And yeah, I don't use any DEs like GNOME or KDE. I'm currently using Hyprland.

ctlaltdefeat:
Yes. exactly

Louis047:
Hey @ctlaltdefeat, I have implemented the requested feature in the PR cited above. Please do test it and let me know if its the intended way. To enable it, follow the below image:

Note that the setting is under `Accessbility` section

<img width="645" height="399" alt="Image" src="https://github.com/user-attachments/assets/f2a499a0-e8ec-4ef3-9a09-72b376a77e4c" />

ctlaltdefeat:
I installed the draft PR version and tested it, works perfectly! Good to merge from my POV

Louis047:
@ctlaltdefeat thanks for the verification!

Louis047:
@ctlaltdefeat Hey, changes have been merged. You can close this issue now.

Issue #4632, opened 2026-06-30
[Simple Window Switcher] Hide close button


rudiedirkx:
Mod author: @Louis047 

I never click-close windows from the window switcher, but when accidentally the mouse is where the alt-tab menu appears, there's a Close X, sometimes in a big red button. I'd love a setting to never show X buttons, not even on hover.

I use the simplest Icon + Label layout ❤️ Just like Win XP had, and Win 10 still supported (and maybe Win 11 until some feature update completely removed it).

Absolutely love this mod, and Windhawk. Windows 11 is crazy without Windhawk.

rudiedirkx:
Claude's new config attempt seems to work (forked, compiled, tested locally):

```diff
diff --git a/simple-window-switcher.wh.cpp b/simple-window-switcher.wh.cpp
index ff60368..850268c 100644
--- a/simple-window-switcher.wh.cpp
+++ b/simple-window-switcher.wh.cpp
@@ -290,4 +290,7 @@ Additional improvements made by [Asteski](https://github.com/Asteski).
           $name: Center Task Icon and Title
           $description: Center the icon and title together in each task row.
+        - showCloseButton: true
+          $name: Show Close Button on Hover
+          $description: Show the X close button on the task under the mouse. When off, hovering never reveals a close button and clicking a task only switches to it (use middle-click to close a window instead).
       $name: Header Content
     - Orientation:
@@ -598,4 +601,5 @@ struct Settings {
     bool perMonitorWindows; bool taskRoundedCorners; bool roundThumbnailCorners; bool roundGroupIndicator; bool roundBadgeIconBackground; bool reverseScrollDirection;
     bool centerTaskContent;
+    bool showCloseButton;
     bool showApplications;
     WCHAR showTitles[32];
@@ -2829,5 +2833,5 @@ static void DrawSwitcherContent(HDC hdc, bool fillBg, HWND hWnd) {
         int btnReserve = 0;
         bool isIconOnly = !g_settings.showThumbnails && !g_settings.showTitle && g_settings.showIcon;
-        if (!HeaderIsVertical()) {
+        if (g_settings.showCloseButton && !HeaderIsVertical()) {
             if (!isIconOnly && !(g_settings.showThumbnails && ThumbnailIsSide())) {
                 btnReserve = ((g_settings.centerTaskContent) || (i == g_hoverIndex && g_hoverWnd == hWnd))
[...]

Issue #4623, opened 2026-06-30
[simple-window-switcher] Fix windows overlaping


bropines:
I fixed the problem with windows overlapping each other, as well as the correct collapsing of qt applications, for example, like Throne.

https://gist.githubusercontent.com/bropines/7a296435736cd3c4a561a453f9dd59c1/raw/9153a18477c7a65e3ad2ad4b23c007c75b043782/simple-window-switcher.wh.cpp

You can compare it with your code using `git diff`

Mod author: @Louis047

bropines:
It looked something like this 

<img width="2486" height="757" alt="Image" src="https://github.com/user-attachments/assets/41f02302-1ed9-4f7a-89a4-0609014ed1b5" />

Louis047:
hey, that's really cool! thanks for the fix. once i'm back to windows, i'll be full on track for the mod progress. currently working on my linux dotfiles and other stuff. 

did you make a PR for this? be sure to cite it so that i can check the code and test it during my free time :)

bropines:
> hey, that's really cool! thanks for the fix. once i'm back to windows, i'll be full on track for the mod progress. currently working on my linux dotfiles and other stuff.
> 
> did you make a PR for this? be sure to cite it so that i can check the code and test it during my free time :)

I'll do PR as soon as I get some of the code in order

Louis047:
> I'll do PR as soon as I get some of the code in order

I'll request to send the code instead so that I can use that as a base to work on further improvements. Thanks!

The reason is that only the related authors to the mod can make changes to their own mods as per the rules. So we have to do it this way.

Louis047:
oh instead, we can do it like this. i'll fork the windhawk-mods repo, then make a PR to that. that would make things simple.

Louis047:
You can do it [here](https://github.com/Louis047/windhawk-mods-fork)

Louis047:
@bropines Hey, still waiting for your PR in my fork.

bropines:
> [@bropines](https://github.com/bropines) Hey, still waiting for your PR in my fork.

I remember I was rewriting other projects at work, now I’ll clean up my bads and open PR

bropines:
sended

Louis047:
@bropines will test it soon 👍

Louis047:
@bropines Hey, I have merged your PR. Thanks, now I can move on to working with other bug fixes :)

Louis047:
[...]

Issue #4583, opened 2026-06-26
[Simple Window Switcher] Inaccurate titles for some apps


sjain882:
Mod source: https://github.com/ramensoftware/windhawk-mods/blob/main/mods/simple-window-switcher.wh.cpp
Mod author: @Louis047 (GitHub profile handle)

Examples:
- WhatsApp (from Microsoft Store) - `WhatsApp` becomes `WhatsApp.Root`
- Windows 10 cmd.exe - `Command Prompt` becomes `Windows Command Prompt`

This does not happen with the stock windows alt-tab menu.

Louis047:
Damn, I thought this was actually fine. Good catch!

Will be fixed in the next minor update. Should be a simple one imo. 

If you know the exact fix, PRs are welcomed.

Louis047:
@sjain882 This is fixed in the new PR, please do check it out and let me know if there's any issue.

sjain882:
Can confirm this issue is fixed on the machine I was having this issue on. Thanks!

Louis047:
@sjain882 Hey there, the changes have been finally merged. You can close this issue now.

PR #4776, opened 2026-07-13
[Simple Window Switcher] Bug Fixes and Minor Feature Additions

Louis047:
### Fixes
- Fixed an issue where the alt-tab interface flashed at the top-left of the screen on launch for single-monitor setups by utilizing absolute negative off-screen coordinates
- Fixed an issue where child popup windows or dialogs overshadowed their root parent windows in the switcher by correctly prioritizing `GW_OWNER`
- Fixed the "Scroll Pages" layout engine wrapping issue to ensure strictly computed page boundaries and alignments without incorrectly pulling from the start of the list
- Fixed the layout and spacing of overflow indicators so they no longer stick flush against the absolute screen edges
- Fixed corner scaling to properly support custom, rounded, or default corner preferences in the switcher background (Credits to @bropines)
- Fixed mirror switcher cleanup to prevent resource leaks and overlap bugs on multi-monitor setups (Credits to @bropines)
- Fixed an issue where only File Explorer tabs would close instead of the entire window by safely utilizing native `SC_CLOSE` commands (Credits to @bropines)

### Updates
- Updated the title fetching logic to prioritize native `GetWindowTextW` fallback for better title extraction and consistency
- Updated the static `...` overflow indicator into dynamic directional chevrons (`▲`, `▼`, `◀`, `▶`) depending on the layout orientation
- Updated keyboard shortcut handling to natively support `Ctrl+W` and `Del` for closing the selected window
- Updated default value for `Virutal Desktop Behavior` to show windows from all virtual displays

### Additions
- Added a new setting `sortMinimizedWindowsToEnd` to prevent game windows that minimize on focus loss from being incorrectly sorted to the end of the list (Credits to @bropines)
- Added `Alt+Backtick` shortcut support with a configurable setting to cycle backwards or filter exclusively to windows of the same active application
- Added a "Show close button" (`showCloseButton`) setting toggle under Appearance -> Thumbnails to easily hide the X button
- Added advanced scroll wheel customization options (Scroll Wheel Action & Secondary Modifiers) to allow seamless toggle between selection scrolling and full-page scrolling

### Issue References:
- #4252
- #4253
- #4262
- #4265
- #4571
- #4572
- #4583
- #4623
- #4632
- #4711

### DIFF
--- a/mods/simple-window-switcher.wh.cpp
+++ b/mods/simple-window-switcher.wh.cpp
@@ -485,7 +485,7 @@ Additional improvements made by [Asteski](https://github.com/Asteski).
       - cursorMonitor: Monitor Based on Cursor Location
     - perMonitorWindows: false
       $name: Display Windows Only from the Monitor Containing the Cursor
-    - virtualDesktopBehavior: currentOnly
+    - virtualDesktopBehavior: allDesktops
       $name: Virtual Desktop Behavior
       $description: Choose which virtual desktops to show windows from.
       $options:
@@ -3800,16 +3800,29 @@ static void SwitchToSelected() {
     if (g_settings.showApplications && g_settings.restoreAllWindows) {
         groupWindows = g_windows[g_selectedIndex].groupWindows;
     }
-    HideSwitcher();
-    for (HWND hw : groupWindows) {
-        if (IsWindow(hw) && hw != hT && IsIconic(hw)) ShowWindow(hw, SW_RESTORE);
-    }
+    
+    // Switch to the selected window FIRST while we still have foreground rights
     if (IsWindow(hT)) {
         HWND hP = GetLastActivePopup(hT);
         HWND hF = IsWindowVisible(hP) ? hP : hT;
         if (IsIconic(hF)) ShowWindow(hF, SW_RESTORE);
-        SwitchToThisWindow(hF, TRUE);
+        
+        // Forcefully attach thread input to guarantee SetForegroundWindow succeeds
+        DWORD fgThread = GetWindowThreadProcessId(GetForegroundWindow(), NULL);
+        DWORD myThread = GetCurrentThreadId();
+        if (fgThread != myThread) AttachThreadInput(myThread, fgThread, TRUE);
+        
+        SetForegroundWindow(hF);
+        
+        if (fgThread != myThread) AttachThreadInput(myThread, fgThread, FALSE);
     }
+
+    for (HWND hw : groupWindows) {
+        if (IsWindow(hw) && hw != hT && IsIconic(hw)) ShowWindow(hw, SW_RESTORE);
+    }
+    
+    // Hide the switcher AFTER we've successfully transferred focus and updated the Z-order
+    HideSwitcher();
 }
 
 // Helper: check if a window is truncated (not placed in current layout)
@@ -4775,11 +4788,11 @@ static void LoadSettings() {
     wcscpy_s(g_settings.altBacktickBehavior, v ? v : L"backward"); Wh_FreeStringSetting(v);
 
     v = Wh_GetStringSetting(L"Accessibility.virtualDesktopBehavior");
-    wcscpy_s(g_settings.virtualDesktopBehavior, v ? v : L"currentOnly");
+    wcscpy_s(g_settings.virtualDesktopBehavior, v ? v : L"allDesktops");
     Wh_FreeStringSetting(v);
     if (wcscmp(g_settings.virtualDesktopBehavior, L"currentOnly") != 0 &&
         wcscmp(g_settings.virtualDesktopBehavior, L"allDesktops") != 0) {
-        wcscpy_s(g_settings.virtualDesktopBehavior, L"currentOnly");
+        wcscpy_s(g_settings.virtualDesktopBehavior, L"allDesktops");
     }
 
     g_settings.rowHeight = Wh_GetIntSetting(L"Dimensions.rowHeight");
--- a/mods/simple-window-switcher.wh.cpp
+++ b/mods/simple-window-switcher.wh.cpp
@@ -266,6 +266,12 @@ Additional improvements made by [Asteski](https://github.com/Asteski).
         - showThumbnails: true
           $name: Show Thumbnails
           $description: Show DWM live thumbnail previews of windows.
+        - showCloseButton: true
+          $name: Show Close Button
+          $description: Show the 'X' button on hover to close windows.
+        - showOverflowIndicator: true
+          $name: Show Overflow Indicator
+          $description: Show '...' at the bottom right when there are more windows off-screen.
         - showHoverBorder: true
           $name: Show Hover Border
           $description: Show a colored border around the thumbnail when hovered.
@@ -432,11 +438,30 @@ Additional improvements made by [Asteski](https://github.com/Asteski).
       $name: Show Delay (ms)
       $description: Delay in milliseconds before showing the switcher (0 = instant).
     - scrollWheelBehavior: never
-      $name: Scroll Wheel to Change Selection
+      $name: Scroll Wheel Activation
+      $description: When the scroll wheel should be active.
       $options:
       - never: Never
       - always: Always
       - stickyOnly: Only in sticky mode
+    - scrollWheelAction: selection
+      $name: Scroll Wheel Action
+      $options:
+      - selection: Change Selection
+      - page: Scroll Pages
+    - scrollSecondaryAction: page
+      $name: Secondary Scroll Wheel Action (With Modifier)
+      $options:
+      - none: None
+      - selection: Change Selection
+      - page: Scroll Pages
+    - scrollSecondaryModifier: shift
+      $name: Secondary Scroll Wheel Modifier Key
+      $options:
+      - none: None
+      - shift: Shift
+      - ctrl: Ctrl
+      - alt: Alt
     - reverseScrollDirection: false
       $name: Reverse Scroll Direction
     - backwardShortcut: altShiftTab
@@ -446,6 +471,12 @@ Additional improvements made by [Asteski](https://github.com/Asteski).
       - altShiftTab: Alt+Shift+Tab (default)
       - altShift: Alt+Shift
       - altBacktick: Alt+Backtick
+    - altBacktickBehavior: backward
+      $name: Alt+Backtick Behavior
+      $description: Action to perform when pressing Alt+` (Backtick).
+      $options:
+      - backward: Cycle Backward
+      - sameApp: Cycle Between Windows of Current Application
     - switcherDisplayBehavior: cursorMonitor
       $name: Switcher Display Behavior
       $options:
@@ -570,7 +601,7 @@ struct WindowEntry {
     int drawnIconSz;           // Size of the drawn icon
 };
 struct Settings {
-    WCHAR theme[32]; WCHAR colorScheme[32]; WCHAR cornerPreference[32]; WCHAR scrollWheelBehavior[32]; WCHAR taskListOrientation[32]; WCHAR headerContentOrientation[32]; WCHAR iconSize[32]; WCHAR backwardShortcut[32]; WCHAR thumbnailPosition[32]; WCHAR thumbnailAlignment[32]; WCHAR switcherDisplayBehavior[32];
+    WCHAR theme[32]; WCHAR colorScheme[32]; WCHAR cornerPreference[32]; WCHAR scrollWheelBehavior[32]; WCHAR scrollWheelAction[32]; WCHAR scrollSecondaryAction[32]; WCHAR scrollSecondaryModifier[32]; WCHAR taskListOrientation[32]; WCHAR headerContentOrientation[32]; WCHAR iconSize[32]; WCHAR backwardShortcut[32]; WCHAR altBacktickBehavior[32]; WCHAR thumbnailPosition[32]; WCHAR thumbnailAlignment[32]; WCHAR switcherDisplayBehavior[32];
     WCHAR virtualDesktopBehavior[32];
     // Global theme settings (apply to both light and dark)
     WCHAR highlightStyle[32]; int opacity;
@@ -593,6 +624,8 @@ struct Settings {
     int rowWidth;
     bool stretchThumbnailsToTaskWidth;
     bool showThumbnails;
+    bool showCloseButton;
+    bool showOverflowIndicator;
     bool showTitle;
     bool showIcon;
     int maxWidthPercent;
@@ -635,6 +668,7 @@ static bool g_showAllMonitors = false;
 static HHOOK g_hMouseHook = NULL;
 static std::vector<WindowEntry> g_windows;
 static int g_selectedIndex = 0, g_hoverIndex = -1;
+static bool g_isPaginatedView = false;
 static HWND g_hoverWnd = NULL;
 static int g_layoutStartIndex = 0; // EP-style: first window index visible in the layout
 // App drill-in: when grouping by application, Ctrl drills into the selected app's
@@ -656,6 +690,7 @@ static int g_autoFitScalePct = 100;
 static int g_winW = 0, g_winH = 0;
 static int g_activePadDivider = 0;
 static bool g_hotkeysRegistered = false;
+static bool g_isAltBacktickSameApp = false;
 static HMONITOR g_hCurrentMonitor = NULL;
 static Settings g_settings;
 static HANDLE g_hSwitcherThread = NULL;
@@ -979,11 +1014,14 @@ static bool ShouldListInAltTab(HWND hwnd) {
     // WS_EX_NOACTIVATE excludes unless WS_EX_APPWINDOW is also set
     if ((ex & WS_EX_NOACTIVATE) && !(ex & WS_EX_APPWINDOW)) return false;
 
-    // Owner chain: if window has a visible, enabled owner, exclude it
-    // unless WS_EX_APPWINDOW forces inclusion
+    // Native Alt+Tab usually prefers the active popup (child dialog).
+    // However, as requested, we want to show the PARENT window instead of the child.
+    // If a window has an owner, it is a child/popup. We skip it unless it explicitly requests to be an app window.
+    // We also do not check IsWindowEnabled(own), so even if the parent is disabled by a modal popup, it still appears.
     HWND own = GetWindow(hwnd, GW_OWNER);
-    bool ownVis = IsWindow(own) && IsWindowEnabled(own) && IsReallyVisible(own);
-    if (ownVis && !(ex & WS_EX_APPWINDOW)) return false;
+    if (IsWindow(own) && IsReallyVisible(own) && !(ex & WS_EX_APPWINDOW)) {
+        return false;
+    }
 
     // Check if an ancestor in the owner chain is a tool window
     if (IsOwnerToolWindow(hwnd)) return false;
@@ -1152,8 +1190,8 @@ static BOOL CALLBACK EnumWindowsProc(HWND hWnd, LPARAM lParam) {
     }
     WindowEntry e = {};
     e.hWnd = hWnd;
-    InternalGetWindowText(hWnd, e.title, 256);
-    if (!e.title[0]) GetWindowTextW(hWnd, e.title, 256);
+    GetWindowTextW(hWnd, e.title, 256);
+    if (!e.title[0]) InternalGetWindowText(hWnd, e.title, 256);
     
     bool excluded = false;
     for (const auto& pat : g_excludeTitlePatterns) {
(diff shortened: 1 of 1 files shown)


Louis047:
Yet to fix the `CyclePage()` function. So keeping the PR as a draft now.

AromaKitsune:
File Explorer window closed instead of individual tabs: confirmed fixed.

Child windows: partially fixed. When VMware Workstation, its Preferences window, and File Explorer (or any app) are on the same virtual desktop, pressing Alt+Tab while the Preferences window is focused selects File Explorer but fails to switch to that window.
When File Explorer is on VD 1 and VMware on VD 2, pressing Alt+Tab while the Preferences window is focused switches to File Explorer, pressing Alt+Tab again selects File Explorer again and switches to the same window instead of VMware.

https://github.com/user-attachments/assets/4750b289-5438-4698-a1b0-8dca6837a60d

Louis047:
@AromaKitsune Fixed the issue, please do test it and let me know if its fully fixed 👍

AromaKitsune:
Child/Dialog windows confirmed fixed.
@sjain882 Please test this updated mod with MFC apps. I tested it only with an ATL/WTL app.

https://github.com/user-attachments/assets/a817847d-029e-4cb8-a49c-31ed4bc731b6

Louis047:
Waiting for @Asteski to add few improvements to Acrylic style.

sjain882:
> Child/Dialog windows confirmed fixed.
> @sjain882 Please test this updated mod with MFC apps. I tested it only with an ATL/WTL app.
> 
> https://github.com/user-attachments/assets/a817847d-029e-4cb8-a49c-31ed4bc731b6
> 
> 

I can confirm that child/dialog windows are fixed for VS2015 and MFC apps

Louis047:
PR is ready for review now, pushing the luminosity addition to the next release

@m417z

m417z:
### Submission review

I haven't verified all the notes, but at least some of them are worth fixing.

---

**1. The Explorer IPC window class is never destroyed or unregistered — Explorer crashes after a mod update/reload.**

[`ExplorerIpcThread`](https://github.com/ramensoftware/windhawk-mods/blob/d41b0f23c214ed5e1d7de7a92c75ad5b61aa0a2e/mods/simple-window-switcher.wh.cpp#L1438-L1457) registers `WindhawkSWS_IpcWindow` with `hInstance = GetModuleHandle(NULL)` (i.e. `explorer.exe`, which never unloads) and creates a message-only window, but neither is torn down — `Wh_ModUninit` only posts `WM_QUIT`. So after the mod is disabled/updated:

* the class registration survives with `lpfnWndProc` still pointing into the now-unmapped mod image;
* the old message-only window survives too, so on the next load `RegisterClassW` silently fails with `ERROR_CLASS_ALREADY_EXISTS`, `CreateWindowExW` succeeds against the **stale** class, and `FindWindowW(L"WindhawkSWS_IpcWindow", NULL)` in `TryGetUwpIconFromExplorer` can return the old window. Either way `SendMessageTimeoutW` dispatches into freed memory → crash in `explorer.exe`.

Keep the `HWND` and clean both up when the loop exits:

```cpp
HWND hIpcWnd = CreateWindowExW(0, L"WindhawkSWS_IpcWindow", L"", 0, 0, 0, 0, 0,
                               HWND_MESSAGE, NULL, wc.hInstance, NULL);
MSG msg;
while (GetMessageW(&msg, NULL, 0, 0)) { ... }

if (hIpcWnd) DestroyWindow(hIpcWnd);
UnregisterClassW(L"WindhawkSWS_IpcWindow", wc.hInstance);
```

(The switcher's own class is handled correctly at [L5244](https://github.com/ramensoftware/windhawk-mods/blob/d41b0f23c214ed5e1d7de7a92c75ad5b61aa0a2e/mods/simple-window-switcher.wh.cpp#L5244) — this one just got missed.)

**2. `SwitchToSelected`: "Restore All Windows" now activates the wrong window.**

The group-restore loop was moved [after](https://github.com/ramensoftware/windhawk-mods/blob/d41b0f23c214ed5e1d7de7a92c75ad5b61aa0a2e/mods/simple-window-switcher.wh.cpp#L3830-L3832) `SetForegroundWindow(hF)`. `SW_RESTORE` *activates* the window it restores, so with `Grouping.restoreAllWindows` enabled the last restored sibling ends up on top instead of the entry the user picked. Restore the group first, then activate `hF` last:

```cpp
for (HWND hw : groupWindows) {
    if (IsWindow(hw) && hw != hT && IsIconic(hw)) ShowWindow(hw, SW_RESTORE);
}
if (IsWindow(hT)) {
    HWND hP = GetLastActivePopup(hT);
    HWND hF = IsWindowVisible(hP) ? hP : hT;
    if (IsIconic(hF)) ShowWindow(hF, SW_RESTORE);
    SetForegroundWindow(hF);
}
HideSwitcher();
```

Side note on the same function: moving `HideSwitcher()` to the end doesn't have the intended effect — `SetForegroundWindow(hF)` sends `WM_ACTIVATE(WA_INACTIVE)` to `g_hSwitcher` on this very thread, and the `WM_ACTIVATE` handler calls `HideSwitcher()` re-entrantly, so the hide still happens *during* the focus transfer.

**3. Drop the `AttachThreadInput` dance in `SwitchToSelected`.**

```cpp
DWORD fgThread = GetWindowThreadProcessId(GetForegroundWindow(), NULL);
DWORD myThread = GetCurrentThreadId();
if (fgThread != myThread) AttachThreadInput(myThread, fgThread, TRUE);
```

At this point the foreground window is normally the switcher itself, so `fgThread == myThread` and the whole block is a no-op — it only ever runs in the abnormal case, which is exactly the dangerous one: attaching to a foreign thread's input queue means a hung target app can wedge the switcher thread inside `SetForegroundWindow`, and since that thread is the only one servicing the Alt+Tab hotkey, Alt+Tab stays broken until the tool process is killed. The switcher already owns the foreground when this runs, so plain `SetForegroundWindow` is sufficient.

**4. The new overflow chevrons are drawn with plain `DrawTextW` — they won't render correctly on the default theme.**

Everywhere else the mod is careful about the alpha channel on the layered (`theme: none`, the default) path: titles go through `DrawThemeTextEx` with `DTT_COMPOSITED`, and `DrawContour` uses GDI+ with an explicit `Color(255, …)`. The [indicator block](https://github.com/ramensoftware/windhawk-mods/blob/d41b0f23c214ed5e1d7de7a92c75ad5b61aa0a2e/mods/simple-window-switcher.wh.cpp#L3007-L3044) uses raw `SetTextColor` + `DrawTextW`, which leaves alpha at 0, so under `UpdateLayeredWindow(..., ULW_ALPHA)` the glyph comes out as a transparent hole rather than a chevron. Use the same pattern as the title text:

```cpp
if (g_hTheme) {
    DTTOPTS opts = { sizeof(DTTOPTS) };
    opts.dwFlags = DTT_COMPOSITED | DTT_TEXTCOLOR;
    opts.crText = g_isDarkMode ? SWS_TEXT_DARK : SWS_TEXT_LIGHT;
    DrawThemeTextEx(g_hTheme, hdc, 0, 0, text, -1, flags, &rc, &opts);
} else {
    SetTextColor(hdc, g_isDarkMode ? SWS_TEXT_DARK : SWS_TEXT_LIGHT);
    DrawTextW(hdc, text, -1, &rc, flags);
}
```

**5. The chevron target rect is too small, so the glyph gets clipped.**

The band is sized from `SWS_PAD_TOP` / `SWS_PAD_LEFT`, i.e. 7 px at 100% DPI:

```cpp
rcPrev.bottom = rcPrev.top + padTop;      // 7 px tall
rcPrev.top    += DpiScale(6, g_dpiY);
rcPrev.bottom += DpiScale(6, g_dpiY);     // ends up as rows 6..13
```

At the default 9 pt font the glyph cell is ~15 px, and `DrawTextW` clips to the rect unless `DT_NOCLIP` is set, so `▲`/`▼`/`◀`/`▶` render as slivers. There's `SWS_MASTER_PADDING` (20 px) of margin available — size the band from that (or measure with `DT_CALCRECT`) instead of from the inner content padding.

**6. `hasPrev` is wrong when the layout wraps and everything is visible.**

```cpp
bool hasPrev = (g_layoutStartIndex > 0);
bool hasNext = ((g_layoutStartIndex + visibleCount) < n);
```

`CycleLinear` can leave `g_layoutStartIndex` non-zero while the wrapping layout still shows every window (`visibleCount == n`). In that state a "more above/left" chevron is shown even though nothing is off-screen. Gate both flags on there actually being something truncated:

```cpp
bool anyTruncated = visibleCount < n;
bool hasPrev = anyTruncated && g_layoutStartIndex > 0;
bool hasNext = anyTruncated;
```

**7. `RecomputeAndReposition` destroys and recreates every mirror window on each navigation step.**

`DestroyMirrorSwitchers()` + `EnumDisplayMonitors(..., MirrorEnumProc, ...)` were added to [`RecomputeAndReposition`](https://github.com/ramensoftware/windhawk-mods/blob/d41b0f23c214ed5e1d7de7a92c75ad5b61aa0a2e/mods/simple-window-switcher.wh.cpp#L3846-L3894), which is called on every Tab/arrow press — and `CycleLinear`/`CycleDirectional` call it *inside a loop* while scrolling to reveal a truncated entry. In "All Monitors" mode that tears down and recreates N−1 top-level windows (each with `ApplyThemeToWindow`, `DwmExtendFrameIntoClientArea`, `SetActiveWindow`) per step, which is both expensive and visibly flickery. Since the mirrors only need new geometry, reposition the existing ones instead and keep the create/destroy in `ShowSwitcher`/`HideSwitcher`:

```cpp
for (HWND hMirror : g_hMirrorSwitchers) {
    if (!IsWindow(hMirror)) continue;
    MONITORINFO mInfo = { sizeof(mInfo) };
    GetMonitorInfoW(MonitorFromWindow(hMirror, MONITOR_DEFAULTTONEAREST), &mInfo);
    int mx, my; GetSwitcherPosition(mInfo.rcWork, &mx, &my);
    SetWindowPos(hMirror, HWND_TOPMOST, mx, my, g_winW, g_winH, SWP_NOACTIVATE);
}
```

**8. `ShouldListInAltTab`: dropping `IsWindowEnabled(own)` can make an app disappear from the switcher entirely.**

```cpp
if (IsWindow(own) && IsReallyVisible(own) && !(ex & WS_EX_APPWINDOW)) return false;
```

The child is now unconditionally dropped in favour of its owner — but nothing checks that the owner will actually be listed. If the owner is filtered out (`ExcludedWindows.excludeByTitle`/`excludeByExe`, `perMonitorWindows` with the owner on another monitor, or `hideMinimizedWindows` with a minimized owner), *neither* window appears and the app is unreachable from the switcher. Either keep the `IsWindowEnabled(own)` test (native Alt+Tab semantics), or only skip the child when the owner itself passes the same listing/exclusion checks.

**9. `` Alt+` `` is now unconditionally hijacked, and the new setting overlaps `backwardShortcut`.**

```cpp
case SWS_HOTKEY_ALTBACKTICK:
    if (wcscmp(g_settings.altBacktickBehavior, L"sameApp") == 0) { ... }
    else if (wcscmp(g_settings.altBacktickBehavior, L"backward") == 0 || UseAltBacktickBackward()) { isBackward = true; }
    else { return 0; }
```

`altBacktickBehavior` defaults to `backward`, so the second branch always wins and the `else` is unreachable — Alt+` now opens the switcher for everyone, including users who left `backwardShortcut` at its `altShiftTab` default and previously got nothing. There's no way to turn it off, and `backwardShortcut: altBacktick` is now redundant with it. Please add a `none` option (or fold the behaviour into `backwardShortcut`), and add the same validation clamp every other string setting in `LoadSettings` has — `altBacktickBehavior` is the only one without one.

**10. `ExplorerIpcWndProc` can wedge Explorer's IPC thread permanently.**

The Win10 AUMID fallback uses bare [`SendMessageW(hCore, WM_GETICON, ...)`](https://github.com/ramensoftware/windhawk-mods/blob/d41b0f23c214ed5e1d7de7a92c75ad5b61aa0a2e/mods/simple-window-switcher.wh.cpp#L1374-L1376) against an arbitrary app window. If that app is hung, the IPC thread blocks forever; the caller's `SMTO_ABORTIFHUNG` timeout only protects the tool process, and every later icon request then costs a full 1 s timeout. Use `SendMessageTimeoutW` with `SMTO_ABORTIFHUNG | SMTO_BLOCK` here too, as `LoadWindowIcon` already does.

<p>

* `ComputeLayout` does `LoadLibraryW(L"shcore.dll")` / `GetProcAddress` / `FreeLibrary` on **every** call ([L1775](https://github.com/ramensoftware/windhawk-mods/blob/d41b0f23c214ed5e1d7de7a92c75ad5b61aa0a2e/mods/simple-window-switcher.wh.cpp#L1775-L1781)), and `CyclePage` now calls it once per page plus twice more per wheel notch. Resolve `GetDpiForMonitor` once into a `static` (or just link `-lshcore` and call it directly — it's Win8.1+). If you keep the dynamic load, use `LoadLibraryExW(L"shcore.dll", nullptr, LOAD_LIBRARY_SEARCH_SYSTEM32)` rather than the bare name.
* `WM_SWS_SCROLL` sets `g_isPaginatedView = true` immediately before `CyclePage(dir)`, which overwrites it on its first dry run — the assignment is dead.
* `isCtrlW` uses `GetAsyncKeyState(VK_CONTROL)` while the surrounding key handling uses `GetKeyState`. In a `WM_KEYDOWN` handler `GetKeyState` is the right call (it's synchronized with the message being processed); `GetAsyncKeyState` reads live physical state and can disagree.
* The comment above `CloseSwitcherEntry` still says "graceful `WM_CLOSE`" after the switch to `SC_CLOSE`. Also, `SafeCloseWindow` is a one-line `PostMessage` wrapper with a name that promises more than it does — inlining it or renaming it to something like `PostCloseCommand` would read better.
* `Wh_GetStringSetting` never returns `NULL` (it returns `L""` on error/unset), so the `v ? v : L"default"` fallbacks throughout `LoadSettings` never fire — and the legacy `enableAltShiftForBackward` migration branch at [L4891](https://github.com/ramensoftware/windhawk-mods/blob/d41b0f23c214ed5e1d7de7a92c75ad5b61aa0a2e/mods/simple-window-switcher.wh.cpp#L4887-L4895) is therefore dead code. If you want the migration to work, test `!*v` instead. `WindhawkUtils::StringSetting` would also remove all the manual `Wh_FreeStringSetting` pairs.
* `showOverflowIndicator` isn't thumbnail-specific (the chevrons show with thumbnails off too), but it lives under Appearance → Thumbnails. Accessibility or a top-level Appearance entry would fit better.
* The README's feature list and screenshots weren't updated for the new features (overflow chevrons, `Ctrl+W`/`Del` close, Alt+` same-app cycling, page scrolling, close-button toggle). The PR body credits bropines — since Windhawk keeps a single `@author`, that credit belongs in the README next to the existing Asteski credit.

</p>
</details>

<p>

* `GetOffscreenDelayPosition` hardcoding `-9999` trades one edge case for another: on a multi-monitor layout whose virtual screen extends further left/up than that (e.g. two or three 4K displays to the left of the primary), `(-9999, -9999)` lands *inside* a real monitor and the 1×1 window flashes there. The original `SM_XVIRTUALSCREEN - 2` was arrangement-agnostic; the actual cause of the reported flash looks like `CreateWindowExW` creating the window at `0, 0`, which you've now fixed separately. Consider keeping the virtual-screen computation and using it for the initial `CreateWindowExW` position too.
* `CyclePage` runs a full `ComputeLayout` per page boundary plus two more passes, followed by `RecomputeAndReposition` (which unregisters and re-registers every DWM thumbnail). Wheel events arrive fast, so with many windows and several pages this is a lot of work per notch. Caching the page map for the lifetime of one switcher session (it only changes when the list or layout changes) would make paging much cheaper.
* The Alt+` `sameApp` filter calls `GetExeFromHwnd` — an `OpenProcess` + `QueryFullProcessImageNameW` — once per window inside the `remove_if` predicate. Note also that all UWP windows resolve to `ApplicationFrameHost.exe`, so `sameApp` will lump unrelated UWP apps together; `GetWindowGroupKey` already handles that case and could be reused here.
* Flipping `virtualDesktopBehavior` to `allDesktops` changes behaviour for every existing user on update and diverges from native Alt+Tab, which is current-desktop-only. Worth calling out in the mod description if you keep it as the default.
* The new rounded-corner `SelectClipRgn` in the layered `PaintSwitcher` path produces a hard, non-antialiased edge, while DWM's own corner rounding (still applied for the `round`/`roundSmall` preferences) is antialiased. The two may not line up perfectly; clipping only for the `custom` preference — where DWM rounding is explicitly disabled — would avoid the double rounding.

</p>
</details>

Louis047:
@m417z done

m417z:

More details:

Louis047:
@m417z I guess this is it. its repeating the same review again.

m417z:
Are you saying that the AI comments about things that are already fixed? If not, I believe "Next steps" should cover the next steps.

Louis047:
> Are you saying that the AI comments about things that are already fixed? If not, I believe "Next steps" should cover the next steps.

Most yeah, and some of them felt unnecessary while checking with Antigravity. The suggestions provided by AI two-three reviews back, are being considered as minor defect by the same AI for some reason. But the solution clearly works fine. Didn't find any rough cases either.

m417z:
Can you provide specific examples? The last review has 2 non-optional notes. Are you saying both are fixes in previous iterations?

Louis047:
Can you provide specific examples? The last review has 2 non-optional notes. Are you saying both are fixes in previous iterations?
 
Can I do it after a few days ? Currently on vacation so don't have the full details.

Louis047:
My apologies for the earlier confusion. I went through the code carefully and realized the bot was re-flagging the 2 items because I had only committed the optional polish points in the previous round while missing those two.
I’ve addressed both issues in the latest commit:
  1. Added a retry loop on `PostThreadMessage(WM_QUIT)` during uninit to prevent potential hangs if unloaded before the worker message queue initializes.
  2. Fixed mirror switcher creation and monitor resolution when a show delay is configured.
  3. Fixed an off-screen parking coordinate regression that left a 1px border artifact at `(0, 0)` during show delay.

Tested everything locally and it's working as expected. Ready for your review!

**Note**: I used AI to develop the mod the whole time and yeah addressing AI-reviews with an AI is really a pain in the ass and took me a while to get back to work on this. I'll be starting the AI review soon. 

@m417z

Louis047:
@m417z All set now 👍

mods/simple-window-switcher.wh.cpp
```diff
@@ -4808,7 +5194,7 @@ static DWORD WINAPI SwitcherThread(LPVOID lpParam) {
     DWORD dwStyle = WS_POPUP | WS_THICKFRAME | WS_CLIPCHILDREN | WS_CLIPSIBLINGS;
     DWORD exStyle = WS_EX_TOOLWINDOW | WS_EX_TOPMOST | WS_EX_LAYERED;
     g_hSwitcher = CreateWindowExW(exStyle, SWS_CLASSNAME, L"",
-        dwStyle, 0, 0, 0, 0, NULL, NULL, GetModuleHandleW(NULL), NULL);
+        dwStyle, -9999, -9999, 1, 1, NULL, NULL, GetModuleHandleW(NULL), NULL);
```
m417z, line 5197: This works in most cases, but somebody who has a large monitor top-left to the primary monitor might still see it. A more correct solution is to make the the window is hidden, not to try and guess coordinates that are far away.

mods/simple-window-switcher.wh.cpp
```diff
@@ -3703,8 +3942,8 @@ static void EnterAppGroup() {
         if (!IsWindow(hw)) continue;
         WindowEntry e = {};
         e.hWnd = hw;
-        InternalGetWindowText(hw, e.title, 256);
-        if (!e.title[0]) GetWindowTextW(hw, e.title, 256);
+        GetWindowTextW(hw, e.title, 256);
+        if (!e.title[0]) InternalGetWindowText(hw, e.title, 256);
```
m417z, line 3946: I wonder why `GetWindowTextW` is prioritized now? When is there a difference between the two?

I'd expect `InternalGetWindowText` to work in all cases, that's what Explorer does.

m417z (commented):
No blockers, I'll merge the update.

mods/simple-window-switcher.wh.cpp
Louis047: `InternalGetWindowText` does work, but as @sjain882 said in this issue: #4583, for some applications like Whatsapp, it captures the wrong title and `GetWindowTextW` seems to be the general choice for window title names.

mods/simple-window-switcher.wh.cpp
Louis047: Will be addressing this in the next update. @Asteski wants to add something new to the mod so I'll be adding the fix with his PR.

PR #4778, opened 2026-07-13
Add new mod: Desktop Marquee Fade (Async) v26.0.4

Twilight0Sparkle:
Adds a smooth, modern fade effect to your desktop selection. Features:

* **Smooth Fade Effect**: The selection rectangle fades out gracefully.
* **Highly Customizable**: Adjust animation speed and duration via mod settings.
* **Smart Opacity**: The mod automatically calculates opacity for a natural look.
* **Smart Interrupt**: If you open windows or interact with the desktop while 
  the animation is playing, the mod instantly speeds up (4x) to avoid 
  interfering with your workflow.
* **Single Instance Mode**: Prevents overlapping animations by ensuring only 
  one selection rectangle is active at a time.
* **Live System Color Customization:** Allows you to override default Windows selection colors (Fill/Border) with custom HEX values. Changes are applied instantly in memory without requiring a system reboot or Explorer restart.
* **Robust Persistence:** Automatically backs up your original Windows registry selection colors on first launch and safely restores them when the mod is disabled or uninstalled.

## Changelog

It's a new mod, so there's now changelog yet.

## Mod authorship

This mod was created by:

- - [x] Claude
- - [ ] ChatGPT
- - [x] Gemini


m417z:
### Submission review

---

**`SetSysColors` + registry writes + the `%APPDATA%\*.bin` file violate Windhawk's reversibility principle.** `ApplySystemColors` (called on every init and settings change) does three things a mod must not do:
- `SetSysColors(2, {COLOR_HIGHLIGHT, COLOR_HOTLIGHT}, ...)` changes those system colors for **every process on the desktop**, not just explorer's marquee — text selection, listbox/listview selection, hyperlinks, etc. everywhere. This is exactly the "prefer hooking the paint over a global side effect" anti-pattern.
- It writes to `HKEY_CURRENT_USER\Control Panel\Colors`, a **persistent registry change**. The registry value survives reboots and is re-applied by Windows on next logon. `Wh_ModUninit` is **not** called when explorer is killed/restarted or the machine reboots, so if the mod ever fails to run its restore path, the user's selection colors stay changed permanently until they disable the mod.
- `InitPersistentDefaults` creates `Windhawk_MarqueeFade_Defaults.bin` in `%APPDATA%` and **never deletes it** — a persistent file left behind.

  The key point: the fade overlay already uses `GetTargetFillColor()`/`GetTargetBorderColor()` directly in `PaintMarqueeDIB`, so the custom colors work on the fade **without** any of this. `SetSysColors`/registry/file exists only to recolor the *live* drag marquee (the split second before you release the button) — a marginal benefit for an enormous, non-reversible, system-wide cost. Please remove the entire `ApplySystemColors` / registry / `.bin` machinery (`ApplySystemColors`, `InitPersistentDefaults`, the `advapi32` dependency, the calls in `Wh_ModInit`/`Wh_ModSettingsChanged`/`Wh_ModUninit`). The custom-color feature keeps working for the fade overlay. If recoloring the live drag marquee is genuinely wanted, it must be done by hooking the desktop's drawing rather than global system colors — see [classic-selection-rectangle-color.wh.cpp](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/classic-selection-rectangle-color.wh.cpp#L178) (hooks `CLVDrawManager::_PaintMarquee` / `FillRectClr` in `comctl32.dll`), which is the reversible way to change the selection-rectangle color.

**The color arguments are swapped, so the mod corrupts colors even with custom colors OFF.** The call is `ApplySystemColors(GetTargetBorderColor(), GetTargetFillColor())`, but the signature is `ApplySystemColors(COLORREF hilight, COLORREF hotTracking)`. With custom colors disabled, `GetTargetBorderColor()` returns `g_defaultHotTracking` and `GetTargetFillColor()` returns `g_defaultHilight`, so enabling the mod sets `COLOR_HIGHLIGHT = g_defaultHotTracking` and `COLOR_HOTLIGHT = g_defaultHilight` — it **swaps the user's two system colors** and persists that swap to the registry. (Removing the `SetSysColors` machinery per the item above makes this moot; flagging it because it shows the color path is buggy.)

**Multiple `explorer.exe` instances → duplicate overlays / animation from the wrong process.** With `@include explorer.exe`, every explorer process runs the mod, installs its own global `WH_MOUSE_LL` hook, and creates its own overlay, so one drag can spawn several stacked fades (and the per-process `single_instance` setting can't dedupe across processes). In `WM_PRECHECK_START` the `sysListViewPid == GetCurrentProcessId()` check is only used to decide whether the click landed on an icon; when the PID doesn't match (i.e. this explorer doesn't own the desktop `SysListView32`) it *skips* the icon test and still sets `isValid = true`. Gate validity on ownership instead — if `sysListViewPid != GetCurrentProcessId()`, set `isValid = false` and bail. That confines the animation to the desktop-owning explorer, fixes the duplicate overlays, and keeps the in-process `LVM_HITTEST` icon detection working. (Because the icon hit-test needs an in-process `SendMessage` with a pointer param, this can't cleanly become a `windhawk.exe` tool mod, so gating to the owning instance is the right fix.)

**Add a screenshot/GIF to the README.** This is a purely visual effect, and the README currently has no image. A short GIF of the fade would make the mod much easier to evaluate and discover. Allowed hosts are `i.imgur.com` and `raw.githubusercontent.com`.

**The README "How to use" steps 1–3 are generic Windhawk instructions** (create mod / paste / compile) that apply to every mod; you can trim them to just the mod-specific steps.
- Version `26.0.4` and the `-v26` suffix baked into the `@id`/`@name` are unusual for a first release, and the id is permanent — consider a plain id/version if this is genuinely the first published version.

**DPI: the overlay is positioned from raw `MSLLHOOKSTRUCT` screen coordinates and drawn with a fixed 1px border.** On mixed-DPI multi-monitor setups the position/size and border thickness can be slightly off; scaling the border by the target monitor's DPI would look more consistent.

<p>

- Use `WindhawkUtils::StringSetting` (RAII) instead of the raw `Wh_GetStringSetting` + manual `Wh_FreeStringSetting`. Also, `Wh_GetStringSetting` never returns `NULL` (it returns `L""` when unset), so the `fillStr ? fillStr : L"0x..."` guards never take the fallback branch — an empty field parses to `0` (black) rather than the documented default. Check for an empty string if you want the default to apply.
- The `#pragma comment(lib, ...)` lines are redundant — the libraries are already linked via `@compilerOptions -luser32 -lgdi32 -lcomctl32 -ladvapi32` (and `advapi32` drops out entirely once the registry code is removed).
- Default mismatch: the settings block declares `initial_alpha: 55` but the code initializes `g_settingInitialAlpha{70}`. `LoadSettings` overwrites it so it's cosmetic, but worth aligning.
- The `use_defaults` setting key is a misnomer for "Accelerate if covered" — a clearer key name would help future maintenance (though note the key can't be renamed after release without resetting users' saved value).
- `Wh_ModUninit` waits only 1000 ms for the worker thread (`WaitForSingleObject(g_hThread, 1000)`) and then `CloseHandle`s and returns regardless. If it ever times out, Windhawk unloads the DLL while the worker thread is still executing mod code (its WndProcs live in the DLL) → crash. The teardown is quick, so `INFINITE` is safer here.

</p>
</details>

<p>

- The precheck is asynchronous — `WM_PRECHECK_START` is posted on `WM_LBUTTONDOWN` and `g_isStartPointValid` is read on `WM_LBUTTONUP`. A very fast press/release could read a stale/incomplete result, but the `w > 10 && h > 10` filter means such tiny gestures don't produce a marquee anyway, so it's benign in practice.
- "Accelerate if covered" speeds up whenever the foreground window's root class isn't `Progman`/`WorkerW` — i.e. essentially any time focus isn't on the bare desktop. That's most of the time, so in practice the fast path triggers frequently; worth confirming that matches the intended "another window covered it" behavior.

</p>
</details>

m417z:
### Submission review

---

Nice, well‑commented mod, and good job avoiding persistent changes — custom colors only touch the mod's own overlay, so the effect is fully reversible. A couple of things to look at before merge:

**Global `WH_MOUSE_LL` hook installed in every `explorer.exe` — a targeted subclass is the idiomatic alternative.** The worker thread installs a system‑wide low‑level mouse hook (line 561). Windhawk injects the mod into *every* `explorer.exe` instance, so every one of them installs its own global `WH_MOUSE_LL` hook and runs the async desktop discovery on each click, only to bail on the PID check in the non‑owning instances. Low‑level mouse hooks sit synchronously on the whole system's input path, and having N of them (one per explorer process) multiplies that cost. The established Windhawk pattern for this is to **subclass the desktop's `SysListView32`** with `WindhawkUtils::SetWindowSubclassFromAnyThread` and observe the band‑selection directly on that window — no global input hook, no cross‑process duplication, no async pre‑check message, and the "is the click on an icon" test becomes a cheap same‑thread `LVM_HITTEST`. [`classic-desktop-icons.wh.cpp`](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/classic-desktop-icons.wh.cpp#L340) is a close reference: it locates the desktop `SysListView32` for the current process ([`FindDesktopWindow`](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/classic-desktop-icons.wh.cpp#L340)), subclasses it ([line 310](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/classic-desktop-icons.wh.cpp#L310)), and hooks `CreateWindowExW` to re‑attach if the desktop view is (re)created later. See also the [Development tips](https://github.com/ramensoftware/windhawk/wiki/Development-tips) wiki page.

If you'd rather not do that redesign now, at minimum stop installing the hook where it can never be the desktop owner: return `FALSE` from `Wh_ModInit` in the File Explorer instances (they're launched with `/factory` / `-Embedding` on the command line — [`desktop-rect-select.wh.cpp` does exactly this check](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/desktop-rect-select.wh.cpp#L251)), so only the shell `explorer.exe` installs the global hook. Return `FALSE` if `GetShellWindow` returns a window that belongs to a different process.

<p>

- **Inconsistent `initial_alpha` default.** The settings YAML defaults `initial_alpha` to `90` (and the README table agrees), but the `$description` says *"Default is 55"*, and the atomic initializer is `g_settingInitialAlpha{55}` (line 113). The atomic initializers `g_settingInitialAlpha{55}` and `g_settingAutoCalcAlpha{true}` (line 114) also don't match their YAML defaults (`90`, `false`) — harmless because `LoadSettings()` overwrites them immediately, but confusing. Pick one value and make the YAML default, the `$description`, and the initializer agree.
- **`-lcomctl32` looks unused.** The mod only sends the `LVM_HITTEST` *message* and uses `LVHITTESTINFO` (both from `<commctrl.h>`); it doesn't call any comctl32 export, so the `-lcomctl32` link flag can be dropped. (If you adopt the subclass approach above, `WindhawkUtils` subclassing is header‑only, so it stays unnecessary.)

</p>
</details>

<p>

- **`PaintMarqueeDIB` recreates the DIB and recomputes every pixel each frame — it doesn't need to.** During a fade the geometry never changes, and the border‑to‑fill alpha ratio is constant (`fillAlpha = borderAlpha * targetFillAlpha/255`), so the whole image just scales uniformly in alpha. You can build the premultiplied bitmap **once** at animation start and then fade it by passing a decreasing `BLENDFUNCTION.SourceConstantAlpha` (255 → 0) to `UpdateLayeredWindow` on each tick, instead of `CreateDIBSection` + a full per‑pixel refill every 16 ms. For a large (e.g. full‑screen 4K) selection that avoids millions of per‑pixel writes and a multi‑MB alloc/free on every frame.
- **The auto‑opacity anchor no longer matches the manual default.** `CalculateSmartAlpha` is documented to return ~55 at the defaults "matching the manual default so toggling auto‑calc changes nothing" (lines 185‑186), but the manual default is now `90`, so toggling *Auto‑calculate Opacity* visibly changes the starting opacity — contrary to the comment. Decide the intended default and align the anchor (`kRefAlpha`) with it.
- **Non‑custom border uses `COLOR_HOTLIGHT`.** When custom colors are off, the fade border is drawn with `COLOR_HOTLIGHT` (the hyperlink/"hot" color), which isn't the color of the live band‑selection rectangle. That can produce a slight color shift at mouse‑up; `COLOR_HIGHLIGHT` (matching the fill) would hand off more seamlessly.

</p>
</details>

PR #4791, opened 2026-07-14
MacOS Minimize Animation v3.1.1: capture/renderer fixes by Potassiumuncher, Excluded programs setting

Abdullah-Masood-05:
## Summary
**The fixes in this release were contributed by @Potassiumuncher** (v1.5 of his genie engine — he's credited in the mod readme as well). I'm currently not able to actively work on this mod (away for a month or more), so this round of fixes is his work, integrated and shipped.

**Potassiumuncher's fixes:**
- **Translucent windows no longer go grey**: the capture was force-flattening `PrintWindow` output to opaque and re-premultiplying already-premultiplied pixels. It now passes premultiplied pixels through and keeps fully-transparent regions transparent, so acrylic / Mica windows keep their see-through areas during the genie.
- **Window-only capture**: snapshots render just the window's own content, so the taskbar / other windows never bleed into the animation on maximized or fullscreen apps.
- **Renderer quality**: per-primitive mask antialiasing, aliased RT modes set once at creation, larger corner radius, retuned tile bloat — smoother silhouette without seams at higher mesh resolutions.
- **~120fps frame pacer**: frames land at even intervals when per-frame cost varies (no duration change; a no-op at ≤120Hz where the DwmFlush gate already paces slower).
- **Animation tile count setting** (8–96, default 35), replacing the fixed 20x20 mesh.

**Mod-side additions:**
- **Excluded programs setting** (fixes the exclusion-list report): Windhawk's per-mod process exclusion only prevents the mod from loading *into* a process, but a window's animation is often driven from a *different* process (Explorer's taskbar, the shell) — so excluded apps still animated (reported with Windows Terminal Quake mode, where the window belongs to WindowsTerminal.exe, not powershell.exe). The new setting matches the process that *owns* the window at the single animation choke point.
- **Non-primary-monitor backdrop fallback**: on secondary displays DWM often returns the whole Mica/acrylic backdrop as fully transparent, which made window backgrounds vanish during the genie. When a majority of captured pixels are fully transparent, the capture now falls back to an opaque backdrop for that snapshot (documented in Known issues; a deeper fix needs investigation).

Readme also documents the reported (not yet reproduced) looping animation on Zen Browser as a known issue, with the exclusion setting as a workaround.

## Test plan
- [ ] Translucent File Explorer minimize/restore on the primary display — transparent regions stay transparent (no grey top)
- [ ] Translucent Explorer on a secondary display — background stays visible (dark fallback, not a hole)
- [ ] Maximized / fullscreen app minimize — no taskbar in the genie
- [ ] Excluded programs: add `WindowsTerminal.exe` → Quake-mode toggle plays no genie
- [ ] Tile count 8 / 35 / 96 — mesh renders correctly at each
- [ ] Classic style unchanged; Modern style default behavior unchanged otherwise

### DIFF
--- a/mods/macos-minimize-animation.wh.cpp
+++ b/mods/macos-minimize-animation.wh.cpp
@@ -2,7 +2,7 @@
 // @id              macos-minimize-animation
 // @name            MacOS Minimize Animation
 // @description     Smooth macOS-style genie minimize and restore (open) animations for every window.
-// @version         3.1.0
+// @version         3.1.1
 // @author          Abdullah Masood
 // @github          https://github.com/Abdullah-Masood-05
 // @include         *
@@ -21,6 +21,13 @@ it warps and flows down into the taskbar; when you restore it, it flows back out
 
 Definitely not inspired from MacOS.
 
+> **Maintenance note:** the current round of fixes - the translucent-window (grey
+> capture) fix, the taskbar capture fixes, the renderer quality / frame pacing
+> improvements and the tile-count setting - was contributed entirely by
+> **Potassiumuncher** (v1.5 of his engine). I'm not actively working on issues at
+> the moment and will be unavailable for a month or more; in the meantime his
+> contributions are keeping this mod moving - thank you!
+
 ## Animation style
 There are two genie styles, selectable in the settings (**Animation style**):
 
@@ -31,12 +38,16 @@ There are two genie styles, selectable in the settings (**Animation style**):
   different look. Kept because some people preferred it.
 
 ## Credits
-The **Modern** genie style - the Direct2D genie rendering engine (a 20x20 mesh warp
-driven by the macOS "lamp" curve), the UI Automation taskbar-button targeting, and
-the taskbar auto-hide handling - was contributed by **Potassiumuncher** -
-<https://github.com/Potassiumuncher>. As of v3.0.0 the default animation is his
-engine, integrated into this mod's hardening (multi-hook capture, flash-free cloak
-restore, safe unload, first-frame sync). The demo GIF above is also his, from his
+The **Modern** genie style - the Direct2D genie rendering engine (a configurable
+mesh warp driven by the macOS "lamp" curve), the UI Automation taskbar-button
+targeting, and the taskbar auto-hide handling - was contributed by
+**Potassiumuncher** - <https://github.com/Potassiumuncher>. As of v3.0.0 the
+default animation is his engine, integrated into this mod's hardening (multi-hook
+capture, flash-free cloak restore, safe unload, first-frame sync). The fixes in
+v3.1.1 are his as well (from v1.5 of his engine): the translucent-window capture
+fix that stops acrylic / Mica windows going grey, the window-only capture that
+keeps the taskbar out of the animation, the smoother mask / mesh rendering, the
+frame pacer, and the tile-count setting. The demo GIF above is also his, from his
 own [MacOS-Animation-for-windows](https://github.com/Potassiumuncher/MacOS-Animation-for-windows)
 repo. Huge thanks for building this, sharing the recording, and generously handing
 it all over - this mod wouldn't look nearly this good without him. The **Classic**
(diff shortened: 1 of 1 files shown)


Abdullah-Masood-05:
To be transparent about authorship: the fixes in this release (the translucent/grey capture fix, the taskbar capture fix, the renderer and pacing improvements, and the tile-count setting) were provided by @Potassiumuncher — not me. I'm not actively working on the mod right now and will be unavailable for a month or more, so I'm integrating and shipping his work with full credit (it's also stated in the mod's readme). Thanks again to him for keeping this moving.

m417z:
### Submission review

I suggest to also read the functionality notes as these seem like real issues, at least the first two.

---

- **DLL-hijack via `LoadLibraryW(L"Shcore.dll")`** (worker thread, `MacGenieAnimThread`, ~line 825 in the head file — this is **pre-existing, not introduced by this PR**, but it's in the same function the frame-pacer change touches, so it's worth fixing now). The mod is `@include *`, so it loads into *every* process, and `Shcore.dll` is **not** a [KnownDLL](https://learn.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-search-order#factors-that-affect-searching). A bare-name `LoadLibraryW` uses the default search order, which includes the target `.exe`'s own directory — so for any app that runs from an attacker-writable folder (a portable app in `Downloads`, an installer, etc.), dropping a same-named `Shcore.dll` next to it yields arbitrary code execution in that process the next time one of its windows minimizes/restores (this is exactly the [#2063](https://github.com/ramensoftware/windhawk-mods/issues/2063) pattern). Fix by restricting the search to System32:
  ```cpp
  HMODULE hShcore = LoadLibraryExW(L"Shcore.dll", nullptr, LOAD_LIBRARY_SEARCH_SYSTEM32);
  ```
  See e.g. [taskbar-auto-hide-when-maximized.wh.cpp#L1525](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/taskbar-auto-hide-when-maximized.wh.cpp#L1525). (Since this fires once per animation, loading it once and caching the module — or using `GetDpiForWindow` from user32, which needs no `Shcore` at all — would also drop the per-animation `LoadLibrary`/`FreeLibrary` churn, but that's a separate cleanup.)

<p>

- **Redundant antialiasing set.** The new `rt->SetAntialiasMode(D2D1_ANTIALIAS_MODE_ALIASED)` at render-target creation (~line 797) is immediately overridden every frame by the identical `rt->SetAntialiasMode(D2D1_ANTIALIAS_MODE_ALIASED)` inside the draw loop (~line 904), so the comment "fix the AA modes once at creation" only really holds for the *text* AA mode on the line below it. Harmless, just dead — you can drop the creation-time geometry-AA line, or drop the per-frame one and keep it at creation.

</p>
</details>

<p>

Non-critical observations about the feature behavior itself — nothing to fix.

- **Frame pacer vs. timer granularity.** `Sleep((DWORD)(remainingMs - 1.0))` in the ~120 fps pacer relies on the default system timer resolution, which without a `timeBeginPeriod(1)` in effect can be ~15 ms — so on a high-refresh (>120 Hz) display `Sleep(7)` may overshoot and the cap can land closer to ~60 fps with some jitter rather than a clean 120. Since the animation is time-based (`elapsedMs`) and `DwmFlush` is the real pace, this only affects smoothness, never duration or correctness — just noting the pacer may not be as precise as intended on fast panels.
- **Excluded apps skip the default *restore* animation.** On the minimize paths the exclusion gate re-enables DWM transitions before the real minimize runs, so an excluded app plays the normal system minimize — good. But on the restore paths (`ShowWindow_Hook` / `DefWindowProcW_Hook` `SC_RESTORE`), `ShowWindow_Original`/`DefWindowProcW_Original` already restored the window with transitions disabled *before* `StartMacGenieAnim` reaches the gate, so an excluded window just appears instantly on restore (no genie, but also no default animation). Cosmetic, and arguably fine for a window the user deliberately excluded.
- **High `tile_count` cost.** At the 96 cap the mesh is 96×96 ≈ 9.2k `CreateQuadGeo`/`FillGeometry`/`Release` calls **per frame** (vs. ~1.2k at the default 35), on top of a per-frame path-geometry rebuild. This is documented as expensive in the setting description, so no change needed — just confirming the per-frame allocation churn is real, so 64+ is genuinely a "fast machine only" choice.

</p>
</details>

Abdullah-Masood-05:
Thanks for the review! Both addressed:
 - `Shcore.dll` is now loaded with `LoadLibraryExW(..., LOAD_LIBRARY_SEARCH_SYSTEM32)` — agreed it was worth fixing while this function was being touched, given `@include *`.
 - Dropped the redundant creation-time geometry-AA set; the per-frame one in the draw loop is the effective one, and the comment now says so.

 On the functionality notes: agreed on all three — the pacer is best-effort on top of the `DwmFlush` gate (duration is time-based either way), the instant restore for excluded apps seems acceptable for windows the user deliberately opted out, and the tile-count cost is why the setting description flags 64+ as expensive. Noting again for transparency that this release's fixes were contributed by @Potassiumuncher.

Issue #4582, opened 2026-06-26
[Auto Custom Titlebar] uwpspy


littlemot:
Mod author: @Louis047 
Winver: 25H2 (OS Build 26200.8524)

not sure if this started happening after the new windows update, but its not possible anymore to target StartMenuExperienceHost.exe using UWPSpy while having the mod enabled

after trying to target, uwpspy closes itself and it makes the start menu not usable for a couple of seconds

Louis047:
Hmm, that's a really odd issue. Kinda shocks me where a mod focusing on a titlebar conflicts with a totally unrelated process. Classic Windows architecture moment imo. 

Will look into this once I get the time. 

Sorry for the very late response.

Louis047:
@littlemot i request you to drop the logs here when the issue occurs, which would make it easy for me to diagnose the issue. Currently working on SWS mod.

littlemot:
[Clipboard (1).txt](https://github.com/user-attachments/files/29986023/Clipboard.1.txt)

Louis047:
@littlemot will take a look into this once i'm done working on sws mod 👍

Louis047:
@littlemot I think I might have fixed the issue for you. I have added few UWP processes to `@exclude` and did a few tweaks. I request you to test the mod and let me know if there's no conflicts with it. 

[auto-custom-titlebar-colors.wh.cpp](https://github.com/user-attachments/files/30041582/auto-custom-titlebar-colors.wh.cpp)

Louis047:
@littlemot I experimented with something and found a proper fix to smart-exclude all UWP apps by default without the `@exclude`. Please test the code and let me know if this doesn't cause any issue. 

[actc-test.wh.cpp](https://github.com/user-attachments/files/30045441/actc-test.wh.cpp)

Louis047:
Never mind, test this. I have made it a bit smarter for excluding browsers: [actc-test.wh.cpp](https://github.com/user-attachments/files/30046234/actc-test.wh.cpp)

littlemot:
the start menu still becomes unusable for some seconds (shorter than before) but it does open in the end

logs: [Clipboard (2).txt](https://github.com/user-attachments/files/30055620/Clipboard.2.txt)

Louis047:
> the start menu still becomes unusable for some seconds (shorter than before) but it does open in the end

I think this is an usual side effect of the mods like this. DO test this version and let me know if that works fine:



[...]

PR #4797, opened 2026-07-15
[Auto Custom Titlebar Colors]: Smart ignore UWP Apps and browser windows with settings caching

Louis047:
## Changelog

- Added a new method to detect and ignore UWP apps by default
- Added classes to ignore browser windows (conflict with Zen Browser which is a special case)
- Added a toggle setting to exclude Zen related browser windows

## Issue References

- https://github.com/ramensoftware/windhawk-mods/issues/4582

**Note:** Custom colours are now enabled by default in v1.2.0 with default dark mode colors updated to `#000000` (active) and `#202020` (inactive). Make sure to update your colors if not using default colors

### DIFF
--- a/mods/auto-custom-titlebar-colors.wh.cpp
+++ b/mods/auto-custom-titlebar-colors.wh.cpp
@@ -48,9 +48,9 @@ Custom colours are only applied when the corresponding "Use Custom Colours" togg
 // ==WindhawkModSettings==
 /*
 - customColours:
-  - light: false
+  - light: true
     $name: "Light Mode"
-  - dark: false
+  - dark: true
     $name: "Dark Mode"
   $name: "Custom Colours"
 
@@ -85,6 +85,16 @@ Custom colours are only applied when the corresponding "Use Custom Colours" togg
     $name: "Use Hex Input"
     $description: "When on, use RRGGBB hex strings. When off, use separate R/G/B fields."
   - activeColour:
+    - hex: "000000"
+      $name: "Hex (RRGGBB)"
+    - r: 0
+      $name: "R (0-255)"
+    - g: 0
+      $name: "G (0-255)"
+    - b: 0
+      $name: "B (0-255)"
+    $name: "Active Window"
+  - inactiveColour:
     - hex: "202020"
       $name: "Hex (RRGGBB)"
     - r: 32
@@ -92,16 +102,6 @@ Custom colours are only applied when the corresponding "Use Custom Colours" togg
     - g: 32
       $name: "G (0-255)"
     - b: 32
-      $name: "B (0-255)"
-    $name: "Active Window"
-  - inactiveColour:
-    - hex: "323232"
-      $name: "Hex (RRGGBB)"
-    - r: 50
-      $name: "R (0-255)"
-    - g: 50
-      $name: "G (0-255)"
-    - b: 50
       $name: "B (0-255)"
     $name: "Inactive Window"
   $name: "Dark Mode Colours"
@@ -170,6 +170,9 @@ static std::mutex g_appDarkModeMutex;
 
 static std::unordered_set<HWND> g_appliedWindows;
 static std::mutex g_appliedMutex;
+
+static std::unordered_set<HWND> g_appliedCustomColorWindows;
+static std::mutex g_appliedCustomColorMutex;
 
 // -----------------------------------------------------------------------------
 // Dark mode detection
@@ -330,10 +333,10 @@ static void LoadSettings()
     BOOL useHexDark = (BOOL)Wh_GetIntSetting(L"darkMode.useHex");
     if (useHexDark) {
         WindhawkUtils::StringSetting hexActive = WindhawkUtils::StringSetting::make(L"darkMode.activeColour.hex");
-        if (!HexToColorref(hexActive.get(), &g_settings.activeDark)) g_settings.activeDark = RGB(32, 32, 32);
+        if (!HexToColorref(hexActive.get(), &g_settings.activeDark)) g_settings.activeDark = RGB(0, 0, 0);
 
         WindhawkUtils::StringSetting hexInactive = WindhawkUtils::StringSetting::make(L"darkMode.inactiveColour.hex");
-        if (!HexToColorref(hexInactive.get(), &g_settings.inactiveDark)) g_settings.inactiveDark = RGB(50, 50, 50);
+        if (!HexToColorref(hexInactive.get(), &g_settings.inactiveDark)) g_settings.inactiveDark = RGB(32, 32, 32);
     } else {
         g_settings.activeDark = RGB(
             (BYTE)std::clamp((int)Wh_GetIntSetting(L"darkMode.activeColour.r"), 0, 255),
@@ -386,6 +389,12 @@ static VOID ApplyTitleBar(HWND hWnd, BOOL isActive, BOOL allowCacheUpdate = TRUE
         }
         if (!wasApplied) return;
 
+        bool hadCustomColor;
+        {
+            std::lock_guard<std::mutex> lock(g_appliedCustomColorMutex);
+            hadCustomColor = g_appliedCustomColorWindows.erase(hWnd) != 0;
+        }
+
         BOOL dark = FALSE;
         {
             std::lock_guard<std::mutex> lockDark(g_appDarkModeMutex);
@@ -401,7 +410,7 @@ static VOID ApplyTitleBar(HWND hWnd, BOOL isActive, BOOL allowCacheUpdate = TRUE
 
         InModGuard guard;
         DwmSetWindowAttribute(hWnd, DWMWA_USE_IMMERSIVE_DARK_MODE, &dark, sizeof(dark));
-        if (!appOwnsColour) {
+        if (!appOwnsColour && hadCustomColor) {
             const COLORREF def = DWMWA_COLOR_DEFAULT;
             DwmSetWindowAttribute(hWnd, DWMWA_CAPTION_COLOR, &def, sizeof(def));
         }
@@ -433,23 +442,34 @@ static VOID ApplyTitleBar(HWND hWnd, BOOL isActive, BOOL allowCacheUpdate = TRUE
     COLORREF colour;
     {
         std::lock_guard<std::mutex> lock(g_settingsMutex);
-        useCustom = g_isDarkMode ? g_settings.useCustomDark : g_settings.useCustomLight;
-        colour = g_isDarkMode 
+        useCustom = darkMode ? g_settings.useCustomDark : g_settings.useCustomLight;
+        colour = darkMode 
             ? (isActive ? g_settings.activeDark : g_settings.inactiveDark)
             : (isActive ? g_settings.activeLight : g_settings.inactiveLight);
     }
 
     if (useCustom) {
+        {
+            std::lock_guard<std::mutex> lock(g_appliedCustomColorMutex);
+            g_appliedCustomColorWindows.insert(hWnd);
+        }
         DwmSetWindowAttribute(hWnd, DWMWA_CAPTION_COLOR,
             &colour, sizeof(colour));
         Wh_Log(L"ApplyTitleBar: hWnd=%p dark=%d active=%d colour=#%02X%02X%02X",
-            hWnd, (BOOL)g_isDarkMode, isActive,
+            hWnd, darkMode, isActive,
             GetRValue(colour), GetGValue(colour), GetBValue(colour));
     } else {
-        const COLORREF def = DWMWA_COLOR_DEFAULT;
-        DwmSetWindowAttribute(hWnd, DWMWA_CAPTION_COLOR, &def, sizeof(def));
-        Wh_Log(L"ApplyTitleBar: hWnd=%p dark=%d active=%d colour=DEFAULT",
-            hWnd, (BOOL)g_isDarkMode, isActive);
+        bool hadCustomColor;
+        {
+            std::lock_guard<std::mutex> lock(g_appliedCustomColorMutex);
+            hadCustomColor = g_appliedCustomColorWindows.erase(hWnd) != 0;
+        }
+        if (hadCustomColor) {
+            const COLORREF def = DWMWA_COLOR_DEFAULT;
+            DwmSetWindowAttribute(hWnd, DWMWA_CAPTION_COLOR, &def, sizeof(def));
+            Wh_Log(L"ApplyTitleBar: hWnd=%p dark=%d active=%d colour=DEFAULT",
+                hWnd, darkMode, isActive);
+        }
     }
 }
 
@@ -497,6 +517,10 @@ static VOID HandleWindowMessage(HWND hWnd, UINT Msg, WPARAM wParam, LPARAM lPara
             {
                 std::lock_guard<std::mutex> lock(g_appliedMutex);
                 g_appliedWindows.erase(hWnd);
+            }
+            {
+                std::lock_guard<std::mutex> lock(g_appliedCustomColorMutex);
+                g_appliedCustomColorWindows.erase(hWnd);
             }
             break;
         }
@@ -737,12 +761,18 @@ VOID Wh_ModUninit()
 {
     Wh_Log(L"[PID %d] Uninit - restoring system defaults", GetCurrentProcessId());
 
-    // Swap out the applied set atomically, then iterate over the captured snapshot.
-    // This avoids holding g_appliedMutex across DWM calls (cross-process, potentially slow).
+    // Swap out the applied sets atomically, then iterate over the captured snapshot.
+    // This avoids holding mutexes across DWM calls (cross-process, potentially slow).
     std::unordered_set<HWND> applied;
     {
         std::lock_guard<std::mutex> lock(g_appliedMutex);
         applied.swap(g_appliedWindows);
+    }
+
+    std::unordered_set<HWND> appliedCustomColors;
+    {
+        std::lock_guard<std::mutex> lock(g_appliedCustomColorMutex);
+        appliedCustomColors.swap(g_appliedCustomColorWindows);
     }
 
     for (HWND hWnd : applied) {
@@ -761,9 +791,11 @@ VOID Wh_ModUninit()
             appOwnsColour = g_appControlledWindows.count(hWnd) != 0;
         }
 
+        bool hadCustomColor = appliedCustomColors.count(hWnd) != 0;
+
         InModGuard guard;
         DwmSetWindowAttribute(hWnd, DWMWA_USE_IMMERSIVE_DARK_MODE, &dark, sizeof(dark));
-        if (!appOwnsColour) {
+        if (!appOwnsColour && hadCustomColor) {
             const COLORREF def = DWMWA_COLOR_DEFAULT;
             DwmSetWindowAttribute(hWnd, DWMWA_CAPTION_COLOR, &def, sizeof(def));
         }
--- a/mods/auto-custom-titlebar-colors.wh.cpp
+++ b/mods/auto-custom-titlebar-colors.wh.cpp
@@ -7,6 +7,11 @@
 // @github          https://github.com/Louis047
 // @include         *
 // @exclude         devenv.exe
+// @exclude         systemsettings.exe
+// @exclude         applicationframehost.exe
+// @exclude         startmenuexperiencehost.exe
+// @exclude         searchhost.exe
+// @exclude         shellexperiencehost.exe
 // @compilerOptions -ldwmapi -luser32
 // ==/WindhawkMod==
 
@@ -39,8 +44,9 @@ Combines automatic dark/light titlebar switching with per-mode, per-state custom
 Custom colours are only applied when the corresponding "Use Custom Colours" toggle is enabled.
 
 ## Notes
-- UWP/WinUI windows (`ApplicationFrameWindow`, WinUI 3, XAML islands) are automatically detected and skipped to avoid conflicts
-- Windows whose caption colour is set by the application itself have that colour left untouched; only the dark/light mode is kept in sync
+- Custom colours are enabled by default in v1.2.0 (`#FFFFFF`/`#E6E6E6` for light mode, `#000000`/`#202020` for dark mode)
+- UWP/WinUI windows (`ApplicationFrameWindow`, WinUI 3, XAML islands) and core shell hosts are automatically skipped to avoid conflicts
+- Windows whose caption colour is set by the application itself while the mod is loaded have that colour left untouched; only the dark/light mode is kept in sync
 - Visual attributes apply seamlessly on window activation and theme changes
 */
 // ==/WindhawkModReadme==
--- a/mods/auto-custom-titlebar-colors.wh.cpp
+++ b/mods/auto-custom-titlebar-colors.wh.cpp
@@ -38,9 +38,7 @@ Custom colours are only applied when the corresponding "Use Custom Colours" togg
 
 ## Notes
 - UWP/WinUI windows and Mozilla-based browsers are now auto-detected and skipped
-- Window redraws are debounced (50ms minimum) to prevent interference with window managers
-- Visual redraws (SetWindowPos) only occur during window activation, not deactivation, to avoid focus-grabbing issues with tiling window managers
-- No forced repaint is issued while a mouse button is held (prevents drag-state corruption)
+- Visual attributes (`DWMWA_CAPTION_COLOR` and `DWMWA_USE_IMMERSIVE_DARK_MODE`) apply seamlessly on window activation and theme changes
 */
 // ==/WindhawkModReadme==
 
@@ -113,6 +111,7 @@ Custom colours are only applied when the corresponding "Use Custom Colours" togg
 
 #include <windows.h>
 #include <dwmapi.h>
+#include <windhawk_utils.h>
 #include <unordered_set>
 #include <unordered_map>
 #include <mutex>
@@ -133,7 +132,7 @@ struct ModSettings {
     COLORREF inactiveLight;
     COLORREF activeDark;
     COLORREF inactiveDark;
-    BOOL excludeMozilla;
+    std::atomic<BOOL> excludeMozilla;
 };
 static ModSettings g_settings;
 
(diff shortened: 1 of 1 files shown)


Louis047:
Keeping this as a draft until the user confirm it works.

Louis047:
The user has confirmed that it works, but with a very minor caveat where they have to wait a second or two to let it work normally

m417z:
### Submission review

---

**1. `Wh_ModUninit` wipes the caption color of app-controlled windows.** The whole point of the new `g_appControlledWindows` set is to *never touch* windows whose owning app set `DWMWA_CAPTION_COLOR` itself. But `UninitEnumWindowsProc` doesn't consult that set — it resets `DWMWA_CAPTION_COLOR` to `DWMWA_COLOR_DEFAULT` (and clears immersive dark mode) on every eligible window:

```cpp
// UninitEnumWindowsProc — mods/auto-custom-titlebar-colors.wh.cpp:663
if (!IsWindowEligible(hWnd)) return TRUE;
BOOL off = FALSE;
DwmSetWindowAttribute(hWnd, DWMWA_USE_IMMERSIVE_DARK_MODE, &off, sizeof(off));
const COLORREF def = DWMWA_COLOR_DEFAULT;
DwmSetWindowAttribute(hWnd, DWMWA_CAPTION_COLOR, &def, sizeof(def));
```

So when the mod is disabled/reloaded, any window the app itself colored gets its color blown away and reset to the system default — the opposite of "leave it alone". Since many apps set their caption color only once (at window creation), it won't come back until the app restarts. Skip windows in `g_appControlledWindows` here (and ideally only reset windows the mod actually colored). This also makes the unload non-reversible for those windows, which is a core Windhawk principle.

**2. `g_appControlledWindows` / `seenWindows` grow forever and key on recyclable `HWND` values.** Neither set ever removes an entry when a window is destroyed:

```cpp
static std::unordered_set<HWND> g_appControlledWindows;   // never erased
...
static std::unordered_set<HWND> seenWindows;              // never erased
```

Two consequences in a long-lived, broadly-injected (`@include *`) process like `explorer.exe`:
- **Stale-HWND false positives:** Windows recycles `HWND` values. Once a tracked window is destroyed, a brand-new, unrelated window can be created with the same handle value — it will be found in `g_appControlledWindows` and silently never colored (item stays until the process dies).
- **Unbounded growth:** both sets accumulate one entry per window ever seen, for the life of the process.

You already hook `DefWindowProcW/A`, so `HandleWindowMessage` can handle `WM_NCDESTROY` and `erase(hWnd)` from both sets to bound the lifetime and fix the reuse hazard.

<p>

- **The two branches in `ApplyTitleBar` are near-identical.** After `Wh_ModInit` succeeds, `DwmSetWindowAttribute_orig` is always non-null, so the `else` branch (lines 380–399) that calls the hooked `DwmSetWindowAttribute` directly is effectively dead and duplicates the whole color-selection block. Since `g_inMod` already guards the hook from re-tracking the mod's own calls, you can drop the branch and just call `DwmSetWindowAttribute` (or `_orig`) once — the `g_inMod` flag makes either safe.

- **Stale README note.** The `## Notes` section still says *"`systemsettings.exe` and `applicationframehost.exe` are excluded to avoid conflicts"* (line 40), but this PR removed those `@exclude` lines in favor of runtime detection. Update or drop that line.

- **`g_settings` is written on the settings-change thread and read on window threads without synchronization.** It's a POD struct, so the worst case is one repaint using a half-updated struct (e.g. `useCustom` toggled but a color not yet stored) right after a settings save — no crash, self-corrects on the next paint. Only worth a mention since the pre-cache code read each value atomically per call.

</p>
</details>

<p>

- **The `DwmGetWindowAttribute` "already-colored" fallback almost certainly never fires.** `DWMWA_CAPTION_COLOR` is a *set-only* attribute — `DwmGetWindowAttribute` returns `E_INVALIDARG` for it on current Windows, so `SUCCEEDED(...)` is false and the inner block (and its `Wh_Log`) never runs:

  ```cpp
  // ApplyTitleBar — mods/auto-custom-titlebar-colors.wh.cpp:349
  if (SUCCEEDED(DwmGetWindowAttribute(hWnd, DWMWA_CAPTION_COLOR, &currentColor, sizeof(currentColor)))) {
      if (currentColor != DWMWA_COLOR_DEFAULT) { ... }   // unreachable
  }
  ```

  That means an app which set its own caption color *before* the mod loaded (mid-session enable) isn't detected via this path — only the `DwmSetWindowAttribute` hook (which catches sets *after* the hook is installed) works. Easy to confirm: enable logging and check whether the "already has custom … ignoring" line ever appears. If it doesn't, the `seenWindows` set + this block are pure overhead and can be removed.

- **`EnumChildWindows` runs on every `ApplyTitleBar` call, in every process.** Each eligibility check that *doesn't* short-circuit on `MozillaWindowClass` walks the full child-window tree calling `GetClassNameW` per descendant, and `ApplyTitleBar` fires on `WM_NCACTIVATE`, `WM_ACTIVATE`, `CreateWindowEx`, etc. across `@include *`. For most windows the tree is small so it's fine, but for complex apps this repeats a full subtree walk on every activation. If it shows up in practice, consider caching a per-`HWND` eligibility verdict (invalidated on `WM_NCDESTROY`, alongside item 2 above).

- **The `DesktopWindowXamlSource` check is broad.** A growing number of otherwise-normal Win32 apps host XAML islands (`DesktopWindowXamlSource`) for parts of their UI while still using a standard caption. Those will now be skipped entirely. That's likely the intended trade-off for the crash fix, but worth being aware that the excluded set is wider than "UWP apps" — it's "anything hosting a XAML island / CoreWindow".

</p>
</details>

Louis047:
Noted, will work on this when I have the time.

Louis047:
@m417z Hope this settles it

m417z:
### Submission review

There are additional notes regarding the mod functionality. I didn't verify them, but they might be worth fixing. Take a look, if you think the mod is ready as is, let me know.

---

**1. The eligibility cache is poisoned at window creation — the structural UWP/WinUI detection never fires for windows created after the mod loads.** `CreateWindowExW/A_hook` calls `ApplyTitleBar` immediately after the top-level window is created ([line 594](https://github.com/ramensoftware/windhawk-mods/blob/3667965a337db7a66d130ceb08ee062be2cee9ce/mods/auto-custom-titlebar-colors.wh.cpp#L594)), at which point the window has **no children yet**. `IsWindowEligible` runs `EnumChildWindows`, finds no `CoreWindow`/`DesktopWindowXamlSource`/`DesktopChildSiteBridge`, and caches `isEligible = TRUE` ([lines 225–233](https://github.com/ramensoftware/windhawk-mods/blob/3667965a337db7a66d130ceb08ee062be2cee9ce/mods/auto-custom-titlebar-colors.wh.cpp#L225-L233)) — an entry that's only evicted at `WM_NCDESTROY`. When the bridge child appears moments later, every subsequent check (`WM_NCACTIVATE`, theme changes, …) hits the stale `TRUE` and the mod colors the window anyway. Net effect: the detection only works for windows that already exist at `Wh_ModAfterInit` time; any UWP/WinUI 3 window opened afterwards is treated as a normal window — reintroducing exactly the conflict from issue 4582. (The create-time `ApplyTitleBar` also colors the window once before detection could possibly succeed, which likely explains the "wait a second or two" caveat the reporter mentioned.)

Suggested fix, in two parts:
- Don't populate the cache from the `CreateWindowEx` path — pass a "don't cache" flag (or skip the `ApplyTitleBar` eligibility caching there), so the first activation-time check, which runs when the children exist, is the one that gets cached.
- Additionally, invalidate on child creation: in the `CreateWindowEx` hooks, `GetClassNameW` the new window, and if it's one of the bridge classes, mark `GetAncestor(hWnd, GA_ROOT)` ineligible in the cache — and reset that root's `DWMWA_CAPTION_COLOR` to `DWMWA_COLOR_DEFAULT`, since the create-time apply may already have colored it. Note this doesn't cover the UWP `ApplicationFrameHost` case, where the `CoreWindow` is created in the app's process and reparented into the frame window cross-process — so the first bullet (defer caching to activation time) is the important one; the child-creation invalidation covers same-process XAML/WinUI 3 windows.

**2. Dialogs never reach `HandleWindowMessage`, so their `WM_NCDESTROY` cleanup never runs.** The `DefDlgProcW/A` hooks exist precisely because dialogs don't route through the exported `DefWindowProcW/A` — but they only special-case `WM_NCACTIVATE` ([lines 553–576](https://github.com/ramensoftware/windhawk-mods/blob/3667965a337db7a66d130ceb08ee062be2cee9ce/mods/auto-custom-titlebar-colors.wh.cpp#L553-L576)). For every dialog, the `g_eligibilityCache` / `g_appControlledWindows` entries are therefore never erased, which brings back the two hazards from the previous review round for this window class: unbounded growth, and stale entries on recycled `HWND` values (a new unrelated window reusing a destroyed dialog's handle inherits its cached eligibility / app-controlled status). Dialogs also miss the `WM_ACTIVATE` / `WM_SETTINGCHANGE` / `WM_DWMCOLORIZATIONCOLORCHANGED` handling. Fix: call `HandleWindowMessage(hWnd, Msg, wParam, lParam)` from both `DefDlgProc` hooks and drop the duplicated `WM_NCACTIVATE` block (double-handling in case both hooks fire for a message is harmless — the erase is idempotent and `ApplyTitleBar` is repeat-safe).

**3. `Wh_ModUninit` still force-resets `DWMWA_USE_IMMERSIVE_DARK_MODE` to `FALSE` on windows whose app set it itself.** `g_appControlledWindows` only tracks `DWMWA_CAPTION_COLOR`, but many modern Win32 apps (terminals, editors, Chromium/Electron apps) opt into dark titlebars themselves via this same attribute, typically once at window creation. Disabling the mod while the system is in dark mode flips those titlebars to light until the app restarts ([lines 677–678](https://github.com/ramensoftware/windhawk-mods/blob/3667965a337db7a66d130ceb08ee062be2cee9ce/mods/auto-custom-titlebar-colors.wh.cpp#L677-L678)) — the same reversibility problem as the caption-color half you already fixed. Suggested fix: in `DwmSetWindowAttribute_hook`, also track app-initiated sets of `DWMWA_USE_IMMERSIVE_DARK_MODE` (e.g. an `HWND → BOOL` map of the app's last value, cleaned up at `WM_NCDESTROY` like the others), and at uninit restore that recorded value instead of unconditionally writing `FALSE`.

<p>

- **Stale README note (carried over from the previous review).** The `## Notes` section still says *"`systemsettings.exe` and `applicationframehost.exe` are excluded to avoid conflicts"* ([line 40](https://github.com/ramensoftware/windhawk-mods/blob/3667965a337db7a66d130ceb08ee062be2cee9ce/mods/auto-custom-titlebar-colors.wh.cpp#L40)), but those `@exclude` lines are gone. While updating it, consider documenting the new behavior instead: UWP/WinUI windows and Mozilla-based browsers are now auto-detected and skipped.
- **`-luxtheme` is unused.** `uxtheme.dll` is only accessed via `GetModuleHandleW` + `GetProcAddress`; nothing links against it, so the `@compilerOptions` flag can be dropped.
- **The debounce comment doesn't match the code.** The comment says "track last SetWindowPos time per window", but `g_lastSetWindowPosTime` is a single global shared by all windows (and written from multiple threads without synchronization) — two windows activating within 50 ms means the second silently skips its frame refresh. Either make it per-window or fix the comment; making the variable `std::atomic<DWORD>` would also address the benign data race.
- **`g_isDarkMode` is written from arbitrary window threads** (`WM_SETTINGCHANGE` handling) and read everywhere; `std::atomic<BOOL>` would make that clean.
- **`EnumWindowsProc` can be simplified**: the `if`/`else` reduces to `ApplyTitleBar(hWnd, isActive, forceRedraw && isActive)`, and the `GetAncestor(GA_PARENT)` check is redundant — `EnumWindows` only enumerates top-level windows.

</p>
</details>

<p>

- **The `GetAsyncKeyState(VK_LBUTTON)` gate suppresses the refresh for the most common activation path.** When a user clicks a window to activate it, the button is still down when `WM_ACTIVATE`/`WM_NCACTIVATE` arrive, so the `SetWindowPos` refresh essentially only ever runs for keyboard activation (Alt-Tab). In practice `DWMWA_CAPTION_COLOR` changes take effect without `SWP_FRAMECHANGED`, so this mostly doesn't show — which raises the question of whether the `SetWindowPos` (and the debounce machinery around it) is needed at all. Worth testing without it. Also, the README note documenting the mouse-button suppression was removed in this PR while the code behavior remains.
- **`MozillaWindowClass` excludes every Gecko browser, not just Zen.** Firefox and Floorp users who previously had titlebar coloring working will silently lose it with this update. If the conflict is specific to Zen's custom styling, consider making the browser exclusion a setting (default on is fine) so users can opt back in. Relatedly, `MozillaCompositorWindowClass` in the child scan is redundant — a compositor child only exists under a `MozillaWindowClass` top-level, which the fast check already rejects.

</p>
</details>

Louis047:
@m417z Done implementing the requested changes 👍

m417z:
### Submission review

1 seems correct and is easy to fix. I haven't verified 2 and 3. Optional notes seem to have some things worth improving too.

---

**1. Deadlock: `SetWindowPos` is called while `g_lastSetWindowPosMutex` is held.** [Lines 412-424](https://github.com/ramensoftware/windhawk-mods/blob/ada41988e82fcd1dd9839de53aa42f3060a73536/mods/auto-custom-titlebar-colors.wh.cpp#L412-L424) take the lock and then call `SetWindowPos` inside the guarded scope:

```cpp
std::lock_guard<std::mutex> lock(g_lastSetWindowPosMutex);
DWORD lastTime = g_lastSetWindowPosTime[hWnd];
...
    SetWindowPos(hWnd, nullptr, 0, 0, 0, 0, SWP_FRAMECHANGED | ...);
```

`SetWindowPos` on a window owned by **another thread** sends `WM_WINDOWPOSCHANGING` / `WM_NCCALCSIZE` / `WM_WINDOWPOSCHANGED` synchronously to that thread and blocks until it dispatches them. That thread, meanwhile, can easily be blocked on the same mutex — e.g. inside `HandleWindowMessage`'s `WM_NCDESTROY` erase ([lines 484-487](https://github.com/ramensoftware/windhawk-mods/blob/ada41988e82fcd1dd9839de53aa42f3060a73536/mods/auto-custom-titlebar-colors.wh.cpp#L484-L487)) or in its own `ApplyTitleBar` debounce block. A thread waiting on a `std::mutex` does not pump messages, so both sides hang: the app's UI freezes and the process has to be killed.

This is reachable in normal use with `@include *` — any process with more than one UI thread (Explorer is the classic case, each window on its own thread) hits it when one thread runs `ApplyToAllWindows` / `ApplyToForegroundWindow` (theme change, `Wh_ModAfterInit`, `Wh_ModSettingsChanged`) while another is activating or destroying a window.

Fix — decide under the lock, call outside it:

```cpp
if (forceRedraw && !(GetAsyncKeyState(VK_LBUTTON) & 0x8000)) {
    DWORD currentTime = GetTickCount();
    bool doRedraw = false;
    {
        std::lock_guard<std::mutex> lock(g_lastSetWindowPosMutex);
        DWORD lastTime = g_lastSetWindowPosTime[hWnd];
        if (lastTime == 0 || currentTime - lastTime >= SETWINDOWPOS_DEBOUNCE_MS) {
            g_lastSetWindowPosTime[hWnd] = currentTime;
            doRedraw = true;
        }
    }
    if (doRedraw) {
        SetWindowPos(hWnd, nullptr, 0, 0, 0, 0,
                     SWP_FRAMECHANGED | SWP_NOMOVE | SWP_NOSIZE |
                     SWP_NOZORDER | SWP_NOOWNERZORDER);
    }
}
```

(The same rule as the subclass-API case in the [development tips](https://github.com/ramensoftware/windhawk/wiki/Development-tips): never make a call that sends a message to another thread while holding a lock that thread can also take.) Note that dropping the `SetWindowPos` refresh altogether — which was suggested last round and is still worth testing, see the functionality notes — would remove this whole code path along with the deadlock.

**2. UWP windows are still colored once, and the color is never taken back.** The create-time apply at [line 622](https://github.com/ramensoftware/windhawk-mods/blob/ada41988e82fcd1dd9839de53aa42f3060a73536/mods/auto-custom-titlebar-colors.wh.cpp#L622) runs when the window has no children yet, so it sets both `DWMWA_USE_IMMERSIVE_DARK_MODE` and `DWMWA_CAPTION_COLOR` on every new top-level window, including UWP/WinUI ones. The new child-creation invalidation only fires when the bridge child is created **in the same process** — for the `ApplicationFrameHost.exe` case the `CoreWindow` is created in the app's process and reparented in, so that path never runs. The later activation-time check then correctly reports the window as ineligible, but `ApplyTitleBar` only *returns early* — it never undoes what was already applied, so the mod's caption color and dark-mode attribute stay on the window for its whole lifetime. That's the conflict from issue 4582 (and matches the reporter's "wait a second or two" caveat, except here it never resolves).

Two things would fix it:
- Keep a set of windows the mod has actually applied to, and when a window is first found **ineligible**, revert it (caption colour → `DWMWA_COLOR_DEFAULT`, dark mode → the value recorded in `g_appDarkModeWindows` if any) and drop it from the set. This covers XAML islands, WinUI 3 and the cross-process UWP case uniformly.
- Additionally, `ApplicationFrameWindow` (the UWP frame class) can be rejected by the same cheap top-level class-name check you already do for `Mozilla*` in `IsWindowEligible`, which avoids ever touching those windows in the first place.

**3. The mod records its own `DwmSetWindowAttribute` call as app-controlled.** The reset in the `CreateWindowEx` hooks ([line 619](https://github.com/ramensoftware/windhawk-mods/blob/ada41988e82fcd1dd9839de53aa42f3060a73536/mods/auto-custom-titlebar-colors.wh.cpp#L619) and [line 653](https://github.com/ramensoftware/windhawk-mods/blob/ada41988e82fcd1dd9839de53aa42f3060a73536/mods/auto-custom-titlebar-colors.wh.cpp#L653)) is not wrapped in `g_inMod`, unlike the calls in `ApplyTitleBar`, so `DwmSetWindowAttribute_hook` inserts that window into `g_appControlledWindows` — the mod flags itself as the app. Today the effect is masked because those windows are also marked ineligible in the cache, but the entry is wrong and it will actively break the revert logic from item 2 (the revert would immediately mark the window app-controlled). Set `g_inMod` around these calls too — an RAII guard next to `ApplyTitleBar`'s usage would keep it from being forgotten again:

```cpp
struct InModGuard {
    InModGuard() { g_inMod = true; }
    ~InModGuard() { g_inMod = false; }
};
```

<p>

- **The two `CreateWindowEx` hooks duplicate the whole bridge-class block, and both miss atom-registered classes.** `lpClassName` is skipped when `IS_INTRESOURCE(lpClassName)` is true, which is the case whenever the caller passes the `ATOM` returned by `RegisterClass` — a fairly common pattern, and one you can't rule out for the XAML bridge classes. Calling `GetClassNameW(hWnd, ...)` on the freshly created window instead works for both cases and lets both hooks share one helper:

  ```cpp
  static void HandleCreatedWindow(HWND hWnd) {
      WCHAR className[256];
      if (GetClassNameW(hWnd, className, 256) &&
          (wcscmp(className, L"DesktopWindowXamlSource") == 0 ||
           wcscmp(className, L"Windows.UI.Core.CoreWindow") == 0 ||
           wcscmp(className, L"Microsoft.UI.Content.DesktopChildSiteBridge") == 0)) {
          ...
      }
      ApplyTitleBar(hWnd, FALSE, FALSE, FALSE);
  }
  ```

  This also removes the `MultiByteToWideChar` round-trip in the A variant.

- **`g_settings.excludeMozilla` is read outside `g_settingsMutex`** in `IsWindowEligible` ([line 234](https://github.com/ramensoftware/windhawk-mods/blob/ada41988e82fcd1dd9839de53aa42f3060a73536/mods/auto-custom-titlebar-colors.wh.cpp#L234)), while every other `g_settings` access is guarded. Harmless in practice for a `BOOL`, but it's an inconsistency — either read it under the lock or make it `std::atomic<BOOL>`.

- **`WindhawkUtils::StringSetting` instead of `Wh_GetStringSetting` + `Wh_FreeStringSetting`.** The four hex reads in `LoadSettings` each pair a get with a manual free; the RAII wrapper (`#include <windhawk_utils.h>`) removes the chance of a missed free:

  ```cpp
  WindhawkUtils::StringSetting hexActive = WindhawkUtils::StringSetting::make(L"lightMode.activeColour.hex");
  if (!HexToColorref(hexActive, &g_settings.activeLight)) g_settings.activeLight = RGB(255, 255, 255);
  ```

- **`XamlSearchContext` wraps a single `BOOL`** — passing `(LPARAM)&hasXaml` directly would do, and `thread_local bool g_inMod` is the only global here without `static` linkage.

- **`|| lastTime == 0` in the debounce is redundant** — with `lastTime == 0`, `currentTime - lastTime` is already `>= 50` for any `GetTickCount()` past the first 50 ms of uptime.

- **No screenshot in the README.** This is a visual mod; a small before/after image (Imgur or `raw.githubusercontent.com`) would make the catalog entry a lot clearer.

</p>
</details>

<p>

- **The `SetWindowPos` refresh is still worth removing entirely.** As noted last round, the `GetAsyncKeyState(VK_LBUTTON)` gate means the refresh is skipped for mouse-driven activation (the button is still down when `WM_ACTIVATE`/`WM_NCACTIVATE` arrive), so it effectively only runs for Alt-Tab — and `DWMWA_CAPTION_COLOR` takes effect without `SWP_FRAMECHANGED` anyway. There's also a side effect worth knowing about: `GetAsyncKeyState` consumes the "pressed since the last call" bit (the low bit of the return value) for whoever asks next, so polling it from inside the host app's own thread can make the app's own `GetAsyncKeyState(...) & 1` checks miss a click. Dropping the `SetWindowPos` call, the debounce map and the key-state probe would simplify the mod considerably and remove item 1 above.

- **`WM_SETTINGCHANGE` is never matched on the ANSI path.** USER32 hands an ANSI `lParam` string to ANSI window procs, so in `DefWindowProcA_hook` the `wcscmp((LPCWSTR)lParam, L"ImmersiveColorSet")` comparison in [`HandleWindowMessage`](https://github.com/ramensoftware/windhawk-mods/blob/ada41988e82fcd1dd9839de53aa42f3060a73536/mods/auto-custom-titlebar-colors.wh.cpp#L516-L519) can't match (it's safe — it fails on the first wide character — just dead). Apps with ANSI window procs therefore only pick up theme changes via the `lParam == NULL` case. If you care about them, pass the message's character set down to `HandleWindowMessage` and use `strcmp` for the A variants.

- **The eligibility verdict is cached for the window's whole lifetime.** Once a window is checked and found eligible, that answer sticks until `WM_NCDESTROY`. An app that creates a XAML island later in the window's life (e.g. lazily, for a settings pane) keeps the stale "eligible" verdict. Caching only the *ineligible* verdict, or re-checking until the window has been shown at least once, would be more robust — same root cause as item 2.

- **Mid-session enable still loses the app's original dark-mode state.** `g_appDarkModeWindows` only records `DWMWA_USE_IMMERSIVE_DARK_MODE` values set *after* the hook is installed, so for windows that already existed when the mod was enabled, `Wh_ModUninit` still writes `FALSE` and flips app-managed dark titlebars to light until the app restarts. Not much can be done (the attribute isn't readable back), but it's worth being aware of — it mostly affects enabling the mod mid-session rather than at process start.

</p>
</details>

Louis047:
@m417z done

m417z:

More details:

Louis047:
Added support for MDI frame windows (`DefFrameProcW/A`), which fixes titlebar activation and color switching for MMC snap-ins (Event Viewer, Device Manager, etc.).

Regarding the latest review:
- **@exclude list:** Skipped re-adding the hardcoded process blacklist (`startmenuexperiencehost.exe`, etc.) in favor of the dynamic class detection (`ApplicationFrameWindow`, `CoreWindow`, `WinUIDesktopWin32WindowClass`, XAML bridges), which already filters out UWP/WinUI surfaces across all hosts without needing an executable-specific exclude list.
- **Creation map cleanup:** Removed the map purge at `HandleCreatedWindow` so custom titlebar colors configured during `WM_CREATE` aren't wiped prematurely (`WM_NCDESTROY` handles cleanup at window destruction).
- **UXTheme fallback:** Switched ordinal 138 to 132 (`ShouldAppsUseDarkMode`) to match the `AppsUseLightTheme` app theme check.

Louis047:
Summary of updates:

1. **Targeted DWM Caption Color Resets:**
   Added tracking via `g_appliedCustomColorWindows` so `DWMWA_COLOR_DEFAULT` is only sent to windows that were previously colored by the mod. Pre-existing windows that manage their own caption colors are no longer overwritten when custom colors are off.

2. **Consistent Theme State:**
   Reused the captured `darkMode` variable throughout `ApplyTitleBar` to prevent mid-function state mismatch.

3. **UWP / Shell Host Verification:**
   Confirmed that the smart class detection in `IsWindowEligible` rejects all UWP/WinUI top-level and bridge surfaces (`ApplicationFrameWindow`, `Windows.UI.Core.CoreWindow`, `DesktopWindowXamlSource`, `Microsoft.UI.Content.DesktopChildSiteBridge`, `WinUIDesktopWin32WindowClass`) before querying or modifying DWM attributes, resolving the conflict from #4582.

4. **MDI Support & Settings Defaults:**
   Hooked `DefFrameProcW/A` for MMC snap-ins, switched fallback to ordinal 132 (`ShouldAppsUseDarkMode`), enabled custom colors by default, and updated dark mode default colors to `#000000` (active) and `#202020` (inactive).

/ai-review

Louis047:
Summary of updates addressing the latest review:

1. **UWP / Shell Host Exclusions:**
   Added `@exclude` directives for core modern Windows shell and UWP container hosts (`systemsettings.exe`, `applicationframehost.exe`, `startmenuexperiencehost.exe`, `searchhost.exe`, `shellexperiencehost.exe`) alongside runtime dynamic class filtering.

2. **Default Settings & README Documentation:**
   Custom colors are enabled by default in v1.2.0 (`#FFFFFF`/`#E6E6E6` for light mode, `#000000`/`#202020` for dark mode). This is documented in the README `## Notes` section and PR changelog.

3. **Targeted DWM Caption Color Resets:**
   `DWMWA_COLOR_DEFAULT` resets are tracked via `g_appliedCustomColorWindows` so they only reset windows that were previously colored by the mod, preserving colors on pre-existing windows when custom colors are toggled off.

4. **MDI Frame Window Support:**
   Hooked `DefFrameProcW/A` to support MMC snap-ins (Device Manager, Event Viewer, etc.).

/ai-review

Louis047:
Tested the build against modern shell surfaces and WinUI/UWP host applications:
- Verified that with the `@exclude` directives and `IsWindowEligible` runtime class detection (`ApplicationFrameWindow`, `Windows.UI.Core.CoreWindow`, `WinUIDesktopWin32WindowClass`, XAML bridges), modern shell hosts and WinUI surfaces run cleanly with no crashes or conflicts.
- Verified titlebar switching and activation on classic Win32 apps, dialogs, and MMC snap-ins (`DefFrameProc`).

/ready-for-reviewer

Louis047:
@m417z Hope this does it now, took way too long than expected

PR #4800, opened 2026-07-15
Add Win32 UI Modernizer

crazyboyybs:
## Mod authorship

This mod was created by:

- - [X] Claude
- - [X] ChatGPT
- - [ ] Gemini


crazyboyybs:
@m417z Hey, I see the PR is requiring manual inspection for the GWLP_WNDPROC

m417z:
The check is there because in the most cases, `WindhawkUtils::SetWindowSubclassFromAnyThread` is preferable. `GWLP_WNDPROC` will cause trouble if multiple mods use it.

crazyboyybs:
> The check is there because in the most cases, `WindhawkUtils::SetWindowSubclassFromAnyThread` is preferable. `GWLP_WNDPROC` will cause trouble if multiple mods use it.

I will change it then, thank you

crazyboyybs:
@m417z A few sites are still on raw GWLP_WNDPROC, for reasons specific to what each one does.

Read-only checks,  not a subclass, nothing to convert
Two call sites only read the current WNDPROC via GetWindowLongPtrW to identify which module owns a CabinetWClass window (Explorer's atlthunk.dll vs. Control Panel/Tools' comctl32.dll), so the mod can branch its dark-mode handling accordingly. Nothing gets written, it's a one-time module-identity check, unrelated to subclassing.

AutoPlay replacement window, one intentional fallback
This mod replaces the AutoPlay flyout with its own Direct2D popup (a private window class the mod creates and owns entirely, no other mod targets it). Install and the normal teardown are now SetWindowSubclassFromAnyThread / self-removal on WM_DESTROY, same as everywhere else.

The one remaining raw swap is a last-resort fallback in teardown. Normal teardown posts a message asking the window to remove its own subclass on its own thread (SendMessageTimeoutW, 1s timeout). If that succeeds, we're done through the subclass API. But if the window's thread is hung and the message never gets processed in time, the code falls back to a direct SetWindowLongPtrW(hwnd, GWLP_WNDPROC, DefWindowProcW) issued from the calling (unload) thread. That fallback only exists because SendMessageTimeoutW already told us the target thread isn't responding. As far as I can tell, SetWindowSubclassFromAnyThread/RemoveWindowSubclassFromAnyThread also marshal their actual work onto the window's owning thread, so calling them here would hit the same stuck thread and could block the same way. A direct SetWindowLongPtrW from another thread doesn't need the target thread's message pump to be alive, it's a synchronous write to an OS-owned window property.

If SetWindowSubclassFromAnyThread/RemoveWindowSubclassFromAnyThread don't actually depend on the target thread's responsiveness internally, I'd like to know, that would let drop this last raw swap too.

m417z:
> But if the window's thread is hung ...

Why not wait for it indefinitely to become unhung? I'm not sure doing something potentially destructive like `SetWindowLongPtrW(hwnd, GWLP_WNDPROC, ...)` is better. Even if you happen to be the only one to subclass the window (if not, it'll crash at some point), you're replacing the pointer under the `RemoveWindowSubclass` feet, and skip that call yourself (although you called `SetWindowSubclass` in the past) which may cause unintended consequences.

crazyboyybs:
> Why not wait for it indefinitely to become unhung? I'm not sure doing something potentially destructive like `SetWindowLongPtrW(hwnd, GWLP_WNDPROC, ...)` is better. Even if you happen to be the only one to subclass the window (if not, it'll crash at some point), you're replacing the pointer under the `RemoveWindowSubclass` feet, and skip that call yourself (although you called `SetWindowSubclass` in the past) which may cause unintended consequences.

@m417z Good point, thanks. Removed the raw GWLP_WNDPROC fallback entirely. It now retries the message-based quarantine (through RemoveWindowSubclassFromAnyThread) a few times instead of falling back to a direct overwrite after one attempt. If it still can't get through, it just hides/closes the window (both safe cross-thread, don't need the target thread's cooperation) and permanently pins the module, never touches the WNDPROC again. No more risk of corrupting comctl32's subclass chain.

m417z:
> permanently pins the module

I'd avoid that too. Windhawk expects the module to be unloaded after calling `Wh_ModUninit` and freeing the mod. Pinning the mod dll violates that contract and is not supported. One example where it can be unsafe - if Windhawk is unloaded, the mod will crash or misbehave on any Windhawk API call since the engine is unloaded.

I think that just waiting for the thread/window to become responsive is the simplest and most robust way to handle this. The user will also get an indication that the mod is stuck during unload, allowing them to understand that something is wrong and report it as an issue and/or handle it by exiting/killing the program.

crazyboyybs:
> > permanently pins the module
> 
> I'd avoid that too. Windhawk expects the module to be unloaded after calling `Wh_ModUninit` and freeing the mod. Pinning the mod dll violates that contract and is not supported. One example where it can be unsafe - if Windhawk is unloaded, the mod will crash or misbehave on any Windhawk API call since the engine is unloaded.
> 
> I think that just waiting for the thread/window to become responsive is the simplest and most robust way to handle this. The user will also get an indication that the mod is stuck during unload, allowing them to understand that something is wrong and report it as an issue and/or handle it by exiting/killing the program.

@m417z Makes sense, thanks, removed the pin (and the leaked LoadLibrary reference that was doing the same thing without the flag) entirely. AutoPlay's quarantine now just waits without a bound instead.

While I was at it, found the exact same pin-as-fallback pattern in one unrelated spot: a FlsFree cleanup on mod unload, guarding against a fiber-local-storage callback outliving the DLL if FlsFree ever fails. Removed the pin there too, but kept it bounded (3 retries then log) rather than unbounded, unlike the AutoPlay case, FlsFree failing on a valid index isn't a "thread's temporarily busy" condition, so waiting longer wouldn't have a real chance of helping.

crazyboyybs:
Sorry for the double commit, removed some excludes that weren't supposed to be there

m417z:
### Submission review

Some other notes
* `IsCurrentProcessDwm`, `IsCurrentProcessMMC` - why not just use `@exclude`?
* `FPUGuard` - perhaps a better solution is to use the `-ffp-exception-behavior=maytrap` command line option. It will be added in the next Windhawk version. Can you check if it works? If it does, let's use it, currently there are ~40 `FPUGuard` usages, and it's easy to miss and difficult to maintain.
* `MSG_WND_CLS` - could use a more unique name.

These are some of the things I noticed from a quick overview. Of course, I didn't read all the source code :) a very impressive mod, well done!

---

**1. `SetPreferredAppMode` is not reset on unload — incomplete revert.** `g_setAppMode(0)` is called on the settings-change path (line 33833), the exclusion-reset path (7546), and the `WM_SETTINGCHANGE` handler (8929), but **not** in `Wh_ModUninit` (33492–33720). So if the user disables the mod while custom dark mode is active, `RestoreDarkSysColors()` reverts the color table (33605) but the process keeps its uxtheme dark-mode preference at ForceDark/AllowDark until it restarts — newly created standard controls (scrollbars, common controls) in a long-lived host like Explorer keep rendering dark after the mod is gone. Reversibility is a core Windhawk principle, so mirror the settings-change path in `Wh_ModUninit`:
```cpp
if (g_setAppMode) g_setAppMode(0);
```
(guard it the same way you do at 33833 if you don't want to clobber another dark-mode mod's preference when it wasn't you who set it).

**2. `MSG_WND_CLS` window-class lifecycle — harden the unchecked `RegisterClassExW`.** `CreateMsgWindow` registers `WH_MW_MsgWnd` (line 8974) with a **mod-code** `lpfnWndProc` (`ThemeMsgWndProc`) and does not check the return value. On a normal unload it's cleaned up fine (`DestroyMsgWindow` → `UnregisterClassW` at 8998). But `Wh_ModUninit` usually runs on a different thread than `Wh_ModInit`, so `DestroyMsgWindow` takes the cross-thread branch: `SendMessageTimeoutW(hwnd, WM_CLOSE, …, SMTO_ABORTIFHUNG, 1000, …)` (8992). If the owning thread's pump is hung/gone, that times out, the window survives, `UnregisterClassW` fails, and on the **next** load `RegisterClassExW` silently fails with `ERROR_CLASS_ALREADY_EXISTS` (return ignored) — leaving `g_msgWnd` bound to a class whose `lpfnWndProc` points into the now-unloaded mod image, so the next `WM_THEMECHANGED`/`WM_SETTINGCHANGE` broadcast calls freed code. You already solved exactly this for the AutoPlay window: register the class with `DefWindowProcW` as the class proc and add behavior per-window via `SetWindowSubclassFromAnyThread` (see `ApRegisterWndClass`, lines 19087–19108 of this file). Doing the same for the message window — or at minimum checking the `RegisterClassExW` result and unregister+re-registering when a stale registration is found — removes the dangling-`WndProc` path entirely.

**3. `ExplorerSection` has no English `$name`.** Lines 643–644 provide `$name:pt` / `$name:es` but there's no base `$name: -- Explorer --`, so English/default-locale users see the raw key `ExplorerSection` as the group header (every other section has the English base name). Add the English `$name`.

<p>

- **A couple of hooks synthesize return values that don't match the API contract.** In `SendMessageW_hook`, when `TVM_SETIMAGELIST(TVSIL_STATE)` is swallowed for the custom pin list (lines 5402, 5433) it `return 0` — i.e. "no previous imagelist" — so a caller that relied on the returned `HIMAGELIST` to free the old one would leak it (low practical risk since Explorer's nav-pane imagelists are system-shared). Similarly `DrawTextExW_hook`'s winver-button suppression path returns `TRUE`/`1` (20277) instead of the real text height. Both are harmless for the callers involved but technically incorrect; returning the value from `_orig` (or the true height) would be cleaner.

</p>
</details>

crazyboyybs:
> ### Submission review

@m417z Thanks for the thorough review! Addressed everything:

- SetPreferredAppMode now reverts in Wh_ModUninit too (guarded on whether the mod is the one who set it).
- MSG_WND_CLS hardened the same way as ApRegisterWndClass, class registered with DefWindowProcW, behavior attached via SetWindowSubclassFromAnyThread, no more dangling-callback path on a timed-out teardown. Also gave it a more unique name.
- ExplorerSection now has its English $name.
- IsCurrentProcessDwm/IsCurrentProcessMMC moved to exclude, good catch, no downside since they already bailed out before any hooks installed.
- Both return-value spots fixed: the pin-list swallow now forwards the real previous HIMAGELIST, and the winver suppression path returns the real text height via DT_CALCRECT instead of a bare TRUE.

On the FPU guard: looked into it, the FPUGuard sites all wrap D2D resource creation specifically (factory/render target creation), guarding against D2D's own internal side effect of changing the process-wide FPU control word. -ffp-exception-behavior=maytrap affects how the compiler codegens our own floating-point expressions, not whether D2D's internal code changes the control word or whether that change gets reverted afterward, so it doesn't look like a drop-in replacement for what FPUGuard is doing here. Keeping it for now but happy to revisit if a live test says otherwise in the future.

PR #4805, opened 2026-07-16
Add Win+Number Virtual Desktop Switch

Cospamos:
Remaps Win+1..9 from the default "activate Nth taskbar app" behavior to Linux-style virtual desktop switching:

- Win+N switches to virtual desktop N, creating it (and any desktops before it) automatically if it doesn't exist yet.
- Win+Shift+N moves the focused window to desktop N (optionally switching the view there too).
- A desktop is removed automatically once you switch away from it and it has no windows left on it.
- Prevents the taskbar flash that can otherwise happen after switching desktops.
- Optionally hides the "Desktop N" name overlay shown on builds where that experimental animation is enabled.

## Changelog

New mod, no changelog yet.

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/virtual-desktop-win-number-switch.wh.cpp
+++ b/mods/virtual-desktop-win-number-switch.wh.cpp
@@ -1,0 +1,1369 @@
+// ==WindhawkMod==
+// @id              virtual-desktop-win-number-switch
+// @name            Win+Number Virtual Desktop Switch
+// @description     Remaps Win+1..9 from "activate Nth taskbar app" to Linux-style virtual desktop switching, with auto-create/auto-remove of desktops
+// @version         2.9
+// @author          Cospamos
+// @github          https://github.com/Cospamos
+// @include         windhawk.exe
+// @compilerOptions -lole32 -loleaut32 -luuid
+// @license         MIT
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Win+Number Virtual Desktop Switch
+
+Replaces the default Windows Win+1..9 behavior (activate the Nth app pinned/open
+on the taskbar) with Linux-style virtual desktop switching:
+
+- **Win+N** switches to virtual desktop N.
+- **Win+Shift+N** moves the currently focused window to desktop N, and
+  switches your view there too (can be turned off in settings to stay put).
+- If desktop N does not exist yet, it (and any desktops before it) is **created
+  automatically**.
+- When you switch away from a desktop that has **no windows left on it**
+  (windows pinned to all desktops don't count), that desktop is **removed
+  automatically**.
+- No taskbar flash asking you to switch back: before switching, the mod hands
+  focus to Explorer's own taskbar window so its internal focus-restore isn't
+  denied and doesn't fall back to flashing a taskbar button, then takes real
+  focus for the actual target window itself right after.
+- No "Desktop N" name overlay popping up on builds where that experimental
+  animation is enabled.
+
+## How it works
+
+This mod runs as a dedicated background process (a Windhawk "tool mod") and
+installs a low-level keyboard hook that intercepts Win+1..9 and Win+Shift+1..9
+system-wide before Explorer's own taskbar shortcut handling sees them, then
+talks to the undocumented Virtual Desktop COM interfaces to
+switch/create/remove desktops and move windows between them.
+
+## Important notes
+
+- The "Desktop N" name overlay is a `XamlExplorerHostIslandWindow` hosted by
+  explorer.exe, and its visibility is toggled through DirectComposition/DWM
+  rather than `CreateWindowExW`/`ShowWindow`/`SetWindowPos` (confirmed by
+  injecting into explorer.exe and hooking those calls directly - they never
+  fired for it), so this mod can't veto it before it's ever painted. Instead
+  it reacts to it via `SetWinEventHook` and hides it immediately, plus
+  proactively re-hides the same window (it's reused, not recreated) right
+  before the next switch starts. This can still show a very brief flicker in
+  rare cases since it's reactive rather than fully preventive. Turn off
+  "Hide the 'current desktop' name overlay" in settings if you'd rather
+  Windows' default behavior.
+- The Virtual Desktop COM interface IDs (GUIDs) and method layout change
+  between Windows builds. This mod auto-detects your build from the registry
+  and picks the matching interface set. If desktop switching silently does
+  nothing, check the mod's debug log (Windhawk -> mod -> Logs) for the
+  detected build number, and try overriding "Windows Version" manually in
+  settings.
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

Let me know what you think about 1.

---

**1. Substantial overlap with `virtual-desktop-helper`.** [`virtual-desktop-helper`](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/virtual-desktop-helper.wh.cpp) is an existing tool mod (also `@include windhawk.exe`) whose core feature set is "switch to desktop 1‑9" and "move active window to desktop 1‑9" via a configurable modifier+number — the same core this mod provides, and it even uses the identical Windows-version COM interface buckets (same `$options` list). It's preferred *extending* an existing mod over merging a near-duplicate, to keep the catalog from fragmenting.

Your mod does add genuinely new capabilities that `virtual-desktop-helper` lacks:
- reliable **Win+N interception** (it swallows the taskbar's built-in Win+number shortcut via a low-level hook — the helper punts on this, its README literally says "you may need to disable these Win-combinations by yourself"),
- **auto-create** missing desktops and **auto-remove** empty ones,
- **taskbar-flash suppression** and the **"Desktop N" overlay hiding**.

So this isn't a pure duplicate. But before it's merged as a standalone mod, please make the case for why these can't be contributed to `virtual-desktop-helper` as options (the Win+N interception in particular is the strongest justification for a separate mod). Coordinate with the maintainer on which way to go — that's a decision for them, but the overlap needs to be called out explicitly.

**2. The keyboard hook and the desktop-switch work share one thread, so every switch stalls system-wide keyboard input.** The `WH_KEYBOARD_LL` hook is installed on `HotkeyThreadProc` (line 1058), and that *same* thread's message loop is what runs `GoToDesktopNum` / `MoveActiveWindowToDesktopNum` (lines 1078/1085). A low-level keyboard hook callback can only be delivered while its owning thread is servicing its message queue — but during a switch that thread is busy inside `GoToDesktopNum`, which includes cross-process COM calls, an `EnumWindows` sweep that makes a COM call per window (`IsDesktopEmpty`), and, worst of all, the `Sleep(50)` × 3 = 150 ms retry loop in `ForceForegroundWindow` (lines 774‑780). While the thread is blocked there, the system can't call the hook, so **all keyboard input everywhere is delayed** (and past `LowLevelHooksTimeout` the hook is bypassed entirely, or Windows may unhook it). The "post the work off the hook" design is the right instinct, but because the hook and the worker are the *same* thread, the work still blocks the hook.

Fix: keep the hook on a dedicated thread whose message loop does only trivial dispatch (`PostThreadMessage`), and run `GoToDesktopNum` / the COM work on a *separate* worker thread. At minimum, get the `Sleep`-based flash retry off the hook thread — a 150 ms synchronous sleep on the thread that gates all keyboard input is the biggest single offender. See the wiki note on not doing blocking work in low-level hooks and the general "post work to a worker thread with its own message loop" guidance.

**3. Remove the `WinReleaseDelayMs` "(TEMP)" setting before merge.** Its `$name` is literally `(TEMP) Win key un-stick delay (ms)` and its `$description` says "Tuning knob, will be removed." (lines 115‑118). Shipping a setting you've already flagged for deletion just confuses users. Either finalize it as a real, plainly-named setting (if the value genuinely needs to be user-tunable) or hard-code the chosen value and drop the setting.

<p>

- **Use `WindhawkUtils::StringSetting`** (RAII) instead of the raw `Wh_GetStringSetting` + `Wh_FreeStringSetting` pair in `LoadSettings`.
- **Drop `-loleaut32`** from `@compilerOptions` if nothing pulls it in — I don't see any oleaut32 API used (`-lole32`/`-luuid` are needed).
- **Stray dev-artifact comments.** Several comments reference context that doesn't exist in the shipped mod — "(see conversation)", "(see git history)". Worth cleaning up for a public submission.
- **Cross-thread flags** (`g_stopHotkeyThread`, `g_pendingWinReleaseTick`, etc.) are `volatile`; `std::atomic<T>` is the idiomatic choice for these.

</p>
</details>

<p>

- **The overlay-hiding heuristic keys off a shared window class.** `XamlExplorerHostIslandWindow` is reused by other shell flyouts (volume, network, etc.), and you hide any instance that appears within the 2 s post-switch window. If the user pops a volume/network flyout right after a Win+N switch, it can get hidden too. Likewise `g_knownDesktopLabelHwnd` caches the last such window regardless of what it actually was, and re-hides it proactively on the next switch. It's a best-effort heuristic — just be aware of the false-positive surface.
- **Global Ctrl injection on each Win+N.** `SendMarkedSyntheticKeystroke(VK_CONTROL)` injects a real Ctrl down/up into whatever app currently has focus. It's there to cancel the Start-menu solo-tap, but a stray Ctrl tap can have side effects in the focused app (games, editors). The whole "swallow the Win key-up then schedule a synthetic release later" scheme is inherently timing-fragile (which is presumably why the TEMP delay knob exists) — it works, but it's the part most likely to misbehave across machines.

</p>
</details>

Cospamos:
Re: 1 (overlap with virtual-desktop-helper)

Just setting the modifier to Win in virtual-desktop-helper's settings doesn't actually solve this - the mod's own README says so directly:

> Win key note: Win+number keys conflict with Windows taskbar shortcuts. If it doesn't work for you, you may need to disable these Win-combinations by yourself.

That's exactly the problem this mod removes. On Windows, disabling the native Win+number taskbar shortcuts isn't a toggle in settings - it's usually a group policy edit or a third-party tool, which is a real barrier for the average user this mod targets. The whole point of the low-level hook here is to make Win+N work without that step.

I went through the source to check whether this could realistically be contributed as an option instead. Its hotkey handling is built entirely on RegisterHotKey/WM_HOTKEY, a fundamentally higher-level mechanism than a low-level keyboard hook. None of what the latter needs exists in that code: no manual modifier-state tracking (RegisterHotKey handles that for free), no Start-Menu-on-solo-tap suppression, and no dedicated hook-only thread.

Happy to reach out to the maintainer directly to discuss it, but I don't think it would go anywhere in practice: supporting Win+N there wouldn't be a setting, it would mean adding a second, parallel hotkey subsystem - close to rewriting the relevant parts of that mod from scratch.

Separately: items 2 and 3 from this review are addressed - the keyboard hook now runs on its own dedicated thread, with the desktop-switch/COM work moved to a separate worker thread, and the WinReleaseDelayMs setting has been removed (hardcoded, since the delay is now adaptive based on how long Win was actually held).

m417z:
> On Windows, disabling the native Win+number taskbar shortcuts isn't a toggle in settings

Right. Have you tried the [Disable Windows Shortcuts](https://windhawk.net/mods/disable-windows-shortcuts) mod?

Cospamos:
Yeah, okay. You're right on this one. Thanks for the honest answer.

PR #4806, opened 2026-07-16
more aggressively hide the navbar

Anixx:
## Changelog

* Changelog item 1...
* Changelog item 2...

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/flexible-explorer-toolbars-deluxe.wh.cpp
+++ b/mods/flexible-explorer-toolbars-deluxe.wh.cpp
@@ -81,7 +81,7 @@ constexpr int UP_BUTTON_ICON_SIZE  = 16;
 constexpr UINT LOCK_TOOLBARS_CMD_ID = 41484;
 
 UINT g_msgDoMove = 0;
-UINT g_msgSuppressNavbar = 0;
+UINT g_msgRefreshLayout = 0;
 
 struct Settings {
     bool moveSearchBand    = true;
@@ -446,11 +446,11 @@ void ExpandShellTabToFillCabinet(HWND cab){
     SetWindowPos(s,NULL,0,0,rc.right,rc.bottom,SWP_NOZORDER|SWP_NOACTIVATE);
 }
 
-// ---- Permanent, non-destructive suppression of the navigation bar container ----
-
 void ForceHideNavWorker(HWND w){
     if(!w||!IsWindow(w))return;
+    bool alreadyHidden = IsForceHidden(w) && !(GetWindowLongPtr(w,GWL_STYLE) & WS_VISIBLE);
     MarkForceHidden(w);
+    if(alreadyHidden)return;
     LONG_PTR style=GetWindowLongPtr(w,GWL_STYLE);
     if(style & WS_VISIBLE)
         SetWindowLongPtr(w,GWL_STYLE,style & ~WS_VISIBLE);
@@ -520,7 +520,7 @@ void ShowToolbarContextMenu(HWND rebar, int x, int y) {
     HWND cab = rebar;
     while (cab) {
         WCHAR cls[64];
-        if (GetClassName(cab, cls, ARRAYSIZE(cls)) && wcscmp(cls, L"CabinetWClass") == 0)
+        if (GetClassName(cab, cls, ARRAYSIZE(cls)) && !wcscmp(cls, L"CabinetWClass"))
             break;
         cab = GetParent(cab);
     }
@@ -529,14 +529,14 @@ void ShowToolbarContextMenu(HWND rebar, int x, int y) {
     HWND workerW = FindMenuBarWorkerW(cab);
     if (!workerW) return;
 
-    HMODULE hMod = LoadLibraryW(L"explorerframe.dll");
+    HMODULE hMod = GetModuleHandleW(L"explorerframe.dll");
     if (!hMod) return;
 
     HMENU hMenu = LoadMenuW(hMod, MAKEINTRESOURCEW(264));
-    if (!hMenu) { FreeLibrary(hMod); return; }
+    if (!hMenu) return;
 
     HMENU hSubMenu = GetSubMenu(hMenu, 0);
-    if (!hSubMenu) { DestroyMenu(hMenu); FreeLibrary(hMod); return; }
+    if (!hSubMenu) { DestroyMenu(hMenu); return; }
 
     bool locked = AreToolbarsLockedByGripper(rebar);
     int itemCount = GetMenuItemCount(hSubMenu);
@@ -563,7 +563,6 @@ void ShowToolbarContextMenu(HWND rebar, int x, int y) {
     PostMessage(workerW, WM_NULL, 0, 0);
 
     DestroyMenu(hMenu);
-    FreeLibrary(hMod);
 }
 
 LRESULT CALLBACK UpButton_SubclassProc(HWND hwnd, UINT msg, WPARAM wP, LPARAM lP, DWORD_PTR) {
@@ -615,6 +614,12 @@ LRESULT CALLBACK BreadcrumbToolbar_SubclassProc(HWND hwnd,UINT msg,WPARAM wP,LPA
 }
 
 LRESULT CALLBACK AddressBandRoot_SubclassProc(HWND hwnd,UINT msg,WPARAM wP,LPARAM lP,DWORD_PTR){
+    if(msg==WM_NCDESTROY){
+        // Self-clean the neutered set on destroy so entries don't leak across
+        // cabinet closes (mirrors the force-hidden worker cleanup below).
+        UnmarkNeutered(hwnd);
+        return DefSubclassProc(hwnd,msg,wP,lP);
+    }
     if(IsNeutered(hwnd))return DefWindowProc(hwnd,msg,wP,lP);
     return DefSubclassProc(hwnd,msg,wP,lP);
 }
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

---

**`NavWorkerW_SubclassProc` now bypasses the original window procedure (regression from v1.0).** In the `IsForceHidden` branch, every message now falls through to `DefWindowProc` instead of `DefSubclassProc` (v1.0 chained to `DefSubclassProc` for the pass-through cases):

```cpp
case WM_NCDESTROY:
    UnmarkForceHidden(hwnd);
    return DefWindowProc(hwnd,msg,wP,lP);   // <-- was DefSubclassProc
...
default:
    return DefWindowProc(hwnd,msg,wP,lP);    // <-- was DefSubclassProc
```

Since a moved cabinet keeps this worker force-hidden for its whole lifetime, this permanently cuts Explorer's real `WorkerW` procedure (and any other subclasser in the chain) out of the loop. The clearest concrete problem is `WM_NCDESTROY`: routing it to `DefWindowProc` means Explorer's own destroy handler never runs and the subclass chain isn't walked, so per-window state the shell frees on destroy is leaked on every cabinet close. Please chain the pass-through/destroy cases through `DefSubclassProc` — you can keep munging the `WINDOWPOS`/`STYLESTRUCT` and swallowing `WM_SIZE`/`WM_MOVE`/`WM_SHOWWINDOW` with `return 0` exactly as now; only the `DefWindowProc` fall-throughs (especially `WM_NCDESTROY`, `WM_WINDOWPOSCHANGING`, `WM_STYLECHANGING`, `default`) should become `DefSubclassProc`, which is what kept the window hidden-but-alive in v1.0.

<p>

- **`g_msgSuppressNavbar` is dead code.** It's registered in `Wh_ModInit` (with its own `return FALSE` failure path) and handled in `Cabinet_SubclassProc`, but it is never posted or sent anywhere, so the handler never fires. `SuppressStrayNavWorkers()` is already called directly from the `WM_SIZE`/`WM_WINDOWPOSCHANGED` path, `DoMoveSearchBandToMenuBar`, etc., so the message route is redundant — remove the global, the registration (and its `return FALSE`), and the handler branch.

- **`WM_APP + 105` on a window you don't own.** You post/handle `WM_APP + 105` on Explorer's `ReBarWindow32`. `WM_APP` values are safe from the control itself, but another mod subclassing the same rebar (e.g. a command-bar / toolbar restorer that your README recommends running alongside this one) could pick the same value and collide. You already use `RegisterWindowMessage` for `g_msgDoMove`; registering this one the same way makes it process-unique and consistent.

- **Stale map entries after the `CleanupCabinetState` trim.** With the WorkerW/neutered-cleanup block removed, `UnmarkNeutered` now has no callers, so `g_neuteredAddressRoots` entries for closed cabinets are never released (small unbounded growth over a long Explorer session; also a theoretical false-positive if an `HWND` value is recycled into a future `Address Band Root`). The force-hidden map is fine now that `WM_NCDESTROY` self-cleans it; consider giving the neutered set the same treatment (or restoring a lightweight cleanup on close).

- **`SuppressStrayNavWorkers` re-hammers already-hidden workers.** It runs on every cabinet `WM_SIZE`/`WM_WINDOWPOSCHANGED`, and `ForceHideNavWorker` unconditionally calls `ShowWindow(SW_HIDE)` + `SetWindowPos(..., SWP_FRAMECHANGED)` (which forces a `WM_NCCALCSIZE`) even when the worker is already hidden. A quick early-out (skip if already `IsForceHidden` and not `WS_VISIBLE`) avoids the redundant per-resize work.

- **`LoadLibraryW(L"explorerframe.dll")` in `ShowToolbarContextMenu`** (pre-existing) — `explorerframe.dll` is always already loaded in `explorer.exe`, so `GetModuleHandleW(L"explorerframe.dll")` gets the same handle without the extra `LoadLibrary`/`FreeLibrary` round-trip (and sidesteps the bare-name search-path concern entirely).

</p>
</details>

<p>

- **The `WM_APP + 105` refresh is fairly heavy.** Every `RB_SETBANDINFO` during folder navigation posts a `WM_APP + 105`, whose handler synthesizes a full `WM_SIZE` to force a re-layout. During navigation Explorer fires many `RB_SETBANDINFO`s, so this can generate a lot of layout churn (each synthetic `WM_SIZE` also re-enters the pending-apply retry logic). It works, but if you see flicker/lag on rapid navigation, coalescing the refreshes (e.g. a short debounce, or `RB_SIZETORECT` instead of a synthetic `WM_SIZE`) would be lighter.

- **Reversibility limitation.** Disabling the mod removes the subclasses but doesn't restore anything — the moved bands stay in the menu rebar and the navbar worker stays hidden (the new `SetWindowLongPtr(GWL_STYLE, … & ~WS_VISIBLE)` makes that stickier than v1.0), so an Explorer restart is needed to get the original layout back. That's inherent to the mod's approach, so there may be no clean fix, but it's worth noting for users in the README if it isn't already implied.

</p>
</details>

m417z:
### Submission review

---

**1. The new "layout refresh" message is registered and posted but never handled — the fix is a no-op.** `g_msgRefreshLayout` is registered in `Wh_ModInit` and posted from `ReBar_SubclassProc` on `RB_SETBANDINFO`:

```cpp
// Explorer triggers RB_SETBANDINFO during folder navigation.
// It fragments the layout, leaving gaps. Post a scheduled layout refresh to fix it.
if (!g_insideApply && !g_insideGripperSync) {
    PostMessage(hwnd, g_msgRefreshLayout, 0, 0);
}
```

But nothing ever handles `g_msgRefreshLayout` — there's no `if(msg==g_msgRefreshLayout)` branch in `ReBar_SubclassProc` (or anywhere else). When the posted message is dispatched it falls through to `DefSubclassProc`/`DefWindowProc` and does nothing, so the intended layout-gap correction during folder navigation never runs; the code just posts a message to the queue on every `RB_SETBANDINFO` for no effect. Either wire up the handler to actually re-apply the layout, e.g.:

```cpp
LRESULT CALLBACK ReBar_SubclassProc(HWND hwnd,UINT msg,WPARAM wP,LPARAM lP,DWORD_PTR){
    if(msg==g_msgRefreshLayout){
        if(IsPendingApply(hwnd)) ReapplyCx(hwnd);   // or ApplySavedLayout(hwnd);
        return 0;
    }
    ...
```

…or remove the registration + post if the refresh isn't needed. As written it's dead code. (If you do wire it up, guard against the obvious feedback loop, since `ReapplyCx`/`ApplySavedLayout` themselves issue `RB_SETBANDINFO` — the existing `g_insideApply` flag should cover that, but worth confirming.)

**2. The `WM_NCDESTROY` "self-clean" for the neutered set is attached to the wrong window class, so `g_neuteredAddressRoots` still leaks.** You removed the explicit neutered/force-hidden cleanup from `CleanupCabinetState` and replaced it with self-cleaning `WM_NCDESTROY` handlers. That works for the force-hidden workers (they *are* hooked with `NavWorkerW_SubclassProc` and *are* force-hidden), but not for the neutered set:

- The only entries ever added to `g_neuteredAddressRoots` come from `MarkNeutered(it->leftover)` in `DoMoveSearchBandToMenuBar`, where `leftover` is a **`Breadcrumb Parent`** or **`UpBand`** window.
- The new cleanup lives in `AddressBandRoot_SubclassProc`, which is only ever installed on **`Address Band Root`** windows (`ProcessWindow`), and those are never neutered.

So the `WM_NCDESTROY` → `UnmarkNeutered(hwnd)` you added runs on windows that were never in the set, while the windows that *are* in the set (`Breadcrumb Parent`/`UpBand`) are never hooked and so never trigger it. The comment ("Self-clean the neutered set on destroy so entries don't leak") doesn't hold — the set grows by up to two stale `HWND`s per cabinet open→close and is never pruned. Since `HWND` values get recycled by the OS and `IsNeutered()` is consulted for every `Address Band Root`, a reused handle can later cause a *live* address bar to be wrongly neutered (it returns `DefWindowProc` and stops working). Restore an explicit `UnmarkNeutered` for the leftovers on cabinet teardown (as the old `CleanupCabinetState` did), or hook the leftover windows with a proc that self-cleans on `WM_NCDESTROY`.

(As an aside, because the neutered leftovers are never hooked, `IsNeutered` is never actually consulted for them either — so it's worth double-checking whether the `MarkNeutered`/`ShowWindow(SW_HIDE)` on the leftovers is even doing what you expect, or whether the neuter mechanism is partly vestigial here.)

<p>

- **Dead function.** `IsNavbarWorkerW` (line 415) is no longer referenced now that `ProcessWindow` uses an inline parent-class check. It can be removed.
- **Stray whitespace.** The new leading blank line at the top of `ReBar_SubclassProc` and the trailing-whitespace lines around the `RB_SETBANDINFO` block are cosmetic; clean them up to keep the diff tidy.

</p>
</details>

<p>

- **Broader WorkerW hiding.** `ProcessWindow` now force-hides *every* `WorkerW` whose direct parent is `CabinetWClass`, dropping the old `IsNavbarWorkerW` requirement that the worker actually contain a `ReBarWindow32`. This hides the nav worker earlier/more reliably (good), but if any build ever puts a second, non-navbar `WorkerW` directly under the cabinet it will be hidden too. If you've confirmed the navbar worker is the only direct-child `WorkerW` on your supported builds, this is fine — just flagging the widened net.
- **Reversibility deepened.** `ForceHideNavWorker` now strips `WS_VISIBLE` via `SetWindowLongPtr(GWL_STYLE, …)` in addition to the `SetWindowPos` hide. `Wh_ModUninit` only removes subclasses — it doesn't restore the style bit, re-show the workers, or move the bands back — so disabling the mod leaves the current Explorer window's navbar hidden and bands relocated until that window is recreated. This is largely inherent to how the mod works (and was already the case in 1.0), but stripping the style bit makes the un-restored state a bit deeper; noting it in case you want a teardown that re-shows the workers.
- **Removed `cxMinChild = 80`.** Per the commit message this is intentional, but note that inserted bands can now collapse well below the previous 80px floor when a saved/ideal width is small; make sure that reads acceptably for the search/breadcrumb bands.

</p>
</details>

Anixx:
`cxMinChild = 80` is the line that Claude inserted without any reason and prompt and that made me spend a day to debug why people could not resize the up button below 80px width.

PR #4810, opened 2026-07-16
Update Island Media Controls to 0.9.211

usho-lear:
## Changelog

- Update Island Media Controls to 0.9.211.
- Add a compact expanded layout that places the album cover beside the title and artist, reducing pointer travel to the playback controls.
- Add taskbar-position awareness, including support for Taskbar on top for Windows 11: the popup direction, component layout, and album wash adapt to the taskbar position.
- Refine expand and collapse animations so the compact island, artwork, progress bar, playback controls, text, corner radius, and live backdrop morph more cohesively.
- Restore display-refresh-rate-synchronized live blur without the temporary 60 FPS animation cap or 15 FPS settled-state cap.
- Optimize Acrylic and Liquid Glass with WGC dirty-region rendering, Direct2D effect-graph composition, cached GPU resources, and fewer full-size intermediate render passes.
- Serialize WGC frame rendering, DirectComposition updates, and live-blur render snapshots under the overlay mutex to avoid free-threaded capture races with UI-thread animation and settings updates.
- Improve Acrylic and Liquid Glass interaction stability, reducing flicker, self-sampling, invisible overlay hit-test issues, and repeated-toggle stalls.
- Refine the Liquid Glass edge refraction and strengthen the light-mode backdrop and playback-control tint for clearer surface shape.
- Increase the playback-control album-wash layer opacity so the control card reads as more strongly blurred.
- Improve light-mode visibility by keeping the album-art wash enabled for Liquid Glass.
- Remove obsolete CPU WGC presentation, native-blur handoff, diagnostic, and unused animation helper paths while preserving the recording-mode static blur fallback.
- Update the default settings to the currently tuned setup while keeping browser thumbnails as the default artwork mode.
- Refresh the README and What's new sections to focus on user-facing features.
- Use SetWindowSubclass/RemoveWindowSubclass for safe popup XAML child cleanup.

## Validation

- Windhawk Clang syntax compilation passes with -Wunreachable-code and -Wunreachable-code-aggressive.
- python .github/pr_validation.py ./mods/island-media-controls.wh.cpp usho-lear passes.
- git diff --check -- mods/island-media-controls.wh.cpp passes.

### DIFF
--- a/mods/island-media-controls.wh.cpp
+++ b/mods/island-media-controls.wh.cpp
@@ -2,7 +2,7 @@
 // @id              island-media-controls
 // @name            Island Media Controls
 // @description     Dynamic island-like media controls for the Windows 11 taskbar.
-// @version         0.9.196
+// @version         0.9.207
 // @author          usho
 // @github          https://github.com/usho-lear
 // @license         MIT
@@ -143,12 +143,6 @@ play/pause, and next controls.
     - "solid": "Solid"
     - "acrylic": "Acrylic / glass"
     - "liquid_glass": "Liquid glass"
-  - BackdropHandoffDelayMs: 0
-    $name: Backdrop handoff delay (ms)
-    $description: Wait after the popup reaches its final shape before switching from fallback blur to live WGC blur.
-  - BackdropHandoffFadeMs: 100
-    $name: Backdrop handoff fade (ms)
-    $description: Crossfade duration from the final fallback blur frame to live WGC blur.
   - BackdropInitialFrameSkip: 2
     $name: Backdrop initial frame skip
     $description: Number of WGC frames to skip before replacing the fallback frame.
@@ -185,6 +179,9 @@ play/pause, and next controls.
 #include <d2d1_1helper.h>
 #include <d2d1effects.h>
 #include <dcomp.h>
+#include <winrt/Windows.Foundation.h>
+#include <winrt/Windows.Foundation.Metadata.h>
+#include <winrt/Windows.Security.Authorization.AppCapabilityAccess.h>
 #include <winrt/Windows.Graphics.Capture.h>
 #include <winrt/Windows.Graphics.DirectX.h>
 #include <winrt/Windows.Graphics.DirectX.Direct3D11.h>
@@ -338,8 +335,6 @@ struct Settings {
     double animationSpeed = 1.0;
     bool displacementDebugPreview = false;
     std::wstring material = L"liquid_glass";
-    int backdropHandoffDelayMs = 0;
-    int backdropHandoffFadeMs = 100;
     int backdropInitialFrameSkip = 2;
     int backdropFallbackBlurPasses = 5;
     int backdropFallbackCaptureScale = 2;
@@ -505,7 +500,6 @@ RECT g_popupLiquidGlassPanelRectPx{};
 int g_popupLiquidGlassPanelRadiusPx = 0;
 bool g_popupLiquidGlassPanelRectValid = false;
 bool g_popupOverlayWgcReadbackHadVisibleFrame = false;
-std::chrono::steady_clock::time_point g_popupOverlayWgcReadbackDiagnosticTime{};
 enum class PopupOverlayWgcDiagnosticState {
     NotStarted,
     StartSkipped,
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

---

**1. Corrupted (mojibake) characters — one of them breaks QQ Music detection.** The file was clearly round-tripped through a non-UTF-8 editor at some point, corrupting the only two non-ASCII lines:

- [`island-media-controls.wh.cpp:1523`](https://github.com/ramensoftware/windhawk-mods/blob/2f91a5e7e47de40437ab77819687f7005b5ebf21/mods/island-media-controls.wh.cpp#L1523) — `L"qq音乐"` became `L"qq闊充箰"`:
  ```cpp
  lower.find(L"qq闊充箰") != std::wstring::npos ||   // was L"qq音乐"
  ```
  This is a functional regression: the QQ Music localized-name fallback in `LooksLikeNativeMusicMediaSource` no longer matches, so a session whose only identifying substring is the Chinese app name falls through to the browser heuristics and can be misclassified. The other substrings (`qqmusic`, `tencent.qqmusic`, `tencentmusic`) still work, so the impact is narrow, but it's an unintended change.
- [`island-media-controls.wh.cpp:9974`](https://github.com/ramensoftware/windhawk-mods/blob/2f91a5e7e47de40437ab77819687f7005b5ebf21/mods/island-media-controls.wh.cpp#L9974) — `progress≈0.985` became `progress鈮?.985` in a comment (cosmetic only, but same root cause).

Restore the correct characters and re-save the file as UTF-8. `grep -nP '[^\x00-\x7F]'` shows exactly these two lines, so they're easy to confirm.

<p>

- **`try`/`catch` glued to a statement.** In `RenderPopupOverlayWgcFrame` the success path ends with `g_popupOverlayWgcFallbackPainted = false;    } catch (winrt::hresult_error const& error) {` on one line ([around L8785](https://github.com/ramensoftware/windhawk-mods/blob/2f91a5e7e47de40437ab77819687f7005b5ebf21/mods/island-media-controls.wh.cpp#L8785)). Purely a formatting nit, but worth splitting for readability.
- A few added lines mix space indentation inside the tab-indented body (e.g. the `artFade.IsHitTestVisible(false);` / `details.IsHitTestVisible(false);` blocks). `git diff --check` is clean, so this is cosmetic only.

- **Untracked detached thread can outlive the mod.** [`RequestPopupOverlayWgcBorderlessAccessAsync`](https://github.com/ramensoftware/windhawk-mods/blob/2f91a5e7e47de40437ab77819687f7005b5ebf21/mods/island-media-controls.wh.cpp#L8497) spawns a fire-and-forget `std::thread(...).detach()` that blocks in `RequestAccessAsync(...).get()`, then touches mod globals and calls `Wh_Log`. Nothing joins or waits for it — `StopPopupOverlayWgcBackdrop`'s spin-wait only covers the WGC render callback (`g_popupOverlayWgcRendering`), not this thread. If the mod is unloaded while `.get()` is still pending, the thread resumes and executes code in the now-unmapped DLL image → crash. In `explorer.exe` (full-trust) the access request usually resolves near-instantly, so the race window is small, but it's real and the thread is entirely untracked. Track the thread and join it before returning from `Wh_ModUninit` (it resolves quickly in shell processes), so it can't outlive the DLL. If you're worried about it ever blocking, guard the join with a bounded wait like the media thread already does.

</p>
</details>

<p>

- **`DisplacementDebugPreview` ("Preview liquid glass displacement map").** This exposes a developer diagnostic — rendering the raw RG displacement texture instead of the material — as an end-user setting. It's harmless and you document it as a tuning aid, but consider whether a "show the debug texture" toggle really belongs in the shipped settings UI, or whether it's better left as a compile-time/internal switch. Not a blocker.
- The taskbar-position awareness (`TaskbarPopupOpensDown`, top-taskbar open-down layout, edge-fade direction) and the new compact expanded layout are feature-behavior changes I couldn't fully exercise from static review — worth a sanity pass on a top-aligned taskbar and a mixed-DPI multi-monitor setup to confirm the popup direction and wash orientation line up.

</p>
</details>

m417z:
### Submission review

Check the note below. Also, fix the changelog on the top of the page so that the update can be submitted.

---

**The free-threaded WGC frame callback now touches DirectComposition objects and shared UI state that the taskbar thread mutates concurrently, without a shared lock.** This PR moves presentation onto the capture callback: `RenderPopupOverlayWgcFrame` (invoked from the `CreateFreeThreaded` frame pool's `FrameArrived` handler at `island-media-controls.wh.cpp:9214`) now runs **outside** `g_popupOverlayWgcMutex` (the mutex is only held to check `g_popupOverlayWgcRunning`, then released before the render — `:9223`/`:9248`), and inside it:

- it calls `g_popupOverlayDcompVisual->SetContent(...)` + `g_popupOverlayDcompDevice->Commit()` (`:8919`–`:8927`), while the taskbar thread concurrently calls `g_popupOverlayDcompVisual->SetTransform(...)` + `g_popupOverlayDcompDevice->Commit()` in `UpdatePopupOverlayDcompVisualTransform` (`:8146`/`:8151`), which `UpdatePopupBackdropOverlayWindow` invokes **without** the mutex during the morph (`:10197`, `:10258`);
- it calls `CalculatePopupBackdropOverlayRect` (`:8614`), which reads `g_settings` (including the `std::wstring material`), `g_layout`, and the animation progress — while `ApplyPendingSettings` replaces `g_settings` wholesale (`g_settings = std::move(*settings)`) on the taskbar thread.

DirectComposition objects are not thread-safe, so a `SetContent`/`Commit` from the pool thread racing a `SetTransform`/`Commit` from the taskbar thread is a data race on the visual/device (worst case a device-removed error or a crash in Explorer), and the `g_settings` read can be a torn read of the `wstring` mid-reassignment. Both windows are the expand/collapse animation path, i.e. normal use. The previous version avoided this by doing the readback on the pool thread and marshaling the actual present to the UI thread via `PostMessage`; that handoff is now gone.

Suggested fix: serialize all DirectComposition visual/device access (and the shared `g_settings`/`g_layout` reads) under `g_popupOverlayWgcMutex` — e.g. hold the mutex across the render in the callback, and take it in `UpdatePopupOverlayDcompVisualTransform` — or snapshot the values the callback needs on the taskbar thread and pass them in. Note that if you extend the callback's lock to cover the whole render, keep `StopPopupOverlayWgcBackdrop`'s existing drain-then-lock ordering so teardown doesn't deadlock against a long blur/present.

PR #4815, opened 2026-07-17
Update Windows 11 Start Menu Buttons 2.2

Salyts:
New Features:

- Added nested submenus support - create submenus within submenus up to 3 levels deep for both left-click and right-click menus
- Added right-click menu support - buttons can now have separate menus for right-click actions (up to 3 levels)
- Submenu icons now support image files (.png, .ico, .jpg, .bmp, .webp) and app paths (.exe, .dll) for icon extraction
- Icon paths now support both quoted and unquoted formats (e.g. "C:\Path\icon.png" or C:\Path\icon.png)
- Added multiple commands support - execute multiple commands in sequence by adding multiple Action entries in Windhawk UI
- Preset buttons now ignore user-defined submenus and right-click menus to maintain consistent preset behavior
- **Added [Visual Configurator](https://salyts.github.io/Windows-11-Start-Menu-Buttons/)** — a web-based tool to configure buttons visually with drag & drop, icon browser, submenu builder, export/import and more

Bug Fixes:

- Fixed "Container left margin" and "Container right margin" not working for Account button
- Fixed quote handling in icon paths - now properly strips outer quotes before processing

Notes:

- When upgrading from v2.1, Action fields may appear empty in UI but will continue to work. To edit them or add multiple commands, re-enter the actions manually in Windhawk settings. See mod description for details.
- **Upgrading from v2.1?** Use the **[Configurator's Migrate tool](https://salyts.github.io/Windows-11-Start-Menu-Buttons/)** to convert your v2.1 config to v2.2 format automatically — paste your old config, click Migrate, and copy the result back to Windhawk.

### DIFF
--- a/mods/windows-11-start-menu-buttons.wh.cpp
+++ b/mods/windows-11-start-menu-buttons.wh.cpp
@@ -10,7 +10,7 @@
 // @include             StartMenuExperienceHost.exe
 // @include             explorer.exe
 // @architecture        x86-64
-// @compilerOptions     -lcomctl32 -lole32 -loleaut32 -lruntimeobject -lshlwapi -lshell32 -luuid -luser32 -lwtsapi32 -lpowrprof -lgdi32 -lgdiplus -lshcore -lcrypt32
+// @compilerOptions     -lole32 -loleaut32 -lruntimeobject -lshlwapi -lshell32 -luuid -luser32 -lwtsapi32 -lpowrprof -lgdi32 -lgdiplus -lshcore
 // ==/WindhawkMod==
 
 // ==WindhawkModReadme==
@@ -161,7 +161,7 @@ If you encounter any issues or have a feature suggestion, please open a report o
 // ==WindhawkModSettings==
 /*
 - preset_language: en
-  $name: Preset language    
+  $name: Preset language
   $description: Language for the tooltip names of preset buttons.
   $options:
     - en: English
@@ -262,7 +262,7 @@ If you encounter any issues or have a feature suggestion, please open a report o
                       $name: Item action
                   $name: Level 3 submenu
               $name: Level 2 submenu
-        $name: Left-click submenu 
+        $name: Left-click submenu
         $description: "Each item can have nested submenus up to 3 levels deep total."
       - rightClickSubmenu:
           - - name: ""
@@ -310,7 +310,7 @@ If you encounter any issues or have a feature suggestion, please open a report o
 #include <shellapi.h>
 #include <shlwapi.h>
 #include <shcore.h>
-#include <wincrypt.h>
+#include <unordered_map>
 
 #undef GetCurrentTime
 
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

Regarding 2, I think we discussed it in another review, and eventually you were able to avoid creating temp files.

---

**1. The multi‑command encoding is fragile and mangles single actions that contain `[` or `]`.**

`BuildButtons`/`LoadSubmenuRecursive` read the action array elements, then flatten them back into one string using `[cmd1][cmd2]` delimiters (`rawAction += L"[" + cmd + L"]"`), store that in `ActionItem::action`, and `ExecuteActionText` re‑parses it with `ParseMultipleCommands` (looks for `[`). This round‑trip collides with `[`/`]` that legitimately appear in a *single* command:

- A single action such as `web:https://example.com/watch[id]`, `~folder[1]`, or a `cmd:`/`shell:` expression containing brackets has `raw.find(L'[') != npos`, so `ParseMultipleCommands` treats it as a multi‑command list. Everything **outside** the brackets is silently discarded — `web:https://example.com/watch[id]` runs the bare process `id` instead of opening the URL.
- The `[a][b]` join is also not escaped, so a genuine multi‑command list where one command contains `]` (e.g. `cmd:echo ]`) splits at the wrong place.

The clean fix is to stop re‑encoding: make the action a `std::vector<std::wstring>` end‑to‑end (`ActionItem` already holds a vector‑worth of data — just keep `actionArray` instead of joining it), and have `ExecuteActionText` take the vector directly. That removes `ParseMultipleCommands` entirely and the bracket‑collision class of bugs with it. This applies to both the top‑level `Action` and the submenu `action` paths.

**2. `.exe`/`.dll` menu icons leak a temp PNG on every flyout open.**

`MakeMenuIcon` now resolves exe/dll icons via `SaveIconToPng`, which writes `%TEMP%\ICOxxxx.tmp.png` and returns the path for `BitmapIcon.UriSource`. The `AutoDelete` guard only deletes the `.tmp` *placeholder* — the actual `ICOxxxx.tmp.png` is never deleted. Because `AddMenuItems` rebuilds the flyout (and calls `MakeMenuIcon`) on **every** menu open, each open of a submenu that contains an exe/dll‑sourced icon drops a new orphaned PNG in `%TEMP%`. (v2.1 rendered menu icons from an in‑memory data URI and didn't touch disk, so this is a regression for menu icons specifically.)

Options, cheapest first:
- Cache the resolved PNG path keyed by the icon string, so a given icon writes at most one file and repeated opens reuse it; delete the cached files in `Wh_ModUninit`.
- Or avoid the file entirely: return a `wuxc::ImageIcon` (it derives from `IconElement`, so it's valid for `MenuFlyoutItem::Icon`) fed by the in‑memory `BitmapSource` you already build in `CreateBitmapSourceFromIcon`/`ResolveExeIconInMemory` — same approach `MakeButtonIcon` already prefers.

<p>

- **Now‑unused dependency.** Removing `CreateDataUriFromIcon` dropped the only user of `CryptBinaryToStringW`, so `#include <wincrypt.h>` (line 313) and `-lcrypt32` in `@compilerOptions` are now dead — remove both. (`-lcomctl32` also has no user in the file, if you want to trim it while you're there.)
- **`LoadSubmenuRecursive`'s `depth` parameter is never used.** It's computed and passed down but never checked, so the "up to 3 levels deep" limit is only enforced by the settings UI shape, not the reader — the code will happily recurse as deep as the settings paths resolve. Either enforce the cap with `depth` or drop the parameter.
- **Trailing whitespace in a couple of `$name` values** (`Preset language    `, `Left-click submenu `). Harmless, just untidy.
- **`Sleep(50)` between multi‑commands runs on the caller thread.** For left‑click submenu items, `ExecuteActionText` runs on the Start menu UI thread, so an N‑command item blocks the UI for ~`50·(N‑1)` ms. Consider sequencing the commands inside a single worker thread instead of sleeping on the caller.

</p>
</details>

<p>

- **Actions run on detached threads that aren't joined on unload.** `ExecuteActionText` does `std::thread(...).detach()` (and some sleep 150 ms before acting). If the mod is disabled while one of these is still in flight, the thread is executing lambda code that lives in the mod image, which can be unmapped underneath it → potential crash. This predates this PR and the window is narrow (user triggers an action and disables the mod within a few ms), so it's an FYI rather than a blocker — but tracking/joining these threads in `Wh_ModBeforeUninit` would close it.

</p>
</details>

Salyts:
@m417z

PR #4823, opened 2026-07-18
UPDATE: Added Control Panel links restoration and enhanced the interface

babamohammed2022:
## Changelog

- Added Ethernet support
- Restored "Connect to a network" and "HomeGroup/sharing" links in Network Center Control Page like in Windows 7
- Enhanced the interface to make it more similar to the original Windows 7 one
- Better connection timeouts and auth error handling
- Enhanced the stability

## Mod authorship

This mod was created by:

- - [X] Claude
- - [ ] ChatGPT
- - [ ] Gemini


m417z:
### Submission review

Personal note: If you find yourself having to embed binary files, it might be a sign that it's not a good fit for a mod, and should be a standalone tool instead. I didn't read the generated review below past item 1.

---

**1. Embedding a compiled PE and writing it to `%TEMP%` is the central concern.** The mod embeds a ~6 KB compiled DLL as a hex byte array (`kIconsDll[]`), writes it to `%TEMP%\Win7NetworkCenterLinks\win7ncl_icons_<pid>.dll` on init, and has DirectUI `LoadLibrary` it (via a process‑wide `LoadLibraryExW` redirect) just to supply two 24×24 icons. Several problems with this:

- **Not auditable.** An opaque compiled binary blob can't be reviewed from the source. Windhawk mods are meant to be source‑auditable, and there's no way for a reviewer (or a user) to confirm what that PE actually contains or does when loaded.
- **Loading an executable from a user‑writable location.** `%TEMP%` is user‑writable; writing a DLL there and loading it is a pattern worth avoiding on principle, even though the redirect to a fixed path + `FileMatchesEmbedded` verification and the per‑user nature of `%TEMP%` keep the practical hijack risk low here.
- **Stale files accumulate.** `CleanupIconsDll()` only runs from `Wh_ModUninit`, which is **not** called when the host process terminates (Explorer restart, sign‑out, reboot, crash). Since the temp file is named per‑PID, every such exit leaves a `win7ncl_icons_<oldpid>.dll` behind, and they pile up in `%TEMP%` over time.
- **It's cosmetic.** There's already a working stock‑icon fallback (`icon(22,…)` / `icon(27,…)` from `netcenter.dll` when `!g_dllReady`), so all of the above buys only icon fidelity.

My recommendation is to reconsider whether the custom icons justify shipping + writing + loading an opaque PE at all — the stock fallback already renders the links. If you do keep it, at minimum: sweep and delete any leftover `win7ncl_icons_*.dll` on init (since uninit won't run at process exit), and do none of this when the feature is disabled (next point).

**2. The DLL is written and all hooks are installed even when `restoreClassicNetworkCenterLinks` is off.** `Init()` calls `EnsureIconsDll()` and `HookAll()` unconditionally, and `SettingsChanged()` calls `EnsureIconsDll()` unconditionally — the setting only gates `Patch()`. So a user who disables the feature still gets the temp DLL written to disk on every process start (and on every settings change). At least gate the disk write on the setting:

```cpp
static bool Init() {
    bool enabled = Wh_GetIntSetting(L"restoreClassicNetworkCenterLinks") != 0;
    g_addConnect = g_addHomegroup = enabled;
    if (enabled) EnsureIconsDll();   // don't touch the disk when disabled
    if (!HookAll()) { CleanupIconsDll(); return false; }
    return true;
}
```

(Keeping the DUI hook installed while disabled is defensible so the toggle works live, but the file write and the redundant `dui70.dll` force‑load shouldn't happen for a feature that's off.)

<p>

- **`SetXML_Hook` is a no‑op installed only to capture the trampoline.** It forwards `SetXML_Orig(t, x, r, h)` unchanged and exists solely so `SetXMLFromResource_Hook` can call the original. `SetXML` is a **public** member (`?SetXML@…@@QEAA…`), so you can just `GetProcAddress` it and cast to `SetXML_t`, and drop the hook — that avoids adding your trampoline to a hot DirectUI path (`SetXML` runs for lots of shell UI in explorer.exe) for no functional reason.

- **`debugEthernet` is dead scaffolding.** The setting is commented out (it lives in a `/* */` block *after* `==/WindhawkModSettings==`), so `Wh_GetStringSetting(L"debugEthernet")` always returns `L""` and `g_Settings.debugEthernet` is permanently `0`. That makes the debug branches in `UpdateEthernetStatus`, the `debugEthernet != 2` guards, and the `oldDebug` tracking in `Wh_ModSettingsChanged` all unreachable. Since `debugEthernet != 2` is always true, `showWifiList` reduces to `g_NetworkCount > 0`. Consider removing the whole `debugEthernet` plumbing (it reads like a leftover AI/testing artifact).

- **Raw `Wh_SetFunctionHook` + `void*` casts** in `HookAll()` — the modern idiom is `WindhawkUtils::SetFunctionHook()` for type safety (0 uses currently vs. 3 raw calls).

- **In `control.exe` the flyout‑only init runs needlessly.** `Wh_ModInit` runs `LoadSystemIcons()`, `InitGdiPlusRendering()`, `DarkContextMenu::Init()`, `CreateMutexW`, `InitializeCriticalSection` before the `IsExplorerProcess()` early‑return, even though only the Network Center hooks are relevant in `control.exe`. Cheap to move those after the gate.

- **`InitGdiPlusRendering()` uses the bare‑name `LoadLibraryW(L"gdiplus.dll")`** (line ~1279) while the rest of the file already uses the safe `LoadLibraryExW(L"gdiplus.dll", NULL, LOAD_LIBRARY_SEARCH_SYSTEM32)` form (line ~1106). `gdiplus.dll` is not a KnownDLL; the hijack surface is negligible since the mod only targets explorer.exe/control.exe (both run from System32), but it's a one‑line consistency fix.

- **`-loleaut32`** appears unused as an import — `SysFreeString` is resolved via `GetProcAddress` and there's no other oleaut32 symbol linked directly. Safe to drop from `@compilerOptions`.

</p>
</details>

<p>

- **The `LoadLibraryExW` hook fires on every DLL load in explorer.exe** (explorer loads a lot of DLLs — shell extensions, thumbnail/property handlers, etc.), running `IsOurDll()` string compares each time. The cost per call is small and, given the `library(win7ncl_icons.dll)` redirect approach, there's no obviously cleaner interception point — just noting it as an inherent cost of this design (another reason to prefer the stock‑icon fallback).

- **Ethernet detection runs `GetAdaptersAddresses` + a full NLM COM network/connection enumeration on every refresh tick** (default 3 s). That's fine functionally (it's on the flyout's COM‑initialized hotkey thread, and `g_pNLM` is created/released consistently there), just heavier than the pure Wi‑Fi path — acceptable while the flyout is open.

</p>
</details>

babamohammed2022:
> ### Submission review
> 
> 
> Personal note: If you find yourself having to embed binary files, it might be a sign that it's not a good fit for a mod, and should be a standalone tool instead. I didn't read the generated review below past item 1.
> 
> **1. Embedding a compiled PE and writing it to `%TEMP%` is the central concern.** The mod embeds a ~6 KB compiled DLL as a hex byte array (`kIconsDll[]`), writes it to `%TEMP%\Win7NetworkCenterLinks\win7ncl_icons_<pid>.dll` on init, and has DirectUI `LoadLibrary` it (via a process‑wide `LoadLibraryExW` redirect) just to supply two 24×24 icons. Several problems with this:
> 
> * **Not auditable.** An opaque compiled binary blob can't be reviewed from the source. Windhawk mods are meant to be source‑auditable, and there's no way for a reviewer (or a user) to confirm what that PE actually contains or does when loaded.
> * **Loading an executable from a user‑writable location.** `%TEMP%` is user‑writable; writing a DLL there and loading it is a pattern worth avoiding on principle, even though the redirect to a fixed path + `FileMatchesEmbedded` verification and the per‑user nature of `%TEMP%` keep the practical hijack risk low here.
> * **Stale files accumulate.** `CleanupIconsDll()` only runs from `Wh_ModUninit`, which is **not** called when the host process terminates (Explorer restart, sign‑out, reboot, crash). Since the temp file is named per‑PID, every such exit leaves a `win7ncl_icons_<oldpid>.dll` behind, and they pile up in `%TEMP%` over time.
> * **It's cosmetic.** There's already a working stock‑icon fallback (`icon(22,…)` / `icon(27,…)` from `netcenter.dll` when `!g_dllReady`), so all of the above buys only icon fidelity.
> 
> My recommendation is to reconsider whether the custom icons justify shipping + writing + loading an opaque PE at all — the stock fallback already renders the links. If you do keep it, at minimum: sweep and delete any leftover `win7ncl_icons_*.dll` on init (since uninit won't run at process exit), and do none of this when the feature is disabled (next point).
> 
> **2. The DLL is written and all hooks are installed even when `restoreClassicNetworkCenterLinks` is off.** `Init()` calls `EnsureIconsDll()` and `HookAll()` unconditionally, and `SettingsChanged()` calls `EnsureIconsDll()` unconditionally — the setting only gates `Patch()`. So a user who disables the feature still gets the temp DLL written to disk on every process start (and on every settings change). At least gate the disk write on the setting:
> 
> ```c++
> static bool Init() {
>     bool enabled = Wh_GetIntSetting(L"restoreClassicNetworkCenterLinks") != 0;
>     g_addConnect = g_addHomegroup = enabled;
>     if (enabled) EnsureIconsDll();   // don't touch the disk when disabled
>     if (!HookAll()) { CleanupIconsDll(); return false; }
>     return true;
> }
> ```
> 
> (Keeping the DUI hook installed while disabled is defensible so the toggle works live, but the file write and the redundant `dui70.dll` force‑load shouldn't happen for a feature that's off.)
> 
> Optional improvements
> 
> 
> * **`SetXML_Hook` is a no‑op installed only to capture the trampoline.** It forwards `SetXML_Orig(t, x, r, h)` unchanged and exists solely so `SetXMLFromResource_Hook` can call the original. `SetXML` is a **public** member (`?SetXML@…@@QEAA…`), so you can just `GetProcAddress` it and cast to `SetXML_t`, and drop the hook — that avoids adding your trampoline to a hot DirectUI path (`SetXML` runs for lots of shell UI in explorer.exe) for no functional reason.
> * **`debugEthernet` is dead scaffolding.** The setting is commented out (it lives in a `/* */` block _after_ `==/WindhawkModSettings==`), so `Wh_GetStringSetting(L"debugEthernet")` always returns `L""` and `g_Settings.debugEthernet` is permanently `0`. That makes the debug branches in `UpdateEthernetStatus`, the `debugEthernet != 2` guards, and the `oldDebug` tracking in `Wh_ModSettingsChanged` all unreachable. Since `debugEthernet != 2` is always true, `showWifiList` reduces to `g_NetworkCount > 0`. Consider removing the whole `debugEthernet` plumbing (it reads like a leftover AI/testing artifact).
> * **Raw `Wh_SetFunctionHook` + `void*` casts** in `HookAll()` — the modern idiom is `WindhawkUtils::SetFunctionHook()` for type safety (0 uses currently vs. 3 raw calls).
> * **In `control.exe` the flyout‑only init runs needlessly.** `Wh_ModInit` runs `LoadSystemIcons()`, `InitGdiPlusRendering()`, `DarkContextMenu::Init()`, `CreateMutexW`, `InitializeCriticalSection` before the `IsExplorerProcess()` early‑return, even though only the Network Center hooks are relevant in `control.exe`. Cheap to move those after the gate.
> * **`InitGdiPlusRendering()` uses the bare‑name `LoadLibraryW(L"gdiplus.dll")`** (line ~1279) while the rest of the file already uses the safe `LoadLibraryExW(L"gdiplus.dll", NULL, LOAD_LIBRARY_SEARCH_SYSTEM32)` form (line ~1106). `gdiplus.dll` is not a KnownDLL; the hijack surface is negligible since the mod only targets explorer.exe/control.exe (both run from System32), but it's a one‑line consistency fix.
> * **`-loleaut32`** appears unused as an import — `SysFreeString` is resolved via `GetProcAddress` and there's no other oleaut32 symbol linked directly. Safe to drop from `@compilerOptions`.
> 
> Functionality notes

You are right, I've removed the DLL and replaced it with base64, initially I used the DLL because it worked decently compared to the base64 but I've enhanced the base64 implementation and made it more robust and I've also tried to address the other reported issues

babamohammed2022:
> ### Submission review
> 
> 
> Personal note: If you find yourself having to embed binary files, it might be a sign that it's not a good fit for a mod, and should be a standalone tool instead. I didn't read the generated review below past item 1.
> 
> **1. Embedding a compiled PE and writing it to `%TEMP%` is the central concern.** The mod embeds a ~6 KB compiled DLL as a hex byte array (`kIconsDll[]`), writes it to `%TEMP%\Win7NetworkCenterLinks\win7ncl_icons_<pid>.dll` on init, and has DirectUI `LoadLibrary` it (via a process‑wide `LoadLibraryExW` redirect) just to supply two 24×24 icons. Several problems with this:
> 
> * **Not auditable.** An opaque compiled binary blob can't be reviewed from the source. Windhawk mods are meant to be source‑auditable, and there's no way for a reviewer (or a user) to confirm what that PE actually contains or does when loaded.
> * **Loading an executable from a user‑writable location.** `%TEMP%` is user‑writable; writing a DLL there and loading it is a pattern worth avoiding on principle, even though the redirect to a fixed path + `FileMatchesEmbedded` verification and the per‑user nature of `%TEMP%` keep the practical hijack risk low here.
> * **Stale files accumulate.** `CleanupIconsDll()` only runs from `Wh_ModUninit`, which is **not** called when the host process terminates (Explorer restart, sign‑out, reboot, crash). Since the temp file is named per‑PID, every such exit leaves a `win7ncl_icons_<oldpid>.dll` behind, and they pile up in `%TEMP%` over time.
> * **It's cosmetic.** There's already a working stock‑icon fallback (`icon(22,…)` / `icon(27,…)` from `netcenter.dll` when `!g_dllReady`), so all of the above buys only icon fidelity.
> 
> My recommendation is to reconsider whether the custom icons justify shipping + writing + loading an opaque PE at all — the stock fallback already renders the links. If you do keep it, at minimum: sweep and delete any leftover `win7ncl_icons_*.dll` on init (since uninit won't run at process exit), and do none of this when the feature is disabled (next point).
> 
> **2. The DLL is written and all hooks are installed even when `restoreClassicNetworkCenterLinks` is off.** `Init()` calls `EnsureIconsDll()` and `HookAll()` unconditionally, and `SettingsChanged()` calls `EnsureIconsDll()` unconditionally — the setting only gates `Patch()`. So a user who disables the feature still gets the temp DLL written to disk on every process start (and on every settings change). At least gate the disk write on the setting:
> 
> ```c++
> static bool Init() {
>     bool enabled = Wh_GetIntSetting(L"restoreClassicNetworkCenterLinks") != 0;
>     g_addConnect = g_addHomegroup = enabled;
>     if (enabled) EnsureIconsDll();   // don't touch the disk when disabled
>     if (!HookAll()) { CleanupIconsDll(); return false; }
>     return true;
> }
> ```
> 
> (Keeping the DUI hook installed while disabled is defensible so the toggle works live, but the file write and the redundant `dui70.dll` force‑load shouldn't happen for a feature that's off.)
> 
> Optional improvements
> 
> 
> * **`SetXML_Hook` is a no‑op installed only to capture the trampoline.** It forwards `SetXML_Orig(t, x, r, h)` unchanged and exists solely so `SetXMLFromResource_Hook` can call the original. `SetXML` is a **public** member (`?SetXML@…@@QEAA…`), so you can just `GetProcAddress` it and cast to `SetXML_t`, and drop the hook — that avoids adding your trampoline to a hot DirectUI path (`SetXML` runs for lots of shell UI in explorer.exe) for no functional reason.
> * **`debugEthernet` is dead scaffolding.** The setting is commented out (it lives in a `/* */` block _after_ `==/WindhawkModSettings==`), so `Wh_GetStringSetting(L"debugEthernet")` always returns `L""` and `g_Settings.debugEthernet` is permanently `0`. That makes the debug branches in `UpdateEthernetStatus`, the `debugEthernet != 2` guards, and the `oldDebug` tracking in `Wh_ModSettingsChanged` all unreachable. Since `debugEthernet != 2` is always true, `showWifiList` reduces to `g_NetworkCount > 0`. Consider removing the whole `debugEthernet` plumbing (it reads like a leftover AI/testing artifact).
> * **Raw `Wh_SetFunctionHook` + `void*` casts** in `HookAll()` — the modern idiom is `WindhawkUtils::SetFunctionHook()` for type safety (0 uses currently vs. 3 raw calls).
> * **In `control.exe` the flyout‑only init runs needlessly.** `Wh_ModInit` runs `LoadSystemIcons()`, `InitGdiPlusRendering()`, `DarkContextMenu::Init()`, `CreateMutexW`, `InitializeCriticalSection` before the `IsExplorerProcess()` early‑return, even though only the Network Center hooks are relevant in `control.exe`. Cheap to move those after the gate.
> * **`InitGdiPlusRendering()` uses the bare‑name `LoadLibraryW(L"gdiplus.dll")`** (line ~1279) while the rest of the file already uses the safe `LoadLibraryExW(L"gdiplus.dll", NULL, LOAD_LIBRARY_SEARCH_SYSTEM32)` form (line ~1106). `gdiplus.dll` is not a KnownDLL; the hijack surface is negligible since the mod only targets explorer.exe/control.exe (both run from System32), but it's a one‑line consistency fix.
> * **`-loleaut32`** appears unused as an import — `SysFreeString` is resolved via `GetProcAddress` and there's no other oleaut32 symbol linked directly. Safe to drop from `@compilerOptions`.
> 
> Functionality notes

If possible, let me know if the issues have been addressed properly now

m417z:
### Submission review

---

- **Remove warning suppression.** Remove these lines:

```cpp
#define WIN32_LEAN_AND_MEAN
#undef _WINSOCKAPI_
#pragma clang diagnostic push
#pragma clang diagnostic ignored "-W#warnings"
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wcpp"
```

```cpp
#pragma GCC diagnostic pop
#pragma clang diagnostic pop
```

The proper fix is adding `-DWIN32_LEAN_AND_MEAN` to `@compilerOptions`.

- **`useRoundedCorners` default contradicts its own description** (settings block, lines 100–102). The default is `true`, but the `$description` says *"Disabled by default for classic theme compatibility."* One of the two is stale — for a Windows 7 classic recreation, defaulting rounded corners **off** (matching the description) seems more consistent, but either way please make the default and the description agree so the setting isn't misleading.

<p>

- **Use `WindhawkUtils::StringSetting` instead of raw `Wh_GetStringSetting` + `Wh_FreeStringSetting`** in `LoadSettings` (lines 194/204 for `language`, 208/213 for `theme`). It's the RAII form and removes the manual free. Related: `Wh_GetStringSetting` never returns `NULL` (it returns `L""` on error/unset), so the `if (lang)` / `if (theme)` guards are always true and can be dropped.
- **Unused includes/libs.** `#include <math.h>` (line 146) and `#include <psapi.h>` (line 151) aren't used (no `math.h`/psapi symbols appear anywhere), and `-lpsapi` in `@compilerOptions` (line 11) is likewise unused. Removing them keeps the build lean.
- **`UnregisterClassW(L"Win7NetFlyoutHidden", ...)` in `Wh_ModUninit`** (line 5201) references a class that is never registered anywhere in the file — dead line, safe to remove.
- **Cross-thread timer calls in `Wh_ModSettingsChanged`** (lines 5177–5183): `KillTimer`/`SetTimer` on `g_hWndFlyout` run on the arbitrary settings-changed thread, but that window is owned by the hotkey thread. This only fires on a settings change and is low-impact, but it would be cleaner to post the refresh-timer update to the owning thread (as `ToggleFlyoutWindow` already does for cross-thread requests via `PostThreadMessageW`).
- **Add a `< g_NetworkCount` upper bound** to the two `g_hWndCheckboxConnect` branches in `WM_CTLCOLORSTATIC`/`WM_CTLCOLORBTN` (lines 2433 and 2477). Today they guard only `g_SelectedRowIndex >= 0`; the index is always ≤ 49 so this is in-bounds (not an OOB), but after a refresh shrinks the list it can read a stale entry's `connState` and pick the wrong color. Matching every other access site (`… && g_SelectedRowIndex < g_NetworkCount`) makes it robust.
- **Defensive `pIfList` deref inconsistency** (RefreshWifiData): line 1758 guards `if (pIfList)` before iterating, but line 1766 dereferences `pIfList->dwNumberOfItems` unguarded. `WlanEnumInterfaces` won't return `ERROR_SUCCESS` with a NULL list, so this is only a consistency nit, but the two sites should match.

</p>
</details>

<p>

- **Network Center link icons don't scale with DPI.** `LoadImageW_Hook` (lines 5011–5015) discards the `width`/`height` DirectUI requests and returns `CopyNetworkCenterIcon(...)`, which hands back a copy of an icon built once at a fixed 24×24 (`CreateIconFromBase64PNG(base64, 24, 24)`, line 1460). The UIFILE requests the icon at a DPI-relative size (`24rp`), so on a 150%/200% display the two restored links get an undersized/upscaled icon while the stock page icons scale correctly. `CreateIconFromBase64PNG` already supports an arbitrary target size, so passing the requested `width`/`height` through (and keying the cache by size, or regenerating when it differs) would make them crisp. Purely cosmetic — no crash.
- **Password dialog is a fixed 420×180 px** (`PromptNetworkPassword`, lines 2570–2577), not DPI-scaled, while the main flyout is DPI-aware via `RecalcDpiMetrics`. On high-DPI setups the dialog frame will look small relative to the rest of the UI. Minor.

</p>
</details>

babamohammed2022:
> ### Submission review
> 
> 
> * **Remove warning suppression.** Remove these lines:
> 
> ```c++
> #define WIN32_LEAN_AND_MEAN
> #undef _WINSOCKAPI_
> #pragma clang diagnostic push
> #pragma clang diagnostic ignored "-W#warnings"
> #pragma GCC diagnostic push
> #pragma GCC diagnostic ignored "-Wcpp"
> ```
> 
> ```c++
> #pragma GCC diagnostic pop
> #pragma clang diagnostic pop
> ```
> 
> The proper fix is adding `-DWIN32_LEAN_AND_MEAN` to `@compilerOptions`.
> 
> * **`useRoundedCorners` default contradicts its own description** (settings block, lines 100–102). The default is `true`, but the `$description` says _"Disabled by default for classic theme compatibility."_ One of the two is stale — for a Windows 7 classic recreation, defaulting rounded corners **off** (matching the description) seems more consistent, but either way please make the default and the description agree so the setting isn't misleading.
> 
> Optional improvements
> 
> 
> * **Use `WindhawkUtils::StringSetting` instead of raw `Wh_GetStringSetting` + `Wh_FreeStringSetting`** in `LoadSettings` (lines 194/204 for `language`, 208/213 for `theme`). It's the RAII form and removes the manual free. Related: `Wh_GetStringSetting` never returns `NULL` (it returns `L""` on error/unset), so the `if (lang)` / `if (theme)` guards are always true and can be dropped.
> * **Unused includes/libs.** `#include <math.h>` (line 146) and `#include <psapi.h>` (line 151) aren't used (no `math.h`/psapi symbols appear anywhere), and `-lpsapi` in `@compilerOptions` (line 11) is likewise unused. Removing them keeps the build lean.
> * **`UnregisterClassW(L"Win7NetFlyoutHidden", ...)` in `Wh_ModUninit`** (line 5201) references a class that is never registered anywhere in the file — dead line, safe to remove.
> * **Cross-thread timer calls in `Wh_ModSettingsChanged`** (lines 5177–5183): `KillTimer`/`SetTimer` on `g_hWndFlyout` run on the arbitrary settings-changed thread, but that window is owned by the hotkey thread. This only fires on a settings change and is low-impact, but it would be cleaner to post the refresh-timer update to the owning thread (as `ToggleFlyoutWindow` already does for cross-thread requests via `PostThreadMessageW`).
> * **Add a `< g_NetworkCount` upper bound** to the two `g_hWndCheckboxConnect` branches in `WM_CTLCOLORSTATIC`/`WM_CTLCOLORBTN` (lines 2433 and 2477). Today they guard only `g_SelectedRowIndex >= 0`; the index is always ≤ 49 so this is in-bounds (not an OOB), but after a refresh shrinks the list it can read a stale entry's `connState` and pick the wrong color. Matching every other access site (`… && g_SelectedRowIndex < g_NetworkCount`) makes it robust.
> * **Defensive `pIfList` deref inconsistency** (RefreshWifiData): line 1758 guards `if (pIfList)` before iterating, but line 1766 dereferences `pIfList->dwNumberOfItems` unguarded. `WlanEnumInterfaces` won't return `ERROR_SUCCESS` with a NULL list, so this is only a consistency nit, but the two sites should match.
> 
> Functionality notes

Thanks for the review. If required, let me know if there are any other issues to address

m417z:
### Submission review

---

**Enabling "Restore classic Network Center links" at runtime does nothing until a reload.** `Win7NetworkCenterLinks::SettingsChanged()` calls `HookAll()` → `WindhawkUtils::SetFunctionHook(...)` when the setting is turned on mid-session (lines ~5121–5128, ~5095–5104), but it never calls `Wh_ApplyHookOperations()`. Per the API header, `Wh_ApplyHookOperations` is *"Called automatically by Windhawk after `Wh_ModInit`"* — so hooks registered later (in `Wh_ModSettingsChanged`) stay pending and never activate. Since the mod uses the `void Wh_ModSettingsChanged()` variant it also doesn't request a reload, so a user who had the option off and switches it on gets a silently non-functional feature until the mod is reloaded / Explorer restarts. (Disabling at runtime works, because `g_addConnect/g_addHomegroup` make `Patch()` a no-op.)

Two ways to fix:
- Simplest / matches the API guidance ("all hooks should be set in `Wh_ModInit`"): install the DirectUI + `LoadImageW` hooks unconditionally in `Init()` and let the existing `g_addConnect/g_addHomegroup` flags gate the behavior (they already turn `Patch` into a no-op when off). This also removes the runtime hook-application concern entirely.
- Or keep the lazy install and add `Wh_ApplyHookOperations();` right after a successful `HookAll()` in `SettingsChanged()`.

<p>

- **Stale version strings.** `@version` is `3.2.0`, but `Wh_ModInit` logs `=== Wh_ModInit v3.1.0 ===` (line 5136), the Network Center Links section header says `v3.1.0` (line ~4749), and the in-file `## Changelog` block (lines ~115–122) still only lists 3.1.0 items with no 3.2.0 entry. Worth syncing so the log/changelog match the shipped version.

- **Dead preprocessor block.** Lines 123–126:
  ```cpp
  #ifndef UNICODE
  #define UNICODE
  #include <psapi.h>
  #endif
  ```
  Windhawk always compiles with `UNICODE` defined, so this block never runs, and `<psapi.h>` is included unconditionally at line 137 anyway. It can be removed.

- **Back-to-back connects can briefly freeze the flyout.** In `AskForPasswordAndConnect` (lines ~2919–2922), starting a new connection does `WaitForSingleObject(g_hConnectThread, 5000)` on the flyout's own (UI) thread. If the previous `WlanConnect` is still in flight, the flyout stops repainting/responding for up to 5 s. Edge case (only on rapid reconnect to a second network), but you could hand the old handle off to be closed asynchronously instead of blocking the UI thread.

</p>
</details>

<p>

- **Context menu "Status" and "Properties" do the same thing.** Both `IDM_STATUS` and `IDM_PROPERTIES` (lines ~3485–3488) `ShellExecute` the same `shell:::{7007ACC7-3202-11D1-AAD2-00805FC1270E}` ("Connect to a Network") folder rather than opening the connection's status / adapter properties. Might be intentional, but the labels imply different destinations.

</p>
</details>

babamohammed2022:
> ### Submission review
> **Enabling "Restore classic Network Center links" at runtime does nothing until a reload.** `Win7NetworkCenterLinks::SettingsChanged()` calls `HookAll()` → `WindhawkUtils::SetFunctionHook(...)` when the setting is turned on mid-session (lines ~5121–5128, ~5095–5104), but it never calls `Wh_ApplyHookOperations()`. Per the API header, `Wh_ApplyHookOperations` is _"Called automatically by Windhawk after `Wh_ModInit`"_ — so hooks registered later (in `Wh_ModSettingsChanged`) stay pending and never activate. Since the mod uses the `void Wh_ModSettingsChanged()` variant it also doesn't request a reload, so a user who had the option off and switches it on gets a silently non-functional feature until the mod is reloaded / Explorer restarts. (Disabling at runtime works, because `g_addConnect/g_addHomegroup` make `Patch()` a no-op.)
> 
> Two ways to fix:
> 
> * Simplest / matches the API guidance ("all hooks should be set in `Wh_ModInit`"): install the DirectUI + `LoadImageW` hooks unconditionally in `Init()` and let the existing `g_addConnect/g_addHomegroup` flags gate the behavior (they already turn `Patch` into a no-op when off). This also removes the runtime hook-application concern entirely.
> * Or keep the lazy install and add `Wh_ApplyHookOperations();` right after a successful `HookAll()` in `SettingsChanged()`.
> 
> Optional improvements
> 
> 
> * **Stale version strings.** `@version` is `3.2.0`, but `Wh_ModInit` logs `=== Wh_ModInit v3.1.0 ===` (line 5136), the Network Center Links section header says `v3.1.0` (line ~4749), and the in-file `## Changelog` block (lines ~115–122) still only lists 3.1.0 items with no 3.2.0 entry. Worth syncing so the log/changelog match the shipped version.
> * **Dead preprocessor block.** Lines 123–126:
>   ```c++
>   #ifndef UNICODE
>   #define UNICODE
>   #include <psapi.h>
>   #endif
>   ```
>   
>   
>       
>         
>       
>   
>         
>       
>   
>       
>     
>   Windhawk always compiles with `UNICODE` defined, so this block never runs, and `<psapi.h>` is included unconditionally at line 137 anyway. It can be removed.
> * **Back-to-back connects can briefly freeze the flyout.** In `AskForPasswordAndConnect` (lines ~2919–2922), starting a new connection does `WaitForSingleObject(g_hConnectThread, 5000)` on the flyout's own (UI) thread. If the previous `WlanConnect` is still in flight, the flyout stops repainting/responding for up to 5 s. Edge case (only on rapid reconnect to a second network), but you could hand the old handle off to be closed asynchronously instead of blocking the UI thread.
> 
> Functionality notes
> 
> 
> * **Context menu "Status" and "Properties" do the same thing.** Both `IDM_STATUS` and `IDM_PROPERTIES` (lines ~3485–3488) `ShellExecute` the same `shell:::{7007ACC7-3202-11D1-AAD2-00805FC1270E}` ("Connect to a Network") folder rather than opening the connection's status / adapter properties. Might be intentional, but the labels imply different destinations.

OK, I've addressed all the reported issues except for the last one because the redirect is intentional (at least for now) as I have not found the proper documentation to open the real "Properties" dialog despite it being still present in Windows 10/11

PR #4825, opened 2026-07-18
Bt battery monitor v1.0.0

BlackPaw21:
## Changelog

* Changelog item 1...
* Changelog item 2...

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini
- - [x] Another AI (please specify): stupid OpenCode

### DIFF
--- a/mods/bt-battery-monitor.wh.cpp
+++ b/mods/bt-battery-monitor.wh.cpp
@@ -8,7 +8,7 @@
 // @donateUrl       https://ko-fi.com/blackpaw21
 // @include         windhawk.exe
 // @license         MIT
-// @compilerOptions -lbluetoothapis -lbthprops -lsetupapi -lcfgmgr32 -lgdi32 -luser32 -lshell32 -lole32 -luuid -lcomctl32 -lcomdlg32
+// @compilerOptions -lbluetoothapis -lbthprops -lsetupapi -lcfgmgr32 -lgdi32 -luser32 -lshell32 -lole32 -luuid -lcomctl32 -lcomdlg32 -ffp-exception-behavior=maytrap
 // ==/WindhawkMod==
 
 // ==WindhawkModReadme==
@@ -32,9 +32,11 @@ See every connected Bluetooth device's battery level right in your system tray �
 4. **Choose Icons** — Pick a preset icon per device type, or click **Browse...** to load a custom `.ico` file.
 5. Click **Save and Apply** — the tray icon updates immediately, no restart needed.
 
-### Left-click to check on your devices
+### Left-click to rescan
 
-Left-click the tray icon to open a neat flyout listing every paired device with its current battery percentage. At a glance you'll see which ones are doing fine and which ones need charging.
+Left-click the tray icon to trigger an immediate Bluetooth rescan. Use it when you've just paired or disconnected a device and want the battery readout to refresh without waiting for the next polling interval.
+
+### Right-click device list
 
 ### Low Battery Warning
 
@@ -274,7 +276,7 @@ static HICON CreateColorIcon(BYTE r, BYTE g, BYTE b, int size) {
     return hIcon;
 }
 
-// Generate a white X icon for the "no connected devices" state
+// Generate a red X icon for the "no connected devices" state
 static HICON CreateXIcon(int size) {
     BITMAPINFO bi = {};
     bi.bmiHeader.biSize = sizeof(BITMAPINFOHEADER);
@@ -296,7 +298,7 @@ static HICON CreateXIcon(int size) {
     for (int y = 0; y < size; y++) {
         for (int x = 0; x < size; x++) {
             if (abs(x - y) <= t / 2 || abs(x + y - last) <= t / 2)
-                pixels[y * size + x] = 0xFFFFFFFF;
+                pixels[y * size + x] = 0xFFFF0000;
         }
     }
 
@@ -307,16 +309,16 @@ static HICON CreateXIcon(int size) {
     if (!hdcMask) { DeleteObject(hbmColor); DeleteObject(hbmMask); return NULL; }
     HBITMAP hOldBmp = (HBITMAP)SelectObject(hdcMask, hbmMask);
     RECT r = {0, 0, size, size};
-    FillRect(hdcMask, &r, (HBRUSH)GetStockObject(WHITE_BRUSH));
-    HBRUSH hbrBlack = CreateSolidBrush(RGB(0, 0, 0));
-    HPEN hBlackPen = CreatePen(PS_SOLID, t, RGB(0, 0, 0));
-    SelectObject(hdcMask, hbrBlack);
-    SelectObject(hdcMask, hBlackPen);
+    FillRect(hdcMask, &r, (HBRUSH)GetStockObject(BLACK_BRUSH));
+    HBRUSH hbrWhite = CreateSolidBrush(RGB(255, 255, 255));
+    HPEN hWhitePen = CreatePen(PS_SOLID, t, RGB(255, 255, 255));
+    SelectObject(hdcMask, hbrWhite);
+    SelectObject(hdcMask, hWhitePen);
     MoveToEx(hdcMask, 0, 0, NULL); LineTo(hdcMask, size, size);
     MoveToEx(hdcMask, size, 0, NULL); LineTo(hdcMask, 0, size);
     SelectObject(hdcMask, hOldBmp);
-    DeleteObject(hbrBlack);
-    DeleteObject(hBlackPen);
+    DeleteObject(hbrWhite);
+    DeleteObject(hWhitePen);
     DeleteDC(hdcMask);
 
     ICONINFO ii = {};
@@ -356,7 +358,8 @@ static bool CreateIcons() {
     DestroyIcons();
     LoadSettings();
 
-    g_hIconDisconnected = CreateXIcon(16);
+    if (!g_hIconDisconnected)
+        g_hIconDisconnected = CreateXIcon(16);
 
     g_hIconLowBatRed = CreateColorIcon(255, 0, 0, 16);
     g_hIconLowBatBlack = CreateColorIcon(0, 0, 0, 16);
@@ -752,7 +755,7 @@ static void RefreshBthleConnectedState(std::vector<DeviceInfo>& devices) {
 static unsigned int __stdcall ScannerProc(void*) {
     HRESULT hrCo = CoInitializeEx(nullptr, COINIT_MULTITHREADED);
     if (FAILED(hrCo) && hrCo != RPC_E_CHANGED_MODE) {
-        Wh_Log(L"BTBat: CoInitializeEx failed: 0x%08X", hrCo);
+        Wh_Log(L"CoInitializeEx failed: 0x%08X", hrCo);
         return 1;
     }
     HANDLE events[] = { g_shutdownEvent, g_rescanEvent };
@@ -812,7 +815,7 @@ static unsigned int __stdcall ScannerProc(void*) {
                 WaitForSingleObject(g_rescanEvent, 0);
             }
         } catch (const std::bad_alloc&) {
-            Wh_Log(L"BTBat: Scanner allocation failure, retrying");
+            Wh_Log(L"Scanner allocation failure, retrying");
             if (WaitForSingleObject(g_shutdownEvent, g_refreshIntervalMs.load()) == WAIT_OBJECT_0)
                 break;
         }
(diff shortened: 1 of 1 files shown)


BlackPaw21:
3rd time's the charm? 🤞 
read through this one carefully, got some extra notes in there as well.
really do hope this one is good enough. sadly im not sure why my AI has been slipping lately, gonna need to work on that before creating more mods.

m417z:
Same question as in the previous PR: Why is the snippet at the bottom different from the snippet in the wiki?

BlackPaw21:
I manually copied and pasted the snippet directly into the code, really hoped it won't come to it but i guess it's time to ditch my AI agent. sorry for the inconvenience once again.

m417z:
Thanks. Below is the review for the rest of the code.

### Submission review

---

**1. The "No Connected Devices" preset setting is dead, and each Save/Apply leaks an icon.** In `CreateIcons()` you call `LoadSettings()` (which sets `g_hIconDisconnected` from the user's `d_iconDisconnected` preset, [L1139‑1140](https://github.com/ramensoftware/windhawk-mods/blob/31104144f8e563feec3c1ec1bb0740ee43804a01/mods/bt-battery-monitor.wh.cpp#L1139)), then immediately overwrite it with a hard-coded white‑X icon:

```cpp
DestroyIcons();
LoadSettings();                          // sets g_hIconDisconnected from the setting
g_hIconDisconnected = CreateXIcon(16);   // L359: clobbers it (leaks the HICON) + ignores the setting
```

So the "No Connected Devices Preset" row in the dashboard (preset picker + custom `.ico` browse) has no effect — the tray always shows the white X — and the icon `LoadSettings` created is leaked on every init and every WM_RELOAD_ALL (settings apply). Either drop the unconditional overwrite so the preset takes effect, or gate it so the X is only a fallback and nothing leaks:

```cpp
LoadSettings();
if (!g_hIconDisconnected)                 // keep X only as the default
    g_hIconDisconnected = CreateXIcon(16);
```

(If you want the white X to remain the *default*, make `CreateXIcon` the fallback inside `LoadCustomIcon`/`LoadSettings` for that slot rather than forcing it here.)

**2. Left-click doesn't open the device flyout the README promises.** The README says *"Left-click the tray icon to open a neat flyout listing every paired device with its current battery percentage."* But the handler only fires a rescan with no visible UI ([L1027‑1029](https://github.com/ramensoftware/windhawk-mods/blob/31104144f8e563feec3c1ec1bb0740ee43804a01/mods/bt-battery-monitor.wh.cpp#L1027)):

```cpp
case WM_TRAY_CALLBACK:
    if (lParam == WM_LBUTTONUP || lParam == NIN_SELECT) {
        SetEvent(g_rescanEvent);          // no menu shown
    } else if (lParam == WM_CONTEXTMENU) {
        ShowPopupMenu();                  // device list is actually here (right-click)
    }
```

The per-device list lives on the *right*-click menu, so a user who left-clicks expecting the flyout sees nothing. Either wire left-click to `ShowPopupMenu()` (or a dedicated flyout), or update the README to describe the actual behavior (left-click = refresh, right-click = device list + settings).

<p>

- **Load system DLLs from System32.** `LoadLibraryW(L"uxtheme.dll")` ([L1242](https://github.com/ramensoftware/windhawk-mods/blob/31104144f8e563feec3c1ec1bb0740ee43804a01/mods/bt-battery-monitor.wh.cpp#L1242)) and `LoadLibraryW(L"dwmapi.dll")` ([L1399](https://github.com/ramensoftware/windhawk-mods/blob/31104144f8e563feec3c1ec1bb0740ee43804a01/mods/bt-battery-monitor.wh.cpp#L1399)) use the default search order; neither is a KnownDLL. The risk is negligible here since the mod only ever runs inside `windhawk.exe` (a protected install dir), so it's just hardening — prefer `LoadLibraryExW(L"uxtheme.dll", nullptr, LOAD_LIBRARY_SEARCH_SYSTEM32)`. (Both are almost always already loaded, so `GetModuleHandleW` succeeds and this is only the fallback path anyway.)
- **Drop the `BTBat:` prefix from `Wh_Log` calls.** Windhawk already prefixes each log line with the mod name, so the manual prefix is redundant.

</p>
</details>

<p>

- **~1 s latency on refresh.** After a rescan signal (left-click, "Rescan Now", or a `WM_DEVICECHANGE`), `ScannerProc` waits an extra 1000 ms before rescanning ([L810‑813](https://github.com/ramensoftware/windhawk-mods/blob/31104144f8e563feec3c1ec1bb0740ee43804a01/mods/bt-battery-monitor.wh.cpp#L810)). That coalesces device-change bursts nicely, but it also makes the manual "Rescan Now" / left-click feel sluggish — you may want to skip the debounce for the explicit-user-request path.
- **Undocumented uxtheme ordinals** (135/133/136 in `ApplyContextMenuTheme`) for the dark context menu may change across Windows builds; the `GetProcAddress`-and-null-check guards keep it safe, but the menu may silently revert to light theming on a future OS.
- **`HasActiveInterface` re-enumerates all HID + GATT interfaces per BLE device** on every poll (nested `SetupDiGetClassDevs` inside the per-device loop). It's on a background thread every N seconds so it's fine, but you could enumerate the interface lists once per scan and match against them.

</p>
</details>

BlackPaw21:
patched and improved. 
I manually reviewed the code multiple times, let me know if i missed anything, thanks.

m417z:
Why did you add `-ffp-exception-behavior=maytrap` to compiler options?

BlackPaw21:
man gonna be fully honest here, just refusing to pay for Claude and it's biting me back in the behind with heavy hallucinations of my agent. just removed it.

m417z:
### Submission review

---

**1. The "No Connected Devices" preset is still inconsistent — the first Save silently replaces the red X with a ddores icon.** The dashboard preview special-cases the `unknown` key to show the programmatic red X ([L1235‑1236](https://github.com/ramensoftware/windhawk-mods/blob/2dcf7f51b106804b87cb5a14a69d98a8a23a4bfe/mods/bt-battery-monitor.wh.cpp#L1235)), but the load path doesn't: `LoadCustomIcon` maps `unknown` to ddores.dll index 130 ([L206](https://github.com/ramensoftware/windhawk-mods/blob/2dcf7f51b106804b87cb5a14a69d98a8a23a4bfe/mods/bt-battery-monitor.wh.cpp#L206), [L1142‑1145](https://github.com/ramensoftware/windhawk-mods/blob/2dcf7f51b106804b87cb5a14a69d98a8a23a4bfe/mods/bt-battery-monitor.wh.cpp#L1142)). Since Save writes **all six** row keys — including a stored value of `unknown` for the disconnected slot even when the user never touched that row ([L1588‑1595](https://github.com/ramensoftware/windhawk-mods/blob/2dcf7f51b106804b87cb5a14a69d98a8a23a4bfe/mods/bt-battery-monitor.wh.cpp#L1588)) — the very first "Save and Apply" makes the stored value non-empty, so `LoadSettings` takes the `LoadCustomIcon` branch and the tray permanently shows ddores icon 130 instead of the advertised red X. From then on the preview (red X) and the actual tray icon (ddores 130) disagree. Fix: special-case `unknown` in the load path the same way `LoadSlotPreview` does, e.g. in `LoadCustomIcon` (or in the `LoadSettings` disconnected branch):

```cpp
if (wcscmp(iconKey, L"unknown") == 0)
    return CreateXIcon(16);
```

This also fixes the same preview/actual mismatch when the user picks "No Connected Devices Preset" for any *other* row.

**2. The README still promises the removed left-click flyout.** The body text was updated, but three leftovers remain:
- The changelog still says *"Left-click opens a per-device flyout menu listing each device with its current battery level"* ([L72](https://github.com/ramensoftware/windhawk-mods/blob/2dcf7f51b106804b87cb5a14a69d98a8a23a4bfe/mods/bt-battery-monitor.wh.cpp#L72)) — left-click is now rescan-only.
- There is an empty section heading `### Right-click device list` with no content, immediately followed by the next heading ([L39](https://github.com/ramensoftware/windhawk-mods/blob/2dcf7f51b106804b87cb5a14a69d98a8a23a4bfe/mods/bt-battery-monitor.wh.cpp#L39)) — either write the section (the device list *is* on right-click now, so it is worth a sentence) or remove the heading.
- The mod description says *"with per-device flyout menu"* ([L4](https://github.com/ramensoftware/windhawk-mods/blob/2dcf7f51b106804b87cb5a14a69d98a8a23a4bfe/mods/bt-battery-monitor.wh.cpp#L4)) — consider rewording to "right-click menu" so the mod page does not set the wrong expectation.

<p>

Minor polish — none of this affects users much, so it is your call.

- **Reset the tray-icon change cache on `WM_RELOAD_ALL`.** `UpdateTrayIcon` skips `Shell_NotifyIconW` when the icon handle and tip are unchanged ([L839‑840](https://github.com/ramensoftware/windhawk-mods/blob/2dcf7f51b106804b87cb5a14a69d98a8a23a4bfe/mods/bt-battery-monitor.wh.cpp#L839), [L899](https://github.com/ramensoftware/windhawk-mods/blob/2dcf7f51b106804b87cb5a14a69d98a8a23a4bfe/mods/bt-battery-monitor.wh.cpp#L899)). After "Save and Apply", `CreateIcons` destroys and immediately recreates all icons, and Windows frequently recycles handle values — if the new `HICON` happens to get the old handle value and the tip is unchanged, the user's new icon choice will not show until some other state change. Make `lastIcon`/`lastTip` resettable (e.g. file-scope statics cleared in the `WM_RELOAD_ALL` handler) so a reload always pushes an update.
- **Handle `NIM_ADD` failure for the GUID icon.** With `guidItem` identity, if a previous instance exited abnormally without `NIM_DELETE`, the shell may still hold the GUID registration and `NIM_ADD` ([L1787](https://github.com/ramensoftware/windhawk-mods/blob/2dcf7f51b106804b87cb5a14a69d98a8a23a4bfe/mods/bt-battery-monitor.wh.cpp#L1787)) fails, leaving the mod with no icon. The usual pattern: if `NIM_ADD` returns FALSE, issue `NIM_DELETE` for the GUID and retry once.
- **Guard `CB_GETCURSEL` for the two top combos.** The icon combos check `is >= 0` before use, but the Save handler indexes `refreshVals[ri]` / `threshVals[ti]` without checking for `CB_ERR` ([L1578‑1585](https://github.com/ramensoftware/windhawk-mods/blob/2dcf7f51b106804b87cb5a14a69d98a8a23a4bfe/mods/bt-battery-monitor.wh.cpp#L1578)) — a `-1` would read out of bounds. Droplists with an initial selection cannot normally lose it, so this is theoretical, but the guard is one line and makes the code consistent.
- **Dead `break;`** after the `MENU_OPEN_SETTINGS` case block ([L999](https://github.com/ramensoftware/windhawk-mods/blob/2dcf7f51b106804b87cb5a14a69d98a8a23a4bfe/mods/bt-battery-monitor.wh.cpp#L999)) — leftover statement, remove it.
- **Unused `#include <sddl.h>`** ([L105](https://github.com/ramensoftware/windhawk-mods/blob/2dcf7f51b106804b87cb5a14a69d98a8a23a4bfe/mods/bt-battery-monitor.wh.cpp#L105)) — nothing in the file uses SDDL.

</p>
</details>

<p>

- **The media-class battery path does not validate the value.** `GetBatteryFromRegistry` and `EnumerateBthleDevices` both check `battery <= 100`, but `GetBatteryFromMediaClass` assigns the raw byte ([L427](https://github.com/ramensoftware/windhawk-mods/blob/2dcf7f51b106804b87cb5a14a69d98a8a23a4bfe/mods/bt-battery-monitor.wh.cpp#L427)) — a device reporting 255 ("unknown") would show as "255%" in the tooltip and menu. Add the same `battery <= 100` check for consistency.
- **The multi-device tooltip always claims a "lowest" value.** `lowest` starts at 100 and only decreases, so `lowest <= 100` ([L893](https://github.com/ramensoftware/windhawk-mods/blob/2dcf7f51b106804b87cb5a14a69d98a8a23a4bfe/mods/bt-battery-monitor.wh.cpp#L893)) is always true and the `"%d devices"` fallback is dead code. When no connected device reports battery at all, the tip reads "N devices — lowest 100%", which is misleading — track whether any reading exists (e.g. init `lowest = -1`) and use the fallback text when none does.
- **Reopening Mod Settings closes and relaunches the dashboard** ([L987‑997](https://github.com/ramensoftware/windhawk-mods/blob/2dcf7f51b106804b87cb5a14a69d98a8a23a4bfe/mods/bt-battery-monitor.wh.cpp#L987)), discarding any unsaved edits. Bringing the existing window to the foreground (`SetForegroundWindow` on the dashboard window when the GUI thread is running) would match what users expect from a settings window.

</p>
</details>

BlackPaw21:
addressed everything and tested. 🤞

PR #4826, opened 2026-07-18
Add new mod: Faster Notifications Settings Page v1.3

mario0318:
### Faster Notifications Settings Page

Adds a new mod (`notifications-settings-page-speedup`).

The Windows 11 Settings, System, Notifications page rebuilds its full per-app list on every visit, including when you edit one app and press Back. With many notification senders this takes minutes each time. This mod keeps the first open as-is (it fills a cache) and makes later visits load in a couple of seconds, by skipping a per-app summary line and caching each app's settings object across visits. The cached objects read live data, so edits still show correctly. Both behaviours default on and revert instantly when turned off.

**Trade-off:** skipping the per-app summary removes the small subtitle that shows banners/sounds status at a glance. Every app still shows its name, icon, and main on/off toggle, and all per-app options remain editable by opening the app. This can be turned off to keep the subtitle.

**Approach:** hooks a couple of internal functions in `SettingsHandlers_Notifications.dll` by symbol and caches the per-app object at the controller call that produces it. The offsets are build-specific, so it may need a refresh after a major Windows release.

**Testing:** verified on Windows 11. First open runs the normal full scan; repeat visits, including after editing an app, load in a few seconds with edited toggles shown correctly.

## Changelog

* Initial release (v1.0) — new mod.

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/notifications-settings-page-speedup.wh.cpp
+++ b/mods/notifications-settings-page-speedup.wh.cpp
@@ -1,0 +1,267 @@
+// ==WindhawkMod==
+// @id              notifications-settings-page-speedup
+// @name            Faster Notifications Settings Page
+// @description     Stops Settings from re-scanning every notification app on each visit, so revisits load in seconds instead of minutes
+// @version         1.0
+// @author          mario0318
+// @github          https://github.com/mario0318
+// @include         SystemSettings.exe
+// @compilerOptions -lshlwapi
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Faster Notifications Settings Page
+
+**Settings, System, Notifications** rebuilds its full app list every time you
+open the page, including when you edit one app and press Back. With many
+notification senders this takes **minutes, on every visit**.
+
+This mod keeps the first open as-is (it fills a cache) and makes every visit
+after that load in a couple of seconds. It skips a per-app summary line (see the
+trade-off below) and caches each app's settings object across visits so the page
+stops repeating a slow per-app cross-process call. The cached objects read live
+data, so **your edits still show correctly** and nothing goes stale.
+
+Both options default on and can be turned off in the mod's settings, which
+reverts the page to normal instantly. Press **Ctrl+Alt+R** to clear the cache
+and force a full rescan.
+
+## The summary trade-off
+
+Most of the speedup comes from skipping a per-app summary. Normally each app in
+the list shows a small subtitle summarizing part of its setup, such as whether
+banners or sounds are on. With this skipped, every app still shows its **name,
+icon, and main on/off toggle**, and you can still open any app to see and change
+**all** of its options. You only lose the at-a-glance subtitle. If you would
+rather keep it, turn off **Skip per-app summary lookup** below; revisits are
+still cached, just a bit slower.
+
+The first open after a fresh sign-in is still slow; only revisits are fast. The
+mod hooks internal functions in `SettingsHandlers_Notifications.dll` by symbol,
+so it may need an update after a major Windows release. Tested on Windows 11.
+*/
+// ==/WindhawkModReadme==
+
+// ==WindhawkModSettings==
+/*
+- skipSummary: true
+  $name: Skip per-app summary lookup
+  $description: The largest single speedup. Each app still shows its name, icon, and main on/off toggle, but loses the small subtitle summarizing banners/sounds status. You can still open any app to see and change those. Turn off to keep the subtitle (revisits stay cached, just a bit slower).
+- cacheSettings: true
+  $name: Cache app settings across visits
+  $description: Makes revisiting the page fast. Cached objects read live data, so your edits are still reflected. Turn off if the list ever renders incorrectly.
+*/
+// ==/WindhawkModSettings==
+
+#include <windhawk_utils.h>
+#include <shlwapi.h>
+#include <string>
+#include <unordered_map>
+#include <shared_mutex>
+#include <atomic>
+
+// ---------------------------------------------------------------------------
+// Settings
+// ---------------------------------------------------------------------------
+static std::atomic<bool> g_skipSummary{true};
+static std::atomic<bool> g_cacheSettings{true};
+
+// ---------------------------------------------------------------------------
+// Per-AUMID cache of each app's IAumidNotificationSettings object.
+//
+// EnumerateAppSettingItems -> _AddEntry(aumid) per app -> controller
+// ->vtable[0x70](aumid, &settings). The controller is an out-of-process COM
+// object (CoCreateInstanceAsUser / MainNotificationController), so that call is
+// a cross-process activation (~150-300ms each). We vtable-hook that slot: the
+// first visit populates the cache, revisits reuse the live objects and skip the
+// activation. Reads still hit the broker, so edits are reflected (no staleness).
+// ---------------------------------------------------------------------------
+static const size_t CTRL_PTR_OFFSET       = 0x108; // helper -> controller ComPtr
+static const size_t GETSETTINGS_VTBL_SLOT = 14;    // 0x70 / sizeof(void*)
+
+static std::unordered_map<std::wstring, IUnknown*> g_cache; // aumid -> settings obj
+static std::shared_mutex g_cacheMutex;
+static void** g_vtblSlot = nullptr;
+static std::atomic<bool> g_vtblHooked{false};
+static volatile bool g_running = true;
+
+static void FlushCache() {
+    std::unique_lock lock(g_cacheMutex);
+    for (auto& kv : g_cache)
+        if (kv.second) kv.second->Release();
+    g_cache.clear();
+    Wh_Log(L"cache cleared (manual rescan)");
+}
+
+// The controller vtable slot: HRESULT(controller, PCWSTR aumid, IUnknown** out)
+using GetSettings_t = HRESULT (STDMETHODCALLTYPE*)(void*, PCWSTR, void**);
+static GetSettings_t GetSettings_Orig = nullptr;
+
+HRESULT STDMETHODCALLTYPE GetSettings_Hook(void* self, PCWSTR aumid, void** out) {
+    if (g_cacheSettings.load() && aumid && out) {
+        {
+            std::shared_lock lk(g_cacheMutex);
+            auto it = g_cache.find(aumid);
+            if (it != g_cache.end()) {
+                it->second->AddRef();
+                *out = it->second;
+                return S_OK;
+            }
+        }
+        HRESULT hr = GetSettings_Orig(self, aumid, out);
+        if (SUCCEEDED(hr) && *out) {
+            IUnknown* p = reinterpret_cast<IUnknown*>(*out);
+            p->AddRef();
+            std::unique_lock lk(g_cacheMutex);
+            g_cache.emplace(aumid, p);
+        }
+        return hr;
+    }
+    return GetSettings_Orig(self, aumid, out);
+}
+
+static void HookControllerVtable(void* helper) {
+    if (g_vtblHooked.load()) return;
+    void* controller = *reinterpret_cast<void**>(
+        reinterpret_cast<char*>(helper) + CTRL_PTR_OFFSET);
+    if (!controller) return;                 // controller not created yet
+    if (g_vtblHooked.exchange(true)) return;
+    void** vtbl = *reinterpret_cast<void***>(controller);
+    g_vtblSlot = &vtbl[GETSETTINGS_VTBL_SLOT];
+    DWORD oldProtect;
+    if (VirtualProtect(g_vtblSlot, sizeof(void*), PAGE_READWRITE, &oldProtect)) {
+        GetSettings_Orig = reinterpret_cast<GetSettings_t>(*g_vtblSlot);
+        *g_vtblSlot = reinterpret_cast<void*>(GetSettings_Hook);
+        VirtualProtect(g_vtblSlot, sizeof(void*), oldProtect, &oldProtect);
+        Wh_Log(L"settings cache active (controller vtable hooked)");
+    } else {
+        g_vtblHooked = false;
+        Wh_Log(L"VirtualProtect failed; settings cache disabled");
+    }
+}
+
+// ---------------------------------------------------------------------------
+// Symbol hooks in SettingsHandlers_Notifications.dll
+// ---------------------------------------------------------------------------
+
+// NotificationsAppListHelper::_AddEntry(aumid, ...) runs once per app. We use it
+// only as a reliable trigger: after the first entry the controller exists at
+// helper+0x108, and we patch its vtable once (the patch persists for later visits).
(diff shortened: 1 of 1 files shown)


mario0318:
Restored the ## Mod authorship section in the description. That was the validation failure.

mario0318:
Replaced the raw Wh_FindFirstSymbol/NextSymbol/CloseSymbol loop with WindhawkUtils::HookSymbols, using the full decorated names for _AddEntry and PopulateSummaryValues. No behavior change, and it clears the manual-inspection warnings. Passes pr_validation.py locally.

mario0318:
tested on 26100 and it works. hooks resolve, cache kicks in, revisits drop from minutes to a couple seconds. green and mergeable whenever you get to it.

m417z:
### Submission review

Additional manual notes:

* `WatchThread` shouldn't be needed if the `LoadLibraryExW` hook works correctly, remove it.
* For `HotkeyThread`, instead of polling, consider using `RegisterHotKey` which is more efficient and more responsive.
* Replace `VirtualProtect` vtable hooks with `WindhawkUtils::SetFunctionHook`. If not possible, please explain why.
* Remove all `Sleep` calls which shouldn't be needed once all feedback is addressed.

---

**1. Worker threads are not joined before unload → crash window.** `Wh_ModUninit` does `g_running = false; Sleep(120);` and then lets the DLL unmap, but neither `WatchThread` nor `HotkeyThread` is actually joined:

- `WatchThread` sleeps in 200 ms increments (`Sleep(200)` > `Sleep(120)`), so it is very likely still inside its `Sleep(200)` when uninit returns; when it wakes, its return address is in the just-unmapped mod image → crash.
- `HotkeyThread` sleeps `Sleep(500)` right after a `FlushCache()`, so if Ctrl+Alt+R was pressed shortly before unload, it wakes inside unmapped code → crash.

`Sleep(120)` is a guess, not a guarantee. Capture the thread handles and *join* them. The idiomatic form is a manual-reset stop event so the wait is immediate rather than polling:

```cpp
static HANDLE g_stopEvent;   // CreateEventW(nullptr, TRUE, FALSE, nullptr)
static HANDLE g_watchThread, g_hotkeyThread;
// threads wait on g_stopEvent instead of Sleep(...) so they exit at once
void Wh_ModUninit() {
    SetEvent(g_stopEvent);
    HANDLE hs[] = { g_watchThread, g_hotkeyThread };
    WaitForMultipleObjects(2, hs, TRUE, INFINITE);
    CloseHandle(g_watchThread); CloseHandle(g_hotkeyThread); CloseHandle(g_stopEvent);
    ...
}
```

See [taskbar-vd-switcher.wh.cpp](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/taskbar-vd-switcher.wh.cpp#L885-L930) for a stop-event + `WaitForSingleObject` join (and note it also drains in-flight worker threads before freeing state they touch). As a side effect this also fixes the handle leak — the `CreateThread` return values are currently discarded, so the handles are never closed.

**2. `LoadLibraryExW` is hooked on the kernel32 import, so the handler-DLL hook can silently fail to install.** `WindhawkUtils::SetFunctionHook((void*)LoadLibraryExW, ...)` binds kernel32's `LoadLibraryExW` export, but the real implementation is in **kernelbase.dll**, and internal/delay-load paths often call kernelbase's copy directly — those loads won't fire your hook. Today the only backstop is `WatchThread`, which polls for **just 120 s** (`600 * 200 ms`). So if `SettingsHandlers_Notifications.dll` is loaded via a kernelbase-direct path *and* the user opens the Notifications page more than ~2 minutes after the mod is enabled, the mod never installs its hooks and the whole speedup quietly does nothing until the mod is toggled. Resolve the kernelbase address explicitly:

```cpp
HMODULE kernelBase = GetModuleHandleW(L"kernelbase.dll");
auto pLoadLibraryExW = (decltype(&LoadLibraryExW))GetProcAddress(kernelBase, "LoadLibraryExW");
WindhawkUtils::SetFunctionHook(pLoadLibraryExW, LoadLibraryExW_Hook, &LoadLibraryExW_Orig);
```

With a reliable loader hook, the arbitrary 120 s cap on the watcher stops mattering (you can shorten or drop it).

**3. Cached COM objects are released off their creating thread.** The `IAumidNotificationSettings` objects in `g_cache` come out of an out-of-process controller (per your comment), i.e. they're apartment-bound proxies. But `FlushCache()` (called from `HotkeyThread`) and the cleanup loop in `Wh_ModUninit` (arbitrary thread) call `->Release()` on them from a thread that is almost certainly not the one that created them. Cross-apartment release marshals the call into the owning STA — at best it works but adds a round-trip, and if that apartment is busy/blocked it can hang, or the ref just leaks. Please verify the apartment these objects live in; if they're STA proxies, post the releases to the owning thread rather than releasing directly from the hotkey/uninit thread. taskbar-vd-switcher does exactly this for WinRT objects — it runs the releases [on the owning UI thread](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/taskbar-vd-switcher.wh.cpp#L2648-L2656).

<p>

- **`volatile bool g_running` → `std::atomic<bool>`.** You already use `std::atomic` for the other flags; `volatile` doesn't provide the cross-thread visibility guarantees you want here.
- **Ref leak on a concurrent same-AUMID miss.** In `GetSettings_Hook`, two threads that both miss the cache for the same `aumid` will both call the original and both `AddRef`, but only the first `g_cache.emplace(...)` inserts — the second returns `false` and its `AddRef`'d pointer is leaked. Check the `emplace` result and `Release()` on failure. Rare (enumeration is normally single-threaded), so low priority.
- **Consider `WindhawkUtils::HookSymbols`** with a `SYMBOL_HOOK` array instead of the manual `Wh_FindFirstSymbol`/`Wh_FindNextSymbol` enumeration + `wcsstr` substring match. `HookSymbols` has built-in caching and is the idiomatic path; it accepts multiple candidate strings per hook if you need tolerance for signature variations. (If the substring match is a deliberate choice to survive parameter-list changes across builds, feel free to keep it.)

</p>
</details>

<p>

Non-critical observations about the approach itself.

- **Inherent fragility of the offset/vtable approach.** Reading the controller at `helper + 0x108`, the vtable slot `14`, and the guessed `_AddEntry` signature (`unsigned long long opt`) are all build-specific — which you already call out in the README. Two caveats worth noting: (a) if the layout shifts on a future build, `*(helper + 0x108)` can be a non-null garbage pointer that you then dereference as a vtable → a crash in SystemSettings rather than a graceful no-op, so a sanity check on `controller`/its vtable before patching would be safer; (b) the manual vtable patch isn't managed by Windhawk, so unlike your symbol hooks it isn't quiesced on unload — an in-flight `GetSettings_Hook` on the UI thread at the moment of unload can crash when the DLL unmaps. There's no clean fix for the manual-patch case, but it's a real (if low-probability) window; the thread-join fix above at least removes the more likely crash sources.
- **The Ctrl+Alt+R hotkey is global and may be unnecessary.** `GetAsyncKeyState` reads global key state, so the flush fires even when SystemSettings isn't focused (and could collide with another app's Ctrl+Alt+R). Since the cached objects read live data, the only case that actually needs a manual flush is apps being added/removed — consider whether the hotkey earns its keep, or scope it to when the window is focused.

</p>
</details>

mario0318:
went through the review, pushed a fix.

the watcher and hotkey threads wait on a stop event and get joined in Wh_ModUninit now instead of the Sleep(120), handles closed too. hotkey moved off GetAsyncKeyState polling to RegisterHotKey with a message loop that exits on WM_QUIT from uninit. the cached IAumidNotificationSettings objects are STA proxies owned by the settings UI thread, so a flush doesn't release them off-thread anymore, it queues them and the next _AddEntry frees them on that thread. also guarded the emplace so a concurrent same-aumid miss can't leak, and dropped the volatile flag.

two I couldn't take. hooking kernelbase's LoadLibraryExW fastfails SystemSettings on launch here (0xc0000409, CFG) since it's hit constantly during startup, so it's still on the imported LoadLibraryExW with the watcher as a backstop (interruptible + joined now). and the controller hook stays an in-place vtable swap rather than SetFunctionHook because it's only reachable from inside the _AddEntry hook and it's a CFG'd out-of-proc proxy. it does sanity-check the controller, its vtable and the target before patching now.

tested on 26100.8968, disables cleanly with the page open.

m417z:
> * LoadLibraryExW on kernelbase: I tried resolving and hooking kernelbase's `LoadLibraryExW` directly. On 26100.8968 it fast-fails SystemSettings on launch (0xc0000409 / CFG) because that function is called all through startup. I kept the imported `LoadLibraryExW` hook and left the watcher as the backstop you mentioned, now interruptible (waits on the stop event) and joined instead of a blind poll.

Please post the code. Many mods do that and it works.

> * The vtable hook stays an in-place swap rather than `SetFunctionHook`: the controller is only reachable from inside the `_AddEntry` hook and it's an out-of-process COM proxy whose method sits behind CFG, so hooking it there wasn't workable.

Why wasn't it workable? Post the code.

mario0318:
follow-up on the controller hook since the offset + slot is the fragile part. it's an out-of-proc COM proxy, vtable[13/14/15] are just consecutive combase ObjectStublessClient thunks, so [14] isn't a real method, it's the shared stub every proxy uses for that slot. SetFunctionHook there would redirect every COM call in the process, which is why it's a scoped vtable swap.

took the fragility off the +0x108 offset though. the controller's found by QueryInterface on its IID now, with the offset as the fast path and a small guarded scan as a fallback if the layout ever shifts. slot 14 stays but it's interface-defined so it holds. still fine on 26100.8968.

m417z:
I'll take a look. What about the second point (`LoadLibraryExW` and kernelbase.dll)?

mario0318:
so I ended up going a pretty different route here. I did go back and test the kernelbase thing in isolation first like you asked though, just flipped that one line over to hook kernelbase's LoadLibraryExW and left everything else alone, and it straight up fastfails SystemSettings the second it launches, before the notifications page is even in the picture. 0xc0000409, faulting in ntdll. probably some packaged app / CFG thing but I didn't really chase down the why since I found a way around the whole thing anyway.

what I did instead was drop the load-function hook entirely and use LdrRegisterDllNotification, so the loader just tells me whenever the dll loads no matter how it got there, which means I don't really have to care about kernel32 vs kernelbase at all anymore. and since nothing's actually getting detoured there's nothing for CFG to trip on, so that crash just can't happen in the first place. the little watcher I had backing it up wasn't really pulling its weight after that so I dropped it too. the callback stays tiny since it runs under the loader lock, just checks the name and pokes a worker thread that does the actual HookSymbols work once it's off the lock.

it's been solid on 8968, launches fine, hooks go in when you open the page, and it unloads clean even if I disable it with the page still open.

m417z:

mario0318:
went through all of it, pushed an update. the teardown ordering and the proxy release on the ui thread i just did the way you laid out.

on the two you left to me i went the simpler way on both. dropped the fallback scan and kept only the verified offset with the QI check, and pulled the Ctrl+Alt+R hotkey so toggling the cache setting is the rescan now, which got rid of that whole thread too. also swapped shlwapi for an exact compare so that's off the loader lock now.

tested on 26100.8972, disabled it with the notifications page still open and it unloaded clean, no hang, no crash. reloaded after and hooks went back in fine too. lemme know if i missed anything.

mario0318:
the ran flag was me misreading my own helper, it only means the send got issued, so with the window already gone nothing runs and I skip the local restore too. it reports through a param struct now, and the send is bounded since that thread can be suspended.

the restore happens before g_vtblSlot gets dropped now, right in the callback. controller and cached proxies are all abandoned rather than released once the dll goes, flagged atomically instead of reaching for the lock in there.

guard was a no-op, storing the id every call meant whoever's enumerating always matched itself. first one wins now. only thing I added on top is if that thread has exited the next one takes over, otherwise a recreated ui thread bypasses the cache forever. also flipped the loader registration ahead of the module check.

mario0318:
took the pinning suggestion, it holds a ref on the dll while the hooks are in and drops it in uninit. that killed the unload branch and the re-hook path and the whole "server is gone" cache thing along with it, which is a fair bit less code than before.

the controller release on takeover was my fault, I abandoned the cache and then turned around and released the one object from the dead apartment anyway. gets dropped with the rest now. slot goes back first while the ref still holds the page mapped, and there's a guard so a restore that didn't take can't end up with the hook calling itself.

liveness is off a handle now instead of the id, kept the id around for GetWindowOnThread. patch is skipped entirely when caching is off. ran flag isn't on the unloading thread's stack anymore either.

mario0318:
all three ordering ones are fixed.

the latch was publishing the thread id ahead of the handle, so a second thread could read the id, find no handle in place yet, and take the live owner for dead. the handle goes out before the id now, and there's a single claim token so only one thread ever latches or takes over, which also closes the window where two threads could both run the takeover and steal it from each other.

the reference on the handler dll is taken before HookSymbols now and the hook runs through that same handle. you're right that the first-run download can sit there for a while and the page can be opened and closed inside it, the old order left the dll able to unload mid-hook with nothing holding it. taking the ref first also means the module that gets hooked and the module that's held are guaranteed to be the same one rather than two by-name lookups.

and uninit checks OwnerAlive before hopping to the ui thread, since once the real owner has exited its id can come back attached to an unrelated live thread, and releasing the proxies there is exactly the cross apartment release the queue and hop exist to avoid. a dead owner just restores the slot locally and leaves the proxies, same as the existing fallback.

mario0318:
yeah, fixed with the thread_local latch.

m417z:
Why use `LdrRegisterDllNotification` and not hook `LoadLibraryExW`, like many other mods do? This approach seems more complex, more fragile, and less efficient. You spin up an extra thread just for that, and the hooks are placed asynchronously, so function calls can be missed.

If the dll isn't loaded via `LoadLibraryExW` (e.g. it's a dependency of another dll), you can still hook `LoadLibraryExW` and after each successful load, check if your target dll is loaded.

Let me know what you think.

mario0318:
yeah, fair point. we ended up on `LdrRegisterDllNotification` because hooking `LoadLibraryExW` straight off kernelbase fast-failed SystemSettings on launch under CFG (it gets called all through startup). hooking the imported `LoadLibraryExW` avoids that, and the after-load check you described covers the dependency case, so i'm switching to it and dropping the loader callback. the worker thread stays, but it's not for the notification, it's to keep the first-run symbol resolution (which can be slow) off the load path. doing the swap now, will confirm it still catches the page load and push.

m417z:
I asked about it previously.

> Please post the code. Many mods do that and it works.

Hooking the kernelbase `LoadLibraryExW` is a better option since it covers more cases. Why would it crash, and how is it related to CFG?

Can you post the code and reproduction steps?

mario0318:
going to withdraw this one. testing it on my current build, the speedup isn't landing anymore, the settings cache leans on some hardcoded offsets in SettingsHandlers_Notifications.dll and those drifted after a windows update, so it quietly does nothing now. that per-build fragility is really the heart of it, it'd need re-deriving after every feature update, which isn't worth carrying. appreciate you taking a look. may come back to it later with a sturdier approach.

PR #4827, opened 2026-07-19
Add files via upload

adriantips:
## Changelog

* Changelog item 1...
* Changelog item 2...

## Mod authorship

This mod was created by:

- - [x] The submitter, without AI assistance - ✅
- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/acrylic-terminals.wh.cpp
+++ b/mods/acrylic-terminals.wh.cpp
@@ -4,7 +4,7 @@
 // @description     Beautiful transparent acrylic glass effect for CMD, PowerShell, and Windows Terminal
 // @version         1.2
 // @author          adrianzgoated
-// @github          https://github.com/adrianzgoated
+// @github          https://github.com/adriantips
 // @include         conhost.exe
 // @include         WindowsTerminal.exe
 // @compilerOptions -ldwmapi
--- a/mods/acrylic-terminals.wh.cpp
+++ b/mods/acrylic-terminals.wh.cpp
@@ -2,7 +2,7 @@
 // @id              acrylic-terminals
 // @name            Acrylic Terminals (CMD & PowerShell)
 // @description     Beautiful transparent acrylic glass effect for CMD, PowerShell, and Windows Terminal
-// @version         1.0
+// @version         1.2
 // @author          adrianzgoated
 // @github          https://github.com/adrianzgoated
 // @include         conhost.exe
@@ -13,16 +13,16 @@
 
 // ==WindhawkModSettings==
 /*
-- Alpha: 0xB0
+- Alpha: 0x80
   $name: Transparency Level
-  $description: Opacity of the acrylic effect (0x00 = invisible, 0xFF = fully opaque)
-- TintR: 12
+  $description: Opacity of the acrylic effect (0x00 = fully transparent, 0xFF = fully opaque)
+- TintR: 10
   $name: Tint Red
   $description: Red component of the tint color (0-255)
-- TintG: 0
+- TintG: 10
   $name: Tint Green
   $description: Green component of the tint color (0-255)
-- TintB: 30
+- TintB: 25
   $name: Tint Blue
   $description: Blue component of the tint color (0-255)
 - ExtendFrame: TRUE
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

---

The main problem is that the mod, as written, doesn't apply its effect in normal use, and it substantially overlaps an existing mod that already does this correctly.

**The effect is never applied to freshly-launched terminals.** The mod only applies acrylic once, inside `Wh_ModInit`, by enumerating the current process's existing windows (`EnumWindows(EnumApplyProc, myPid)`). But Windhawk runs `Wh_ModInit` *before the target process starts executing* for every newly launched `conhost.exe` / `WindowsTerminal.exe` — at that point the console window doesn't exist yet, so `EnumWindows` finds nothing and the effect is never applied. It only works in the narrow case where you enable (or reload) the mod while a terminal window is *already* open, and even then any terminal you open afterward gets nothing. In practice this makes the mod non-functional for the common case (open a new console → no effect).

Fix: apply the effect when the window actually appears, not once at init. Hook window creation (`NtUserCreateWindowEx` / `CreateWindowExW`) or use `SetWinEventHook(EVENT_OBJECT_SHOW`/`EVENT_OBJECT_CREATE)` and apply from there, keeping the `EnumWindows` pass to catch windows that already exist when the mod loads. The [`translucent-windows`](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/translucent-windows.wh.cpp) mod does exactly this — it enumerates existing windows on load ([around L6016](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/translucent-windows.wh.cpp#L6016)) *and* hooks `NtUserCreateWindowEx` ([L6338](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/translucent-windows.wh.cpp#L6338)) so later windows also get the effect.

**Substantial overlap with `translucent-windows`.** That mod (actively maintained, `@include *`) already applies acrylic/blur/mica to arbitrary windows — including console windows — via the same `SetWindowCompositionAttribute` accent-policy path, with configurable tint/opacity, per-process rules, correct new-window handling, and clean teardown. This submission is a thin, console-scoped subset of it, and as written is less correct. The maintainer strongly prefers extending an existing mod over merging a near-duplicate, since duplicates fragment the catalog. Please either use / contribute a console-preset option to `translucent-windows`, or clearly state what unique capability this mod adds that couldn't live there. If it stays a separate mod, it needs to at least reach functional parity (the window-creation fix above).

**Incomplete reversibility — the forced dark titlebar is never restored.** `ApplyAcrylic` sets `DWMWA_USE_IMMERSIVE_DARK_MODE` (attribute `20`) to `TRUE`, but `RemoveAcrylic` only clears the accent policy and (conditionally) the extended frame — it never reverts the dark titlebar. After the mod is disabled, the forced dark titlebar persists on the live window until it happens to be recreated. A mod's effects must fully disappear when it's disabled, so `RemoveAcrylic` should also set `DWMWA_USE_IMMERSIVE_DARK_MODE` back to `FALSE` (and reset the extended-frame margins unconditionally rather than gated on the current `ExtendFrame` setting, so a setting toggled between apply and remove doesn't leave state behind).

**`@include WindowsTerminal.exe` is likely non-functional / inconsistent with the mod's stated purpose.** Windows Terminal draws its content on an opaque XAML/DirectX surface that covers the whole client area, and it already exposes its own acrylic/opacity via its settings — so a blur-behind applied to its top-level HWND will generally be hidden behind Terminal's own rendering (and may conflict with WT's own composition). The README describes only `conhost.exe`. Please verify the effect is actually visible on `WindowsTerminal.exe`; if it isn't, drop that `@include` and reconcile the README so the advertised targets match what works.

<p>

- **Signed-int shift is technically UB.** In `ApplyAcrylic`, `alpha << 24` with `alpha` an `int` and a default of `0xB0` (176) shifts into the sign bit — left-shifting a signed value past `INT_MAX` is UB. It works in practice, but compute the color in an unsigned type and make `ACCENT_POLICY::GradientColor` a `DWORD`/`unsigned` (and mask `alpha & 0xFF` like the other channels).
- **`AccentFlags = 2` is an unexplained magic number.** Give it a named constant or a short comment describing which borders/behavior it selects.
- **README/`@license` mismatch.** `@license` is `MIT`, but the README footer says "License: Free to use, modify, and distribute." Make the README consistent with the declared MIT license.
- **`EnumApplyProc`/`EnumRemoveProc` touch every visible window of the process.** For `conhost.exe` this is usually just the console window, but it will also apply to any other visible top-level window the process owns. If you keep the enumeration approach, consider narrowing to the actual console/main window.

</p>
</details>

<p>

- **conhost only has a visible window when it's the default terminal host.** On Windows 11 with Windows Terminal set as the default terminal application, launching cmd/PowerShell opens them inside Windows Terminal via ConPTY, and `conhost.exe` runs headless (no visible window) — so the `conhost.exe` path only has a target when the legacy console host is the default terminal. Worth documenting as a limitation so users understand when the effect will and won't appear.

</p>
</details>

Issue #4838, opened 2026-07-19
[Customize Windows notifications placement] Animation direction option does not effect


mguludag:
Hello,
The mod's animation direction options don't change the direction. I put some debug logs and found the problem. It tries to find `RevealGrid2` but on my system I can only find `RevealGrid`. 
I use Win11 LTSC and it might be different than regular versions. 

Mod author: @m417z

m417z:
Fixed in version 1.2.3.

PR #4839, opened 2026-07-19
Enhance FindChildByName to support partial matches

mguludag:
## Changelog

* #4838  Fix animation direction option by partial matching of "RevealGrid" name for **Customize Windows notifications placement** mod


m417z:
Fixed in version 1.2.3.

PR #4843, opened 2026-07-20
Add Tray Privacy Indicator Anchor v2.0

sb4ssman:
Reserves stable system-tray space for the location, microphone, camera and Copilot indicators, so the tray stops shifting every time Windows briefly shows or hides one of its own. The placeholders are dim when idle, bright when in use, and slashed when blocked — and because they are always present, nothing beside them moves.

This replaces the v1.0 that was previously in this PR. Since v1.0 was never published, v2.0 marks the settings contract rather than a release history: the whole settings surface moved to grouped keys (Placement, Content, Layout, Size, Adjust, Surface, Behavior) and a single nestable **Arrangement** expression that supersedes `itemOrder`, the grid-mode family, fixed rows and columns, short row/column placement, and the per-icon nudge settings.

### Review items from the previous round

All five are fixed:

- **Use-after-free** — a `std::wstring_view` was taken over a destroyed `hstring` temporary on the hot scan path. It now takes an owned `std::wstring` copy.
- **`cameraHardwareDetection` defaulted true**, so the mod touched the camera on every Explorer start. It is now opt-in and defaults to false.
- **`[[clang::no_destroy]]`** applied across the XAML globals, with the optional-backed holders `reset()` on the UI thread in `Wh_ModUninit`, and the no-taskbar path deliberately retaining state rather than releasing from Windhawk's unload thread.
- **`SYMBOL_HOOK` array naming** — renamed to name their target modules (`taskbarDllHooks`, `systemTrayModuleHooks` with the three modules in a comment above it).
- **Physical-pixel/DIP mixing** in the row heuristic — row capacity is now computed in DIPs, after reserving vertical padding, via a shared helper that also returns the taskbar's orientation.

### Also in this version

- **Native properties are leased, not guessed.** Everything the mod writes to a native icon is snapshotted before the first write and restored to its exact prior local value on teardown — including "there was no local value here", where the correct restore is `ClearValue` so the native binding resumes. The previous code re-derived what the value *should* have been from the glyph's text, which could not express that case.
- **Stands down on a vertical taskbar.** `taskbar-vertical` applies a `RotateTransform` to the same tray elements this mod writes a `TranslateTransform` to. Rather than fight over one dependency property, the mod detects a vertical taskbar from the tray's own aspect, leaves everything as found, and logs once.
- **Every `$options` value is parsed to an enum at load**, so no downstream code compares a setting against a string literal that a later rename could silently invalidate.
- **The camera hardware monitor says when it is off.** It is opt-in (per the review above) and it is the only path that can detect a physical shutter or kill switch, so both the log and the camera's own tooltip name the setting rather than leaving the capability silently absent.
- Privacy state is driven by registry/device/endpoint notifications rather than a periodic global sweep; only Copilot process activity, a five-minute reconciliation, and backed-off setup retries remain timer-driven.

### Testing

Live-tested on Windows 11: all four indicators idle/active/blocked, the camera hardware shutter via `CameraOcclusionInfo`, microphone endpoint mute versus privacy denial, evidence tooltips and click-through to the matching `ms-settings:` pages, automatic and hand-written arrangements, tray placements and the Start-adjacent positions, Explorer restart, and disable → native restore → re-enable.

## Changelog

* Replaced the settings surface with grouped keys and one nestable `Arrangement` expression; `itemOrder`, grid modes, smart layout, fixed rows/columns, short row/column placement, and the per-icon nudge settings are gone
* Fixed a use-after-free on the scan path (`wstring_view` over a destroyed `hstring`)
* Made camera hardware detection opt-in, defaulting to off
* Applied `[[clang::no_destroy]]` to the XAML globals with a UI-thread reset
* Renamed the symbol hook arrays to name their target modules
* Fixed physical-pixel/DIP mixing in the row-capacity heuristic
* Added an exact property-lease restore for every native property written
* Parsed every `$options` value to an enum at load

## Mod authorship

This mod was created by:

- - [x] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/tray-privacy-indicator-anchor.wh.cpp
+++ b/mods/tray-privacy-indicator-anchor.wh.cpp
@@ -1,0 +1,5711 @@
+// ==WindhawkMod==
+// @id              tray-privacy-indicator-anchor
+// @name            Tray Privacy Indicator Anchor
+// @description     Permanently shows location/microphone/camera/Copilot icons in the system tray — dim when idle, bright when in use — preventing taskbar layout shifts.
+// @version         2.0
+// @author          sb4ssman
+// @github          https://github.com/sb4ssman
+// @include         explorer.exe
+// @architecture    x86-64
+// @compilerOptions -lole32 -loleaut32 -lruntimeobject -lversion -lsetupapi -lcfgmgr32 -lshell32
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Tray Privacy Indicator Anchor
+
+A Windhawk mod for Windows 11 that reserves stable tray space for privacy and
+status indicators. Location, microphone, camera, and Copilot placeholders stay
+visible in the system tray: dim when idle, bright when active.
+
+The goal is to stop taskbar tray sections from shifting when Windows briefly
+shows or hides privacy indicators, especially when Windows Web Experience Pack
+or Widgets frequently access location.
+
+## Gallery
+
+Idle location and microphone placeholders reserve their tray space without
+demanding attention:
+
+![Idle location and microphone placeholders](https://raw.githubusercontent.com/sb4ssman/Windhawk-Mod-Lab/main/privacy-indicator-anchor/assets/location-mic-availble-not-in-use.png)
+
+All four unavailable indicators in a single row:
+
+![All four privacy indicators unavailable in one row](https://raw.githubusercontent.com/sb4ssman/Windhawk-Mod-Lab/main/privacy-indicator-anchor/assets/all-4-disabled.png)
+
+The same four indicators in a compact block, which is what `auto` picks when
+the taskbar is tall enough for two rows:
+
+![All four unavailable indicators in a compact grid](https://raw.githubusercontent.com/sb4ssman/Windhawk-Mod-Lab/main/privacy-indicator-anchor/assets/all-4-disabled-grid.png)
+
+Activity highlighted in red — the microphone in use, and the slashed camera
+reporting attempted use while its hardware switch blocks it:
+
+![Active microphone in red beside a red slashed camera blocked by its hardware switch](https://raw.githubusercontent.com/sb4ssman/Windhawk-Mod-Lab/main/privacy-indicator-anchor/assets/camera-mic-in-use-highlighted.png)
+
+The active glow treatment provides a more emphatic alternative:
+
+![Active microphone and camera with glow](https://raw.githubusercontent.com/sb4ssman/Windhawk-Mod-Lab/main/privacy-indicator-anchor/assets/cam-mic-in-use-highlight-glow.png)
+
+Active or requested pathways can remain conspicuous even while blocked:
+
+![Requested or active privacy pathways shown while blocked](https://raw.githubusercontent.com/sb4ssman/Windhawk-Mod-Lab/main/privacy-indicator-anchor/assets/requested-or-active-pathways-disabled.png)
+
+The location tooltip explains the access-denied reason and opens the matching
+Windows privacy page when clicked:
+
+![Location evidence tooltip and Windows Location settings](https://raw.githubusercontent.com/sb4ssman/Windhawk-Mod-Lab/main/privacy-indicator-anchor/assets/location-tooltip-and-win-settings.png)
+
+Microphone evidence distinguishes an endpoint mute from privacy denial:
+
+![Microphone endpoint-mute evidence tooltip](https://raw.githubusercontent.com/sb4ssman/Windhawk-Mod-Lab/main/privacy-indicator-anchor/assets/mic-tooltip.png)
+
+Supported camera drivers report their hardware privacy-control evidence:
+
+![Camera hardware privacy-control evidence tooltip](https://raw.githubusercontent.com/sb4ssman/Windhawk-Mod-Lab/main/privacy-indicator-anchor/assets/camera-tooltip.png)
+
+Copilot reports its installation state and links to the relevant settings:
+
+![Copilot installation-state tooltip](https://raw.githubusercontent.com/sb4ssman/Windhawk-Mod-Lab/main/privacy-indicator-anchor/assets/copilot-tooltip.png)
+
+## Features
+
+- Persistent placeholder icons for location, microphone, camera, and Copilot
+- Idle opacity setting so inactive icons can be subtle but still reserve space
+- One nestable **Arrangement** expression places the icons in any shape —
+  fitted to your taskbar height automatically, or written out by hand
+- Turn any of the four icons off individually
+- Tray placement before icons, before OmniButton, before clock, after clock, or after Show Desktop
+- Experimental placement immediately left or right of Start
+- Per-icon and per-group pixel nudges inside the arrangement expression, plus a
+  whole-group offset
+- Independent idle, active, disabled, glow, and slash colors
+- Steady, breathing, or radiating active emphasis with reach/speed controls
+- Disabled slash overlays for blocked or unavailable privacy devices
+- Hardware camera shutter/kill-switch detection on supported Windows 11 camera drivers
+- Evidence-specific tooltips instead of a generic "hardware disabled" label
+- Click-through to the relevant Windows privacy, input, camera, taskbar, or app settings
+- Optional testing toggle to let Windows' native privacy indicators appear
+
+## Why this starts at 2.0
+
+Version 1.0 was never published — it existed only as a pull request. The 2.0 in
+the version field marks the settings contract, not a history of releases: every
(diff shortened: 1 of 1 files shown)


m417z:
In addition to the notes below, improve the `[[clang::no_destroy]]` usage as discussed in the other PRs.

Also `WindhawkUtils::SYMBOL_HOOK systemTrayDllHooks` applies for 3 modules, rename the variable and list the modules in a comment.

### Submission review

---

**1. Dangling `std::wstring_view` over an `hstring` temporary (undefined behavior).** In `ApplyPrivacyIndicatorBehavior` and its text-changed callback:

```cpp
std::wstring_view text = tb.Text();          // line 3830
...
std::wstring_view newText = tbRef.Text();    // line 3854
```

`TextBlock::Text()` returns a `winrt::hstring` *by value*. The temporary is destroyed at the end of the statement, freeing the `HSTRING` buffer, and `text`/`newText` are `wstring_view`s that keep pointing at that freed buffer. Every subsequent read (`text.empty()`, `text[0]`, `DetectPrivacyType(text)`, `IsPrivacyText(text)`) is a use-after-free. This runs on the hot native-indicator scan path (every icon load and every text change), so it's not a corner case — it only "works" today because the freed buffer usually still holds intact bytes. Every other taskbar mod keeps the `hstring` alive, e.g. [taskbar-tray-system-icon-tweaks.wh.cpp#L496](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/taskbar-tray-system-icon-tweaks.wh.cpp#L496) uses `auto text = innerTextBlock.Text();`. Fix: take a `std::wstring` copy (you already do this for names via `std::wstring(fe.Name())`) or keep an `hstring`/`auto` and index it directly.

**2. `cameraHardwareDetection` defaults to `true`, silently initializing the camera on every Explorer start.** With the default on, the worker thread calls `MediaCapture::InitializeAsync()` on the system's default camera at every load/restart, for every user who installs the mod. Your own README warns this "activates its LED/indicator" on some devices — so by default a "reserve tray space" mod can make the camera light blink on Explorer restart, sign-in, etc., which users won't connect to this mod. This is exactly the kind of hardware side effect that should be opt-in. Recommend defaulting it to `false` (the feature is already fully documented as experimental, so opt-in fits). It's your call, but a surprising camera-touching default is a real user-facing concern.

<p>

- **`MicPrivacyMonitor::Cleanup()` ordering + unsynchronized endpoint pointers.** `IMMNotificationClient` callbacks are delivered on an audio-system thread, and they call `AttachDefaultEndpoint()`, which mutates `m_device`/`m_volume`. `Cleanup()` runs on the worker thread and calls `DetachEndpointVolume()` *before* `UnregisterEndpointNotificationCallback(this)` — so a notification arriving in that window re-attaches `m_device`/`m_volume` that then leak, and the two threads touch those raw pointers without a lock. It's a narrow unload/device-churn race (the notification callbacks are effectively serialized against each other), but unregister-then-detach, plus a small lock or atomic swap around the endpoint pointers, would close it.

- **`GetStringSetting` could be `WindhawkUtils::StringSetting`.** The raw `Wh_GetStringSetting` + `Wh_FreeStringSetting` helper reinvents the RAII wrapper; `WindhawkUtils::StringSetting` (from `windhawk_utils.h`) does the same with automatic cleanup.

- **Copilot activity polling enumerates every process once a minute.** `CheckCopilotActive()` takes a full `CreateToolhelp32Snapshot(TH32CS_SNAPPROCESS)` every 60s just to look for four exe names. It's cheap enough, but it's the only real polling left in an otherwise event-driven design — worth a comment noting there's no event source, or a lower cadence.

</p>
</details>

<p>

- **Smart-grid row capacity mixes physical pixels with DIPs.** In `InjectSyntheticIcons`, `taskbarH` comes from `GetWindowRect` (physical pixels), but `iconSize`/`buttonSpacing` (and therefore `pitch`) are XAML DIPs, and `availableRows = (taskbarH + spacing) / pitch` divides one by the other. On any non-100% display scaling — i.e. most laptops at 125–150% — the computed row count won't match the intent ("single row on standard taskbars"). Worth verifying the auto layout on a scaled display; the clean fix is to derive the height in DIPs (e.g. from the taskbar XAML element's `ActualHeight`, or divide the window rect by the monitor DPI scale) before dividing by the DIP pitch.

- **Only the primary taskbar gets the anchor bar.** `FindCurrentProcessTaskbarWnd` matches `Shell_TrayWnd` only, so secondary-monitor taskbars are untouched. That's consistent with where the system tray lives and the README says "primary taskbar," so just confirming it's intentional.

</p>
</details>

sb4ssman:

sb4ssman:
Thanks — both blocking items were real, and item 1 was the worst bug this mod could have. Addressed in this push:

**1. Hidden with nothing in its place.** Suppression is now a single decision, `ApplyNativeSuppression`, made on every path (initial track, glyph change, visibility change). A native indicator is collapsed only when the anchor bar is live **and** has a slot for that indicator's type (`Both` needs the location and microphone slots). Otherwise anything the mod wrote on it is handed back through the lease. That covers all three routes you listed:
- The `Loaded` path only tracks icons when `g_syntheticGrid` exists after the injection attempt, so a failed, stood-down or empty injection leaves the icon alone.
- With all four `Content` toggles off, `ApplyStyle` returns before `ScanMainStack`.
- With one toggle off, that device's native glyph is tracked but never collapsed. The README now says so.

As suggested, the type is no longer guessed. An empty glyph is tracked with `typeKnown = false`: it sets no active flag and is never suppressed, and the decision is made in the text callback once the glyph arrives. `UntrackPrivacyElement` only clears a flag that a known type set. That also resolves the "guessed `Location`" functionality note. An already-tracked icon is re-evaluated when `ApplyPrivacyIndicatorBehavior` sees it again, so an icon tracked before the bar appeared is suppressed once the bar exists.

One detail worth noting: restoring writes `Visibility`, which raises that icon's visibility callback synchronously while `RestoreObject` is still iterating its snapshots. A re-entrancy guard makes the nested call a no-op.

**2. Vertical stand-down bypassed.** The check is now `TaskbarRequiresStandDown()`, called at the top of `InjectSyntheticIcons`, so every injection path honours it. `ApplyOnTaskbarThread` clears `g_taskbarRestarted` before standing down and calls `RemoveModUi()` if anything is already injected. That makes the stand-down complete in both directions (your functionality note).

**Optional items taken:**
- Dropped `NonActivatableStack` from the host-stack filter. For the record, it was added last round on the previous review's advice, but you're right that the scan only walks `MainStack`, and the two should agree.
- Removed the unleased `iconView.IsHitTestVisible(true)`.
- `g_uiHostWnd` is now set where injection succeeds, in `InjectSyntheticIcons`.
- Removed the redundant `catch { Release; throw; }` around `PlaceChild`.
- The glow tree is only built when `Surface.GlowEnabled` is on.
- `g_taskbarWnd` and `g_uiHostWnd` are `std::atomic<HWND>`, loaded once per use. The state worker's double read, a null check followed by a use, could see a rebuild in between.
- `g_settings` is loaded into a copy on Windhawk's thread and published on the taskbar thread. If that dispatch never runs, it is published directly, so a change is never lost.
- `g_startLease` is a `std::optional` reset on the UI thread, like its neighbours.
- AppModel-driven Copilot re-checks are coalesced to one per 2 s.
- The unused template surface is gone. The mod is now assembled from ten small components (6383 → ~6040 lines), with its own focused tree walk instead of the shared one. `tbh::RetryLoop`, `sio::StringSetting`, `vtw::*`, `ple::Lease::{Abandon,Count,Empty}` and `start_placement::Side::Over` no longer exist in the file.

**Not changed, with reasons:**
- **`-loleaut32` is required.** `LogCurrentUiException` calls `winrt::hresult_error::message()`, and C++/WinRT stores that message in a `BSTR`. Removing the flag fails the link with undefined `SysStringLen` and `SysFreeString`. We tried it on 2026-09-19.
- **`OpenSettingsForItem` stays on the UI thread.** Moving `ShellExecuteW` to a worker would need a thread that `Wh_ModUninit` joins, or the launch could still be running in this image after unload. It runs only on a user's tap and hands an `ms-settings:` URI to the protocol broker, so we judged the extra lifecycle surface not worth it.
- **`auto` rows on a 48 px taskbar:** deliberate, and the README was what was wrong. Two 16 px rows do fit a standard taskbar, and the live-tested gallery shot of the 2×2 block was taken on one. The README now says that `auto` gives 2×2 on a standard taskbar and shows the single-row expression beside the single-row screenshot.
- **Overlap with tray-utility-customizer:** each mod leases its own named tray slot and restores only that slot on unload. They group different things (utilities vs privacy indicators), so a user who wants both gets two deliberate slots rather than a conflict.

PR #4844, opened 2026-07-20
Update Taskbar Virtual Desktop Switcher to v2.0

### LINKED ISSUE #4830: [Taskbar Virtual Desktop Switcher] Add custom indicator
Mod author: @sb4ssman 

We would love a custom indicator that works like the already implemented toggle dots indicator, so users can define their own symbols/letters. Example: 🟢🔴🔴.

Image

Btw thanks for the recent update!

sb4ssman:
Updates `taskbar-vd-switcher` from v1.7 to v1.8 and resolves #4830.

The existing indicator mode is now fully customizable: users can choose independent active and inactive symbols or text (for example, `🟢` and `🔴`) while preserving `●` and `○` as the defaults. Desktop indicators and the optional Task View button also have independent font-family settings.

Desktop entries are now native XAML `ToggleButton` controls. The active desktop owns the real `Checked`, `CheckedPointerOver`, and `CheckedPressed` states, allowing Windows 11 Taskbar Styler rules to target the active state directly instead of inferring it from a color.

Default tray placement

Four desktops with the Task View button

## Tested

* Live-tested in Explorer: custom active/inactive indicators, independent fonts, checked-state styling, desktop switching, settings reapply, and teardown behave as intended
* Windhawk Clang syntax compilation passes
* Exit-time-destructor audit passes
* Settings YAML parsing and embedded/folder README parity pass
* Current upstream PR validator passes without warnings

## Changelog

* Added customizable active and inactive indicator symbols/text to the existing indicator mode, with `●` and `○` retained as defaults
* Added independent font-family settings for desktop indicators and the Task View button
* Replaced desktop buttons with `ToggleButton` controls whose checked state follows the current virtual desktop; exposed native `Checked`, `CheckedPointerOver`, and `CheckedPressed` states for Taskbar Styler
* Kept click feedback synchronized with the actual desktop while switching asynchronously
* Hardened XAML lifecycle handling: button delegates, tooltips, and boxed content are cleared before grid removal; process-shutdown XAML owners use intentional lifetime retention; UI callbacks contain exceptions
* Expanded the readme with a copy-ready `🟢`/`🔴` preset, the new settings, and the exact Taskbar Styler target

## Mod authorship

This mod was created by:

- - [x] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/taskbar-vd-switcher.wh.cpp
+++ b/mods/taskbar-vd-switcher.wh.cpp
@@ -77,8 +77,8 @@ A [Windhawk](https://windhawk.net) mod for Windows 11 that injects clickable but
 | Button spacing | 2 px | Gap between buttons in the grid |
 | Label format | Numbers | Numbers · Roman numerals · Indicator symbols · Custom labels |
 | Custom labels | *(empty)* | Comma-separated, e.g. `H,W,M` |
-| Active indicator | ● | Symbol shown for the current desktop in Indicator symbols mode |
-| Inactive indicator | ○ | Symbol shown for other desktops in Indicator symbols mode |
+| Active indicator | ● | Current desktop's symbol in Indicator symbols mode; e.g. 🟢 for a red/green light look |
+| Inactive indicator | ○ | Other desktops' symbol in Indicator symbols mode; e.g. 🔴 |
 | Font size | 10 pt | Button label size |
 | Indicator font family | *(native)* | Font family for desktop labels and indicator symbols |
 | Active text color | *(native)* | Current desktop's label color |
@@ -103,10 +103,6 @@ A [Windhawk](https://windhawk.net) mod for Windows 11 that injects clickable but
 | Task View button position | After | Column before/after desktop buttons, or sliver row above/below |
 | Task View button sliver height | 6 px | Height of the Task View button when used as a sliver row |
 | Task View button column width | 14 px | Width of the Task View button when used as a side column |
-
-For a copy-ready colored indicator preset, choose **Indicator symbols** and set
-**Active indicator symbol** to `🟢` and **Inactive indicator symbol** to `🔴`.
-Swap either emoji for any symbol, letter, or text you prefer.
 
 All color settings accept `#RRGGBB` or `#AARRGGBB` hex (the alpha byte is
 honored), the generics `accent`, `accentLight`, and `accentDark` for the
@@ -248,11 +244,15 @@ This mod builds directly on patterns established by several community mods:
 
 - activeIndicator: "●"
   $name: Active indicator symbol
-  $description: Symbol or text shown for the current desktop when Label format is Indicator symbols.
+  $description: >-
+    Symbol or text shown for the current desktop when Label format is Indicator
+    symbols. For a red/green light look, paste 🟢 here and 🔴 below.
 
 - inactiveIndicator: "○"
   $name: Inactive indicator symbol
-  $description: Symbol or text shown for other desktops when Label format is Indicator symbols.
+  $description: >-
+    Symbol or text shown for other desktops when Label format is Indicator
+    symbols. For a red/green light look, paste 🔴 here and 🟢 above.
 
 - fontSize: 10
   $name: Font size (pt)
--- a/mods/taskbar-vd-switcher.wh.cpp
+++ b/mods/taskbar-vd-switcher.wh.cpp
@@ -477,7 +477,10 @@ struct ModSettings {
     int          masterButtonSpacing  = 0;
     int          gridVerticalOffset   = 0;
 };
-[[clang::no_destroy]] ModSettings g_settings;
+// ModSettings holds only std::wstring/int/bool, so its destructor is safe to
+// run at process shutdown; no [[clang::no_destroy]] needed, and adding it would
+// leak the string buffers on every normal unload.
+ModSettings g_settings;
 
 static void LoadSettings() {
     auto Str = [](const wchar_t* k) {
--- a/mods/taskbar-vd-switcher.wh.cpp
+++ b/mods/taskbar-vd-switcher.wh.cpp
@@ -2,7 +2,7 @@
 // @id              taskbar-vd-switcher
 // @name            Taskbar Virtual Desktop Switcher
 // @description     Injects clickable buttons into the taskbar — one per virtual desktop — with configurable grid arrangement for direct switching.
-// @version         1.7
+// @version         1.8
 // @author          sb4ssman
 // @github          https://github.com/sb4ssman
 // @include         explorer.exe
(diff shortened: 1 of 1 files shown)


m417z:
A question about `[[clang::no_destroy]]`: did you add it, or did Claude add it? Using it is tricky, because using it too much may cause memory leaks, but not using it enough may cause hangs and crashes when the process terminates.

For example, using it with `std::vector` will leak memory because even if the vector is freed, the memory isn't. A solution that works is using `[[clang::no_destroy]] std::optional<std::vector<ButtonEventState>>`, but it makes the usage a bit less convenient.

I'm still exploring how to make it more convenient. Perhaps I should include `WindhawkUtils::NoDestroy` that's similar but with a nicer syntax.

### Submission review

---

- **`[[clang::no_destroy]]` on `g_settings` is unnecessary and leaks its strings on every unload.** `ModSettings` holds only `std::wstring`/`int`/`bool` members — its destructor just frees heap, which is safe to run at process shutdown (the OS reclaims it anyway), so it doesn't need the attribute. Suppressing the destructor instead means the ~30 `std::wstring` buffers are leaked on every *normal* unload (disable/reload), where `~ModSettings` would otherwise have freed them cleanly. Keep the attribute on the WinRT-holding globals (`g_buttonGrid`, `g_injectionParent`, `g_buttonEventStates`, `g_secondaryBars`, etc. — those genuinely need it), and drop it from `g_settings`:
  ```cpp
  ModSettings g_settings;   // no [[clang::no_destroy]] needed
  ```

sb4ssman:
Claude added it, not me — and it added it too broadly. These mods are written
with Claude assistance, and after an earlier review raised exit-time destructor
safety, it applied `[[clang::no_destroy]]` to every namespace-scope global with
a non-trivial destructor rather than only to the ones holding WinRT references.
`g_settings` was collateral from that sweep. Your explanation of the tradeoff is
what made the distinction concrete, so thank you for asking rather than just
flagging it.

The required fix is already in the current head (`c5995bd3`): `g_settings` is a
plain `ModSettings g_settings;` again, with a comment recording *why* it must
not be annotated so the sweep doesn't come back.

On the containers: `g_autoRevokerList`, `g_buttonEventStates`, and
`g_secondaryBars` here are still the bare form that keeps its heap buffer alive
after every unload. I've adopted the
`[[clang::no_destroy]] std::optional<std::vector<...>>` + `reset()` shape you
described in #4485 and #4841, and this mod should follow. I'd rather not push
that change here on top of a green PR without live-testing it first, so I'll
fold it into the next tested update rather than sneak it in now.

`WindhawkUtils::NoDestroy` would help. The failure mode I keep hitting is that
the correct choice depends on what the global *owns* — heap-only, a nullable
WinRT handle, or a container of WinRT references — and the attribute alone
doesn't express that. A wrapper whose reset/release semantics are part of its
type would make the right thing the easy thing.

m417z:
> `WindhawkUtils::NoDestroy` would help

I thought about it, but it wouldn't be different from `[[clang::no_destroy]]` or `[[clang::no_destroy]] std::optional<...>`. It's not difficult because of the syntax, it's difficult because it's truly tricky. Some destructors must be skipped on termination, but not all. Some objects must be freed on the correct thread, and for others it doesn't matter. Let me know if you have a better solution than the current one - using `[[clang::no_destroy]]` with `std::optional`.

m417z:
> I'd rather not push
>  that change here on top of a green PR without live-testing it first, so I'll
>  fold it into the next tested update rather than sneak it in now.

To clarify: are you going to update the PR, or are you saying that you prefer getting it merged as is?

sb4ssman:
I am updating it, we can hold off on the merge. 

And I see the other comments. I am working them too.

sb4ssman:

sb4ssman:
Thanks for the careful trace — both blocking items were real, and item 2 is why we removed the bridge rather than patching it.

**1 + 2. The 1.7 bridge is removed.** `HasLegacySettings`, `PreferCurrentOrLegacy*` and `Behavior.UseLegacySettings` are gone. Your item 2 is the root problem, and neither of the two fixes you offered removes it:
- Windhawk has no API to write a setting. `Wh_SetIntValue`/`Wh_SetStringValue` are mod storage and invisible to the settings page. So any bridge, including "until first save" or a storage snapshot, leaves a stretch where the settings page shows one value and the bar uses another.
- The "default means unset" rule makes a carried value impossible to set back to its default (your `beforeClock` → `afterClock` example).
- The mapping gaps in item 1 (`"dot"`, `gridVerticalOffset` → `PadY`, the Task View keys) showed that a partial bridge produces *wrong* results. A wrong migration is worse for a user than an obvious one-time reset.

The README's "Upgrading from 1.x" section, which you noted contradicted the bridge, is now accurate again. It says settings reset once on update, and now also tells users to copy their settings in Textual mode first. This also makes your "please test Save with undeclared keys" question moot.

**Optional items taken:**
- **Dead template code:** the mod is now assembled from eight small components, with its own focused tree walk (5,432 → 5,119 lines). Gone from the file: `property_lease`, `ContentAlong`/`AvailableRows`/`PixelsToDip`/`ResolveArrangement`/uncached `Measure`/`Arrange`, `vtw::*`, `sio::LoadString`/`sio::StringSetting` (now `WindhawkUtils::StringSetting`), `GetWindowsAccentBrush` and `VdPositionName`, along with the commentary about other mods and lab files.
- **Retry:** it now **uses** `RetryLoop`; the hand-rolled `CreateThread` worker is gone.
- **Deferred settings reapply freezing the bar:** fixed as you suggested. A reapply that could not run sets `g_reapplyPending`. While it is set, each retry attempt removes and rebuilds the old bar rather than treating "a grid exists" as done. The notification thread is restarted meanwhile, so the old bar keeps following desktop switches.
- **`RunFromWindowThread` running a callback once per concurrent hook:** fixed in the shared dispatch component with a `ran` flag in `Dispatch`, so every mod built from it gets the fix.
- **`!tornDown` branch:** the code now matches the rationale. Handler records hold strong XAML references, so they are retained with the other owners rather than cleared from Windhawk's thread. The preview window class is unregistered from the current thread when the popup's own thread can't be reached (it fails harmlessly if a window still exists).
- **`s_reapplied`:** replaced by a small struct passed through the `void*`. `InvalidateButtonGridSizeEstimate()` now runs on the UI thread, next to `RemoveButtonGrid()`.
- **Small stuff:** duplicate includes and `<thread>` removed; `<dwmapi.h>`/`<shobjidl.h>` moved to the top; the "NEW in v2.1" comment rewritten; the init log uses `WH_MOD_VERSION`. The embedded README's lab-relative link now points at the repository, and our README sync tooling does that for every lab link from now on.

**Functionality notes:**
- **Previews on secondary taskbars:** fixed. The hosting taskbar is now whichever of this process's taskbars contains the cursor (the hover that scheduled the preview is over the button), instead of a XamlRoot match that only works on the primary taskbar.
- **Untitled windows** are now drawn and counted, just without a caption. **Past the 48-window cap**, the extra windows are still counted in the footer. **`DWM_TNP_SOURCECLIENTAREAONLY`** is no longer set: `rcSource` is computed from the window rect, and the frame-included default is what we want.
- **Left as is:** the per-hover `CoCreateInstance` and registry reads. They run once per hover after the 400 ms delay, and caching a COM object in a namespace-scope holder would add a thread-affine release to unload. We also kept the 100 ms cursor poll while the popup is up. `PointerExited` covers the normal exit, but the poll is the backstop when the pointer leaves through the popup, which is `HTTRANSPARENT`, or when the button is rebuilt underneath it.

**Not changed, with reasons:**
- **`-loleaut32` is required.** `LogCurrentUiException` calls `winrt::hresult_error::message()`, which C++/WinRT stores in a `BSTR`. Removing the flag fails the link on `SysStringLen`/`SysFreeString`.

PR #4853, opened 2026-07-21
Update for the mod

babamohammed2022:
## Changelog

- Added the new Auto/Light/Dark theme setting with live updates
- Adjusted the flyout size to match Windows 7/8.1 original
- Made the window appearance cleaner
- Added the Battery warning notification (if it is below 20%)
- Added the pending Windows Update reboot notification
- Added the BitLocker encryption status notification
- Added Portuguese and German languages
- Added automatic language detection
- Added targeted MSDT diagnostics (network, security, Windows Update, battery) for relevant problem types, falling back to the generic troubleshooter when unavailable
- Added partial multi-monitor support
- Various stability improvements

## Mod authorship

This mod was created by:

- - [X] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/win7-action-center-recreation.wh.cpp
+++ b/mods/win7-action-center-recreation.wh.cpp
@@ -77,7 +77,7 @@ The mod has been tested on Windows 10 1809, Windows 10 21H2 and Windows 11 23H2 
 /*
 - useRoundedCorners: true
   $name: Rounded corners
-  $description: Soft rounded edges on the flyout (Windows 7 look). Turn this off for Classic theme or other styles that need square corners.
+  $description: This setting enables rounded edges on the flyout (Windows 7 look). Turn this off for Classic theme or other styles that need square corners.
 - refreshInterval: 5000
   $name: Status check interval (ms)
   $description: How often the tray icon re-checks security and maintenance (milliseconds). Use at least 1000. Set 0 to check only when Windows reports a change.
@@ -89,7 +89,7 @@ The mod has been tested on Windows 10 1809, Windows 10 21H2 and Windows 11 23H2 
   $description: When hotkeys are enabled, Ctrl+N shows a sample balloon and Ctrl+Shift+N clears it. Useful only for testing.
 - privacyMode: false
   $name: Privacy mode
-  $description: Always show a green "all good" tray icon and hide problems in the flyout. Handy on a shared screen.
+  $description: Always show the neutral tray icon and hide problems in the flyout. Handy on a shared screen.
 - language: auto
   $name: Language
   $description: Language for the tray icon, flyout, and balloons. "Auto" follows Windows.
@@ -1101,6 +1101,75 @@ enum ProblemType {
 };
 
 // ============================================================================
+// MSDT Diagnostic Packs (with fallback to generic troubleshooter)
+// ============================================================================
+
+// Verifica se un diagnostic pack MSDT esiste sul sistema corrente prima di
+// invocarlo. I pack vivono in %SystemRoot%\diagnostics\system\<PackId>.
+// Se il pack non c'e' (rimosso in build piu' recenti, o SKU diverso),
+// OpenMsdtDiagnostic ricade sullo shell folder generico dei troubleshooter,
+// esattamente come fa gia' OpenProblemAction per i problemi non mappati.
+static BOOL IsMsdtPackAvailable(const WCHAR* packId) {
+    WCHAR path[MAX_PATH];
+    WCHAR sysRoot[MAX_PATH];
+    if (!GetEnvironmentVariableW(L"SystemRoot", sysRoot, MAX_PATH)) {
+        StringCchCopyW(sysRoot, MAX_PATH, L"C:\\Windows");
+    }
+    StringCchPrintfW(path, MAX_PATH, L"%s\\diagnostics\\system\\%s", sysRoot, packId);
+    DWORD attrs = GetFileAttributesW(path);
+    return (attrs != INVALID_FILE_ATTRIBUTES) && (attrs & FILE_ATTRIBUTE_DIRECTORY);
+}
+
+static void OpenGenericTroubleshooter(void) {
+    SHELLEXECUTEINFOW sei = { sizeof(sei) };
+    sei.fMask = SEE_MASK_FLAG_NO_UI | SEE_MASK_INVOKEIDLIST;
+    sei.lpVerb = L"open";
+    sei.lpFile = L"explorer.exe";
+    sei.lpParameters = L"shell:::{C58C4893-3BE0-4B45-ABB5-A63E4B8C8651}";
+    sei.nShow = SW_SHOWNORMAL;
+
+    if (!ShellExecuteExW(&sei)) {
+        Wh_Log(L"Troubleshooting shell command failed, using fallback");
+        ShellExecuteW(NULL, L"open", L"control.exe",
+                     L"/name Microsoft.Troubleshooting", NULL, SW_SHOWNORMAL);
+    }
+}
+
+// Lancia un diagnostic pack MSDT specifico se disponibile, altrimenti ricade
+// sull'elenco troubleshooter generico (stesso comportamento del ramo
+// "ALL OTHER PROBLEMS" gia' presente in OpenProblemAction).
+static void OpenMsdtDiagnostic(const WCHAR* packId) {
+    if (IsMsdtPackAvailable(packId)) {
+        WCHAR sysDir[MAX_PATH];
+        WCHAR msdtPath[MAX_PATH];
+        WCHAR params[64];
+        UINT len = GetSystemDirectoryW(sysDir, MAX_PATH);
+        if (len > 0 && len < MAX_PATH) {
+            StringCchCopyW(msdtPath, MAX_PATH, sysDir);
+            StringCchCatW(msdtPath, MAX_PATH, L"\\msdt.exe");
+        } else {
+            StringCchCopyW(msdtPath, MAX_PATH, L"msdt.exe");
+        }
+        StringCchPrintfW(params, ARRAYSIZE(params), L"-id %s", packId);
+
+        SHELLEXECUTEINFOW sei = { sizeof(sei) };
+        sei.fMask = SEE_MASK_FLAG_NO_UI;
+        sei.lpVerb = L"open";
+        sei.lpFile = msdtPath;
+        sei.lpParameters = params;
+        sei.nShow = SW_SHOWNORMAL;
+
+        if (ShellExecuteExW(&sei)) {
+            return; // successo
+        }
+        Wh_Log(L"msdt.exe -id %s failed to launch, falling back to generic troubleshooter", packId);
+    } else {
+        Wh_Log(L"MSDT pack %s not available on this system, falling back", packId);
+    }
+    OpenGenericTroubleshooter();
+}
+
+// ============================================================================
 // Open Problem Action (Firewall or Troubleshooting)
 // ============================================================================
 void OpenProblemAction(int problemType) {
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

---

1. **`CplHookRetryThread` can crash the process on unload, and its hooks would never activate anyway — remove it.** Three independent problems:
   - The thread runs for up to ~20 seconds and is never tracked or waited on: `CleanupModResources` doesn't know about `g_cplRetryThread`, so if the mod is disabled/updated while the thread is sleeping, the mod DLL is unmapped while the thread is still executing mod code — a crash in explorer.exe/control.exe. (The thread handle also leaks.)
   - Even a "successful" retry does nothing: `Wh_SetFunctionHook` called after `Wh_ModInit` returns only *registers* the hook operation; it isn't applied until `Wh_ApplyHookOperations` is called, which the thread never does. See the note in `windhawk_api.h`: hook operations are "applied with `Wh_ApplyHookOperations`", which is "called automatically by Windhawk after `Wh_ModInit`" — later calls need to be applied explicitly.
   - The retry can't help in the first place: `CplHookDui` force-loads `dui70.dll` itself via `LoadLibraryExW`, so the only remaining failure mode is `GetProcAddress` not finding the decorated names — and retrying one second later won't change that result.

   Since all `Wh_SetFunctionHook` calls happen inside `Wh_ModInit` → `CplInit` on the first attempt (where they're applied automatically), the retry thread adds only risk. Please delete it.

2. **The BitLocker check cannot work from unelevated Explorer, yet pays a blocking WMI round-trip on the tray UI thread every refresh.** The `ROOT\CIMV2\Security\MicrosoftVolumeEncryption` namespace (`Win32_EncryptableVolume`) requires administrator privileges — the code comment claiming it is "queryable unelevated" is incorrect. In unelevated `explorer.exe`, `ConnectServer`/`ExecQuery` fail with access denied, so the check silently never reports anything. Meanwhile, `CheckBitLocker` still runs `CoCreateInstance` + a cross-process RPC connect attempt synchronously on the tray thread (the thread that owns the flyout and notification windows) on every refresh — default every 5 s, and as often as every 1 s with the minimum `refreshInterval`.

   A working unelevated alternative is the `System.Volume.BitLockerProtection` shell property (`SHCreateItemFromParsingName` + `IShellItem2::GetProperty`) — this is what Explorer itself uses to draw the drive padlock overlays as a standard user. And since encryption status doesn't change minute-to-minute, query it once at startup (or very infrequently) and cache the result instead of re-checking on every refresh tick. If you keep any variant of the current code, at minimum it must not run a WMI query per refresh on the UI thread.

3. **`SetXML_Hook` now copies every XML string DirectUI parses.** The previous version had a cheap `wcsstr` prefilter *before any allocation*; the new code constructs `std::wstring xml(pszXML)` for every `DUIXmlParser::SetXML` call in every hooked process, and only then filters. DirectUI parses a lot of XML in the shell, so this is a per-call heap copy of potentially large buffers for unrelated UI. (The comment above the line even says "avoid building std::wstring work for irrelevant fragments" — but that's exactly what it does.) Please restore a raw-pointer prefilter, e.g. `if (!wcsstr(pszXML, L"atom(HavingAProblem)")) return CallOriginalSetXML(...);` before creating the `std::wstring`.

4. **`ValidateHubXml` now accepts up to 5 unbalanced `<Element>` tags — please make it strict again.** The old validator rejected any open/close mismatch; the new one only fails when the counts differ by more than 5, so a patch that corrupts the XML structure (up to 5 dangling elements) is passed to the Control Panel renderer as "valid". This tolerance looks like it papers over false mismatches from the old self-close detection — but `IsSelfClosingTag` now correctly handles whitespace before `/>`, so the original cause of false mismatches is fixed and the check can go back to `opens != closes → fail`. This matters more now that `LooksLikeActionCenterHub` also accepts partial fragments, where a bad patch has fewer surrounding invariants to catch it.

5. **Why `@include systemsettings.exe`?** The Security and Maintenance hub is a DirectUI Control Panel page hosted in `explorer.exe`/`control.exe`; the immersive Settings app doesn't render it. As written, the mod injects into SystemSettings only to run `CplInit`, which force-loads `dui70.dll` and installs the DUI hooks in a process that never shows the patched page. If there's a concrete scenario where SystemSettings hosts this page, please mention it; otherwise drop the include (the `control.exe` addition, in contrast, makes sense).

6. **Leftover AI artifacts and dead code:**
   - `OpenActionCenterFromFlyout()` is never called from anywhere. It's also preceded by the instruction-comment `// Aggiungi questa funzione dopo OpenProblemAction()` ("add this function after OpenProblemAction()"), which is clearly a prompt/instruction that leaked into the code, and internally it matches a window by the English-only title `L"Action Center"` and calls `Sleep(100)`. Please remove the whole function.
   - The free-form notes right after the settings block ("`// in the new the interface has been enhanced...`", "`// i assume that it is almost correct now...`") are changelog/working notes, not code documentation — the PR description already covers this. Please remove them.

<p>

- **`CheckRdpNla`:** the extra `(*criticalCount)++` runs even when `AddProblem` returned `FALSE` (problem list full, or already detected), inflating the critical count with no corresponding entry. Guard it: `if (AddProblem(PROB_RDP_NLA, idx, criticalCount)) (*criticalCount)++;`. Also, the comment says "AddProblem already counts it once" — it doesn't (`AddProblem` only bumps the count for firewall/autoupdate/antivirus), so the comment is misleading as written.
- **`CheckBatteryStatus`:** the comment documents a two-tier scheme ("<=10% = critical (STATE_ALERT), <=20% = warning") but the code only implements the single `pct <= 20` warning. Either implement the critical tier or fix the comment.
- **Eager decoding in `CplInit`:** `EnsureSolutionTemplate()` and `EnsureEmbeddedUifile()` are now decoded unconditionally at init in every included process; previously the embedded UIFILE was decoded lazily on the hook path. Reverting to lazy decoding avoids paying the conversion (and keeping the wide copy resident) in processes/sessions that never open the Security and Maintenance page.
- **`IsSystemOlderThanWin1021H2` comment:** `VerifyVersionInfoW` *is* subject to manifest-based version compatibility since Windows 8.1 — it works here only because explorer/control are manifested for Windows 10. The claim that it is "NOT affected by the compatibility shim" is inaccurate; worth correcting so nobody copies it.
- **README/changelog sync:** the Features section doesn't mention the new BitLocker and RDP-without-NLA checks, and the PR changelog omits the RDP NLA check entirely.

</p>
</details>

<p>

- **`PendingFileRenameOperations` is a noisy signal.** That value is set by many installers, uninstallers, and antivirus products for reasons unrelated to Windows Update, and can linger. Users are likely to see persistent "Windows updates are waiting. Restart your computer" warnings with no update pending. Consider limiting the pending-reboot check to the two dedicated keys (CBS `RebootPending` and WU `RebootRequired`), which are exactly the "reboot required to finish installing updates" markers the README describes.
- **The startup balloon fires on every mod load, not only at session start.** `g_isStartupCheck` is reset whenever the mod DLL is loaded, so the "startup" notification also appears on mod enable/update/reload and on every Explorer restart. Your top-of-file note suggests this is known; just flagging that the README's "on every Windows session start" undersells how often users with a standing warning (e.g. the battery one) will see it.
- **`useIconPosition` rejects legitimate icon rectangles containing a 0 coordinate** (`rcIcon.left != 0 && rcIcon.top != 0`): with a top- or left-docked taskbar, a valid tray-icon rect can have `top == 0` or `left == 0`, misclassifying it and falling back to the coarser edge anchoring. Checking only for a non-empty rect (`right > left && bottom > top`, plus `hrRect == S_OK`) would be more accurate.
- **Multi-monitor is primary-taskbar only:** `GetTaskbarEdge` and `AdjustWindowPosForTaskbar` find `Shell_TrayWnd` (primary taskbar) only; secondary taskbars (`Shell_SecondaryTrayWnd`) aren't considered. The tray icon itself lives on the primary taskbar so this is mostly fine, and the README already calls multi-monitor support partial — just an FYI for future iterations.

</p>
</details>

babamohammed2022:
> ### Submission review
> 
> 
> 1. **`CplHookRetryThread` can crash the process on unload, and its hooks would never activate anyway — remove it.** Three independent problems:
>    
>    * The thread runs for up to ~20 seconds and is never tracked or waited on: `CleanupModResources` doesn't know about `g_cplRetryThread`, so if the mod is disabled/updated while the thread is sleeping, the mod DLL is unmapped while the thread is still executing mod code — a crash in explorer.exe/control.exe. (The thread handle also leaks.)
>    * Even a "successful" retry does nothing: `Wh_SetFunctionHook` called after `Wh_ModInit` returns only _registers_ the hook operation; it isn't applied until `Wh_ApplyHookOperations` is called, which the thread never does. See the note in `windhawk_api.h`: hook operations are "applied with `Wh_ApplyHookOperations`", which is "called automatically by Windhawk after `Wh_ModInit`" — later calls need to be applied explicitly.
>    * The retry can't help in the first place: `CplHookDui` force-loads `dui70.dll` itself via `LoadLibraryExW`, so the only remaining failure mode is `GetProcAddress` not finding the decorated names — and retrying one second later won't change that result.
>    
>    Since all `Wh_SetFunctionHook` calls happen inside `Wh_ModInit` → `CplInit` on the first attempt (where they're applied automatically), the retry thread adds only risk. Please delete it.
> 2. **The BitLocker check cannot work from unelevated Explorer, yet pays a blocking WMI round-trip on the tray UI thread every refresh.** The `ROOT\CIMV2\Security\MicrosoftVolumeEncryption` namespace (`Win32_EncryptableVolume`) requires administrator privileges — the code comment claiming it is "queryable unelevated" is incorrect. In unelevated `explorer.exe`, `ConnectServer`/`ExecQuery` fail with access denied, so the check silently never reports anything. Meanwhile, `CheckBitLocker` still runs `CoCreateInstance` + a cross-process RPC connect attempt synchronously on the tray thread (the thread that owns the flyout and notification windows) on every refresh — default every 5 s, and as often as every 1 s with the minimum `refreshInterval`.
>    A working unelevated alternative is the `System.Volume.BitLockerProtection` shell property (`SHCreateItemFromParsingName` + `IShellItem2::GetProperty`) — this is what Explorer itself uses to draw the drive padlock overlays as a standard user. And since encryption status doesn't change minute-to-minute, query it once at startup (or very infrequently) and cache the result instead of re-checking on every refresh tick. If you keep any variant of the current code, at minimum it must not run a WMI query per refresh on the UI thread.
> 3. **`SetXML_Hook` now copies every XML string DirectUI parses.** The previous version had a cheap `wcsstr` prefilter _before any allocation_; the new code constructs `std::wstring xml(pszXML)` for every `DUIXmlParser::SetXML` call in every hooked process, and only then filters. DirectUI parses a lot of XML in the shell, so this is a per-call heap copy of potentially large buffers for unrelated UI. (The comment above the line even says "avoid building std::wstring work for irrelevant fragments" — but that's exactly what it does.) Please restore a raw-pointer prefilter, e.g. `if (!wcsstr(pszXML, L"atom(HavingAProblem)")) return CallOriginalSetXML(...);` before creating the `std::wstring`.
> 4. **`ValidateHubXml` now accepts up to 5 unbalanced `<Element>` tags — please make it strict again.** The old validator rejected any open/close mismatch; the new one only fails when the counts differ by more than 5, so a patch that corrupts the XML structure (up to 5 dangling elements) is passed to the Control Panel renderer as "valid". This tolerance looks like it papers over false mismatches from the old self-close detection — but `IsSelfClosingTag` now correctly handles whitespace before `/>`, so the original cause of false mismatches is fixed and the check can go back to `opens != closes → fail`. This matters more now that `LooksLikeActionCenterHub` also accepts partial fragments, where a bad patch has fewer surrounding invariants to catch it.
> 5. **Why `@include systemsettings.exe`?** The Security and Maintenance hub is a DirectUI Control Panel page hosted in `explorer.exe`/`control.exe`; the immersive Settings app doesn't render it. As written, the mod injects into SystemSettings only to run `CplInit`, which force-loads `dui70.dll` and installs the DUI hooks in a process that never shows the patched page. If there's a concrete scenario where SystemSettings hosts this page, please mention it; otherwise drop the include (the `control.exe` addition, in contrast, makes sense).
> 6. **Leftover AI artifacts and dead code:**
>    
>    * `OpenActionCenterFromFlyout()` is never called from anywhere. It's also preceded by the instruction-comment `// Aggiungi questa funzione dopo OpenProblemAction()` ("add this function after OpenProblemAction()"), which is clearly a prompt/instruction that leaked into the code, and internally it matches a window by the English-only title `L"Action Center"` and calls `Sleep(100)`. Please remove the whole function.
>    * The free-form notes right after the settings block ("`// in the new the interface has been enhanced...`", "`// i assume that it is almost correct now...`") are changelog/working notes, not code documentation — the PR description already covers this. Please remove them.
> 
> Optional improvements
> 
> Minor polish — none of this affects users much, so it's your call.
> 
> * **`CheckRdpNla`:** the extra `(*criticalCount)++` runs even when `AddProblem` returned `FALSE` (problem list full, or already detected), inflating the critical count with no corresponding entry. Guard it: `if (AddProblem(PROB_RDP_NLA, idx, criticalCount)) (*criticalCount)++;`. Also, the comment says "AddProblem already counts it once" — it doesn't (`AddProblem` only bumps the count for firewall/autoupdate/antivirus), so the comment is misleading as written.
> * **`CheckBatteryStatus`:** the comment documents a two-tier scheme ("<=10% = critical (STATE_ALERT), <=20% = warning") but the code only implements the single `pct <= 20` warning. Either implement the critical tier or fix the comment.
> * **Eager decoding in `CplInit`:** `EnsureSolutionTemplate()` and `EnsureEmbeddedUifile()` are now decoded unconditionally at init in every included process; previously the embedded UIFILE was decoded lazily on the hook path. Reverting to lazy decoding avoids paying the conversion (and keeping the wide copy resident) in processes/sessions that never open the Security and Maintenance page.
> * **`IsSystemOlderThanWin1021H2` comment:** `VerifyVersionInfoW` _is_ subject to manifest-based version compatibility since Windows 8.1 — it works here only because explorer/control are manifested for Windows 10. The claim that it is "NOT affected by the compatibility shim" is inaccurate; worth correcting so nobody copies it.
> * **README/changelog sync:** the Features section doesn't mention the new BitLocker and RDP-without-NLA checks, and the PR changelog omits the RDP NLA check entirely.
> 
> Functionality notes

Thanks for the review. If possible, let me know if there are any other issues to address

PR #4855, opened 2026-07-21
Add OmniButton Customizer v2.0

sb4ssman:
Takes the native Windows 11 OmniButton — the network / volume / battery cluster that opens Quick Settings — and lets you arrange its items into any shape, hide the ones you don't want, and restyle each one independently. The button stays in its native tray position, so other mods' "before OmniButton" anchors still mean what they always did, and it uses no XAML Diagnostics, so it coexists with Windows 11 Taskbar Styler.

This replaces the v1.0 that was previously in this PR. Since v1.0 was never published, v2.0 marks the settings contract rather than a release history: the whole settings surface moved to grouped keys (Placement, Content, Layout, Size, Adjust, Surface) and a single nestable **Arrangement** expression that supersedes `itemOrder`, the grid-mode family, the coupled battery mode, and all eight per-item nudge settings.

### Highlights

- **One Arrangement field.** `auto` fits the available items to the taskbar height; anything else is an expression — `|` places side by side, `,` stacks, parentheses nest, and `name[dx,dy]` nudges one item or a whole group. `auto` logs the expression it generated, so it can be pasted back into the field and edited.
- **Only the controls each item can honor.** Network, volume and the battery are each drawn by a *stack* of glyphs layered exactly on top of one another — that is how they show signal strength, a mute slash, or a charge level — so resizing one layer pulls it off the others. Size and font family are therefore offered only on the battery percentage, the one item that is genuinely a single piece of text. Color and opacity are available on all four.
- **Recoloring works on template-bound tray glyphs.** `InnerTextBlock` inside a `SystemTray.IconView` is template-bound, so a local write is re-asserted away; the mod styles the outermost `Control` between the leaf and the host instead.
- **Content sizing.** `Item width: 0` measures each item and reserves exactly what it needs. The percentage is always measured and treats that value as a floor, so "9%" growing to "100%" cannot clip at the group edge.
- **Writes no Windows settings and no registry values.** An earlier build drove the Windows battery-percentage setting; it was removed, because Explorer only sometimes re-read the value and the Settings page never refreshed, so the control worked once and then appeared dead.
- **Stands down on vertical taskbars.** `taskbar-vertical` applies a `RotateTransform` to the same elements this mod writes a `TranslateTransform` to. Rather than fight over one dependency property, the mod detects a vertical taskbar by the tray's own aspect, leaves everything as found, and logs once.

### Review items from the previous round

Both blocking items are fixed:

- **DPI mixing** — row capacity was dividing physical pixels by a DIP pitch. Metrics now come from a shared helper returning orientation, DPI, and the constrained extent in DIPs, and `2 × padY` is reserved before dividing.
- **`SYMBOL_HOOK` array naming** — renamed, with all three target modules named in a comment directly above the array.

Also fixed from the optional list: `[[clang::no_destroy]]` across the XAML globals with a controlled UI-thread `reset()` in `Wh_ModUninit`; the optional+reset form on `g_settings`; the duplicate `LoadColorSetting`; the `slotWidth` 1–15 surprise; and the unused `slotW` parameter.

Deliberately not taken, to avoid bundling untested changes into a tested build: the retry start/stop mutex, `IsWindow()` revalidation, structural network/volume detection, and a `WM_DPICHANGED` recompute. Happy to take any of these as follow-ups.

### Testing

Live-tested on Windows 11 at 100%, 125% and 150% display scaling: automatic and hand-written arrangements, per-item and per-group nudges, hiding individual items, negative item spacing, per-item color, percentage size, single- and double-height taskbars, running alongside several other taskbar and tray mods, and enable/disable/Explorer-restart teardown.

## Changelog

* Replaced the entire settings surface with grouped keys and one nestable `Arrangement` expression; `itemOrder`, grid modes, smart layout, fixed rows/columns, slot width and height, the coupled battery mode, and the eight per-item nudge settings are gone
* Fixed physical-pixel/DIP mixing in the row-capacity heuristic
* Fixed recoloring on template-bound system-tray glyphs
* Fixed the battery percentage clipping at the group edge
* Added fit-to-content item sizing and negative item spacing
* Removed the Windows battery-percentage setting write
* Renamed the symbol hook array and named all three target modules
* Applied `[[clang::no_destroy]]` to the XAML globals with a UI-thread reset

## Mod authorship

This mod was created by:

- - [x] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/omnibutton-customizer.wh.cpp
+++ b/mods/omnibutton-customizer.wh.cpp
@@ -1,0 +1,4196 @@
+// ==WindhawkMod==
+// @id              omnibutton-customizer
+// @name            OmniButton Customizer
+// @description     Arrange the Windows 11 OmniButton's network, volume, battery and percentage into any layout you write, hide any of them, and recolor them independently
+// @version         2.0
+// @author          sb4ssman
+// @github          https://github.com/sb4ssman
+// @include         explorer.exe
+// @architecture    x86-64
+// @compilerOptions -lole32 -loleaut32 -lruntimeobject -lversion
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# OmniButton Customizer
+
+A [Windhawk](https://windhawk.net) mod for Windows 11 that takes the native
+OmniButton — the network / volume / battery cluster that opens Quick Settings —
+and lets you arrange its items into any shape you like, hide the ones you don't
+want, and restyle each one independently.
+
+These native icons are compound and a little weird — each is several glyphs
+layered on top of one another, each with its own built-in visual origin — so a
+mathematically correct grid does not necessarily look like one. `auto` gets the
+shape right, but it does not make pleasing arrangements; expect to adjust.
+Every example below is a real arrangement string, nudges and all.
+
+![All four items as a 2×2 block on a single-height taskbar](https://raw.githubusercontent.com/sb4ssman/Windhawk-Mod-Lab/main/omnibutton-customizer/assets/single-height-auto.png)
+*Straight out of the box: `Arrangement` left at `auto`, which fits the four
+native items to the taskbar height and settles on a 2×2 block.*
+
+## Showcase
+
+![The same 2×2 block, written by hand](https://raw.githubusercontent.com/sb4ssman/Windhawk-Mod-Lab/main/omnibutton-customizer/assets/single-height-arranged-2x2.png)
+*The same shape written out and optically corrected:*
+
+```text
+(network[-6,2] | volume[-2,4]), (battery[1,0] | percent[4,-2])
+```
+
+*Two rows joined into a column — the parentheses matter, because `,` binds
+tighter than `|`. Every time `auto` runs it logs the arrangement it generated,
+so you can paste that line into the field and start nudging from there; the
+automatic and hand-written paths are the same field and the same syntax.*
+
+![Network and volume above a centered battery](https://raw.githubusercontent.com/sb4ssman/Windhawk-Mod-Lab/main/omnibutton-customizer/assets/single-height-arranged-2-over-1.png)
+*Three items with the percentage left out: two across the top, the battery
+centered below. `Short row or column` decides how a ragged last group lines up.*
+
+![The items in a single row](https://raw.githubusercontent.com/sb4ssman/Windhawk-Mod-Lab/main/omnibutton-customizer/assets/single-height-arranged-reverse.png)
+*A single row, in an order you choose rather than the native one.*
+
+![A tight two-high stack](https://raw.githubusercontent.com/sb4ssman/Windhawk-Mod-Lab/main/omnibutton-customizer/assets/single-height-compact-stack.png)
+*Pulled in close with a negative `Size` → `Item spacing`. That is the setting
+that tightens a cluster — horizontal padding only reserves space at the two
+outside edges and can never change the distance between items.*
+
+![The four items arranged as a diamond, with the native tooltip showing](https://raw.githubusercontent.com/sb4ssman/Windhawk-Mod-Lab/main/omnibutton-customizer/assets/single-height-diamond-adjusted-with-hover.png)
+*A diamond — volume on top, battery below, network and the percentage on the
+sides:*
+
+```text
+network[4,-2] | volume[2,-4], battery[0,4] | percent[0,-2]
+```
+
+*Nesting one stacked pair between two single items is all it takes; no
+parentheses are needed here because `,` already binds tighter than `|`. It is
+still the native button, so the hover tooltip and the click through to Quick
+Settings behave exactly as they always did.*
+
+![A tight cluster with an enlarged battery percentage](https://raw.githubusercontent.com/sb4ssman/Windhawk-Mod-Lab/main/omnibutton-customizer/assets/single-height-stacked-with-percent-emphasis.png)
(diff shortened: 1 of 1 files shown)


m417z:
In addition to the notes below, improve the `[[clang::no_destroy]]` usage as discussed in the other PRs.

Regarding 1, it's generally preferable to test mods in different DPI settings.

### Submission review

---

1. **`ResolveGeometry` mixes physical pixels with XAML DIPs — the grid overflows the taskbar on any display scaling above 100%.** `GetWindowRect` returns physical pixels (Explorer is per-monitor-DPI aware), but the result is divided by the 24 DIP row pitch (or the `slotHeight` setting) to get `availableRows`, and `taskbarH / rows` is then assigned to XAML `Height`, which is in DIPs. At 100% scaling this cancels out; at 125%/150% (the default on most laptops) the taskbar is 60/72 physical px but only 48 DIPs, so `availableRows` is overestimated and `slotH` comes out as 30-32 DIPs — a 2-row grid becomes 60-64 DIPs tall on a 48-DIP taskbar and gets clipped. Mixed-DPI multi-monitor setups hit the same math. Convert to DIPs before comparing, as [taskbar-icon-size](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/taskbar-icon-size.wh.cpp#L1811) does with `MulDiv` + `GetDpiForWindow`:

   ```cpp
   int taskbarH = 48;
   if (hTaskbarWnd) {
       RECT r{};
       if (GetWindowRect(hTaskbarWnd, &r)) {
           UINT dpi = GetDpiForWindow(hTaskbarWnd);
           taskbarH = MulDiv(r.bottom - r.top, 96, dpi ? dpi : 96);
       }
   }
   ```

2. **The `systemTrayDllHooks` array name only declares `SystemTray.dll`, but the hooks are resolved against three possible modules.** `GetSystemTrayModuleHandle()` can return `SystemTray.dll`, `Taskbar.View.dll`, or `ExplorerExtensions.dll`, and `HookSystemTraySymbols` applies the array to whichever one is found. By convention the array's name or a comment above it must declare exactly the module(s) it's resolved against. Use a neutral name with a comment listing all of them, as [taskbar-tray-system-icon-tweaks](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/taskbar-tray-system-icon-tweaks.wh.cpp#L1586) does:

   ```cpp
   // SystemTray.dll, Taskbar.View.dll, ExplorerExtensions.dll
   WindhawkUtils::SYMBOL_HOOK symbolHooks[] = {{
   ```

<p>

- `[[clang::no_destroy]]` on `g_settings` is unneeded: `ModSettings` holds only arrays/ints/enums, so it's trivially destructible and safe at process shutdown as-is. Unneeded suppression is noise — remove the attribute there.
- `g_autoRevokerList` and `g_propertySnapshots` carry a bare `[[clang::no_destroy]]` on container types. The preferred form for non-nullable containers is `[[clang::no_destroy]] std::optional<...>` with `reset()` on the unload path, which makes the destruction explicit. In this specific code it's only conformance polish: `clear()` on a `list`/`vector` does destroy the elements (releasing the XAML refs on the taskbar thread), and the residual buffer is plain heap.
- `StartRetryThread`/`StopRetryThread` manipulate `g_retryThread`/`g_retryStopEvent` without synchronization, and can run concurrently from `Wh_ModSettingsChanged` (arbitrary thread) and `TrayUI_StartTaskbar_Hook` (taskbar thread) — two simultaneous `StopRetryThread` calls could double-close the handles. Only reachable when a settings change coincides with a taskbar restart, hence optional; a small mutex around start/stop closes it.
- `ApplyOnTaskbarWindowThread` and `Wh_ModUninit` prefer the cached `g_taskbarWnd` without validating it. If `Shell_TrayWnd` is recreated in-process, the retry thread keeps dispatching to a dead handle (the `IconView` Loaded path eventually recovers, since `ApplyAllSettings` re-finds the window). An `IsWindow()` check — or always calling `FindCurrentProcessTaskbarWnd()` — makes the retry path self-healing.
- `LoadColorSetting` is an exact duplicate of `LoadStringSetting<32>` — the template alone suffices.
- The `slotW` parameter of `WalkSetupBatteryInnerPanel` is unused, and `<cmath>` appears to be an unused include.

</p>
</details>

<p>

- `slotWidth` values 1-15 silently jump to the default 32 (`sw < 16 ? 32 : sw`) instead of clamping to the 16 minimum — a user lowering the value below 16 gets a *wider* slot. Clamping to 16 would be less surprising.
- Wifi and volume are identified purely positionally (items-host children 0 and 1), while battery is detected structurally. If a future Windows build reorders or inserts slots, the first two would be mislabeled; the "Unknown slot" logging is a good safety net, but structural detection for all items would be more robust.
- The layout isn't recomputed when the taskbar height or monitor DPI changes at runtime (e.g. `WM_DPICHANGED`, taskbar size mods toggling) — only on settings changes, `TrayUI::StartTaskbar`, or new `IconView` creation. Probably fine in practice; noting it as a known limitation.

</p>
</details>

radioactive-exe:
Hello @sb4ssman! This mod seems very interesting and useful (I never say no to more screen realestate by merging or stacking icons) - just seems like it's been a while, so I was curious if you plan to continue working on it? Would love to see how it shapes up and if/when it gets added to the store later on.

sb4ssman:
Hi @radioactive-exe I have not abandoned my mod family. I've been building things and have more to contribute. Stay tuned!

PR #4857, opened 2026-07-21
Add Energy Saver Power Plan Link mod

Yusseter:
Adds a new Windhawk mod that links Windows 11 Energy Saver to the standard Power Saver power plan.

The mod saves the active power scheme when Energy Saver is enabled, switches to Power Saver, and restores the previous scheme when Energy Saver is disabled. Manual power plan changes made while Energy Saver is active are preserved.

Tested with Ultimate Performance, Balanced, and Power Saver plans, including an Explorer restart.

## Changelog

* Initial release of Energy Saver Power Plan Link.

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [x] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/energy-saver-power-plan-link.wh.cpp
+++ b/mods/energy-saver-power-plan-link.wh.cpp
@@ -7,9 +7,8 @@
 // @github          https://github.com/Yusseter
 // @homepage        https://github.com/Yusseter/energy-saver-power-plan-link
 // @license         MIT
-// @include         explorer.exe
-// @architecture    x86-64
-// @compilerOptions -lpowrprof
+// @include         windhawk.exe
+// @compilerOptions -lpowrprof -lshell32
 // ==/WindhawkMod==
 
 // ==WindhawkModReadme==
@@ -27,16 +26,18 @@ When Energy Saver is disabled:
 - If the user manually selected another plan, that selection is preserved.
 
 Disabling the mod also restores the previous plan when appropriate.
+
+> On some Modern Standby systems, the classic Power Saver plan might not be
+> available. In that case, Windows can't switch to that plan and the mod leaves
+> the current plan unchanged.
 */
 // ==/WindhawkModReadme==
-
-#ifndef _WIN32_WINNT
-#define _WIN32_WINNT 0x0A00
-#endif
 
 #include <windows.h>
 #include <powrprof.h>
 #include <powersetting.h>
+#include <shellapi.h>
+#include <stdio.h>
 
 #include <cstring>
 #include <mutex>
@@ -77,6 +78,8 @@ constexpr GUID kPowerSaverSchemeGuid = {
 constexpr wchar_t kPreviousSchemeValueName[] = L"PreviousPowerScheme";
 
 HPOWERNOTIFY g_notificationHandle = nullptr;
+HANDLE g_keepAliveEvent = nullptr;
+HANDLE g_keepAliveThread = nullptr;
 std::mutex g_stateMutex;
 
 bool GetActivePowerScheme(GUID* scheme) {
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

---

* **Convert the mod to a "tool mod" running in a dedicated process.** The mod doesn't hook anything — it only uses `PowerSettingRegisterNotification`, which works from any process — so there's no reason to inject into `explorer.exe`. Hosting it there has two concrete downsides:
  * There can be multiple `explorer.exe` processes (e.g. with the "Launch folder windows in a separate process" option), so the mod runs multiple times: each instance registers its own notification and races the others over the shared `PreviousPowerScheme` value. The `std::mutex` is per-process and doesn't help across instances.
  * Any instability in the mod affects the whole Windows shell.

  See [Mods as tools: Running mods in a dedicated process](https://github.com/ramensoftware/windhawk/wiki/Mods-as-tools:-Running-mods-in-a-dedicated-process). Concretely:
  * Change `@include explorer.exe` to `@include windhawk.exe`.
  * Remove the `@architecture x86-64` line — `windhawk.exe` is a 32-bit process, so with `x86-64` the tool mod would never load. Existing tool mods omit `@architecture` entirely.
  * Rename the callbacks to the `WhTool_*` equivalents. The tool-mod pattern supports `WhTool_ModInit`, `WhTool_ModSettingsChanged`, and `WhTool_ModUninit` — there's no `BeforeUninit` variant, so move the unregister-and-restore logic from `Wh_ModBeforeUninit` into `WhTool_ModUninit` (this mod has no Windhawk hooks, so nothing changes semantically).
  * Paste the launcher snippet from the wiki **verbatim** at the end of the mod — please don't refactor or trim it, so it can be diffed against the reference.

  Two merged mods do exactly this — register power-setting notifications from a tool mod — and can serve as references: [lid-sleep-delay.wh.cpp](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/lid-sleep-delay.wh.cpp#L499) and [low-battery-alarm.wh.cpp](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/low-battery-alarm.wh.cpp#L2932).

<p>

* The `#ifndef _WIN32_WINNT` block (lines 33-35) is unnecessary — Windhawk's toolchain already targets a modern Windows SDK level. Harmless, but it can be dropped.
* The hand-rolled `DeviceNotifySubscribeParameters` struct is fine (other mods, e.g. [fix-missing-taskbar-icons.wh.cpp](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/fix-missing-taskbar-icons.wh.cpp#L93), define the same thing because the toolchain headers may lack it), but consider guarding it the same way that mod does, or using the SDK's `DEVICE_NOTIFY_SUBSCRIBE_PARAMETERS` name if it compiles, so readers recognize the standard type.

</p>
</details>

<p>

* On Modern Standby (S0) systems — most recent laptops — the classic power plans other than Balanced are typically hidden or absent, so switching to the Power Saver scheme may fail and the mod will effectively do nothing (it handles the failure gracefully by logging and clearing the saved record, which is good). Consider mentioning this limitation in the README so users on such machines know why nothing happens.

</p>
</details>

Yusseter:
Addressed the review:

- Converted the mod to a dedicated-process tool mod targeting windhawk.exe.
- Removed the architecture restriction.
- Moved initialization and cleanup to the WhTool callbacks.
- Added the official tool-mod launcher verbatim.
- Removed the unnecessary _WIN32_WINNT block.
- Added a Modern Standby limitation note.
- Added a low-overhead keep-alive thread so the dedicated tool process remains active.

Retested power-plan switching and restoration, manual power-plan changes, mod disable cleanup, and Explorer restart behavior.

PR #4863, opened 2026-07-22
Live Video Wallpaper

sysakshay:
# Live Video Wallpaper v1.1.0

This pull request updates the `live-video-wallpaper` mod with major performance optimizations, audio volume controls, aspect ratio scaling fixes, zero-latency desktop context menu responsiveness, and non-blocking file picker enhancements.

preview2
preview3
preview4

## Changelog

* **Audio & Volume Control:** Added `audioVolume` settings dropdown in Windhawk UI (`0%` to `100%`) with automatic audio stream unmuting and volume restoration when switching focused applications.
* **Aspect Ratio & Display Fixes:** Fixed aspect ratio scaling and vertical stretching on wide, ultrawide (21:9 / 32:9), and 1:1 square monitors across `Fill`, `Cover`, and `Fit` modes (`pDst = &dest`).
* **Zero-Latency Occlusion Events:** Integrated `EVENT_SYSTEM_FOREGROUND` event hooks (`WINEVENT_SKIPOWNPROCESS`) for instant (< 16ms) audio/video pause when application windows are focused or maximized. Added configurable `occlusionInterval` setting (`100ms`, `250ms`, `500ms`).
* **Desktop Context Menu Responsiveness:** Implemented `HTTRANSPARENT`, `WS_EX_TRANSPARENT`, and `WS_DISABLED` window styles so right-clicking desktop icons or background opens context menus with 0ms latency.
* **Async Non-Blocking File Picker:** Refactored `Ctrl + Alt + G` interactive file selection to run on a background thread without pausing the 60 FPS video wallpaper loop. Added support for `.mp4`, `.m4v`, `.mov`, `.wmv`, `.webm`, and automatic path sync with Windhawk UI settings.
* **Low-Level Performance & Memory Optimizations:** Cached D3D11 render target views (`tempRTV`) while handling DXGI flip-discard swap-chain buffer rotation in `Fit` mode. Replaced `std::vector` heap allocations in profiler percentile calculations with zero-allocation 60 Hz stack buffers. Cached `WorkerW` window handles ($O(1)$) and gated GDI texture flags.

## Mod authorship

This mod was created by:

- [ ] Claude
- [ ] ChatGPT
- [ ] Gemini
- [ ] Other (please specify):


m417z:
### Submission review

---

1. **Make this a tool mod instead of injecting into explorer.exe.** The mod installs no function hooks at all — everything it does (`SetWinEventHook`, `RegisterHotKey`, `EnumWindows`, `SetParent` into WorkerW, timers, D3D/Media Foundation rendering) works from any process, including cross-process parenting into WorkerW (this is exactly how standalone wallpaper apps like Lively work). Running a D3D11 device, a Media Foundation engine decoding arbitrary user-supplied video files, and a 60 FPS render loop *inside the shell process* means any decoder/driver crash or hang takes Explorer down with it — your own comment in `Wh_ModUninit` ("restart explorer.exe to recover") describes the failure mode. There's also a duplication bug: with "Launch folder windows in a separate process" enabled, a second explorer.exe instance spins up a second complete wallpaper stack (the hotkey fallback chains would then silently bind alternate combos in that instance — see the Functionality notes). The tool-mod framework's launcher handles single-instancing for you. Please follow [Mods as tools: Running mods in a dedicated process](https://github.com/ramensoftware/windhawk/wiki/Mods-as-tools:-Running-mods-in-a-dedicated-process): `@include windhawk.exe`, rename the lifecycle callbacks to `WhTool_*`, and paste the launcher boilerplate verbatim (see [explorer-folder-hover-menu.wh.cpp](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/explorer-folder-hover-menu.wh.cpp) for a complete example). For comparison, [desktop-live-overlay](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/desktop-live-overlay.wh.cpp#L2937) hooks `CreateWindowExW` and so has a genuine reason to live in explorer.exe — this mod doesn't.

2. **`Wh_ModInit` delays Explorer startup by ~10 seconds on every logon** (line 3640). `Wh_ModInit` runs before the target process starts executing, and `WaitForSingleObject(g_threadReadyEvent, 10000)` waits for the wallpaper thread — which itself waits for the `Progman` window, which Explorer cannot create until `Wh_ModInit` returns. At boot/Explorer restart this circular wait always burns the full 10 s timeout, delaying the shell. The wait result isn't even used; just remove the wait and let the thread initialize asynchronously.

3. **Mod threads can outlive the DLL and crash Explorer on unload.**
   - The `Wh_ModUninit` timeout path (lines 3656–3681) returns while the wallpaper thread is still running, on the assumption that "Windows cannot unload this DLL while this thread is still alive." That assumption is incorrect: after `Wh_ModUninit` returns, Windhawk unloads the DLL, and `FreeLibrary` does *not* wait for threads executing inside the module — when the stuck thread resumes, it executes unmapped code and crashes explorer.exe. And the timeout is easy to hit: the boot-phase `Sleep(500)` x 120 Progman loop (lines 3390–3394) and the `Sleep(200)` WorkerW retries are not interruptible, so disabling the mod during that window guarantees the 3 s wait expires. Fix: use an exit event and `WaitForSingleObject(exitEvent, 500)` instead of `Sleep` so the thread reacts to shutdown promptly, and don't return from `Wh_ModUninit` until the thread has actually exited (a slow unload is recoverable; a crash isn't).
   - `FilePickerThreadProc` (lines 3324–3329) is fire-and-forget: its handle is closed immediately and it's never joined. If the mod is unloaded while the `GetOpenFileNameW` dialog is open, the thread survives the DLL and crashes; `Wh_ModUninit` also calls `DeleteCriticalSection(&g_pathLock)` (line 3685) while the picker thread may still acquire it. Track the thread handle, signal/join it in `Wh_ModUninit`, and while you're at it, guard against spawning multiple concurrent dialogs from repeated hotkey presses.

4. **Persistent registry writes violate the reversibility principle.** The mod writes `HKCU\Software\Windhawk\LiveVideoWallpaper\LastPickedVideoPath` (lines 1625–1634, 3307–3315) — this survives disabling and uninstalling the mod, and squats in Windhawk's own key namespace. Windhawk provides mod-scoped persistent storage for exactly this: use `Wh_SetStringValue(L"LastPickedVideoPath", ...)` / `Wh_GetStringValue` (see `windhawk_api.h`) instead.

5. **The picked video silently reverts when any unrelated setting changes.** `FilePickerThreadProc` sets `s_lastWindhawkSettingVideoPath = fileBuffer` (line 3304). After picking file B while the settings UI still shows path A, changing *any* setting (e.g. volume) makes `LoadSettings` see `settingPath (A) != s_lastWindhawkSettingVideoPath (B)`, conclude the user "changed" the path, and reload video A. Fix: don't overwrite `s_lastWindhawkSettingVideoPath` in the picker — it should track only the last value *seen from the settings UI*, so an unchanged UI value is correctly ignored while a real user edit still wins.

6. **Settings block: numeric defaults with `$options` and mismatched read APIs.** `targetFps: 60` and `audioVolume: 100` (lines 116–141) have numeric defaults, which makes them number settings — but `$options` is only valid for string settings, and the code reads both with `Wh_GetStringSetting`. Suggested fix: make `audioVolume` a plain number setting (0–100, no `$options`) read with `Wh_GetIntSetting`, and make `targetFps` a number setting where `0` means "monitor refresh rate" (documented in `$description`), also read with `Wh_GetIntSetting`.

7. **Global D3D/MF-owning objects have non-trivial destructors that run at process shutdown.** `g_player` (line 2677) and the profiler's namespace-scope `ComPtr` globals (`s_dxgiAdapter3`, `s_d3dDevice`, `s_d3dContext`, `s_overlayTexture`) release DXGI swap chains, the D3D device, and the media engine from CRT destructors during `DLL_PROCESS_DETACH` at process termination (sign-out, reboot, Explorer restart) — under the loader lock, after all other threads have been killed. Releasing DXGI/D3D objects there is documented-unsafe and can hang or crash the process on every exit. Your controlled-unload cleanup already exists (the wallpaper thread calls `g_player.Shutdown()` before exiting, leaving these destructors as no-ops on that path), so just suppress the automatic destructors: `[[clang::no_destroy]]` on the nullable `ComPtr` globals, and `[[clang::no_destroy]] std::optional<VideoPlayer> g_player;` (released via the existing `Shutdown()` on the wallpaper thread). This becomes much less critical once the mod is a tool mod (item 1), since the terminating process is then its own dedicated process rather than Explorer.

8. **Don't treat `ERROR_CLASS_ALREADY_EXISTS` as success in `RegisterClassW`** (line 3408). The thread-exit path correctly calls `UnregisterClassW`, so the class can only "already exist" when a previous mod instance failed to clean up — and in that case the stale registration's `lpfnWndProc` points into a stale (possibly unmapped) module, so creating a window with it dispatches messages to a dangling pointer. Registration failure should be treated as fatal for this thread, not papered over.

<p>

- **`MediaEngineNotify` is leaked on every init/recovery cycle.** It's created with `refCount = 1` (line 1988), `SetUnknown` adds a reference, and the mod never releases its own initial reference — so the refcount never reaches zero. Call `notify->Release()` after handing it to the attributes (and also on the early-error paths in `InitMediaEngine`).
- **The profiler isn't thread-safe despite its header claims.** `EventNotify` calls `Profiler::BeginSection`/`EndSection` from Media Foundation work-queue threads (line 1775) while the wallpaper thread uses the same unsynchronized `s_sections` / `s_sectionStack` / `s_stackDepth` state. Simplest fix: drop the section bracket from `EventNotify` (it only measures a `PostMessage`) and keep profiling wallpaper-thread-only.
- **The `#ifdef DEBUG` diagnostics never compile** — `DEBUG` is not defined in Windhawk builds, so all those throttled gate/stall logs are dead code. `Wh_Log` already compiles to a cheap check and is disabled unless the user turns logging on, so you can use it directly (keep the time-based throttling if you want).
- **Unused libraries in `@compilerOptions`**: `-ldxgi` (no dxgi.dll free functions are called — all DXGI interfaces come from `QueryInterface`) and `-lpsapi` (`K32GetProcessMemoryInfo` is a kernel32 export).
- `Wh_GetStringSetting` never returns `NULL` (it returns `L""` on error), so the scattered null checks on its results (e.g. lines 1612–1615) are unnecessary — and inconsistent, since `batteryMode` isn't checked at all.
- **README visual**: the PR description has nice screenshots, but the README itself has none. Since the mod has a very visible effect, please add a screenshot/GIF to the README — note that only `i.imgur.com` and `raw.githubusercontent.com` are allowed image hosts.

</p>
</details>

<p>

- **Multi-monitor**: a single window spans the whole virtual screen, so one video is stretched/letterboxed across the union of all monitors rather than rendered per-monitor, and mixed refresh rates/DPI aren't considered. Worth documenting as a limitation in the README if per-monitor rendering isn't planned.
- **Ctrl+Alt+H only hides the window** — `Tick()` isn't gated on `g_wallpaperHidden`, so the mod keeps decoding, presenting, and playing audio at full rate while invisible. Consider pausing the engine and killing the render timer while hidden.
- **WebM/MOV support is overstated**: Media Foundation has no built-in VP8/VP9/WebM decoder (it needs the optional Store codec extensions), and `.mov` support is partial. The picker filter and README promise formats that will silently fail on a default Windows install — consider trimming the list or noting the codec requirement.
- **The hotkey fallback chains** (lines 3505–3513 etc.) silently bind a different combo than the documented one when registration fails; the user can only discover which via the logs. Consider making the hotkeys configurable in settings instead, or dropping the fallbacks so behavior matches the README.
- **The profiler HUD presents fabricated numbers as live telemetry**: `+30` "internal COM objects" and `+4` decoder surfaces (lines 2599–2605), a hardcoded battery level of 50% (line 1116), and a hardcoded `"H.264 (Hardware Decode)"` / 60 FPS codec string regardless of the actual stream. For a diagnostics feature this is misleading — either compute real values or label them as estimates.

</p>
</details>

PR #4869, opened 2026-07-22
UPDATE: Enhanced the UI to make it more similar to the Windows 7 flyout

babamohammed2022:
## Changelog

- Added network location icons (Home/Public/Work based on the type of the network) to make it more similar to the original Windows 7 flyout
- Added option to disable network location icons
- Enhanced the interface
- Enhanced the image loading logic
- Fixed the refresh button position
- Updated the README

## Mod authorship

This mod was created by:

- - [X] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/win7-network-flyout-recreation.wh.cpp
+++ b/mods/win7-network-flyout-recreation.wh.cpp
@@ -117,6 +117,10 @@ If you encounter issues, please report them on the author of the mod.
 */
 // ==/WindhawkModSettings==
 // ## Changelog 
+// - van.dll alignment: row height 30rp + 24rp name + padding
+//   rect(8rp,3rp,10rp,3rp), signal icon re-centered (matches the
+//   real Windows 7 van.dll UIFILE). Refresh button, footer link
+//   alignment, separator and all hover effects left unchanged.
 // - 3.4.0: Network location detection now only runs while the flyout is
 //   visible (it was previously re-run on every 3s auto-refresh tick even
 //   while hidden), and the last detected category is kept instead of reset
@@ -188,7 +192,7 @@ If you encounter issues, please report them on the author of the mod.
 #define WINDOW_HEIGHT_BASE       405
 #define HEADER_HEIGHT_BASE       105
 #define FOOTER_HEIGHT_BASE       60
-#define ROW_HEIGHT_NORMAL_BASE   26
+#define ROW_HEIGHT_NORMAL_BASE   30
 #define ROW_HEIGHT_EXPANDED_BASE 74
 
 static UINT g_dpi = 96;
@@ -1259,12 +1263,12 @@ static const LocalePack g_Locales[] = {
         L"Connessione a %s fallita",
         L"Rete %d",
         L"Tipo di sicurezza:",
-        L"Intensit\u00E0 del segnale:",
+        L"Potenza segnale:",
         L"Tipo di radio:",
         L"Eccellente",
-        L"Buono",
-        L"Discreto",
-        L"Scarso",
+        L"Buona",
+        L"Sufficiente",
+        L"Scarsa",
         L"Nessun segnale",
         L"Connessione in corso...",
         L"Disconnessione in corso...",
@@ -1277,7 +1281,7 @@ static const LocalePack g_Locales[] = {
         L"Timeout durante la connessione",
         L"Il tentativo di connessione \u00E8 scaduto. La rete potrebbe essere fuori portata.",
         L"Inserire una chiave di sicurezza di rete.",
-        L"Risoluzione dei problemi",
+        L"Risoluzione problemi",
         L"Apri Centro connessioni di rete e condivisione",
     }},
     { 0x040A, {
@@ -4107,7 +4111,7 @@ void DrawNativeSignalIcon(HDC hdc, int right, int top, ULONG quality) {
     else if (quality > 0)  idx = 1;
     int iconSize = ScaleDpi(20);
     int xPos = right - iconSize - 4;
-    int yPos = top + (ScaleDpi(26) - iconSize) / 2;
+    int yPos = top + (ScaleDpi(30) - iconSize) / 2;  // ROW_HEIGHT_NORMAL_BASE=30 (van.dll)
     if (g_hIconSignalBars[idx]) {
         DrawIconBicubic(hdc, xPos, yPos, iconSize, iconSize,
                         g_hIconSignalBars[idx], &g_pBitmapSignalBars[idx]);
@@ -5087,7 +5091,7 @@ LRESULT CALLBACK FlyoutWndProc(HWND hwnd, UINT uMsg, WPARAM wParam, LPARAM lPara
                     BOOL isSelected = (i == g_SelectedRowIndex);
                     BOOL isHovered  = (i == g_HoveredRowIndex);
                     BOOL hasKeyboardFocus = (i == g_KeyboardSelectedIndex);
-                    
+
                     if (isSelected || isHovered) {
                         RECT rcFullRow = rcRow; rcFullRow.left = 0; rcFullRow.right = WINDOW_WIDTH - 5;
                         COLORREF bgColor = isSelected ? GetRowSelectedColor() : GetRowHoverColor();
@@ -5107,8 +5111,8 @@ LRESULT CALLBACK FlyoutWndProc(HWND hwnd, UINT uMsg, WPARAM wParam, LPARAM lPara
                     BOOL isConnected = (g_NetworkList[i].connState == CONN_STATE_CONNECTED);
                     SelectObject(hdc, isConnected ? g_hFontBold : g_hFontNormal);
                     SetTextColor(hdc, GetNetworkNameColor());
-                    DrawTextWithWrap(hdc, ssidBuf, rcRow.left - ScaleDpi(2), rcRow.top+6, 
-                                     rcRow.right - rcRow.left - 10, 18);       
+                    DrawTextWithWrap(hdc, ssidBuf, rcRow.left - ScaleDpi(2), rcRow.top + ScaleDpi(3),
+                                     rcRow.right - rcRow.left - 10, ScaleDpi(24));       
                     WifiNetworkItem* item = &g_NetworkList[i];
                     BOOL isTransitioning = (item->connState == CONN_STATE_CONNECTING ||
                                             item->connState == CONN_STATE_DISCONNECTING);
@@ -5156,7 +5160,7 @@ LRESULT CALLBACK FlyoutWndProc(HWND hwnd, UINT uMsg, WPARAM wParam, LPARAM lPara
         RECT rcClient;
         GetClientRect(hwnd, &rcClient);
         int footerTop = rcClient.bottom - FOOTER_HEIGHT;
-        int centerX = (rcClient.right - textSize.cx) / 2;
+                int centerX = (rcClient.right - textSize.cx) / 2;
         int footerTextYC = footerTop + (FOOTER_HEIGHT - textSize.cy) / 2;
         if (g_Settings.useRoundedCorners) {
             footerTextYC += (FOOTER_HEIGHT * 15) / 100;
--- a/mods/win7-network-flyout-recreation.wh.cpp
+++ b/mods/win7-network-flyout-recreation.wh.cpp
@@ -4924,8 +4924,7 @@ LRESULT CALLBACK FlyoutWndProc(HWND hwnd, UINT uMsg, WPARAM wParam, LPARAM lPara
         } else {
             hLargeIcon = g_hIconSignalBars[0];  // No connection - show empty signal bars
         }
-        if (hLargeIcon) DrawIconEx(hdc, ScaleDpi(12), ScaleDpi(37), hLargeIcon, iconSize, iconSize, 0, NULL, DI_NORMAL);
-
+if (hLargeIcon) DrawIconEx(hdc, ScaleDpi(12), ScaleDpi(37) - 3, hLargeIcon, iconSize, iconSize, 0, NULL, DI_NORMAL);
         if (showWifiList) {
             int totalHeight = GetTotalListHeight();
             int visibleHeight = LIST_Y_END - LIST_Y_START;
--- a/mods/win7-network-flyout-recreation.wh.cpp
+++ b/mods/win7-network-flyout-recreation.wh.cpp
@@ -2,13 +2,13 @@
 // @id             win7-network-flyout-recreation
 // @name           Windows 7 Network Flyout Recreation
 // @description    This mod recreates the Windows 7 network flyout for Windows 10 and 11 along with some more configurable restorations
-// @version        3.2.0
+// @version        3.4.0
 // @author         babamohammed
 // @github         https://github.com/babamohammed2022
 // @include        explorer.exe
 // @include        control.exe
 // @architecture   x86-64
-// @compilerOptions -DWIN32_LEAN_AND_MEAN -lgdi32 -ldwmapi -luxtheme -lole32 -lshell32 -luser32 -lcomctl32 -liphlpapi -lwlanapi -luuid -lshlwapi
+// @compilerOptions -DWIN32_LEAN_AND_MEAN -finput-charset=UTF-8 -lgdi32 -ldwmapi -luxtheme -lole32 -lshell32 -luser32 -lcomctl32 -liphlpapi -lwlanapi -luuid -lshlwapi
 // ==/WindhawkMod==
 
 // ==WindhawkModReadme==
@@ -19,7 +19,7 @@ This mod recreates the classic Windows 7 network flyout on Windows 10 and 11, re
 
 Screenshot of the light theme:
 
-![Screenshot](https://raw.githubusercontent.com/babamohammed2022/babamohammed2022/main/pic.PNG)
+![Screenshot](https://raw.githubusercontent.com/babamohammed2022/babamohammed2022/main/networkflyoutwin7.PNG)
 
 Screenshot of the dark theme:
 
@@ -36,7 +36,7 @@ The mod has been tested on Windows 10 21H2, Windows 10 1809, Windows 11 23H2, Wi
 - **Wi-Fi network list**: Shows all available networks with live signal strength
 - **Connect/Disconnect**: Connect to secured and open networks with password support
 - **Privacy mode**: Hide real network names (shows as Network 1, Network 2...)
-- **Windows 7 style tooltips**: Full network info on hover (SSID, signal, security type)
+- **Classic tooltips**: Full network info on hover (SSID, signal, security type)
 - **Right-click context menu**: Quick access to network status and properties
 - **Keyboard navigation**: Full Arrow keys, Enter, and Escape support
 - **Auto-refresh**: Periodically refreshes the network list at a configurable interval
@@ -46,6 +46,7 @@ The mod has been tested on Windows 10 21H2, Windows 10 1809, Windows 11 23H2, Wi
 - **Dual Theme Support**: Includes both light and dark themes, with the dark theme created specifically for late-night use and, if present, dark Aero theme.
 - **NEW: Ethernet Support**: The mod should now properly show the flyout for Ethernet connection.
 - **NEW: Classic Network Center links**: Optionally restores the Windows 7 “Connect to a network” and HomeGroup/sharing links with their custom artwork.
+- **Classic artwork refresh**: Uses cleaned embedded PNG artwork for the refresh button, chevrons, and classic Home/Public/Work network location icons.
 
 
 ## Requirements
@@ -65,9 +66,9 @@ The mod has been tested on Windows 10 21H2, Windows 10 1809, Windows 11 23H2, Wi
 | **Ctrl+H** | Toggle network flyout (disabled by default) |
 
 ## Credits
-- **m417z** — Code review
-- **Anixx** — Testing on Windows 11 23H2 and feedback
-- **sebastian08dm08-cpu** — Testing on Windows 10 1809
+- **m417z** - Code review
+- **Anixx** - Testing on Windows 11 23H2 and providing feedback
+- **sebastian08dm08-cpu** - Testing on Windows 10 1809
 
 If you encounter issues, please report them on the author of the mod.
 */
@@ -104,6 +105,9 @@ If you encounter issues, please report them on the author of the mod.
 - restoreClassicNetworkCenterLinks: true
   $name: Restore classic Network Center links
   $description: Add the Windows 7 “Connect to a network” and HomeGroup/sharing links to the ''Network and Sharing Center'' page in the Control Panel.
+- useNetworkLocationIcons: true
+  $name: Network location icons (Home / Public / Work)
+  $description: Show the classic Windows 7 network location icon in the flyout header based on the active network profile (house = Home, bench = Public, buildings = Work). Disable to restore the original generic network icon.
 - theme: light
   $name: Theme
   $description: Select the network flyout's theme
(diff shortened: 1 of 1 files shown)


babamohammed2022:
Hello @m417z, if possible, if you can, please let me know if the code has issues to address

m417z:
### Submission review

---

- **The new network-location detection runs on every refresh tick, even while the flyout is hidden.** The auto-refresh timer ([`WM_TIMER` 1000](https://github.com/ramensoftware/windhawk-mods/blob/580e83a8d46f142a3fc8d7581d9760aeb4b23cfa/mods/win7-network-flyout-recreation.wh.cpp#L4345-L4350), default every 3 s for the lifetime of explorer) calls `RefreshNetworkData`, which now also runs [`DetectNetworkLocationCategory()`](https://github.com/ramensoftware/windhawk-mods/blob/580e83a8d46f142a3fc8d7581d9760aeb4b23cfa/mods/win7-network-flyout-recreation.wh.cpp#L2929-L2941): a registry enumeration of all `NetworkList\Profiles`, and — whenever the name match fails — up to four NLM enumerations (out-of-process COM calls into the netprofm service) plus a full `GetAdaptersAddresses` pass per connection in the fallback scan. The result is only consumed at paint time, so all of this work is wasted while the flyout is not visible. Since `WM_SHOW_FLYOUT` already calls `RefreshNetworkData()` when the flyout opens, gating the detection on visibility is safe and cheap:

  ```cpp
  BOOL flyoutVisible = g_hWndFlyout && IsWindow(g_hWndFlyout) && IsWindowVisible(g_hWndFlyout);
  if (isAnyConnected && g_Settings.useNetworkLocationIcons && flyoutVisible) {
      g_CurrentNetworkCategory = DetectNetworkLocationCategory();
  }
  ```

  (Keeping the previous `g_CurrentNetworkCategory` instead of resetting it to `-1` while hidden would also avoid a generic-icon flash on the first paint after opening.) Constant background polling/enumeration in explorer is a recurring maintainer concern — please gate the work on relevance.

<p>

- **Dead GDI+ cache plumbing around `GetNetworkLocationIcon`.** The function takes a `void*** pppOutCache` out-parameter and its comment says the cache is passed to `DrawIconBicubic` "for high-quality scaling" — but the only caller [discards it and draws with `DrawIconEx`](https://github.com/ramensoftware/windhawk-mods/blob/580e83a8d46f142a3fc8d7581d9760aeb4b23cfa/mods/win7-network-flyout-recreation.wh.cpp#L4645-L4652). As a result, [`g_pBitmapNetLocHome/Public/Work` and `g_pBitmapNetworkMap`](https://github.com/ramensoftware/windhawk-mods/blob/580e83a8d46f142a3fc8d7581d9760aeb4b23cfa/mods/win7-network-flyout-recreation.wh.cpp#L1052-L1055) are declared and freed in two cleanup paths but never populated — dead code. Either actually draw the header icon through `DrawIconBicubic(hdc, ..., hLargeIcon, ppCache)` (which would also fix the slight blur noted in the functionality section below), or drop the four globals, the out-parameter, and their cleanup. If this plumbing is an AI leftover rather than intentional, it's worth trimming.
- **Em-dashes were stripped from log strings, leaving double spaces**, e.g. [`L"Win11: SysPager found but no ToolbarWindow32  partial legacy taskbar"`](https://github.com/ramensoftware/windhawk-mods/blob/580e83a8d46f142a3fc8d7581d9760aeb4b23cfa/mods/win7-network-flyout-recreation.wh.cpp#L952) and several similar `Wh_Log` calls. The README credits were converted from `—` to `-`; doing the same in these strings (rather than deleting the dash outright) would keep them readable.
- **Changelog doesn't cover the shipped version.** `@version` is `3.4.0`, but the changelog comment's newest entry is `3.3.3`. Also, the previously released `3.2.0`/`3.1.0` entries were deleted rather than appended to — keeping the history intact makes it easier to track what changed between merged versions.
- **Registry string reads:** `RegQueryValueExW` doesn't guarantee null termination for `REG_SZ` values; `TryGetCategoryFromRegistryProfileName` passes the buffer straight to `_wcsicmp`. In practice Windows writes these values terminated, but `RegGetValueW` (which guarantees termination) is the more robust choice here.
- **`IsAdapterGuidIgnoredByIpHelper` re-runs a full `GetAdaptersAddresses` for every connection** examined by the fallback scan. If you keep the fallback scan, fetching the adapter table once per scan and looking adapters up in it would avoid the repeated 15 KB alloc + enumeration.
- The `refreshLeftOffset = (WINDOW_WIDTH * 7) / 1000` expression is duplicated in [`InitRefreshButtonRect`](https://github.com/ramensoftware/windhawk-mods/blob/580e83a8d46f142a3fc8d7581d9760aeb4b23cfa/mods/win7-network-flyout-recreation.wh.cpp#L2217) and the [paint handler](https://github.com/ramensoftware/windhawk-mods/blob/580e83a8d46f142a3fc8d7581d9760aeb4b23cfa/mods/win7-network-flyout-recreation.wh.cpp#L4663); a shared helper/constant would keep hit-testing and painting from drifting apart.

</p>
</details>

<p>

- **Category detection matches registry profiles by display name, which isn't a reliable key.** Profile names aren't unique — a machine that has seen several networks named e.g. "Network" / "Network 2" can have stale profiles whose name equals the current NLM friendly name, and the first name match wins, potentially showing the wrong Home/Public/Work icon. A robust join that removes name matching entirely: for the active adapter, get its `INetwork` (via `GetNetworkConnections` + `GetAdapterId`), call `INetwork::GetNetworkId()`, and open `NetworkList\Profiles\{that GUID}` directly — the `Profiles` subkey names *are* the network GUIDs, so you get the exact `Category` for the exact network, with the same "registry wins over NLM" property you wanted from the 3.3.3 reordering.
- **Ordering nit in the same area:** when Ethernet and Wi-Fi are connected simultaneously and the Ethernet registry lookup misses, the Wi-Fi registry lookup runs *before* the Ethernet exact-adapter NLM check — so the icon can reflect the Wi-Fi network even though the header shows the Ethernet connection. The GUID-based join above would resolve this too.
- **Header icon is drawn slightly stretched.** The Home/Public/Work icons are decoded at `ScaleDpi(35)` but drawn at `iconSize = ScaleDpi(35*1.05)`, so `DrawIconEx` upscales them ~5% with low-quality stretching — a mild blur on the headline feature of this update. Decode at the draw size (or route through `DrawIconBicubic`, see the cache item above).
- **Refresh button hover size jump on high DPI.** The normal refresh icon is drawn at its natural 16 px (no DPI scaling, `DrawIconEx(..., 0, 0, ...)`), while the new hover icon is drawn at `ScaleDpi(22)`. At 96 DPI that's a 16→22 px jump (possibly intentional, since the hover art includes the button frame), but at 150% DPI the normal icon stays a tiny 16 px while the hover icon becomes 33 px. Consider decoding the normal icon at `ScaleDpi(16)` as well.

</p>
</details>

babamohammed2022:
> ### Submission review
> 
> 
> * **The new network-location detection runs on every refresh tick, even while the flyout is hidden.** The auto-refresh timer ([`WM_TIMER` 1000](https://github.com/ramensoftware/windhawk-mods/blob/580e83a8d46f142a3fc8d7581d9760aeb4b23cfa/mods/win7-network-flyout-recreation.wh.cpp#L4345-L4350), default every 3 s for the lifetime of explorer) calls `RefreshNetworkData`, which now also runs [`DetectNetworkLocationCategory()`](https://github.com/ramensoftware/windhawk-mods/blob/580e83a8d46f142a3fc8d7581d9760aeb4b23cfa/mods/win7-network-flyout-recreation.wh.cpp#L2929-L2941): a registry enumeration of all `NetworkList\Profiles`, and — whenever the name match fails — up to four NLM enumerations (out-of-process COM calls into the netprofm service) plus a full `GetAdaptersAddresses` pass per connection in the fallback scan. The result is only consumed at paint time, so all of this work is wasted while the flyout is not visible. Since `WM_SHOW_FLYOUT` already calls `RefreshNetworkData()` when the flyout opens, gating the detection on visibility is safe and cheap:
>   ```c++
>   BOOL flyoutVisible = g_hWndFlyout && IsWindow(g_hWndFlyout) && IsWindowVisible(g_hWndFlyout);
>   if (isAnyConnected && g_Settings.useNetworkLocationIcons && flyoutVisible) {
>       g_CurrentNetworkCategory = DetectNetworkLocationCategory();
>   }
>   ```
>   
>   
>       
>         
>       
>   
>         
>       
>   
>       
>     
>   (Keeping the previous `g_CurrentNetworkCategory` instead of resetting it to `-1` while hidden would also avoid a generic-icon flash on the first paint after opening.) Constant background polling/enumeration in explorer is a recurring maintainer concern — please gate the work on relevance.
> 
> Optional improvements
> 
> 
> * **Dead GDI+ cache plumbing around `GetNetworkLocationIcon`.** The function takes a `void*** pppOutCache` out-parameter and its comment says the cache is passed to `DrawIconBicubic` "for high-quality scaling" — but the only caller [discards it and draws with `DrawIconEx`](https://github.com/ramensoftware/windhawk-mods/blob/580e83a8d46f142a3fc8d7581d9760aeb4b23cfa/mods/win7-network-flyout-recreation.wh.cpp#L4645-L4652). As a result, [`g_pBitmapNetLocHome/Public/Work` and `g_pBitmapNetworkMap`](https://github.com/ramensoftware/windhawk-mods/blob/580e83a8d46f142a3fc8d7581d9760aeb4b23cfa/mods/win7-network-flyout-recreation.wh.cpp#L1052-L1055) are declared and freed in two cleanup paths but never populated — dead code. Either actually draw the header icon through `DrawIconBicubic(hdc, ..., hLargeIcon, ppCache)` (which would also fix the slight blur noted in the functionality section below), or drop the four globals, the out-parameter, and their cleanup. If this plumbing is an AI leftover rather than intentional, it's worth trimming.
> * **Em-dashes were stripped from log strings, leaving double spaces**, e.g. [`L"Win11: SysPager found but no ToolbarWindow32  partial legacy taskbar"`](https://github.com/ramensoftware/windhawk-mods/blob/580e83a8d46f142a3fc8d7581d9760aeb4b23cfa/mods/win7-network-flyout-recreation.wh.cpp#L952) and several similar `Wh_Log` calls. The README credits were converted from `—` to `-`; doing the same in these strings (rather than deleting the dash outright) would keep them readable.
> * **Changelog doesn't cover the shipped version.** `@version` is `3.4.0`, but the changelog comment's newest entry is `3.3.3`. Also, the previously released `3.2.0`/`3.1.0` entries were deleted rather than appended to — keeping the history intact makes it easier to track what changed between merged versions.
> * **Registry string reads:** `RegQueryValueExW` doesn't guarantee null termination for `REG_SZ` values; `TryGetCategoryFromRegistryProfileName` passes the buffer straight to `_wcsicmp`. In practice Windows writes these values terminated, but `RegGetValueW` (which guarantees termination) is the more robust choice here.
> * **`IsAdapterGuidIgnoredByIpHelper` re-runs a full `GetAdaptersAddresses` for every connection** examined by the fallback scan. If you keep the fallback scan, fetching the adapter table once per scan and looking adapters up in it would avoid the repeated 15 KB alloc + enumeration.
> * The `refreshLeftOffset = (WINDOW_WIDTH * 7) / 1000` expression is duplicated in [`InitRefreshButtonRect`](https://github.com/ramensoftware/windhawk-mods/blob/580e83a8d46f142a3fc8d7581d9760aeb4b23cfa/mods/win7-network-flyout-recreation.wh.cpp#L2217) and the [paint handler](https://github.com/ramensoftware/windhawk-mods/blob/580e83a8d46f142a3fc8d7581d9760aeb4b23cfa/mods/win7-network-flyout-recreation.wh.cpp#L4663); a shared helper/constant would keep hit-testing and painting from drifting apart.
> 
> Functionality notes

Thanks for the review. I've tried to address the reported issues. If necessary, let me know if there are any other issues to address. 
Note: the refresh button has been kept as it is since the hover image is slightly bigger for simplicity

m417z:
Just fix the two incorrect `HotkeyThreadProc` calls, then test it, and I'll merge it. You can also check the other notes.

### Submission review

---

1. **The new detection/priming calls run on threads that have no COM apartment — keep them on the flyout thread instead.** `CoInitializeEx` is only ever called in `HotkeyThreadProc` ([line 5755](https://github.com/ramensoftware/windhawk-mods/blob/95c77676a7bf9f30874908f5f899403222c5a403/mods/win7-network-flyout-recreation.wh.cpp#L5755)), yet this PR adds two direct calls elsewhere:
   - `Wh_ModInit` ([line 6292](https://github.com/ramensoftware/windhawk-mods/blob/95c77676a7bf9f30874908f5f899403222c5a403/mods/win7-network-flyout-recreation.wh.cpp#L6292)) calls `RefreshNetworkData(TRUE)` on the Windhawk init thread, before the hotkey thread even exists. On that thread `CoCreateInstance(CLSID_NetworkListManager, ...)` fails with `CO_E_NOTINITIALIZED`, so the NLM part of the "priming" silently no-ops (and the Wi-Fi list is empty since the WLAN handle is opened later on the hotkey thread) — while the parts that *do* run (`GetAdaptersAddresses` plus a full `NetworkList\Profiles` registry enumeration) execute synchronously during explorer startup, adding injection-time latency for little benefit. And if COM ever *is* already initialized on that thread, `g_pNLM` gets created in the wrong apartment for the STA hotkey thread that uses it afterwards.
   - `Wh_ModSettingsChanged` ([line 6319](https://github.com/ramensoftware/windhawk-mods/blob/95c77676a7bf9f30874908f5f899403222c5a403/mods/win7-network-flyout-recreation.wh.cpp#L6319)) calls `RefreshNetworkData(TRUE)` on a Windhawk callback thread. If `g_pNLM` already exists it was created on the `COINIT_APARTMENTTHREADED` hotkey thread, so this uses an STA-created interface pointer from a foreign thread without marshaling — a COM apartment-rules violation. It also introduces a second concurrent writer to the shared state (`g_EthernetConnected`/`g_EthernetNetworkName`, `g_NetworkList` via `RefreshWifiData`, and the two unsynchronized `g_pNLM` lazy-create sites, one of which leaks the losing instance) racing the flyout thread's auto-refresh timer and paint handler.

   The fix follows the pattern the mod already uses: before this PR, `Wh_ModSettingsChanged` marshaled refresh work to the flyout thread with `PostMessageW(g_hWndFlyout, WM_REFRESH_DATA, ...)` — do the same for the forced re-detection (e.g. a `wParam` force flag on `WM_REFRESH_DATA`), and move the startup priming to the top of `HotkeyThreadProc` right after `CoInitializeEx` instead of doing it in `Wh_ModInit`. That resolves the apartment issues, the races, and the startup latency in one move.

<p>

- **Dead cache plumbing and an inaccurate changelog entry.** `GetNetworkLocationIcon(void***)` returns a GDI+ cache slot that the only caller ignores ([lines 4918–4919](https://github.com/ramensoftware/windhawk-mods/blob/95c77676a7bf9f30874908f5f899403222c5a403/mods/win7-network-flyout-recreation.wh.cpp#L4918): `ppDummy` is never used), so `g_pBitmapNetLocHome/Public/Work` and `g_pBitmapNetworkMap` are only ever freed, never populated. The 3.4.0 changelog entry says the header icon is "drawn through the existing bicubic GDI+ path", but the paint code uses plain `DrawIconEx`. Either actually draw through `DrawIconBicubic` like the signal bars do (line 4116), or drop the cache parameter/globals and correct the changelog line. (Decoding at `ScaleDpi(35)` does remove the 5% upscale, so the visible fix is real either way.)
- **`InitPniduiInfo` path cascade is redundant.** `GetModuleHandleW` with a bare name matches a loaded module by base name regardless of the directory it was loaded from, so the final `GetModuleHandleW(L"pnidui.dll")` already covers all the AppData/ProgramData/System32 probes; the extra paths only matter if two differently-pathed `pnidui.dll` modules are loaded at once, which the original two-step lookup (ExplorerPatcher full path, then bare name) already handled. Also, the "mod's own directory" step ([line 4311](https://github.com/ramensoftware/windhawk-mods/blob/95c77676a7bf9f30874908f5f899403222c5a403/mods/win7-network-flyout-recreation.wh.cpp#L4311)) actually computes explorer.exe's directory (`GetModuleFileNameW(GetModuleHandle(NULL), ...)`), and the failure log "Try installing ExplorerPatcher" (line 4342) is odd advice to give users of a Windhawk mod. Suggest reverting to the original two lookups.
- **Include hygiene.** `<shlwapi.h>` is now included twice; `#include <shlobj.h>` carries trailing whitespace; the `CSIDL_LOCAL_APPDATA`/`CSIDL_COMMON_APPDATA` fallback `#define`s are dead — `shlobj.h` always defines them.
- **Magic layout numbers.** `DrawIconEx(hdc, ScaleDpi(12), ScaleDpi(37) - 3, ...)` ([line 4927](https://github.com/ramensoftware/windhawk-mods/blob/95c77676a7bf9f30874908f5f899403222c5a403/mods/win7-network-flyout-recreation.wh.cpp#L4927)) has an unscaled `-3`, and `int offsetY = (WINDOW_HEIGHT * 1.01) / 100;` (duplicated at lines 5112 and 5124) is floating-point math for what is effectively a 1% offset — a shared `ScaleDpi`-based constant would be clearer and DPI-consistent.
- **Changelog block.** The first entry ("van.dll alignment: …") has no version prefix and sits above the 3.4.0 entries, and the 3.3.x entries describe intermediate versions that were never published. Consider folding everything since 3.2.0 under the single 3.4.0 release users will actually receive.


- **Leftover AI-chat artifacts and mangled paste damage.** Several fragments of an AI conversation were pasted into the file as-is:
   - [Line 1085](https://github.com/ramensoftware/windhawk-mods/blob/95c77676a7bf9f30874908f5f899403222c5a403/mods/win7-network-flyout-recreation.wh.cpp#L1085): `// Aggiungi prima questa struttura (una volta sola, all'inizio del file)` and [line 1093](https://github.com/ramensoftware/windhawk-mods/blob/95c77676a7bf9f30874908f5f899403222c5a403/mods/win7-network-flyout-recreation.wh.cpp#L1093): `// Poi sostituisci la dichiarazione con:` — these are instructions *to* the author ("Add this struct first…", "Then replace the declaration with:"), not code comments, and they're in Italian (comments should be in English). Same for [line 4261](https://github.com/ramensoftware/windhawk-mods/blob/95c77676a7bf9f30874908f5f899403222c5a403/mods/win7-network-flyout-recreation.wh.cpp#L4261) `// Fallback per CSIDL se non definiti`.
   - The same paste mangled the code: [line 1094](https://github.com/ramensoftware/windhawk-mods/blob/95c77676a7bf9f30874908f5f899403222c5a403/mods/win7-network-flyout-recreation.wh.cpp#L1094) `static WinHandle g_hConnectMutex;static HMODULE g_hGdiPlus = NULL;` and [line 6281](https://github.com/ramensoftware/windhawk-mods/blob/95c77676a7bf9f30874908f5f899403222c5a403/mods/win7-network-flyout-recreation.wh.cpp#L6281) (`.reset(...)` and `g_uTaskbarCreated = ...` merged onto one line) put two statements on one line, and several blocks in `FlyoutWndProc` dropped to column-0 indentation ([line 4927](https://github.com/ramensoftware/windhawk-mods/blob/95c77676a7bf9f30874908f5f899403222c5a403/mods/win7-network-flyout-recreation.wh.cpp#L4927), lines 5110–5133), plus stray re-indents at lines 5163, 5845, 6296. Please clean these up — one statement per line, restore the surrounding indentation.

</p>
</details>

<p>

- **Cached icons don't survive a DPI change.** All the newly cached images — the Home/Public/Work icons (decoded at `ScaleDpi(35)`), the chevrons and refresh icons (`ScaleDpi(16)`/`ScaleDpi(22)`) — are decoded once at the DPI in effect on first use and never recreated: `RecalcDpiMetrics` doesn't free them, and `FreeSystemIcons` only runs at unload. Move the flyout to a monitor with a different DPI and `DrawIconEx` stretches them again — the same blur the 3.4.0 changelog entry fixes for the first monitor, and the README advertises mixed-DPI support. Consider destroying the cached icons/bitmaps in `RecalcDpiMetrics` so they're re-decoded at the new DPI on next use.
- **First-open priming via the Ethernet name fallback rarely hits.** When NLM is unavailable (as it is on the `Wh_ModInit` path today), `g_EthernetNetworkName` falls back to the adapter's friendly name (e.g. "Ethernet"), which usually isn't a `NetworkList` profile name, so `TryGetCategoryFromRegistryProfileName` misses and the generic icon can still appear on the very first open. Moving the priming to the COM-initialized hotkey thread (item 2 above) also fixes this, since the exact-GUID join will then work.

</p>
</details>

babamohammed2022:
> Just fix the two incorrect `HotkeyThreadProc` calls, then test it, and I'll merge it. You can also check the other notes.
> 
> ### Submission review
> 
> 
> 1. **The new detection/priming calls run on threads that have no COM apartment — keep them on the flyout thread instead.** `CoInitializeEx` is only ever called in `HotkeyThreadProc` ([line 5755](https://github.com/ramensoftware/windhawk-mods/blob/95c77676a7bf9f30874908f5f899403222c5a403/mods/win7-network-flyout-recreation.wh.cpp#L5755)), yet this PR adds two direct calls elsewhere:
>    
>    * `Wh_ModInit` ([line 6292](https://github.com/ramensoftware/windhawk-mods/blob/95c77676a7bf9f30874908f5f899403222c5a403/mods/win7-network-flyout-recreation.wh.cpp#L6292)) calls `RefreshNetworkData(TRUE)` on the Windhawk init thread, before the hotkey thread even exists. On that thread `CoCreateInstance(CLSID_NetworkListManager, ...)` fails with `CO_E_NOTINITIALIZED`, so the NLM part of the "priming" silently no-ops (and the Wi-Fi list is empty since the WLAN handle is opened later on the hotkey thread) — while the parts that _do_ run (`GetAdaptersAddresses` plus a full `NetworkList\Profiles` registry enumeration) execute synchronously during explorer startup, adding injection-time latency for little benefit. And if COM ever _is_ already initialized on that thread, `g_pNLM` gets created in the wrong apartment for the STA hotkey thread that uses it afterwards.
>    * `Wh_ModSettingsChanged` ([line 6319](https://github.com/ramensoftware/windhawk-mods/blob/95c77676a7bf9f30874908f5f899403222c5a403/mods/win7-network-flyout-recreation.wh.cpp#L6319)) calls `RefreshNetworkData(TRUE)` on a Windhawk callback thread. If `g_pNLM` already exists it was created on the `COINIT_APARTMENTTHREADED` hotkey thread, so this uses an STA-created interface pointer from a foreign thread without marshaling — a COM apartment-rules violation. It also introduces a second concurrent writer to the shared state (`g_EthernetConnected`/`g_EthernetNetworkName`, `g_NetworkList` via `RefreshWifiData`, and the two unsynchronized `g_pNLM` lazy-create sites, one of which leaks the losing instance) racing the flyout thread's auto-refresh timer and paint handler.
>    
>    The fix follows the pattern the mod already uses: before this PR, `Wh_ModSettingsChanged` marshaled refresh work to the flyout thread with `PostMessageW(g_hWndFlyout, WM_REFRESH_DATA, ...)` — do the same for the forced re-detection (e.g. a `wParam` force flag on `WM_REFRESH_DATA`), and move the startup priming to the top of `HotkeyThreadProc` right after `CoInitializeEx` instead of doing it in `Wh_ModInit`. That resolves the apartment issues, the races, and the startup latency in one move.
> 
> Optional improvements
> 
> 
> * **Dead cache plumbing and an inaccurate changelog entry.** `GetNetworkLocationIcon(void***)` returns a GDI+ cache slot that the only caller ignores ([lines 4918–4919](https://github.com/ramensoftware/windhawk-mods/blob/95c77676a7bf9f30874908f5f899403222c5a403/mods/win7-network-flyout-recreation.wh.cpp#L4918): `ppDummy` is never used), so `g_pBitmapNetLocHome/Public/Work` and `g_pBitmapNetworkMap` are only ever freed, never populated. The 3.4.0 changelog entry says the header icon is "drawn through the existing bicubic GDI+ path", but the paint code uses plain `DrawIconEx`. Either actually draw through `DrawIconBicubic` like the signal bars do (line 4116), or drop the cache parameter/globals and correct the changelog line. (Decoding at `ScaleDpi(35)` does remove the 5% upscale, so the visible fix is real either way.)
> * **`InitPniduiInfo` path cascade is redundant.** `GetModuleHandleW` with a bare name matches a loaded module by base name regardless of the directory it was loaded from, so the final `GetModuleHandleW(L"pnidui.dll")` already covers all the AppData/ProgramData/System32 probes; the extra paths only matter if two differently-pathed `pnidui.dll` modules are loaded at once, which the original two-step lookup (ExplorerPatcher full path, then bare name) already handled. Also, the "mod's own directory" step ([line 4311](https://github.com/ramensoftware/windhawk-mods/blob/95c77676a7bf9f30874908f5f899403222c5a403/mods/win7-network-flyout-recreation.wh.cpp#L4311)) actually computes explorer.exe's directory (`GetModuleFileNameW(GetModuleHandle(NULL), ...)`), and the failure log "Try installing ExplorerPatcher" (line 4342) is odd advice to give users of a Windhawk mod. Suggest reverting to the original two lookups.
> * **Include hygiene.** `<shlwapi.h>` is now included twice; `#include <shlobj.h>` carries trailing whitespace; the `CSIDL_LOCAL_APPDATA`/`CSIDL_COMMON_APPDATA` fallback `#define`s are dead — `shlobj.h` always defines them.
> * **Magic layout numbers.** `DrawIconEx(hdc, ScaleDpi(12), ScaleDpi(37) - 3, ...)` ([line 4927](https://github.com/ramensoftware/windhawk-mods/blob/95c77676a7bf9f30874908f5f899403222c5a403/mods/win7-network-flyout-recreation.wh.cpp#L4927)) has an unscaled `-3`, and `int offsetY = (WINDOW_HEIGHT * 1.01) / 100;` (duplicated at lines 5112 and 5124) is floating-point math for what is effectively a 1% offset — a shared `ScaleDpi`-based constant would be clearer and DPI-consistent.
> * **Changelog block.** The first entry ("van.dll alignment: …") has no version prefix and sits above the 3.4.0 entries, and the 3.3.x entries describe intermediate versions that were never published. Consider folding everything since 3.2.0 under the single 3.4.0 release users will actually receive.
> * **Leftover AI-chat artifacts and mangled paste damage.** Several fragments of an AI conversation were pasted into the file as-is:
>   
>   * [Line 1085](https://github.com/ramensoftware/windhawk-mods/blob/95c77676a7bf9f30874908f5f899403222c5a403/mods/win7-network-flyout-recreation.wh.cpp#L1085): `// Aggiungi prima questa struttura (una volta sola, all'inizio del file)` and [line 1093](https://github.com/ramensoftware/windhawk-mods/blob/95c77676a7bf9f30874908f5f899403222c5a403/mods/win7-network-flyout-recreation.wh.cpp#L1093): `// Poi sostituisci la dichiarazione con:` — these are instructions _to_ the author ("Add this struct first…", "Then replace the declaration with:"), not code comments, and they're in Italian (comments should be in English). Same for [line 4261](https://github.com/ramensoftware/windhawk-mods/blob/95c77676a7bf9f30874908f5f899403222c5a403/mods/win7-network-flyout-recreation.wh.cpp#L4261) `// Fallback per CSIDL se non definiti`.
>   * The same paste mangled the code: [line 1094](https://github.com/ramensoftware/windhawk-mods/blob/95c77676a7bf9f30874908f5f899403222c5a403/mods/win7-network-flyout-recreation.wh.cpp#L1094) `static WinHandle g_hConnectMutex;static HMODULE g_hGdiPlus = NULL;` and [line 6281](https://github.com/ramensoftware/windhawk-mods/blob/95c77676a7bf9f30874908f5f899403222c5a403/mods/win7-network-flyout-recreation.wh.cpp#L6281) (`.reset(...)` and `g_uTaskbarCreated = ...` merged onto one line) put two statements on one line, and several blocks in `FlyoutWndProc` dropped to column-0 indentation ([line 4927](https://github.com/ramensoftware/windhawk-mods/blob/95c77676a7bf9f30874908f5f899403222c5a403/mods/win7-network-flyout-recreation.wh.cpp#L4927), lines 5110–5133), plus stray re-indents at lines 5163, 5845, 6296. Please clean these up — one statement per line, restore the surrounding indentation.
> 
> Functionality notes

I've addressed all the required issues and most of the optional ones except for the Ethernet one as I currently do not possess a dedicated cable to test if a user has an issue with it, I think that it will be reported anyways and the cache issue has been addressed partially, I tried the bicubic draw suggestion but it corrupted the icon artwork, so I reverted that part to avoid regression

PR #4870, opened 2026-07-22
Update vlc-discord-rpc to v1.2.0

ciizerr:
## ✨ Added

- Added a **multi-engine metadata scraper** for more accurate movie and TV information.
- Added **TVMaze support** for TV shows, providing faster lookups and improving metadata accuracy.
- Added richer metadata, including:
  - Official title
  - Release year
  - ⭐ Rating
  - Genres
  - Runtime
  - High-quality poster artwork
- Added a **30-day local metadata cache** to speed up repeat lookups and reduce unnecessary network requests.
- Reorganized mod settings into clear categories for easier navigation.

## 🚀 Improved

- Improved metadata lookup speed with faster loading times.
- Improved title matching for files containing scene names, quality tags, release groups, and other unnecessary text.
- Improved overall metadata accuracy for both movies and TV shows.
- Improved reliability when fetching metadata and artwork.
- Automatically removes expired cached metadata to keep the local cache clean and up to date.
- Improved compatibility with existing settings so previous configurations continue to work after updating.
- Improved overall stability when enabling, disabling, or reloading the mod.

## 🛠️ Fixed

- Fixed incorrect posters being displayed for movies that are part of a collection (box sets). The correct movie poster is now shown.
- Fixed metadata matching issues caused by noisy or poorly formatted filenames.
- Fixed cases where title cleaning could accidentally remove parts of valid movie or TV show names.
- Fixed duplicate metadata notifications in some situations.

> **Note:** This is the **final release supporting Windhawk v1.x.x**.


m417z:
### Submission review

Some of the items might be opinionated (like 5), let me know if you have a preference to leave some of them as is.

---

1. **Detached threads keep running mod code after the mod is unloaded — crash in vlc.exe.** The scraper thread ([line 2242](https://github.com/ramensoftware/windhawk-mods/blob/ac6104d71c4a9b06fd5b9c3edf005fca55626880/mods/vlc-discord-rpc.wh.cpp#L2242)), the Uguu upload thread ([line 2343](https://github.com/ramensoftware/windhawk-mods/blob/ac6104d71c4a9b06fd5b9c3edf005fca55626880/mods/vlc-discord-rpc.wh.cpp#L2343)), the two artwork-search threads ([line 2354](https://github.com/ramensoftware/windhawk-mods/blob/ac6104d71c4a9b06fd5b9c3edf005fca55626880/mods/vlc-discord-rpc.wh.cpp#L2354), [line 2386](https://github.com/ramensoftware/windhawk-mods/blob/ac6104d71c4a9b06fd5b9c3edf005fca55626880/mods/vlc-discord-rpc.wh.cpp#L2386)) and the toast thread ([line 1657](https://github.com/ramensoftware/windhawk-mods/blob/ac6104d71c4a9b06fd5b9c3edf005fca55626880/mods/vlc-discord-rpc.wh.cpp#L1657)) are all `.detach()`ed and never joined. `Wh_ModUninit` joins only `g_workerThread`, then Windhawk unmaps the mod DLL — any detached thread still inside a fetch is then executing unmapped code, which crashes VLC. The window is not theoretical: a scraper run is several sequential HTTPS requests with 3-second timeouts each, and the toast thread pumps messages for 5 seconds — disabling or updating the mod right after a track change lands squarely in it. The threads also use resources that `Wh_ModUninit` tears down underneath them: `g_hWinHttpSession` is closed (and can even be re-created by a racing `GetSharedWinHttpSession()` call after the close), and the toast thread uses GDI+ after `GdiplusShutdown`.
   Fix: don't detach. Keep the threads in a mutex-guarded `std::vector<std::thread>` (or use a pending-thread counter + event), have them check `g_stopThread` between network calls, and join them all in `Wh_ModUninit` before closing the WinHttp session and shutting down GDI+.

2. **Global `std::thread g_workerThread` aborts VLC on every normal exit.** When the process itself exits, `Wh_ModUninit` is *not* called — the OS terminates the other threads and then runs the mod's global destructors. At that point `g_workerThread` ([line 255](https://github.com/ramensoftware/windhawk-mods/blob/ac6104d71c4a9b06fd5b9c3edf005fca55626880/mods/vlc-discord-rpc.wh.cpp#L255)) is still `joinable()`, so `~std::thread()` calls `std::terminate()` and vlc.exe aborts during shutdown. Use the standard pattern:

   ```cpp
   // Joined and reset in Wh_ModUninit; must not run its destructor at process exit.
   [[clang::no_destroy]] std::optional<std::thread> g_workerThread;

   BOOL Wh_ModInit() {
       ...
       g_workerThread.emplace(Worker);
       return TRUE;
   }

   void Wh_ModUninit() {
       g_stopThread = true;
       if (g_workerThread && g_workerThread->joinable()) g_workerThread->join();
       g_workerThread.reset();
       ...
   }
   ```

3. **The legacy-settings migration layer is partly dead code and partly harmful.**
   - The string fallbacks never execute: `Wh_GetStringSetting` never returns NULL — it returns an empty string on error (see the doc comment in `windhawk_api.h`). So every `PCWSTR s = Wh_GetStringSetting(L"Group.Name"); if (!s) s = Wh_GetStringSetting(L"Name");` pair ([lines 1784–1871](https://github.com/ramensoftware/windhawk-mods/blob/ac6104d71c4a9b06fd5b9c3edf005fca55626880/mods/vlc-discord-rpc.wh.cpp#L1784-L1871)) always takes the first read, and old `ClientId`/`Theme`/`Provider`/`CustomUrl`/`ButtonLabel`/`CustomJunkWords`/`MusicActivityName` values do not migrate at all.
   - The boolean pattern `if (!bX && Wh_GetIntSetting(L"OldX")) bX = true;` can only ever force a setting *on*: for as long as an old `true` value is still readable from storage, unchecking the new grouped setting has no effect — a user upgrading from v1.1.x can end up unable to disable Cover Art, Quality Tags, etc.
   - The `TvMazeGroup.*` fallback reads ([lines 1804–1813](https://github.com/ramensoftware/windhawk-mods/blob/ac6104d71c4a9b06fd5b9c3edf005fca55626880/mods/vlc-discord-rpc.wh.cpp#L1804-L1813)) reference a settings schema that never existed in any published version of the mod — dead code.

   Recommendation: drop the migration layer entirely and read only the names declared in the current settings block. Settings resetting to defaults on a schema reorganization is normal for mod updates, and far preferable to settings that can't be turned off.

4. **The VLC HTTP password is logged in plaintext.** [Line 1893](https://github.com/ramensoftware/windhawk-mods/blob/ac6104d71c4a9b06fd5b9c3edf005fca55626880/mods/vlc-discord-rpc.wh.cpp#L1893) logs `Password=%S`, and the README's troubleshooting section explicitly tells users to consult (and, in practice, share) the mod logs. Log only whether a password was found; the `rawPassword` out-parameter of `ReadVlcConfig` exists solely for this log line and can be removed.

5. **Bing artwork search explicitly disables SafeSearch.** The fallback poster search uses `&adlt=off` ([line 783](https://github.com/ramensoftware/windhawk-mods/blob/ac6104d71c4a9b06fd5b9c3edf005fca55626880/mods/vlc-discord-rpc.wh.cpp#L783)) and blindly takes the first image result. A wrong title match can put an explicit image on the user's public Discord profile (and in the toast popup). Use `adlt=strict` — posters and album covers don't need the filter off.

<p>

- **`metadata_cache.txt` grows without bound.** `SaveMetadataToDisk` only appends; expired lines are skipped on read but never deleted, and every refetch after expiry appends another record for the same key. Consider rewriting the file (dropping expired/duplicate lines) periodically or on load. Also, the `|||` record delimiter isn't escaped — a filename-derived cache key containing `|` sequences would corrupt the record.
- The `[VLC-RPC]` prefix in every `Wh_Log` call is unnecessary — Windhawk already attributes log lines to the mod.
- `#pragma comment(lib, "shell32.lib")` (line 248) is redundant with `-lshell32` already in `@compilerOptions`.
- The raw `Wh_GetStringSetting` + `Wh_FreeStringSetting` pairs would be simpler with the RAII `WindhawkUtils::StringSetting` from `windhawk_utils.h`.

</p>
</details>

<p>

- **Embedded shared TMDB API key**: the base64 "obfuscation" is trivially reversible in an open-source mod, so the community key is effectively public. All users share its rate limit, and if TMDB revokes it for abuse, the scraper silently breaks for everyone. Worth being aware that this is fragile (and check TMDB's terms about key sharing).
- **Port probing treats any 401 as "VLC found"**: a 401 response sets `requestSuccess = true`, so a non-VLC local service that requires auth on a candidate port (dev servers on 8080 are common) permanently stops the scan from ever reaching VLC on a later fallback port. Consider also matching something VLC-specific (e.g. the `WWW-Authenticate` realm) before locking onto a port. Related: the probe sends the user's VLC Basic-auth credentials to whatever is listening on each candidate port — localhost-only, but worth keeping in mind.
- **Substring-based junk-word filtering can mangle legitimate titles**: the word list is matched as raw substrings, so e.g. `psa` matches inside "Upsala" and `yts` inside "Analytics". Matching on word boundaries (surrounded by separators) would avoid this.
- **Toast timing**: the toast only fires when a status update is pushed to Discord. With cover art and both scrapers disabled, nothing changes after the initial update, so a new track's toast waits for the ~30-iteration heartbeat instead of the intended ~3 seconds.
- **Uguu retention**: uguu.se advertises short retention (their site says hours, the changelog says 24 h) — an uploaded cover URL can die mid-session on long playlists. Discord's media proxy usually caches it, but the changelog claim is worth double-checking.

</p>
</details>

ciizerr:
Thank you for the detailed review! I've addressed the feedback in this update:

1. **Thread crashes:** Removed all detached threads. Network requests and toast notifications are now stored in a `g_asyncThreads` vector and are properly joined during `Wh_ModUninit()`.
2. **Port probing:** Added logic to inspect the `WWW-Authenticate` header to verify that the 401 response is actually coming from VLC before reporting a password error.
3. **Memory leaks:** Replaced all manual string setting allocations. *(Note: `WindhawkUtils::StringSetting` caused immediate startup crashes during my testing, so I implemented a custom `GetStringSettingSafe` wrapper that correctly frees the allocated memory without triggering those crashes.)*
4. **Junk filtering:** Added word boundary checks to prevent the cleaner from accidentally truncating valid titles.
5. **Security & logging:** Updated Bing searches to use `adlt=strict`, removed the plaintext VLC password from logs, and obfuscated the TMDB API key in logged URLs.
6. **TMDB API key:** I intentionally obfuscated the default TMDB API key to provide a basic layer of protection against automated scraping. I also added a setting that allows users to provide their own TMDB API key if the default key ever becomes rate-limited.

Please let me know if there's anything else that should be adjusted or improved. Thanks again for taking the time to review the code!

m417z:
### Submission review

---

1. **The metadata scraper thread is still detached — crash risk on disable/update.** The changelog says all background threads are now tracked and joined, but the TMDB/TVMaze scraper thread is still `.detach()`ed (line 2304). It performs several HTTP round-trips (each with a 3-second timeout, so potentially 10–20 seconds total) and touches `g_cacheMutex`, `g_tvMetadataCache`, `g_imageCache`, `g_tvFetchInProgress`, and the shared WinHttp session. On disable/update, `Wh_ModUninit` joins the worker and the tracked threads, closes `g_hWinHttpSession`, and then the DLL is unmapped — while the detached thread may still be running, which means it uses a closed session handle and ultimately executes unmapped code, crashing VLC. Track it in `g_asyncThreads` exactly like the upload/artwork/toast threads.

2. **The "Custom VLC Port" setting has no effect.** The settings block declares `AdvancedGroup.CustomPort` (line 219), but the code reads `Wh_GetIntSetting(L"AdvancedGroup.CustomVlcPort")` (line 1902) — an undeclared key, which always returns 0. So the custom-port escape hatch the README's troubleshooting section tells users to use silently does nothing. Fix the key name (and note the related dead variable `bUseCustomPort`, which is set but never used).

3. **`g_asyncThreads` grows without bound during normal playback.** Every toast, artwork search, and cover upload appends a `std::thread` that is never reaped until unload — finished threads stay `joinable()`, keeping their kernel thread handles open for the entire VLC session. A long listening session with many track changes accumulates hundreds of open handles. Reap completed threads when appending a new one, e.g. pair each thread with a done-flag:

   ```cpp
   struct AsyncTask {
       std::thread thread;
       std::shared_ptr<std::atomic<bool>> done;
   };
   [[clang::no_destroy]] std::optional<std::vector<AsyncTask>> g_asyncTasks{std::in_place};

   template <typename F>
   void RunAsync(F&& f) {
       auto done = std::make_shared<std::atomic<bool>>(false);
       std::thread t([f = std::forward<F>(f), done]() mutable {
           f();
           done->store(true);
       });
       std::lock_guard<std::mutex> lock(g_asyncMutex);
       for (auto it = g_asyncTasks->begin(); it != g_asyncTasks->end();) {
           if (it->done->load()) {
               it->thread.join();
               it = g_asyncTasks->erase(it);
           } else {
               ++it;
           }
       }
       g_asyncTasks->push_back({std::move(t), std::move(done)});
   }
   ```

   Using this helper for the scraper thread also fixes item 1.

4. **Moving all settings into groups silently resets every existing user's configuration.** Settings values are stored under the key names the code reads, and those all changed from flat (`ShowCoverArt`, `CustomPort`, `ClientId`, `Theme`, …) to dotted group names (`DisplayGroup.ShowCoverArt`, …). After the update, previously stored values are simply never read again — anyone with a custom port, custom Discord application ID, custom TMDB key, dark theme, etc. gets reverted to defaults. This directly contradicts the changelog claim that "previous configurations continue to work after updating." Either keep the flat key names, or if you accept the reset, correct the changelog so users know to re-apply their settings.

5. `[[clang::no_destroy]] std::vector<std::thread> g_asyncThreads` (line 257) uses the bare attribute on a container. For container types the convention is `[[clang::no_destroy]] std::optional<std::vector<...>>`, released with `reset()` in `Wh_ModUninit` — `.clear()` keeps the heap buffer alive, which leaks it on every mod unload/reload. (The `g_workerThread` declaration right above it already follows the correct pattern.)

<p>

- The port scan sends the VLC HTTP password (Basic `Authorization` header, line 1963) to every candidate localhost port before knowing the responder is VLC — the new `WWW-Authenticate: ... VLC` check only gates the 401 handling, after credentials were already transmitted. If another local app happens to listen on 8080/8081/9090, it receives the user's VLC password. Low severity (localhost-only), but consider probing without the header first and only authenticating once the responder is identified as VLC.
- `displayImage` — a URL scraped from TMDB/TVMaze/Bing/uguu responses — is concatenated into the Discord IPC JSON without `SanitizeString` (line 2503), unlike the text fields around it. A stray quote in a scraped URL would corrupt the payload; sanitize or validate it like the rest.
- `GetAudioLanguages` (lines 1474–1476): `lang.substr(0, 2)` can return a single character, and `shortLang[1] -= 32` then writes to `s[s.size()]`, which is undefined behavior. Guard on the length.
- `Wh_ModUninit` can block for ~10–20 seconds joining threads that are mid-network-request (several sequential fetches at 3-second timeouts each). Checking `g_stopThread` between the HTTP calls in the scraper/artwork paths would keep disable/update snappy.
- Unneeded/missing includes: `#include <windhawk_utils.h>` is included but nothing from `WindhawkUtils` is used; conversely `std::chrono` (line 2449) and `time()`/`difftime()` rely on transitive includes — add `<chrono>` and `<ctime>` explicitly.
- In the artwork-fallback path the worker acquires `g_asyncMutex` and constructs a thread while still holding `g_cacheMutex` (lines 2423–2441). It happens to be safe today (only the worker ever holds both), but the nesting is fragile — copy what you need, close the `g_cacheMutex` scope, then spawn/register the thread, like the other spawn sites do.

</p>
</details>

<p>

- The Discord IPC client only ever writes to the pipe and never reads responses (handshake replies, ACKs, pings). Discord tolerates this in practice, but if its outbound pipe buffer ever fills, writes on Discord's side can stall; draining the pipe occasionally (a non-blocking `PeekNamedPipe`/`ReadFile`) would make the client better-behaved.

</p>
</details>

ciizerr:
I've addressed the remaining issues:

* Fixed all settings reads to use the correct grouped setting paths. This also fixes the **Custom Port** setting, which wasn't being applied.
* Improved mod shutdown so disabling or reloading is much faster and no longer waits for pending network requests.
* Reverted the VLC port detection to the simpler and more reliable authentication flow used in **v1.1.5**, fixing an issue where the mod could keep scanning ports even when VLC was already running.
* Tested everything locally to confirm the fixes are working as expected.

Thanks again for taking the time to review everything!

PR #4871, opened 2026-07-22
Add virtual-desktop-popups-disable

Amrsatrio:
## Changelog

Initial release.

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/virtual-desktop-popups-disable.wh.cpp
+++ b/mods/virtual-desktop-popups-disable.wh.cpp
@@ -7,11 +7,15 @@
 // @github          https://github.com/Amrsatrio
 // @twitter         https://twitter.com/amrsatrio
 // @include         explorer.exe
+// @architecture    x86-64
 // ==/WindhawkMod==
 
 // ==WindhawkModReadme==
 /*
 # Disable Virtual Desktop Popups
+
+![Screenshot of Desktop 1 popup with the popup crossed out](https://i.imgur.com/uena3MS.png)
+
 This simple mod disables the popups saying "Desktop N" that appear when the current virtual desktop is changed on Windows 11 build 226xx.2050+ and 23493+.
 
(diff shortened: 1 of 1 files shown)


m417z:
Some minor nice-to-have notes below.

### Submission review

---

- **Add a screenshot to the README**: since the mod's whole purpose is to remove the "Desktop N" popup, a screenshot of the popup it disables would help users browsing the catalog recognize what this mod is about. Allowed image hosts are `i.imgur.com` and `raw.githubusercontent.com`.

- **Missing null check on the module handle**: if `LoadLibraryExW(L"twinui.dll", ...)` ever fails, `HookSymbols` is called with a null `HMODULE`, and the failure path becomes slower and less clear than it needs to be. A fail-fast check with its own log message is cheap:

  ```cpp
  HMODULE hTwinUI = LoadLibraryExW(L"twinui.dll", nullptr, LOAD_LIBRARY_SEARCH_SYSTEM32);
  if (!hTwinUI) {
      Wh_Log(L"Failed to load twinui.dll");
      return FALSE;
  }
  ```

- **Misleading variable name**: `hTwinUIPCShell` holds a handle to `twinui.dll`, not `twinui.pcshell.dll` — looks like a leftover from code targeting the latter (cf. [virtual-desktop-hotkey-cycle.wh.cpp](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/virtual-desktop-hotkey-cycle.wh.cpp#L92), where the same name matches the actual module). Rename to `hTwinUI`.

- **Consider `@architecture x86-64`**: the mod targets Windows 11 only, which has no 32-bit builds, so the x86 compilation target is dead weight. Declaring `x86-64` (which also covers ARM64, where you tested) would let you drop the `SSTDCALL` macro and write `__cdecl` directly in the symbol strings.

PR #4881, opened 2026-07-23
Add Genie + Friends minimize animation pack

akilluminati47:
GPU-accelerated minimize/restore animation pack featuring a true mesh-warped Genie (Magic Lamp) bend plus nine additional effects: Vacuum, Glide, Pop, Slide, Free Fall, Warp, Squash, Roll-Up, and Swirl. Each effect has an independent duration slider and a master enable switch.

Rendering uses DirectComposition -- each frame pushes only a transform + opacity, or for Genie, a per-frame HLSL-warped tessellated mesh. This keeps performance smooth on high-refresh displays (120+ Hz) with minimal CPU usage. When the GPU/shader path cannot initialize in a given process, the mod falls back silently to the original GDI renderer so no window is left without animation. Multi-monitor setups are fully supported via virtual-screen rendering.

## Changelog

* **Double-animation guard:** per-window `g_inSysCmd` guard so `ShowWindow_Hook` skips animation when nested inside `DefWindowProcW_Hook`'s `SC_MINIMIZE`/`SC_RESTORE` handling, preventing a grey frame flash.
* **DwmFlush before ShowWindow:** first-frame `DwmFlush()` before `ShowWindow()` in all three render paths (CPU, GPU, Genie GPU).
* **Initial Commit + DwmFlush in GPU setup:** `dcompDevice->Commit()` + `DwmFlush()` after visual-tree setup to prevent a blank first frame.
* **Restore cloak:** `DWMWA_CLOAK` set before restoring the real window; `wasCloaked` field in `GhostAnimData`; uncloaked in `FinalizeRealWindow`.
* **Eager GPU init:** `EnsureGpuDevice()` + `EnsureGenieGpuResources()` called in `Wh_ModInit()` to avoid first-frame stall.
* **Ghost window background:** `SetClassLongPtr(hGhost, GCLP_HBRBACKGROUND, GetStockObject(NULL_BRUSH))` for both GPU and CPU path to suppress WM_ERASEBKGND flashes.
* **Target locks to taskbar icon centre (not cursor):** `GetTaskbarIconCenter` function finds the exact taskbar button rect via `ToolbarWindow32` → `TB_BUTTONCOUNT` / `TB_GETBUTTON` / `TB_GETRECT`, matching by `dwData == HWND`. The horizontal centre of the icon button is used as the animation X target; the vertical target stays at the top edge of the taskbar (`rcWork.bottom`). Removed stale cursor-tracking cache.
* **Animation shrinks to icon button width:** `GetTaskbarIconCenter` now also returns the full clickable button width. For Genie/Vacuum/Warp/Swirl the final `scaleX` is clamped to `iconButtonWidth / windowWidth`, and the Genie mesh `neckW` uses the icon width directly, so the ribbon contracts to exactly the taskbar icon block instead of a hardcoded 3-5% of the window size.

## Mod authorship

This mod was created by:

### DIFF
--- a/mods/genie-and-friends-minimize-animation.wh.cpp
+++ b/mods/genie-and-friends-minimize-animation.wh.cpp
@@ -1,9 +1,9 @@
 // ==WindhawkMod==
-// @id              genie-minimize-animation-fork
+// @id              genie-and-friends-minimize-animation
 // @name            Genie + Friends minimize animation pack
-// @description     GPU-accelerated macOS/Compiz-style minimize & restore effects — Genie (true mesh bend), Vacuum, Glide, Pop, Slide, Free Fall, Warp, Squash, Roll-Up & Swirl.
-// @version         2.2.1
-// @author          lolstijl & akilluminati47
+// @description     GPU-accelerated macOS/Compiz-style minimize & restore effects: Genie (true mesh bend), Vacuum, Glide, Pop, Slide, Free Fall, Warp, Squash, Roll-Up & Swirl.
+// @version         2.2.2
+// @author          akilluminati47
 // @github          https://github.com/akilluminati47
 // @include         *
 // @compilerOptions -ldwmapi -lgdi32 -ld3d11 -ldxgi -ldcomp -ld3dcompiler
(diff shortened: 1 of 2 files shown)


m417z:
### Submission review

---

**1. This would be the third minimize-animation mod in the catalog - please consolidate instead of forking.** The repo already has [genie-minimize-animation](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/genie-minimize-animation.wh.cpp) (lolstijl's mod, no longer maintained, which this forks) and [macos-minimize-animation](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/macos-minimize-animation.wh.cpp), which already ships a GPU mesh-warp genie engine (Direct2D), taskbar-button targeting, multi-hook minimize/restore coverage, and safe unload. The strong preference is to *extend* an existing mod rather than merge a near-duplicate. A separate mod is acceptable only if there's a very good reason, and then it should stand on its own: not named/id'd as a "fork" (e.g. `genie-friends-minimize-animation` would match the mod's display name), and with a clear statement of what it does that the two existing mods don't.

**2. `@author` must be the single responsible author (the submitter).** Windhawk's ownership model is one author per mod - `@author lolstijl & akilluminati47` should be just `akilluminati47` (matching `@github`), with lolstijl credited in the README (which the Credits section already does).

m417z:

akilluminati47:
after weekend ill fix these. broke it

PR #4882, opened 2026-07-23
Add pinned apps to all taskbars while keeping per-monitor windows

Divc09:
## What changed

Adds `pinned-apps-on-all-taskbars`, a Windows 11 taskbar mod that shows pinned launchers on every monitor while keeping running-window buttons on the taskbar for their monitor.

## Why

Windows' built-in "Taskbar where window is open" mode keeps running windows local but omits pinned launchers from secondary taskbars. The previous filter-only approach created non-functional placeholder buttons and left stale state when disabled.

This implementation uses Explorer's native "All taskbars" controller path for launcher creation, click handling, ordering, pin/unpin updates, and the transition back to a pinned launcher after the last window closes. It then preserves the taskbar monitor handle and applies the real-window monitor filter during filter creation, combining the two desired behaviors.

The mod also broadcasts the taskbar settings refresh after initialization and after Windhawk has removed the hooks, so enabling/disabling rebuilds the taskbar models without requiring an Explorer restart.

## Validation

- Compiled with the official Windhawk compiler for x86-64 and ARM64 targets.
- Hook symbols validated against the current Windows 11 `taskbar.dll` symbols.
- Submission is a single file under `mods/`, as required by this repository.

## User setup

Keep Windows set to **All taskbars** under Taskbar behaviors,, then enable the mod.

## Changelog

* New mod submission.

## Mod authorship

This mod was created by:
Divc09

### DIFF
--- a/mods/pinned-apps-on-all-taskbars.wh.cpp
+++ b/mods/pinned-apps-on-all-taskbars.wh.cpp
@@ -2,7 +2,7 @@
 // @id              pinned-apps-on-all-taskbars
 // @name            Pinned apps on all taskbars
 // @description     Shows pinned apps on every monitor while keeping running apps on the taskbar where their window is open
-// @version         2.1.0
+// @version         1.0.0
 // @author          Div
 // @github          https://github.com/Divc09
 // @include         explorer.exe
--- a/mods/pinned-apps-on-all-taskbars.wh.cpp
+++ b/mods/pinned-apps-on-all-taskbars.wh.cpp
@@ -141,7 +141,7 @@ BOOL Wh_ModInit() {
         return FALSE;
     }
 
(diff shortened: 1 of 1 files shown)


m417z:
> Enabling and disabling the mod rebuilds the taskbar models automatically; an Explorer restart isn't required.

I tried the mod, and it didn't work for me. Does it work for you?

After enabling the mod, I only saw it being applied after changing the "When using multiple displays, show my taskbar apps on" option or restarting explorer. Same for pinning/unpinning an item - it stays on the secondary taskbar until the taskbar is rebuilt via settings or an explorer restart.

Divc09:
"When using multiple displays, show my taskbar apps on" - This option needs to be set to all taskbars for this to work, but the behavior is still slightly buggy. I'll continue working on it and update the mod to be more seamless soon

PR #4889, opened 2026-07-24
Update Always On Top Windows to v1.2

AhmedAwad7:
## Changelog

**🚀 What's New**
* 1- Added the ability to enable/disable the pin/unpin sound and the option to change it to another .wav sound
* 2- Added floating text mode to know that pin/unpin is working and the option to enable or disable it
* 3- Added support for Windows 7

**🚀 ما الجديد**
* 1- تم إضافة إمكانية تعطيل/تفعيل صوت التثبيت/إلغاء التثبيت وإمكانية تغييره بصوت أخر بصيغة .wav
* 2- تم إضافة وضع النص العائم لمعرفة أن التثبيت/إلغاء التثبيت يعملان وإمكانية تعطيله أو تفعيله
* 3- إضافة دعم لويندوز 7

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini


m417z:
### Submission review

---

1. **The tool-mod launcher boilerplate was modified — please restore the verbatim wiki snippet.** Version 1.1 contained the [wiki launcher](https://github.com/ramensoftware/windhawk/wiki/Mods-as-tools:-Running-mods-in-a-dedicated-process) byte-for-byte, but this update rewrites `Wh_ModAfterInit` (lines 547–624): the module lookup order is flipped to `kernel32.dll`-first, a plain `CreateProcess` fallback is added, extra logging is inserted, and the brace style in `Wh_ModSettingsChanged`/`Wh_ModUninit` is compressed. The launcher snippet should stay identical to the wiki so it's easy to review and to update across all mods — see [explorer-folder-hover-menu.wh.cpp](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/explorer-folder-hover-menu.wh.cpp) for a mod that keeps it verbatim. If your Windows 7 testing showed that the template fails there (e.g. `CreateProcessInternalW` not resolving from `kernelbase.dll`), please mention that in the PR discussion so the shared template can be fixed for everyone, rather than shipping a privately diverged copy.

2. **Settings names/descriptions embed two languages in one string — use the localization syntax instead** (lines 78–109). The multi-line `$name: |` blocks mean *every* user sees the English and Arabic text stacked together in the settings UI. User-facing strings should default to English, with other languages provided via the localization keys:
   ```yaml
   - soundEnabled: true
     $name: Enable sounds
     $name:ar-SA: تفعيل الصوت
     $description: Enable or disable sound notifications
     $description:ar-SA: تفعيل أو تعطيل إشعارات الصوت
   ```
   See [accent-color-sync.wh.cpp](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/accent-color-sync.wh.cpp#L50-L53) for an example.

3. **`SetWindowPos` results are ignored, so failures are reported as success** (lines 319, 329). The tool process runs at medium integrity, so pinning an elevated window (e.g. Task Manager, an admin terminal) fails due to UIPI — yet the mod still plays the "pinned" sound, shows the "📌 Pinned" toast, and adds the window to the tracked list. Check the return value and give distinct feedback (or at least skip the success sound/notification and don't track the window) when it fails.

<p>

- **`Wh_GetStringSetting`/`Wh_GetIntSetting` don't take a default value** (lines 148–156). The variadic arguments are printf-style formatting for the *value name* (e.g. `L"key[%d]"`), so the second arguments here are silently ignored — defaults come from the `==WindhawkModSettings==` block. Also, `Wh_GetStringSetting` never returns NULL (it returns `L""` on error), so the `if (onStr)` checks are unnecessary. Relatedly, the string literals `L"C:\Windows\Media\Speech On.wav"` (lines 140–141, 150, 153) contain invalid escape sequences (`\W`, `\M`, `\S`) — they compile to `C:WindowsMediaSpeech On.wav` with compiler warnings. Since these values are never actually used, the simplest fix is to drop them entirely:
  ```cpp
  PCWSTR onStr = Wh_GetStringSetting(L"soundOn");
  g_settings.soundOn = onStr;
  Wh_FreeStringSetting(onStr);
  ```
- **Settings reload races with the worker thread** (line 430). `WhTool_ModSettingsChanged` runs `LoadSettings` on a Windhawk callback thread while `MainThread` may be reading `g_settings.soundOn`/`soundOff` in `PlaySystemSound` — a concurrent `std::wstring` write/read is undefined behavior. You already have `g_cs`; guard `g_settings` with it (copy the string out under the lock before calling `PlaySound`).
- Conversely, as currently written the critical section around `g_trackedWindows` is unnecessary: the list is only touched from `MainThread` (hotkey handler) and in `WhTool_ModUninit` *after* the thread has been joined. Either drop it or repurpose it for the settings race above.
- **Dead/ineffective code**: `WM_HIDE_NOTIFICATION` (line 117) is never used. The `DestroyWindow(g_notificationHwnd)` block in `WhTool_ModUninit` (lines 452–455) is also ineffective — by that point `MainThread` has exited, the system has already destroyed the windows it owned, and `DestroyWindow` can only destroy windows created by the calling thread anyway. Remove it, or destroy the notification window in `MainThread` after the message loop.
- **`RegisterClassEx` is called on every notification** (line 275) — every call after the first fails with `ERROR_CLASS_ALREADY_EXISTS` and the result is ignored. Register the class once (e.g. in `MainThread` next to `TopMostModClass`), set `wc.hInstance`, and unregister it on thread exit like you do for the main class. While at it, the `SetProp`/`GetProp` round-trip for the text can be replaced by a simple global — there are only two possible literal strings, and props should be removed with `RemoveProp` before destruction.

</p>
</details>

<p>

- **Consider using the real `WS_EX_TOPMOST` state instead of a tracked list as the source of truth** for the toggle: `GetWindowLongPtr(hWnd, GWL_EXSTYLE) & WS_EX_TOPMOST`. Today, after the tool process restarts (mod update or Windhawk restart) the list is lost, so a previously pinned window needs *two* hotkey presses to unpin (the first press "re-pins" it). The style check also lets users unpin windows made topmost by the app itself or by other tools. The list would then only be needed for unpin-on-unload (item 3 above).
- **The toast always appears on the primary monitor**: `SystemParametersInfo(SPI_GETWORKAREA, ...)` (line 259) returns the primary monitor's work area. On multi-monitor setups it would be more useful near the affected window — use `MonitorFromWindow(hWnd, MONITOR_DEFAULTTONEAREST)` + `GetMonitorInfo` and its `rcWork`.
- **The toast is not DPI-aware and uses the stock device-context font**: the hard-coded 200×40 box and default font will look tiny on high-DPI displays, and the stock font may not render the 📌/📍 emoji (especially on Windows 7, which this update claims to support). Consider sizing from the window's DPI and selecting the message font from `SystemParametersInfo(SPI_GETNONCLIENTMETRICS, ...)`.
- **The hotkey is hard-coded to Ctrl+Alt+T**, and a `RegisterHotKey` failure (line 388) is only logged — if another app owns the combination, the mod is silently inert. A settings option for the modifier/key would solve both (reload the hotkey in `WhTool_ModSettingsChanged`).

</p>
</details>

AhmedAwad7:
@m417z 

* The tool-mod launcher has been adjusted Similar to what was mentioned in the wiki


* The overlap between Arabic and English in the settings has been fixed and they have been made separate.

* The part related to starting the window pinned has been fixed so that no sound occurs or floating text appears when the window pinned fails.

PR #4892, opened 2026-07-24
Alt+Tab Plus

BlueFinch3000:
Adds Alt+Tab Plus, a Windows 11 mod that augments the native Alt+Tab experience with:

- Commonly used and optionally pinned programs.
- Recently modified files from configurable folders.
- Program-specific recent files sourced from Windows Jump Lists.
- Column, top-strip, and bottom-strip layouts.
- An adjustable native thumbnail area for the column layout.
- Companion panels compatible with other XAML styling mods.

The mod performs discovery locally and does not make network requests. It compiles successfully with Windhawk's bundled compiler on Windows 11.

https://github.com/user-attachments/assets/fae83d29-1930-4501-9819-7a780c6530ac

### Compatibility

The companion panels do not use Explorer's XAML diagnostics connection. This allows the mod to coexist with tools and mods such as Windows 11 Taskbar Styler, Windows 11 File Explorer Styler, and UWPSpy. The previously experimental integrated XAML renderer has been removed.

## Changelog

* Not applicable—this pull request introduces a new mod.

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [x] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/alt-tab-plus.wh.cpp
+++ b/mods/alt-tab-plus.wh.cpp
@@ -4,7 +4,7 @@
 // @description     Adds common programs and recently modified files to the Windows 11 Alt+Tab overlay
 // @version         0.3.1
 // @author          BlueFinch
-// @github 			BlueFinch3000
+// @github          https://github.com/BlueFinch3000
 // @include         explorer.exe
 // @architecture    x86-64
 // @compilerOptions -lole32 -loleaut32 -lruntimeobject -lshell32 -lversion -ldwmapi -lgdi32
@@ -3111,7 +3111,7 @@ bool HookAltTabShow() {
         return false;
     }
 
-    WindhawkUtils::SYMBOL_HOOK hooks[] = {
+    WindhawkUtils::SYMBOL_HOOK twinuiPcshellDllHooks[] = {
         {
             {LR"(public: virtual long __cdecl XamlAltTabViewHost::Show(struct IImmersiveMonitor *,enum ALT_TAB_VIEW_FLAGS,struct IApplicationView *))"},
             &XamlAltTabViewHost_Show_Original,
@@ -3128,7 +3128,9 @@ bool HookAltTabShow() {
             true,
         },
     };
-    if (!WindhawkUtils::HookSymbols(twinui, hooks, ARRAYSIZE(hooks))) {
+    if (!WindhawkUtils::HookSymbols(
+            twinui, twinuiPcshellDllHooks,
+            ARRAYSIZE(twinuiPcshellDllHooks))) {
         Wh_Log(L"Couldn't hook XamlAltTabViewHost::Show");
         return false;
     }
--- a/mods/alt-tab-plus.wh.cpp
+++ b/mods/alt-tab-plus.wh.cpp
@@ -2,18 +2,15 @@
 // @id              alt-tab-plus
 // @name            Alt+Tab Plus
 // @description     Adds common programs and recently modified files to the Windows 11 Alt+Tab overlay
-// @version         0.3.1
+// @version         0.4.1
 // @author          BlueFinch
 // @github          https://github.com/BlueFinch3000
 // @include         explorer.exe
 // @architecture    x86-64
-// @compilerOptions -lole32 -loleaut32 -lruntimeobject -lshell32 -lversion -ldwmapi -lgdi32
+// @compilerOptions -lole32 -lshell32 -lversion -ldwmapi -lgdi32
 // ==/WindhawkMod==
 
 // SPDX-License-Identifier: GPL-3.0-only
-//
-// The XAML diagnostics attachment pattern is adapted from Windows 11 Taskbar
-// Styler by m417z and the ExplorerTAP work it credits.
 
 // ==WindhawkModReadme==
 /*
@@ -52,10 +49,9 @@ files are skipped.
 This mod supports the Windows 11 XAML Alt+Tab experience. It doesn't support the
 classic Windows 10 switcher or third-party switcher replacements.
 
-The default companion-window renderer doesn't use XAML diagnostics and can
-coexist with Windows 11 Taskbar Styler, File Explorer Styler, UWPSpy,
-ExplorerBlurMica, and TranslucentTB. An optional integrated XAML renderer is
-available, but Windows allows only one XAML diagnostics consumer per process.
+The companion panels don't use XAML diagnostics and can coexist with Windows 11
+Taskbar Styler, File Explorer Styler, UWPSpy, ExplorerBlurMica, and
+TranslucentTB.
 
 After installing the source as a local Windhawk mod, restart Explorer once if
 the cards don't appear on the first Alt+Tab invocation.
@@ -64,12 +60,6 @@ the cards don't appear on the first Alt+Tab invocation.
 
 // ==WindhawkModSettings==
 /*
-- renderMode: companion
-  $name: Renderer
-  $description: Companion is compatible with other XAML styling mods. Integrated XAML is experimental and requires exclusive access to XAML diagnostics.
-  $options:
-  - companion: Compatible companion panels
-  - xaml: Integrated XAML panels (experimental)
 - layout: columns
   $name: Layout
   $description: Placement of common programs, open windows, and recent files.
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

---

1. **The global `std::thread g_worker` crashes Explorer at process exit.** `Wh_ModUninit` (and thus `StopWorker`) only runs on a controlled mod unload — it is *not* called when the host process terminates (Explorer restart, sign-out, reboot). In that case the OS kills the worker thread, then the CRT runs the destructor of the global `g_worker`, which is still `joinable()`, so `~thread()` calls `std::terminate()` and aborts explorer.exe on every shutdown. Use the standard pattern:

   ```cpp
   [[clang::no_destroy]] std::optional<std::thread> g_worker;
   ```

   and in `StopWorker`: `g_worker->join(); g_worker.reset();`. See [vlc-discord-rpc.wh.cpp#L256](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/vlc-discord-rpc.wh.cpp#L256) for the same pattern.

2. **`ShellExecuteW` is called on a thread that never initialized COM.** In companion mode, a click runs `Launch` on the companion thread, which never calls `CoInitializeEx`. The `ShellExecuteEx` documentation requires COM to be initialized (STA with `COINIT_DISABLE_OLE1DDE`), and `shell:AppsFolder\...` targets — which the settings explicitly advertise for pinned programs — depend on it. Initialize COM in `CompanionThreadProc` (with a matching `CoUninitialize`), or better, run `Launch` on a short-lived helper thread with STA COM. The latter also helps the XAML renderer, where the `Click` handler currently runs `ShellExecuteW` synchronously on the Alt+Tab UI thread and can block it (e.g. a target resolving over the network).

3. **Constant polling that runs for the mod's whole lifetime.** Two instances:
   - The companion thread's 50 ms `SetTimer` (`CompanionThreadProc`) fires ~20×/second forever — including while the panels are hidden, and even when the renderer setting is `xaml` and the companion windows are never shown. Create the timer when the panels are shown and kill it on hide.
   - The worker wakes every 2 s to poll `GetForegroundWindow` for usage tracking. The preferred design is event-driven: `SetWinEventHook(EVENT_SYSTEM_FOREGROUND, ...)` (see [taskbar-auto-hide-when-maximized.wh.cpp#L1101](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/taskbar-auto-hide-when-maximized.wh.cpp#L1101)); the companion thread's message loop can host it. The worker then only needs to wake for the refresh interval.

4. **When the Alt+Tab hook can't be installed, the mod stays resident doing background work with no user-visible function.** `HookAltTabShow`'s failure return is ignored in `Wh_ModInit` (e.g. on Windows 10, where the mod is documented as unsupported but still injects into explorer.exe, or after a future symbol change). In the default companion mode nothing will ever be shown, yet the worker keeps scanning folders and polling the foreground window forever. Install the hook first, and if it fails, return `FALSE` from `Wh_ModInit` without starting the threads — Windhawk will retry after each settings change, so this is a valid pattern.

5. **Add a screenshot or GIF to the README.** This is a highly visual mod; a screenshot of the layouts would help users a lot. Allowed image hosts are `i.imgur.com` and `raw.githubusercontent.com`.

<p>

- `CompanionThreadProc` calls `PeekMessageW(nullptr, ...)` with a null `LPMSG`. It happens to be safe only because no message can exist yet; pass a real `MSG` struct for the queue-creation idiom.
- `FindVisibleNativeAltTabWindow` / `OnWindowCreated` treat `Shell_InputSwitchTopLevelWindow` (the input-language switcher flyout) as a candidate for `g_nativeAltTabHostWindow`, i.e. as a window the mod may reposition via `ConstrainNativeAltTabWindow`. Including that class for TAP initialization matches Windows 11 Taskbar Styler, but as a *resize target* it's wrong — if the flyout is ever picked, the mod would move the language switcher. Restrict the corridor-resize candidates to `XamlExplorerHostIslandWindow`.
- `SaveUsageRecords` writes the full ~66 KB `UsageStore` blob every refresh interval (down to every 10 s) even when nothing changed. Track a dirty flag and only persist on change.
- With "Launch folder windows in a separate process" (or multiple explorer.exe instances generally), every instance runs its own worker and companion thread, and all of them scan folders and overwrite the same `programUsageV1` value, while only the shell instance ever shows Alt+Tab. Consider bailing out (`return FALSE`) when the current process isn't the shell instance (e.g. compare the PID of `GetShellWindow()`).
- `#include <set>` is unused.

</p>
</details>

<p>

- **The reparse-point filter breaks OneDrive folders.** `DiscoverRecentFiles` skips anything with `FILE_ATTRIBUTE_REPARSE_POINT`, but OneDrive cloud placeholder files carry that attribute — and the default folders (`Desktop`, `Documents`) are exactly the ones OneDrive commonly redirects. Many users would see a permanently empty "Recent files" card. Skip reparse points only for directories (the cycle-avoidance case), or check the reparse tag for the `IO_REPARSE_TAG_CLOUD*` family and keep those files. Relatedly, the `%USERPROFILE%\Documents` defaults don't match redirected known folders — consider resolving them via `SHGetKnownFolderPath` instead.
- In `DiscoverCommonPrograms`, the `stable_sort` skip count uses `settings.commonPrograms.size()` (the configured count), but duplicate configured commands are deduplicated, so fewer pinned entries may actually be inserted — then the first auto-ranked entries escape sorting. Since pinned entries already have `score = ULLONG_MAX`, sorting the whole vector is simpler and always correct.
- The hide logic (`elapsed > 250 && !(GetAsyncKeyState(VK_MENU) & 0x8000)`) doesn't handle the persistent switcher: with Ctrl+Alt+Tab the native overlay stays open after all keys are released, but the companion panels vanish after 250 ms. Tracking the visibility of the native host window would cover both cases.
- `GetJumpListObjectPath` uses `SLGP_RAWPATH`, which can return paths with unexpanded environment variables; `PathIsExistingFile` then fails and the entry is silently dropped. Expand the path (or drop `SLGP_RAWPATH`) before the existence check.
- `PaintCompanionWindow` extracts every item's icon via `SHGetFileInfoW` on each `WM_PAINT` (and creates/destroys several fonts per item). Icon extraction can touch disk and even block on network paths, and hover changes repaint the whole panel. Cache the `HICON`s (and fonts) per entry list, and destroy them when entries change.
- Auto-ranking from `FeatureUsage\AppSwitched` filters value names through `PathIsExistingFile`, which drops all packaged/UWP apps (their value names are AUMIDs, not paths). If that's not intentional, AUMID entries could be mapped to `shell:AppsFolder\<AUMID>` commands, which `Launch` already supports.

</p>
</details>

mods/alt-tab-plus.wh.cpp
```diff
@@ -0,0 +1,3219 @@
...
+        {
+            {LR"(public: virtual long __cdecl XamlAltTabViewHost::Show(struct IImmersiveMonitor *,enum ALT_TAB_VIEW_FLAGS,struct IApplicationView *))"},
+            &XamlAltTabViewHost_Show_Original,
+            XamlAltTabViewHost_Show_Hook,
+        },
+        {
+            {
+                LR"(public: __cdecl winrt::impl::consume_Windows_Internal_Shell_TaskGroups_ITaskGroupWindowInformation::Position(struct winrt::Windows::Foundation::Rect const &)const )",
+                LR"(public: __cdecl winrt::impl::consume_Windows_Internal_Shell_TaskGroups_ITaskGroupWindowInformation::Position(struct winrt::Windows::Foundation::Rect const &)const)",
+                LR"(public: __cdecl winrt::impl::consume_Windows_Internal_Shell_TaskGroups_ITaskGroupWindowInformation<struct winrt::Windows::Internal::Shell::TaskGroups::ITaskGroupWindowInformation>::Position(struct winrt::Windows::Foundation::Rect const &)const)",
```
m417z, line 2683: The other mods use a symbol string that's not listed here:
https://github.com/ramensoftware/windhawk-mods/blob/069f254f818e34abc367301c24d59e698240958a/mods/alt-tab-per-monitor.wh.cpp#L552-L556

Does it work on your computer? It's also marked as optional, perhaps because it never works.

mods/alt-tab-plus.wh.cpp
```diff
@@ -0,0 +1,3219 @@
...
+- Added All and Program specific recent-file sources.
+- Added up to four program-specific Jump List sections with an independent
+  recent-file count for each program.
+
+## 0.2.0
+
+- Added compatible Win32 companion panels as the default renderer.
+- Fixed compatibility with Windows 11 Taskbar Styler and File Explorer Styler,
+  which already occupy Explorer's XAML diagnostics connection.
+- Kept the integrated XAML renderer as an optional experimental setting.
```
m417z, line 193: What are the advantages of using XAML mode? If there are none, perhaps it's better to remove it, since it conflicts with Windows 11 Taskbar Styler - the most popular Windhawk mod - and may cause confusion for users who missed it in the readme.

BlueFinch3000: I removed XAML mode. You're right. Definitely don't want to step on Taskbar Styler's toes!

m417z:

PR #4895, opened 2026-07-24
Explorer Command Bar 1.0.0

DanRotaru:
# Explorer Command Bar 1.0.0

Add your own buttons and dropdown menus to the **Windows 11 File Explorer command bar** (the toolbar with New / Sort / View), and hide the built‑in buttons, separators and spacing you don't want.

Explorer Command Bar demo

Designed for Windows 11 24H2 / 25H2 with the WinAppSDK (WinUI 3) File Explorer.

## Features

- **Custom buttons** — add as many toolbar buttons as you like, each running a command of your choice.
- **Dropdown menus & submenus** — a button can open a menu, and menu entries can themselves be submenus (three levels deep).
- **Path & selection placeholders** — `%path%` (active tab folder) and `%sel%` (selected file/folder) are substituted into the command parameters.
- **Flexible icons** — a Segoe Fluent Icons glyph, an `.exe` / `.dll` / `.ico` file, a Store‑app icon (`shell:AppsFolder\…`), the command executable's own icon, or no icon at all.
- **Hide built‑in elements** — individually hide New, Cut, Copy, Paste, Rename, Share, Delete, Sort, View, the group separators, the "See more" (…) overflow menu, the contextual commands (Set as background, Rotate left, Rotate right, Extract all) and the Details pane toggle.
- **Custom item spacing** — set the exact spacing between the command bar buttons, in pixels.
- **Open menus on hover** — optionally open dropdowns on hover, with a configurable delay.
- **Rock solid** — buttons are re‑applied automatically across tab switches, navigation, new tabs and new windows, and everything is cleanly restored when the mod is disabled.

## Screenshots

Hiding built‑in buttons and separators:

Hide buttons

You may hide even all options, and use only your custom ones:

Hide All buttons

## Command parameters

The following placeholders can be used in an item's parameters:

* `%path%` — the folder path of the currently active tab.
* `%sel%` — the full path of the selected file or folder in the active tab (the
  first one, if several are selected).

Wrap placeholders in quotes so paths with spaces work, e.g. `-d "%path%"`. If a used placeholder has no value (nothing selected, or a non‑filesystem location like *This PC*), the command is launched without any parameters. Commands run with the active tab's folder as their working directory.

## Icons

The **Icon glyph or icon path** field accepts several forms:

* **A glyph** — a hex code point of a [Segoe Fluent Icons](https://learn.microsoft.com/en-us/windows/apps/design/iconography/segoe-fluent-icons-font) glyph, e.g. `E756`.
* **A file path** — an `.exe`, `.dll` or `.ico` file to take the icon from, optionally with an icon index, e.g. `C:\Windows\System32\shell32.dll,3`.
* **A Store app** — `shell:AppsFolder\<AppUserModelID>` to use a modern Store app's icon (useful for apps whose `.exe` stub carries a legacy icon, such as Notepad and Calculator).
* **Empty** — the icon is extracted from the command's executable (app execution aliases such as `wt.exe` are resolved to their real target).
* **Hide icon** — enable the toggle to show no icon at all.

## Default configuration

Out of the box the mod adds:

* **Open in Terminal** — `wt.exe -d "%path%"`
* **Open in Notepad** — `notepad.exe "%sel%"`
* **Additional** ▾ (dropdown)
  * Open in VS Code — `code.exe "%path%"`
  * Open Paint — `mspaint.exe`
  * Open Calculator — `calc.exe`
  * **Commands** ▸ — `vite`, `npm init`, `npm install`, `npm run dev`,
    `npm run build`, `npm run start`
  * **AI** ▸ — `Claude`, `Codex`

All of the built‑in buttons stay visible by default. Everything is configurable in the mod settings.

## How it works

The mod injects a XAML diagnostics provider into Explorer and watches the WinUI 3 visual tree for the File Explorer command bar. When it appears, the configured buttons are inserted and the visibility / spacing settings are applied. The mod listens for the command bar being rebuilt so the buttons stay in place across navigation, new tabs and new windows, and it restores the original state of any element it touches when disabled.

## Changelog

* No.

## Mod authorship

This mod was created by:

- - [x] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/explorer-command-bar.wh.cpp
+++ b/mods/explorer-command-bar.wh.cpp
@@ -4,7 +4,7 @@
 // @description     Add your own buttons and dropdown menus to the Windows 11 File Explorer command bar, and hide the built-in ones
 // @version         1.0.0
 // @author          DanRotaru
-// @github          https://github.com/DanRotaru
+// @github          https://github.com/danrotaru
 // @homepage        https://dan13.me/
 // @include         explorer.exe
 // @architecture    x86-64
@@ -20,7 +20,7 @@ Add your own buttons and dropdown menus to the **Windows 11 File Explorer
 command bar** (the toolbar with New / Sort / View), and hide the built-in
 buttons, separators and spacing you don't want.
 
-![Explorer Command Bar demo](https://cepret.md/img/placeholder.svg)
+![Explorer Command Bar demo](https://github.com/DanRotaru/windhawk-mods/blob/master/exporer-command-bar/screenshots/main.gif?raw=true)
 
 Designed for Windows 11 24H2 / 25H2 with the WinAppSDK (WinUI 3) File
 Explorer.
@@ -50,17 +50,14 @@ Explorer.
 
 ## Screenshots
 
-Custom buttons on the command bar:
 
-![Buttons](https://cepret.md/img/placeholder.svg)
+Hiding built‑in buttons and separators:
 
-Dropdown menu with submenus:
+![Hide buttons](https://github.com/DanRotaru/windhawk-mods/blob/master/exporer-command-bar/screenshots/hide-buttons.gif?raw=true)
 
-![Dropdown menu](https://cepret.md/img/placeholder.svg)
+You may hide even all options, and use only your custom ones:
 
-Hiding built-in buttons and separators:
-
-![Hide buttons](https://cepret.md/img/placeholder.svg)
+![Hide All buttons](https://github.com/DanRotaru/windhawk-mods/blob/master/exporer-command-bar/screenshots/hide-all-buttons.jpg?raw=true)
 
 ## Command parameters
 
@@ -110,12 +107,23 @@ in the mod settings.
 
 ## How it works
 
-The mod injects a XAML diagnostics provider into Explorer and watches the WinUI
-3 visual tree for the File Explorer command bar. When it appears, the configured
-buttons are inserted and the visibility / spacing settings are applied. The mod
-listens for the command bar being rebuilt so the buttons stay in place across
-navigation, new tabs and new windows, and it restores the original state of any
-element it touches when disabled.
+The mod hooks a couple of functions of File Explorer's own WinUI 3 code
+(`FileExplorerExtensions.dll`) which run when the command bar is built, and
+finds the command bars from there by walking the XAML tree. The configured
+buttons are then inserted and the visibility / spacing settings are applied. The
+mod also listens for the command bar being rebuilt so the buttons stay in place
+across navigation, new tabs and new windows, and it restores the original state
+of any element it touches when disabled.
+
+Notably, the mod does **not** use XAML Diagnostics
+(`InitializeXamlDiagnosticsEx`), since only one XAML diagnostics consumer can be
+active in a process at a time. That makes it compatible with mods and tools
+which do use it, such as **Windows 11 File Explorer Styler**, ExplorerBlurMica
+and TranslucentTB.
+
+File Explorer windows which are already open when the mod is enabled are
+handled too, but if the buttons don't show up in one of them right away, opening
+a new tab or navigating to another folder makes them appear.
 
 */
 // ==/WindhawkModReadme==
@@ -386,12 +394,13 @@ element it touches when disabled.
 
 #include <windows.h>
 
+#include <windhawk_utils.h>
+
 #include <exdisp.h>
 #include <servprov.h>
 #include <shellapi.h>
 #include <shlobj.h>
 #include <shobjidl.h>
-#include <xamlom.h>
 
 #include <atomic>
 #include <memory>
@@ -402,15 +411,7 @@ element it touches when disabled.
 #include <unordered_set>
 #include <vector>
 
-std::atomic<bool> g_initialized;
 std::atomic<bool> g_unloading;
-
-// {7B0A0F2C-3D1E-4C8B-9A65-2E8F41D0B7A3}
-static constexpr CLSID CLSID_WindhawkTAP = {
-    0x7b0a0f2c,
-    0x3d1e,
-    0x4c8b,
-    {0x9a, 0x65, 0x2e, 0x8f, 0x41, 0xd0, 0xb7, 0xa3}};
 
 static constexpr CLSID kCLSID_ShellWindows = {
     0x9ba05972,
@@ -480,17 +481,6 @@ struct {
     std::vector<ActionItem> items;
 } g_settings;
 
-HMODULE GetCurrentModuleHandle() {
-    HMODULE module;
-    if (!GetModuleHandleEx(GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS |
-                               GET_MODULE_HANDLE_EX_FLAG_UNCHANGED_REFCOUNT,
-                           L"", &module)) {
-        return nullptr;
-    }
-
-    return module;
-}
-
 ////////////////////////////////////////////////////////////////////////////////
 // clang-format off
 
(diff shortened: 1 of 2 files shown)


mods/explorer-command-bar.wh.cpp
```diff
@@ -0,0 +1,2900 @@
...
+  * **Commands** ▸ - `vite`, `npm init`, `npm install`, `npm run dev`,
+    `npm run build`, `npm run start`
+  * **AI** ▸ - `Claude`, `Codex`
+
+All of the built-in buttons stay visible by default. Everything is configurable
+in the mod settings.
+
+## How it works
+
+The mod injects a XAML diagnostics provider into Explorer and watches the WinUI
```
m417z: Using XAML Diagnostics (`InitializeXamlDiagnosticsEx`) makes the mod incompatible with other mods or customization tools that use it. Specifically, it makes the mod incompatible with [Windows 11 File Explorer Styler](https://windhawk.net/mods/windows-11-file-explorer-styler), a popular generic styling mod.

If possible, please try to avoid using XAML Diagnostics. There are no existing File Explorer mods which do that, so it might require some research.

If you can't find an alternative solution, let me know. It's less preferable, as it will make it less useful for users who also want to style their explorer.

DanRotaru: I understand that it may be incompatible, but I think it's ok to let user decide which one to use, I may add an disclaimer in the mod description, which will be ok I suppose. Using the XAML Diagnostics is best solution in my case I think, so I would not replace it with an alternative. For me the functional part is more useful than styling the file explorer, but again, it depends on user. 

Thank you!

m417z: > Using the XAML Diagnostics is best solution in my case I think, so I would not replace it with an alternative.

Do you think so because you'd rather not spend time looking for an alternative way, or is there another reason? XAML Diagnostics is powerful, and allows to implement generic styling capabilities which are difficult to implement otherwise. But for specific mods, it's usually an overkill. For example, for the taskbar, there's the taskbar styler mod, and there are many other taskbar mods which customize specific parts of the taskbar XAML without using XAML Diagnostics. The same could be done with File Explorer.

If we'll get to the conclusion that XAML Diagnostics is the way to go, then yes, it's best to add information about the incompatibility to the readme. The users can choose what they prefer, but I'm also afraid that some users will skip the readme and then wonder why things are buggy.

DanRotaru: Hey, I thought the best solution is to use the XAML Diagnostics, but I was wrong, I found another way to implement it. I've updated it and now it works great with the other mods, without any conflict. I've tested the Windows 11 File Explorer Styler and seems like everything is ok, all is good, no crashes, slow rendering and so on, all looks native like.
image


## How it works

The mod hooks a couple of functions of File Explorer's own WinUI 3 code (`FileExplorerExtensions.dll`) which run when the command bar is built, and finds the command bars from there by walking the XAML tree with `VisualTreeHelper`. The configured buttons are then inserted and the visibility / spacing settings are applied. The mod also listens for the command bar being rebuilt so the buttons stay in place across navigation, new tabs and new windows, and it restores the original state of any element it touches when disabled.

### Compatibility

Notably, the mod does **not** use XAML Diagnostics (`InitializeXamlDiagnosticsEx`). Only one XAML diagnostics consumer can be active in a process at a time, so a mod which uses it stops working as soon as another one is enabled. Avoiding it means this mod can be used together with:

* **Windows 11 File Explorer Styler**
* ExplorerBlurMica, TranslucentTB and similar tools
* UWPSpy and other XAML inspection tools

File Explorer windows which are already open when the mod is enabled are handled too, but if the buttons don't show up in one of them right away, opening a new tab or navigating to another folder makes them appear.

m417z:

DanRotaru:
Thanks for the detailed review. All six required items are addressed, and I took all of the optional improvements plus most of the functionality notes. Summary below, in the order of the review.

## Required items

**1. README image host.** All three images now use `raw.githubusercontent.com`:

```
https://raw.githubusercontent.com/DanRotaru/windhawk-mods/master/exporer-command-bar/screenshots/main.gif
https://raw.githubusercontent.com/DanRotaru/windhawk-mods/master/exporer-command-bar/screenshots/hide-buttons.gif
https://raw.githubusercontent.com/DanRotaru/windhawk-mods/master/exporer-command-bar/screenshots/hide-all-buttons.jpg
```

**2. Raw ABI pointers as element identity.** Fixed the way you suggested, modeled on `compact-start-menu`. The four address-keyed maps (`g_originalVisibility`, `g_originalOverflow`, `g_originalMargin`, `g_originalMinWidth`) and the separate `g_watches` list are gone, replaced by a single registry of per-element entries:

```cpp
struct ManagedElement {
    DWORD threadId;
    winrt::weak_ref<mux::UIElement> element;

    bool hasOriginalVisibility = false;
    mux::Visibility originalVisibility = mux::Visibility::Visible;

    bool hasOriginalSpacing = false;
    mux::Thickness originalMargin{};
    double originalMinWidth = 0;

    bool hasOriginalOverflow = false;   // Command bars only.
    muxc::CommandBarOverflowButtonVisibility originalOverflow =
        muxc::CommandBarOverflowButtonVisibility::Auto;

    bool watched = false;
    int64_t visibilityToken = 0;
};
```

* Entries are matched by resolving the weak reference (`entry.element.get() == element`), never by address. C++/WinRT's `operator==` falls back to comparing `QueryInterface(IUnknown)` results, so this is a real COM identity check and works even when the lookup passes an `AppBarButton` while the entry holds a `UIElement`.
* Lookups are restricted to the current thread's entries and prune this thread's dead entries on every call, so the list no longer grows for the whole Explorer session, and a recycled address can't be matched: a stale entry is dropped before any comparison happens.
* `SetManagedVisibility`, `ApplyItemSpacing`, `EffectiveVisibility`, the `OverflowButtonVisibility` original and the watcher token all go through this one registry.
* The "already watched" check is now a flag on the entry, evaluated under the lock, so the stale-key early return that skipped registering a watcher is gone.
* Unwatching and forgetting are separate steps: `RemoveButtonsForCurrentThread` unregisters the watchers first, restores from the entries, and only then drops this thread's entries.

**3. Hardcoded "Open in Terminal" fallback.** Dropped - an empty item list is a supported configuration. One follow-up this exposed: with no custom buttons, *Vertical separator after View* had no separator to bind to, because it was identified as "the separator right before our first custom button". It now falls back to the first separator after the View button, so the hide-only configuration from the third screenshot keeps that option working.

**4. Synchronous, uncached icon extraction on the UI thread.** Two changes:

* Decoding is split from XAML. `DecodedIcon` holds premultiplied BGRA pixels plus width/height, is resolved once per `(icon setting, command)` pair into a global cache, and only the `WriteableBitmap` is constructed per call (per `XamlRoot`). The cache is dropped in `LoadSettings`, so editing the settings still picks up changed icons. A cache entry with no pixels is a remembered failure, so the `code.exe`-style full fallback chain is walked at most once per process.
* Menus are built lazily: `MenuFlyout.Items` is populated on the flyout's `Opening` event instead of while the window or tab is being built, so a command bar build no longer walks the whole menu an submenu tree. As a safety net (and to get the work done a little before the click) the entries are also created on the button's `PointerEntered`.

**5. Launch worker thread lifetime.** The thread handles are kept in `g_launchThreads`, finished ones are reaped whenever a new launch starts, and `Wh_ModUninit` waits for the rest. The wait is the last thing uninit does, after the per-window restore via `RunFromWindowThread`, so it can't deadlock against a UI thread we're sending a message to. `OnActionInvoked` also returns early when the mod is already unloading.

**6. Defaults which don't exist on a typical machine.** The defaults are now stock Windows 11 only:

* **Open in Terminal** - `wt.exe -d "%path%"`
* **Open in Notepad** - `notepad.exe "%sel%"`
* **Additional** ▾ - Paint, Calculator, **System** ▸ (Task Manager, Control Panel)

The submenu is kept so the three-level feature is still visible out of the box, and the developer set (VS Code, the `npm` commands, the AI CLIs) moved to a new **Example configurations** section in the README.

## Optional improvements

* **`WindhawkUtils::StringSetting`** - `LoadActionItem` uses `StringSetting::make` with the same format arguments; the five manual `Wh_FreeStringSetting` calls are gone.
* **`Wh_ModUninit` unregistering watchers** - watchers are now unregistered per thread in `RemoveButtonsForCurrentThread`, and the hover `DispatcherTimer`s are tracked per thread and `Stop()`ed there too (and when the buttons are recreated after a settings change). What's left in the registry at the end of uninit can only belong to a thread `RunFromWindowThread` failed to reach - that can't be unregistered from another thread, so uninit logs the window it couldn't reach and how many elements were left, instead of silently clearing.
* **`VectorChanged` coalescing** - the recompute mechanism was generalized into `QueueCommandBarUpdate(commandBar, fullUpdate)`, and the `VectorChanged` handler uses it, so appending N buttons results in one full update instead of N passes over ~30 commands. A pending recompute is upgraded to a full update if both are requested.
* **`g_symbolsHooked`** - reset on the failure path, so a transient symbol-resolution failure no longer disables the mod for the whole process.
* **Tree walk short-circuit** - `CollectCommandBars` stops as soon as both `FileExplorerCommandBar` and `FileExplorerSecondaryCommandBar` are found.
* **Includes** - `<chrono>`, `<cwctype>`, `<cstdlib>`, `<cstdio>` and `<cstring>` added, `<initguid.h>` removed.
* **Dead setting read** - `%s.type` is only read when `depth < kMaxMenuDepth`.
* **README nits** - the empty paragraph under `## Screenshots` is gone. I left the `exporer-` directory name as is for now, since renaming it breaks the image URLs of the already-published screenshots; the mod id and file name themselves are `explorer-command-bar`.

## Functionality notes

Addressed:

* **Item spacing in DIPs** - the setting is renamed to *Item spacing (pixels at 100% scaling)* and the description now says the value is in effective pixels (1.5x as many physical pixels at 150% scaling) and that any value other than -1 also removes the built-in buttons' minimum width, shrinking their click target to the icon.
* **Monochrome icons** - `DecodeIcon` now handles icons with no color bitmap by reading the AND and XOR halves of the mask, instead of falling through to the empty-glyph path.
* **Unclassifiable buttons** - a primary command that can't be classified is logged with its icon URI and its automation id, so an SVG file name churning in a future Windows build is diagnosable from the log.
* **`GetExplorerContext` is not really off the UI thread** - correct, the comment was misleading and now says what the worker thread actually buys (not blocking inside the click handler), and that the shell calls marshal back to the Explorer UI thread.

Not changed, with reasons:

* **A failed icon leaves a blank button** - this one is already handled: the fallback glyph passed to `CreateIconElement` is `U+E783` (ErrorBadge), not an empty string. It's a private-use character, so it renders as nothing in a diff or a browser, which I suspect is how it read as `L""`.
* **Icon sharpness at 150%+** - skipped for now. Requesting a size derived from `RasterizationScale` means keying the icon cache per scale factor, which I'd rather not add in the same change as the caching itself. Happy to do it as a follow-up if you'd prefer it in now.
* **Spacing applied unevenly** - left as is deliberately. Applying the margin to every unrecognized `AppBarButton` would also apply it to the ~20 zero-width contextual commands, which would then take up space in the bar.
* **Identifying buttons by SVG file name** - kept, but now logged when it fails (see above). Automation ids are only exposed for a couple of the buttons, so the icon file name is still the only thing that covers all of them.
* **Already-open windows depend on a focused XAML element** - agreed, and it matches the limitation documented in the README. I don't have a better entry point for that case yet.

## Verification

The mod compiles clean with Windhawk's clang (`-Wall`; the only remaining `-Wextra` notes are
pre-existing missing-field-initializer warnings for `ManagedTarget`'s defaulted `group` member). Tested on Windows 11 25H2: buttons and menus across new windows, new tabs and navigation, hiding each built-in element, the contextual commands as the selection changes, the item spacing option, disabling the mod (everything restored), and changing the settings while windows are open.

DanRotaru:
All three required items are fixed, and I took every optional improvement plus most of the functionality notes. One deliberate revert from the previous round is called out at the end.

## Required items

**1. `OnCommandBarAdded` resolving other threads' weak references.** Correct, and the fix went the way you suggested at the end of that item rather than the minimal one: `g_entries`, `g_managedElements`, `g_hoverTimers` and the pending-update map are now `thread_local`, and the three mutexes that guarded them are gone.

* Every one of these containers was only ever touched from the owning UI thread, so the cross-thread resolve is now impossible by construction instead of being a rule the code has to remember - the `threadId` field and every `threadId ==` filter disappeared with it.
* The storage goes away with the window's thread, so a closed Explorer window no longer leaves its ~30-40 `ManagedElement` records behind for the rest of the session, and lookups no longer scan other windows' entries.
* `Wh_ModUninit` and `Wh_ModSettingsChanged` already reach every thread through `RunFromWindowThread`, and that is where the per-thread teardown happens (`RemoveButtonsForCurrentThread`: unregister watchers, stop hover timers, restore, then clear the thread's state). Uninit no longer clears anything cross-thread, and the "couldn't be restored" bookkeeping I had added for that case is gone since there is nothing left to log.
* `g_iconCache`, `g_launchThreads` and the settings stay global with their mutexes - they're genuinely shared and hold no thread-affine objects.

**2. Modal error box blocking `Wh_ModUninit`.** Switched to `ShellExecuteExW` with
`SEE_MASK_FLAG_NO_UI | SEE_MASK_NOASYNC`, and the failure is logged with `GetLastError()`. Good catch - with the defaults shipping commands that may not be installed, this was reachable by just clicking one of them and disabling the mod while the dialog was up.

**3. Bare command name resolved against the browsed folder.** Fixed: `LaunchItemForWindow` now passes `ResolveCommandPath(item.command)` as `lpFile`, so the command is resolved through the search path and the App Paths keys before the launch, and `lpDirectory` is only the working directory.

One addition on top of your one-liner: if the command still can't be resolved to a path (nothing on the search path, no App Paths entry - e.g. a tool that isn't installed), the launch now also drops `lpDirectory`, so the unresolved name can't be resolved against the browsed folder either. That case is logged, and it only affects commands which were about to fail anyway.

## Optional improvements

* **`Wh_ModBeforeUninit`** added, setting `g_unloading` one callback earlier.
* **`ParseGlyphSetting`** now requires every character to be a hex digit (and logs when it isn't), so
  `abc` and `E756x` end up as no icon instead of an unrelated glyph.
* **Focus-hook scan** is bounded: a `thread_local` flag records that a scan of the thread's island found its command bars, and `CommandBarControlGotFocusHandler` skips the tree walk while a tracked command bar is still alive. It's deliberately not conditioned on both bars being found, exactly because of the layout you mention where there is no secondary bar. A rebuilt command bar still comes back through `OnApplyTemplate` / `Loaded` / `CommandBarManager::CommandBar`, and the flag is cleared on teardown.
* **`<cstdint>`** added.
* **`MAXIMUM_REPARSE_DATA_BUFFER_SIZE`** instead of `16 * 1024`.
* **The two over-long lines** wrapped.

## Functionality notes

* **"Vertical separator after View" binding to a different element.** Good point, and the fallback I added last round was worse than I thought - it took over the group separator in front of the contextual commands, which then stopped being auto-hidden. The no-custom-buttons fallback now picks the *last* of Explorer's own separators, i.e. the same element our first button would have followed, so the option targets the same line in both configurations and the earlier-group separators keep their `GroupSeparator` classification.
* **The watcher's stale group snapshot.** The watcher no longer decides anything for a `GroupSeparator` target: it updates the remembered baseline and queues the recompute, which re-collects the group. So the stale snapshot is never acted on, instead of being acted on and then corrected.
* **`HasNoAlphaChannel` can't tell "no alpha channel" from "fully transparent".** True; commented, including why the ambiguity is resolved in favor of opaque (a visible button beats an invisible one).
* **No overlap with existing mods** - agreed, and thanks for checking `remove-command-bar` too.

## Reverted from the previous round: the default items

I've restored the original default configuration (VS Code, the **Commands** submenu with the `npm` entries, and the **AI** submenu), reverting the change I made for item 6 of the previous review, and removed the *Example configurations* README section that existed only to hold them.

The reasoning: the defaults are what makes the mod's shape obvious at a glance - a plain button, a dropdown, and a nested submenu, with a mix of glyph icons, Store-app icons and executable icons - and paring them down to four stock apps loses that. An entry whose command isn't installed is inert: clicking it launches nothing and logs the failure (no error box at all now, after item 2 above), and removing an item is two clicks in the settings. The README now says exactly that under *Default configuration*, so nobody is left wondering why a button does nothing. If this is a blocker for acceptance rather than a preference, say so and I'll trim them - but I'd rather ship the defaults that demonstrate the feature.

## Verification

The mod compiles clean with Windhawk's clang at `-Wall` (the remaining `-Wextra` notes are
pre-existing missing-field-initializer warnings for `ManagedTarget`'s defaulted `group` member).

DanRotaru:
Item 1 is fixed. Item 2 I'd like to keep as it is, with the reasoning below; you're right on both factual points, so I'm not going to repeat the "they're inert" claim from my last comment.

## 1. Symbol hook retried on every `LoadLibraryExW`

Fixed, both halves.

**The retry is now bounded by the failure mode.** `HookFileExplorerExtensionsSymbols` returns a result instead of a bool, so the two cases are no longer conflated:

```cpp
enum class SymbolHookResult {
    Success,
    // Symbol resolution itself failed, e.g. the PDB couldn't be downloaded.
    // Transient, so trying again later can succeed.
    ResolutionFailed,
    // The symbols resolved, but this build has none of the functions the mod
    // needs. Permanent for this Explorer, so it must not be retried: each
    // HookSymbols call for a module invalidates its symbol cache and forces a
    // re-resolution.
    NoSymbolFound,
};
```

`g_symbolsHooked` is reset only for `ResolutionFailed`, which is the transient case the previous round was about. `NoSymbolFound` keeps the flag set, so an unsupported Windows build costs exactly one symbol resolution instead of one per module load - which, as you say, is the realistic trigger and the one that turned "the mod does nothing" into "Explorer is slow".

**The `LoadLibraryExW` hook filters by module name**, so the common path no longer even calls
`GetModuleHandle`:

```cpp
HMODULE module = LoadLibraryExW_Original(lpLibFileName, hFile, dwFlags);
if (!module || g_unloading || !lpLibFileName) {
    return module;
}

PCWSTR fileName = lpLibFileName;
for (PCWSTR p = lpLibFileName; *p; p++) {
    if (*p == L'\\' || *p == L'/') {
        fileName = p + 1;
    }
}

// LoadLibraryEx appends the default extension itself, so the caller may have
// left it out.
if (_wcsicmp(fileName, L"FileExplorerExtensions.dll") == 0 ||
    _wcsicmp(fileName, L"FileExplorerExtensions") == 0) {
    HookFileExplorerExtensionsIfLoaded(/*applyHooks=*/true);
}
```

Two small differences from your snippet: forward slashes are treated as separators too, and the extension-less spelling is accepted, since `LoadLibraryEx` appends `.dll` for the caller and both forms reach the same module.

## 2. The default configuration

Your corrections are right, and I withdraw the "inert" argument from my previous comment: `lpFile` is `cmd.exe`, so those launches succeed, the user gets a console with `'npm' is not recognized`, and nothing is logged. And `npm install` / `npm run …` do execute what the browsed folder's `package.json` declares.

I'd still like to ship the defaults as they are, as an explicit author's call:

* The command is not run silently or automatically - it's a toolbar button the user has to click, in a submenu they have to open first, from a list they configured (or chose to keep). The mod's whole purpose is "run this command against the current folder"; a user who removes `npm install` and adds `mytool.exe "%path%"` has exactly the same exposure to whatever that tool does with the folder.
* `cmd.exe /k` was chosen precisely so nothing is hidden: the console stays open with the command line and its output visible, rather than running detached.
* The defaults are also the documentation. The `README` section now describes what each default is, and the entries are removable in two clicks from the settings.

I don't think that resolves your objection - a default is a default, and the mod's shape can indeed be demonstrated with stock Windows apps - so if this blocks acceptance, say so and I'll drop the `cmd.exe` entries and move them to the README as an example, which is where they were in the previous revision. I'd just rather that be an explicit request than me guessing at the bar.

## Optional items and functionality notes

Not changed in this round; noting where I agree so they're not lost:

* **Icons without an alpha channel come out as opaque squares** - agreed, this is a real bug for the documented "point at an old `.exe`/`.dll`" case, and using `hbmMask` as the alpha (falling back to `MakeOpaque` when the mask can't be read) is the right fix. Happy to include it now if you'd like it before merge, otherwise it's the first thing in a follow-up.
* **An open menu flyout surviving `RemoveOurButtons`** - agreed, `flyout.Hide()` there is the same reasoning as stopping the hover timers.
* **`GetManagedElement` pruning on every lookup** - agreed, the prune and the match can share one scan, or the prune can run once per pass.
* **`ParseGlyphSetting` taking a one-character setting literally** - fair; the literal form should be either documented or dropped.
* **`<cwchar>` / `<cstring>`** - noted.
* **`SEE_MASK_FLAG_NO_UI` not covering UAC or the "open with" picker** - understood, and thanks for confirming the restore-first / wait-last ordering is the right trade-off there.

## Verification

The mod compiles clean with Windhawk's clang at `-Wall` (the remaining `-Wextra` notes are
pre-existing missing-field-initializer warnings for `ManagedTarget`'s defaulted `group` member).

m417z:
> the mod's shape can indeed be demonstrated with stock Windows apps - so if this blocks acceptance, say so and I'll drop the `cmd.exe` entries and move them to the README as an example, which is where they were in the previous revision. I'd just rather that be an explicit request than me guessing at the bar.

I think that it'll be a better first experience to have built-in things that works for everyone. "Blocks" is a strong word, but I tend to agree it's a better option. Why did you revert it?

I also wonder why you (or Claude) included the `__ptr64` symbols, they were only used in very very old Windhawk versions. It doesn't hurt, just redundant.

DanRotaru:
I chosen that default options cause I want to display more options, and to show app possibilities. The user may easily to edit, remove some items that don't want. 
Regarding __ptr64, I may fix it in future versions.

I also have a question for you. I also have locally developed two more mods, and want to understand the mods should be developed as two more additional mods or merged into this (Explorer Command Bar) mod (as an update)?
Both are working with Explorer Command Bar.

Some words about mods:

The first mod - replace/add the New+ (which is coming from PowerToys, with custom shortcuts, files, templates) instead of default New, which is less useful.

The second mod - add the context menu button to the explorer command bar. Instead of right click to the folder/file, click or hover to the item in explorer command bar.

m417z:
> the mods should be developed as two more additional mods or merged into this (Explorer Command Bar) mod (as an update)?

There's no single right answer, it's mostly up to you.

mods/explorer-command-bar.wh.cpp
```diff
@@ -0,0 +1,3272 @@
...
+    - parameters: ""
+    - iconGlyph: E81E
+    - separatorAfter: false
+    - subItems:
+      - - type: button
+        - name: Open in VS Code
+        - command: code.exe
+        - parameters: '"%path%"'
+        - iconGlyph: ""
+        - separatorAfter: false
```
m417z, line 250: This part is rejected by Windhawk 2.0 alpha 2 because the first item in the array doesn't have `subItems`. It's a bug in the alpha version, but adding this here fixes it:
```yaml
        - subItems:
          - - name: ""
            - command: ""
            - parameters: ""
            - iconGlyph: ""
            - hideIcon: false
            - separatorAfter: false
```
You might want to add it in the next update.

m417z (commented):
I tried the mod and it seems to crash on unload because some of the events aren't revoked:
* `button.Click`
* `menuItem.Click`
* `menu.Opening`
* `button.PointerEntered`
* `timer.Tick`
* `button.PointerExited`
* `button.PointerCanceled`

Please address this in the next update as well.

PR #4896, opened 2026-07-24
Rewrite taskbar-start-button-corner-fix.wh.cpp

alchemyyy:
## Changelog

* Replaced the bad UIA approach with a direct hook on the invisible InputSite window. Added some robustness around this hook to grab it through multiple InputSite window "pathways", along with retry and lifecycle management logic.

* Also improved the "corner candidacy geometry" check.

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/taskbar-start-button-corner-fix.wh.cpp
+++ b/mods/taskbar-start-button-corner-fix.wh.cpp
@@ -2,40 +2,38 @@
 // @id              taskbar-start-button-corner-fix
 // @name            Start Menu Corner Click Fix
 // @description     Fixes the issue where clicking in the corner of the taskbar doesn't open the Start menu on multi monitor setups.
-// @version         1.1
+// @version         2.0
 // @author          Alchemy
 // @github          https://github.com/alchemyyy
 // @license         MIT
 // @include         explorer.exe
 // @architecture    x86-64
-// @compilerOptions -lole32 -loleaut32 -luser32
+// @compilerOptions -lole32 -loleaut32
 // ==/WindhawkMod==
 
 // ==WindhawkModReadme==
 /*
 # Start Menu Corner Click Fix
 
-Fixes an issue on Windows 11 where clicking in the very corner of the taskbar doesn't open the Start menu
-on left-aligned taskbars when the start menu button is "in a sticky corner".
+Fixes the issue where clicking the bottom-left corner of a left-aligned taskbar
+doesn't reliably open the Start menu when another monitor is adjacent to that edge.
 
-## The Problem
+## How it works
 
-Windows 11's taskbar has an InputSite window (Windows.UI.Input.InputSite.WindowClass)
-that intercepts mouse clicks before they reach the Start button.
+The Windows 11 taskbar receives mouse input as `WM_POINTER` messages through an
+InputSite window. The mod intercepts the pointer pair in the configured corner
+region, prevents the original sticky-corner path from handling it, and invokes
+the Start button belonging to the clicked taskbar after mouse-up.
 
-This seems to be caused by the "sticky corners" feature, and thus only affects multi-monitor setups.
-
-## The Solution
-
-This mod hooks `CreateWindowInBand` to detect when the InputSite window is created,
-then installs a thread-specific mouse hook on its thread. This hook intercepts mouse
-clicks at the corner region and invokes the Start menu via UI Automation when the
-real button doesn't handle the click.
+UI Automation is initialized lazily on a dedicated worker thread, never on the
+taskbar input thread. If the taskbar's Start button can't be found or invoked,
+no fallback input is generated.
 
 ## Requirements
 
 - Windows 11 64-bit (tested on 25H2)
-- Windhawk v1.4 or later
+- A left-aligned taskbar
+- Windhawk v1.5 or later
 */
 // ==/WindhawkModReadme==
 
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

---

1. **Don't retry `Wh_ApplyHookOperations` with `Sleep` on the taskbar UI thread.** `HookInputSiteWindowProc` runs synchronously inside `CreateWindowInBand_Hook`, i.e. on the taskbar's UI thread in the middle of window creation. On an apply failure it sleeps 25 ms up to two times while holding `g_hookOperationLock` — freezing that thread for up to ~75 ms plus apply time, and blocking a concurrent mod unload for the same duration (`Wh_ModBeforeUninit` waits on the same lock). A `Wh_ApplyHookOperations` failure isn't a transient condition that a 25 ms wait fixes, and the code already has the correct fallback: the slot stays `pending` and "the next taskbar scan will retry". Call it once, log on failure, and drop the retry loop — see how [taskbar-empty-space-clicks](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/taskbar-empty-space-clicks.wh.cpp#L1389-L1393) applies the InputSite hook with a single call.

2. **Check the message type before querying the window tree.** `HandleInputSiteWindowMessage` performs `GetAncestor` + `GetWindowThreadProcessId` + up to two `GetClassNameW` calls for *every* message before the `switch`. Since the hooked window procedure is shared by every `Windows.UI.Input.InputSite.WindowClass` window in explorer.exe (not just the taskbar's), this overhead lands on all pointer-move, paint, and timer traffic of every XAML-island surface in the shell, forever. Only a handful of messages are relevant — switch on `message` first (or early-return for messages you don't handle) and do the taskbar identification only inside the relevant cases. [taskbar-empty-space-clicks](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/taskbar-empty-space-clicks.wh.cpp#L1357-L1360) early-outs on a cached taskbar thread id first for the same reason; caching the identified taskbar/InputSite window handles instead of re-deriving them per message also works.

3. **The four-slot hook machinery can collapse to a single hook.** All `Windows.UI.Input.InputSite.WindowClass` windows share one window procedure, which is why existing mods keep a single `InputSiteWindowProc_Original` and just guard against double-hooking with a small set — see [taskbar-empty-space-clicks](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/taskbar-empty-space-clicks.wh.cpp#L1378-L1396) (its comment there also documents why wndproc-hooking is used instead of subclassing). Four identical hook functions plus slot bookkeeping plus the `HookLifecycle` enum is a lot of code for a case that doesn't occur in practice; if this generality is an artifact of how the rewrite was generated rather than something you observed (multiple distinct InputSite window procedures), consider simplifying. Relatedly, `g_hookLifecycle` is only ever read and written under `g_hookOperationLock`, so it doesn't need to be `std::atomic`.

4. **Drop the `debugLogging` setting.** `Wh_Log` already compiles to a cheap conditional that does nothing unless Windhawk's own logging is enabled, so a second mod-level toggle only means diagnostics require flipping two switches. Call `Wh_Log` unconditionally and remove the setting (and the `g_settings.debugLogging` atomic).

<p>

- **With the default `edgeThickness == edgeLength == 5`, the region is a plain 5×5 square and `edgeThickness` has no effect.** The L-shape only materializes when `edgeLength > edgeThickness`. If the L-shape is the intended default behavior, consider a longer default arm length; otherwise the two settings/descriptions could be simplified to a single corner-size value.
- **Verify `SC_TASKLIST` behavior for secondary taskbars.** `QueueStartMenuActivation` posts `WM_SYSCOMMAND`/`SC_TASKLIST` to `Shell_SecondaryTrayWnd` as well. Historically that command opens the Start menu on the primary monitor (if the secondary tray handles it at all), so a corner click on a secondary taskbar may open the menu on the wrong monitor or do nothing — this is why [taskbar-empty-space-clicks](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/taskbar-empty-space-clicks.wh.cpp#L3252-L3258) clicks the Start button element of the clicked taskbar and falls back to a Win-key press. If you've verified it works per-monitor on current builds, all good — otherwise consider that approach (from a helper thread, not the input thread).

</p>
</details>

alchemyyy:
1. Removed the `Sleep` retry loop. `Wh_ApplyHookOperations()` is attempted once. Failures remain pending for the next taskbar scan.
2. Irrelevant messages are now filtered before any window-tree or taskbar-identification queries are performed.
3. Collapsed the four hook slots and callbacks into one shared InputSite WNDPROC hook. The lifecycle state is non-atomic and protected by the existing lock.
4. Removed `debugLogging` and its conditional guards. Relevant `Wh_Log` calls are unconditional.

Functionality:

1. Changed the defaults to an actual L-shaped region: `5`-pixel thickness and `15`-pixel arm length.
2. Replaced `SC_TASKLIST` with taskbar-specific UI Automation on a dedicated worker thread. It toggles the clicked taskbar’s Start button and falls back to the Windows key if necessary.

m417z:
### Submission review

Some of the optional notes might be worth addressing, such as high DPI support and better multi-monitor support.

---

- **The worker thread, COM and the UI Automation client are created unconditionally, in every `explorer.exe`, even when the mod can never fire.**

`Wh_ModInit` always calls `InitializeStartMenuActivationWorker()`, which spawns a thread, does `CoInitializeEx`, and creates a `CUIAutomation` instance (pulling `UIAutomationCore.dll` into the process) for the lifetime of the mod. On a single-monitor setup — or any setup with no monitor left-adjacent to the taskbar — `IsInCornerRegion` can never return true, so that thread and that UIA client sit idle forever. The same happens in secondary `explorer.exe` instances that host no taskbar at all.

Create the worker lazily instead — on the first intercepted corner press, or from `QueueStartMenuActivation` — under the `g_startMenuActivationLock` you already have. `Wh_ModInit` then only needs the existing "neither hook could be configured" check to decide whether to return `FALSE`.

- **Redundant `Wh_ApplyHookOperations()` on the `Wh_ModAfterInit` pass.** If an InputSite window already existed at init, `g_inputSiteWindowProcHookPending` stays `true` across the implicit apply that Windhawk performs when `Wh_ModInit` returns, so the rescan in `Wh_ModAfterInit` applies hook operations a second time for nothing. Clearing the flag together with the lifecycle transition fixes it:

  ```cpp
  AcquireSRWLockExclusive(&g_hookOperationLock);
  if (g_hookLifecycle != HookLifecycle::Unloading) {
      g_hookLifecycle = HookLifecycle::Running;
      g_inputSiteWindowProcHookPending = false;  // Applied implicitly already.
  }
  ReleaseSRWLockExclusive(&g_hookOperationLock);
  ```

<p>

* **`windhawk_utils.h` is included but never used.** Both hooks go through raw `Wh_SetFunctionHook` with `void*` casts; `WindhawkUtils::SetFunctionHook` is type-safe and checks the prototypes for you:

  ```cpp
  auto pCreateWindowInBand =
      (CreateWindowInBand_t)GetProcAddress(user32Module, "CreateWindowInBand");
  WindhawkUtils::SetFunctionHook(pCreateWindowInBand, CreateWindowInBand_Hook,
                                 &g_createWindowInBandOriginal);
  ...
  auto wndProc = (WNDPROC)GetWindowLongPtrW(inputSiteWindow, GWLP_WNDPROC);
  WindhawkUtils::SetFunctionHook(wndProc, InputSiteWindowProc_Hook,
                                 &g_inputSiteWindowProcOriginal);
  ```

* **The gesture state can get stuck if the pointer-up is never delivered.** Because the down is suppressed, XAML never takes pointer capture, so a press in the corner followed by a release outside the window relies entirely on `WM_POINTERLEAVE` arriving to clear `suppressing`. If it doesn't, every subsequent `WM_POINTERUPDATE` for that pointer is swallowed and taskbar hover/tooltips stop working until the next click. Cheap belt-and-braces: in the `WM_POINTERUPDATE` case, clear the state and fall through when the first button is no longer down.

  ```cpp
  case WM_POINTERUPDATE: {
      UINT32 pointerID = GET_POINTERID_WPARAM(wParam);
      if (IsActiveCornerPointer(window, pointerID)) {
          if (!IS_POINTER_FIRSTBUTTON_WPARAM(wParam)) {
              ClearCornerPointerState();  // Missed the up; don't keep eating moves.
              break;
          }
          ...
  ```

* **The hook bookkeeping is heavier than it needs to be.** `HookLifecycle`, `g_inputSiteWindowProcHookPending` and the retry-on-apply-failure path add a fair amount of machinery for what other taskbar mods do in a few lines (`if (g_initialized) Wh_ApplyHookOperations();`, see [taskbar-scroll-actions.wh.cpp#L1329-L1337](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/taskbar-scroll-actions.wh.cpp#L1329-L1337)). The `Unloading` guard is worth keeping; the rest could probably go.

* **The file now has mixed line endings** — the original was CRLF throughout, and the rewritten parts are LF (357 of 881 lines still end with CRLF). The rest of the repo is LF; converting the whole file would keep future diffs clean.

</p>
</details>

<p>

* **The corner interception isn't gated on the Start button actually being in the corner.**

`IsInCornerRegion` only checks the geometry (bottom-left of a taskbar monitor that has a left-adjacent monitor). It never checks taskbar alignment, so with the Windows 11 default *centered* alignment the mod swallows the `WM_POINTERDOWN`/`WM_POINTERUP` pair in the bottom-left corner — which is plain empty taskbar space there — and opens the Start menu. That silently changes behavior for a configuration the mod isn't meant for, and it takes those clicks away from other mods that act on empty taskbar space (e.g. Taskbar Empty Space Clicks), since the message never reaches the original window procedure.

* **The `InvokePattern` fallback was dropped.** v1.1 tried `UIA_TogglePatternId` and then `UIA_InvokePatternId`; `ActivateTaskbarStartButton` now goes straight from a missing toggle pattern to the Windows key. Worth keeping the invoke attempt in between — it's a few lines and covers builds where the button doesn't expose a toggle pattern.

* **The Windows-key fallback ignores which taskbar was clicked.** `SendWindowsKeyPress` opens Start on the primary monitor, so on a secondary taskbar the fallback silently does the wrong thing. Probably worth saying so in the README, or skipping the fallback entirely for secondary taskbars.

* **Each corner click walks the whole taskbar UIA tree.** `ElementFromHandle` + `FindFirst(TreeScope_Descendants, ...)` runs per click and marshals to the taskbar UI thread; that's both perceptible latency before the menu appears and the one call that `Wh_ModBeforeUninit` may end up waiting on (the worker join is `INFINITE`). Caching the `IUIAutomationElement` per taskbar `HWND` and only re-resolving when `Toggle()` fails would cut both.

* **Toggling when the menu is already open may race with light dismiss.** v1.1 checked `IsStartMenuOpen()` first; now a corner click while Start is open calls `Toggle()` unconditionally, which can collide with the shell's own dismiss-on-outside-click and produce a close/reopen flicker. Worth testing that case.

* **The corner region is in raw pixels.** `edgeThickness`/`edgeLength` aren't DPI-scaled, so 5/15 px is a noticeably smaller physical target on a 150–200% monitor than on a 100% one. Scaling by the taskbar monitor's DPI (`GetDpiForWindow`) would make the defaults behave the same everywhere.

</p>
</details>

alchemyyy:
1. The UI Automation worker is now created lazily from `QueueStartMenuActivation`, under `g_startMenuActivationLock`. `Wh_ModInit` no longer creates the thread or UIA client.
2. Removed the pending-hook state entirely. Existing hooks rely on Windhawk’s implicit post-init application, while hooks configured later call `Wh_ApplyHookOperations()` once.
3. Switched both hooks to the type-safe `WindhawkUtils::SetFunctionHook`.
4. Added recovery for a missed `WM_POINTERUP` by clearing suppression when `WM_POINTERUPDATE` reports that the first button is no longer down.
5. Normalized the file to LF line endings.

Also addressed several functionality notes:

- Corner interception now requires the Windows taskbar alignment setting to be left-aligned.
- Corner dimensions are scaled using the clicked taskbar’s DPI.
- Restored the `InvokePattern` fallback after `TogglePattern`.
- The Windows-key fallback is only used for the primary taskbar, avoiding activation on the wrong monitor.
- Start-button UIA elements are cached per taskbar and refreshed after activation failure.

I also revisited the taskbar-detection guidance from my earlier PRs. The current implementation retains the same class-name and current-process validation, but derives the root taskbar from the exact InputSite window being handled. I kept this approach because it naturally supports secondary taskbars, avoids selecting an arbitrary global taskbar, and doesn’t retain a taskbar/thread identity that can become stale after recreation.

Validated with x64 and ARM64 builds and the repository PR validation script.

m417z:

More details:

alchemyyy:
- **Removed the redundant startup `Wh_ApplyHookOperations()` call.** Hooks configured during `Wh_ModInit` are left for Windhawk to apply automatically when initialization returns. Only an InputSite hook discovered after initialization is applied explicitly.
- **Preserved retry behavior for runtime hook failures.** A failed runtime application remains pending and is retried during subsequent InputSite scans instead of becoming permanently stuck.
- **Kept a minimal unloading guard.** The `CreateWindowInBand` detour can still execute until Windhawk applies hook removals after `Wh_ModBeforeUninit`. The guard prevents a late callback from configuring or applying another hook during that interval.
- **Converted UI Automation ownership to `winrt::com_ptr`.** This removes the manual `Release()` bookkeeping and the associated cleanup helpers. `RemoveInvalidCachedStartButtons` remains because RAII manages COM lifetimes, but it does not evict entries for destroyed or recreated taskbar windows.
- **Deferred the registry query until after the geometry checks.** Normal taskbar clicks now exit through the cheap coordinate checks without calling `RegGetValueW`; taskbar alignment is queried only for an actual corner candidate.
- **Replaced the I/O completion port with an auto-reset event and atomic `HWND` slot.** The worker is created lazily once, and clicks arriving while it is busy are deliberately coalesced to the newest taskbar. This avoids repeated open/close toggling when Explorer or UI Automation is severely delayed.
- **Added UI Automation connection and transaction timeouts.** The worker now creates `CUIAutomation8`, uses `IUIAutomation2`, and requests one-second connection and transaction timeouts. Both settings are best-effort: failures are logged, but Start-button condition creation and activation still proceed using the platform defaults.
- **Simplified linker options.** Removed redundant `advapi32`, `user32`, and `runtimeobject` entries. The mod only explicitly links `ole32` and `oleaut32`.
- **Corrected metadata and wording.** The README now requires Windhawk v1.5 because that is when `WindhawkUtils::SetFunctionHook` was introduced, and the settings describe values as pixels at 100% scaling rather than implying that users must supply pre-scaled values.

PR #4905, opened 2026-07-25
Add Desktop Draggable Widgets

Stunning-dev:
## Mod authorship

This mod was created by:

- - [ ] Claude
- - [x] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/desktop-draggable-widgets.wh.cpp
+++ b/mods/desktop-draggable-widgets.wh.cpp
@@ -1,0 +1,1541 @@
+// ==WindhawkMod==
+// @id              desktop-draggable-widgets
+// @name            JonaOS Draggable Desktop Widgets
+// @description     Draggable Windows desktop widgets with transparent, light, and dark themes for calendar, clock, music, volume, battery, search, personalization, WiFi, Bluetooth, hotspot, and accessibility.
+// @version         1.9
+// @author          Jona like it, code it
+// @github          https://github.com/Stunning-dev
+// @include         windhawk.exe
+// @compilerOptions -ldwmapi -lgdiplus -lgdi32 -lole32 -lshell32
+// @license         MIT
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# JonaOS Draggable Transparent Widgets
+![Demo](https://i.imgur.com/3PpoXyw.gif)
+*Widgets can be dragged and made vertical or horizontal*
+![image](https://i.imgur.com/IntMcNV.png)
+*White Transparent (Better for Dark Wallpapers)*
+![image](https://i.imgur.com/lZCz1Yx.png)
+*Smocky*
+![image](https://i.imgur.com/zyL5hKj.png)
+*PassOver*
+![image](https://i.imgur.com/eNTYZt8.png)
+*Monochrome*
+![image](https://i.imgur.com/Y5GPtQ0.png)
+*Honey Dawn*
+![image](https://i.imgur.com/y2BR0Hj.png)
+*Pink*
+![image](https://i.imgur.com/K1vNfOh.png)
+*CharcoalCream*
+![image](https://i.imgur.com/cwhcLEo.png)
+*NavyLinen*
+![image](https://i.imgur.com/w8EBiFY.png)
+*OliveSand*
+![image](https://i.imgur.com/UmBUw5y.png)
+*CrimsomBlush*
+![image](https://i.imgur.com/JepCRa8.png)
+*DeepTeal*
+![image](https://i.imgur.com/HWn0dNW.png)
+*MerlotPeach*
+![image](https://i.imgur.com/EvRLWky.png)
+*OceanSand*
+![image](https://i.imgur.com/LYWTqNs.png)
+*IndigoMist*
+![image](https://i.imgur.com/opBfLJA.png)
+*SpruceMist*
+![image](https://i.imgur.com/E7O6IXT.png)
+*PlumCream*
+![image](https://i.imgur.com/p1je0M5.png)
+*MahoganyGold*
+![image](https://i.imgur.com/zbc7nN1.png)
+*ForestSage*
+![image](https://i.imgur.com/hWodzr2.png)
+*PineFrost*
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

---

**1. Keep the tool-mod launcher boilerplate verbatim — move the custom logic into `WhTool_ModSettingsChanged`.** The launcher block is a near-exact copy of the wiki snippet, except that `Wh_ModSettingsChanged` was edited to contain mod-specific logic. The snippet should stay byte-for-byte identical (it makes review and future maintenance much easier), and it already supports a `WhTool_ModSettingsChanged` callback for exactly this purpose — see [explorer-folder-hover-menu.wh.cpp](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/explorer-folder-hover-menu.wh.cpp) for reference. Concretely, add next to your other `WhTool_*` functions:

```cpp
void WhTool_ModSettingsChanged() {
    if (g_hwnd) {
        PostMessageW(g_hwnd, kSettingsChangedMessage, 0, 0);
    } else {
        LoadSettings();
    }
}
```

and restore the boilerplate's version:

```cpp
    if (g_isToolModProcessLauncher) {
        return;
    }

    WhTool_ModSettingsChanged();
}
```

**2. Trailing commas in the theme `$options` labels.** The last 15 theme entries end with a comma (`- charcoalCream: CharcoalCream,` … `- onyxAmber: OnyxAmber,`), and in YAML that comma is part of the value — so the dropdown will literally show "CharcoalCream,", "NavyLinen,", etc. The first six entries don't have it. Remove the commas (and see the optional section for label polish on the same lines).

**3. Widgets can drift off-screen when the "Widget size" setting changes.** `LogicalDesktopBounds()` depends on `RenderScale()`, so switching Compact → Comfortable shrinks the logical desktop, and widgets the user dragged near the right/bottom edge can end up partly or fully outside the window. Since `WM_NCHITTEST` returns `HTTRANSPARENT` outside widget bounds, a fully off-screen widget can't even be dragged back — it's lost until a display change or restart. `ApplyOrientationResizeIfNeeded` already clamps on orientation changes; also clamp every widget after `LoadSettings()` in the `kSettingsChangedMessage` handler:

```cpp
LoadSettings();
ApplyOrientationResizeIfNeeded(...);
for (Widget& widget : g_widgets) {
    ClampWidgetToDesktop(widget);
}
Render();
```

<p>

Minor polish — none of this affects users much, so it's your call. The first four carry over from the previous review of pull request 4766.

- **Don't recreate the audio endpoint every second.** The 1 s `WM_TIMER` still does a fresh `CoCreateInstance(MMDeviceEnumerator)` + `GetDefaultAudioEndpoint` + `Activate` on every tick (and on every `WM_MOUSEMOVE` while dragging the slider) just to read/write one float. Cache the `IAudioEndpointVolume` and recreate it only when the default device changes (`IMMNotificationClient`), or subscribe to `IAudioEndpointVolumeCallback` and drop the volume polling entirely. Relatedly, consider skipping the full re-render when nothing visible changed — each tick currently allocates a work-area-sized 32-bpp DIB and redraws everything with GDI+, even though the drawn state only changes when the volume/battery/minute value changes.
- `volatile bool g_unloading` — prefer `std::atomic<bool>` for cross-thread signalling.
- `#include <shlobj.h>` appears unused (`ShellExecuteW` comes from `<shellapi.h>`).
- `OpenMusicFolder()` can pass `L"shell:My Music"` directly to `ShellExecuteW` as the file, without launching `explorer.exe` explicitly.
- **Theme label polish** (same lines as item 2): the newer labels have no spaces ("CharcoalCream", "NavyLinen", …) while the earlier ones do ("White transparent", "Pass Over") — consider "Charcoal Cream" etc. for consistency. Also "Crimsom" is a typo for "Crimson" (at least fix the display label; the stored value is invisible to users, but if you fix it too, keep the code's comparison string in sync).
- **"Widget size" percentages don't add up**: internally Comfortable is 0.75 and Compact is 0.5625, so Compact is 75% of Comfortable — but the labels say "100%" and "56.25%". Whichever baseline you intend, make the two labels consistent (e.g. "Compact (75%)" if Comfortable is the 100% reference).
- `DwmExtendFrameIntoClientArea` with `MARGINS margins = {-1}` only sets `cxLeftWidth` (the sheet-of-glass idiom is all four members at -1), and since the window is painted via `UpdateLayeredWindow` the DWM frame extension has no visible effect anyway — it can likely be removed, along with `-ldwmapi` if nothing else needs it.

</p>
</details>

<p>

Non-critical observations and ideas about the feature behavior itself — the first three carry over from the previous review.

- **Widget positions still aren't persisted.** Dragging updates `g_widgets[i].rc` in memory only, so every reload resets the layout — and `WM_DISPLAYCHANGE` calls `ApplyDefaultWidgetLayout()`, so a resolution change (or monitor sleep/wake on some setups) also discards the user's arrangement. Consider persisting positions with `Wh_SetIntValue`/`Wh_GetIntValue` and restoring them (clamped) instead of resetting.
- **Single-monitor / DPI assumptions.** The panel covers the primary monitor's work area only, and `UpdateScale()` derives scale from `GetSystemMetrics(SM_CXSCREEN/CYSCREEN)` against a hard-coded 1920×1080. On multi-monitor or per-monitor-DPI setups this can scale oddly — no trivial fix, just an FYI.
- **`OpenWindowsSearch()` synthesizes Win+S via `SendInput`.** Works, but injecting global keystrokes can race with the user's real modifier state. There's no clean public API for the search flyout, so this may remain the pragmatic choice.
- **`g_onBatteryCapable` is computed but never read.** On a desktop PC the battery widget just shows a full, non-interactive bar forever. Since you already detect the no-battery case, consider using it — hide the battery widget (and reflow), or draw a "no battery" state.
- **The volume-slider hit test runs before the widget z-order hit test** (`WM_LBUTTONDOWN` checks `VolumeSliderRect()` first). If the user drags another widget on top of the volume slider, clicks in the overlap still adjust the volume instead of hitting the top widget.

</p>
</details>

Stunning-dev:
close this pull request

PR #4907, opened 2026-07-26
Create dark-theme-flash-fix.wh.cpp

grandrange:
## Changelog

* Changelog item 1...
* Changelog item 2...

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/dark-theme-flash-fix.wh.cpp
+++ b/mods/dark-theme-flash-fix.wh.cpp
@@ -52,6 +52,8 @@ App calls ShowWindow(hWnd)
 - The delay adds a small perceived latency.
   I recommend 100–200 ms it blends with the native Windows 11
   window animation and feels natural.
+- In some programs, the mod may break the window appearance animation.
+  I do not recommend applying the mod globally.
 */
 // ==/WindhawkModReadme==
 
--- a/mods/dark-theme-flash-fix.wh.cpp
+++ b/mods/dark-theme-flash-fix.wh.cpp
@@ -4,6 +4,7 @@
 // @description Eliminates white flash when opening windows in dark mode via DWM cloaking
 // @version 1.0
 // @author grandrange
+// @github https://github.com/grandrange
 // @compilerOptions -lGdi32
 // ==/WindhawkMod==
 
--- a/mods/dark-theme-flash-fix.wh.cpp
+++ b/mods/dark-theme-flash-fix.wh.cpp
@@ -5,7 +5,6 @@
 // @version 1.0
 // @author grandrange
 // @github https://github.com/grandrange
-// @compilerOptions -lGdi32
 // ==/WindhawkMod==
 
 // ==WindhawkModReadme==
--- a/mods/dark-theme-flash-fix.wh.cpp
+++ b/mods/dark-theme-flash-fix.wh.cpp
@@ -1,0 +1,153 @@
+// ==WindhawkMod==
+// @id dark-theme-flash-fix
+// @name Dark Theme Flash Fix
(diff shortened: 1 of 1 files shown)


Anixx:
How is this different from this?
https://windhawk.net/mods/fix-explorer-white-flash

Undisputed00x:
Actaully, this doesn't work, for example, explorer dialogs still briefly display a white background when opened. The internal function that paints the background white uses the system brush COLOR_WINDOW (which is white by default) to paint the background white on the first frame of every window in the system, but the problem is that bypasses GetSysColorBrush API and retrieves the system brush pointer directly from the memory in the kernel. The function that causes this effect is likely located in the kernel. I've seen this happen with windows that use system colors/brushes directly from memory instead of using the getter APIs. This is what i have found in Translucent Windows mod: https://github.com/ramensoftware/windhawk-mods/blob/109589023dde428deaee2fe80e4ce446283a7935/mods/translucent-windows.wh.cpp#L5463
As a result this is also related to this: https://github.com/ramensoftware/windhawk-mods/issues/2010#issuecomment-3544379735

grandrange:
The Explorer fix mod and my mod solve completely different problems. That mod fixes the flash when opening new tabs in an already running program, while my mod fixes the flash when launching a program. Programs like OBS Studio or qBittorrent always flash when launching or opening from the tray. This mod solves this problem, and yes, it works. I've personally tested it and use it, so I decided to share it, as it's quite an annoying issue if you use the dark theme in Windows 11.

Undisputed00x:
@grandrange You forgot to add `// @include *`, but in any case, this is not a **fix** but rather a workaround, as it actually adds a visual delay when the window opens and as a big disadvantage the opening animation is skipped completely even when the delay setting is set to 1ms. Claude generated the same code for me when I asked it for a possible solution to this problem a while ago.

grandrange:
> @grandrange You forgot to add `// @include *`, but in any case, this is not a **fix** but rather a workaround, as it actually adds a visual delay when the window opens and as a big disadvantage the opening animation is skipped completely even when the delay setting is set to 1ms. Claude generated the same code for me when I asked it for a possible solution to this problem a while ago.

I did this on purpose so that by default the mod won't affect any process, as the idea is to only add the processes that have the problem.

In 99% of cases, the flash occurs in the first milliseconds of startup. A delay of 100-200 milliseconds doesn't significantly affect anything, but it does remove the flash. I know this doesn't fix the root of the problem, because the dark theme in Windows 11 is inherently poorly designed, but it works and doesn't break anything, unlike editing the theme file itself.

grandrange:
This mod only works with window opening. Opening windows in Windows 11 doesn't have a smooth animation, like minimizing and maximizing a window to the taskbar. The mod doesn't create a delay before the window opens, but instead makes the window invisible for a few milliseconds. This makes the window appear more smoothly, as it's already fully rendered.

Undisputed00x:
@grandrange Here is a video example of how the mod completely hides the windows opening animation even on 1ms setting.

<details><summary>Video</summary>
<p>


https://github.com/user-attachments/assets/bd425fb2-a377-42aa-80d4-ebe7dd7268b4



</p>
</details>

grandrange:
> @grandrange Here is a video example of how the mod completely hides the windows opening animation even on 1ms setting.
> 
> Video

That's an interesting observation, thank you for sharing it. That's why I didn't do: // @include * The mod doesn't need to be applied globally, only to those programs that flash when the app opens.

I understand this is a workaround, but it's better than changing the Windows 11 theme and breaking the display in other Win 32 apps.

PR #4908, opened 2026-07-26
Add instant-taskbar-thumbnail-previews.wh.cpp

alchemyyy:
## Changelog

## Mod authorship

This mod removes / controls the delays around "taskbar thumbnail previews" i.e. these:

image

It also contains two optional features:

* A toggle which instantly closes open thumbnail previews when clicking outside the taskbar and thumbnail surface
* A toggle which instantly closes open thumbnail previews when hovering over the Start button

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/instant-taskbar-thumbnail-previews.wh.cpp
+++ b/mods/instant-taskbar-thumbnail-previews.wh.cpp
@@ -1,0 +1,1318 @@
+// ==WindhawkMod==
+// @id              instant-taskbar-thumbnail-previews
+// @name            Instant Taskbar Thumbnail Previews
+// @description     Controls taskbar thumbnail preview show and close behavior.
+// @version         1.0.0
+// @author          Alchemy
+// @github          https://github.com/alchemyyy
+// @license         MIT
+// @include         explorer.exe
+// @architecture    x86-64
+// @compilerOptions -lole32 -loleaut32 -lruntimeobject
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Instant Taskbar Thumbnail Previews
+
+Adds the ability to configure or remove the delays before Windows 11 shows and closes taskbar thumbnail previews,
+as well as the ability to configure mouse actions that close them.
+
+Both delay settings replace the corresponding Windows value, so they can make the transitions either shorter or longer than the stock delays.
+
+The optional outside-click behavior arms the taskbar's native light-dismiss action while a thumbnail flyout is open.
+
+The optional Start-button hover behavior closes an open thumbnail flyout directly from the Start button's XAML pointer-enter handler.
+
+This mod targets the new Windows 11 taskbar used by Windows 11 24H2 and later.
+
+*/
+// ==/WindhawkModReadme==
+
+// ==WindhawkModSettings==
+/*
+- delayMs: 1
+  $name: Thumbnail hover delay
+  $description: Exact delay in milliseconds. The minimum value is 1, which is effectively immediate.
+- closeDelayMs: 1
+  $name: Thumbnail close delay
+  $description: Exact delay after the pointer leaves the thumbnail area. The minimum value is 1, which is effectively immediate.
+- closeOnOutsideClick: false
+  $name: Close on outside click
+  $description: Immediately close open thumbnail previews when clicking outside the taskbar and thumbnail surface.
+- closeOnStartButtonHover: false
+  $name: Close on Start button hover
+  $description: Immediately close open thumbnail previews when hovering over the Start button.
+*/
+// ==/WindhawkModSettings==
+
+#include <windhawk_utils.h>
+
+#include <windows.h>
+
+#include <atomic>
+#include <unordered_set>
+
+#undef GetCurrentTime
+
+#include <winrt/Windows.UI.Xaml.Automation.h>
+#include <winrt/Windows.UI.Xaml.h>
+
+#define MINIMUM_DELAY_MS 1
+#define TIME_SPAN_TICKS_PER_MILLISECOND 10000LL
+
+namespace {
+
+std::atomic<LONGLONG> g_hoverDelayTimeSpan{
+    MINIMUM_DELAY_MS * TIME_SPAN_TICKS_PER_MILLISECOND};
+std::atomic<UINT> g_closeDelayMilliseconds{MINIMUM_DELAY_MS};
+std::atomic<bool> g_closeOnOutsideClick{false};
+std::atomic<bool> g_closeOnStartButtonHover{false};
+std::atomic<bool> g_taskbarHooksQueued{false};
+std::atomic<bool> g_taskbarViewHooksQueued{false};
+std::atomic<bool> g_initialized{false};
+std::atomic<bool> g_windowsLightDismissTrackingReliable{true};
+SRWLOCK g_windowsLightDismissTaskbarHostsLock = SRWLOCK_INIT;
+std::unordered_set<void*> g_windowsLightDismissTaskbarHosts;
+thread_local bool g_inDismissTransition = false;
+thread_local HWND g_activeTaskbarWindow = nullptr;
+thread_local bool g_previewLightDismissArmed = false;
+thread_local HWND g_previewLightDismissTaskbarWindow = nullptr;
+thread_local UINT_PTR g_previewLightDismissRearmTimer = 0;
+thread_local void* g_activeHoverFlyoutController = nullptr;
+thread_local void* g_pointerOverFlyoutFrameModel = nullptr;
+thread_local void* g_lastUnregisteredTaskbarHost = nullptr;
+
+void* CTaskBand_ITaskListWndSite_Vftable;
+void* CSecondaryTaskBand_ITaskListWndSite_Vftable;
+
+using CTaskBandGetTaskbarHost_t =
+    void*(WINAPI*)(void* object, void** taskbarHostSharedPointer);
+CTaskBandGetTaskbarHost_t CTaskBandGetTaskbarHost_Original;
+CTaskBandGetTaskbarHost_t CSecondaryTaskBandGetTaskbarHost_Original;
+
+using ReferenceCountBaseDecrement_t = void(WINAPI*)(void* object);
+ReferenceCountBaseDecrement_t ReferenceCountBaseDecrement_Original;
+
+using TaskbarHostRegisterLightDismiss_t = void(WINAPI*)(void* object);
+TaskbarHostRegisterLightDismiss_t TaskbarHostRegisterLightDismiss_Original;
+
+using TaskbarHostUnregisterLightDismiss_t = void(WINAPI*)(void* object);
+TaskbarHostUnregisterLightDismiss_t TaskbarHostUnregisterLightDismiss_Original;
+
+using TaskbarControllerOnLightDismissTriggered_t = void(WINAPI*)(
+    void* object,
+    const void* taskbarModel,
+    const void* eventArguments);
+TaskbarControllerOnLightDismissTriggered_t
+    TaskbarControllerOnLightDismissTriggered_Original;
+
+constexpr int TASK_BAND_INTERFACE_SEARCH_LIMIT = 20;
+
+enum class TaskbarWindowType {
+    invalid,
+    primary,
+    secondary,
+};
+
+TaskbarWindowType GetTaskbarWindowType(HWND taskbarWindow) {
+    WCHAR className[32] = {};
+    if (!GetClassNameW(taskbarWindow, className, ARRAYSIZE(className))) {
+        return TaskbarWindowType::invalid;
+    }
+
+    if (_wcsicmp(className, L"Shell_TrayWnd") == 0) {
+        return TaskbarWindowType::primary;
+    }
+
+    if (_wcsicmp(className, L"Shell_SecondaryTrayWnd") == 0) {
+        return TaskbarWindowType::secondary;
+    }
+
+    return TaskbarWindowType::invalid;
+}
+
+HWND NormalizeTaskbarWindow(HWND window) {
+    if (!window) {
+        return nullptr;
+    }
+
+    HWND rootWindow = GetAncestor(window, GA_ROOT);
+    if (!rootWindow) {
+        rootWindow = window;
+    }
+
+    if (GetTaskbarWindowType(rootWindow) ==
+        TaskbarWindowType::invalid) {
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

---

**1. Raw pointers to taskbar objects are cached indefinitely and can dangle.**

`g_taskbarHost` and `g_activeHoverFlyoutController` hold raw pointers to Explorer-owned objects with no reference held and no invalidation:

* `GetTaskbarHostFromWindow` takes the `shared_ptr<TaskbarHost>` returned by `CTaskBand::GetTaskbarHost`, immediately releases it via `_Ref_count_base::_Decref`, and returns the bare pointer, which `ArmPreviewLightDismiss` then caches in `g_taskbarHost` **forever** and later passes to `TaskbarHost::RegisterLightDismiss`. When a `TaskbarHost` goes away — most easily reproduced by unplugging/replugging a monitor, which destroys that `Shell_SecondaryTrayWnd`'s `CSecondaryTaskBand`/`TaskbarHost` — the next `ArmPreviewLightDismiss` calls a member function on freed memory and takes Explorer down with it.
* Same for `g_activeHoverFlyoutController`, which is stashed in `UpdateFlyoutWindowPosition_Hook` and dereferenced later in `DismissActiveHoverFlyoutImmediately`. It is only cleared if `CommitDismissFlyout` or `HideAllHoverFlyouts` happens to be called for that exact object; any other destruction path leaves a dangling pointer.

The established pattern in this repo is to resolve the host fresh, use it within the call, and never store it — see [hide-start-button.wh.cpp#L124-L146](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/hide-start-button.wh.cpp#L124-L146) and [taskbar-multirow.wh.cpp#L295-L340](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/taskbar-multirow.wh.cpp#L295-L340). Please re-resolve on each use instead of caching.

**2. The taskbar host is picked arbitrarily, so multi-monitor arms the wrong taskbar.**

`FindTaskbarHostForCurrentThread` walks `EnumThreadWindows` and returns the **first** window that yields a host. `Shell_TrayWnd` and `Shell_SecondaryTrayWnd` live on the same Explorer thread (this is why [hide-start-button.wh.cpp#L159-L178](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/hide-start-button.wh.cpp#L159-L178) checks both class names inside a single `EnumThreadWindows`), so on a multi-monitor setup which taskbar you get is z-order luck — and then it's cached for the rest of the session. Since `RegisterLightDismiss` watches *that* taskbar's InputSite, a flyout opened from the other monitor's taskbar is armed against the wrong surface. The host should be derived from the taskbar that actually owns the flyout, not from "whatever `EnumThreadWindows` returned first".

**3. `RegisterLightDismiss` is never balanced with `UnregisterLightDismiss`.**

The mod calls `TaskbarHost::RegisterLightDismiss` itself, but the only place `UnregisterLightDismiss` appears is as a hook used to *observe* Windows' own calls. Consequences:

* When a thumbnail flyout closes normally (pointer leave → `CommitDismissFlyout`), the mod sets `g_previewLightDismissArmed = false` but leaves the registration in place. Windows has no reason to unregister it — it never registered it. So the taskbar stays armed with nothing open, and the next unrelated click anywhere delivers a spurious `TaskbarController::OnLightDismissTriggered`, which can dismiss other taskbar surfaces.
* Because the flag was cleared, the next flyout calls `RegisterLightDismiss` **again** on the same host, so registrations accumulate over a session.
* On unload the registration is left behind, which breaks reversibility — disabling the mod doesn't fully undo its effect on the taskbar.

Please call `TaskbarHostUnregisterLightDismiss_Original` yourself whenever the mod disarms: in `ClearActiveHoverFlyoutController`, in `DismissActiveHoverFlyoutImmediately`, when `closeOnOutsideClick` is turned off in `Wh_ModSettingsChanged`, and in `Wh_ModUninit`.

Note that the current `Wh_ModUninit` body doesn't help here: `g_previewLightDismissArmed`, `g_activeHoverFlyoutController`, `g_pointerOverFlyoutFrame` and `g_taskbarHost` are `thread_local`, and `Wh_ModUninit` runs on an arbitrary Windhawk thread, so those four assignments only clear that thread's copies and never touch the taskbar thread's state. The teardown has to be posted to the taskbar thread (e.g. the `RegisterWindowMessage` + `WH_CALLWNDPROC` trick in [hide-start-button.wh.cpp#L180-L207](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/hide-start-button.wh.cpp#L180-L207)) to actually unregister.

**4. `openThumbnailsOnGroupedClick` duplicates and conflicts with an existing mod.**

[Cycle through taskbar windows on click](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/taskbar-left-click-cycle.wh.cpp#L80-L88) already owns this behavior axis and hooks the same `bool __cdecl ShouldCycleWindows(void)` symbol — it returns `!ctrlKeyDown`, this mod returns `controlKeyDown`. Two mods hooking the same function with opposite intent will fight, and the feature is unrelated to this mod's stated purpose (thumbnail delays). Better to drop it here and propose it as an option (e.g. an inverted mode) on `taskbar-left-click-cycle`, per the usual preference for extending an existing mod over shipping a competing one.

Also, unlike that mod, this hook discards `ShouldCycleWindows_Original()` entirely instead of inverting it, so the result no longer follows whatever the stock policy is for the user's taskbar configuration.

**5. Add a screenshot or GIF to the README.**

The PR body has a good screenshot but the README has none. A short GIF showing the instant show/close is the most useful thing here — the mod's whole point is a timing difference. Only `i.imgur.com` and `raw.githubusercontent.com` are accepted image hosts.

**6. How does this relate to your existing [Taskbar Thumbnail Hover Fix](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/taskbar-thumbnail-hover-fix.wh.cpp) mod?** Both tweak taskbar thumbnail hover behavior; it's worth saying in the README why they're separate mods rather than one, or folding them together.

<p>

* `LightDismissalScope` / `g_committingLightDismissal` looks like dead code. `DismissActiveHoverFlyoutImmediately` calls `CommitDismissFlyout_Original` (not the hook) and already nulls `g_activeHoverFlyoutController` before the call, so the early-out on `!hoverFlyoutController` makes re-entry impossible and the scope guard can never do anything. If it's guarding a case you actually hit, a comment would help; otherwise it can go.
* Similarly, `g_dismissTransitionDepth` is a counter for a path that can't recurse — a `bool` would read more clearly — and the `g_initialized` checks inside `UpdateFlyoutWindowPosition_Hook` / `DismissActiveHoverFlyoutImmediately` are redundant: Windhawk installs hooks only after `Wh_ModInit` returns and removes them before `Wh_ModUninit`, so a hook can't run outside the initialized window.
* `FIND_TASKBAR_HOST_CONTEXT` wraps a single `void*`; passing `&taskbarHost` as the `LPARAM` directly would be simpler.
* `Wh_Log(L"Grouped click policy: thumbnails=%d control=%d cycle=%d", 1, ...)` passes a hardcoded `1` for the first field — presumably a leftover.
* `-luser32` isn't needed; user32 is linked by default. Compare [hide-start-button.wh.cpp#L12](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/hide-start-button.wh.cpp#L12), which uses `GetClassName`/`FindWindowEx`/`SetWindowsHookEx`/`SendMessage` with only `-lole32 -loleaut32 -lruntimeobject`.
* The `Taskbar.dll` symbols are resolved unconditionally even though they're only used by `closeOnOutsideClick`, which defaults to off — that's an avoidable symbol-resolution (and possibly PDB download) cost for most users. You could gate `HookTaskbarSymbols` on the setting and switch to `BOOL Wh_ModSettingsChanged(BOOL* bReload)` with `*bReload = TRUE` when that setting toggles, so enabling it later still installs the hooks.
* When you add the module name/comment that CI is asking for on the two symbol arrays, note that the second one resolves against `Taskbar.View.dll` *or* `ExplorerExtensions.dll`, so both belong in the comment — see [hide-start-button.wh.cpp#L288-L289](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/hide-start-button.wh.cpp#L288-L289).
* `WindhawkUtils::SetFunctionHook` is preferred over raw `Wh_SetFunctionHook` with `reinterpret_cast<void*>` in `HookLoadLibraryExW`; it gives you type checking on the hook signature.
* `HandleLoadedModule` calls `GetModuleHandleW` on every single `LoadLibraryExW` in the process before testing the already-hooked flags; testing the flags first would avoid that.
* The README heading is `# Instant Taskbar Thumbnails` while `@name` is `Instant Taskbar Thumbnail Previews` — worth matching. Also, `TrackMouseEvent` is listed among "registry values", but it's an API.

</p>
</details>

<p>

* `TransitionToFlyoutVisiblePendingState_Hook` replaces the incoming delay unconditionally. If Windows ever passes a short or zero delay for an "already showing, pointer moved to the adjacent button" re-show, the mod will impose the full configured delay on it, which would feel worse than stock for anyone who sets a *longer* delay. Preserving a zero/near-zero incoming value (`if (delayTimeSpan > 0)`) would keep those transitions instant.
* `delayMs` / `closeDelayMs` are clamped to a minimum of 1 — is that deliberate? If `0` works it'd be worth allowing, and if it doesn't, a note in the `$description` would explain why 1 is the floor.
* `SetIsPointerOverFlyoutFrame_Hook` ignores its `object` argument and writes to a single thread-wide flag, and `ClearActiveHoverFlyoutController` resets that flag based on the *controller* identity rather than the model's. If more than one `HoverFlyoutModel` is ever live, the flag can go stale, and the `pointerOverFlyoutFrame` branch in `TaskbarControllerOnLightDismissTriggered_Hook` would then make the wrong call about whether the click landed inside the preview island.
* `UpdateFlyoutWindowPosition` covers all hover flyouts, not just thumbnails — the Task View virtual-desktop switcher flyout goes through `HoverFlyoutController` too (see [taskbar-thumbnails.wh.cpp#L246-L270](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/taskbar-thumbnails.wh.cpp#L246-L270), which has a separate setting for it). So `closeOnOutsideClick` will apply there as well; probably fine, but worth mentioning in the README.
* `TaskbarController::OnLightDismissTriggered` fires for light dismisses the mod didn't arm (Start menu opening, other taskbar flyouts), so an open thumbnail preview will also be closed by those. That's likely desirable, just noting it's broader than "clicked outside".

</p>
</details>

alchemyyy:
1. Removed cached `TaskbarHost` pointers. The host is now resolved fresh for each operation, with its `shared_ptr` retained until the operation completes. Cached controller/model pointers are invalidated through destructor hooks.
2. Flyout ownership is derived from `TaskbarModel::HostWindowId`, then validated as a primary or secondary taskbar owned by the current process and taskbar UI thread. This selects the correct taskbar on multi-monitor systems instead of relying on enumeration order.
3. `RegisterLightDismiss` is now balanced with `UnregisterLightDismiss` on normal dismissal, immediate dismissal, setting disable, and mod unload. Teardown is dispatched synchronously to the actual taskbar thread.
4. Removed `openThumbnailsOnGroupedClick`, its setting, and the conflicting `ShouldCycleWindows` hook.
5. Added the PR screenshot to the README.
6. Documented how this mod differs from and can be used alongside Taskbar Thumbnail Hover Fix.

Also addressed the optional and functionality notes:

- Removed the unused light-dismiss re-entry guard.
- Replaced the dismissal depth counter with a boolean.
- Removed the obsolete taskbar-host search context and grouped-click logging.
- Removed the unnecessary `-luser32`.
- Taskbar.dll symbol resolution is now conditional on `closeOnOutsideClick`; changing that setting reloads the mod.
- Replaced raw `Wh_SetFunctionHook` calls with `WindhawkUtils::SetFunctionHook`.
- Avoided redundant module lookups after hooks are queued.
- Corrected the README heading and `TrackMouseEvent` wording.
- Preserved Windows’ non-positive immediate-transition delays.
- Documented the deliberate 1 ms minimum.
- Pointer-over state now tracks model identity and is cleared when that model is destroyed.
- Documented Task View flyouts and broader native light-dismiss behavior.

Validated with x64 and ARM64 builds and the repository PR validation script.

m417z:

More details:

alchemyyy:
1. **Light-dismiss cleanup**
   - Every registration requested by the preview path is now paired with `UnregisterLightDismiss` when the preview closes, changes taskbars, the relevant setting reloads, or the mod unloads.
   - The exact taskbar window used for registration is retained so multi-monitor transitions disarm the previous host before arming the next.
   - `CleanupTaskbarThreads` now performs real cleanup on each taskbar thread, justifying the temporary thread-specific `WH_CALLWNDPROC` machinery.
   - I verified the implementation of `RegisterLightDismiss`: it is idempotent and returns immediately when an action already exists. Repeated calls therefore neither allocate additional actions nor replace an existing action.
   - Since the function returns `void`, a successful call cannot formally prove ownership when Windows already has an action. Reliably distinguishing that case would require decoding the function’s null guard and inspecting a private `TaskbarHost` member. I rejected that approach as too fragile across Windows/compiler versions and documented the trade-off in the code.

2. **Process-wide hooks removed**
   - Removed the process-wide `DispatchMessageW` hook. Deferred rearming now uses a taskbar-thread `SetTimer` callback, allowing the current pointer dispatch to unwind first.
   - Removed the process-wide `MonitorFromWindow` hook. The associated taskbar is now found using `GetCursorPos`, `WindowFromPoint`, `GetAncestor`, and a monitor-matched current-thread taskbar fallback.

3. **Optional hooks**
   - Taskbar.View hooks are now assembled dynamically according to the enabled settings while retaining a single `HookSymbols` call.
   - The default configuration installs only the three timing hooks.
   - Taskbar light-dismiss hooks are installed only for `closeOnOutsideClick`.
   - Both `closeOnOutsideClick` and `closeOnStartButtonHover` now request a reload when changed so the installed hook set remains accurate.

4. **Lifecycle and state fixes**
   - `DismissTransitionScope` now saves and restores its previous state, making nested transitions safe.
   - Failed symbol-resolution attempts leave their queued flags set, preventing repeated resolution and symbol-cache invalidation for the same module.
   - Removed the queued-flag resets from `Wh_ModUninit`, where they had no effect.
   - Retained the `catch (...)` blocks only for C++/WinRT exceptions; they are not treated as protection against access violations.

PR #4909, opened 2026-07-26
Make the mod much shorter, control band appearance via context menu (…

Anixx:
…instead of settings)

Make the mod much shorter, control band appearance via context menu (instead of settings)

## Changelog

* Make the code 2.5 times shorter
* The toolbar appearance is now controlled via toolbar context menus, not via mod's settings.

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/flexible-explorer-toolbars-deluxe.wh.cpp
+++ b/mods/flexible-explorer-toolbars-deluxe.wh.cpp
@@ -376,11 +376,6 @@ void ToggleBand(HWND cab, BandType type, bool enable) {
             ForceCabinetRelayout(cab);
             PostMessage(cab, g_msgFixContent, (WPARAM)ch, 0);
             SaveBandPositions(mr, false);
-        } else {
-            // вставка не удалась — вернуть в скрытое состояние
-            SetPropW(ch, L"FlexTbIsHidden", (HANDLE)1);
-            ShowWindow(ch, SW_HIDE);
-            SetHiddenBand(cab, type, ch);
         }
     }
 }
--- a/mods/flexible-explorer-toolbars-deluxe.wh.cpp
+++ b/mods/flexible-explorer-toolbars-deluxe.wh.cpp
@@ -27,8 +27,6 @@ This mod hides the Navigation Bar and instead creates the following optional too
 * The Breadcrumbs Bar
 * The Up Buton
 
-The toolbars can be enabled and disabled via context menus.
-
 The toolbars can be locked and unlocked.
 If you are using this mod together with Classic Explorer toolbar (Open Shell), enable that toolbar before enabling this mod, otherwise its enabled state will not be remembered.
 
--- a/mods/flexible-explorer-toolbars-deluxe.wh.cpp
+++ b/mods/flexible-explorer-toolbars-deluxe.wh.cpp
@@ -2,7 +2,7 @@
 // @id              flexible-explorer-toolbars-deluxe
 // @name            Flexible Explorer Toolbars Deluxe
 // @description     Makes Search Bar, Breadcrumb Bar and others into movable toolbars
-// @version         1.3.4
+// @version         1.3.6
 // @author          Anixx
 // @github          https://github.com/Anixx
 // @include         explorer.exe
@@ -99,7 +99,10 @@ template<typename F> void EnumBands(HWND rb, UINT mask, F f) {
 
 int GetSavedRank(const wchar_t* cls) { WCHAR ok[160]; swprintf(ok,160,L"OrderRank_%s",cls); return Wh_GetIntValue(ok, INT_MAX); }
 void SetSavedRank(const wchar_t* cls, int r) { WCHAR ok[160]; swprintf(ok,160,L"OrderRank_%s",cls); Wh_SetIntValue(ok, r); }
-struct BandState{UINT cx; bool brk;};
+
+// FIX #2: default-initialize so any accidental use-before-LoadBandState() is well-defined
+// instead of indeterminate (width/style must never end up random).
+struct BandState { UINT cx = 0; bool brk = false; };
 bool LoadBandState(const wchar_t* cls, BandState& out) {
     WCHAR ck[160], bk[160]; swprintf(ck,160,L"Cx_%s",cls); swprintf(bk,160,L"Break_%s",cls);
     int cx = Wh_GetIntValue(ck, -1); if(cx<20 || cx>8000) return false;
@@ -163,6 +166,19 @@ void ApplySavedLayout(HWND rb) {
 HWND FindByClass(HWND p, LPCWSTR c) { return FindWindowEx(p, NULL, c, NULL); }
 HWND GetCabinet(HWND h) { while(h){ WCHAR c[64]; if(GetClassName(h,c,64) && !wcscmp(c,L"CabinetWClass")) return h; h=GetParent(h); } return NULL; }
 HWND FindNavRb(HWND c) { for(HWND w=FindByClass(c,L"WorkerW"); w; w=FindWindowEx(c,w,L"WorkerW",NULL)) if(HWND r=FindByClass(w,L"ReBarWindow32")) return r; return NULL; }
+
+// FIX #3: enumerate only this process's CabinetWClass windows instead of HWND_BROADCAST,
+// which would hit every top-level window in every process on the desktop.
+BOOL CALLBACK EnumSyncCabinets_Proc(HWND w, LPARAM) {
+    DWORD pid = 0; GetWindowThreadProcessId(w, &pid);
+    if(pid == GetCurrentProcessId()) {
+        WCHAR c[64];
+        if(GetClassName(w, c, 64) && !wcscmp(c, L"CabinetWClass"))
+            PostMessage(w, g_msgSyncSettings, 0, 0);
+    }
+    return TRUE;
+}
+void PostSyncToAllCabinets() { EnumWindows(EnumSyncCabinets_Proc, 0); }
 
 BOOL CALLBACK EnumFindClass_Proc(HWND h, LPARAM l) {
     auto* p = (std::pair<LPCWSTR, bool*>*)l; WCHAR c[256];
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

Most notes are about the mod functionality, I haven't verified them. Let me know if you prefer to skip some of them.

---

**1. `ProcessWnd` now force-hides the menu-bar `WorkerW`, not just the navigation-bar one.**

```cpp
else if(!wcscmp(c,L"WorkerW") && GetCabinet(h)) { HookWindow(h,WorkerW_Proc); ForceHideWorker(h); }
```

`GetCabinet()` walks the whole ancestor chain, so this also matches `CabinetWClass > ShellTabWindowClass > WorkerW` — which is exactly the window that hosts the menu rebar the mod moves its toolbars into (`mRb`, in the `g_msgDoMove` handler). Once `FlexTbForceHidden` is set on it, `WorkerW_Proc` pins it hidden permanently (`WM_STYLECHANGING` / `WM_WINDOWPOSCHANGING` / `WM_SHOWWINDOW`), so the menu bar and every moved toolbar go invisible. It's reachable both from `Wh_ModInit`'s `EnumChildWindows` (i.e. enabling the mod while an Explorer window is already open) and from `Hook_CWExW`.

v1.1 required the *direct* parent to be `CabinetWClass` ([v1.1 L970-L977](https://github.com/ramensoftware/windhawk-mods/blob/e6e6322957e02141338ec74bd7320b4b6537f205/mods/flexible-explorer-toolbars-deluxe.wh.cpp#L970-L977)); please restore that:

```cpp
else if(!wcscmp(c,L"WorkerW")) {
    HWND p = GetParent(h); WCHAR pc[64];
    if(p && GetClassName(p,pc,64) && !wcscmp(pc,L"CabinetWClass")) {
        HookWindow(h,WorkerW_Proc); ForceHideWorker(h);
    }
}
```

**2. `Rb_Proc` swallows `WM_CONTEXTMENU` for every `ReBarWindow32` in explorer.exe, including the taskbar's.**

```cpp
if(m == WM_CONTEXTMENU) {
    ...
    if(HWND cab = GetCabinet(h))
        if(HWND ww = ...) ...
    return 0;   // <-- outside the `if (cab)`
}
```

`ProcessWnd` subclasses *every* rebar in the process (no cabinet check, unlike `ShellTabWindowClass`/`WorkerW`), and the taskbar is `Shell_TrayWnd > ReBarWindow32` on Windows 10 (and on Windows 11 with ExplorerPatcher). There `GetCabinet()` returns `NULL`, no menu is shown — and the message is eaten anyway, so right-clicking the taskbar rebar (Quick Launch / other deskbands / empty rebar area) does nothing at all. The last commit ("don't intercept the menus of taskbar toolbars") added the cabinet check but not the fall-through to `DefSubclassProc`.

Two fixes needed: scope the subclass to cabinet rebars in `ProcessWnd`, and only `return 0` when the menu was actually shown.

Related: v1.1 hit-tested the band and only took over the menu when the band was one of the mod's own (`RB_HITTEST` + `CF_MOVED`, [v1.1 L688-L723](https://github.com/ramensoftware/windhawk-mods/blob/e6e6322957e02141338ec74bd7320b4b6537f205/mods/flexible-explorer-toolbars-deluxe.wh.cpp#L688-L723)). The new code replaces the context menu of *any* band in a cabinet rebar, so e.g. Open-Shell's Classic Explorer Bar loses its own menu. Please keep the hit-test.

**3. `ToggleBand` uses an uninitialized `BandState`.**

```cpp
WCHAR c[256]; GetEffClass(ch, c, 256); BandState bs; LoadBandState(c, bs);
...
rbi.fStyle = GetRefGripper(mr) | (bs.brk ? RBBS_BREAK : 0);
rbi.hwndChild=ch; ...; rbi.cx=rbi.cxIdeal=bs.cx;
```

`struct BandState{UINT cx; bool brk;};` has no initializers and `LoadBandState`'s return value is ignored, so when there is no stored `Cx_<class>` the band is inserted with an indeterminate width and break flag. This is hit on the very first enable from the new context menu: widths are only written by `SaveBandPositions(rb, /*saveWidth=*/true)`, and the initial move plus every `ToggleBand` call pass `false` — so a band that starts hidden (breadcrumb and up button, which are now off by default) has no saved width the first time it's shown.

```cpp
struct BandState { UINT cx = 0; bool brk = false; };
...
BandState bs;
if (!LoadBandState(c, bs)) {
    bs.cx = (type == BandType::UpButton) ? 30 : (type == BandType::Breadcrumb ? 250 : 200);
    bs.brk = true;
}
```

(The same defaults the `g_msgDoMove` path already uses via `hasSv`.)

**4. Moving visibility out of the settings into the context menu can leave the user with no way back.**

The only UI that can re-enable a toolbar is the right-click menu of a band in the menu rebar. If the menu bar is disabled (the README explicitly treats it as optional) and the user unchecks the last visible band, the rebar has no bands left to right-click and there is no setting to fall back to — the mod can then only be recovered by resetting its storage. Note also that the search band is not subclassed (`if(f != CF_SEARCH) HookWindow(ch, Tbar_Proc);`), so with the default configuration (search bar only, menu bar off) right-clicking the search box gives the edit control's menu, not the toolbar menu.

Either keep the three toggles in `==WindhawkModSettings==` as well, or refuse to hide the last remaining band.

**5. Nothing is undone on unload.**

`Wh_ModUninit` only removes the subclasses. Disabling the mod leaves the navigation bar `WorkerW` hidden, the search/breadcrumb/up windows reparented into the menu rebar, the `Address Band Root` neutered, and all `FlexTb*` window properties set — so the Explorer window stays broken until it is closed and reopened. This is pre-existing behaviour from v1.1, but it is worth fixing now that the mod is being restructured: move the bands back to the nav rebar, clear the `FlexTb*` props, and drop `FlexTbForceHidden` + re-show the worker. A mod's effects should disappear when it is disabled.

<p>

* Use `WindhawkUtils::SetFunctionHook()` instead of raw `Wh_SetFunctionHook` with `void*` casts — it is type-safe and would catch a signature mismatch for free:
  ```cpp
  WindhawkUtils::SetFunctionHook(TrackPopupMenu, Hook_TPM, &origTPM);
  ```
* `Wh_SetFunctionHook` return values and the `RegisterWindowMessage` results aren't checked. v1.1 returned `FALSE` from `Wh_ModInit` when `RegisterWindowMessage` failed; that check was dropped (`g_msgSyncSettings == 0` would then alias `WM_NULL` in `Cab_Proc`). `GetProcAddress(..., "NtSetValueKey")` is also passed to `Wh_SetFunctionHook` unchecked.
* `PostMessage(HWND_BROADCAST, g_msgSyncSettings, 0, 0)` broadcasts to every top-level window on the desktop, in every process. Only this process's `CabinetWClass` windows care — enumerate them instead (you already do exactly that in `Wh_ModInit`).
* `GetCabinet` walks with `GetParent`, which returns the *owner* for top-level windows and can therefore walk out of the window's own tree. `GetAncestor(h, GA_PARENT)` stops at the real parent.
* `swprintf(ok, 160, L"OrderRank_%s", cls)` — `cls` comes from `GetClassName` into a 256-char buffer, so the key can be silently truncated. Either size the key buffers accordingly or build them with `std::wstring`/`std::format`.
* `GetRefGripper` keeps overwriting `style` inside the enumeration, so it ends up with the *last* non-moved band's gripper style rather than the first one found; an early-out would make the intent clearer.
* `CMD_TOGGLE_*` = `0xF101`-`0xF103` sit in the range Windows reserves for system commands, and they are injected into a menu owned by ExplorerFrame. Picking IDs from an unused lower range reduces the chance of colliding with a future ExplorerFrame item.
* Missing includes for symbols used directly: `<climits>` (`INT_MAX`) and `<cwctype>` (`towlower`) — currently relying on transitive includes.
* The identifiers are now down to 1-2 characters (`g_set.s/b/u`, `mv`, `bs`, `hh/mm/ww/ll`, `h2`, `c2`), with several statements per line and nested ternaries. Shorter isn't a goal in itself, and it makes the mod noticeably harder to review and to maintain — v1.1's naming was easier to follow.

</p>
</details>

<p>

* **Upgrade path.** v1.1 read `MoveSearchBand`/`MoveBreadcrumb`/`MoveUpButton` via `Wh_GetIntSetting`, with all three defaulting to `true`. The new code reads the same names via `Wh_GetIntValue`, which is a completely different store, and defaults to search-only — so existing users silently lose the breadcrumb and up bands on update. Worth a note in the README, or seeding the storage values from a first-run default of "all on".
* **`Hook_TPM`/`Hook_TPMEx` change menu semantics.** They force `TPM_RETURNCMD` and then `PostMessage` the `WM_COMMAND`, whereas `TrackPopupMenu` normally delivers it synchronously before returning. That applies to *every* cabinet-owned menu containing command 41484, not just to the three inserted items. It also ignores `TPM_NONOTIFY` (a caller that asked for no notification now gets one) and drops the high word of the command id via `MAKEWPARAM(LOWORD(cmd), 0)`. Forwarding through the original unmodified when the command isn't one of yours would avoid all of that.
* **Menu items are inserted into a menu the mod doesn't own.** `SyncMenu` calls `InsertMenuW` on the `HMENU` passed to `TrackPopupMenu`. In your own `Rb_Proc`/`Tbar_Proc` path the menu is freshly loaded and destroyed, but in the Explorer-initiated path it is ExplorerFrame's menu — if that handle is cached, the three items stay in it, including after the mod is disabled (at which point clicking them does nothing).
* **Hardcoded ExplorerFrame resource IDs** (menu 264, strings 34304 / 49952 / 14347, command 41484) are fragile across builds, and "Address"/"Up" are borrowed from unrelated resources, so the labels may not match the bands on some languages/builds. The English fallbacks in `GetLocalStr` cover the missing-resource case, but not the wrong-string case.
* **`Hook_NtSet` ignores the key handle** — it drops *any* `ITBar7Layout` write anywhere in explorer.exe, not just the one under the Explorer toolbar key. Checking the key (e.g. via `NtQueryKey`) would narrow the blast radius.
* The README still shows only the v1.1 screenshots; a shot of the new context menu with the three checkable items would help, since that is now the only way to configure the mod.
* Small README typos, in case you're touching it anyway: "curently", "shoud", "altough", "Comand", "Buton", "screnshot".

</p>
</details>

m417z:

More details:

Anixx:
/ready-for-reviewer

I did not address the issue of the toolbar may be disabled so there would be no way to bring them back because of the following:
1. The menubar is activated elsewhere
2. To bring the search bar back one has to reset the mod (disable and re-enable)

m417z:
OK, I'll take a look soon.

Anixx:
I do not undo everything in the navbar on mod unload because it gets restored on the window reopening.

PR #4914, opened 2026-07-27
Publish the first version of the mod

babamohammed2022:
This is the first version of the mod. If possible, point out all issues with it so that I can enhance it with the feedback given.

## Changelog

* Changelog item 1...
* Changelog item 2...

## Mod authorship

This mod was created by:

- - [X] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/aero-flip3d-recreation.wh.cpp
+++ b/mods/aero-flip3d-recreation.wh.cpp
@@ -1,78 +1,90 @@
 // ==WindhawkMod==
 // @id              aero-flip3d-recreation
 // @name            Aero Flip 3D Recreation
-// @description     This mod recreates the Windows Vista/7 Aero Flip 3D for modern Windows versions
+// @description     This mod recreates the classic Windows Vista/7 Flip 3D effect in modern Windows versions
 // @version         1.0.0
 // @author          babamohammed
 // @github          https://github.com/babamohammed2022
-// @include         explorer.exe
+// @include         windhawk.exe
 // @architecture    x86-64
-// @compilerOptions -ldwmapi -luser32 -lgdi32 -lmsimg32
+// @compilerOptions -ldwmapi -luser32 -lgdi32 -lmsimg32 -lwinmm
 // ==/WindhawkMod==
 
 // ==WindhawkModReadme==
 /*
 # Aero Flip 3D Recreation
 
-This mod recreates the Windows Vista/7 **Flip 3D** window switcher on Windows 10/11.
-The mod is based on live DWM thumbnails and only public, documented Windows APIs and system files are not modified.
-Note: This is a first version so it might not be 100% accurate but it will improve over time with feedback.
+This mod recreates the classic Windows Vista/7 Flip 3D window switcher on modern Windows versions.
+
+It keeps window previews live and displays them in a 3D-style cascade similar to the original effect.
+
+This is the first version, so some details may still be improved. To help the author improve the mod, feel free to send suggestions.
+
 The mod has been tested on Windows 10 1809.
 
 ## Keyboard shortcuts
 
 | Shortcut | Action |
 | --- | --- |
-| `Win+Tab` | Open Flip 3D. Press `Tab` to cycle through windows; keep `Win` held to auto-cycle like the original. **Releasing `Win` activates** the selected window. |
-| `Ctrl+Win+Tab` | Open Flip 3D in **persistent mode**: the switcher stays open after you release the keys. |
-| `Ctrl+G` or `Ctrl+Alt+F12` | Toggle the switcher on/off (persistent mode). Useful fallback if `Win+Tab` is taken by another app. |
+| `Win+Tab` | Opens Aero Flip 3D |
+| `Ctrl+Alt+F12` | Opens or closes Flip 3D manually. Use this as a fallback. |
 
-### While the switcher is open
+### While Flip 3D is open
 
-- **Navigate**: `Tab`, the arrow keys, or the mouse wheel.
-- **Confirm**: `Enter`, `Space`, or click.
-- **Cancel**: `Esc`.
-- **Emergency exit**: `Ctrl+Shift+Esc` always reaches Task Manager.
+- Navigate: `Tab`, arrow keys, or mouse wheel.
+- Select: `Enter` or `Space`.
+- Close: `Esc`.
+- Emergency exit: `Ctrl+Shift+Esc` opens Task Manager.
 
 ## Settings
 
-- **Realistic perspective cards** (`usePerspective`, on by default): deform each
-  card into a perspective trapezoid (Windows 7-like). Uncheck it for the
-  original **flat** 2.5D card cascade. The change applies live if the switcher
-  is open.
+- Simulated 3D cards: This setting uses the classic DWM thumbnail-strip 3D effect.
+- Performance profile: This setting changes animation speed and the number of visible windows.
+
+## Performance
+
+Live DWM thumbnails are used only during animations.
+
+More thumbnail strips are used to make the edges smoother while keeping the original pseudo-3D window shape.
 
 ## Requirements
 
-- Windows 10 (64 bit) or Windows 11 (64 bit)
-- DWM composition enabled (on by default)
+- Windows 10 version 1809 or later (64-bit)
+- DWM enabled (default on Windows)
+- 2-4 GB RAM 
 
 ## Notes
 
-- The switcher is drawn on the **primary monitor**.
-- The native `Alt+Tab` switcher is never intercepted.
-- While the mod is enabled it owns the `Win+Tab` hotkey, replacing the modern
-  Task View. Disable or uninstall the mod to get Task View back.
-
-
+- Aero Flip 3D appears on the main monitor, where all available windows are displayed in the 3D switcher interface.
+- The native Alt+Tab window switcher is not modified and continues to work normally.
+- While the mod is enabled, the Win+Tab shortcut is redirected from the modern Task View interface to the classic Flip 3D experience.
+- To restore the original Windows Task View behavior, simply disable or uninstall the mod.
 */
 // ==/WindhawkModReadme==
 
 // ==WindhawkModSettings==
 /*
 - usePerspective: true
-  $name: Realistic perspective cards
-  $description: If this setting is enabled, the mod deforms each card into a perspective trapezoid using live thumbnail strips (Windows 7-like). If disabled, the mod uses the original simplified flat 2.5D card cascade (scale, offset and opacity, no perspective).
-- verboseLogging: false
-  $name: Verbose debug logging
-  $description: This setting logs detailed step-by-step diagnostics (window enumeration, thumbnail registration, hook install steps, layout computation) to the Windhawk log. Disable it for normal use; turn it on only when troubleshooting since it is more verbose than the default logging.
+  $name: Simulated 3D cards
+  $description: This setting uses the classic DWM thumbnail-strip pseudo-3D effect.
+
+- performanceProfile: auto
+  $name: Performance profile
+  $description: This setting changes the animation speed and the number of visible windows.
+
+  $options:
+  - auto: Auto
+  - high: High quality
+  - low: Low
+  - verylow: Very low
 */
 // ==/WindhawkModSettings==
 
 // -----------------------------------------------------------------------------
 // Defensive rewrite (hardened build)
 //
-// This mod runs inside explorer.exe, a critical process: any unhandled fault
-// restarts the shell. The code follows an aggressive defensive strategy:
+// This mod runs in Windhawk's dedicated tool process. The code follows a
+// conventional RAII and explicit-error-checking strategy:
 //
 //  1. RAII everywhere: HDC, HBITMAP, HBRUSH, HHOOK, HANDLE, HTHUMBNAIL and
 //     BeginPaint/EndPaint are owned by small move-only wrapper classes, so
@@ -80,14 +92,13 @@ The mod has been tested on Windows 10 1809.
 //  2. Aggressive validation: every HWND is checked with IsWindow() before use,
 //     every pointer with a null check, every Win32/DWM return value with an
 //     explicit test before its result is trusted.
-//  3. No exception escapes a Windows boundary: every callback invoked by the
-//     system (WndProcs, low-level hooks, thread proc, EnumWindows callbacks,
-//     Wh_ModInit/Wh_ModUninit) is wrapped in try/catch. On failure the mod
-//     logs, cleans up and degrades gracefully instead of crashing Explorer.
+//  3. UI and input work run on distinct threads: low-level hook callbacks
+//     only classify/post input, while enumeration, DWM and animation run on
+//     the UI thread.
 //  4. Minimal logging: Wh_Log is used only for init/uninit, unrecoverable
 //     failures and caught exceptions — never for per-frame or per-key events.
-//  5. Non-blocking init: Wh_ModInit starts a background hook thread and waits
-//     at most 200 ms, so Explorer's watchdog can never restart the shell.
+//  5. Tool-process teardown joins the UI and input threads before code is
+//     unloaded, so no hook or window procedure can outlive the module.
 //  6. Isolated components with fallbacks: if the desktop snapshot, the
 //     wallpaper bitmap or DWM thumbnails fail, the backdrop degrades to a
 //     plain Aero gradient and the mod keeps working.
@@ -97,73 +108,45 @@ The mod has been tested on Windows 10 1809.
 //     small number of failed DWM updates; failure logs are rate-limited; the
 //     overlay has a safe-destroy fallback.
 //
-// Only public, documented APIs are used: DwmRegisterThumbnail,
-// DwmUpdateThumbnailProperties, DwmUnregisterThumbnail, EnumWindows,
-// RegisterHotKey and low-level input hooks.
+// The standard fallback uses public DWM thumbnail APIs. The optional live
+// DirectComposition backend is explicitly gated to Windows 10 1809 and checks
+// its private DWM exports before it is enabled.
 // -----------------------------------------------------------------------------
 
 // -----------------------------------------------------------------------------
-// How the REAL Flip 3D worked (from the decompiled Windows 7 RTM dwmapi.dll)
+// How the REAL Flip 3D worked, and this public-API implementation
 //
-// dwmapi.dll was only an RPC client: the Flip 3D engine lived inside the DWM
-// compositor process (dwm.exe, managed by the uxsms service).
+// Windows 7's dwmapi.dll was an RPC client: the original Flip 3D engine lived
+// in dwm.exe and therefore had compositor-private perspective transforms. On
+// Windows 10/11 that engine no longer exists. Earlier revisions of this mod
+// approximated it by registering many cropped DWM thumbnail strips per card.
+// That design has been removed: every strip required a synchronous ALPC update,
+// could not rotate continuously, and exposed DWM's uniform-fit seams.
 //
-// - Toggling Flip 3D: exported ordinal #105 (the famous
-//   `rundll32 dwmapi #105` trick) simply sends ALPC message 0x4000002B to
-//   dwm.exe, with no window arguments. dwm.exe then rendered the 3D card
-//   stack itself, with true perspective transforms — which is exactly why it
-//   cannot be replicated 1:1 with public APIs: DWM thumbnails only support a
-//   destination rect + opacity, so this mod draws a pseudo-3D stack instead.
-// - Thumbnail APIs used by this mod and their ALPC message IDs:
-//     DwmRegisterThumbnail          -> 0x4000001D (destination must be a
-//                                        top-level window owned by the
-//                                        calling process — our overlay —
-//                                        source may be any top-level window;
-//                                        HTHUMBNAIL is a client-local ID)
-//     DwmUpdateThumbnailProperties  -> 0x4000001E (45-byte property payload;
-//                                        the real DWM rejects flags == 0 and
-//                                        inverted rects — right < left or
-//                                        bottom < top — with E_INVALIDARG;
-//                                        ApplyThumbnailProperties mirrors
-//                                        this validation exactly)
-//     DwmUnregisterThumbnail        -> 0x4000001F
-//     DwmQueryThumbnailSourceSize   -> 0x40000021
-// - Every thumbnail update is one LPC round-trip to dwm.exe: that is why
-//   updates are batched once per animation frame and z-order rebuilds happen
-//   only when the selection changes, never per-frame.
-// - The Flip 3D policy attribute was DWMWA_FLIP3D_POLICY (public id 11,
-//   internal DWM attribute 17 per the attribute translation table).
-//   DWMWA_CLOAKED (14) did not exist on Windows 7 — it is a Windows 8+
-//   attribute, used here only to skip cloaked UWP windows on modern systems.
+// The normal path uses only documented DWM thumbnails. A window is registered
+// in a small number of cropped vertical strips; their independent destination
+// heights create the original pseudo-3D trapezoid while the source remains live.
 //
-// The Windows 10/11 DWM no longer implements message 0x4000002B (and
-// Windhawk itself requires Windows 10+), so this mod re-implements the
-// switcher inside explorer.exe using only the still-supported thumbnail
-// messages listed above.
-//
-// Perspective approximation: since a thumbnail can only be scaled/moved
-// (no affine or perspective transforms), each card's live content is
-// registered up to kPerspectiveStripCount times (fewer for cards deep in
-// the stack, none beyond depth 3), each thumbnail cropped to a vertical
-// source strip via DWM_TNP_RECTSOURCE and mapped onto a destination strip
-// whose height follows the trapezoid of a card rotated around a vertical
-// axis. The tilt amount is animated and grows with stack depth (front
-// card nearly flat, deep cards angled — CoverFlow-style), so the morphs
-// are seamless and the large front card shows no visible strip
-// staircase. If DWM rejects the strips, the card falls back to a single
-// flat thumbnail. All of this stays within public, documented APIs.
+// The overlay backdrop remains a cached desktop snapshot with documented GDI
+// fallbacks. All rendering APIs used are available on Windows 10 version 1809.
+// -----------------------------------------------------------------------------
 // -----------------------------------------------------------------------------
 
 #include <windows.h>
+#include <shellapi.h>
 #include <dwmapi.h>
+#include <timeapi.h>
+
 
 #include <algorithm>
 #include <atomic>
 #include <cmath>
 #include <cstdint>
+#include <cstring>
 #include <cstdarg>
+#include <cstdlib>
+#include <cwchar>
 #include <utility>
-#include <setjmp.h>
 #include <vector>
 
 #ifndef DWMWA_EXTENDED_FRAME_BOUNDS
@@ -197,10 +180,6 @@ The mod has been tested on Windows 10 1809.
 // Never log per-frame animation, per-key input or per-thumbnail updates.
 // -----------------------------------------------------------------------------
 
-static void LogNote(PCWSTR what) {
-    Wh_Log(L"[AeroFlip3D] %s", what);
-}
-
 // Failure logs are rate-limited to at most one per second: a recurring fault
 // (e.g. an exception in a WM_PAINT handler) must not flood the log or burn
 // CPU inside a low-level hook, where the system enforces a response timeout.
@@ -223,8 +202,7 @@ static void LogFailureCode(PCWSTR what, DWORD code) {
 // Verbose, opt-in diagnostics (the "verboseLogging" setting). These are the
 // logs to turn on when troubleshooting a specific problem (mod not
 // activating, hook install failing, thumbnails not registering, layout
-// looking wrong, etc.) — they are deliberately noisier than LogNote/
-// LogFailure and are off by default so normal use stays quiet.
+// looking wrong, etc.) — they are deliberately noisier than LogFailure and are off by default so normal use stays quiet.
 static std::atomic<bool> g_verboseLogging{false};
 
 static void LogDebug(PCWSTR what) {
@@ -620,196 +598,11 @@ static ScopeGuard<Fn> MakeScopeGuard(Fn fn) {
     return ScopeGuard<Fn>(std::move(fn));
 }
 
-// Runs fn(); if any exception escapes, logs `what` and returns failValue.
-// Used to keep exceptions from crossing system boundaries.
-template <typename R, typename Fn>
-static R SafeExec(PCWSTR what, R failValue, Fn&& fn) {
-    try {
-        return fn();
-    } catch (...) {
-        LogFailure(what);
-        return failValue;
-    }
-}
-
 // -----------------------------------------------------------------------------
-// Anti-crash "fault fence" (public Win32 APIs: AddVectoredExceptionHandler,
-// TlsAlloc/GetValue/SetValue, longjmp, kernel32 file I/O).
-//
-// Why this exists: the mod runs inside explorer.exe, so any UNHANDLED hardware
-// fault (access violation, illegal instruction, divide-by-zero, ...) on a
-// thread we own, or inside a Windows callback we installed, terminates the
-// shell. The C++ try/catch blocks elsewhere only catch *thrown* exceptions;
-// whether they also catch hardware faults depends on the compiler's exception
-// model, which we cannot control here. A Vectored Exception Handler combined
-// with setjmp/longjmp catches hardware faults regardless of that model.
-//
-// Design (kept deliberately conservative so it can never disturb explorer or
-// other mods):
-//   - A per-thread stack of "fences" (setjmp buffers + a label) is kept in TLS.
-//   - Each system boundary (WndProcs, low-level hooks, thread proc, the
-//     ActivateFlip3D path, EnumWindows callbacks, Wh_ModInit/Uninit) opens a
-//     fence with FLIP3D_FENCE_BEGIN / FLIP3D_FENCE_ELSE / FLIP3D_FENCE_END.
-//   - The VEH returns EXCEPTION_CONTINUE_SEARCH for everything EXCEPT a small
-//     set of hardware fault codes, AND only when the faulting thread currently
-//     has an active fence. In that case it appends a one-line record to
-//     %TEMP%\AeroFlip3D_crash.log (the fence label + exception code + RIP, so
-//     the offending boundary is identified) and longjmps back to the fence's
-//     recovery block, which returns a safe value. explorer.exe keeps running.
-//   - C++ thrown exceptions (0xE06D7363), breakpoints and stack overflows are
-//     NEVER intercepted here: the former are handled by the existing try/catch
-//     (so RAII destructors still run on those paths), the latter cannot be
-//     recovered from safely.
-//   - The VEH does no heap allocation and only reads the TLS slot, so it is
-//     safe to run on any thread at any time.
-//
-// Residual, accepted risk: if a fault happens *inside* a Win32/DWM call while
-// that call holds one of its own internal locks, the longjmp will not release
-// it; the worst case is that *our* hook thread later hangs (Flip 3D stops
-// working) while the rest of explorer stays alive -- which is strictly better
-// than a shell restart, and the crash log records the event.
+// No vectored exception handler or non-local fault recovery is installed. A tool
+// mod must never attempt to continue after an access violation: RAII cleanup
+// and ordinary error-return handling are the only supported recovery paths.
 // -----------------------------------------------------------------------------
-
-struct FenceFrame {
-    jmp_buf* jb;
-    const char* label;
-};
-
-struct FenceStack {
-    FenceFrame frames[16];
-    int top;
-};
-
-static DWORD g_fenceTls = TLS_OUT_OF_INDEXES;
-static PVOID g_veh = nullptr;
-
-static FenceStack* GetFenceStack(bool create) {
-    if (g_fenceTls == TLS_OUT_OF_INDEXES) {
-        return nullptr;
-    }
-    FenceStack* fs = static_cast<FenceStack*>(TlsGetValue(g_fenceTls));
-    if (!fs && create) {
-        fs = static_cast<FenceStack*>(calloc(1, sizeof(FenceStack)));
-        if (fs) {
-            TlsSetValue(g_fenceTls, fs);
-        }
-    }
-    return fs;
-}
-
-static void Flip3DFencePush(jmp_buf* jb, const char* label) {
-    FenceStack* fs = GetFenceStack(true);
-    if (fs && fs->top < 16) {
-        fs->frames[fs->top].jb = jb;
-        fs->frames[fs->top].label = label;
-        fs->top++;
-    }
-}
-
-static void Flip3DFencePop() {
-    FenceStack* fs = GetFenceStack(false);
-    if (fs && fs->top > 0) {
-        fs->top--;
-    }
-}
-
-// Auto-pop on the normal (non-recovery) return path, so a boundary with many
-// `return` statements cannot leak a fence level.
-class Flip3DFenceAutoPop {
-public:
-    ~Flip3DFenceAutoPop() {
-        Flip3DFencePop();
-    }
-};
-
-// Appends a single diagnostic line using only kernel32/user32 (no CRT heap),
-// so it is safe to call from the VEH fault path.
-static void WriteCrashLog(DWORD code, PEXCEPTION_POINTERS ep, const char* label) {
-    wchar_t path[MAX_PATH] = L"";
-    DWORD len = GetTempPathW(MAX_PATH, path);
-    if (len == 0 || len >= MAX_PATH) {
-        return;
-    }
-    if (lstrcatW(path, L"AeroFlip3D_crash.log") == nullptr) {
-        return;
-    }
-
-#if defined(_M_ARM64) || defined(_M_ARM64EC)
-    ULONG_PTR rip = ep && ep->ContextRecord ? static_cast<ULONG_PTR>(ep->ContextRecord->Pc) : 0;
-#elif defined(_M_IX86)
-    ULONG_PTR rip = ep && ep->ContextRecord ? static_cast<ULONG_PTR>(ep->ContextRecord->Eip) : 0;
-#else
-    ULONG_PTR rip = ep && ep->ContextRecord ? static_cast<ULONG_PTR>(ep->ContextRecord->Rip) : 0;
-#endif
-    wchar_t line[384];
-    wsprintfW(line,
-              L"[AeroFlip3D] fence recovered | fence=%S | code=0x%08X | rip=0x%016IX | tick=%u\r\n",
-              label ? label : "?",
-              code,
-              rip,
-              static_cast<unsigned>(GetTickCount64() & 0xFFFFFFFFu));
-
-    HANDLE h = CreateFileW(path, FILE_APPEND_DATA,
-                           FILE_SHARE_READ | FILE_SHARE_WRITE,
-                           nullptr, OPEN_ALWAYS, FILE_ATTRIBUTE_NORMAL, nullptr);
-    if (h != INVALID_HANDLE_VALUE) {
-        DWORD written = 0;
-        DWORD bytes = static_cast<DWORD>(wcslen(line) * sizeof(wchar_t));
-        WriteFile(h, line, bytes, &written, nullptr);
-        CloseHandle(h);
-    }
-}
-
-static LONG NTAPI Flip3D_Veh(PEXCEPTION_POINTERS ep) {
-    if (!ep || !ep->ExceptionRecord) {
-        return EXCEPTION_CONTINUE_SEARCH;
-    }
-
-    const DWORD code = ep->ExceptionRecord->ExceptionCode;
-
-    // Only recover from hardware faults we can survive. Explicitly exclude C++
-    // throws, breakpoints and stack overflows.
-    if (code != EXCEPTION_ACCESS_VIOLATION &&
-        code != EXCEPTION_ILLEGAL_INSTRUCTION &&
-        code != EXCEPTION_INT_DIVIDE_BY_ZERO &&
-        code != EXCEPTION_INT_OVERFLOW &&
-        code != EXCEPTION_DATATYPE_MISALIGNMENT &&
-        code != EXCEPTION_PRIV_INSTRUCTION &&
-        code != EXCEPTION_ARRAY_BOUNDS_EXCEEDED &&
-        code != EXCEPTION_FLT_DENORMAL_OPERAND &&
-        code != EXCEPTION_FLT_DIVIDE_BY_ZERO &&
-        code != EXCEPTION_FLT_INVALID_OPERATION) {
-        return EXCEPTION_CONTINUE_SEARCH;
-    }
-
-    FenceStack* fs = GetFenceStack(false);
-    if (!fs || fs->top <= 0) {
-        return EXCEPTION_CONTINUE_SEARCH;  // Not our guarded code: leave it alone.
-    }
-
-    const char* label = fs->frames[fs->top - 1].label;
-    jmp_buf* jb = fs->frames[fs->top - 1].jb;
-    if (!jb) {
-        return EXCEPTION_CONTINUE_SEARCH;
-    }
-
-    WriteCrashLog(code, ep, label);
-    longjmp(*jb, 1);  // Transfer control to the boundary's recovery block.
-}
-
-// The fence must be opened with these three macros INSIDE the boundary
-// function (setjmp has to live in the frame we longjmp back to). The recovery
-// block (between BEGIN and ELSE) must end with a `return`. The normal body
-// (between ELSE and END) may return freely; its fence level is auto-popped.
-#define FLIP3D_FENCE_BEGIN(label)        \
-    jmp_buf _flip3d_fjb;                 \
-    Flip3DFencePush(&_flip3d_fjb, label); \
-    if (setjmp(_flip3d_fjb)) {           \
-        Flip3DFencePop();
-#define FLIP3D_FENCE_ELSE                \
-    } else {                             \
-        Flip3DFenceAutoPop _flip3d_fpop;
-#define FLIP3D_FENCE_END }
 
 // -----------------------------------------------------------------------------
 // State
@@ -817,10 +610,10 @@ static LONG NTAPI Flip3D_Veh(PEXCEPTION_POINTERS ep) {
 
 struct FlipWindowEntry {
     HWND hwnd = nullptr;
-    ThumbnailHandle thumbnail;          // Used only in flat-fallback mode.
-    std::vector<ThumbnailHandle> slices;  // Perspective strips (one live thumbnail each).
-    int stripCount = 0;                   // 0 = this card uses the flat single thumbnail.
-    bool stripsHidden = false;            // True once hidden strips were told to hide (skip per-frame work).
+    ThumbnailHandle thumbnail;          // Flat fallback / simulated deep card.
+    std::vector<ThumbnailHandle> slices;  // Simulated 3D vertical DWM strips.
+    int stripCount = 0;
+    bool stripsHidden = false;
     SIZE sourceSize = {0, 0};
     RECT sourceRect = {0, 0, 0, 0};
 
@@ -839,6 +632,17 @@ struct FlipWindowEntry {
     double currentTilt = 0.0;
     double startTilt = 0.0;
     double targetTilt = 0.0;
+
+    // Last pose actually pushed to DWM. Every thumbnail update is a
+    // synchronous ALPC round-trip to dwm.exe, so a card whose geometry did
+    // not move since the previous frame must not be re-sent: on a slow
+    // machine those redundant calls are the single biggest source of stutter
+    // (the animation easing settles long before the timer stops, and hidden
+    // or unchanged deep cards would otherwise keep paying full price).
+    RECT appliedRect = {0, 0, 0, 0};
+    BYTE appliedOpacity = 0;
+    double appliedTilt = -1.0;  // -1 = nothing applied yet.
+    bool everApplied = false;
 };
 
 enum class TriggerModifier : int {
@@ -854,28 +658,32 @@ enum class FlipAnimationKind : int {
     Exit,
 };
 
-static HWND g_hControllerWnd = nullptr;  // message-only window on hook/UI thread
+static HWND g_hControllerWnd = nullptr;  // message-only window on UI thread
 static HWND g_hOverlayWnd = nullptr;     // visible fullscreen stack overlay
+static bool g_overlayClassRegistered = false;
+static bool g_controllerClassRegistered = false;
 static HWND g_hDesktopThumbnailSource = nullptr;
 static HTHUMBNAIL g_desktopThumbnail = nullptr;
-static ScopedBitmap g_desktopSnapshotBitmap;    // real desktop captured via GDI
+[[clang::no_destroy]] static ScopedBitmap g_desktopSnapshotBitmap;    // real desktop captured via GDI
 static SIZE g_desktopSnapshotSize = {0, 0};
-static ScopedBitmap g_backdropCompositeBitmap;  // snapshot + veil, pre-blended once
+[[clang::no_destroy]] static ScopedBitmap g_backdropCompositeBitmap;  // snapshot + veil, pre-blended once
 static SIZE g_backdropCompositeSize = {0, 0};
 static DWORD g_hookThreadId = 0;
 static DWORD g_overlayThreadId = 0;
-static ScopedHandle g_hookThread;
-static ScopedHandle g_hookReadyEvent;
+[[clang::no_destroy]] static ScopedHandle g_hookThread;       // UI/controller thread.
+[[clang::no_destroy]] static ScopedHandle g_inputThread;      // LL-hook-only thread.
+static DWORD g_inputThreadId = 0;
+[[clang::no_destroy]] static ScopedHandle g_hookReadyEvent;
 static std::atomic<bool> g_hookInstallOk{false};
 static std::atomic<bool> g_hookSessionActive{false};
 
-static ScopedHook g_keyboardHook;
-static ScopedHook g_mouseHook;
+[[clang::no_destroy]] static ScopedHook g_keyboardHook;
+[[clang::no_destroy]] static ScopedHook g_mouseHook;
 
-static std::vector<FlipWindowEntry> g_windows;
+[[clang::no_destroy]] static std::vector<FlipWindowEntry> g_windows;
 static int g_selectedIndex = 0;
 static bool g_desktopSelected = false;  // true = front slot empty, desktop showing (Win7-style entry)
-static bool g_persistentMode = false;   // true = Ctrl+Win+Tab or Ctrl+G: switcher stays open after key release
+static bool g_persistentMode = false;   // true = Win+Tab: switcher stays open after key release
 static int g_activeWindowIndex = -1;    // index of the foreground window at activation time
 static HWND g_initialForegroundHwnd = nullptr;
 static bool g_isActive = false;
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

---

**1. This should be a tool mod, not an `explorer.exe` mod.**
The mod installs zero function hooks. Everything it uses — `SetWindowsHookEx(WH_KEYBOARD_LL/WH_MOUSE_LL)`, `RegisterHotKey`, `EnumWindows`, `DwmRegisterThumbnail`, its own overlay window — works from *any* process. There is no reason to be inside the shell, and being inside it is what forced the whole defensive apparatus (item 2) to exist in the first place. It also creates a real bug today: with "Launch folder windows in a separate process" enabled there are multiple `explorer.exe` processes, and the mod will load into each one, each installing its own LL keyboard hook, its own VEH, and racing for the same hotkeys.

Please convert to the ["mods as tools" pattern](https://github.com/ramensoftware/windhawk/wiki/Mods-as-tools:-Running-mods-in-a-dedicated-process): `@include windhawk.exe`, rename `Wh_ModInit`/`Wh_ModSettingsChanged`/`Wh_ModUninit` → `WhTool_*`, and paste the launcher boilerplate from the wiki **verbatim**. The closest existing mod, [Simple Window Switcher](https://github.com/ramensoftware/windhawk-mods/blob/109589023dde428deaee2fe80e4ce446283a7935/mods/simple-window-switcher.wh.cpp#L4864), is exactly this shape and is worth reading end to end: it runs its switcher UI + DWM thumbnails + hotkeys in the dedicated tool process, and only injects into `explorer.exe` for the one thing that genuinely needs injection (hooking `RegisterHotKey` so the shell doesn't grab the combo first). [Always On Top](https://github.com/ramensoftware/windhawk-mods/blob/109589023dde428deaee2fe80e4ce446283a7935/mods/always-on-top.wh.cpp#L431) is a smaller example.

**2. Remove the VEH + `setjmp`/`longjmp` "fault fence" entirely.**
This is the biggest blocker. Lines 673-806 / 3876 install a process-wide vectored exception handler in `explorer.exe` that intercepts access violations at first chance and `longjmp`s out of the faulting frame. Concretely:

- `longjmp` out of a hardware fault does **not** unwind — none of the `Scoped*` destructors run, so every RAII object the whole design relies on leaks, and any lock held by the code you jumped out of (user32, GDI, the DWM RPC channel, the loader lock) stays held forever. The comment at line 666 calls this an "accepted risk", but a leaked user32/loader lock hangs *Explorer*, not just the mod's thread.
- `longjmp` with non-trivial destructors in scope is **undefined behavior** in C++ ([csetjmp.syn]) — and every fenced function has vectors/`Scoped*` objects live.
- A first-chance VEH that swallows `EXCEPTION_ACCESS_VIOLATION` also hijacks AVs that downstream code legitimately handles with `__try`/`__except`, aborting operations that would have recovered fine.
- Values read in the recovery block (e.g. `msg` in `OverlayWndProc`) are non-`volatile` locals/params, so they are indeterminate after `longjmp`.
- The fence stack itself is buggy: `Flip3DFencePush` silently drops the frame when `top == 16`, but `Flip3DFenceAutoPop` pops unconditionally — so an overflow desynchronizes the stack and a later fault can `longjmp` into a frame that no longer exists.
- `WriteCrashLog` has a stack buffer overflow: `GetTempPathW` may return up to 259 chars into `wchar_t path[MAX_PATH]`, then `lstrcatW` appends 20 more without bounds checking (line 733).
- It writes a log file to `%TEMP%\AeroFlip3D_crash.log`. Mods should not write files; `Wh_Log` is the mechanism.
- The per-thread `FenceStack` is `calloc`'d per thread and only freed for the thread that happens to run `Wh_ModUninit`, so every other thread's block leaks on each load/unload cycle.

Continuing execution after an access violation leaves the process in an unknown state — that is strictly worse than the crash it's trying to prevent. No mod in the catalog does this. Delete the fence, delete `#include <setjmp.h>`, and rely on writing correct code (moving to a tool mod also removes the "a fault restarts the shell" motivation). The blanket `try`/`catch (...)` wrappers can go with it.

**3. Window classes are never unregistered, and the `ERROR_CLASS_ALREADY_EXISTS` fallback makes it worse.**
`CreateOverlayWindow` (line 2681) and `CreateControllerWindow` (line 3336) both treat `ERROR_CLASS_ALREADY_EXISTS` as success, and nothing ever calls `UnregisterClass`. A class registered by the mod DLL is *not* unregistered when the DLL unloads, so on the next load (mod toggle / update / settings reload) the stale registration is still there with `lpfnWndProc` pointing into the unmapped previous image. Depending on where the new image lands you either get `CreateWindowExW` failing with `ERROR_CANNOT_FIND_WND_CLASS` (switcher silently never opens again until Explorer restarts) or a window dispatching messages to a dangling `WndProc` — a crash in the shell. Register fresh on every load and `UnregisterClassW(className, hInstance)` on teardown, after the windows are destroyed; see [Simple Window Switcher](https://github.com/ramensoftware/windhawk-mods/blob/109589023dde428deaee2fe80e4ce446283a7935/mods/simple-window-switcher.wh.cpp#L4844) and the [window-class note in the wiki](https://github.com/ramensoftware/windhawk/wiki/Global-objects-and-process-shutdown#7-raii-wrappers-that-call-teardown-apis).

**4. Heavy work runs on the low-level-hook thread, stalling system-wide input.**
`WH_KEYBOARD_LL`/`WH_MOUSE_LL` callbacks are dispatched on the installing thread, and *all* system input is blocked until that thread returns (or `LowLevelHooksTimeout`, 300 ms by default, elapses — after which Windows may silently drop the hook). This mod puts the entire UI on that same thread:

- `CanStartFlip3DSession()` is called directly from inside `LowLevelKeyboardProc` (lines 3426, 3463). It runs a full `EnumWindows` and, per window, `DwmGetWindowAttribute(DWMWA_CLOAKED)` — dozens of cross-process round-trips to `dwm.exe`, inside the input hook.
- Every navigation calls `RebuildThumbnailZOrder()`, which unregisters and re-registers *every* thumbnail — up to 14 + 10 + 7 + 6 strips plus the flat cards — each one an LPC round-trip to `dwm.exe`, synchronously on the hook thread.
- While the switcher is open, the 60 fps timer pushes up to ~37 `DwmUpdateThumbnailProperties` calls per frame on that same thread.

Split the input hook onto its own dedicated thread that does nothing but classify the keystroke and `PostMessage` to the UI thread, and keep the enumeration/DWM/animation work off it entirely.

**5. `Ctrl+G` is registered as a system-wide hotkey.**
Line 3694: `RegisterHotKey(g_hControllerWnd, kToggleHotkeyId, MOD_CONTROL, 'G')`. `RegisterHotKey` combos are system-wide and the keystroke never reaches the focused app, so while this mod is enabled `Ctrl+G` stops working in every program on the machine (find-next in browsers, go-to-line in editors, …). A bare `Ctrl`+letter combo is not an acceptable global hotkey. Drop it, or make the hotkey a setting with a modifier combo that isn't in general use.

**6. The mod can lock out the keyboard system-wide.**
The LL hook sets `g_hookSessionActive = true` *before* posting the activate message (line 3433/3471), and while that flag is set it swallows essentially every keystroke (`return 1` at line 3546). The only things that clear it are the controller/overlay message handlers and the 60 s failsafe timer — but the failsafe timer lives on the *overlay* window, which only exists after a successful activation. If `PostToController` doesn't reach a live controller window (window destroyed, thread gone, thread wedged), the flag stays set and the user's keyboard is dead everywhere with no recovery except disabling the mod by mouse. Don't set the "active" flag from the hook — set it only once the UI thread has actually opened the switcher, and add a hook-side timeout that clears it if no confirmation arrives.

**7. `Wh_ModUninit` can leave the hooks installed while the DLL unloads.**
`Wh_ModUninitImpl` waits 5 s for the hook thread (line 3917) and then just logs on timeout. `g_keyboardHook`/`g_mouseHook` are only reset in the `else` branch, so in the timeout path the LL hooks stay installed, the controller window keeps its `WndProc`, and the thread keeps running — all pointing into the mod image that Windhawk is about to unmap. That's an Explorer crash. Unhook, destroy the windows and unregister the classes unconditionally in `Wh_ModUninit`, regardless of how the wait ended.

**8. Globals with non-trivial destructors run at process shutdown, when `Wh_ModUninit` does not.**
On Explorer restart / sign-out / reboot, `Wh_ModUninit` is not called; the CRT runs global destructors alone on the shutdown thread, after every other thread was already terminated. Affected globals here (lines 855-869):

- `g_windows` — `std::vector<FlipWindowEntry>`, and each `ThumbnailHandle` destructor calls `DwmUnregisterThumbnail`, i.e. an LPC round-trip into `dwm.exe` at shutdown. This is the dangerous one.
- `g_keyboardHook` / `g_mouseHook` — `UnhookWindowsHookEx` under the loader lock.
- `g_hookThread`, `g_hookReadyEvent` — `CloseHandle`.
- `g_desktopSnapshotBitmap`, `g_backdropCompositeBitmap` — `DeleteObject`.

Mark them `[[clang::no_destroy]]` and keep (or add) the explicit release in `Wh_ModUninit`. The `Scoped*` wrappers are nullable/handle-like so they can take the bare attribute; `g_windows` is a container of resource-owning elements, so it needs the wrapper form — `[[clang::no_destroy]] std::optional<std::vector<FlipWindowEntry>>`, released with `.reset()` (not `.clear()`) on the thread that registered the thumbnails. See https://github.com/ramensoftware/windhawk/wiki/Global-objects-and-process-shutdown, sections [5](https://github.com/ramensoftware/windhawk/wiki/Global-objects-and-process-shutdown#5-containers-of-resource-owning-or-thread-affine-elements) and [7](https://github.com/ramensoftware/windhawk/wiki/Global-objects-and-process-shutdown#7-raii-wrappers-that-call-teardown-apis).

**9. The 4-second hotkey "reclaim" timer.**
`kHotkeyReclaimTimerId` (lines 3230-3248) unregisters and re-registers Win+Tab and Ctrl+Win+Tab every 4 seconds, forever, whether or not the switcher is in use. That is a permanent polling loop fighting Task View / Game Bar over a global hotkey, and it opens a small window every 4 s where the combo is unowned and Task View pops up instead. Register once at startup and leave it; if the hotkey is genuinely lost, react to an event (e.g. re-register when the hook fallback observes the combo) rather than polling.

**10. `IsClassBlacklisted` hides unrelated applications.**
Lines 1730-1814 blacklist window classes by product name, and several are not product-specific: `Qt5QWindowIcon` is the class name of essentially *every* Qt 5 application window, so the switcher will silently omit qBittorrent, VirtualBox, VLC, and so on; `Steam` and `Action` are similarly generic. It's also unclear why a window switcher needs a screen-recorder blacklist at all — was this intentional, or an AI artifact? I'd drop the whole product list and keep only the shell/system classes and the mod's own windows.

**11. README issues.**
- No screenshot or GIF. This is a purely visual mod — please add one (only `i.imgur.com` and `raw.githubusercontent.com` are allowed image hosts).
- The README documents "**Confirm**: `Enter`, `Space`, or click", but `OverlayWndProc` has no mouse-button handling at all — clicking does nothing. Either implement click-to-select (with hit-testing against the card rects, which is also what makes the deck feel like the original) or remove the claim.
- `Ctrl+Win+F` is swallowed by the LL hook (line 3414) but isn't documented anywhere.

**12. Use `Wh_Log` directly; drop the custom logging layer and the `verboseLogging` setting.**
`LogNote`/`LogFailure`/`LogFailureCode`/`LogDebug`/`LogDebugFmt` (lines 200-247) reimplement things Windhawk already provides: logging is off by default and toggled from the Windhawk UI, and the mod name is already prefixed, so the `[AeroFlip3D]` prefixes and the `verboseLogging` setting are redundant (`Wh_Log` also takes printf-style arguments directly, so `LogDebugFmt`'s buffer is unnecessary). Call `Wh_Log` at the call sites and delete the setting — `Wh_Log` compiles down to a cheap flag check, so there's no cost to leaving the calls in.

<p>

- **`@compilerOptions`** (line 10): `-O3 -DNDEBUG -fomit-frame-pointer -funroll-loops` aren't needed — nothing here is compute-bound, and `-fomit-frame-pointer` only makes a crash inside Explorer harder to diagnose. Just keep the `-l` flags you actually need.
- **`LoadLibraryW(L"msimg32.dll")`** (line 1548): `msimg32` is already in `@compilerOptions`, so you can call `AlphaBlend` directly and drop the `GetProcAddress` dance. If you'd rather keep the dynamic load, use `GetModuleHandleW` (it's already mapped) or `LoadLibraryExW(..., LOAD_LIBRARY_SEARCH_SYSTEM32)` — `msimg32` isn't a KnownDLL, so a bare-name load is a (here low-risk, since the host is `explorer.exe`) DLL-hijacking path.
- **Dead code**: `kEnableDesktopDwmBackdrop` is a compile-time `false`, which makes `RegisterDesktopThumbnail`, `FindDesktopThumbnailSourceWindow`, `FindDesktopHostProc`, `ApplyDesktopThumbnailProperties` and `g_desktopThumbnail` unreachable; `kEnableLowLevelInputHooks`/`kEnableMinimalHotkeyHook` are compile-time `true`; `DrawFaux3DFrames` is an empty function called twice per paint; `g_hookReadyEvent` is never created (only ever `SetEvent`'d behind a null check); `g_hookInstallOk` is stored but never read. The comment at line 90 ("waits at most 200 ms") no longer matches the code, which doesn't wait at all.
- **`PreloadSymbols`**: Windows binds imports at load time (there's no lazy binding without delay-load), so the `GetProcAddress` calls don't prevent anything. The function can go.
- **Comments are partly in Italian** ("Se già abbiamo un bitmap…", "Finestre del mod stesso", "SOFTWARE DI REGISTRAZIONE SCHERMO", "Controllo cache con tolleranza DPI"). Please keep the source in English.
- **`SetThreadPriority(..., THREAD_PRIORITY_ABOVE_NORMAL)`** (line 3659) isn't needed for a helper thread.
- **`ActivateSelectedWindow`**: the overlay is already the foreground window at that point, so plain `SetForegroundWindow` should work without the `AttachThreadInput` dance, which is deadlock-prone.
- **Missing includes**: `<cstdlib>` (`calloc`/`free`/`abs`) and `<cwchar>` (`wcslen`) are used via transitive includes.
- **`SetStretchBltMode(HALFTONE)`** should be followed by `SetBrushOrgEx` to avoid brush misalignment (both in `EnsureBackdropComposite` and `PaintStaticWallpaperBackdrop`).
- **`HasUwpContent`** returns `true` for every window that *isn't* an `ApplicationFrameWindow` — the name says the opposite of what it does.

</p>
</details>

<p>

- **Auto-cycle while Win is held** (`kAutoCycleIntervalMs = 250`) isn't what Vista/7 did — the original advanced on each `Tab` press (with normal key auto-repeat when Tab was held), it didn't spin on its own while `Win` was held. As implemented you can't hold `Win` to look at the deck without it walking away from your selection.
- **The 60 s failsafe** also fires in persistent mode, so a `Ctrl+Win+Tab` switcher closes itself after a minute.
- **Per-frame full-screen repaint**: `UpdateAnimationFrame` calls `InvalidateRect(g_hOverlayWnd, nullptr, FALSE)` every 16 ms, causing a full-screen `BitBlt` of the (static) backdrop at 60 fps. Only the DWM thumbnails move — the backdrop only needs repainting when the overlay is first shown or resized.
- **Rebuild flicker**: `RebuildThumbnailZOrder` unregisters *all* thumbnails and re-registers them on every navigation, so there's a frame with nothing on screen, plus the LPC cost. Consider registering once per session and only re-registering the cards whose relative z-order actually changed.
- **Stale backdrop cache**: `EnsureBackdropComposite` only invalidates on a size change, so if `CaptureDesktopSnapshot` ever takes its reuse path the composite would keep showing the previous desktop. (Today `ReleaseDesktopSnapshot` runs on every close, so the reuse path is effectively dead code.)
- **Blank-capture detection** treats "all 9 sampled pixels are black" as a failed capture — a plain black wallpaper falls through all three capture paths to the Aero gradient.
- **Single monitor**: the deck is drawn on the primary monitor only, and there's no `WM_DISPLAYCHANGE`/`WM_DPICHANGED` handling if the layout changes while the switcher is open.
- **Strip-based perspective** is a neat workaround for DWM thumbnails having no transform, but 14 live thumbnails for one card is a lot of DWM state for the effect. Worth measuring whether ~6-8 strips on the front card looks materially worse; it would cut the per-frame ALPC traffic substantially.
- **`SwallowModifierRelease`** injects `VK 0xFF` to defuse the Start menu. It's a known trick, but it's fragile and it synthesizes input from inside a low-level hook; if it ever misfires the user sees a stray Start menu. Worth a comment explaining the failure mode, or a guard that skips it when the Start menu isn't at risk.

</p>
</details>

babamohammed2022:
> ### Submission review
> 
> 
> **1. This should be a tool mod, not an `explorer.exe` mod.** The mod installs zero function hooks. Everything it uses — `SetWindowsHookEx(WH_KEYBOARD_LL/WH_MOUSE_LL)`, `RegisterHotKey`, `EnumWindows`, `DwmRegisterThumbnail`, its own overlay window — works from _any_ process. There is no reason to be inside the shell, and being inside it is what forced the whole defensive apparatus (item 2) to exist in the first place. It also creates a real bug today: with "Launch folder windows in a separate process" enabled there are multiple `explorer.exe` processes, and the mod will load into each one, each installing its own LL keyboard hook, its own VEH, and racing for the same hotkeys.
> 
> Please convert to the ["mods as tools" pattern](https://github.com/ramensoftware/windhawk/wiki/Mods-as-tools:-Running-mods-in-a-dedicated-process): `@include windhawk.exe`, rename `Wh_ModInit`/`Wh_ModSettingsChanged`/`Wh_ModUninit` → `WhTool_*`, and paste the launcher boilerplate from the wiki **verbatim**. The closest existing mod, [Simple Window Switcher](https://github.com/ramensoftware/windhawk-mods/blob/109589023dde428deaee2fe80e4ce446283a7935/mods/simple-window-switcher.wh.cpp#L4864), is exactly this shape and is worth reading end to end: it runs its switcher UI + DWM thumbnails + hotkeys in the dedicated tool process, and only injects into `explorer.exe` for the one thing that genuinely needs injection (hooking `RegisterHotKey` so the shell doesn't grab the combo first). [Always On Top](https://github.com/ramensoftware/windhawk-mods/blob/109589023dde428deaee2fe80e4ce446283a7935/mods/always-on-top.wh.cpp#L431) is a smaller example.
> 
> **2. Remove the VEH + `setjmp`/`longjmp` "fault fence" entirely.** This is the biggest blocker. Lines 673-806 / 3876 install a process-wide vectored exception handler in `explorer.exe` that intercepts access violations at first chance and `longjmp`s out of the faulting frame. Concretely:
> 
> * `longjmp` out of a hardware fault does **not** unwind — none of the `Scoped*` destructors run, so every RAII object the whole design relies on leaks, and any lock held by the code you jumped out of (user32, GDI, the DWM RPC channel, the loader lock) stays held forever. The comment at line 666 calls this an "accepted risk", but a leaked user32/loader lock hangs _Explorer_, not just the mod's thread.
> * `longjmp` with non-trivial destructors in scope is **undefined behavior** in C++ ([csetjmp.syn]) — and every fenced function has vectors/`Scoped*` objects live.
> * A first-chance VEH that swallows `EXCEPTION_ACCESS_VIOLATION` also hijacks AVs that downstream code legitimately handles with `__try`/`__except`, aborting operations that would have recovered fine.
> * Values read in the recovery block (e.g. `msg` in `OverlayWndProc`) are non-`volatile` locals/params, so they are indeterminate after `longjmp`.
> * The fence stack itself is buggy: `Flip3DFencePush` silently drops the frame when `top == 16`, but `Flip3DFenceAutoPop` pops unconditionally — so an overflow desynchronizes the stack and a later fault can `longjmp` into a frame that no longer exists.
> * `WriteCrashLog` has a stack buffer overflow: `GetTempPathW` may return up to 259 chars into `wchar_t path[MAX_PATH]`, then `lstrcatW` appends 20 more without bounds checking (line 733).
> * It writes a log file to `%TEMP%\AeroFlip3D_crash.log`. Mods should not write files; `Wh_Log` is the mechanism.
> * The per-thread `FenceStack` is `calloc`'d per thread and only freed for the thread that happens to run `Wh_ModUninit`, so every other thread's block leaks on each load/unload cycle.
> 
> Continuing execution after an access violation leaves the process in an unknown state — that is strictly worse than the crash it's trying to prevent. No mod in the catalog does this. Delete the fence, delete `#include <setjmp.h>`, and rely on writing correct code (moving to a tool mod also removes the "a fault restarts the shell" motivation). The blanket `try`/`catch (...)` wrappers can go with it.
> 
> **3. Window classes are never unregistered, and the `ERROR_CLASS_ALREADY_EXISTS` fallback makes it worse.** `CreateOverlayWindow` (line 2681) and `CreateControllerWindow` (line 3336) both treat `ERROR_CLASS_ALREADY_EXISTS` as success, and nothing ever calls `UnregisterClass`. A class registered by the mod DLL is _not_ unregistered when the DLL unloads, so on the next load (mod toggle / update / settings reload) the stale registration is still there with `lpfnWndProc` pointing into the unmapped previous image. Depending on where the new image lands you either get `CreateWindowExW` failing with `ERROR_CANNOT_FIND_WND_CLASS` (switcher silently never opens again until Explorer restarts) or a window dispatching messages to a dangling `WndProc` — a crash in the shell. Register fresh on every load and `UnregisterClassW(className, hInstance)` on teardown, after the windows are destroyed; see [Simple Window Switcher](https://github.com/ramensoftware/windhawk-mods/blob/109589023dde428deaee2fe80e4ce446283a7935/mods/simple-window-switcher.wh.cpp#L4844) and the [window-class note in the wiki](https://github.com/ramensoftware/windhawk/wiki/Global-objects-and-process-shutdown#7-raii-wrappers-that-call-teardown-apis).
> 
> **4. Heavy work runs on the low-level-hook thread, stalling system-wide input.** `WH_KEYBOARD_LL`/`WH_MOUSE_LL` callbacks are dispatched on the installing thread, and _all_ system input is blocked until that thread returns (or `LowLevelHooksTimeout`, 300 ms by default, elapses — after which Windows may silently drop the hook). This mod puts the entire UI on that same thread:
> 
> * `CanStartFlip3DSession()` is called directly from inside `LowLevelKeyboardProc` (lines 3426, 3463). It runs a full `EnumWindows` and, per window, `DwmGetWindowAttribute(DWMWA_CLOAKED)` — dozens of cross-process round-trips to `dwm.exe`, inside the input hook.
> * Every navigation calls `RebuildThumbnailZOrder()`, which unregisters and re-registers _every_ thumbnail — up to 14 + 10 + 7 + 6 strips plus the flat cards — each one an LPC round-trip to `dwm.exe`, synchronously on the hook thread.
> * While the switcher is open, the 60 fps timer pushes up to ~37 `DwmUpdateThumbnailProperties` calls per frame on that same thread.
> 
> Split the input hook onto its own dedicated thread that does nothing but classify the keystroke and `PostMessage` to the UI thread, and keep the enumeration/DWM/animation work off it entirely.
> 
> **5. `Ctrl+G` is registered as a system-wide hotkey.** Line 3694: `RegisterHotKey(g_hControllerWnd, kToggleHotkeyId, MOD_CONTROL, 'G')`. `RegisterHotKey` combos are system-wide and the keystroke never reaches the focused app, so while this mod is enabled `Ctrl+G` stops working in every program on the machine (find-next in browsers, go-to-line in editors, …). A bare `Ctrl`+letter combo is not an acceptable global hotkey. Drop it, or make the hotkey a setting with a modifier combo that isn't in general use.
> 
> **6. The mod can lock out the keyboard system-wide.** The LL hook sets `g_hookSessionActive = true` _before_ posting the activate message (line 3433/3471), and while that flag is set it swallows essentially every keystroke (`return 1` at line 3546). The only things that clear it are the controller/overlay message handlers and the 60 s failsafe timer — but the failsafe timer lives on the _overlay_ window, which only exists after a successful activation. If `PostToController` doesn't reach a live controller window (window destroyed, thread gone, thread wedged), the flag stays set and the user's keyboard is dead everywhere with no recovery except disabling the mod by mouse. Don't set the "active" flag from the hook — set it only once the UI thread has actually opened the switcher, and add a hook-side timeout that clears it if no confirmation arrives.
> 
> **7. `Wh_ModUninit` can leave the hooks installed while the DLL unloads.** `Wh_ModUninitImpl` waits 5 s for the hook thread (line 3917) and then just logs on timeout. `g_keyboardHook`/`g_mouseHook` are only reset in the `else` branch, so in the timeout path the LL hooks stay installed, the controller window keeps its `WndProc`, and the thread keeps running — all pointing into the mod image that Windhawk is about to unmap. That's an Explorer crash. Unhook, destroy the windows and unregister the classes unconditionally in `Wh_ModUninit`, regardless of how the wait ended.
> 
> **8. Globals with non-trivial destructors run at process shutdown, when `Wh_ModUninit` does not.** On Explorer restart / sign-out / reboot, `Wh_ModUninit` is not called; the CRT runs global destructors alone on the shutdown thread, after every other thread was already terminated. Affected globals here (lines 855-869):
> 
> * `g_windows` — `std::vector<FlipWindowEntry>`, and each `ThumbnailHandle` destructor calls `DwmUnregisterThumbnail`, i.e. an LPC round-trip into `dwm.exe` at shutdown. This is the dangerous one.
> * `g_keyboardHook` / `g_mouseHook` — `UnhookWindowsHookEx` under the loader lock.
> * `g_hookThread`, `g_hookReadyEvent` — `CloseHandle`.
> * `g_desktopSnapshotBitmap`, `g_backdropCompositeBitmap` — `DeleteObject`.
> 
> Mark them `[[clang::no_destroy]]` and keep (or add) the explicit release in `Wh_ModUninit`. The `Scoped*` wrappers are nullable/handle-like so they can take the bare attribute; `g_windows` is a container of resource-owning elements, so it needs the wrapper form — `[[clang::no_destroy]] std::optional<std::vector<FlipWindowEntry>>`, released with `.reset()` (not `.clear()`) on the thread that registered the thumbnails. See https://github.com/ramensoftware/windhawk/wiki/Global-objects-and-process-shutdown, sections [5](https://github.com/ramensoftware/windhawk/wiki/Global-objects-and-process-shutdown#5-containers-of-resource-owning-or-thread-affine-elements) and [7](https://github.com/ramensoftware/windhawk/wiki/Global-objects-and-process-shutdown#7-raii-wrappers-that-call-teardown-apis).
> 
> **9. The 4-second hotkey "reclaim" timer.** `kHotkeyReclaimTimerId` (lines 3230-3248) unregisters and re-registers Win+Tab and Ctrl+Win+Tab every 4 seconds, forever, whether or not the switcher is in use. That is a permanent polling loop fighting Task View / Game Bar over a global hotkey, and it opens a small window every 4 s where the combo is unowned and Task View pops up instead. Register once at startup and leave it; if the hotkey is genuinely lost, react to an event (e.g. re-register when the hook fallback observes the combo) rather than polling.
> 
> **10. `IsClassBlacklisted` hides unrelated applications.** Lines 1730-1814 blacklist window classes by product name, and several are not product-specific: `Qt5QWindowIcon` is the class name of essentially _every_ Qt 5 application window, so the switcher will silently omit qBittorrent, VirtualBox, VLC, and so on; `Steam` and `Action` are similarly generic. It's also unclear why a window switcher needs a screen-recorder blacklist at all — was this intentional, or an AI artifact? I'd drop the whole product list and keep only the shell/system classes and the mod's own windows.
> 
> **11. README issues.**
> 
> * No screenshot or GIF. This is a purely visual mod — please add one (only `i.imgur.com` and `raw.githubusercontent.com` are allowed image hosts).
> * The README documents "**Confirm**: `Enter`, `Space`, or click", but `OverlayWndProc` has no mouse-button handling at all — clicking does nothing. Either implement click-to-select (with hit-testing against the card rects, which is also what makes the deck feel like the original) or remove the claim.
> * `Ctrl+Win+F` is swallowed by the LL hook (line 3414) but isn't documented anywhere.
> 
> **12. Use `Wh_Log` directly; drop the custom logging layer and the `verboseLogging` setting.** `LogNote`/`LogFailure`/`LogFailureCode`/`LogDebug`/`LogDebugFmt` (lines 200-247) reimplement things Windhawk already provides: logging is off by default and toggled from the Windhawk UI, and the mod name is already prefixed, so the `[AeroFlip3D]` prefixes and the `verboseLogging` setting are redundant (`Wh_Log` also takes printf-style arguments directly, so `LogDebugFmt`'s buffer is unnecessary). Call `Wh_Log` at the call sites and delete the setting — `Wh_Log` compiles down to a cheap flag check, so there's no cost to leaving the calls in.
> 
> Optional improvements
> 
> 
> * **`@compilerOptions`** (line 10): `-O3 -DNDEBUG -fomit-frame-pointer -funroll-loops` aren't needed — nothing here is compute-bound, and `-fomit-frame-pointer` only makes a crash inside Explorer harder to diagnose. Just keep the `-l` flags you actually need.
> * **`LoadLibraryW(L"msimg32.dll")`** (line 1548): `msimg32` is already in `@compilerOptions`, so you can call `AlphaBlend` directly and drop the `GetProcAddress` dance. If you'd rather keep the dynamic load, use `GetModuleHandleW` (it's already mapped) or `LoadLibraryExW(..., LOAD_LIBRARY_SEARCH_SYSTEM32)` — `msimg32` isn't a KnownDLL, so a bare-name load is a (here low-risk, since the host is `explorer.exe`) DLL-hijacking path.
> * **Dead code**: `kEnableDesktopDwmBackdrop` is a compile-time `false`, which makes `RegisterDesktopThumbnail`, `FindDesktopThumbnailSourceWindow`, `FindDesktopHostProc`, `ApplyDesktopThumbnailProperties` and `g_desktopThumbnail` unreachable; `kEnableLowLevelInputHooks`/`kEnableMinimalHotkeyHook` are compile-time `true`; `DrawFaux3DFrames` is an empty function called twice per paint; `g_hookReadyEvent` is never created (only ever `SetEvent`'d behind a null check); `g_hookInstallOk` is stored but never read. The comment at line 90 ("waits at most 200 ms") no longer matches the code, which doesn't wait at all.
> * **`PreloadSymbols`**: Windows binds imports at load time (there's no lazy binding without delay-load), so the `GetProcAddress` calls don't prevent anything. The function can go.
> * **Comments are partly in Italian** ("Se già abbiamo un bitmap…", "Finestre del mod stesso", "SOFTWARE DI REGISTRAZIONE SCHERMO", "Controllo cache con tolleranza DPI"). Please keep the source in English.
> * **`SetThreadPriority(..., THREAD_PRIORITY_ABOVE_NORMAL)`** (line 3659) isn't needed for a helper thread.
> * **`ActivateSelectedWindow`**: the overlay is already the foreground window at that point, so plain `SetForegroundWindow` should work without the `AttachThreadInput` dance, which is deadlock-prone.
> * **Missing includes**: `<cstdlib>` (`calloc`/`free`/`abs`) and `<cwchar>` (`wcslen`) are used via transitive includes.
> * **`SetStretchBltMode(HALFTONE)`** should be followed by `SetBrushOrgEx` to avoid brush misalignment (both in `EnsureBackdropComposite` and `PaintStaticWallpaperBackdrop`).
> * **`HasUwpContent`** returns `true` for every window that _isn't_ an `ApplicationFrameWindow` — the name says the opposite of what it does.
> 
> Functionality notes

OK, thanks for the review. If possible, let me know if there are any other issues to address to improve the mod
Note: Screenshots will be added once the mod is approved for merging and the code is finalized. I want to ensure the visual appearance matches the final implementation before capturing images

m417z:
The new review is currently only enabled for selected pull requests. Enabled for this one too:
/ai-review

babamohammed2022:
> ### Submission review
> 
> 
> 
> Thanks for the rewrite — the tool-mod conversion, the removal of the VEH/`longjmp` fence, the class unregistration, the split input thread, the bounded hotkey claim and the `Ctrl+Alt+F12` hotkey all look good, and the tool-mod launcher boilerplate at the bottom matches the wiki snippet verbatim. The remaining items are mostly about the keyboard state machine and settings that don't do what they say.
> 
> **1. `Tab` never navigates the deck — the only way to advance is the 250 ms auto-cycle timer.** `LowLevelKeyboardProc` swallows every `Tab` while the switcher is open, before any of the navigation paths can run:
> 
> ```c++
> if (active && vk == VK_TAB && winDown) {
>     return 1;                       // line 3399-3403
> }
> ...
> if (active) {
>     ...
>     if (vk == VK_TAB) {
>         return 1;                   // line 3463-3465
>     }
> ```
> 
> Consequences:
> 
> * The `WM_FLIP3D_NAVIGATE` post at line 3433 (the `Win`-held Tab-advance path) is unreachable — line 3399 always returns first.
> * `WM_HOTKEY`/`kHotkeyWinTab` -> `NavigateSelection(1)` (line 3228) is unreachable too, because an LL hook returning nonzero also suppresses `RegisterHotKey` delivery — which is exactly what the mod relies on elsewhere to block Task View.
> * `OverlayWndProc`'s `WM_KEYDOWN` -> `VK_TAB` -> `NavigateSelection(1)` (line 3113) never runs either, since the hook consumes the key before it reaches any queue.
> 
> So the README's "Navigate: `Tab`, arrow keys, or mouse wheel" is only true for arrows and the wheel, and holding `Win` just spins through the deck on `kAutoCycleTimerId` at 4 cards/s (see also the functionality note below). The fix is to handle `Tab` in the `if (active)` block like the arrow keys instead of dropping it:
> 
> ```c++
> if (active) {
>     ...
>     if (keyDown && vk == VK_TAB) {
>         PostToController(WM_FLIP3D_NAVIGATE,
>                          static_cast<WPARAM>(static_cast<INT_PTR>(shiftDown ? -1 : 1)), 0);
>         return 1;
>     }
>     if (vk == VK_TAB) {
>         return 1;  // swallow the key-up only
>     }
> ```
> 
> **2. `Ctrl+Alt+F12` cannot close the switcher, only open it.** The README documents it as "Opens or closes Flip 3D manually", but the close half never works in the normal case:
> 
> * `manualShortcut` (line 3376-3378) is gated on `!stickyOwnedElsewhere`, and `RegisterHotKey(Ctrl+Alt+F12)` normally succeeds, so the hook never handles the combo itself.
> * While `active`, the `Alt` key-down is swallowed by the catch-all `return 1` at line 3500 (`IsAltKey` is _not_ in the pass-through list at line 3366), so `GetAsyncKeyState(VK_MENU)` never even reports `Alt` as down; the `F12` key-down is then swallowed by the same `return 1`, so `WM_HOTKEY` never fires.
> 
> Either let the combo through while active (add `VK_F12` + `IsAltKey` to the pass-through cases so `WM_HOTKEY` can be delivered), or handle it in the hook regardless of `g_stickyHotkeyOwned` when `active` is true. This is currently the documented fallback for getting out of persistent mode, so it matters more than a normal shortcut.
> 
> **3. State shared between the input thread and the UI thread is unsynchronized.** The mod correctly made `g_hookSessionActive`, `g_lastNavigationTick` and the hotkey-ownership flags `std::atomic`, but several other globals are read by `LowLevelKeyboardProc`/`LowLevelMouseProc` (input thread) while being written by the UI thread, with no atomics or barriers (lines 697-704):
> 
> * `g_triggerModifier` — written in `ActivateFlip3DImpl` (line 2910) right before the `relaxed` store to `g_hookSessionActive` (line 2911). The hook can observe `active == true` with a stale `TriggerModifier::None`, in which case `IsTriggerModifierKey` at line 3439 fails and releasing `Win` does **not** confirm/close the switcher — it stays open until `Esc` or the 60 s failsafe.
> * `g_persistentMode` (read at line 3440), `g_suppressNextModifierRelease` / `g_suppressReleaseModifier` (written from _both_ threads: line 3385-3386 in the hook, `PreparePendingModifierReleaseSuppression` from `OverlayWndProc`).
> * `g_wheelAccum`, `g_lastWheelPostMs`, `g_lastWheelDir` — written by the mouse hook, reset by `ResetWheelState()` on the UI thread.
> 
> Make these `std::atomic`, and use `memory_order_release`/`acquire` around the `g_hookSessionActive` handshake so everything the UI thread published before activating is visible to the hook.
> 
> **4. The 128-window cap makes the switcher stop working entirely instead of capping the deck.** `EnumWindowsForFlipProc` returns `FALSE` when the cap is reached (line 1793-1794), and `EnumerateFlipEligibleWindows` treats _any_ early stop as a failure and throws the whole list away (line 1813-1817):
> 
> ```c++
> if (result->size() >= kMaxFlipWindows) {
>     return FALSE;  // Deck cap reached: stop enumerating (DoS guard).
> }
> ...
> if (!EnumWindows(EnumWindowsForFlipProc, reinterpret_cast<LPARAM>(&result))) {
>     result.clear();
> }
> ```
> 
> So on a machine with 128+ eligible top-level windows, `eligible` comes back empty, `ActivateFlip3D` bails with "not enough eligible windows", and `Win+Tab` does nothing at all (the key is already swallowed, so Task View doesn't appear either). Distinguish "stopped because the cap was hit" (keep the results) from "stopped because of an error" (discard), e.g. with a flag in the callback's context struct.
> 
> **5. Unload can hang: `WM_QUIT` is posted through thread-ID globals that the threads set themselves.** `WhTool_ModUninitImpl` (line 3810-3821) does:
> 
> ```c++
> if (g_hookThreadId) {
>     PostThreadMessageW(g_hookThreadId, WM_QUIT, 0, 0);
> }
> if (g_hookThread) {
>     WaitForSingleObject(g_hookThread.get(), INFINITE);
> ```
> 
> `g_hookThreadId` is only assigned inside `HookThreadProcImpl` (line 3680). If the mod is disabled before the new thread has executed that line, the `if` is skipped, no `WM_QUIT` is ever sent, and the `INFINITE` wait never returns — the tool process hangs with the LL keyboard hook still installed. The hook thread's own cleanup has the same pattern for `g_inputThreadId` (line 3690, set at line 3585). Capture the thread ID from `CreateThread`'s `lpThreadId` out-parameter (or retry `PostThreadMessageW` until it succeeds, since it fails only until the target has a message queue) so the quit request can never be lost.
> 
> **6. `performanceProfile` doesn't change the number of visible windows.** `DetectPerformanceProfile` computes `p.maxDeckCards` (8 / 6 / 5, lines 748, 802, 806), but nothing ever reads it — `ComputeSimulatedStackLayout` hard-codes the cap:
> 
> ```c++
> const int maxVisibleDepth = std::min(count, 8) + (desktopSelected ? 1 : 0);  // line 2485
> ```
> 
> `StripCountForDepth` (line 898) is likewise profile-independent, so "Low"/"Very low" still register 20 + 14 + 10 + 8 strips. That leaves the setting's `$description` ("changes the animation speed and the number of visible windows") and the README's Performance section describing behavior the code doesn't implement — only `frameIntervalMs` and `allowSnapshot` actually change. Either use `g_perf.maxDeckCards` (and scale `StripCountForDepth` by the profile), or drop the unused field and reword the description.
> 
> **7. With "Simulated 3D cards" disabled, the mod still pays the full strip cost for a flat result.** `RebuildThumbnailZOrder` always uses `StripCountForDepth` (line 2238) regardless of `g_settings.perspective`; the setting only zeroes `targetTilt` (line 2567). With tilt 0 every strip of a card gets the same height, so ~52 registered thumbnails and ~52 synchronous `DwmUpdateThumbnailProperties` ALPC round-trips per animation frame produce exactly the image a single thumbnail per card would. Gate the strip path on the setting:
> 
> ```c++
> const int stripsWanted = g_settings.perspective ? StripCountForDepth(depth) : 0;
> ```
> 
> **8. The custom logging layer is still there, and its verbose half is dead code.** This was point 12 of the previous review. `LogFailure`/`LogFailureCode`/`LogDebug`/`LogDebugFmt` (lines 186-225) still wrap `Wh_Log` with `[AeroFlip3D]` / `[AeroFlip3D][debug]` prefixes that Windhawk already adds, and `LogDebugFmt` hand-formats into a buffer even though `Wh_Log` takes printf-style arguments directly. On top of that the `verboseLogging` setting was removed from the settings block but `g_verboseLogging` remained, hard-wired to `false` (line 206) and never written — so every `LogDebug`/`LogDebugFmt` call in the file (30+ of them, including the whole activation trace) can never emit anything, and the comment at line 202 still refers to a setting that no longer exists. Call `Wh_Log` directly at the call sites and delete the wrappers; `Wh_Log` is a cheap flag check and is off by default, so there is no reason to gate or throttle it.
> 
> **9. README still has no screenshot or GIF.** I saw the note that visuals will come once the code is finalized — flagging it only so it doesn't get lost, since a purely visual mod really needs one before merge (allowed image hosts are `i.imgur.com` and `raw.githubusercontent.com`). Please also fix the "Navigate: `Tab`" claim (item 1) and the `Ctrl+Alt+F12` "or closes" claim (item 2), whichever way you resolve them.
> 
> Optional improvements
> 
> 
> * **The blanket `try`/`catch (...)` wrappers can go.** Now that the VEH/`longjmp` fence is gone, the remaining `try { ... } catch (...)` blocks around `OverlayWndProc`, `ControllerWndProc`, both LL hooks, `ActivateFlip3DImpl`, `HookThreadProcImpl`, `EnumWindowsForFlipProc`, `FindDesktopHostProc` and the `reserve()` calls catch nothing that actually happens: Win32/DWM report failure through return codes, and with the Clang/mingw toolchain `catch (...)` does **not** catch SEH faults such as access violations. They add nesting and an odd control flow (`EnumWindowsForFlipProc` returning `FALSE` "on exception" is what makes item 4 fire) without buying robustness.
> * **`g_windows` and `[[clang::no_destroy]]`** (line 683): the attribute is warranted (each `ThumbnailHandle` destructor is a `DwmUnregisterThumbnail` ALPC call, which must not run at process shutdown), but `std::vector` isn't nullable, so the bare attribute leaves no way to fully release it — `CleanupThumbnails()` uses `g_windows.clear()` (line 2416), which destroys the elements but keeps the heap buffer. The recommended form is the wrapper, released with `reset()`: `[[clang::no_destroy]] std::optional<std::vector<FlipWindowEntry>> g_windows;`. Low impact here since `Wh_ModUninit` ends in `ExitProcess(0)`, but see https://github.com/ramensoftware/windhawk/wiki/Global-objects-and-process-shutdown ([section 5](https://github.com/ramensoftware/windhawk/wiki/Global-objects-and-process-shutdown#5-containers-of-resource-owning-or-thread-affine-elements)). The other `[[clang::no_destroy]]` globals (the `Scoped*` handle wrappers) are handle-like, so the bare attribute is right for them.
> * **`LoadLibraryW(L"msimg32.dll")`** (line 1465): `msimg32` is already in `@compilerOptions`, so `AlphaBlend` can be called directly and the `GetProcAddress` dance dropped. If you keep the dynamic load, use `LoadLibraryExW(..., nullptr, LOAD_LIBRARY_SEARCH_SYSTEM32)` — `msimg32` is not a KnownDLL, so a bare-name load searches the executable's directory first; low risk in `windhawk.exe`, but free to avoid.
> * **Dead code / stale comments** still present:
>   
>   * `kEnableDesktopDwmBackdrop` is `constexpr false` (line 880), so `RegisterDesktopThumbnail`, `FindDesktopThumbnailSourceWindow`, `FindDesktopHostProc`, `ApplyDesktopThumbnailProperties`, `g_desktopThumbnail` and `g_hDesktopThumbnailSource` are unreachable; `kEnableLowLevelInputHooks` is `constexpr true`, so the branch at line 3405 is dead.
>   * `DrawFaux3DFrames` (line 1613) is an empty function called twice per paint.
>   * `g_hookReadyEvent` is never created — only `SetEvent`'d behind a null check (line 3731) — and `g_hookInstallOk` is stored but never read, so the "ready event is ALWAYS signaled" comment at line 3753 and the "Maximum time `Wh_ModInit` may wait" comment at line 911 no longer describe anything.
>   * `PostToController`'s `PostThreadMessageW(g_hookThreadId, ...)` fallback (line 1156) can't work: a thread message has `msg.hwnd == nullptr`, so `DispatchMessageW` drops it and no window proc ever sees `WM_FLIP3D_*`.
>   * Several comments still talk about `explorer.exe` (lines 1546, 2241, 3029, 3753) or reference source material that no longer applies.
> * **Comments still partly in Italian**: line 1290 ("Se non abbiamo un bitmap valido, creane uno nuovo") and line 1418 ("Cache valida"). Please keep the source in English.
> * **`LogFailure` throttling** is global (one message per second for _all_ callers), so an unrelated failure right after a common one is silently dropped. Another reason to just call `Wh_Log` directly.
> * **`FallbackToSingleThumbnail`** (line 1905) logs "Perspective strips unavailable" unconditionally, including on the success path.
> * **`ApplySingleThumbnailProperties`** hard-codes `>= 5` (line 1886) where the rest of the code uses `kMaxThumbnailUpdateFailures`.
> * **`Wh_GetStringSetting` never returns NULL** (it returns `L""` on error or when unset), so the `if (PCWSTR choice = ...)` check at line 826 is redundant. `WindhawkUtils::StringSetting` (from `windhawk_utils.h`) would also remove the manual `Wh_FreeStringSetting`.
> * **Redundant thumbnail registration at open**: `ActivateFlip3DImpl` registers a thumbnail per window (line 2937) just to call `DwmQueryThumbnailSourceSize`, and `RebuildThumbnailZOrder` (line 3017) immediately unregisters all of them and starts over. That is one extra register/unregister round-trip per window on every activation.
> * **`ActivateSelectedWindow`** (line 2710): the overlay is the foreground window at that point, so plain `SetForegroundWindow` should work without the `AttachThreadInput` dance, which can block on a hung target thread.
> * **`SetStretchBltMode(HALFTONE)`** should be followed by `SetBrushOrgEx` to avoid brush misalignment (lines 1458 and 1572).
> * **`HasUwpContent`** (line 1679) returns `true` for every window that _isn't_ an `ApplicationFrameWindow` — the name says the opposite of what it does.
> * **`WhTool_ModSettingsChanged`** runs on an arbitrary thread and writes `g_settings`/`g_perf` while the UI thread reads them; posting the reload to the controller window would keep all of that on one thread.
> * **`WM_ERASEBKGND` and `WM_PAINT`** both call `PaintStaticWallpaperBackdrop`, so every paint cycle blits the full-screen backdrop twice. Returning `1` from `WM_ERASEBKGND` without painting, and painting only in `WM_PAINT`, halves that.
> 
> Functionality notes

Excluding the screenshot, most of the issues have been addressed. If there are any other problems to address let me know
/ai-review

babamohammed2022:
> ### Submission review
> 
> 
> 
> Thanks for the iteration — the custom logging layer is gone, the enumeration cap no longer discards the whole list, `maxDeckCards` is now actually used, the strip path is gated on the perspective setting, and the tool-mod launcher boilerplate still matches the wiki snippet verbatim. Most of what's left is in the keyboard state machine again: the `Ctrl+Alt+F12` change went the wrong way and the shortcut is now dead in the normal case.
> 
> **1. `Ctrl+Alt+F12` no longer works at all in the normal case — the hook swallows it and nothing handles it.** The previous `manualShortcut` condition contained `&& !stickyOwnedElsewhere`, so when `RegisterHotKey(Ctrl+Alt+F12)` succeeded the hook let the key through and `WM_HOTKEY` did the work. The new version (line 3238) dropped that term but kept the unconditional `return 1`:
> 
> ```c++
> const bool manualShortcut =
>     (vk == VK_F12 && ctrlDown && altDown && !winDown) ||   // line 3239
>     false;
> if (manualShortcut && keyDown) {
>     ...
>     if (active) { /* close */ }
>     else if (!stickyOwnedElsewhere) { /* open */ }
>     return 1;                                              // line 3258 — always
> }
> ```
> 
> `RegisterHotKey(kStickyHotkeyId, MOD_CONTROL | MOD_ALT, VK_F12)` normally succeeds, so `stickyOwnedElsewhere` is `true`. When the switcher is closed, neither branch runs — and the `return 1` suppresses the key, so `WM_HOTKEY`/`kStickyHotkeyId` (line 3095) is never delivered either. The documented "fallback" shortcut is a no-op: it can't open the switcher.
> 
> The close direction is likely broken too, for a second reason. While `active`, only `Ctrl` and `Shift` key-downs are passed through (line 3226); `Alt` is not, so its key-down hits the catch-all `return 1` at line 3362. A key suppressed by a low-level hook is never handed to the rest of the system, so neither `GetAsyncKeyState(VK_MENU)` nor `LLKHF_ALTDOWN` reliably reports `Alt` as down afterwards — `altDown` is then false, `manualShortcut` is false, and the `F12` key-down is swallowed by the same catch-all. Worth verifying on a real machine, but the fix removes the doubt either way.
> 
> Suggested shape — handle the combo entirely in the hook (so it works whether or not `RegisterHotKey` succeeded), track the modifier state from the events the hook itself observes rather than from `GetAsyncKeyState`, and only `return 1` when the hook actually consumed the key:
> 
> ```c++
> if (manualShortcut && keyDown) {
>     ...
>     if (active) { /* close */ } else { /* open */ }
>     return 1;   // consumed
> }
> // otherwise fall through to CallNextHookEx so WM_HOTKEY can still fire
> ```
> 
> If you'd rather keep `RegisterHotKey` as the primary path, then the hook must _not_ swallow `F12` when `stickyOwnedElsewhere` is true, and `Alt` must be added to the pass-through list at line 3226 so the combo can be recognised while the switcher is open.
> 
> **2. Unload can still hang: the `WM_QUIT` post is unchecked before an `INFINITE` wait.** Capturing the thread ID from `CreateThread` (lines 3666-3671, 3589-3593) makes the ID available immediately, but `PostThreadMessageW` _fails_ with `ERROR_INVALID_THREAD_ID` until the target thread has created its message queue (which happens inside the thread, at the `PeekMessageW` bootstrap). The return value is ignored:
> 
> ```c++
> if (g_hookThreadId) {
>     PostThreadMessageW(g_hookThreadId, WM_QUIT, 0, 0);   // line 3684 — may fail
> }
> if (g_hookThread) {
>     WaitForSingleObject(g_hookThread.get(), INFINITE);   // line 3687 — never returns
> ```
> 
> If the mod is disabled before the freshly created thread has run its first line, the quit request is lost and the tool process hangs forever with the LL keyboard hook installed. Same pattern for `g_inputThreadId` at line 3690 and in the hook thread's cleanup guard (line 3553). Retry until it succeeds, bounded by the thread actually exiting:
> 
> ```c++
> while (g_hookThreadId && !PostThreadMessageW(g_hookThreadId, WM_QUIT, 0, 0) &&
>        WaitForSingleObject(g_hookThread.get(), 10) == WAIT_TIMEOUT) {
> }
> ```
> 
> **3. `Tab` behaves differently depending on which trigger path is active, and neither matches the README.** The README now says "`Tab` is reserved: it is swallowed to prevent Task View interference", and the comment at line 3321 says the 250 ms auto-cycle is "the sole navigation mechanism while `Win` is held". That is only true when `g_winTabHotkeyOwned` is true. In the normal case the shell owns `Win+Tab`, so all 20 `RegisterHotKey` attempts fail, `winTabOwnedElsewhere` is false, and line 3277 is reached while the switcher is open:
> 
> ```c++
> if (vk == VK_TAB && keyDown && winDown && !altDown && !ctrlDown && !winTabOwnedElsewhere) {
>     ...
>     PostToController(WM_FLIP3D_NAVIGATE, ... shiftDown ? -1 : 1, 0);   // line 3291
>     return 1;
> }
> ```
> 
> So on a normal machine `Tab` _does_ advance the deck — on top of the auto-cycle timer, which is exactly the "advances excessively fast" combination the comment says it wants to avoid. And in the rare case where the hotkey registration succeeds, `Tab` does nothing at all (line 3325 swallows it, which also blocks `WM_HOTKEY`/`kHotkeyWinTab` -> `NavigateSelection(1)` at line 3088). Pick one behaviour and make both paths do it, then make the README match. If `Tab` is meant to advance, the auto-cycle timer is redundant and should go (see the functionality notes).
> 
> **4. The README still has no screenshot or GIF.** Noted that you're waiting for the visuals to be final — flagging it again so it isn't lost, since a purely visual mod needs one before merge. Allowed image hosts are `i.imgur.com` and `raw.githubusercontent.com`.
> 
> Optional improvements
> 
> 
> * **`[AeroFlip3D]` prefixes**: the wrapper functions are gone, but 44 `Wh_Log` calls still hard-code the `[AeroFlip3D] ` prefix. Windhawk already prefixes every log line with the mod name, so it's duplicated in the log output. Also, the header comment at line 90 ("Minimal logging: `Wh_Log` is used only for init/uninit, unrecoverable failures and caught exceptions") no longer matches — `RecomputeTargetsForCurrentSelection` logs on every navigation, which is fine (`Wh_Log` is a cheap flag check), but the comment should say so.
> * **`-lmsimg32`** is now unused: `AlphaBlend` is only reached through `GetProcAddress` (line 1375). Either drop the `-l` flag, or drop the `LoadLibraryExW`/`GetProcAddress` dance and call `AlphaBlend` directly through the import.
> * **Dead code / stale comments** still present:
>   
>   * `kEnableDesktopDwmBackdrop` is `constexpr false` (line 784), so `RegisterDesktopThumbnail`, `FindDesktopThumbnailSourceWindow`, `FindDesktopHostProc`, `ApplyDesktopThumbnailProperties`, `g_desktopThumbnail` and `g_hDesktopThumbnailSource` are unreachable; `kEnableLowLevelInputHooks` is `constexpr true`, so the branch at line 3263 is dead.
>   * `g_hookReadyEvent` (line 618) is never created — only `SetEvent`'d behind a null check — and `g_hookInstallOk` (line 619) is stored three times but never read. The comments at lines 815-816 ("Maximum time `Wh_ModInit` may wait…", with no constant after it) and 3619 ("The ready event is ALWAYS signaled so `Wh_ModInit` can never hang") describe machinery that no longer exists.
>   * `p.allowSnapshot` is now `true` on every path, so the `!g_perf.allowSnapshot` branch at line 1153 and the field itself are dead.
>   * Empty `if` body left behind at lines 3086-3087 inside `case kHotkeyWinTab`.
>   * `ComputeStackLayout` (line 2445) is a pure pass-through to `ComputeSimulatedStackLayout`.
>   * `OverlayWndProc`'s `WM_MOUSEWHEEL` handler (line 2947) is unreachable: `LowLevelMouseProc` returns 1 for every wheel event while the session is active.
>   * Comments still referring to `explorer.exe` / the shell: lines 1452-1453, 2112, 2890-2891, 3618-3620, and the "Explorer must never be blocked" note at line 815.
> * **The blanket `try`/`catch (...)` wrappers** (14 of them) can go: Win32/DWM report failure through return codes, and with the Clang/mingw toolchain `catch (...)` does not catch SEH faults such as access violations. They only add nesting.
> * **`g_windows` and `[[clang::no_destroy]]`** (line 625): the attribute is warranted (each `ThumbnailHandle` destructor is a `DwmUnregisterThumbnail` ALPC call), but `std::vector` isn't nullable, so `CleanupThumbnails()`'s `g_windows.clear()` (line 2284) destroys the elements while keeping the heap buffer. The recommended form is the wrapper, released with `reset()`: `[[clang::no_destroy]] std::optional<std::vector<FlipWindowEntry>> g_windows;` — see https://github.com/ramensoftware/windhawk/wiki/Global-objects-and-process-shutdown ([section 5](https://github.com/ramensoftware/windhawk/wiki/Global-objects-and-process-shutdown#5-containers-of-resource-owning-or-thread-affine-elements)). Low impact here since `Wh_ModUninit` ends in `ExitProcess(0)`. The other `[[clang::no_destroy]]` globals are handle-like, so the bare attribute is right for them.
> * **`FallbackToSingleThumbnail`** (line 1821) still logs "Perspective strips unavailable" unconditionally, including on the success path.
> * **Redundant thumbnail registration at open**: `ActivateFlip3DImpl` registers a thumbnail per window (line 2799) only to call `DwmQueryThumbnailSourceSize`, and `RebuildThumbnailZOrder` (line 2879) immediately unregisters all of them and starts over — one extra register/unregister round-trip per window on every activation.
> * **`WhTool_ModSettingsChanged`** (line 3708) runs on an arbitrary thread and writes `g_settings`/`g_perf` while the UI thread reads them; posting the reload to the controller window would keep all of that on one thread.
> * **`ActivateSelectedWindow`** (line 2572): the overlay is the foreground window at that point, so plain `SetForegroundWindow` should work without the `AttachThreadInput` dance, which can block on a hung target thread.
> * **README "Requirements: 2-4 GB RAM"** reads like an upper bound; presumably you mean a minimum.
> 
> Functionality notes

Thanks for the review. If there are any other problems to address let me know. If not, I will include a picture inside the README of the mod
/ai-review

babamohammed2022:
> ### Submission review
> 
> 
> 
> Good foundation: this is correctly built as a tool mod, the launcher boilerplate is a verbatim copy of the wiki snippet, the window classes are unregistered on teardown, the LL hooks only classify/post (all real work happens on the UI thread), and the one setting is declared and read correctly. The items below are the ones worth fixing before merge.
> 
> **1. Add a screenshot or GIF to the README.** This mod is purely visual — a Flip 3D recreation — and the README has no image at all. Users pick mods from the catalog by their screenshots, and reviewers can't judge how close the result is to the real thing without one. Only `i.imgur.com` and `raw.githubusercontent.com` are allowed as image hosts.
> 
> **2. Navigation is timer-driven only, so the user can't actually pick a window.** While the switcher is open, `VK_TAB` is swallowed unconditionally (`aero-flip3d-recreation.wh.cpp:3278`) and the only thing that advances the deck is `kAutoCycleTimerId`, which fires every 250 ms for as long as Win is held (`:2999`). Combined with the entry state (`g_desktopSelected = true`, so the front slot starts empty, `:2811`), a Win+Tab press followed by a normal-speed release either selects nothing or selects whatever card the timer happened to land on. There is no way to step deterministically to the _n_-th window — the user has to release Win inside a 250 ms window.
> 
> The original behaviour is one card per `Tab` press while Win is held. Please make `Tab` advance the selection (and `Shift+Tab` go back), and either drop the auto-cycle timer or make it a setting that is off by default. Note that `kLayoutAnimationDurationMs` is 400 ms while the auto-cycle period is 250 ms, so today every cycle restarts a longer animation and the deck never settles. [`simple-window-switcher`](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/simple-window-switcher.wh.cpp#L3749) is a good reference for the linear Tab/Shift+Tab/arrow navigation model.
> 
> **3. The DWM thumbnail budget is unbounded, and every navigation step re-registers all of it.**
> 
> Two separate problems that compound:
> 
> * `g_perf.maxDeckCards` doesn't limit how many cards exist — `ComputeSimulatedStackLayout` only clamps `visualDepth` (`:2368`), so with 30 open windows all 30 get a live DWM thumbnail, all 22 cards past the cap land on the _same_ rect at `targetOpacity = 255` (`:2431`), and all of them are pushed to DWM every animation frame. They're invisible (stacked behind each other) but cost full price. `kMaxFlipWindows` is 128, so the worst case is ~52 strips (20+14+10+8 for the front four) plus ~124 flat thumbnails. Cards deeper than the cap should get `targetOpacity = 0` (which takes the existing `!cardVisible` fast path in `ApplyCardThumbnails`) or simply not be registered.
> * `RebuildThumbnailZOrder` (`:2059`) unregisters _and re-registers every thumbnail of every card_ on each selection change, and with the auto-cycle timer that happens every 250 ms. Each `DwmRegisterThumbnail`/`DwmUnregisterThumbnail` is a synchronous ALPC round-trip to `dwm.exe` — you're doing 100–350 of them per second on a busy desktop. Bounding the registered set to the visible deck (previous bullet) is the main fix; also consider lowering `StripCountForDepth` (20 strips for the front card is a lot for a per-navigation full rebuild).
> 
> Both of these matter most exactly on the low-end machines the perf profile is written for.
> 
> **4. A Win+Tab that doesn't open the switcher opens the Start menu instead.** The LL hook swallows the `Tab` keydown and returns 1 (`:3243`) before the controller has decided anything. If `ActivateFlip3DImpl` then bails — fewer than 2 eligible windows (`:2732`), DWM composition off (`:2723`), overlay creation failed — nothing is active, so the later Win keyup takes the plain pass-through path and Windows sees a Win press with no intervening key: the Start menu pops up. Arm the release suppression at the moment you swallow the `Tab` (i.e. call `PreparePendingModifierReleaseSuppression()` right before `PostToController(WM_FLIP3D_ACTIVATE, ...)`), so the Win release is neutralised whether or not the switcher ends up opening.
> 
> **5. Persistent mode swallows the entire keyboard with no failsafe.** While `g_hookSessionActive` is set, `LowLevelKeyboardProc` returns 1 for every key that isn't explicitly handled (`:3315`) — that's system-wide. The `kFailsafeTimerId` auto-close is only armed when `!g_persistentMode` (`:2845`), so the `Ctrl+Alt+F12` session has no automatic recovery at all: if the UI thread ever wedges (it does cross-process work — `AttachThreadInput`, `SetForegroundWindow`, `DwmRegisterThumbnail` against arbitrary windows), the user is left with a dead keyboard. Arm the failsafe timer in persistent mode too (with a longer period if you want), and consider having the hook stop swallowing if the controller window stops acknowledging posted messages.
> 
> Optional improvements
> 
> 
> **Dead code.** There's a fair amount of it, and it's the kind of thing that makes a 3850-line mod harder to review than it needs to be:
> 
> * `kEnableDesktopDwmBackdrop` is a hard-coded `false` (`:784`), which makes the entire DWM-backdrop subsystem unreachable: `FindDesktopHostProc`, `FindDesktopThumbnailSourceWindow`, `g_desktopHostSearchResult`, `RegisterDesktopThumbnail`, `ApplyDesktopThumbnailProperties`, `UnregisterDesktopThumbnail`, `g_desktopThumbnail`, `g_hDesktopThumbnailSource` (~90 lines).
> * `g_hookReadyEvent` is declared and checked but never created — nothing ever calls `CreateEvent`, so `SetEvent` at `:3554` never runs.
> * `g_hookInstallOk`, `g_stickyHotkeyOwned` and `FlipWindowEntry::stripCount` are only ever written, never read.
> * `ActivateFlip3DImpl`'s `initialDelta` parameter is ignored on the activation path (only used in the already-active branch), so the value the hook computes from `shiftDown` at `:3239` has no effect.
> * The cross-thread branch at the top of `DeactivateFlip3D` (`:2645`) can't be reached — the overlay lives on the same thread as both window procs that call it.
> * `#include <cstring>` and `#include <cstdarg>` are unused.
> 
> **Stale comments that describe code that isn't there.** The header block claims "The optional live DirectComposition backend is explicitly gated to Windows 10 1809 and checks its private DWM exports" (`:103`), `:2747` says "The normal DirectComposition path installs a live shell thumbnail", and `:2757` says "D3D resources are created while the overlay is still hidden" — there is no DirectComposition or D3D code in the mod. Also worth removing: the `message (4).txt` references (`:797`, `:2283`), and the `(Fix #4)` / `(Fix #5)` / `(Fix #6)` / `(Fix #7)` / `(Optional fix)` / `(Z-order fix)` review-iteration markers scattered through the file — they're meaningless to a future reader.
> 
> **The `try`/`catch (...)` scaffolding.** There are 14 catch-alls wrapping window procs, hook procs, `EnumWindows` callbacks and thread entries. The comments say they guard against "hardware faults" (`:1518`, `:2867`) — they don't; a C++ `catch (...)` doesn't catch SEH access violations under clang/mingw. The only thing that can realistically be thrown in this code is `std::bad_alloc`, and the `try { result.reserve(32); } catch (...) {}` blocks around `vector::reserve` (`:1709`, `:1996`, `:2107`) don't buy anything either. Dropping most of these would shorten the file noticeably without losing safety.
> 
> **`[[clang::no_destroy]]` audit.** Only `g_windows` actually needs it — its element destructors call `DwmUnregisterThumbnail`, which is a cross-process call into `dwm.exe` and shouldn't run from the CRT shutdown path. The others (`g_desktopSnapshotBitmap`, `g_backdropCompositeBitmap`, `g_hookThread`, `g_inputThread`, `g_hookReadyEvent`, `g_keyboardHook`, `g_mouseHook`) only do `DeleteObject` / `CloseHandle` / `UnhookWindowsHookEx`, which are safe at shutdown, so the attribute is just noise there. For `g_windows`, the recommended form for a container is `[[clang::no_destroy]] std::optional<std::vector<FlipWindowEntry>>` with `reset()` in the cleanup path — `.clear()` releases the thumbnails but keeps the vector's heap buffer. See https://github.com/ramensoftware/windhawk/wiki/Global-objects-and-process-shutdown for the full picture.
> 
> **`WhTool_ModSettingsChanged` races with the UI thread.** It runs on Windhawk's thread and rewrites `g_settings` and `g_perf` (`:3672`) while the UI thread may be mid-frame reading `g_settings.perspective` and `g_perf.frameIntervalMs`. Posting the reload to the controller window and doing `LoadSettings()`/`DetectPerformanceProfile()` there would remove the race. (Also: re-running hardware detection on every settings change isn't needed — the RAM and core count don't change.)
> 
> **Small ones:**
> 
> * `GetDpiForWindow` is resolved dynamically via `GetProcAddress` (`:866`), but it's available since Windows 1607 and the mod already requires 1809+ — just call it directly.
> * `AlphaBlend` is `LoadLibraryEx`'d from `msimg32.dll` (`:1360`); adding `-lmsimg32` to `@compilerOptions` and calling it directly is simpler. (The `LOAD_LIBRARY_SEARCH_SYSTEM32` flag is correct as written, so this is purely a simplification.)
> * `int h = std::max(minW, ...)` (`:2402`) clamps a _height_ with a width-derived minimum. It works, but the name makes it read like a bug.
> * `IsFlipEligibleWindow` is re-evaluated for every hwnd in `ActivateFlip3DImpl` (`:2768`) right after `EnumerateFlipEligibleWindows` already filtered on it, and `GetRepresentativeWindowRect` runs twice per window as a result.
> * `wParam == kAnimationTimerId || wParam == g_animationTimerId` (`:2995`) — `SetTimer` with an explicit ID returns that ID, so the two are always equal.
> * The local `winTabOwnedElsewhere` (`:3235`) actually holds "owned by _us_"; the name says the opposite.
> 
> Functionality notes

I've addressed most of the reported issues. I think that one last AI review command should be enough.
/ai-review

babamohammed2022:
Thanks for review. I think that after multiple iteractions the mod is overall decent for a first version and it works correctly on my setup. I'd prefer to address the other issues in a future update along with user feedback since I'm mantaining multiple mods at the same time.
I'm ready for the human review.
/ready-for-reviewer

PR #4921, opened 2026-07-28
Add Taskbar audio device switcher

umnik1:
Adds `mods/taskbar-audio-device-switcher.wh.cpp`, a new mod.

The mod adds one tray icon per connected audio device, using the icon of the device itself (`PKEY_DeviceClass_IconPath`), so headphones, speakers, an HDMI output and a microphone are easy to tell apart. The icon of the current default device is drawn normally, the other ones are faded and grayscale.

* **Left click** — make the device the default one (console + multimedia, optionally communications).
* **Middle click** — set it as the communications device.
* **Right click** — a menu with the default/communications device options, a **Shown devices** submenu to pick which devices get an icon, and shortcuts to the sound settings, the volume mixer and `mmsys.cpl`.

The list of devices is kept up to date through `IMMNotificationClient`, so plugging a device in or out, enabling/disabling it, or changing the default device elsewhere is picked up automatically. Switching the default device is done with the undocumented `IPolicyConfig` interface, with a fallback to the Vista variant.

Devices hidden from the tray menu are stored in the mod storage rather than in the settings, since a mod can only read its settings. Tray icon ids are derived from the device id, so the Windows 11 "show this icon on the taskbar" choice survives restarts and reconnects.

The mod hooks nothing — it only runs its own hidden window and message loop in a thread inside `explorer.exe`, and cleans up its window, class, icons and COM references in `Wh_ModUninit`.

## Changelog

* New mod, no changelog.

## Mod authorship

This mod was created by:

- - [x] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/taskbar-audio-device-switcher.wh.cpp
+++ b/mods/taskbar-audio-device-switcher.wh.cpp
@@ -2,10 +2,10 @@
 // @id              taskbar-audio-device-switcher
 // @name            Taskbar audio device switcher
 // @description     Shows a tray icon for every connected audio device and switches the default device with a click
-// @version         1.1.0
+// @version         1.2.0
 // @author          Maksim Chingin
 // @github          https://github.com/umnik1
-// @include         explorer.exe
+// @include         windhawk.exe
 // @architecture    x86-64
 // @compilerOptions -lole32 -loleaut32 -lshell32 -lgdi32 -luser32
 // ==/WindhawkMod==
@@ -22,6 +22,20 @@ options.
 
 Each device uses its own icon, the one Windows shows in the sound settings, so
 headphones, speakers, HDMI output and a microphone are easy to tell apart.
+
+## How this differs from the other audio switching mods
+
+The catalog already has several mods for switching audio devices, but the
+interaction model here is different: instead of one icon that you click through
+or scroll on, every device gets its own icon, with its own graphic, and you
+click the device you want directly. Nothing is cycled, and no list of devices
+has to be configured in advance.
+
+* **Audioswap** / **Microswap** — a single tray icon which cycles through a
+  preconfigured list of up to 6 devices.
+* **Mic Tray Control** — one microphone icon, focused on mute and volume.
+* **Audio scroll switcher** — switches devices by scrolling over the taskbar,
+  without adding any icon.
 
 ## Choosing which devices are shown
 
@@ -40,14 +54,20 @@ New tray icons are hidden inside the overflow ("^") flyout by default. To make
 the audio device icons always visible, either drag them out of the flyout onto
 the taskbar, or go to
 *Settings → Personalization → Taskbar → Other system tray icons* and turn the
-relevant `explorer.exe` entries on.
+`windhawk.exe` entries on.
 
 ## Notes
 
+* The mod runs in a dedicated `windhawk.exe` process and doesn't hook anything,
+  see [Mods as
+  tools](https://github.com/ramensoftware/windhawk/wiki/Mods-as-tools:-Running-mods-in-a-dedicated-process).
 * Devices are refreshed automatically when a device is plugged in/out, enabled,
   disabled, or when the default device changes from somewhere else.
 * Switching the default device uses the undocumented `IPolicyConfig` interface,
   the same one used by nircmd, SoundSwitch and similar tools.
+* Middle click sets the device as the communications device only. With the
+  default settings a left click already does that too, so middle click is only
+  useful if the *Switch the communications device too* option is turned off.
 */
 // ==/WindhawkModReadme==
 
@@ -78,11 +98,15 @@ relevant `explorer.exe` entries on.
 */
 // ==/WindhawkModSettings==
 
+#include <windhawk_utils.h>
+
 #include <windows.h>
 #include <shellapi.h>
 #include <propsys.h>
 #include <mmdeviceapi.h>
+#include <stdio.h>
 
+#include <atomic>
 #include <cstring>
 #include <cwchar>
 #include <string>
@@ -199,15 +223,15 @@ struct AudioDevice {
     bool isDefaultComm = false;
     // Hidden through the tray menu, can be toggled back from there.
     bool hidden = false;
-    // Hidden through the "Hidden devices" setting, can only be brought back by
-    // changing the setting.
+    // Hidden through the "Always hidden devices" setting, can only be brought
+    // back by changing the setting.
     bool hiddenBySetting = false;
     UINT trayId = 0;
 };
 
 struct TrayIcon {
     UINT trayId = 0;
-    HICON icon = nullptr;
+    HICON icon = nullptr;  // owned by the icon cache, not by this struct
     std::wstring tip;
 };
 
@@ -223,12 +247,23 @@ Settings g_settings;
 
 HANDLE g_thread;
 HANDLE g_windowReadyEvent;
-HWND g_hWnd;
+// Written by the mod thread, read from the Windhawk callbacks and from the
+// MMDevice notification threads.
+std::atomic<HWND> g_hWnd;
 UINT g_taskbarCreatedMessage;
+int g_iconSize = 16;
 
 std::vector<AudioDevice> g_devices;  // all devices, hidden ones included
 std::unordered_map<std::wstring, TrayIcon> g_trayIcons;  // device id -> icon
 std::vector<std::wstring> g_hiddenIds;
+
+// Extracting an icon from a DLL is expensive and the device list is rebuilt on
+// every notification, so the built icons are kept around. The cache owns the
+// handles.
+std::unordered_map<std::wstring, HICON> g_iconCache;
+
+// Kept alive for the notification callback and reused for the enumeration.
+IMMDeviceEnumerator* g_enumerator;
 
 using PrivateExtractIconsW_t = UINT(WINAPI*)(LPCWSTR, int, int, int, HICON*, UINT*, UINT, UINT);
 PrivateExtractIconsW_t g_pPrivateExtractIcons;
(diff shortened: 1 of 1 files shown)


m417z:

umnik1:
Thanks for the review. Pushed as `v1.2.0`.

### 1. Tool mod — done

Agreed, and the duplicate-icons argument is the decisive one. Converted to `@include windhawk.exe`, renamed the entry points to `WhTool_ModInit` / `WhTool_ModSettingsChanged` / `WhTool_ModUninit`, and added the launcher block. I took the block from `mic-tray-control.wh.cpp` rather than retyping it, and diffed it against `theme-toggler-tray.wh.cpp` — the two are byte-identical, so it should be the canonical snippet.

### 2. Unload path — done

Resolved by the conversion: `WhTool_ModUninit` is followed by `ExitProcess(0)`, so a worker thread that overruns the wait can no longer end up running inside an unloaded image.

I also dropped the `ERROR_CLASS_ALREADY_EXISTS` re-registration fallback rather than keeping it. Your reading of it was right — it could only ever trigger as a consequence of the race above, and in a tool mod the process is fresh on every load, so the class cannot pre-exist.

### 3. Screenshot — not done yet

I'll add screenshots of the tray row and of the right-click menu in a follow-up commit, hosted on one of the two allowed hosts. Please treat this item as open until then.

### 4. Comparison with the existing mods — done

Added a *How this differs from the other audio switching mods* section naming audioswap, microswap, mic-tray-control and audio-scroll-switcher, and describing the one-icon-per-device model.

### Optional items — all applied

* `WindhawkUtils::StringSetting` in `LoadSettings` (the `i < 64` cap is gone).
* `g_hWnd` is now `std::atomic<HWND>`.
* The `IMMDeviceEnumerator` held for the notification callback is reused for the enumeration.
* `SetDefaultDevice` takes the device id **by value**, which also removes the aliasing you pointed at — the copy is made at the call site, before any outgoing COM call.
* `ExpandEnvironmentStringsW`'s return value is checked against the buffer size.
* `mmsys.cpl` is opened directly instead of going through `rundll32.exe` on the search path.
* `RemoveAllTrayIcons(true)` on a settings change is gone.
* `WhTool_ModSettingsChanged` waits on the ready event, so an early settings change isn't dropped.
* Fixed the comments that still called the setting "Hidden devices".

### Functionality notes — applied, except one

Applied: an icon cache keyed by `(size, dimmed, flow, iconPath)`; refreshes on `WM_SETTINGCHANGE` / `WM_DISPLAYCHANGE` / `WM_DPICHANGED` now only happen when the icon size actually changed, which removes the pathological case you describe; `GetSystemMetricsForDpi` with the DPI of the taskbar's monitor; only a trailing `,<number>` is treated as an icon index; `SetDefaultEndpoint`'s own `HRESULT` is checked before falling back to the Vista interface; `SndVol.exe` fallback for `ms-settings:apps-volume`; a readme line about middle click.

**Not applied: the premultiplied alpha suggestion.** You suggested checking it before changing anything, so I measured it. I extracted the real `mmres.dll,-3010` icon at 16x16, read its colour bitmap with `GetDIBits`, composited the icon over a solid `#808080` background through the normal icon drawing path, and compared each partially transparent pixel against both models:

```
icon 16x16, background=128
raw BGRA           actual   straight  premul
185,185,183,159    R=162    162.3     231.2
169,160,151, 16    R=129    129.4     255.0
162,153,144,239    R=142    143.0     152.0
...
partially transparent pixels: 12
closer to straight alpha : 12 px (mean abs err 0.61)
closer to premultiplied  : 0 px (mean abs err 60.63)
pixels with a channel > alpha: 7 (non-zero rules out premultiplied)
```

The data is straight, non-premultiplied alpha, and it is composited as such: 7 pixels have a colour channel greater than their alpha, which cannot happen in valid premultiplied data. Scaling R/G/B along with the alpha would blend the icon toward black rather than fade it, so I kept the alpha-only scaling and added a comment in `MakeDimmedIcon` recording why.

Happy to be corrected if you're measuring something different, e.g. if the shell's own drawing path differs from `DrawIconEx` here.

umnik1:
Item 3 is done now — screenshots added in f66c7c2, so all four review items are addressed.

Tray icons \
*Two playback devices: the default one is drawn normally, the other one dimmed*

Right click menu \
*The right click menu of a device*

The images are hosted on an orphan `assets` branch of my fork rather than committed here, so that the pull request stays the single `mods/taskbar-audio-device-switcher.wh.cpp` file. I checked that both URLs are served anonymously and that they match the `URL_PATTERN` in `scripts/archive_mod_images.py`, so `archive_mod_images.py` should pick them up.

Note that the two workflow runs on the latest commit are sitting in `action_required` and need a maintainer to approve them. The previous commit (41c1675) passed everything, and this one only adds the six readme lines above, with no code change.

umnik1:
Thanks — pushed as `v1.3.0` (b1568d3).

### 1. Re-entrancy in `WindowProc` — fixed

Agreed, and the mechanism holds up: `TaskbarCreated` is broadcast with `SendNotifyMessage`, which for another thread is delivered as an asynchronously *sent* message, so it can be dispatched while the thread is parked inside `Shell_NotifyIconW` — and the handler cleared the very container the caller was walking. Deferred it to the refresh timer and made `RefreshDevices` non-reentrant with the `g_refreshing` / `g_refreshPending` / `g_trayIconsLost` flags, as suggested. That also covers the `WM_DESTROY` → `RemoveAllTrayIcons(true)` loop.

### 2. Icon size never re-derived — fixed

Recomputed on `TaskbarCreated`, and I went slightly further: `GetTrayIconSize` now reports whether the value actually came from the taskbar, and any refresh that follows a fallback value re-derives it. So recovery doesn't depend on a single event arriving.

### Optional items — all applied

`DestroyWindow` before clearing `g_hWnd`; middle click short-circuits on `isDefaultComm`; exclusion patterns lowercased once in `LoadSettings`; `@architecture` dropped — nothing here is 64-bit specific, and I verified the mod compiles for `i686`, `amd64` and `arm64`.

On the `.cpl` verb: you asked me to check whether the current call actually works, and it does not. `AssocQueryStringW(ASSOCSTR_COMMAND, L".cpl", L"open", ...)` returns `0x80070483` (`ERROR_NO_ASSOCIATION`), and `cplfile\shell` only has `cplopen` and `runas`. So the item was broken by my previous round when I replaced `rundll32` with the `open` verb. Now using the default verb with an absolute `System32` path, same for the `SndVol.exe` fallback. Good catch.

### One item doesn't hold up: the `ExtractIconExW` fallback

> negative indices (resource IDs) are only supported by `PrivateExtractIconsW`; `ExtractIconExW` treats the index as an ordinal

I wrote the `LoadLibraryEx` + `LoadImage` replacement first, then measured before keeping it — and reverted it. Comparing the 32bpp bits of the icons each path returns for `mmres.dll`:

```
ExtractIconExW small vs resource id 3010: 16x16, differing pixels: 0 of 256
PrivateExtractIconsW  vs resource id 3010: 16x16, differing pixels: 0 of 256
resource id 3011      vs resource id 3010: 16x16, differing pixels: 139 of 256
```

`ExtractIconExW(path, -3010, ...)` returns an icon that is pixel-identical to resource id 3010, and the last line is the control showing the comparison discriminates. This matches the documented behaviour: an `nIconIndex` that is negative and not `-1` is the resource id. So the fallback does handle the form that occurs, and I kept it as it was rather than adding code for a non-problem. I did add a comment saying so, so it doesn't get "fixed" later.

### Functionality notes — two deliberate keeps

* **Hidden ids are never pruned** — intentional, and it's the behaviour you flagged as possibly wanted: a USB DAC that gets unplugged and plugged back in should stay hidden. Kept as is.
* **`NIF_GUID`** — with `NIF_GUID` the `uID` member is documented as ignored, and the version 4 callback identifies the icon through `uID` in `HIWORD(lParam)`. Switching would remove the field the click handler uses to tell devices apart, and would need one window per icon to compensate. The 16-bit hash plus collision fixup works, so I left it.

The remaining two notes (middle click, shared `HICON` for a shared `iconPath`) are handled and intentional respectively.

umnik1:
Thanks — pushed as `v1.4.0` (62f882b). The finding is right, and it needed one more change than proposed.

### Fixed at the root: the refresh owns `g_iconSize`

Rather than making `PurgeStaleIcons` tolerate a size change mid-refresh, I removed the possibility. The handlers that notice a change now only *read* the size and set a `g_iconSizeStale` flag; the refresh re-derives it at its start, where no iterator is held and nothing is in flight:

```cpp
// WM_SETTINGCHANGE / WM_DISPLAYCHANGE / WM_DPICHANGED
if (IconSizeChanged()) {        // reads, never stores
    g_iconSizeStale = true;
    SetTimer(hWnd, kRefreshTimerId, 500, nullptr);
}
```

`g_iconSize` now has exactly one write site, reachable only from `WhTool_ModInit`'s thread startup and from the top of `RefreshDevices`, so the invariant in the comment holds by construction. I applied this to the `TaskbarCreated` arm as you suggested, and to the `WM_SETTINGCHANGE` arm as well — that one is the trigger in your own scenario.

### The `inUse` guard alone doesn't cover the second case

Applied it, but it does not fix the `icon == nullptr` path on its own. That path assigns the stored handle *before* the purge runs:

```cpp
Shell_NotifyIconW(NIM_MODIFY, &nid);   // no NIF_ICON, shell keeps the old icon
it->second.icon = icon;                // icon == nullptr
```

So by the time `PurgeStaleIcons` walks the cache, nothing in `g_trayIcons` references the old handle any more — `entry.second.icon == it->second` compares `nullptr` against it and fails, `inUse` stays false, and the handle is destroyed while the shell is still displaying it. Exactly the case the guard was meant to catch.

It needs the recorded handle to keep reflecting what the shell actually shows:

```cpp
Shell_NotifyIconW(NIM_MODIFY, &nid);
if (icon) {
    it->second.icon = icon;
}
```

With both changes together the guard protects both paths. Worth noting in case the same suggestion goes out on another submission.

Verified the mod still compiles for `i686`, `amd64` and `arm64`, and links.

umnik1:
Thanks — pushed as `v1.5.0` (d7d6b17). Good catch, and applied as proposed.

The trap is real: the `visibleCount <= 1` guard only holds at menu-build time, and the device it protects can disappear afterwards. Since the manual list lives in the mod storage, which the settings UI cannot clear, the only ways out were reconnecting the device or removing the mod. The fallback now runs at refresh time, with `hiddenBySetting` left alone.

It composes well with the existing menu guard, which I checked: after the fallback kicks in, the single remaining device is drawn both `MF_CHECKED` and `MF_GRAYED` — shown, and not hideable — which is exactly right. *Show all devices* doesn't appear at that moment because nothing in the current list is hidden, but it comes back as soon as the other device returns and the manual list applies again, which is also when it's meaningful. The stored list is deliberately left untouched, so the original intent survives the reconnect.

I brute-forced the invariant rather than reasoning about it. Pulling the block out and running it over every combination of flags for lists of 0 to 4 devices:

```
checked 121 combinations, violations: 0
```

Two conditions checked: a visible device always exists except for an empty list or an all-`hiddenBySetting` one, and the fallback never overrides the setting.

Also added a readme line about the behaviour so the docs don't drift from the code, and verified the mod still compiles for `i686`, `amd64` and `arm64`, and links.

PR #4923, opened 2026-07-28
Add Explorer Visual Tweaks Dark

VitalSkib:
Explorer Visual Tweaks Dark primarily fixes inconsistent colors and visual states in File Explorer, while providing coordinated customization for the affected elements in one mod.

Features:

* consistent Preview Pane frame and content backgrounds;
* matching plain-text preview background, text color, and scrollbar;
* customizable ItemsView and Navigation Pane selection backgrounds;
* customizable Navigation Pane focus indicator;
* customizable Explorer progress resources with gradients, borders, and corner rounding;
* support for Explorer-based Open/Save dialogs through Windhawk's Inclusion List.

### Design notes

Selection and Progress rendering is filtered by theme classes and specific part/state identifiers. Progress rendering intentionally follows the corresponding `PROGRESS` resources wherever Explorer uses them, without window-tree or size heuristics.

The Selection renderer also suppresses a reproducible Explorer visual artifact in which the previously selected item can remain highlighted while the newly clicked item receives a second, brighter frame, resembling multi-selection without Ctrl or Shift. I reported this through both [Microsoft Q&A](https://learn.microsoft.com/en-us/answers/questions/5916895/file-explorer-shows-false-multi-selection-with-blu) and [Feedback Hub](https://aka.ms/AA13icjy). The mod changes the affected visual states without modifying Explorer's underlying selection logic. Navigation Pane selection without focus retains its native appearance.

The Preview Pane class brush is installed only when the target window class is registered. After successful registration, ownership follows the documented [WNDCLASSEXW::hbrBackground contract](https://learn.microsoft.com/en-us/windows/win32/api/winuser/ns-winuser-wndclassexw): Windows automatically deletes a class background brush when the class is unregistered, and applications should not delete it themselves. The mod therefore intentionally leaves the brush class-owned. Preview Pane setting changes and complete restoration after disabling the mod require restarting File Explorer and any included host applications. Selection and Progress settings apply immediately.

On Windows 11, the private UxTheme functions `GetThemeClass`, `AllowDarkModeForWindow`, and `SetPreferredAppMode` are resolved directly through export ordinals 74, 133, and 135, following the [System Informer reference](https://github.com/winsiderss/systeminformer/blob/1ab5a1b98bf1de5ea92d189a29d1f28c476e27a4/phlib/guisup.c#L128-L139). Required function addresses are checked during initialization. DirectUI functions are resolved by exact decorated export names. No PDB downloads, hardcoded RVAs, or structure offsets are used.

`DUI70.dll` is loaded from System32 and pinned at the end of successful initialization because a hook target resides in that module.

### Screenshots

Explorer selections and drive progress

Preview Pane and plain-text preview

## Changelog

N/A (New mod)

## Mod authorship

This mod was created by:

* [ ] The submitter, without AI assistance
* [x] The submitter, with AI assistance
* [x] Claude
* [x] ChatGPT
* [ ] Gemini
* [ ] Another AI (please specify):
* [ ] Other (please specify):


m417z:

VitalSkib:
The implementation has been substantially rewritten since the previous AI review. That review targeted commit `fddcdce85b90a30d990cce35be2bc0bed0586b03`; the current implementation is at `66dfd79f85be4342cc3ad9b1c64e3ee322f2c28d`.

In particular, the current version no longer performs live class-brush replacement or deletion, doesn't use loader-notification callbacks, doesn't modify undocumented DirectUI value structures, and synchronizes shared rendering resources. Symbol resolution no longer uses RVAs or ordinals, the Preview Pane lifecycle follows the documented class-brush ownership contract, and the README now documents the restart requirement, Inclusion List scope, and screenshots.

The combined scope is intentional: the primary purpose is to correct related Explorer color inconsistencies in a coordinated way, with customization provided as an additional benefit. Progress replacement deliberately targets Explorer's exact `PROGRESS` theme resources wherever they are used.

Please review the current implementation independently from the earlier version.

VitalSkib:
Thank you for the detailed review. I went through each point separately and made changes where the finding represented a real issue. For the remaining points, I performed additional runtime diagnostics or clarified the intended lifecycle contract.

### 1. Overlap with existing mods

I understand the concern about catalog fragmentation. The overlap is intentional, but the mod is not positioned as another general-purpose progress-bar or TreeView customizer.

Its primary purpose is to correct several related visual inconsistencies in File Explorer with one coordinated palette:

* inconsistent Preview Pane frame and content backgrounds;
* plain-text preview background, text color, and scrollbar;
* the stale/false selection transition in ItemsView;
* inconsistent ItemsView and Navigation Pane selection states;
* drive progress resources that otherwise retain unrelated theme colors.

The Preview Pane/plain-text-preview correction and ItemsView ghost-selection handling aren't covered by the referenced mods.

The integrated Progress renderer also provides exact-size rendering with independently configurable left/right gradient colors, full-state gradient colors, background, border, inset, and corner rounding. It intentionally replaces the corresponding `PROGRESS` theme resources wherever Explorer uses them, without window-tree, size, or caller-address heuristics.

The Navigation Pane support is coordinated with the ItemsView selection colors and adds the narrow WinUI-style focus pill. Keeping these components together also avoids requiring users to combine several mods that hook the same theme-rendering APIs and configure matching colors independently.

I have clarified this consistency/fix-first purpose in the PR description and included screenshots showing the affected areas together.

### 2. Qualified theme class names

Addressed.

`IsThemeClass` now accepts either the exact class name or a class-list-qualified name ending in `::ExpectedClass`:

```cpp
PCWSTR tail = className + classLength - expectedLength;
if (_wcsicmp(tail, expected) != 0)
    return false;

return tail == className ||
       (tail - className >= 2 &&
        tail[-2] == L':' && tail[-1] == L':');
```

The `::` boundary prevents unrelated names that merely end with the same text from matching.

Runtime diagnostics on my current Explorer build return the bare `TreeView`, `ListView`, and `PROGRESS` names, but the code now also handles forms such as `Explorer::TreeView` and `DarkMode_Explorer::TreeView`.

### 3. DirectUI queries outside a DirectUI paint

Addressed.

`IsTransientSingleSelectionGhost` now returns before calling any DirectUI getters unless the current thread is inside the hooked `DirectUI::Element::PaintBackground` call:

```cpp
void* oldElement = g_currentPaintElement;
if (!oldElement)
    return false;
```

The ItemsView transition branch applies the same condition before calling `GetPendingMouseSelectionElement()`.

Therefore, process-wide theme calls on unrelated Explorer window threads no longer query or dereference a DirectUI element without a matching thread-local paint context.

### 4. `TREIS_SELECTED` suppression

I verified this with two diagnostic builds rather than relying on visual inspection.

The second diagnostic correlated every TreeView theme call with:

* the actual `SysTreeView32` window;
* `TreeView_GetSelection`;
* the selected `HTREEITEM`;
* `TreeView_GetItemRect`;
* keyboard focus;
* the item under the pointer.

On the tested Windows 11 build, a genuinely selected Navigation Pane item uses `TREIS_HOTSELECTED` while `SysTreeView32` has keyboard focus, even when the pointer is over another item.

Examples from the trace had all of the following simultaneously:

* `focusClass='SysTreeView32'`;
* the themed row matched the item returned by `TreeView_GetSelection`;
* `pointerOnSelected=0`;
* the state remained `TREIS_HOTSELECTED`.

In that complete run the counts were:

* `TREIS_HOTSELECTED`: 19;
* `TREIS_SELECTEDNOTFOCUS`: 95;
* `TREIS_SELECTED`: 0.

The earlier diagnostic captured one rare unmodified `TREIS_SELECTED` call. Its ordering showed that it belonged to the previously selected row during the transition to a new row:

1. the old row was drawn as `HOTSELECTED`;
2. the user selected another row;
3. the old row received one `SELECTED` draw;
4. the new row became `HOTSELECTED`.

The newer diagnostic also showed Explorer drawing transition states for a row after `TreeView_GetSelection` had already moved to a different item.

Therefore, on the tested build, the unmodified `TREIS_SELECTED` call is the stale old-row transition that the mod intentionally suppresses, not the normal focused selection background. Returning the original renderer there would restore the visual artifact.

`TREIS_SELECTEDNOTFOCUS` is now passed through to the original renderer unchanged, so the legitimate unfocused selected state retains its native appearance.

### 5. Process-wide preferred app mode

This is intentional and limited to `prevhost.exe`, not `explorer.exe` or arbitrary Explorer windows.

Runtime testing showed that applying `AllowDarkModeForWindow` to the RichEdit window without setting the preferred app mode left its scrollbar light. The narrower per-window theme alternative was also insufficient for the complete preview-host behavior. `ForceDark` was retained because it was the only tested configuration that consistently produced the intended RichEdit scrollbar in the isolated preview host.

Preview configuration follows the same static process-lifetime contract as the class background brush. Restoring only the preferred app mode while the class-owned background remains installed would create a partially restored and internally inconsistent Preview Pane.

The README therefore explicitly requires restarting File Explorer after changing or disabling Preview settings. I also verified at runtime that restarting Explorer terminates the old `prevhost.exe`; the previous PID no longer exists afterward. Thus the process-wide state is discarded with the dedicated preview-host process.

Selection and Progress remain independently reloadable without a restart.

### 6. Preview Pane class-brush ownership

The class brush is intentionally installed only during registration of the two exact target classes.

After a successful `RegisterClassExW`, ownership follows the documented [`WNDCLASSEXW::hbrBackground` contract](https://learn.microsoft.com/en-us/windows/win32/api/winuser/ns-winuser-wndclassexw): Windows automatically deletes class background brushes when the class is unregistered, and applications should not delete them themselves.

The mod therefore atomically relinquishes its local ownership only after successful class registration. If registration fails or the brush was never transferred, the mod deletes it itself. This avoids both double deletion and leaking an unowned brush.

I also logged the original brushes used by both target classes:

* `Shell Preview Extension Host` used the system `COLOR_WINDOW` brush;
* `Shell Preview Extension Host Previewer` used the corresponding system-color pseudo-brush value.

Neither observed original is a caller-created GDI brush that the mod could orphan.

I considered and tested the subclass/`WM_ERASEBKGND` direction while investigating live updates. It changes the implementation from supplying the class's native background resource to manually replacing part of its painting behavior. It also introduces per-window subclass lifecycle, cross-thread removal, and in-flight paint/resource-lifetime problems without preserving the original class-brush semantics.

For that reason, the mod deliberately uses the documented class-registration ownership model and accepts the explicit restart requirement instead of adding a more invasive live-update mechanism.

### 7. Preview settings reload

The absence of Preview reload in `Wh_ModSettingsChanged` is intentional.

The Preview background, ReadingPane fill, RichEdit colors, scrollbar mode, and class brush are treated as one consistent configuration snapshot. Reloading only `GetThemeColor`/`GetSysColor` values would update some areas while the class-owned Host/Previewer background remained on the old color until process teardown.

That partial update was tested and produced visibly inconsistent Preview Pane regions.

Consequently:

* Selection and Progress settings reload immediately;
* Preview settings remain immutable after initialization;
* all Preview changes, as well as complete restoration after disabling the mod, require restarting Explorer and the associated preview host.

This behavior is stated explicitly in the README rather than silently presenting a partial live update as complete support.

VitalSkib:
Thank you. The updated review confirms that the two actionable code issues were addressed: qualified theme class names are now handled, and DirectUI element queries are gated by the thread-local paint context.

The remaining four required findings were addressed in detail in my previous comment, including runtime diagnostics. In summary:

1. The overlap is intentional: this is a coordinated visual-consistency mod rather than a standalone progress or TreeView customizer. The Preview Pane correction and ItemsView stale-selection suppression are unique parts of the implementation, while the integrated renderers provide one consistent palette without requiring several mods that hook the same APIs.

2. The Preview class brush deliberately follows the documented class-registration ownership contract. `SetClassLongPtr` was investigated and rejected because restoring and deleting a class-wide brush cannot be made safe while independent Explorer UI threads may still be painting with the old handle. The static process-lifetime behavior and required restart are documented.

3. `ForceDark` is confined to the dedicated `prevhost.exe` process. Runtime testing showed that `AllowDarkModeForWindow` without the preferred app mode left the RichEdit scrollbar light. Restarting Explorer was verified to terminate the old preview-host process, which discards this process-wide state.

4. Preview settings intentionally form one immutable configuration snapshot. Partially reloading RichEdit/theme colors while the class-owned background remains unchanged produced inconsistent Preview Pane regions. Selection and Progress reload immediately; Preview changes require the documented restart.

I also verified the `TREIS_SELECTED` concern with two diagnostic builds. A genuinely selected, focused Navigation Pane item uses `TREIS_HOTSELECTED` even when the pointer is elsewhere. The rare unmodified `TREIS_SELECTED` call belonged to the previously selected row during the native transition to another item, so suppressing it prevents the stale-selection artifact. `TREIS_SELECTEDNOTFOCUS` is delegated to the native renderer.

No further code changes are planned for these findings.

m417z:
These functions are exported by ordinal:
```
    WindhawkUtils::SYMBOL_HOOK uxThemeDllHooks[] = {
        {{L"GetThemeClass"}, &g_getThemeClass},
        {{L"AllowDarkModeForWindow"}, &g_allowDarkModeForWindow, nullptr,
         !requireDarkMode},
        {{L"SetPreferredAppMode"}, &g_setPreferredAppMode, nullptr,
         !requireDarkMode},
    };
```
You don't need to use debug symbols which are slower and require internet connection.

Here's a reference that contains all the export numbers:
https://github.com/winsiderss/systeminformer/blob/1ab5a1b98bf1de5ea92d189a29d1f28c476e27a4/phlib/guisup.c#L128-L139

VitalSkib:
@m417z 
Thanks for the reference. I understand that these functions can be resolved directly through ordinals 74, 133, and 135, avoiding symbol loading.

Using `HookSymbols` was a deliberate choice: it identifies these private functions by their expected names rather than relying on a fixed ordinal-to-function mapping. If an ordinal were reassigned in a future Windows build, a non-null result from `GetProcAddress` would not establish that it still points to the intended function. Name-based resolution avoids that particular risk and lets initialization fail if a required symbol is no longer available.

I’m not suggesting that these ordinals change frequently, or that symbol names guarantee ABI compatibility. A function’s signature or behavior could still change under the same name. This is simply the specific compatibility risk the current approach aims to avoid.

I acknowledge the startup and network cost when the required symbols aren’t cached; `HookSymbols` caching reduces repeated work but doesn’t eliminate that dependency.

Would keeping name-based resolution be acceptable given this trade-off, or is direct ordinal resolution required for acceptance?

m417z:
I don't think these ordinals will change soon, or ever. Too many apps are using it, and it's being suggested in StackOverflow etc., example:
https://stackoverflow.com/questions/53501268/win10-dark-theme-how-to-use-in-winapi

Symbols can change too, and since the symbols don't contain the function signature, a function's ABI can change with a symbol query as much as it can with an ordinal, so I don't think there's a great advantage in using symbols.

VitalSkib:
Thanks for the explanation. I've updated all three functions to use `GetProcAddress` with ordinals 74, 133, and 135, following the reference you provided.

The existing required-function checks are preserved, including the conditional requirement for the dark-mode functions. I also removed the unused `windhawk_utils.h` include and updated the PR description. No rendering or resource-lifetime changes were made.

PR #4927, opened 2026-07-28
Files 2 Folders v2.6.2 (#4654)

triatomic:
Two commits: a bugfix release (v2.5.3) and a new feature (v2.6).

## v2.5.3 — bugfixes

**The main fix.** Joining a drive root with a subfolder name produced a doubled separator — `"C:\"` + `"\"` + name gives `C:\New Folder`. `MoveFileExW` tolerates that, so "Move - fast" kept working, but `SHCreateItemFromParsingName` rejects it with `E_INVALIDARG`. Every destination item failed to queue, `PerformOperations` was never reached, and the operation completed with no files moved and no error shown.

Also in that commit:

- **Undo works for "Move - safe" and "Copy"** — `IFileOperation` records no undo information unless `FOFX_ADDUNDORECORD` is set.
- **Protected destinations elevate** via `FOFX_SHOWELEVATIONPROMPT`, letting the shell's own copy engine raise the permission prompt. Nothing is probed up front; the direct call is tried first and the shell is only asked to elevate after a real permission error. Elevation is decided per run, so a protected destination costs one consent prompt rather than one per file, and declining once stops the run.
- **"Move - fast" cannot elevate**, so it no longer creates the destination elevated either — it reports which modes can, before anything is created.
- **Fixed a hang**: the elevated-create parent walk treated a drive root as its own parent, so a root that `GetFileAttributesW` cannot stat spun forever on Explorer's UI thread.
- `ParentDir` and the `ExtractFolderAndSelection` fallback kept only `"C:"` for a drive-root item — a drive-*relative* path resolving against the current directory.
- The failure message box is built as a `std::wstring` rather than `wsprintfW` into a fixed `WCHAR[1024]`.
- `Wh_ModSettingsChanged` stops the hotkey thread before reloading settings.

## v2.6 — Defolder

A fourth button in the dialog's Operation control, doing the reverse of the existing operation: it moves the contents of the selected folders **out** into the folder being browsed, then offers to remove the folders left empty. Recursive — files at any depth are pulled up and the inner structure is discarded.

Since it has no destination to choose, selecting it greys out the four subfolder-mode radios and their inputs, and the header changes to describe what will happen.

It asks for one summary confirmation covering both halves ("Move 47 files out of 3 folders into this one, and then remove the folders that end up empty?"), defaulting to No.

Safety properties:

- Folders are removed with `RemoveDirectoryW`, never a recursive delete — that call fails on a non-empty directory, so anything that could not be moved keeps its folder alive. Nothing is deleted with contents still in it.
- Reparse points (junctions, symlinks) are skipped rather than followed, so a link inside the selection cannot pull in outside files or loop.
- Recursion is capped at 64 levels.
- Only folders in the selection are processed; selected files are ignored.

A new **"Defolder: use the safe move engine"** setting (on by default) chooses between the shell engine (Ctrl+Z undo, conflict and permission prompts) and the fast direct move (auto-renames collisions, cannot be undone).

### Mod authorship

- [x] Claude
- [ ] The submitter, manually

### DIFF
--- a/mods/files-2-folders.wh.cpp
+++ b/mods/files-2-folders.wh.cpp
@@ -2,7 +2,7 @@
 // @id              files-2-folders
 // @name            Files 2 Folders
 // @description     Move or copy one or more selected files in Explorer into a subfolder (named — nested paths with "/" supported, by extension, by name, or by date), with a workaround hotkey for other file managers
-// @version         2.5.3
+// @version         2.6
 // @author          tria
 // @github          https://github.com/triatomic
 // @include         explorer.exe
@@ -98,6 +98,31 @@ The **Operation** setting controls what happens to the selection:
   leaving the originals in place. Same shell progress / undo / conflict UI as
   the safe move.
 
+## Defolder — the reverse operation
+**Defolder** is the fourth button in the dialog's Operation control. Instead of
+moving files *into* a new subfolder, it moves the contents of the folders you
+selected *out* into the folder you are browsing, and then offers to remove the
+folders that were left empty.
+
+It is **recursive**: files at any depth are pulled up to the top level, so the
+inner folder structure is discarded. Select a few folders, pick **Defolder**,
+and confirm the summary ("Move 47 files out of 3 folders...") — the subfolder
+mode options are greyed out while it is selected, because there is no
+destination to choose.
+
+Notes:
+
+- Only the folders in your selection are touched; any files you also selected
+  are ignored.
+- Folders are removed only if they actually end up empty. Anything that could
+  not be moved keeps its folder alive, so nothing is deleted with contents
+  still in it.
+- Junctions and symbolic links are skipped rather than followed, so a link
+  inside the selection can't pull in files from elsewhere.
+- The **Defolder: use the safe move engine** setting controls whether the files
+  move via the shell (Ctrl+Z undo, conflict prompts, permission prompts) or the
+  fast direct move (auto-renames collisions, cannot be undone).
+
 In copy mode the dialog options say "Copy" rather than "Move", so you can tell
 the configured operation at a glance.
 
@@ -160,6 +185,11 @@ Forbidden characters in folder names (`* : ? " < > | / \`) are replaced with
     On (default): in "Fixed name" mode, if a folder of that name already exists, files are moved into it (e.g. everything goes into "archive").
     Off: a new numbered folder is created instead ("archive", "archive (2)", "archive (3)"…), the way Explorer numbers duplicates.
     Only affects the "Fixed name" mode. "By date" always creates a new numbered folder; "by name"/"by extension" always reuse.
+- defolderUseShell: true
+  $name: "Defolder: use the safe move engine"
+  $description: |-
+    On (default): Defolder moves files with the same Windows file-operation engine as "Move - safe", so you get Ctrl+Z undo, the "Replace or skip files?" prompt, and a permission prompt for protected locations.
+    Off: uses the fast direct move instead, which auto-renames collisions ("file (2).txt") and is quicker for very large folders, but cannot be undone.
 - showNestedPreview: true
   $name: "Fixed name: show nested-folder preview"
   $description: |-
@@ -207,10 +237,12 @@ Forbidden characters in folder names (`* : ? " < > | / \`) are replaced with
     Move - fast (default): instant same-volume rename via MoveFileExW — but no Ctrl+Z undo, no progress bar, and it can't move into locations that need administrator permission.
     Move - safe: moves go through the standard Windows file-operation system, so you get the familiar progress dialog, Ctrl+Z undo, the "Replace or skip files?" prompt for conflicts, and Windows' own permission prompt for protected locations. Noticeably slower, especially with hundreds of files. (A move carried out with administrator permission can't be undone with Ctrl+Z — Windows doesn't record it on your undo stack.)
     Copy: copies the items into the subfolder and leaves the originals in place (same shell progress / undo / conflict UI as the safe move).
+    Defolder: the reverse operation - moves the contents of the selected folders OUT into the folder you are browsing, then offers to remove the folders left empty. Recursive: files at any depth are pulled up to the top level and the inner folder structure is discarded.
   $options:
   - moveFast: "Move - fast (instant)"
   - moveSafe: "Move - safe (progress, Ctrl+Z undo)"
   - copy: "Copy (leave originals)"
+  - defolder: "Defolder (unpack folders)"
 - hotkeyEnabled: true
   $name: "Workaround for other programs"
   $description: |-
@@ -360,12 +392,18 @@ enum F2FOperation {
     F2F_MOVE_FAST = 0,
     F2F_MOVE_SAFE = 1,
     F2F_COPY      = 2,
+    // Not a destination-building operation like the three above: Defolder moves
+    // the *contents* of the selected folders up into the folder being browsed
+    // and then offers to remove the emptied folders. It ignores the subfolder
+    // mode radios entirely, which the dialog greys out while it is selected.
+    F2F_DEFOLDER  = 3,
 };
 
 static const struct { const wchar_t* key; F2FOperation op; } kOperationKeys[] = {
     { L"moveFast", F2F_MOVE_FAST },
     { L"moveSafe", F2F_MOVE_SAFE },
     { L"copy",     F2F_COPY },
+    { L"defolder", F2F_DEFOLDER },
 };
 
 static F2FOperation OperationFromKey(const std::wstring& key) {
@@ -381,6 +419,7 @@ static F2FOperation OperationFromKey(const std::wstring& key) {
 struct ModSettings {
     std::wstring defaultSubfolderName;
     bool fixedNameReuse;
+    bool defolderUseShell;
     bool showNestedPreview;
     bool showAddLevelButton;
     std::wstring dateFormat;
@@ -408,6 +447,7 @@ static void LoadSettings() {
     if (s) Wh_FreeStringSetting(s);
 
     g_settings.fixedNameReuse = Wh_GetIntSetting(L"fixedNameReuse") != 0;
+    g_settings.defolderUseShell = Wh_GetIntSetting(L"defolderUseShell") != 0;
     g_settings.showNestedPreview = Wh_GetIntSetting(L"showNestedPreview") != 0;
     g_settings.showAddLevelButton = Wh_GetIntSetting(L"showAddLevelButton") != 0;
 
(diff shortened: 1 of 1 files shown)


m417z:
### Submission review

---

**1. The `IsDirWritable` probe tests the wrong right, and it has side effects on every run.**

`IsDirWritable` calls `GetTempFileNameW`, i.e. it tests whether a **file** can be created directly in `folder`. The mod never does that — it creates a **subfolder** and writes into that subfolder. On a default Windows install those are different rights on `C:\`: the root ACL grants `Authenticated Users:(AD)` (add subdirectory) on the folder itself and inherit-only `Modify` to children, but no add-file right. So for items in a drive root — the exact scenario this PR fixes — the probe returns false and `needElevation` becomes true, even though creating `C:\New folder` and moving files into it succeeds unelevated (as it did in 2.5, and as Explorer does).

The visible result with the default **Move - fast**: a UAC prompt, an empty admin-created folder in `C:\`, and then a message box saying the operation can't be done. With **Move - safe**: two UAC prompts and a move that is no longer undoable — where nothing needed elevating.

Other false positives from the same probe: `GetTempFileNameW` fails when `dir` is longer than `MAX_PATH - 14`, and it fails on read-only/removable media and on network shares where the user lacks write access — in all of those cases the mod claims "requires administrator permission", and elevation cannot help. It also creates and deletes a temp file in the user's folder on *every* invocation, including the overwhelmingly common writable case, which means sync/AV/`ReadDirectoryChangesW` churn (OneDrive-backed folders in particular) for a question that usually doesn't need asking.

Drop the probe and escalate only after a real failure — that's also what a merged mod in the repo does for the same problem: [explorer-ctrln-newfile.wh.cpp#L912](https://github.com/ramensoftware/windhawk-mods/blob/109589023dde428deaee2fe80e4ce446283a7935/mods/explorer-ctrln-newfile.wh.cpp#L912) tries the direct path first and only calls the shell-elevation helper when `IsPermissionError(err)`:

```cpp
static bool EnsureDir(HWND owner, const std::wstring& path, bool* elevated) {
    if (IsDirectoryPath(path)) return true;
    DWORD err = (DWORD)SHCreateDirectoryExW(nullptr, path.c_str(), nullptr);
    if (err == ERROR_SUCCESS || IsDirectoryPath(path)) return true;
    if (err != ERROR_ACCESS_DENIED) { /* log + fail */ return false; }
    *elevated = true;             // remembered for the move pass
    return CreateDirShellElevated(owner, path);
}
```

**2. Consider dropping the elevation moniker entirely and letting the shell elevate itself.**

The same reference mod gets an elevated file operation without `CoGetObject`: it creates the normal in-process `CLSID_FileOperation` and sets [`FOFX_SHOWELEVATIONPROMPT`](https://github.com/ramensoftware/windhawk-mods/blob/109589023dde428deaee2fe80e4ce446283a7935/mods/explorer-ctrln-newfile.wh.cpp#L698) (`0x00040000`), which makes the shell copy engine raise the elevation prompt and run the operation elevated itself — the same "You'll need to provide administrator permission… Continue" flow Explorer shows. Since the error UI is already enabled here (no `FOF_NOERRORUI`), adding that one flag to the `SetOperationFlags` call in `RunItemsShell` should cover the move pass, and a `NewItem`-based fallback with the same flag covers the folder that `SHCreateDirectoryExW` couldn't create. That removes `CreateElevatedFileOperation`, `CreateDirElevated`, `IsDirWritable` and the `needElevation` plumbing — roughly 150 lines — and items 3 and 4 below disappear with them.

**3. `needElevation` costs one UAC consent per created folder, not one per action.**

The comment in `DoFiles2Folder` says "both then use the same elevated server, so one user gesture costs at most one consent", but the code creates a fresh elevated object each time: once inside `CreateDirElevated` (called from `EnsureDir`) and once inside `RunItemsShell`. Each `CoGetObject` elevation-moniker activation goes through AppInfo and raises its own consent dialog; nothing is shared between the two calls. In **by name** / **by extension** mode `EnsureDir` runs per item, so a selection landing in five new folders means five consent dialogs plus one for the move. If the moniker approach is kept, the elevated `IFileOperation*` has to be created once per action and passed to both passes.

**4. `CreateDirElevated` can only create one missing level.**

`IFileOperation::NewItem` needs a parent `IShellItem`, and `SHCreateItemFromParsingName` fails for a path that doesn't exist yet. So on the second missing level the parent lookup fails, that `NewItem` is never queued, and the leaf is skipped as well — despite the comment stating that the missing ancestors are queued top-down before the leaf. With a nested fixed name (`archive/2024/incoming`, a documented feature) in a protected location, the user gets a UAC prompt, one stray folder, and then "Could not create destination folder.". Either create the levels one `PerformOperations` at a time, or use the `FOFX_SHOWELEVATIONPROMPT` + `SHCreateDirectoryExW` approach from item 2, which handles the whole chain in one call.

**5. `ParentDir` drops the root separator, so the hotkey path probes the wrong directory.**

`ParentDir(L"C:\\file.txt")` returns `L"C:"` (a drive-relative path), not `L"C:\\"`. That value becomes `folder` for the whole hotkey flow, so `IsDirWritable(L"C:")` probes Explorer's current directory on drive C (usually `C:\Windows`) rather than the drive root — an all-but-guaranteed spurious "requires administrator permission" for files sitting in a drive root, which is the case this PR is about. It's also what `SHChangeNotify` gets at the end. Since the release is fixing root-path handling, worth fixing here too (and in the `folderOut = first.substr(0, slash)` fallback in `ExtractFolderAndSelection`, which has the same issue):

```cpp
static std::wstring ParentDir(const std::wstring& path) {
    size_t slash = path.find_last_of(L"\\/");
    if (slash == std::wstring::npos) return L"";
    if (slash == 2 && path[1] == L':') return path.substr(0, 3);  // keep "C:\"
    return path.substr(0, slash);
}
```

**6. In "Move - fast" the UAC prompt and the folder creation happen before the "cannot elevate" message.**

`EnsureDir` runs for every mode before the `switch (effectiveOp)`, so with `needElevation` the fast path prompts for consent, creates the destination folder elevated, and only then tells the user that "Move - fast" cannot request administrator permission — leaving an empty folder behind that the user can't even delete without elevating again. The check belongs before anything is created (or, better, after `MoveFileExW` actually returns `ERROR_ACCESS_DENIED` — `MoveItemsFast` already captures the first error code, so the message can be raised from there).

<p>

- The new log lines are prefixed with `F2F: `. Windhawk already prefixes every `Wh_Log` line with the mod name, so the prefix is redundant (and it's inconsistent with the existing `Files2Folders: ` lines — I'd drop it from both).
- The README and the `operation` setting description still say "Move - fast … no prompt if a destination needs admin rights (it just fails)", which no longer matches the new message box, and neither mentions that an elevated "Move - safe" is not undoable. The PR description explains both; the README is where users will look.
- `RunItemsShell`'s return value is ignored at both call sites, so the new `queued = 0;` assignments on the failed/aborted paths only affect the log line. Either use the count (e.g. to decide whether to send `SHChangeNotify`) or drop the return value.

</p>
</details>

<p>

- Dropping `FOFX_NOCOPYHOOKS` means third-party copy hooks now participate in every safe move/copy. That does match plain Explorer behavior, but it also means a misbehaving shell extension can now veto or stall the operation — intentional per the comment, just noting the change in blast radius.
- The rationale for `PathJoin` over `PathCombineW` is long-path support, but paths aren't actually long-path-clean end to end: the selection is read through `StrRetToBufW` into `WCHAR[MAX_PATH]` and the folder through `SHGetPathFromIDListW` into `WCHAR[MAX_PATH]`. `PathJoin` is still the right call, just don't count on `>MAX_PATH` working today.
- When the hotkey path is used, `owner` is `nullptr`, so the elevation consent dialog (and the error message boxes) have no owner window and can end up behind other windows. Passing the hotkey thread's hidden message window won't help (it's `HWND_MESSAGE`), but the F2F dialog's own HWND would, if the elevation is triggered from that flow.

</p>
</details>

triatomic:
Thanks for the detailed review — every point was correct. All six required items are addressed, plus the three optional ones. The branch is now a single squashed commit (`f4d6f10d`).

**1. `IsDirWritable` probe — removed.** You were right about the mechanism, and I confirmed it locally before rewriting: `C:\` grants `Authenticated Users` `CreateDirectories` directly with `Modify` only `InheritOnly` to children, so `GetTempFileNameW("C:\\")` returns `ERROR_ACCESS_DENIED` while creating a subfolder there and writing into it both succeed unelevated. The probe reported "needs elevation" for exactly the drive-root case this release exists to fix. It's gone, along with the temp-file churn on every run.

**2. Elevation moniker — dropped.** Adopted the `explorer-ctrln-newfile` approach: plain in-process `CoCreateInstance` plus `FOFX_SHOWELEVATIONPROMPT`, with the direct call tried first and the shell asked to elevate only after a real permission error (`EnsureDir` now takes a `bool* elevated` out-param as in your snippet). This removed `CreateElevatedFileOperation`, `CreateDirElevated`, `IsDirWritable` and all `needElevation` plumbing. My earlier reading that the flag was inert without `FOF_NOERRORUI` was simply wrong — a shipping mod disproves it.

**3 and 4** — resolved by item 2, as you predicted.

**5. `ParentDir`** — applied your fix verbatim, and the same root check in the `ExtractFolderAndSelection` fallback.

**6. "Move - fast" message** — moved into `MoveItemsFast` and keyed on the actual `ERROR_ACCESS_DENIED`, so nothing is created before the user is told.

**Optional items** — all three done: log prefixes dropped from both the new and the pre-existing lines; README and the `operation` setting description now cover the elevation behavior and note that an elevated move isn't undoable; the move/copy count is now used to skip `SHChangeNotify` when nothing changed.

---

Two things from the functionality notes I want to flag rather than quietly leave:

- **`owner` is `nullptr` on the hotkey path** (`RunFiles2Folder(nullptr, ...)`), so an elevation consent or error dialog triggered from that flow can still end up behind other windows. Not fixed here — threading the F2F dialog's own HWND through is a reasonable follow-up, but it touches the hotkey flow rather than the bug this PR is about, and I'd rather not widen the diff late in review. Happy to do it here if you'd prefer.
- **`CreateDirShellElevated` creates one level per `PerformOperations`.** You suggested `SHCreateDirectoryExW` with the flag as the alternative for the whole chain, but that function takes no `IFileOperation` flags and so can't carry `FOFX_SHOWELEVATIONPROMPT`; the per-level loop was the way I could make a nested path work correctly. It does mean a nested fixed name in a protected location can prompt per level. If there's a way to get the whole chain into one elevated call that I've missed, I'd rather use it.

Also noted and agreed: dropping `FOFX_NOCOPYHOOKS` does widen the blast radius to third-party copy hooks (intentional, to match Explorer), and paths aren't long-path-clean end to end given the `MAX_PATH` buffers in the selection/folder reads — `PathJoin` isn't relied on for `>MAX_PATH` today.

The drive-root fix, undo, and elevation into `C:\Program Files` are all confirmed working locally.

m417z:

More details:

triatomic:
Thanks — I verified all four against the code and every one holds up. Two of them are regressions I introduced in this PR. Pushed as `d1508719`, version bumped to **2.5.3** since the reviewed 2.5.2 contained the hang.

**1. Infinite loop in `CreateDirShellElevated` — fixed.** Confirmed: for `"Z:\"` the slash is at index 2, so `cur.substr(0, 3)` returns `cur` unchanged and the loop never terminates while `missing` grows without bound — on Explorer's UI thread for the context-menu path. Applied your `if (parent == cur) break;` guard. The point that `IsDirectoryPath` is false whenever `GetFileAttributesW` fails *for any reason* is what makes this reachable rather than theoretical, since access-denied on a mapped drive is exactly what routes into this function.

**2. "Move - fast" prompting and leaving an empty folder — fixed.** Correct, and it directly contradicted my own PR description. `EnsureDir` escalated before the `switch (effectiveOp)` ever looked at the operation. Elevation is now gated on the operation (`elev.allowed = effectiveOp != F2F_MOVE_FAST`) and the "use Move - safe or Copy" message is raised from the creation failure, before anything is created or prompted for, rather than from `MoveItemsFast` afterwards.

**3. One consent per folder, and cancelling didn't stop the next — fixed, both halves.** Added a run-scoped `ElevationState { allowed, used, refused }`: `refused` latches the first decline so the remaining destinations don't re-prompt. And per your better suggestion, the by-name / by-extension modes now pre-create their destinations in a **single** `IFileOperation` (`CreateDirsShellElevatedBatch`) since they're all flat siblings of one existing parent — 20 files in a protected folder is now one prompt for the whole run. Nested chains still go one level per `PerformOperations`, since `NewItem` needs an existing parent.

**4. Unbounded `wsprintfW` — fixed.** Replaced with a `std::wstring` built by append and passed via `.c_str()`.

**Optional items taken:** the `Wh_ModSettingsChanged` ordering race (now `StopHotkeyThread(); LoadSettings(); StartHotkeyThread();` — a real threading bug worth taking), and the two stale comments on `UniqueDests` and `EnsureDir`. `RestartHotkeyThread` became unused and was removed.

**Optional items not taken,** to keep an already-large diff from growing further into pre-existing code: the `WindhawkUtils` migration, `LOAD_LIBRARY_SEARCH_SYSTEM32` hardening, dropping `<algorithm>` and the unused `-lcomdlg32`/`-loleaut32` flags, and `std::atomic` for the shutdown flag. I did confirm those last two are accurate — there are no comdlg32 or oleaut32 calls and no `std::` algorithm uses — so I'm happy to fold them in if you'd rather they land here than in a follow-up.

From the functionality notes, one I agree with and did **not** fix: `ERROR_ACCESS_DENIED` in `MoveItemsFast` can be source-side, so that message could still misdirect. It now only fires from the destination-creation path, which narrows it considerably, but the source-side case remains. Also still open: the hotkey path passes `owner == nullptr`, so its dialogs are unowned, and cancelling can leave a freshly created destination folder behind.

Not yet recompiled against these changes — I'll confirm before this is merged.

triatomic:
Heads-up on a scope change: I've added a second commit adding a new feature, **Defolder** (`1a854ce1`), and retitled the PR to v2.6. The v2.5.3 bugfix commit you reviewed is unchanged and still the first commit — this was a fast-forward push, not a rewrite, so the review history lines up with what's there.

Defolder is the reverse of the existing operation: a fourth button in the dialog's Operation control that moves the contents of the selected folders out into the folder being browsed, then offers to remove the folders left empty. It's recursive, so it asks for one summary confirmation before doing anything, defaulting to No.

The parts most worth your attention, since the operation deletes directories:

- Removal uses `RemoveDirectoryW` per directory, never a recursive delete. It fails on a non-empty directory by design, so a file that couldn't move keeps its folder alive.
- Reparse points are skipped rather than followed, so a junction inside the selection can't reach outside it or loop.
- Recursion is capped at 64 levels.

If you'd rather review the bugfix on its own, I'm happy to drop the second commit and bring Defolder back as a separate PR once v2.5.3 lands — it's already sitting on its own branch. Just say which you prefer.

One caveat I should state plainly: v2.6 compiles but I haven't finished exercising it at runtime yet, so please don't merge on my account until I confirm. The v2.5.3 changes have been tested (drive-root move, undo, and elevation into a protected location).

triatomic:
All five addressed in `05be2038` (v2.6.1). I verified each against the code first — every one held up, and the three Defolder findings were genuine data-loss bugs.

**1. Junction/symlink roots — fixed.** Confirmed the mechanism: `IsDirectoryPath` is `GetFileAttributesW(...) & FILE_ATTRIBUTE_DIRECTORY`, which is true for a junction since it carries `DIRECTORY | REPARSE_POINT`. So a directly-selected junction was walked through, the target's files moved out from wherever it pointed, and the link itself removed — exactly what the README claims cannot happen. The roots now get the same check as the walk.

**2. Deleting folders when nothing moved — fixed.** Your repro is right: cancel at the conflict dialog and every already-empty subfolder was still gone, past the Recycle Bin. Cleanup is now skipped entirely when `done == 0`.

**3. Silent mode — fixed.** Both confirmations are unconditional now. Your framing was the convincing part: the menu entry carries the shell's localized *"Move to a folder"* label, so a single click silently flattening a tree and deleting the folders is a gap between what the entry says and what it does that no setting should be able to open. I kept the operation available in silent mode rather than refusing it, but it always asks.

**4. Misleading empty-selection feedback — fixed.** Added a folder-skip count and an `ElevationState::denied` flag so "these modes only apply to files" is distinguished from a real permission refusal. Previously Move-fast blamed permissions and the other two said nothing at all.

**5. Window class leak — fixed.** `RegisterClassW`'s result is now checked, the class is unregistered on the `CreateWindowExW` failure path, and it registers with the mod's own module handle via a `GetCurrentModuleHandle()` helper rather than explorer.exe's. Worth noting the class *was* already unregistered on the normal exit path — the leak was only the failure path — but the `hInstance` mismatch was real regardless.

**Optional items taken:** dropped the unused `<windowsx.h>` / `<algorithm>` includes and the `-loleaut32` / `-lcomdlg32` flags (verified nothing uses `SysAlloc*`, `Variant*`, or any common dialog), added the `<cwctype>` include `towupper` was only reaching transitively, and updated the docs for the fourth Operation button, the `@description`, and Defolder's partial undo.

**Not taken,** to keep this from growing further into pre-existing code: the `WindhawkUtils::SetFunctionHook` / `StringSetting` migration, `LOAD_LIBRARY_SEARCH_SYSTEM32` hardening, and `std::atomic` for the shutdown flag. All reasonable; happy to fold them in here if you'd prefer that to a follow-up. The README GIF still shows the three-button control — that needs re-recording, which I can't do as part of the diff.

From the functionality notes, two I agree with and have not fixed: `RunItemsShell` returns the queued count rather than the moved count (so "Moved X of Y" can over-report), and nested selections where both `A` and `A\B` are selected enumerate twice. Both need a real progress sink or root de-duplication respectively; say the word if you want them in this PR.

Testing status, to be straight about it: v2.5.3's changes are confirmed working (drive-root move, undo, elevation into a protected location). v2.6/v2.6.1 compile, but Defolder hasn't been fully exercised at runtime yet — I'll confirm before this is merged.

triatomic:
Picked up three of the optional items in `0d36e34e` (v2.6.2):

- **`WindhawkUtils::SetFunctionHook`** for both hooks, replacing the raw `Wh_SetFunctionHook` with `void*` casts. Worth taking for the compile-time signature check — the casts were hiding exactly the kind of mismatch that only shows up as a crash on the first call.
- **`LOAD_LIBRARY_SEARCH_SYSTEM32`** on the `dwmapi.dll`, `uxtheme.dll` and `shell32.dll` loads (five call sites).
- **`std::atomic<bool>`** for the hotkey shutdown flag.

**`WindhawkUtils::StringSetting` I've deliberately left**, and I'd rather be explicit about why than defer it silently again: it's the one item that isn't mechanical. Converting the seven `Wh_GetStringSetting` sites means changing how `LoadSettings` populates `g_settings`, and the benefit is removing manual free calls and dead null guards from code that works. In a PR that's four rounds deep, with a destructive feature I'm still runtime-testing, that trade doesn't look right to me. Happy to be overruled if you'd rather it land here — it's a contained change, just not a free one.

You were right that `Wh_GetStringSetting` never returns NULL, so the null guards are dead weight regardless; that's worth cleaning up whenever the RAII conversion happens.

Also still open from the earlier functionality notes, both acknowledged and not fixed: `RunItemsShell` returns the queued count rather than the moved count (so "Moved X of Y" can over-report), and nested selections where both `A` and `A\B` are picked enumerate twice. Say the word if you want either in this PR rather than a follow-up.

triatomic:
/ready-for-reviewer

All findings from both AI review rounds are addressed. Summary of where things landed, and what's deliberately left.

**What this PR contains** — four commits on one file:

- `d1508719` **v2.5.3** — the original bug plus the fixes from round one. Drive-root paths produced a doubled separator that `SHCreateItemFromParsingName` rejected, so "Move - safe" and "Copy" silently moved nothing while "Move - fast" worked. Also: `FOFX_ADDUNDORECORD` for Ctrl+Z undo, elevation via `FOFX_SHOWELEVATIONPROMPT` decided once per run, the drive-root infinite loop in the parent walk, `ParentDir` dropping the root separator, and the unbounded `wsprintfW`.
- `1a854ce1` **v2.6** — Defolder, a fourth Operation button that moves folder contents out into the browsed folder and offers to remove the emptied folders.
- `05be2038` **v2.6.1** — round two. The three Defolder data-loss bugs (junction roots being followed and deleted, folders removed after a cancelled move, confirmations suppressed by silent mode), plus the misleading empty-selection message and the window-class leak.
- `0d36e34e` **v2.6.2** — the optional hardening: `WindhawkUtils::SetFunctionHook`, `LOAD_LIBRARY_SEARCH_SYSTEM32`, `std::atomic<bool>`.

**Tested:** the drive-root move and Ctrl+Z undo, elevation into a protected location, and Defolder end to end including the cancel and junction paths.

**Knowingly not addressed**, all from the functionality notes:

- `WindhawkUtils::StringSetting` — the RAII conversion restructures how `LoadSettings` populates `g_settings` across seven call sites, to remove manual free calls and dead null guards from working code. The only optional item that isn't mechanical, so I left it. (The observation that `Wh_GetStringSetting` never returns NULL is correct and worth acting on whenever that happens.)
- `RunItemsShell` returns the queued count, not the moved count, so Defolder's "Moved X of Y" can over-report. A real count needs `IFileOperationProgressSink` or a post-move existence check.
- Nested selections (both `A` and `A\B` selected via the clipboard/hotkey path) enumerate twice, producing a spurious "could not be moved" warning on the second attempt. Needs root de-duplication.
- Enumeration failures in `CollectTreeContents` are silent — the folder correctly survives, but the user isn't told why.
- The README GIF still shows the three-button Operation control; it needs re-recording, which I can't do as part of the diff.

Happy to fold any of these in rather than leave them for a follow-up — just say which.

PR #4933, opened 2026-07-28
Add windows-animations mod v1.1.0

redrag2105:
Introduces a new lightweight window transition suite for Windows.
- Smart Hybrid Engine utilizing GDI and DWM Thumbnail API.
- Physics-based cinematic close effects (Square Shatter, Thanos Snap, Perlin Dissolve, Cyber Glitchy).
- MacOS Genie, Ink Splash minimize and restore animations.
- Soft scale and fade-in animations exclusively for the Alt+Tab menu.

## Changelog

* Changelog item 1...
* Changelog item 2...

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini


redrag2105:
## ✨ Key Features

* **🚀 Smart Hybrid Engine:** Intelligently seamlessly switches between lightning-fast GDI (for pixel-perfect destruction physics) and the native DWM Thumbnail API (to perfectly preserve Windows 11 rounded corners and drop shadows).

* **🎬 Cinematic Close Effects:** Transform how you close applications with four physics-based animations:
  * **Square Shatter:** The window violently explodes outward into digital blocks before drifting into the void.
    
    Square Shatter Preview

  * **Thanos Snap:** A disintegration wave sweeps across the window, turning it into thousands of tiny dust particles that curve away into the wind.
    
    Thanos Snap Preview

  * **Perlin Dissolve:** The window organically melts and dissolves into thin air using a smooth Perlin noise map.
    
    Perlin Dissolve Preview

  * **Cyber Glitch:** The window fractures into irregular horizontal data strips that shudder, then tear left and right like a corrupted hologram.

    Cyber Glitch Preview

* **🧞 Fluid Minimize & Restore:** Choose between the classic Genie "suck into the taskbar" effect (derived from macos-minimize-animation by Abdullah Masood under MIT license) and **Ink Splash** — an organic blot reveal/retract.
  
  Genie Minimize Preview
  
  Ink Splash Preview

* **🔄 Soft Switch Animation:** A pristine scale and fade-in animation triggered *exclusively* when actively switching windows via the Alt+Tab menu.
  * **Before:**
    
    Alt Tab Switch Before
  * **After:**
    
    Alt Tab Switch After
* **🛡️ Rock-Solid Stability:** Features lifecycle management for System Tray apps (like Discord, Steam, etc.)—handling background apps gracefully to minimize ghosting or stuck transparency. Tray apps only *hide* their window instead of closing it, so animating them is opt-in via the **Treat hiding a window as closing it** setting.

Potassiumuncher:
yo keep everything but the macos animation it already exists

m417z:

redrag2105:
Here is a breakdown of the fixes and some context regarding the necessary design trade-offs:

### ✅ Addressed & Fixed

*   **1. Fork/Conflict with `macos-minimize-animation` & 7. Missing toggle for minimize**
    **Fixed:** I've added a toggle in the settings to explicitly enable/disable the Minimize/Restore animations. This addresses the conflict concern: users who prefer the original `macos-minimize-animation` can simply toggle off the minimize feature in this suite, while safely enjoying the standalone Close and Switch effects without any hook interference.
*   **3. Closed window popping back up (Ghost Window)**
    **Fixed:** Modified the teardown logic. The mod now only uncloaks the window if `IsWindowVisible` genuinely returns `false`, preventing unresponsive ghost windows if the app's message loop has already stopped.
*   **4. Unbounded waits hanging the host process on Unload**
    **Fixed:** Replaced the unbounded waits and unsafe `PostThreadMessageW` logic. It now safely retries posting `WM_QUIT` using the thread handle and waits safely, preventing unload deadlocks.
*   **6. Polling threads running regardless of settings**
    **Fixed:** Background threads and hooks are now strictly gated by their respective settings.

---

### ⚖️ Necessary Design Trade-offs

*   **2. Close animation blocking the target app's UI thread**
    **Trade-off:** Removing the `WaitForSingleObject` entirely breaks the capture process because the window destroys its GDI surface before the animation engine can grab a frame. Pausing the app's teardown sequence is a known, necessary trade-off to guarantee the physical destruction effects render correctly in Win32 before the surface disappears.
*   **5. Heavy synchronous work / `PrintWindow` Re-entrancy**
    **Trade-off:** I understand the architectural concern about synchronous `WM_PRINT` during `WM_CLOSE` or `WM_DESTROY`. However, using `BitBlt` exclusively fails to correctly capture DWM-composed, hardware-accelerated, or transparent windows (such as UWP, WinUI 3, and certain Terminal configurations)—often resulting in completely black or broken captures. Calling `PrintWindow(..., PW_RENDERFULLCONTENT)` is currently the only reliable way to grab an accurate visual snapshot of these modern windows right before teardown.

primez-x:
Sweet mod, animations look super polished and very well designed.

redrag2105:
> Sweet mod, animations look super polished and very well designed.

Thanks a lot! I'm glad you like the animations! 🙌

redrag2105:
### 1. Largely a fork of `macos-minimize-animation`?
I want to make a comprehensive animations suite for Windows. Removing the minimize/restore animation or adding this whole suite to macos-**_minimize-animation_** does not make sense. Not to mention that there is also an Ink Splash minimize/restore animation in my mod, not just the genie. Besides, there is an option to toggle off the minimize/restore animations anyway.

### 2. UI Automation dock-X lookup + 120 ms wait
Genie minimize/restore targets the taskbar button. Button X moves when apps open, close, or pin/unpin, so a one-shot cache alone drifts and the suck aims at the wrong place. The async UIA worker plus a short wait is meant to refresh that aim when we can, then fall back to cache/heuristic so the animation still starts.

We may later skip the wait when a fresh cache hit already exists (fire-and-forget refresh), but the current behavior is a deliberate “prefer a better dock X when cheap enough” choice, not an accidental sync UIA call.

### 3. `PrintWindow` on the launch worker + unload join
Launch animation has no prior snapshot. Capture has to happen after the window exists and preferably after first paint; doing that work off the showing path keeps `ShowWindow` from blocking on a long capture. The worker join on unload exists so animation threads are not abandoned mid-frame with GDI/UIA still touching process state (safer than tearing the DLL out under them).

### 4. Alt+Tab polling in Explorer + shared memory
Switch animation must run only for real Alt+Tab switches, not every foreground change (click, Win+number, notification focus, etc.). Explorer is the natural place to observe the gesture; other processes need a cheap “was this an Alt+Tab?” signal. The named section + generation counter is that signal: one writer (Explorer), many readers, no per-process task-switcher UI scraping.

### 5. Close animation on `DestroyWindow` (wide scope)
I will make this a setting in a future version.

With all the issues addressed, I believe the mod is now ready for the final review.
/ready-for-reviewer

m417z:
Thanks for the submission. Since the mod is based on the work of @Abdullah-Masood-05, I'd like to check with him regarding plans to add a similar functionality. If `macos-minimize-animation` won't expand beyond the minimize/restore animation, it's fine to merge this as a separate mod. Otherwise, it might be better to join forces if you both think it's a good idea.

@Abdullah-Masood-05 what do you think?

AyrtonAlbuquerque:
@redrag2105 Great Mod! I would like to suggest to you 2 improvements:

1 - When using the WIN + D to minimize all windows, play the animation on all windows. Right now it just play it in one.

2 - Unhide the taskbar when minimizing a window and offset the the bottom on the animation to match the taskbar height, so that the animation looks like it is going into the icon and not the edge of the screen. Regarding this i would also like to warn you that this unhiding can cause a loop bug in some windows (i think it happens in windows that hold multiple processes, like a browser for example). I reported the same bug to @Abdullah-Masood-05 in his mod, where when the unhide option is enabled the bug occurs.

I have been following this mod and @Abdullah-Masood-05 mod for a while now and they are both fantastic. Right now im favoring this mod more because of 2 things. First the addition of closing animations are very nice (Although i think you should add more discrete ones) and second is a bug im having with @Abdullah-Masood-05 mod where it bugs some borderless fullscreen windows when they are restored and this mod does not.

Good job, i hope it gets merged!

redrag2105:
> @redrag2105 Great Mod! I would like to suggest to you 2 improvements:
> 
> 1 - When using the WIN + D to minimize all windows, play the animation on all windows. Right now it just play it in one.
> 
> 2 - Unhide the taskbar when minimizing a window and offset the the bottom on the animation to match the taskbar height, so that the animation looks like it is going into the icon and not the edge of the screen. Regarding this i would also like to warn you that this unhiding can cause a loop bug in some windows (i think it happens in windows that hold multiple processes, like a browser for example). I reported the same bug to @Abdullah-Masood-05 in his mod, where when the unhide option is enabled the bug occurs.
> 
> I have been following this mod and @Abdullah-Masood-05 mod for a while now and they are both fantastic. Right now im favoring this mod more because of 2 things. First the addition of closing animations are very nice (Although i think you should add more discrete ones) and second is a bug im having with @Abdullah-Masood-05 mod where it bugs some borderless fullscreen windows when they are restored and this mod does not.
> 
> Good job, i hope it gets merged!

I'm glad you like the mod, and thank you for your suggestions!

Regarding your first point, there is actually a known bug where some apps (like Chrome, Discord, etc.) don't trigger the minimize animation when users press WIN+D. That is why it looks like the animation is only playing for one window.

As for the second suggestion, I actually thought about adding this but eventually let it go. The problem is that when minimizing an app, users would have to wait for the taskbar to gradually unhide before the animation could even start. This causes an annoying delay that disrupts the workflow. I might still add it as an option in the settings in a future release, though.

Finally, thanks for the feedback on the close animations! I will definitely consider adding some simpler, more discreet options in future updates

Abdullah-Masood-05:
Thanks for the tag @m417z. macos-minimize-animation isn't expanding beyond minimize/restore/launch, so I don't see a conflict here. Fine by me if @redrag2105's mod merges as its own thing. Thanks for crediting the genie engine properly. Nice work on the close effects.

redrag2105:
> Thanks for the tag @m417z. macos-minimize-animation isn't expanding beyond minimize/restore/launch, so I don't see a conflict here. Fine by me if @redrag2105's mod merges as its own thing. Thanks for crediting the genie engine properly. Nice work on the close effects.

Thank you for the kind words and the support! Your work on the genie engine is brilliant and inspired me a lot while building this suite. I really appreciate you letting this merge independently. Cheers!

PR #4941, opened 2026-07-29
Add Taskbar Blob Shape mod

Deen-0x:
## Taskbar Blob Shape

Add Taskbar Blob Shape mod which replaces the rounded-rectangle indicator behind Windows 11 taskbar buttons with a parametric "blob" — a tab-like shape whose top edge is flat and wide, with concave flares at the top corners so each button reads as a tab merging into the desktop above it.

## Attribution
- Originally based on [Taskbar Elastic WinUI Pill](https://github.com/ramensoftware/windhawk-mods/pull/4685) & the yet released **Taskbar Elastic Border** by @Lockframe 

## Use case
- See Taskbar Styling [PR #617](https://github.com/ramensoftware/windows-11-taskbar-styling-guide/pull/617).

blob
lhSj7xL9B2

## Features

- **Parametric shape** — top corner radius (the concave flares; also sets how far the shape extends up and outward), bottom corner radius, custom width/height (`auto` matches the button's background element), and margin offsets.
- **Color** — custom hex fill with multi-color gradient support and separate `light | dark` values, per-theme opacity multipliers, and automatic fallback to the system accent color when no custom color is set.
- Works across multiple monitors and taskbars.

## Implementation notes

- Hooks `TaskListButton::UpdateVisualStates` (Taskbar.View.dll) to observe running indicator state changes; activation is a pure per-button opacity toggle.
- Each button gets its own `Path` element, hosted in the taskbar's `RootGrid` (above the task list's clipping region so the flare tips render fully) and glued to its button with a one-time composition `ExpressionAnimation` on the button's visual offset chain — the shape tracks reordering, reflow, and taskbar animations on the render thread with no per-frame UI-thread work.
- Full button lifecycle tracking: an `Unloaded` handler per button hides its shape and invalidates the binding when the button leaves the tree (window moved to another monitor, app closed, container recycled), and reused containers re-resolve their hosting grid, re-parent the shape if needed, and rebind — shapes never outlive or detach from their buttons.
- The native `BackgroundElement` is suppressed via its hand-off visual's `IsVisible` flag rather than a local `Opacity` value, so visual-state transition storyboards (which outrank local values in XAML property precedence) can't bring it back on top of the shape.
- Late layout is handled with a `SizeChanged` subscription per button — no timers, no polling.
- Unloading detaches all event handlers, stops the composition animations, removes the injected elements, and restores the native indicators.

## Tested on

- Windows 11 25H2 (OS Build 26200.8894), single and dual monitor
- Reordering pinned items, launching pinned apps, minimize/restore, dragging windows between monitors in both directions (including button removal and container reuse), overflow-free and mixed light/dark themes, enable/disable cycles

## Mod authorship

This mod was created by: Claude

### DIFF
--- a/mods/taskbar-blob-shape.wh.cpp
+++ b/mods/taskbar-blob-shape.wh.cpp
@@ -16,7 +16,7 @@
 
 Adds a blob shape behind active taskbar buttons.
 
-![Taskbar Blob Shape](https://i.imgur.com/REPLACE-ME.png)
+![Taskbar Blob Shape](https://raw.githubusercontent.com/Deen-0x/windhawk-assets/refs/heads/main/taskbar-blob-shape/demo.gif)
 
 Based on **Taskbar Elastic WinUI Pill** and the unreleased **Taskbar Elastic
 Border** by **Lockframe**.
--- a/mods/taskbar-blob-shape.wh.cpp
+++ b/mods/taskbar-blob-shape.wh.cpp
@@ -1,0 +1,990 @@
+// ==WindhawkMod==
+// @id              taskbar-blob-shape
+// @name            Taskbar Blob Shape
+// @description     Injects a customizable blob shape behind active taskbar items.
+// @version         0.2.0
+// @author          Deen-0x
+// @github          https://github.com/Deen-0x
+// @include         explorer.exe
+// @architecture    x86-64
+// @compilerOptions -lole32 -loleaut32 -lruntimeobject
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Taskbar Blob Shape
+
+Adds a blob shape behind active taskbar buttons.
+
+Every taskbar button gets its own blob shape, shown or hidden as the
+button's running indicator state changes. The shapes are hosted in the
+taskbar's RootGrid (above the task list's clipping region, so the flares
+render fully) and each one is glued to its button with a composition
+expression, so it follows the button through reordering, reflow, and
+animations on the render thread.
+
+The blob shape is a rounded rectangle whose top extends upward and flares
+outward with concave (outside) corner radii, ending in a flat top edge:
+
+- **Width / Height**: the main area of the blob shape ('auto' matches the
+  button's background element). The blob shape is anchored so this area is
+  centered on it.
+- **Top corner radius**: the radius of the concave top flares. The flares
+  are circular quarter arcs, so this also sets how far the blob shape
+  extends upward and outward. Total size is
+  (Width + 2*TopRadius) x (Height + TopRadius).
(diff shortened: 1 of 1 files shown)


Deen-0x:
Regarding Point 2:
Fair question - this did begin as a fork of that lineage, and the settings vocabulary is deliberately inherited. But the shapes are topologically different, and that forced a different implementation.

The elastic pill is a convex rounded rectangle sliding behind the active button. The blob is a tab silhouette with concave top corners and a flat top edge anchored flush to the taskbar's top — it reads as the button merging into the taskbar surface, not as an indicator behind it. No parameter values turn one into the other: a Border's CornerRadius can't produce concave corners, and the flares extend beyond the button's bounds.

That geometry dictates the architecture: a Path with programmatic geometry instead of a styled Border; hosted in RootGrid above the task list's clip region (button-hosted shapes get their flares clipped — verified); one static shape per button toggled by indicator state instead of one shared shape animating between them, each glued to its button with an X-only composition expression while Y is structurally anchored to the taskbar top. Merging this into the pill wouldn't be a shape enum — it's a second rendering element, hosting model, and lifecycle model behind a mode switch, sharing little beyond the hook.


Regarding functionality notes:

1. Windows already has press icon sizing feedback, there's no need for hover feedback here imo.

2. That's intended, a blob isn't designed to be displayed in overflow, it would look weird and unintentional.

3. It's good and tbh intended that the blob behaves the same way.

4. After extensive testing, this solution proved to be more consistent and bug free (no stuttering and animation laggs)

5. Intended solution to combat y axis stuttering.

Deen-0x:
Regarding the relationship to Taskbar Elastic WinUI Pill. This began as a fork of that lineage, which is why the hook, helpers, and settings vocabulary are shared — kept deliberately so users find familiar knobs. The mods diverge at the geometry, and the geometry forced a different architecture. The pill is a convex rounded Border that slides between buttons; the blob is a tab silhouette with concave top flares and a flat top edge anchored to the taskbar's top — no CornerRadius values produce it, and the flares extend beyond the button's bounds. That requires a Path with programmatic geometry, hosting in RootGrid above the task list's clip region (button-hosted shapes get their flares clipped — verified), one static per-button shape with full lifecycle tracking instead of one animated shared element, and no animation surface at all. Merging it into the pill would mean a second rendering element, hosting model, positioning model, and lifecycle model behind a mode switch — a maintenance burden on Lockframe's mod larger than a standalone mod that shares its vocabulary. That said, I defer to the maintainer and to Lockframe: if a Shape: pill | blob option upstream is preferred, I'll contribute this there instead. The mods should not be enabled together (both suppress the native indicator), which I've noted in the README — happy to also add a mutual note to the pill's README if Lockframe agrees.

PR #4945, opened 2026-07-30
Add Command on Window focus mod

rom4ster:
## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini

This mod was not vibe coded. AI assistance was used to assist with, not replace development.

### DIFF
--- a/mods/command-on-window-focus.wh.cpp
+++ b/mods/command-on-window-focus.wh.cpp
@@ -1,0 +1,263 @@
+// ==WindhawkMod==
+// @id              command-on-window-focus
+// @name            Run command on window focus
+// @description     Run a command on window focus matching a filter and optionally a command when not matching
+// @version         0.1
+// @author          rom4ster
+// @github          https://github.com/rom4ster
+// @include         explorer.exe
+// @compilerOptions -lcomdlg32
+// @license         MIT
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Your Awesome Mod
+This is a place for useful information about your mod. Use it to describe the
+mod, explain why it's useful, and add any other relevant details. You can use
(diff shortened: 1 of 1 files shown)


m417z:

rom4ster:
Alright at this point it seems pretty ok. I am not going to implement any state tracking in this mod despite the AI's suggestion. Stateful systems are more prone to error than stateless ones, even if that means a couple of extra threads get spawned. You can in many situations update a command or script to better handle at that level if need be. 

I do think future state for a mod like this will be something more akin to list of [ {filter list, command, negate} ] for settings allowing for more flexible/granular control but as a version 1 I think the current functionality is fine.

rom4ster:
See comment above on why I wont do state changes. 

I am not too concerned on integrity processes or the silent fail, especially since admin processes seem to work just fine from my testing. 

I also dont mind the 10 second stall but lmk if it really is too much.

m417z:
Please address this, there's no reason for these defines which make overall maintenance more difficult:

> **`#define Init WhTool_ModInit`** (and `SettingsChanged` / `Uninit`) puts three very generic identifiers into the preprocessor for the rest of the translation unit. It happens to be harmless here, but it will bite the moment a header or a future edit uses one of those names. Just name the functions `WhTool_ModInit` / `WhTool_ModSettingsChanged` / `WhTool_ModUninit` directly, as in the wiki snippet.

There are other review notes that I think are worth doing, such as "Compile the regexes once, at load time", but I'll leave it to you to decide.

rom4ster:
If regex matching errors out, let it bubble, let it throw, that type of unexpected error is exactly what I think should happen. 
I say we keep the child windows doing checks and see what happens. I may end up adding a window title filter anyways. 
Default settings are fine. If someone enables this mod, they probably should see the settings.

PR #4947, opened 2026-07-30
Add instant-taskbar-thumbnail-previews.wh.cpp

alchemyyy:
## Changelog

## Mod authorship

This mod removes / controls the delays around "taskbar thumbnail previews" i.e. these:

image

It also contains two optional features:

* A toggle which instantly closes open thumbnail previews when clicking outside the taskbar and thumbnail surface
* A toggle which instantly closes open thumbnail previews when hovering over the Start button

Implementation notes and known limitations

- **Stable symbol set:** All optional hooks are resolved unconditionally so toggling settings does not rewrite Windhawk’s symbol cache, trigger PDB re-resolution, or require a mod reload. Disabled features return through inexpensive runtime guards.

- **Light-dismiss ownership:** Native `RegisterLightDismiss` and `UnregisterLightDismiss` calls are tracked per `TaskbarHost`. Calls made by the mod use the original trampolines, allowing it to avoid unregistering an action Windows subsequently depends on.

- **Preexisting registration limitation:** Registrations created before the hooks are installed cannot be observed. Solving that completely would require inspecting a private `TaskbarHost` member through hardcoded offsets or runtime instruction decoding, which would be pretty fragile across Windows builds.

- **Cleanup threading:** This mod briefly installs a thread-specific `WH_CALLWNDPROC` hook during settings cleanup or unload solely to execute cleanup on the taskbar UI thread, then immediately removes it.

- **Deferred rearming:** When Windows fires light-dismiss, it first unregisters the current action. If the pointer remains over the thumbnail flyout, the mod must register a replacement so future outside clicks remain detectable. `SetTimer` queues this rearming on the taskbar thread so it normally runs after the current pointer dispatch finishes. A nested modal message loop could theoretically run the timer earlier, but that is not expected in this pointer-input path. This is a simpler alternative to hooking `DispatchMessageW` and tracking dispatch depth process-wide.

- **Private implementation dependency:** This mod hooks undocumented Taskbar and Taskbar.View symbols and retrieves `TaskbarHost` through internal task-band interfaces. Windows updates can break these paths; optional features fail closed when their symbols are unavailable.

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/instant-taskbar-thumbnail-previews.wh.cpp
+++ b/mods/instant-taskbar-thumbnail-previews.wh.cpp
@@ -1,0 +1,1318 @@
+// ==WindhawkMod==
+// @id              instant-taskbar-thumbnail-previews
+// @name            Instant Taskbar Thumbnail Previews
+// @description     Controls taskbar thumbnail preview show and close behavior.
+// @version         1.0.0
+// @author          Alchemy
+// @github          https://github.com/alchemyyy
+// @license         MIT
+// @include         explorer.exe
+// @architecture    x86-64
+// @compilerOptions -lole32 -loleaut32 -lruntimeobject
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Instant Taskbar Thumbnail Previews
+
+Adds the ability to configure or remove the delays before Windows 11 shows and closes taskbar thumbnail previews,
+as well as the ability to configure mouse actions that close them.
+
+Both delay settings replace the corresponding Windows value, so they can make the transitions either shorter or longer than the stock delays.
+
+The optional outside-click behavior arms the taskbar's native light-dismiss action while a thumbnail flyout is open.
+
+The optional Start-button hover behavior closes an open thumbnail flyout directly from the Start button's XAML pointer-enter handler.
+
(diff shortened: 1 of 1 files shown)


alchemyyy:
Issue 1 seems to be a false positive.

In `Taskbar.View.dll` - Windows’ own generated `ExperienceToggleButton::OnPointerEntered` code:

```asm
movq 0x8(%rcx), %rcx
```

It reads the pointer at `this + 8`, equivalent to:

```cpp
reinterpret_cast<IUnknown**>(object)[1]
```

It then calls `QueryInterface` through that pointer. Another `ExperienceToggleButton` helper, `TransformToVisual`, uses the same field for XAML interface queries. Therefore, slot 1 is the composable inner XAML object on this build, not a static vtable.

The `+2` and `+3` approaches found in other mods refer to different embedded C++/WinRT interface subobjects. They can also reach `FrameworkElement` through `QueryInterface`, but their existence does not make the current `+8` access incorrect.

The bot is correct that `catch (...)` does not catch access violations under Clang. However, that observation does not establish an invalid access here; the decompiled code and native disassembly validate the layout being used. All private class offsets remain inherently Windows-version-dependent, including the bot’s suggested replacement.

PR #4948, opened 2026-07-30
Rewrite taskbar-start-button-corner-fix.wh.cpp

alchemyyy:
## Changelog

Reworked corner-click handling to hook the taskbar’s invisible InputSite window procedure and intercept its `WM_POINTER` messages directly. Existing InputSite windows are found through taskbar scans, while `CreateWindowInBand` covers future taskbar recreation. Runtime hook operations are applied only when necessary, with retry and unload protection.

Improved corner detection by validating the taskbar bounds, monitor ownership, adjacent-monitor geometry, DPI-scaled L-shaped region, pointer lifecycle, and left-aligned taskbar setting. The registry check now runs only after the cheaper geometry checks pass.

UI Automation is now limited to Start-button activation on a dedicated worker thread:

- Uses `CUIAutomation8`, COM RAII, provider timeouts, and per-taskbar element caching.
- Coalesces clicks received during UIA lag to avoid repeatedly toggling Start open and closed.
- Adds the `openStartWhenButtonUnavailable` setting, enabled by default. When UIA successfully searches the taskbar but finds no Start button, the mod falls back to pressing the Windows key.
- Does not trigger the fallback for COM errors, UIA provider failures, activation failures, stale taskbars, or worker-queue failures.

The fallback applies to both primary and secondary taskbars. Since Windows normally opens a keyboard-triggered Start menu on the primary monitor, the documentation recommends the [Start menu open location](https://windhawk.net/mods/start-menu-open-location) mod for cursor-monitor placement.

The UIA bounding rectangle is not queried before interception because that would require provider calls on the taskbar input thread or depend on potentially stale cached geometry.

Also updated the Windhawk requirement and settings descriptions, and removed redundant linker options.

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/taskbar-start-button-corner-fix.wh.cpp
+++ b/mods/taskbar-start-button-corner-fix.wh.cpp
@@ -2,40 +2,34 @@
 // @id              taskbar-start-button-corner-fix
 // @name            Start Menu Corner Click Fix
 // @description     Fixes the issue where clicking in the corner of the taskbar doesn't open the Start menu on multi monitor setups.
-// @version         1.1
+// @version         2.0
 // @author          Alchemy
 // @github          https://github.com/alchemyyy
 // @license         MIT
 // @include         explorer.exe
 // @architecture    x86-64
-// @compilerOptions -lole32 -loleaut32 -luser32
+// @compilerOptions -lole32 -loleaut32
 // ==/WindhawkMod==
 
 // ==WindhawkModReadme==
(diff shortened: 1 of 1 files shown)


alchemyyy:
The corner is already an explicit deadzone in Windows., hence the click consumption there.

Issue #1504, opened 2025-02-09
[Windows 11 Taskbar Styler] Memory leak while using Task View


AromaKitsune:
Hi @m417z , I found a bug in your "Windows 11 Taskbar Styler" mod.

While the "Windows 11 Taskbar Styler" mod is enabled, pressing Win+Tab to bring up the Task View several times causes Windows Explorer's memory usage to increase more.

I usually multi-task with Alt+Tab, Task View, and Virtual Desktops. The more I use Task View the slower the Task View becomes, and pressing Win+Tab would take several seconds for the Task View to show up. Windows Explorer would eventually crash if the memory usage goes up any further.

On my PC, Windows Explorer starts with ~140 MB (after restarting `explorer.exe`).
![Image](https://github.com/user-attachments/assets/03bf1505-e01c-4d3b-a158-114287a5016a)

After using Task View at least 20 times (with mod disabled), it stays at ~170 MB, this is expected.
![Image](https://github.com/user-attachments/assets/d0cac49a-0567-4c74-a561-a339518317e0)

After using Task View at least 20 times (with mod enabled), it goes up to ~320 MB, this is where the memory leak happens.
![Image](https://github.com/user-attachments/assets/7437041d-240c-4cb7-9e41-78048b2bfb00)

It keeps increasing when using Task View 30 more times, resulting in an increased delay of the Win+Tab hotkey, and if the memory usage goes up any further, Windows Explorer crashes.
![Image](https://github.com/user-attachments/assets/55e2dbfa-725e-49f7-86d7-8811c6fc9221)

m417z:
Hi, I can reproduce it. The mod uses the Windows XAML Diagnostics API to receive events about UI elements that are added and removed. I tried to change the mod to subscribe to the events and do absolutely nothing else, and I could still observe the memory leak. So it seems that the memory leak is an issue with the XAML Diagnostics API, and there's not much the mod can do about it.

The XAML Diagnostics API is designed for debugging, and it has several other issues. For example, it was observed to cause crashes in some cases. Since the mod is designed around that API, it comes with these issues. Hopefully Microsoft will improve it in the future, although I wouldn't hold my breath for it.

AromaKitsune:
That's unfortunate... I really like this mod, especially the Translucent Taskbar theme. Thanks for taking your time troubleshooting this bug.

JusticeWay:
Has the problem been fixed?

m417z:
[...]

Issue #354, opened 2023-11-01
[Windows 11 Taskbar Styler] Cause Explorer crash when switch virtual desktop


clover-yan:
**Windows 11 Taskbar Styler by @m417z**

When switch virtual desktop, Explorer may randomly crash and restart.

I use Windows 11 22H2 `22621.2134`, with StartAllBack using. Disabling it and all other mods didn't help.

<details>
<summary>My mod settings</summary>

```json
{"resourceVariables[0].variableKey":"TaskbarContextMenuMargin","resourceVariables[0].value":"0","promptForExplorerRestart":"1","controlStyles[0].target":"taskbar:TaskListButtonPanel > Border#BackgroundElement","controlStyles[0].styles[0]":"Background=black"}
```
</details>
<details>
<summary>Log from Event Viewer</summary>

```xml
日志名称:          Application
来源:            Application Error
日期:            2023/11/1 16:57:08
事件 ID:         1000
任务类别:          应用程序崩溃事件
级别:            错误
关键字:           
用户:            DESKTOP-0U8LNJD\seewo
计算机:           DESKTOP-0U8LNJD
描述:
错误应用程序名称: explorer.exe，版本: 10.0.22621.2070，时间戳: 0x2bd7252c
错误模块名称: Windows.UI.Xaml.dll，版本: 10.0.22621.2070，时间戳: 0xa1d6b8fe
异常代码: 0xc0000005
错误偏移量: 0x0000000000343080
错误进程 ID: 0x0x39D0
错误应用程序启动时间: 0x0x1DA0C9FE7548377
错误应用程序路径: C:\WINDOWS\explorer.exe
错误模块路径: C:\Windows\System32\Windows.UI.Xaml.dll
报告 ID: b92d8a24-9a70-4315-8d52-c153516967cf
错误程序包全名: 
错误程序包相对应用程序 ID: 
事件 Xml:
<Event xmlns="http://schemas.microsoft.com/win/2004/08/events/event">
  <System>
    <Provider Name="Application Error" Guid="{a0e9b465-b939-57d7-b27d-95d8e925ff57}" />
    <EventID>1000</EventID>
    <Version>0</Version>
    <Level>2</Level>
    <Task>100</Task>
    <Opcode>0</Opcode>
    <Keywords>0x8000000000000000</Keywords>
    <TimeCreated SystemTime="2023-11-01T08:57:08.7838832Z" />
    <EventRecordID>11732</EventRecordID>
    <Correlation />
[...]

PR #4950, opened 2026-07-30
Add CJK Spacer mod

aenerv7:
Adds the CJK Spacer mod, which inserts spaces between CJK characters and Unicode letters or digits in context-menu and tooltip text hosted by `explorer.exe`; existing whitespace remains unchanged.

Classic Win32 menu handling is limited to the lifetime of `TrackPopupMenu` / `TrackPopupMenuEx`. Existing items are rewritten immediately before display, dynamically added or changed items are handled only while the popup is active, and recorded strings are restored as soon as the popup closes. Restoration treats the recorded index as a hint and falls back to locating the spaced text if insertions shifted menu indices. The insertion hooks also cover append operations and `MF_POPUP | MF_BYCOMMAND` forms.

Classic tooltips are scoped to uxtheme handles opened for the `TOOLTIP` class. `GetThemeTextExtent` handles measurement, while `DrawThemeText` and `DrawThemeTextEx` handle drawing. The hot path first rejects untracked theme handles; when the HDC maps to a window, that window must also be `tooltips_class32`, which protects against a stale or reused theme handle affecting another control. Existing tooltip windows are discovered when the mod is enabled in a running Explorer. Unthemed tooltips drawn directly with `DrawTextW` remain outside this path.

The opt-in `modernUiText` path uses XAML Diagnostics for both Windows.UI.Xaml and Microsoft.UI.Xaml. It changes source properties on standard `MenuFlyoutItem` / `MenuFlyoutSubItem` controls and `ToolTip.Content`, rather than assigning presenter `TextBlock.Text`. Only values that `ReadLocalValue` identifies as plain local strings are eligible; bindings, styles, expressions, unrelated XAML text, and taskbar thumbnail previews remain untouched.

Modern sources are now processed directly from their XAML visual-tree lifetime: a supported source is changed on its Add mutation and restored on Remove or mod unload. There is no popup-HWND ownership guess, pending set, accessibility event hook, or sticky cached-popup mapping. State remains per UI thread, and unload restoration uses the standard blocking `SendMessage` dispatch so its stack parameter cannot outlive the call. The thread registry stores each thread's creation-time fingerprint and explicitly prunes dead or reused IDs during registration and unload snapshotting, without putting a mutex-taking unregister operation in a thread-local destructor.

XAML Diagnostics connection requests come from `Wh_ModAfterInit` and a hook on the real `kernelbase.dll!LoadLibraryExW` implementation, covering both already-loaded and late-loaded Windows.UI.Xaml/Microsoft.UI.Xaml modules. The loader hook only schedules a single-flight worker; requests arriving while it runs remain pending and are consumed before the worker exits. WUX and MUX connection state is tracked independently, and `SetSite(nullptr)` clears and re-requests only the connection that was lost. COM activation and TAP loading happen outside the loader path. Each worker pins the mod image until `FreeLibraryAndExitThread`, so unload never joins a possibly blocked connection thread and the worker never executes unloaded code. The visual-tree advise thread also pins the image for its own lifetime. Reentrancy protection lives inside `EnsureModernXamlDiagnostics`, covering every caller, and the 10,000-name probe observes the stop flag. A connection is marked successful only when `SetSite` actually adds a watcher, so a diagnostics-handling mod that returns `S_OK` while blocking the call cannot create a false-positive connection. Watcher replacement and `SetSite(nullptr)` explicitly unadvise and release the old watcher, including an advise/unadvise race guard. The setting and README name Windows 11 Taskbar Styler and Windows 11 File Explorer Styler as potential exclusive-connection conflicts.

Hook groups are installed only for enabled features. Settings changes request a full Windhawk reload so disabling a feature removes its hooks and tracked state. The mod is injected only into `explorer.exe`; Start, Search, and shell flyouts hosted by other processes are explicitly documented as out of scope.

The implementation was written by ChatGPT from requirements provided by @aenerv7, then compiled and tested by @aenerv7 in Windhawk.

## Changelog

Not applicable (new mod).

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [x] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/cjk-spacer.wh.cpp
+++ b/mods/cjk-spacer.wh.cpp
@@ -2,7 +2,7 @@
 // @id              cjk-spacer
 // @name            CJK Spacer
 // @description     Add spaces between CJK characters and letters or digits in Explorer context menus and tooltips
-// @version         0.1.10
+// @version         0.1.11
 // @author          aenerv7
 // @github          https://github.com/aenerv7
 // @license         GPL-3.0
@@ -130,11 +130,9 @@ Rewritten strings are recorded and restored when the popup closes.
 
 #include <winrt/base.h>
 #include <winrt/Microsoft.UI.Xaml.Controls.h>
-#include <winrt/Microsoft.UI.Xaml.Media.h>
 #include <winrt/Microsoft.UI.Xaml.h>
 #include <winrt/Windows.Foundation.h>
 #include <winrt/Windows.UI.Xaml.Controls.h>
-#include <winrt/Windows.UI.Xaml.Media.h>
 #include <winrt/Windows.UI.Xaml.h>
 
 namespace {
@@ -1041,7 +1039,9 @@ HTHEME WINAPI OpenThemeDataForDpiHook(HWND window,
 }
 
 HRESULT WINAPI CloseThemeDataHook(HTHEME theme) {
-    UntrackClassicTooltipTheme(theme);
+    if (g_classicTooltipThemeCount.load(std::memory_order_relaxed) != 0) {
+        UntrackClassicTooltipTheme(theme);
+    }
     return g_originalCloseThemeData(theme);
 }
 
@@ -1340,14 +1340,16 @@ struct TooltipContentAccess {
             return false;
         }
 
-        try {
-            const auto value =
-                winrt::unbox_value<winrt::hstring>(content);
-            text->assign(value.c_str(), value.size());
-            return true;
-        } catch (const winrt::hresult_error&) {
+        const auto propertyValue =
+            content.template try_as<wf::IPropertyValue>();
+        if (!propertyValue ||
+            propertyValue.Type() != wf::PropertyType::String) {
             return false;
         }
+
+        const auto value = propertyValue.GetString();
+        text->assign(value.c_str(), value.size());
+        return true;
     }
 
     template <typename Element>
@@ -1384,7 +1386,7 @@ public:
 
         try {
             auto element = m_element.get();
-            if (!element || !element.IsLoaded()) {
+            if (!element) {
                 return;
             }
 
@@ -1505,13 +1507,11 @@ using ModernTextStateMap =
                        std::shared_ptr<ModernTextStateBase>,
                        ModernElementKeyHash>;
 
-[[clang::no_destroy]] thread_local
-    std::optional<ModernTextStateMap>
-        g_modernTextStatesForThread{std::in_place};
+thread_local std::optional<ModernTextStateMap>
+    g_modernTextStatesForThread{std::in_place};
 
 std::mutex g_modernTextStateThreadsMutex;
-[[clang::no_destroy]] std::unordered_set<DWORD>
-    g_modernTextStateThreads;
+std::unordered_set<DWORD> g_modernTextStateThreads;
 
 void RegisterModernTextStateThread() {
     std::lock_guard<std::mutex> guard(
@@ -1753,14 +1753,17 @@ class WindhawkTap
 public:
     HRESULT STDMETHODCALLTYPE SetSite(IUnknown* site) override try {
         m_site.copy_from(site);
-        if (!site || g_stoppingModernUi.load(
-                         std::memory_order_relaxed)) {
+        if (!site) {
             return S_OK;
         }
 
         // Balance the reference added when XAML Diagnostics loads this module.
         if (HMODULE module = GetCurrentModuleHandle()) {
             FreeLibrary(module);
+        }
+
+        if (g_stoppingModernUi.load(std::memory_order_relaxed)) {
+            return S_OK;
         }
 
         auto watcher =
@@ -1855,6 +1858,7 @@ using InitializeXamlDiagnosticsEx_t =
     decltype(&InitializeXamlDiagnosticsEx);
 
 std::mutex g_xamlDiagnosticsInitializationMutex;
+std::mutex g_xamlDiagnosticsWorkerMutex;
 bool g_windowsUiXamlDiagnosticsConnected;
 bool g_microsoftUiXamlDiagnosticsConnected;
 HANDLE g_xamlDiagnosticsWakeEvent;
@@ -1942,11 +1946,12 @@ void EnsureModernXamlDiagnostics() {
     }
 }
 
-DWORD WINAPI XamlDiagnosticsWorkerProc(LPVOID) {
+DWORD WINAPI XamlDiagnosticsWorkerProc(LPVOID parameter) {
+    const HANDLE wakeEvent = static_cast<HANDLE>(parameter);
     const HRESULT initializeResult =
         CoInitializeEx(nullptr, COINIT_MULTITHREADED);
-    while (WaitForSingleObject(
-               g_xamlDiagnosticsWakeEvent, INFINITE) == WAIT_OBJECT_0) {
+    while (WaitForSingleObject(wakeEvent, INFINITE) ==
+           WAIT_OBJECT_0) {
         if (g_stopXamlDiagnosticsWorker.load(
                 std::memory_order_relaxed)) {
             break;
@@ -1963,24 +1968,31 @@ DWORD WINAPI XamlDiagnosticsWorkerProc(LPVOID) {
 bool StartXamlDiagnosticsWorker() {
     g_stopXamlDiagnosticsWorker.store(
         false, std::memory_order_relaxed);
-    g_xamlDiagnosticsWakeEvent =
+    const HANDLE wakeEvent =
         CreateEventW(nullptr, FALSE, FALSE, nullptr);
-    if (!g_xamlDiagnosticsWakeEvent) {
+    if (!wakeEvent) {
         return false;
     }
 
-    g_xamlDiagnosticsWorkerThread = CreateThread(
-        nullptr, 0, XamlDiagnosticsWorkerProc, nullptr, 0, nullptr);
-    if (!g_xamlDiagnosticsWorkerThread) {
-        CloseHandle(g_xamlDiagnosticsWakeEvent);
-        g_xamlDiagnosticsWakeEvent = nullptr;
+    const HANDLE workerThread = CreateThread(
+        nullptr, 0, XamlDiagnosticsWorkerProc, wakeEvent, 0, nullptr);
+    if (!workerThread) {
+        CloseHandle(wakeEvent);
         return false;
     }
 
+    {
+        std::lock_guard<std::mutex> guard(
+            g_xamlDiagnosticsWorkerMutex);
+        g_xamlDiagnosticsWakeEvent = wakeEvent;
+        g_xamlDiagnosticsWorkerThread = workerThread;
+    }
     return true;
 }
 
 void RequestXamlDiagnosticsInitialization() {
+    std::lock_guard<std::mutex> guard(
+        g_xamlDiagnosticsWorkerMutex);
     if (!g_xamlDiagnosticsWakeEvent ||
         g_stoppingModernUi.load(std::memory_order_relaxed)) {
         return;
@@ -2000,18 +2012,27 @@ void RequestXamlDiagnosticsInitialization() {
 }
 
 void StopXamlDiagnosticsWorker() {
-    if (!g_xamlDiagnosticsWorkerThread) {
-        return;
+    HANDLE workerThread;
+    HANDLE wakeEvent;
+    {
+        std::lock_guard<std::mutex> guard(
+            g_xamlDiagnosticsWorkerMutex);
+        workerThread = g_xamlDiagnosticsWorkerThread;
+        wakeEvent = g_xamlDiagnosticsWakeEvent;
+        if (!workerThread) {
+            return;
+        }
+
+        g_xamlDiagnosticsWorkerThread = nullptr;
+        g_xamlDiagnosticsWakeEvent = nullptr;
+        g_stopXamlDiagnosticsWorker.store(
+            true, std::memory_order_relaxed);
+        SetEvent(wakeEvent);
     }
 
-    g_stopXamlDiagnosticsWorker.store(
-        true, std::memory_order_relaxed);
-    SetEvent(g_xamlDiagnosticsWakeEvent);
-    WaitForSingleObject(g_xamlDiagnosticsWorkerThread, INFINITE);
-    CloseHandle(g_xamlDiagnosticsWorkerThread);
-    CloseHandle(g_xamlDiagnosticsWakeEvent);
-    g_xamlDiagnosticsWorkerThread = nullptr;
-    g_xamlDiagnosticsWakeEvent = nullptr;
+    WaitForSingleObject(workerThread, INFINITE);
+    CloseHandle(workerThread);
+    CloseHandle(wakeEvent);
 }
 
 void CALLBACK ModernPopupWinEventProc(
@@ -2065,7 +2086,7 @@ bool RunFromWindowThread(HWND window,
                          RunFromWindowThreadProc_t procedure,
                          PVOID parameter) {
     static const UINT message = RegisterWindowMessageW(
-        L"Windhawk_RunFromWindowThread_CJKSpacer");
+        L"Windhawk_RunFromWindowThread_" WH_MOD_ID);
 
     struct RunParameter {
         RunFromWindowThreadProc_t procedure;
@@ -2193,9 +2214,8 @@ void UninitializeModernUi() {
                     CleanupModernTextStatesForCurrentThread();
                 },
                 nullptr)) {
-            // The thread-local state is deliberately no_destroy. If its UI
-            // thread can't be reached, retaining weak references is safer than
-            // releasing thread-affine XAML objects from this thread.
+            // Don't call Restore from the wrong thread. The state contains
+            // weak references only and is released when its UI thread exits.
             Wh_Log(L"Leaving modern XAML state for unreachable thread %u",
                    threadId);
         }
--- a/mods/cjk-spacer.wh.cpp
+++ b/mods/cjk-spacer.wh.cpp
@@ -2,7 +2,7 @@
 // @id              cjk-spacer
 // @name            CJK Spacer
 // @description     Add spaces between CJK characters and letters or digits in Explorer context menus and tooltips
-// @version         0.1.14
+// @version         0.1.15
 // @author          aenerv7
 // @github          https://github.com/aenerv7
 // @license         GPL-3.0
--- a/mods/cjk-spacer.wh.cpp
+++ b/mods/cjk-spacer.wh.cpp
@@ -2,7 +2,7 @@
 // @id              cjk-spacer
 // @name            CJK Spacer
 // @description     Add spaces between CJK characters and letters or digits in Explorer menus and tooltips
-// @version         0.1.1
+// @version         0.1.2
 // @author          aenerv7
 // @github          https://github.com/aenerv7
 // @license         GPL-3.0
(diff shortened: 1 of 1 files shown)


aenerv7:
Updated in `6435d59` to address the review on `bdf7b7d`:

- removed the unimplemented WinUI 3 `PopupWindowSiteBridge`/DWriteCore path and collapsed the DirectWrite helper to the verified system `dwrite.dll` path;
- restore pre-display classic menu rewrites through the public `SetMenuItemInfoW` API, avoiding a hook trampoline during `Wh_ModUninit`;
- added CJK pre-scans to classic menu hooks, an empty-tracker fast path, and same-process window filtering;
- replaced the 16-entry classic-tooltip theme array with an unbounded tracked set;
- excluded U+30A0 and U+30FB punctuation from CJK boundaries and documented menu-displayed file names.

Validated with Windhawk's real non-`WH_EDITING` macro mode (syntax and object compilation), editor mode, the repository PR validator, and local unit tests. The currently queued AI review was requested for the previous `2e2094d` head; a new review will be requested for `6435d59` after that request completes.

aenerv7:
Updated in `2cf5c60` to address the two main findings from the review on `2e2094d`:

- replaced `ShowWindow` / `SetWindowPos` / `DestroyWindow` discovery with creation-time tracking through `CreateWindowExW`, `CreateWindowInBand`, and `CreateWindowInBandEx`, plus a one-time same-process `EnumWindows` pass in `Wh_ModAfterInit`;
- install menu, themed-tooltip, and modern DirectWrite/window hooks only when their respective settings are enabled; settings changes now request a full mod reload;
- moved the allocation-free CJK pre-scan before the current-thread popup/window-state query;
- clarified the same-UI-thread DirectWrite scope, themed-only classic tooltip support, and non-`explorer.exe` shell surfaces in both the embedded and project README.

Validated with Windhawk's real non-`WH_EDITING` macro mode (syntax and object compilation), editor mode, the repository PR validator, and local unit tests.

aenerv7:
Updated in `9080ea8` to narrow the mod's scope after runtime testing:

- classic menu text is now transformed only during `TrackPopupMenu` / `TrackPopupMenuEx`; dynamic changes are limited to that active popup, and recorded strings are restored immediately when it closes;
- taskbar-thumbnail suppression is now process-wide rather than same-thread-only;
- modern rewriting is also suppressed for XAML popups owned by thumbnail hosts and whenever the pointer is over a taskbar-thumbnail surface;
- the metadata and both README copies now describe only context menus and pointer tooltips, and version is bumped to `0.1.6`.

Validated in Windhawk's real non-`WH_EDITING` macro mode (syntax and object compilation), editor mode, the repository PR validator, and local unit tests including immediate classic-menu restoration.

The currently queued AI review was requested for the previous `2cf5c60` head. A fresh review will be requested for `9080ea8` after that queued request completes.

aenerv7:
Updated in `8014f24` to address all three main findings from the review on `9080ea8`:

- resolved `GetWindowTheme` and enumerate existing same-process top-level and child `tooltips_class32` windows in `Wh_ModAfterInit`, so classic tooltips also work when enabled into a running Explorer;
- fixed `(UINT)-1` menu appends and made restoration fall back from the recorded index to locating the recorded spaced text; rewrite records are retained independently and restored in reverse order so index collisions don't discard earlier entries;
- restricted themed tooltip rewriting to `TTP_STANDARD`, `TTP_STANDARDTITLE`, `TTP_BALLOON`, and `TTP_BALLOONTITLE` in addition to the tracked theme handle.

Also applied the low-risk suggestions: shared read locking for tooltip themes, deferred cursor probing, HWND class revalidation, `GetModuleHandleW` for user32, and a 16-level submenu recursion cap. Version is now `0.1.7`.

Validated with Windhawk's real non-`WH_EDITING` object compilation, editor mode, the repository PR validator, and expanded local tests covering shifted-index restoration, colliding recorded indices, `(UINT)-1`, and tooltip part filtering.

aenerv7:
Updated in `6a865fb` to address both main findings from the review on `8014f24`:

- changed `modernUiText` to opt-in (`false` by default) in metadata, runtime defaults, and documentation, so the stable classic menu and tooltip paths are the out-of-box behavior;
- added a separate `g_trackedModernPopupCount` and use it for the render-path fast exit, so the taskbar's persistent `TaskListThumbnailWnd` no longer defeats the no-popup path.

Also applied the directly actionable optional suggestions:

- pre-display menu rewriting calls the original `SetMenuItemInfoW` trampoline when available, avoiding redundant hook re-entry;
- `InsertMenuW` now records `MF_POPUP | MF_BYCOMMAND` insertions by their spaced text;
- existing-tooltip discovery filters top-level windows by process before enumerating children;
- `IsUnicodeLetterOrDigit` now has an ASCII fast path.

Version is now `0.1.8`. Validated with Windhawk's real non-`WH_EDITING` syntax and object compilation, editor mode, the repository PR validator, and expanded local tests for the opt-in default, popup-only counter, and `MF_POPUP` restoration.

aenerv7:
Updated the PR with the XAML-targeted modern UI implementation requested after review:

- Replaced the thread-wide DirectWrite hooks with XAML Diagnostics visual-tree handling for both Windows.UI.Xaml and Microsoft.UI.Xaml.
- Only TextBlock elements below MenuFlyoutPresenter or ToolTip are modified; unrelated XAML text and taskbar thumbnails are no longer in scope.
- Added EVENT_OBJECT_SHOW/HIDE/DESTROY lifecycle tracking so hidden/reused popup HWNDs are removed from the visible map and restored before reuse.
- Added WindowFromDC confirmation for classic themed tooltip drawing, with theme-handle fallback only when the HDC has no associated window.
- Updated the readme, bumped the mod to 0.1.9, and added regression coverage for modern target classification and classic tooltip HDC confirmation.

Validation completed locally: Windhawk editor flags, real WH_MOD object compilation, PR validator, and the standalone tests all pass.

aenerv7:
Addressed the latest submission review in 0.1.10:

- Replaced the shared watcher state map with thread-local XAML source state keyed by watcher and instance handle.
- Removed presenter `TextBlock.Text` writes and the associated Loaded/Unloaded delegates; modern handling now updates `MenuFlyoutItem.Text`, `MenuFlyoutSubItem.Text`, or string-valued `ToolTip.Content`.
- Associated each state with its specific popup HWND, so HIDE/DESTROY restores only that popup's sources.
- Added UI-thread teardown via `RunFromWindowThread`; unreachable no-destroy thread-local state is retained instead of released from the wrong thread.
- Made the strong visual-tree watcher container `[[clang::no_destroy]]` with explicit unadvise/release.
- Moved XAML Diagnostics connection attempts to a dedicated worker with retry backoff and removed the ineffective out-of-context WinEvent fallback.
- Put tracked theme handles first in the classic tooltip fast path, retained `WindowFromDC` class confirmation, removed unused `-lcomctl32`, and refreshed the README and PR description.
- Added regression coverage for source type classification and popup-specific restoration.

Validation completed locally: Windhawk's compiler produced x64 and ARM64 builds, the standalone test suite passed, the upstream PR validator reported no warnings, `diff --check` passed, and the repository/local source copies are identical.

aenerv7:
Addressed the latest lifecycle review in 0.1.11:

- `WindhawkTap::SetSite` now always balances the XAML Diagnostics module reference for a non-null site before checking the stopping flag.
- Removed unnecessary `[[clang::no_destroy]]` from the thread-local weak-only state map and the plain `DWORD` thread registry; the strong COM watcher container remains protected.
- Removed the `IsLoaded()` gate so source dependency properties are updated deterministically regardless of XAML load ordering.
- Closed the diagnostics wake-event race with a short mutex covering handle lookup, signaling, detachment, and shutdown. The worker receives its event handle directly, and shutdown detaches the globals before joining and closing the handles.
- Replaced exception-based tooltip string unboxing with `IPropertyValue`/`PropertyType::String`.
- Switched the registered message suffix to `WH_MOD_ID`, removed unused XAML Media headers, and added the zero-count fast path to `CloseThemeDataHook`.
- Updated the standalone README and PR description, and bumped the mod to 0.1.11.

Validation completed locally: Windhawk's compiler produced x64 and ARM64 builds, the standalone tests passed, the upstream PR validator reported no warnings, `diff --check` passed, and both source copies are identical.

aenerv7:
Updated the PR with 0.1.12 to address both remaining required findings and the low-risk follow-ups:

- Replaced the thread-wide SHOW sweep with explicit per-thread `pending` and per-HWND `ownedByPopup` sets. New sources are drained once into the popup that opens; HIDE restores while retaining ownership for HWND reuse, and DESTROY clears ownership and re-pends surviving sources.
- Added dependency-property binding protection. Modern menu and tooltip sources are modified only when `ReadLocalValue` returns a plain `PropertyType::String`; bindings, styles, expressions, and other content are skipped.
- Moved DESTROY cleanup before HWND thread/process lookup so invalidated popup handles still release their source ownership on the in-context UI thread.
- Avoided accumulating `uxtheme.dll` references by using `GetModuleHandleW` first and explicitly releasing the fallback load.
- Removed the dead child-window tooltip enumeration, added the explicit `<functional>` include, restored past-end menu insertions correctly, and prevented command-ID lookup from selecting submenu entries.
- Documented the local-string limitation, popup ownership lifecycle, and classic-menu text-matching interoperability.
- Added regression coverage for pending-to-owner transfer, HIDE versus DESTROY behavior, past-end insertion restore, and submenu/command ID collisions.

Validation completed locally: Windhawk's compiler produced x64 and ARM64 builds, the expanded standalone tests passed, the upstream PR validator reported no warnings, `diff --check` passed, and both source copies are identical.

aenerv7:
Updated in fff49d5 (v0.1.13) to address the latest submission review:

- Replaced the pending/owned popup heuristic with exact XAML source Add/Remove lifetime handling, removing popup HWND attribution and WinEvent tracking entirely.
- Removed the joined diagnostics worker. Initial connection now runs from `Wh_ModAfterInit`, while a `LoadLibraryExW` hook handles late XAML module loads; the connection-name loop also observes the stop flag.
- A diagnostics connection is recorded only when `SetSite` actually registers a new watcher generation, so a blocked call returning `S_OK` is no longer treated as connected. The setting and README now name the two Windows 11 Styler mods as possible conflicts.
- Switched `RunFromWindowThread` to the canonical blocking `SendMessage` form.
- `SetSite(nullptr)` and watcher replacement now remove/unadvise the old watcher, with advise/unadvise race handling.
- Dead or reused UI-thread IDs are pruned explicitly using thread creation-time fingerprints. This intentionally avoids a mutex-taking `ModernThreadState` destructor.

Validation completed locally: Windhawk x64 and ARM64 compilation, the CJKSpacer test executable, source parity, `git diff --check`, and `.github/pr_validation.py` all pass.

aenerv7:
Updated in 1b902af (v0.1.14) for the latest submission review:

- The late-load hook now resolves and hooks the real `kernelbase.dll!LoadLibraryExW` implementation instead of the kernel32 forwarder.
- The loader hook no longer performs XAML Diagnostics initialization inline. It only requests a single-flight, short-lived worker, keeping TAP loading and COM activation outside the loader path.
- Each worker obtains an extra reference to the mod image before it starts and releases it with `FreeLibraryAndExitThread`; unload doesn't join the worker, while the worker can't continue in unloaded code.
- The thread-local reentrancy guard now lives at the top of `EnsureModernXamlDiagnostics`, so it covers every entry point.
- Removed the unreachable already-modified branch from `ModernTextState::Apply` and stopped pruning all registered UI threads on each new registration; dead/reused IDs remain fingerprinted and are pruned during unload.
- README and tests were updated for the new loader/worker lifecycle.

Validation completed: Windhawk x64 and ARM64 compilation, local CJKSpacer tests, source parity, `git diff --check`, and `.github/pr_validation.py` all pass.

aenerv7:
Updated in c2ba635 (v0.1.15) for the latest submission review:

- Added a pending request flag. Requests arriving while the single-flight worker is running are consumed before it exits, including the final race after the worker clears its running state.
- Tracked WUX and MUX diagnostics connections independently. `WindhawkTap` records the connection flavor, clears only the corresponding connected flag on `SetSite(nullptr)`, and schedules that flavor for reconnection.
- Added an explicit module reference to every `VisualTreeWatcher` Advise thread and terminate it with `FreeLibraryAndExitThread`; watcher creation now fails cleanly if the reference or thread cannot be acquired.
- Added the `g_stoppingModernUi` check to the visual-tree Add path.
- Optimized `CloseThemeDataHook` to take the exclusive theme-set lock only for a tracked handle, and removed redundant `CJK Spacer` prefixes from lifecycle logs.
- Updated the embedded README, standalone README, and PR description.

Validation completed: Windhawk x64 and ARM64 compilation, local CJKSpacer tests, source parity, `git diff --check`, and `.github/pr_validation.py` all pass.

aenerv7:
Follow-up metadata fix: bumped the module metadata and mirrored source to v0.1.15 in 4514999; validation and source-parity checks still pass.

aenerv7:
保留现代 XAML 路径，不拆分为独立 PR，理由如下：

- Windows 11 主资源管理器和桌面右键菜单的主要实现就是 XAML；移除该路径会让模组默认覆盖范围退回“显示更多选项”和经典 Tooltip，无法完整实现 CJK Spacer 的目标。
- 现代路径已经默认关闭，并明确标注为 best-effort；与其他 XAML Diagnostics 模组的竞争是可见、可关闭的用户选择，不会影响经典菜单和 Tooltip 路径。
- 当前实现已经包含失败重试抑制、连接断开重试、UI 线程恢复和跨代状态保护。将它拆出会增加发布和使用复杂度，而不会消除底层单消费者限制。

README 和设置描述已同步说明这一决策及 live HMENU/Text Replace 兼容性。

aenerv7:
Addressed the latest AI review in `7ea7a10`:

- Removed the per-load modern-UI generation machinery. The stop flag now gates callbacks and detached workers directly.
- Kept the thread creation-time fingerprint, with an explanatory code comment. A dead UI thread ID can be recycled by a new thread which also owns windows; checking only the ID could dispatch unload cleanup into that unrelated thread.
- Removed the `CreateWindowExW`, `CreateWindowInBand`, and `CreateWindowInBandEx` initialization hooks. Existing modules are handled from `Wh_ModAfterInit`, and late/static-import paths are handled by `kernelbase!LoadLibraryExW`, including the `CoreMessagingXP.dll` trigger.
- Made the classic menu, classic Tooltip, and modern loader hook groups fail initialization on any required registration failure, avoiding partial measure/draw or rewrite/restore behavior.
- Added stale popup-handle pruning and a cheap `DestroyMenu` fast path.
- Removed unreachable dynamic `uxtheme.dll` loading bookkeeping and clarified imported `CloseThemeData` use during teardown.
- Reduced locking for unrelated theme closes.
- Closed the modern-state unload registration race and added `Wh_ModBeforeUninit` stopping.
- Made XAML unadvising non-throwing and made state restoration structurally reentrancy-safe.
- Updated the implementation documentation and bumped the version to `0.1.25`.

Validation completed locally: Windhawk x64 syntax check passed, source copies are byte-identical, `git diff --check` passed, and the statically linked regression suite reports `All CJKSpacer spacing tests passed.`

aenerv7:
Addressed the review for `7ea7a10` in `39a4d05`:

- Made `OpenThemeDataForDpi` optional on systems where the Windows 10 1703+ export is absent. If the export exists, hook registration must still succeed; all Vista-era Tooltip hooks remain required as one atomic group.
- Removed the redundant classic-menu and classic-Tooltip setting loads from installed hooks.
- Added the popup-count fast path to `IsCreatedClassicPopupMenu` and moved cheap string/type checks ahead of popup tracking locks.
- Changed `DestroyMenuHook` to skip tree traversal when there are no rewritten records and the specific menu isn't tracked.
- Wrapped `AdviseVisualTreeChange` interface acquisition/call so an exception can't escape the raw worker thread procedure.
- Removed the unused `<new>` include.
- Added timed cleanup for construction-time records that never reach `TrackPopupMenu*`: old pending live records are restored, invalid handles are discarded, scans are throttled to once per 30 seconds, and a 1024-record hard threshold trims pending records toward 768. The current menu is excluded from cleanup so a newly written value can't be immediately reverted.
- Updated the embedded documentation and project README to state that menu-bar dropdowns created through `CreatePopupMenu` can be processed even when user32 displays them without `TrackPopupMenu*`.
- Bumped the version to `0.1.26` and added a regression test for restoring an expired pending menu record.

Validation: Windhawk x64 syntax check passed, source copies are byte-identical, `git diff --check` passed, and the statically linked regression suite reports `All CJKSpacer spacing tests passed.`

m417z:
Looks good overall.

> The Win32 menu hook technique is based in part on the open-source Windhawk
> "Text Replace" mod by m417z.

> On Windows 11, the primary Explorer and desktop context menus are modern XAML
> menus. The `modernUiText` setting is disabled by default because each named
> XAML Diagnostics connection permits only one consumer, and tools such as
> Taskbar Styler or File Explorer Styler may already use or block those
> connections.

Have you looked at `HookD2DAndDWrite` and `HookGdiplus` in Text Replace? It should work for XAML elements as well, perhaps that will make XAML Diagnostics unnecessary.

aenerv7:
Thanks for the review.

The main reason I prefer XAML Diagnostics than HookD2DAndDWrite and HookGdiplus is I just want to narrow the scope down. This mod focused on MenuFlyoutItem.Text, MenuFlyoutSubItem.Text, ToggleMenuFlyoutItem.Text, RadioMenuFlyoutItem.Text and string format ToolTip.Content. Therefore, this mod can be a supplement of Text Replace and that is why I create this mod and how I actually use this mod.

m417z:
If both works, I think XAML Diagnostics has the largest impact and is best avoided. aside from the incompatibility with styler mods, it has some side effects such as crashes and memory leaks, examples:

* https://github.com/ramensoftware/windhawk-mods/issues/354
* https://github.com/ramensoftware/windhawk-mods/issues/1504

Windows 11 Taskbar Styler is a generic styling mod and it has you use it, but you might not have to.

Let me know what you think.

aenerv7:
OK, I Understood your concerns. I will remove XAML Diagnostics part to furtherly narrow this mod's scope down to only classic context menu and win32 tooltip.

m417z:
I think it's fine to have it as an option. I'm just saying that if there's another way to solve it (e.g. HookD2DAndDWrite and HookGdiplus), I think it's preferable.

aenerv7:
As I know so far, XAML Diagnostics is the only way to precisely point out the modern UI is context menu or tooltip. Other way to distinguish what the UI is in modern UI is not reliable. The original intention of this mod is to fix "bad manner" mixed CJK and non-CJK texts in classic context menu and win32 tooltip. So modern UI support for me is optional. I will do nothing if you think an option for modern UI support is acceptable. Otherwise, remove XAML Diagnostics part is also acceptable for me.

m417z:
In this case it's fine as it. I'll review and, if nothing else comes up, merge it soon. Thanks.

aenerv7:
My pleasure and I will keep digging to try to find more appropriate way to implement this mod in modern UI. If I found something I will update this mod ASAP.

m417z:
Looks good. Please do this one small change:

The Windhawk engine expects a mod to be unloaded after `Wh_ModUninit`. Remove the `AcquireCurrentModuleReference` and `FreeLibraryAndExitThread` usage, and instead, wait for all threads to end in `Wh_ModUninit`, so that when `Wh_ModUninit` returns, there's a single reference to the mod module, and it can be unloaded with a single `FreeLibrary` call.

aenerv7:
Just to confirm: did you mean Wh_ModUninit instead of Wh_ModInit?

m417z:
Yes :)

aenerv7:
I removed the AcquireCurrentModuleReference and FreeLibraryAndExitThread usage.

The XAML Diagnostics connection worker is now a managed thread created during modern UI initialization. It is signaled to stop and joined during shutdown. The asynchronous visual-tree watcher setup is preserved to avoid blocking SetSite, but every watcher thread handle is now tracked and joined as well.

Shutdown begins in Wh_ModBeforeUninit so that these threads stop before hook removal, and the same cleanup is called idempotently from Wh_ModUninit. Therefore, by the time Wh_ModUninit returns, no mod-created threads remain and only Windhawk’s single module reference is left.

PR #4953, opened 2026-07-30
Create fix-10-taskbar-notifications.wh.cpp

bozohi:
This is a mod that applies runtime patches regarding 11's notifications so it doesn't crash anymore in Windows 10 explorer mod created by @Anixx .

## Changelog

None

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/fix-10-taskbar-notifications.wh.cpp
+++ b/mods/fix-10-taskbar-notifications.wh.cpp
@@ -5,7 +5,7 @@
 // @version         0.1
 // @author          ilovethisgame
 // @github          https://github.com/bozohi
-// @include         explorer.exe
+// @include         %ProgramData%\Windhawk\Engine\ModsWritable\LegacyStore\explorer.exe
 // @architecture    amd64
 // @license         MIT
 // ==/WindhawkMod==
@@ -14,10 +14,8 @@
 /*
 This mod patches the Explorer shell provided by the Win10 taskbar mod at runtime.
 
-**IF YOUR WINDOWS VERSION IS NOT WINDOWS 11 24H2 OR ABOVE, DO NOT INSTALL THIS!**
-
 **!Important! You MUST install, follow the instructions and run atleast 2 times the [Win10 taskbar mod.](https://windhawk.net/mods/win10-taskbar-on-win11-24h2)**
-**If you don't follow these instructions carefully, explorer may behave in an unexpected (and likely unpleasant) manner.**
+**If you don't follow these instructions carefully, the mod will not apply correctly.**
 
 # Technical details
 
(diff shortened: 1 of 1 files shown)


Anixx:
Do I understand correctly, this fixes the Notification Center when running the Win10 taskbar on Win11, so that receiving a notification would not crash Explorer any more, yes?

bozohi:
> Do I understand correctly, this fixes the Notification Center when running the Win10 taskbar on Win11, 

@Anixx  It likely does not let the notification arrive to the notification center.

> so that receiving a notification would not crash Explorer any more, yes?

@Anixx  Yes! This is the verified intended behavior.

Anixx:
> "4. Worth folding into the parent mod rather than shipping alongside it"

I would voice a strong disagreement here. I as an author of that mod, I do not have an idea of the mechanism and cannot take the responsibility for the functioning of this mod. 

Additionally, the mods have
* Different target process (userinit.exe versus explorer.exe)
* Different methods (WINAPI hooking versus memory patching)
* Different residency status (that mod is active only at startup and quits when userinit.exe quits)

mods/fix-10-taskbar-notifications.wh.cpp
```diff
@@ -0,0 +1,130 @@
+// ==WindhawkMod==
+// @id              fix-10-taskbar-notifications
+// @name            Fix 11's notifications crash in Win10 taskbar under Win11 24H2+
+// @description     Stops notifications from crashing explorer
+// @version         0.1
+// @author          ilovethisgame
+// @github          https://github.com/bozohi
+// @include         %ProgramData%\Windhawk\Engine\ModsWritable\LegacyStore\explorer.exe
+// @architecture    amd64
```
m417z: After you switch to hooking, change this to `x86-64`.

mods/fix-10-taskbar-notifications.wh.cpp
```diff
@@ -0,0 +1,130 @@
+// ==WindhawkMod==
+// @id              fix-10-taskbar-notifications
+// @name            Fix 11's notifications crash in Win10 taskbar under Win11 24H2+
+// @description     Stops notifications from crashing explorer
+// @version         0.1
+// @author          ilovethisgame
+// @github          https://github.com/bozohi
+// @include         %ProgramData%\Windhawk\Engine\ModsWritable\LegacyStore\explorer.exe
```
m417z: This won't work with the portable Windhawk version. The win10-taskbar-on-win11-24h2 mod doesn't work too, but it might be updated in the future to support both the portable Windhawk version and ARM64.

Target `explorer.exe` and check the module version like the review suggested.

m417z (changes requested):
I agree about "serves users better inside that mod [...] than as a separate catalog entry they have to find", but it's up to @Anixx.

> 1. Nothing still verifies that the image being patched is the build 0x14DC40 came from
> 2. Go through the hooking engine instead of writing bytes into live code

Please address these items - check the module version and use function hooking (you can keep using a hardcoded function pointer address) instead of memory patching.

Anixx:
@m417z  I agree about "serves users better inside that mod [...] than as a separate catalog entry

I thought, mod patterns/metapackages were planned?

m417z:
Windhawk 2.0 makes it possible to import multiple mods from a JSON, that's a start. Maybe more related features will be added in the future to improve this, but nothing specific is planned for now.

mods/fix-10-taskbar-notifications.wh.cpp
bozohi: I see.

I'll take care of these concerns tomorrow, thanks for the review

m417z (approved):

bozohi:
@Anixx here's the link to my mod: https://windhawk.net/mods/fix-10-taskbar-notifications

PR #4955, opened 2026-07-30
Add foreground window primary monitor hotkey

fjdiazt:
## Changelog

* Add Ctrl+Alt+Shift+W action that moves only the foreground window to the primary monitor.

### DIFF
--- a/mods/gather-windows-to-monitor.wh.cpp
+++ b/mods/gather-windows-to-monitor.wh.cpp
@@ -2,7 +2,7 @@
 // @id              gather-windows-to-monitor
 // @name            Gather Windows To Monitor
 // @description     Move eligible open windows to a chosen monitor with global hotkeys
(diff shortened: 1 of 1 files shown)


fjdiazt:
/ready-for-reviewer

Leaving the latest AI findings unaddressed:
- The DPI-awareness finding doesn’t apply because windhawk.exe declares PerMonitorV2, and worker threads inherit the process default.
- The overlap/generalization suggestion is intentionally out of scope. This update adds one foreground-to-primary action alongside the existing bulk-gather workflow.
- The optional improvements and functionality notes are deferred or intentional behavior, not required for this update.

m417z:
Please do the first two optional improvements that were missed in the previous submission:

> `Enabled` setting is redundant

> `DebugLogging` setting is redundant too

PR #4960, opened 2026-07-31
v2.7.0: Fix verb truncation, & stripping, atomic SendInput; enhance context menu matching

LiHua81:
## Summary

v2.7.0 update for Click on Empty Explorer.

### Bugfixes
- **Verb truncation:** `GCS_VERB` → `GCS_VERBA` — Unicode build was writing wide chars into `CHAR*` buffer, truncating all verbs to 1 character
- **& markers:** `StrContainsNorm` now strips `&` so "Git Bash" matches "Open Git Ba&sh here"
- **SendInput:** merged split press/release calls into single atomic call

### Improvements
- **Perf:** subclass procs skip `CopySettings()` for unhandled messages
- **Debug:** dump shows normalized match text per entry (`→ match:`)
- **Docs:** expanded Context Menu Match section with rules table
- **Open in Terminal:** now launches Windows Terminal directly via `wt.exe -d <folder>` instead of matching the context-menu entry. Works on non-English Windows and survives shell updates that rename the menu verb/text. Requires Windows Terminal to be installed.

## Test Plan

- Configure a trigger (e.g. double-click) to "Open in Terminal", then click empty space in a normal folder → Windows Terminal opens in that folder
- Repeat with a folder whose path contains spaces → terminal opens in the correct folder (path is quoted)
- Switch the system display language to a non-English locale → "Open in Terminal" still works (no longer depends on localized menu text)
- Virtual folders (This PC, Recycle Bin) → no terminal opens; mod logs a note instead of crashing
- Other actions (Go Up, Duplicate Tab, Open in VS Code, Custom Hotkey, etc.) still function as before
- Enable the Windhawk debug log and confirm no errors/exceptions during the above

## MOD AUTHORSHIP

- [ ] Claude
- [ ] Gemini
- [ ] ChatGPT
- [x] Another AI (please specify): CodeBuddy Code
- [ ] Other (please specify): ____


LiHua81:
Regarding the `openInVSCode` default: keeping `L"Code"` deliberately.

The VS Code background verb is install-dependent — on my machine it is registered as `AnyCode` (`HKCR\Directory\Background\shell\AnyCode`), not `VSCode`, so switching the default to `L"VSCode"` would break the action on standard installs: the exact pass would miss, and the substring pass cannot match `vscode` inside `anycode` anyway. A substring default of `"Code"` is the only one that reliably matches the entry across install variants and localized labels, at the cost of the documented Encode/Decode-style collision risk, which users can avoid with the per-trigger *Context Menu Match* override or by typing the entry verb.

m417z:
Any reason not to address the finding in the last review with the suggested fix?

> Fix by taking an owning reference for the duration of the call — the object then survives a nested release, and the nested `g_pUIAutomation = nullptr` still does the right thing for the next click:
> 
> ```
> winrt::com_ptr<IUIAutomation> pUIA;
> pUIA.copy_from(GetUIAutomation());
> if (!pUIA)
>     return DefSubclassProc(hWnd, uMsg, wParam, lParam);
> ```

LiHua81:
Regarding the openInVSCode finding: left as-is deliberately. The suggested exact match `VSCode` does not hold on this machine — the only VS-related folder-background verb registered here is `AnyCode` (Visual Studio's VSLauncher, ProgrammaticAccessOnly), and no `VSCode` verb exists in either HKCR or HKCU, so `VSCode` would also fall through to the substring pass. The `Code` preset is equivalent to `openWithContextMenu` with a preset string; users wanting an exact hit can set the per-trigger Context Menu Match. Functionality is frozen for this revision.

PR #4979, opened 2026-08-01
Add Control Panel in Modern File Explorer

crazyboyybs:
## Changelog

* Changelog item 1...
* Changelog item 2...

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [X] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/control-panel-in-modern-file-explorer.wh.cpp
+++ b/mods/control-panel-in-modern-file-explorer.wh.cpp
@@ -48,6 +48,10 @@ customizations.
   continue to follow that setting.
 * Pinning a Control Panel window pins File Explorer because the window shares
   File Explorer's taskbar identity.
+* Control Panel launched via Run (`Win+R` → `control`) or `control.exe`
+  directly is also covered — the mod hooks `control.exe` too, since it goes
+  through the same host-selection functions before launching its own
+  Explorer window.
 
 ---
 
@@ -81,6 +85,10 @@ do Explorador legado.
   As pastas comuns continuam seguindo essa configuração.
 * Fixar uma janela do Painel de Controle fixa o Explorador de Arquivos, pois a
   janela compartilha a identidade do Explorador na barra de tarefas.
+* O Painel de Controle aberto pelo Executar (`Win+R` → `control`) ou
+  diretamente pelo `control.exe` também é coberto — o mod também intercepta
+  o `control.exe`, pois ele passa pelas mesmas funções de seleção de host
+  antes de abrir sua própria janela do Explorer.
 
 ---
 
@@ -115,6 +123,10 @@ personalizaciones heredadas de Explorer.
   proceso independiente. Las carpetas normales siguen respetando esa opción.
 * Anclar una ventana del Panel de control ancla el Explorador de archivos porque
   la ventana comparte su identidad en la barra de tareas.
+* El Panel de control abierto mediante Ejecutar (`Win+R` → `control`) o
+  directamente con `control.exe` también está cubierto — el mod también
+  engancha `control.exe`, ya que pasa por las mismas funciones de selección
+  de host antes de abrir su propia ventana de Explorer.
 */
 // ==/WindhawkModReadme==
 
@@ -154,8 +166,11 @@ static LoadLibraryExW_t LoadLibraryExW_Original;
 static thread_local IPropertyStore* g_controlPanelPropertyStore;
 static std::atomic<bool> g_explorerFrameInitializationStarted;
 static std::atomic<bool> g_setValueHookInstalled;
+static std::atomic<WindowPropertyStore_SetValue_t> g_hookedSetValue;
+static std::atomic<bool> g_propertyStoreMismatchLogged;
 static std::atomic<bool> g_unloading;
 static HMODULE g_explorerFrameRef;
+static HMODULE g_setValueModuleRef;
 
 static bool IsCopiedAppIdentityProperty(const PROPERTYKEY& key) {
     return IsEqualPropertyKey(key, PKEY_AppUserModel_RelaunchCommand) ||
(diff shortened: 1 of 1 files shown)


crazyboyybs:
/ai-review

Addressed all required review items. The `IPropertyStore::SetValue` hook was retained because runtime logs confirmed that `PKEY_AppUserModel_ID` is written directly and bypasses `CopyPropertyFromItemToPropStore`, it is now best-effort, COM-safe, and scoped to the exact Control Panel property store. The README now explains how the mod differs from aerexplorer and includes a screenshot

crazyboyybs:
/ai-review

Addressed all required findings and optional improvements. The additional `CExplorerLauncher::GetHostFromTarget` hook was not added because the existing `UseSeparateProcess` hook already handles Control Panel specifically, and testing confirmed that “Launch folder windows in a separate process” does not interfere with the mod. A generic symbol-resolution retry was also omitted because it could duplicate partially registered hooks without reliably distinguishing transient failures from missing symbols. All other suggested safety, late-loading, logging, and cleanup improvements were applied

crazyboyybs:
/ai-review

Addressed all required findings and optional improvements. The additional `CExplorerLauncher::GetHostFromTarget` hook was not added because the existing `UseSeparateProcess` hook already handles Control Panel specifically, and testing confirmed that “Launch folder windows in a separate process” does not interfere with the mod. A generic symbol-resolution retry was also omitted because it could duplicate partially registered hooks without reliably distinguishing transient failures from missing symbols. All other suggested safety, late-loading, logging, and cleanup improvements were applied

Anixx:
How is this mod different from this one? https://windhawk.net/mods/cpl-classic-view-lite

crazyboyybs:
> How is this mod different from this one? https://windhawk.net/mods/cpl-classic-view-lite

@Anixx  They do two different things, mine forces the control panel to open inside the new file explorer with the winui bar and makes it not look like a different program on the taskbar. Also, both can be used together for better integration inside file explorer

Anixx:
Hmmm. What is "new file explorer"? As far as I know, there is only one so far.

Your mod makes File Explorer to hang if Control Panel is opened from OpenShell start menu on my system (does not hang if opened another way).

crazyboyybs:
> Hmmm. What is "new file explorer"? As far as I know, there is only one so far.
> 
> Your mod makes File Explorer to hang if Control Panel is opened from OpenShell start menu on my system (does not hang if opened another way).

Control Panel opens in a separate window with the old ribbon interface. This mod redirects it to the one with new winui top part and forces it to not have it's own app icon or separate process on the taskbar. As for the hang, I can probably look into it in a new version, is not top priority as open shell is a third party program

crazyboyybs:
/ai-review

`Wh_ModAfterInit` no longer calls `Wh_ApplyHookOperations()` unconditionally, it now tracks whether anything was actually registered before applying. Also added null guards for `IsControlPanel_Original` in both `UseSeparateProcess_Hook` and `CopyPropertyFromItemToPropStore_Hook`, so a partial `HookSymbols` failure can no longer reach a null call. Re-tested `SHGetPropertyStoreForWindow` with the `HWND_MESSAGE` probe window (logging enabled, Explorer restarted): no `"Failed to hook IPropertyStore::SetValue"`, and `"Suppressed Control Panel AppUserModel.ID"` logs for every Control Panel window, taskbar grouping confirmed working

crazyboyybs:
/ai-review

Dropped the `HWND_MESSAGE` probe window entirely. The SetValue hook is now installed lazily from the real `IPropertyStore` the first time `CopyPropertyFromItemToPropStore_Hook` sees a Control Panel item, no `CoInitializeEx`, no window, no -lole32 (needed -luuid instead, for `GUID_NULL`). `GET_MODULE_HANDLE_EX_FLAG_PIN` replaced with an ordinary reference, released in a new Wh_ModUninit (and in Wh_ModInit's own failure path, since Wh_ModUninit doesn't run there). `CopyPropertyFromItemToPropStore` is now optional=true, losing that symbol on a future build only drops taskbar-identity suppression, not the mod's main feature

crazyboyybs:
The lazy `SetValue` install now checks `g_unloading` before calling `SetFunctionHook/ApplyHookOperations`, closing the narrow teardown-window crash you flagged (verified both API contracts directly in windhawk_api.h). Confirmed via log that control.exe's `Wh_ModInit` does reach `InitializeExplorerFrameHooks`, documented in the README's Known behavior section (all three languages)

Also found, while confirming the above, that the `GetHostFromTarget` CLSID substitution has never actually fired in any tested launch path (Start menu, Win+R, before or after include control.exe), `UseSeparateProcess_Hook` alone, once active inside control.exe too, is what fixes Win+R: `GetHostFromTarget`'s own internal logic already returns `CLSID_DesktopExplorerHost` once `UseSeparateProcess` says "don't use a separate process", confirmed directly from a control.exe log line. Downgraded the hook to diagnostic-only (logs if the legacy CLSID is ever seen, no longer mutates it) rather than ship a substitution that's
never been exercised, matches your own note that a wrong sentinel there could break Control Panel entirely in some untested config

Took the rest of the optional items: removed the dead `if (!result)` check (verified against the x64 ABI, a function returning a struct by value always returns the hidden-return-slot pointer, never null), added a module reference for whatever DLL implements `IPropertyStore::SetValue` (symmetric with the existing ExplorerFrame.dll one, matters more now with the short-lived control.exe host), and both `Wh_ApplyHookOperations()` call sites now log on failure. Restored the "different `SetValue` implementation" diagnostic without the probe window, compares against what was actually installed instead of a probe-derived pointer

Kept the `SHCreateItemFromIDList` ordering as it is, the diagnostic value proved itself this round, and `GetHostFromTarget` turns out to only fire for the four special-cased host CLSIDs, not on every window open, so the trade-off is smaller than it looked

/ai-review

crazyboyybs:
Re-tested the Open-Shell hang against current code and is confirmed fixed. The suspected cause (`CoInitializeEx` -> `CreateWindowExW` -> `SHGetPropertyStoreForWindow` -> `CoUninitialize` running synchronously at mod startup) was already removed in the previous round's probe-window deletion. Removed `GetHostFromTarget_Hook` entirely, it had been reduced to log-only and never fired in any tested path, so it was pure cost (a symbol resolution, a hook, and a `SHCreateItemFromIDList` + `IsControlPanel` pair per host-selection call) for nothing. Also dropped wrl/client.h and the entire @compilerOptions line with it, verified via a real link pass that nothing in the file needs any extra library anymore. Investigated hooking the real ExplorerFrame caller instead of `IPropertyStore::SetValue`. Found it, `CShellBrowser::OnNavigationResult`, confirmed by disassembling it directly (WinDbg). It's exactly what you mentioned, `SHGetPropertyStoreForWindow`, four `CopyPropertyFromItemToPropStore` calls, then a direct `SetValue(PKEY_AppUserModel_ID)` through vtable slot 6. Decided not to migrate: the HWND it operates on comes from a private `CShellBrowser` field (`GetAncestor`-derived, no public accessor confirmed, would need to verify `CShellBrowser` implements `IOleWindow::GetWindow()`, which I couldn't confirm). Migrating would trade a verified, tested mechanism (with the `g_unloading` fix already closing the one concrete crash risk) for one resting on an unconfirmed private-ABI assumption, to satisfy a soft "avoid if possible" performance note rather than a proven bug, especially now that the hang this was meant to help diagnose is confirmed fixed by an unrelated change. Kept the current shape plus the mismatch diagnostic, per your own fallback allowance.

Took both optional items: `g_explorerFrameRef` is now `std::atomic<HMODULE>` (real cross-thread race, now consistent with the rest of the file), and `RefSetValueModule`/`g_setValueModuleRef` are gone per your own follow-up note that the reference wasn't buying enough to justify it. Added the cpl-classic-view-lite differentiation sentence to the README (all three languages)

/ai-review

PR #4982, opened 2026-08-02
Add startup-app.delay-fix.wh.cpp

Meteony:
## Changelog

Windows intentionally delays startup applications after sign-in to reduce
system load, which can cause apps to open minutes after startup.

Common fix (which the mod intends to emulate): 
Common Fix

This mod redirects Explorer's startup-serialization key to a dedicated,
volatile key where both `WaitforIdleState` and `StartupDelayInMSec` are zero.
It never creates, modifies, or deletes the user's real `Serialize` key.

The redirect key is in-memory only. The mod deletes it on a normal disable,
and Windows discards it when the user registry hive unloads at sign-out or
shutdown if Explorer exits unexpectedly.

Windows reads the startup-delay values when Explorer starts, so enabling or
disabling the mod takes effect after the next sign-in or Explorer restart.

This only affects the delay Explorer applies to startup-item processing. It
doesn't change delays configured by other systems, such as Task Scheduler.

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/startup-app-delay-fix.wh.cpp
+++ b/mods/startup-app-delay-fix.wh.cpp
@@ -67,8 +67,6 @@ EXTERN_C NTSYSAPI NTSTATUS NTAPI NtQueryKey(
     HANDLE keyHandle, KEY_INFORMATION_CLASS keyInformationClass,
     PVOID keyInformation, ULONG length, PULONG resultLength);
 
-EXTERN_C NTSYSAPI NTSTATUS NTAPI RtlNtStatusFromDosError(ULONG dosError);
-
 using NtOpenKey_t = NTSTATUS(NTAPI*)(PHANDLE keyHandle,
                                      ACCESS_MASK desiredAccess,
                                      POBJECT_ATTRIBUTES objectAttributes);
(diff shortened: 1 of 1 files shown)


Meteoni:
Verified after a full reboot with the Serialize key absent. Explorer used NtOpenKeyEx; the mod substituted a read-only virtual handle and successfully returned WaitforIdleState=0 through KeyValuePartialInformation.

m417z:
I agree with functionality note 1. How about creating a regular, non-volatile key, if missing, and then remove it on unload only if it stays empty? I think that as a side effect, its better to have a leftover empty key than having actual data deleted.

It'd be even better to avoid both (e.g. hook key open and simulate success or redirect), but it'd a bit more complicated. If you'd like to try it, I think the easiest way is to create a dedicated volatile key Explorer doesn't use, and redirect `Serialize` opens to it. In fact, it might even be simpler because you don't need `GetKeyPath`, since you hook the key open, you can save the handle and check for it instead. Let me know what you think.

Meteoni:
> I agree with functionality note 1. How about creating a regular, non-volatile key, if missing, and then remove it on unload only if it stays empty? I think that as a side effect, its better to have a leftover empty key than having actual data deleted.
> 
> It'd be even better to avoid both (e.g. hook key open and simulate success or redirect), but it'd a bit more complicated. If you'd like to try it, I think the easiest way is to create a dedicated volatile key Explorer doesn't use, and redirect `Serialize` opens to it. In fact, it might even be simpler because you don't need `GetKeyPath`, since you hook the key open, you can save the handle and check for it instead. Let me know what you think.

You can be right. I'll look into it.

PR #4988, opened 2026-08-02
MacOS Minimize Animation v3.1.3: shrink modern genie canvas to fix AMD stutter

Abdullah-Masood-05:
## Summary

A user reported that the modern (Direct2D) genie is slow and choppy on AMD GPUs - on their RX 6900 XT and on a Vega 8 iGPU - while the same setup with a GTX 1080 ran it perfectly, and the classic style was smooth everywhere. The cause turned out to be the modern engine's render canvas: it was the full virtual desktop (every monitor combined), so every frame called `UpdateLayeredWindow` on a whole-screen layered window, which AMD drivers handle slowly.

The fix, which the same user tested and confirmed smooth on both AMD machines, is to shrink the modern canvas to the tight bounding box the classic engine already uses: the union of the window's monitor and the dock target's monitor, clamped to the window rect and dock X with a small margin. The genie mesh math only ever occupies the region between the window rect and the dock target, so the tighter canvas loses nothing visually.

Changes:
- Modern engine now computes its canvas from `MonitorFromRect` (window) and `data->hMon` (dock), clamped to both monitors, instead of `GetSystemMetrics(SM_*VIRTUALSCREEN)`.
- Version bumped to 3.1.3.

## Test plan
- [x] Modern style minimize/restore on AMD (6900 XT and Vega 8 iGPU reported fixed)
- [x] Modern style still smooth on NVIDIA (GTX 1080 reported fine before, must not regress)
- [x] Classic style unaffected (already used the tight canvas)
- [ ] Multi-monitor: window and dock on different monitors still animates correctly

### DIFF
--- a/mods/macos-minimize-animation.wh.cpp
+++ b/mods/macos-minimize-animation.wh.cpp
@@ -818,6 +818,15 @@ DWORD WINAPI MacGenieAnimThread(LPVOID lpParam) {
     if (boundTop    < boundMon.top)    boundTop    = boundMon.top;
     if (boundBottom > boundMon.bottom) boundBottom = boundMon.bottom;
 
+    // The monitor clamp is only there to stop the box growing beyond the screens
+    // involved - it must never clip the window itself or the dock target away.
+    if (boundLeft  > data->targetRect.left)   boundLeft  = data->targetRect.left;
+    if (boundRight < data->targetRect.right)  boundRight = data->targetRect.right;
+    if (boundTop   > data->targetRect.top)    boundTop   = data->targetRect.top;
+    if (boundBottom < data->targetRect.bottom) boundBottom = data->targetRect.bottom;
+    if (boundLeft  > dockXB - 16)             boundLeft  = dockXB - 16;   // iGeom is dockX ± 11
+    if (boundRight < dockXB + 16)             boundRight = dockXB + 16;
+
     int vLeft = boundLeft;
     int vTop = boundTop;
     int vWidth = boundRight - boundLeft;
--- a/mods/macos-minimize-animation.wh.cpp
+++ b/mods/macos-minimize-animation.wh.cpp
@@ -752,6 +752,58 @@ static void CalculateLampVertexMacOS(float tx, float ty, float p, const Geometry
     *outY = w.y + y + offsetY;
 }
 
+// Shared by both engines: union of the monitors involved (the window's own
+// monitor and the monitor the genie funnels to). With Multi-monitor support off
+// (the default) data->hMon is the primary monitor, so a window sitting on a
+// secondary display still needs its own monitor in the union or the canvas would
+// collapse onto the primary and clip the first half of the animation.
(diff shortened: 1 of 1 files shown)


Abdullah-Masood-05:
Thanks, both clipping cases are real. I applied the re-expansion after the monitor clamp in both engines' canvas logic:

```cpp
// The monitor clamp is only there to stop the box growing beyond the screens
// involved - it must never clip the window itself or the dock target away.
if (boundLeft  > data->targetRect.left)   boundLeft  = data->targetRect.left;
if (boundRight < data->targetRect.right)  boundRight = data->targetRect.right;
if (boundTop   > data->targetRect.top)    boundTop   = data->targetRect.top;
if (boundBottom < data->targetRect.bottom) boundBottom = data->targetRect.bottom;
if (boundLeft  > dockXB - 16)             boundLeft  = dockXB - 16;   // iGeom is dockX ± 11
if (boundRight < dockXB + 16)             boundRight = dockXB + 16;
```

I also added the `boundBottom` re-expansion (your snippet covered left/right/top, but a window straddling stacked monitors with the larger part on the lower one can push its bottom edge below `boundMon.bottom` too - same identity-frame clipping you described for the horizontal case).

Good catch on `targetDockX` not being keyed to `hMon` by the way - restoring the primary's taskbar button for a secondary-monitor window really does aim the funnel at a column the old box would have cut off.

Abdullah-Masood-05:
Thanks for the thorough review. Fixed all three in `a3a2452d` — I went with the exact sway bound rather than the mesh sampling, to keep the change small:

1. **Sway pad** — the horizontal pad is now the exact bound of the mesh's sway term in `CalculateLampVertexMacOS` (amplitude `max(|w.x-i.x|, |w.right-i.right|)/7`, proportional to the window-to-dock distance), so a narrow window far from the taskbar column no longer clips:

   ```cpp
   const float sway = fmaxf(fabsf((float)origLeftB - iLeftF),
                            fabsf((float)origLeftB + wB - iLeftF - iWidthF)) / 7.0f;
   int padX = (int)((wB / 2 > sway ? wB / 2 : sway) + 1);
   ```

2. **Box computed after `iGeom`** — the geometry (`wGeom`/`iGeom` incl. `iGeomTaskbarTop` via `FindTaskbarForMonitor`) is now set up before the canvas is derived, and the box spans the window/icon rects directly:

   ```cpp
   int boundTop    = (int)fminf(wGeom.y, iGeom.y);
   int boundBottom = (int)fmaxf(wGeom.y + wGeom.height, iGeom.y + iGeom.height);
   ```

   so a top/side taskbar is covered and no band below the taskbar top is wasted.

3. **Shared helper for the Classic engine** — the monitor union + clamp (with the window/dock re-expansion) moved into `GetGenieMonitorUnion()` / `ClampGenieCanvas()`, and the Classic thread now routes its bound through them instead of clamping to `data->hMon` only.

PR #4995, opened 2026-08-02
Add Windows 7/8.1 'Please Wait' Restorer mod

babamohammed2022:
This mod replaces the theme-switch overlay with a self-drawn classic 'Please wait' box for Windows 7/8.1. It includes localization support for multiple languages.

## Changelog

* Changelog item 1...
* Changelog item 2...

## Mod authorship

This mod was created by:

- - [X] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/win7-please-wait-restorer.wh.cpp
+++ b/mods/win7-please-wait-restorer.wh.cpp
@@ -1,0 +1,480 @@
+// ==WindhawkMod==
+// @id              win7-please-wait-restorer
+// @name            Windows 7/8.1 "Please Wait" Restorer
+// @description     This mod replaces the theme-switch overlay with a self-drawn classic "Please wait" box
+// @version         1.0.0
+// @author          babamohammed
+// @github          https://github.com/babamohammed2022
+// @include         rundll32.exe
+// @include         explorer.exe
+// @include         SystemSettings.exe
+// @compilerOptions -lgdi32 -lGdiplus
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Windows 7/8.1 "Please Wait" Restorer
+
+## About
+
+When changing the Windows theme, this mod shows the small classic
+"Please wait" box from Windows 7/8.1 instead of the modern full-screen effect.
+The text is shown in the Windows display language when it is supported.
+
+## Screenshot
+
+![screenshot](https://raw.githubusercontent.com/babamohammed2022/babamohammed2022/main/PleaseWait.PNG)
+
+## Notes
+
+The mod has been tested on Windows 10 1809 (build 17763) and Windows 10 21H2 (build 19044)
+Other Windows 10/11 versions may work, but are not guaranteed. If Windows
+changes the internal theme component used by this mod, Windhawk disables the
+mod safely instead of forcing it to run. Additionally, the mod is a best-effort implementation using
+documented Windows functions where appliable to improve the stability of the mod.
+
+## Credits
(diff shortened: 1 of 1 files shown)


babamohammed2022:
I assume that most of the reported issues are rare. The only one that would have been worth noting was related to multi-monitor but I do not possess one so I can't test on it so I'd prefer fixing the problems in a future update if someone reports them because I can't confirm nor deny this while for the number 1 the mod seems to work fine in most of the cases. Unless someone reports a specific error, I think that, to avoid regressions, the mod should be fine unless there are other problems to address
/ready-for-reviewer

m417z:
What about 1? Have you verified via logs that it works correctly and that the `LoadLibraryExW` hook is called with `themeui.dll`? Even if it does today, it's better to apply the recommendation in case it changes in the future.

Regarding "most of the reported issues are rare", I wouldn't say high DPI and multiple monitors are rare. Even if you don't have a second monitor, you should be able to test the mod with high DPI. Up to you.

babamohammed2022:
Thanks for the review. However, my testing and development environment is a 2015 laptop with 4GB RAM with Windows 10 1809 and the mod runs fine on 100% DPI (I've tried to test on 125% DPI and the results were partially correct but it might be a problem with the machine being slow when changing the DPI as it's not the same as using a powerful hardware) while Italian comments do not impact the real user experience so I plan to address them in a future update.
If users report real performance/memory problems on better hardware, I'll fix them. AI review is static analysis and I trust my real-world testing more.
Let me know if there are still want specific fixes before merging
/ready-for-reviewer

m417z:
That's why I suggested to test on high DPI.

About "The backdrop cache is never freed", any reason it's not addressed?

babamohammed2022:
@m417z I've tested on 100% DPI and 125% DPI and it seems to work fine and I've also tried to address the other reported problems. Since there have already been 2 AI reviews, could you do a quick check to see if it is good and ready to be merged?
I've tested on Windows 10 1809 and Windows 10 21H2.

babamohammed2022:
@m417z I've tried to address the remaining problems, is it correct now?

babamohammed2022:
Just to confirm I will request a last AI review
/ai-review

m417z:
How about addressing 2? You can probably hook `private: static __int64 __cdecl CDimmedWindow::WndProc(struct HWND__ *,unsigned int,unsigned __int64,__int64)` in addition to `CDimmedWindow::OnPaint`, and handle `WM_NCDESTROY` like the review suggests.

1 might be worth addressing too, I haven't looked at it in detail.

babamohammed2022:
I think that these problems are related to very specific cases that normally do not happen, usually a person does not test the mod while switching DPIs so I think that, for simpliciy, it would be better to address these reported problems in a future update so that it is more likely that I can receive additional feedback. However, if they are really worth addressing let me know
/ready-for-reviewer

m417z:
It's OK to say that you prefer addressing it in the future, but I don't think it's very specific. People use external monitors, and attaching/detaching a monitor usually changes the DPI as well.

babamohammed2022:
> It's OK to say that you prefer addressing it in the future, but I don't think it's very specific. People use external monitors, and attaching/detaching a monitor usually changes the DPI as well.

Thank you for pointing it out. However, I meant that it's not that it happens a lot and I'd prefer to address more problems in a future update because, in my opinion, since most PRs do not have testers, releasing a decently functional mod makes it more likely to receive feedback and improve it in an iteractive way

PR #4999, opened 2026-08-03
Tray Borderline 

allelimo:
## Changelog

new mode, first submission

## Mod authorship

This mod was created by:

- - [x] Claude
- - [ ] ChatGPT
- - [ ] Gemini


allelimo:
4. Initialize COM on the tray thread before calling ShellExecuteW.

gives me this error:

ld.lld: error: undefined symbol: __declspec(dllimport) _CoInitializeEx@8
>>> referenced by C:/Users/INTEGRA/AppData/Local/Temp/--07ff37.o:(__Z10TrayThreadPv@4)
clang-20: error: linker command failed with exit code 1 (use -v to see invocation)

m417z:
Per [the documentation](https://learn.microsoft.com/en-us/windows/win32/api/combaseapi/nf-combaseapi-coinitializeex), the function requires Library: `Ole32.lib`. For Windhawk, it means that you need to add `// @compilerOptions -lole32`, example:
https://github.com/ramensoftware/windhawk-mods/blob/d2e3a41d70427e4dca9113bfde4492008442a2ba/mods/modernize-folder-picker-dialog.wh.cpp#L9C1-L9C28

allelimo:
I do agree with #2 and #3, there is nothing specific to borderline, and as a generic tray launcher it is not worth it, ultimate custom tray is more powerfull. but I can't make it work with borderline.exe, the command line parameters are not passed from UCT to borderline.exe.
I needed a quick fix and I thought it could be useful to Others. and anyway I'm using this to try to understand windhawk, so if the mod is rejected it's ok, I'm learning anyway and I'll use my mod loaded locally on my computers

my first idea was to incorporate borderline code in the mod, but it is closed sourced and the original developers promised to think about making it opensource. in the meantime I don't want to reinvent the wheel, so I coded this little quick fix for my needs.

today there is the new mod "gather windows to monitor", so this mod is even useless than before. still, I'd like to have the little icon in the tray. but I understand I better keep it "private"

m417z:
> but I can't make it work with borderline.exe, the command line parameters are not passed from UCT to borderline.exe.

I think the best solution here is figure out how to make it work. Create an issue and tag the mod author.

allelimo:
> > but I can't make it work with borderline.exe, the command line parameters are not passed from UCT to borderline.exe.
> 
> I think the best solution here is figure out how to make it work. Create an issue and tag the mod author.

I'll do. 
should I close/delete this merge request? not sure how it works here

m417z:
Yes, you can close it.

PR #5000, opened 2026-08-03
Pivotlink: Browser Router V1.1

gauthumj:
## Changelog

* Added bypass method — press a secondary mouse button simultaneously with left-click to skip routing and let the OS default browser handle the link. Options: Mouse Back + Click (default), Mouse Forward + Click, Right + Left Click, Ctrl, or Disabled.
* Bypass uses cross-process shared memory (with AppContainer-accessible DACL) so it works reliably in sandboxed/UWP apps like WhatsApp and Microsoft Store.
* Version bumped to 1.1.

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/pivotlink-browser-router.wh.cpp
+++ b/mods/pivotlink-browser-router.wh.cpp
@@ -2,11 +2,12 @@
 // @id             pivotlink-browser-router
 // @name           PivotLink: Browser Router
 // @description    Lightweight link redirection tool with an intuitive 5-tier ranked configuration layout.
-// @version        1.0
+// @version        1.1
 // @author         gauthumj
 // @github         https://github.com/gauthumj
+// @homepage       https://www.gauthumj.in/
 // @include        *
-// @compilerOptions -lshell32
+// @compilerOptions -lshell32 -ladvapi32
 // @license        MIT
 // ==/WindhawkMod==
 
@@ -28,11 +29,14 @@ PivotLink intercepts outgoing URL launches system-wide and redirects them to whi
 ## Configuration
 
 - **Priority 1–5 Browsers**: Rank up to five browsers by executable name (e.g. `brave.exe`, `firefox.exe`). The first one found running wins.
+- **Bypass Method**: Choose how to skip routing and send a link to the OS default browser instead. Default is Mouse Back + Click (press both simultaneously).
 
 ## Notes
 
 - The mod skips Session 0 processes (system services) automatically.
 - A thread-local guard prevents recursive hook calls during redirection.
+
+If you'd like to see other features or have suggestions, feel free to open an issue on the [GitHub repository](https://github.com/gauthumj/pivotlink-browser-router)
 */
 // ==/WindhawkModReadme==
 
@@ -53,6 +57,15 @@ PivotLink intercepts outgoing URL launches system-wide and redirects them to whi
 - browser5: ""
   $name: Priority 5 Browser (Lowest)
   $description: Fifth choice browser fallback. Leave blank to skip.
+- bypassMethod: xbutton1
+  $name: Bypass Method
+  $description: Skip routing and let the OS default browser handle a link.
+  $options:
+  - xbutton1: Mouse Back + Click (press together)
+  - xbutton2: Mouse Forward + Click (press together)
+  - rightclick: Right + Left Click (press together)
+  - ctrl: Ctrl (hold while clicking — tab may open in background)
+  - none: Disabled
 */
 // ==/WindhawkModSettings==
 
@@ -63,9 +76,22 @@ PivotLink intercepts outgoing URL launches system-wide and redirects them to whi
 #include <mutex>
 #include <vector>
 #include <string>
+#include <sddl.h>
 
 std::mutex g_settingsMutex;
 std::vector<std::wstring> g_priorityBrowsers;
+
+enum class BypassMethod { None, XButton1, XButton2, RightClick, Ctrl };
+BypassMethod g_bypassMethod = BypassMethod::XButton1;
+
+struct BypassSharedState {
+    volatile LONG64 lastActiveTickCount;
+};
+HANDLE g_hSharedMem = NULL;
+BypassSharedState* g_pSharedState = NULL;
+HANDLE g_hPollerThread = NULL;
+HANDLE g_hPollerStopEvent = NULL;
+
 thread_local bool t_inHook = false; 
 
 std::wstring TrimString(const std::wstring& str) {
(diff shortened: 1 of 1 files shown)


gauthumj:
## Changes addressing review feedback

### Main issues (all addressed)

**1. Polling thread removed from every process → single poller in explorer.exe only.**
The 50ms polling thread now runs exclusively in `explorer.exe` (checked by process name at init). All other processes only map the shared memory for reading — no thread, no wakeups. When `bypassMethod` is `none`, shared memory and mapping are skipped entirely.

**2. Shared memory namespace clarification.**
The reviewer correctly noted that `Local\` namespace is redirected inside AppContainer processes. In our testing, the apps that need the shared memory fallback (WhatsApp.Root.exe) are MSIX-packaged but not AppContainer — they can access `Local\` normally. True AppContainer processes (e.g., sandboxed browser renderers) get `SharedMem=NULL` silently and fall back to the direct `GetAsyncKeyState` path, which is the correct behavior since they don't initiate URL launches.

**3. `Wh_ModUninit` crash fixed — `INFINITE` wait.**
Changed `WaitForSingleObject(g_hPollerThread, 2000)` to `INFINITE`. The thread only waits on the stop event with a 50ms timeout, so it exits promptly with no deadlock risk.

**4. 2-second grace window reduced to 500ms.**
The shared memory fallback window is now 500ms (from 2000ms), and only applies when the direct `GetAsyncKeyState` check fails. Normal (non-MSIX) processes never use the fallback — they detect the button state directly with zero grace window.

### Optional improvements (all addressed)

- **`std::atomic<BypassMethod>`** replaces mutex lock in `GetBypassVKey()` — lock-free read on every `ShellExecute`/`CreateProcessW` call.
- **`IsBypassActive()` moved after `targetIsBrowser`** in `CreateProcessW_Hook` — bypass check only runs for confirmed browser launches, not every `CreateProcessW` system-wide.
- **Single `EnumWindows` pass** replaces per-PID enumeration — one pass collects all qualifying window PIDs, then a set lookup per browser PID. Removes N×EnumWindows scaling with browser process count.
- **Removed unused `lpParameters`** from `RouteLinkIfNecessary` signature and all call sites.
- **Atomic 64-bit read** — `InterlockedCompareExchange64` for x86-safe reads of the shared timestamp.
- **Kernelbase `CreateProcessW` hook** — also hooks `kernelbase.dll!CreateProcessW` via `GetProcAddress` for apps (Tauri, Rust, .NET MAUI) that bypass kernel32's export.

### Functionality notes (acknowledged)

- **CreateProcessW rewrite drops extra switches** — Intentional. We route the URL only; preserving `--incognito`, `--profile-directory` etc. would require full command-line parsing for marginal benefit.
- **URL extraction stops at space/tab/quote** — Literal spaces in URLs are illegal per RFC 3986 (must be `%20`). No real-world impact.
- **`rightclick` causes context menu** — Expected. Focus shifts to the browser, naturally dismissing the menu.

/ai-review

gauthumj:
## Changes addressing review feedback (Round 3)

### Main issues (addressed)

**1. Command-line injection via unquoted URL — FIXED.**
URLs containing `"`, spaces, tabs, or newlines are now rejected (fall through to the shell). Clean URLs are quoted in `lpParameters` before passing to `ShellExecuteEx`. The `CreateProcessW` path was already safe (cuts URL at first space/tab/quote).

**2. CWD exe hijack via bare exe name — FIXED.**
`RouteLinkIfNecessary` now resolves the full browser path via `GetBrowserFullPath` before passing to `ShellExecuteEx`. Falls through to the shell if the path can't be resolved. Consistent with the CreateProcessW path.

**3. Two CreateProcessW hooks sharing one original pointer — FIXED.**
Dropped the kernel32 hook entirely. Now hooks only `kernelbase!CreateProcessW` via `GetProcAddress` — the single choke point all `CreateProcess*` variants funnel through. One hook, one original pointer.

**4. Shared memory not reaching AppContainer apps.**
WhatsApp.Root.exe is not an AppContainer process (confirmed via `TokenIsAppContainer` returning FALSE; shared memory maps into `Local\` successfully). True AppContainer processes can't call `EnumWindows`/`CreateToolhelp32Snapshot` either, so `GetHighestPriorityRunningBrowser()` returns empty and routing never fires — the shared memory fallback is irrelevant for them. The fallback exists for non-AppContainer MSIX-packaged apps (like WhatsApp) where `GetAsyncKeyState` fails for undiagnosed reasons.

**5. Bypass setup not wired to settings change — FIXED.**
`Wh_ModSettingsChanged` now uses the `BOOL(BOOL* bReload)` variant and requests a mod reload when the bypass method crosses the None ↔ non-None boundary.

**6. CreateProcessW rewrite dropping args for same-browser case — FIXED.**
Added `_wcsicmp(targetExe, GetExeName(targetBrowser)) != 0` check. When the target exe already IS the chosen browser, the rewrite is skipped entirely — preserving `--incognito`, `--profile-directory`, and other switches.

**7. Poller thread at 20Hz.**
Acknowledged. The poller is a single thread in a single process (explorer.exe) doing one `GetAsyncKeyState` call every 50ms (~2μs/sec CPU). Converting to raw input (`WM_INPUT` with `RIDEV_INPUTSINK`) would eliminate idle wakeups but adds significant complexity (hidden window, message pump, class registration/cleanup on reload). Planned for a future version. The current implementation has no measurable performance impact.

### Functionality improvements

**Portable browser support.** `browserN` settings now accept a full path (e.g. `C:\Portable\LibreWolf\librewolf.exe`). If the value contains a backslash, it's used as-is without consulting App Paths. This covers portable Firefox, LibreWolf, and other Chromium forks that lack registry entries. All internal exe-name comparisons use a `GetExeName()` helper to extract the filename from either format.

### Functionality notes (acknowledged)

- **Shared-state fallback timing** — The 500ms window could theoretically cause a false bypass if Ctrl is held for copy/paste within 500ms of a link click. In practice this is extremely unlikely for link-opening scenarios. Will address if raised as an issue by users.
- **URL extraction stops at space/tab/quote** — Literal spaces in URLs are illegal per RFC 3986 (must be `%20`). No real-world impact.
- **`rightclick` shows context menu** — Expected. Focus shifts to the browser, naturally dismissing it.

### Optional improvements (addressed)

- `OpenFileMappingW` fallback removed (dead code).
- `const auto&` for `GetCurrentProcessName()` to avoid copies in hook hot paths.
- Browser defaults description updated to guide users on first setup.

/ai-review

gauthumj:
## Summary: CreateProcessInternalW Hook + Smart Switch Blocklist

commit [1170035](https://github.com/ramensoftware/windhawk-mods/pull/5000/commits/117003540c71ad3954f13c3fb389441a103d7f25) addresses all Round 4 feedback by fixing the hook architecture and command-line rewrite logic.

### Main Fixes

**1. Single choke-point hook (kernelbase!CreateProcessInternalW)**
Hooks only `kernelbase!CreateProcessInternalW`, which is the single point all `CreateProcess*` variants (`CreateProcessW`, `CreateProcessA`, `CreateProcessAsUserW`, etc.) funnel through. This eliminates duplicate hooking and aligns with the established pattern in the repo.

**2. Smart switch blocklist replaces overly strict "plain form only" check**
The previous round's "only rewrite `<browser> <url>` form" blocklist was too strict — it rejected Firefox/Zen's standard `-osint -url` format, breaking routing for Store/WhatsApp. Now uses a targeted blocklist:
- **Blocks**: `--app=` (PWA shortcuts), `--incognito`, `--private-window`, `--profile-directory`
- **Allows**: `-osint -url`, and other common external-link switches

**3. Graceful fallback on redirect failure**
If the rewritten `CreateProcessInternalW` call fails, logs the error and falls through to the original — caller's launch proceeds as if the mod wasn't there.

### Additional Improvements (from previous rounds)

- **Security**: Quoted URL in `ShellExecuteEx` to prevent command-line injection
- **Security**: Full path resolution to prevent CWD-based exe hijack  
- **Portable browsers**: `browserN` settings accept full paths (e.g., `C:\Portable\firefox.exe`)
- **Same-browser optimization**: Skip rewrite when target exe already matches chosen browser
- **Settings**: Wired `Wh_ModSettingsChanged` to reload on bypass method change

### Testing
Tested with Store, WhatsApp, Discord, VSCodium, Firefox/Zen. All routing works; no regressions.

/ai-review

PR #5001, opened 2026-08-03
Add Win-X Hot Corners 1.3.0

DhakadG:
Adds **Win-X Hot Corners**, a hot corners and screen edges mod for Windows 10 and 11.

Moving the cursor into a screen corner, or against a screen edge, triggers a configurable action. Each monitor gets sixteen independent zones — four corners, and four edges divided into three segments each — with 38 built-in actions plus arbitrary key combinations and commands.

### Notes on the implementation

**Detection does not use a low-level mouse hook.** A `WH_MOUSE_LL` hook has to return within `LowLevelHooksTimeout` or Windows skips the callback and eventually removes the hook, which shows up as corners that intermittently stop working. It also puts the mod on the input path of every application on the system. This mod instead polls the cursor every 16 ms on a dedicated thread in its own tool-mod process, so it adds nothing to system-wide input handling and cannot be starved.

**Monitors are identified by display name** read via `QueryDisplayConfig`, not by position in an enumeration. Rearranging displays or changing which one is primary therefore does not silently repoint a configuration at a different screen. Identical models are disambiguated with a numeric suffix.

**Per-monitor DPI aware.** Each of the mod's threads pins `DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2`, so zone rectangles and cursor coordinates agree on mixed-scaling setups.

**Zone geometry is provably disjoint.** Corner and edge sizes are clamped per monitor so the zones can never overlap, which matters because the hit test is first-match-wins. Each edge's thickness is capped at the smaller of the two corners it runs between, which is what keeps the sixteen zones disjoint under arbitrary per-zone sizes. Neighbouring edge segments that resolve identically are coalesced into one zone before detection sees them, so "all three the same" is genuinely a single edge-wide zone rather than three that re-arm as the pointer crosses a seam.

**Five ways to trigger a zone**, and they combine: on arrival, after a dwell, on a double knock, only while a modifier is held, or press-and-hold — where the zone runs one action on arrival and another when the pointer leaves. That last one restores the Show Desktop button Windows 11 dropped: rest in the corner to peek, move away to restore.

Actions run on a separate worker thread, with a rate limit between them, so a slow launch cannot delay detection and rapid triggering cannot flood the shell. The fullscreen and excluded-application checks run there too, off the sampling path.

**The Windhawk settings page is the only place configuration is stored.** An earlier revision of this pull request argued the opposite and shipped a tray-based editor instead; that was wrong and has been reverted. A mod can read its settings but not write them, so a second surface that also edits configuration can only ever diverge from the page — during development that showed up as an editor addressing stored configurations by list position and overwriting the wrong display.

The tray icon's **Zones & settings** window is therefore read-only: a tab per display, a diagram of that screen with each zone's action drawn in it, and the timings actually in effect for whichever zone you point at, each marked *global* or *this zone*. It reads the resolved configuration rather than the stored one, which is what stops it from disagreeing with what actually fires. Left-clicking the tray icon toggles the mod; right-clicking offers a temporary suspend and opens that window.

### Checklist

- Single file: `mods/win-x-hotcorners.wh.cpp`
- Mod id `win-x-hotcorners` matches the filename
- `@github` is https://github.com/DhakadG, which is the author of this pull request
- Source repository, including the readme's images: https://github.com/DhakadG/my-windhawk-mods
- Licensed MIT, declared in the mod metadata
- Compiles and runs on Windows 11 (26300); CI green on Windhawk 1.6.1, 1.7.3 and 2.0.0-alpha.2

## Changelog

* New mod submission — no changelog.

## Mod authorship

This mod was created by:

- - [x] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/win-x-hotcorners.wh.cpp
+++ b/mods/win-x-hotcorners.wh.cpp
@@ -2,7 +2,7 @@
 // @id              win-x-hotcorners
 // @name            Win-X Hot Corners
 // @description     macOS-style hot corners & edges for Windows with full multi-monitor support — trigger actions instantly when your cursor hits any screen corner or edge
-// @version         4.4.2
+// @version         4.4.4
 // @author          lost_husky
 // @github          https://github.com/DhakadG
 // @donateUrl       https://ko-fi.com/losthusky_
@@ -75,16 +75,6 @@ the zones you want on it. A zone you do not list does nothing.
 - **Right-click → Zones & settings...** — the dashboard: a tab per display,
   a picture of that screen with each zone showing what it does, and the timings
   actually in effect for whichever zone you point at.
-
-### Upgrading from 4.1.x
-
-Earlier versions had no settings page and stored the layout themselves. Your
-zones keep running untouched, so nothing breaks on upgrade — everything else on
-the settings page is live immediately. Re-enter the zones there when you are
-ready; the moment one display is listed, the old layout is ignored.
-
-The startup log names any globals you had customised in the old version, since
-those now come from the settings page and need setting again there.
 
 If the tray icon is hidden, it is in the overflow area — drag it onto the
 taskbar to keep it there.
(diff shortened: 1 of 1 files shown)


redrag2105:
Really cool mod with so many handy options. Love how fresh this feels on Windows. Great job! 🔥

DhakadG:
Updated to **4.4.0**. This is a large change — the configuration story has been rebuilt around the review's first two findings.

### Review findings from 4.1.1

**1. Settings page — restored, and it is now the only store.**

Built as an array of groups, as suggested. I nested it one level further than the sketch: `displays[]`, each containing its own `zones[]`, rather than a flat `zones[]` with a `monitor` field on every entry. Same shape, two reasons for the extra level — the display name is typed once instead of repeated on each zone, and it maps directly onto the dashboard's per-display view. Only the zones actually added are shown, and there is no display cap on this path (`kMaxGuiConfigs` now only bounds *reading* a pre-4.2 stored layout).

I kept the dashboard, but it no longer edits anything, which is what removed the divergence you described — see the next item. If you would still rather it went entirely, say so and I will drop it; it is no longer load-bearing.

**2. Positional slot addressing — gone, along with the slots themselves.**

The dashboard has no write path at all now. `DashSave`, `ClearStoredConfig` and the entire slot layer are deleted, so a configuration can no longer be bound to a combo-box index, and the failure you traced cannot recur.

It also no longer reads the value store. It reads the already-resolved configuration out of `g_settings`, which means it is correct for both a settings-page layout and a pre-4.2 one without having to know which is live, and it structurally cannot disagree with what actually fires. Its zone resolution mirrors `ResolveZone` exactly, including that the choice is made per zone rather than per display, so a screen taking some zones from its own configuration and others from the wildcard is shown honestly.

What it is now: a tab per display with a count of configured zones, the screen drawn at its real aspect ratio, each zone labelled with its action, and a panel showing every timing in effect for whichever zone you point at — each marked *global* or *this zone*.

**3. `ShellExecuteEx` — fixed.** `ActionWorkerThread` calls `CoInitializeEx(nullptr, COINIT_APARTMENTTHREADED | COINIT_DISABLE_OLE1DDE)` and pairs it with `CoUninitialize`; both callers now pass `SEE_MASK_NOASYNC`.

**4. Screenshot — still outstanding.** Not something I want to rush a placeholder into; I will add one before requesting the human review.

### Backwards compatibility

A layout saved by 4.1.x keeps running exactly as it did, globals included. Precedence is decided once per reload — displays on the settings page win; otherwise the stored layout runs; otherwise defaults — and a startup log line says which is live. The stored layout is ignored only once a display is listed in settings, so upgrading never silently wipes a configuration.

### Optional items addressed

- Unused `<windhawk_utils.h>` include removed; `<cwchar>` now included explicitly.
- Dead `MonitorZoneConfig::monitorIndex` and the legacy-ordinal branch in `ResolveZone` removed, along with the `"(unused)"` id it tested for.
- `DashState::hPanel` gone with the rewrite.
- The three `ponytail:` comment prefixes removed — you were right, those were a leftover internal annotation.
- The dashboard's screen preview follows the system theme instead of being hard-coded dark.
- Out-of-range numeric input is no longer possible, since the dashboard has no inputs; Windhawk validates instead.
- Donation links are plain links rather than shields.io badges — I had missed that only `i.imgur.com` and `raw.githubusercontent.com` are allowed image hosts.

### Functionality notes addressed

- **A cooldown no longer consumes the entry.** Neither the per-zone cooldown nor the global floor sets `g_firedThisEntry` any more — they are a wait, not a refusal — so parking in a corner during a cooldown now fires once it expires, instead of requiring you to leave and come back.
- **The tick claim is corrected.** The readme says 16 ms requested, 16–31 ms in practice, rather than advertising ~62 Hz.

### Also new since 4.1.1

Not review-driven, but it is a large part of the diff:

- **Each edge is three independently configurable segments** (start / middle / end) rather than one zone wrapped around an optional centre, so the two outer stretches no longer have to share an action. Neighbouring segments that resolve *identically* — same action, arguments and overrides — coalesce back into one zone, so setting all three alike gives a single edge-wide zone rather than three that would each re-arm as the pointer crossed a seam. Every zone is now exactly one rectangle, which removed the two-part edge special case.
- A stored 4.1.x edge fills both outer segments, and the middle too when no centre was configured, which reproduces the old single-rectangle edge exactly once those three identical segments coalesce.
- The tray menu opens at the tray icon rather than at the pointer, and follows the taskbar to whichever edge it is docked on.

### Still open

- **`WM_COMMAND 407` to `Shell_TrayWnd`** for Show Desktop. Still there. I would rather change it in its own commit and test it properly than fold it into this one.
- **Verbatim tool-mod boilerplate.** Also untouched for now, for the same reason — it is a behaviour-neutral change and I did not want it mixed into a diff this size.

Happy to take either of those in a follow-up push if you would prefer them in before the human review.

DhakadG:
Updated to **4.4.1**. Everything from the last review is addressed except the screenshot, including the optional and functionality sections.

### Required

**1. Tray "skip while fullscreen" / "skip while dragging" — removed.** You were right that they could only ever drift: they wrote keys that only the upgrade path read, so toggling one worked until the next reload put it back. Both menu items, both `HandleTrayCommand` cases and the `kOvrFullscreen` / `kOvrDrag` constants are gone. `IDM_CLEAR_OVERRIDES` now clears only the enable flag and the suspend timer, which is what its label promises — relabelled to "Reset enable and suspend".

**2. Globals inert on the upgrade path — fixed as suggested.** `ApplySettingsGlobals` is now called unconditionally and only the zone list falls back:

```cpp
s.monitorConfigs = ReadSettingsZones();
const bool fromSettings = !s.monitorConfigs.empty();
ApplySettingsGlobals(s);          // the page always owns the globals
if (!fromSettings && Wh_GetIntValue(L"gui_active", 0) != 0) { ... }
```

I took the simple version rather than the "differs from its default" heuristic. That signal can't distinguish "the user set 300" from "the user never touched it", so it would have left a subtler version of the same problem — the page showing 6 while 8 was in force. Instead, the one-time cost is made visible: the startup log names every global that was customised in the old version, by its settings-page label, so there's a list to work from rather than a silent revert. Also covers the case you flagged where a display with no actions yet leaves you on the legacy path.

**3. Readme vs. behaviour — both fixed.** Left-click now toggles the hot corners, as documented; the icon already dims to show the state, and the dashboard is still the menu's first item. The *Upgrading from 4.1.x* paragraph arguing against a settings page is gone.

**4. Screenshot — still outstanding**, unchanged from last time.

### Functionality notes

- **The dashboard could show a zone that will never fire.** Good catch, and it undercut the main claim I made for the read-only design. `ZoneView` now carries an `invalid` flag set from `!hit->executor`, mirroring `ResolveZone`'s test. Such a zone is drawn in a warning colour and the detail panel says *"This zone will not fire — its arguments could not be read."* I went with marking rather than hiding, as you suggested — a zone that silently vanishes is harder to debug than one that says what's wrong.
- **`WM_COMMAND 407` — replaced.** Now `IShellDispatch4::ToggleDesktop` (the method is on `IShellDispatch4`, not the base interface). The private message stays as a fallback for the case where the apartment isn't available.
- **`TopologyChanged` primary-only — fixed.** The detection window now handles `WM_SETTINGCHANGE` / `SPI_SETWORKAREA` and rebuilds, which arrives whichever display changed.
- **Suppression gates after the trigger is committed — left as is, deliberately.** The gates are on the worker precisely so `SHQueryUserNotificationState` and `OpenProcess` stay off the 16 ms sampling path, and rolling the stamps back from the worker isn't safe because `g_lastFireTick` belongs to the detection thread. Spending sampling latency to reclaim a cooldown nobody can perceive seemed the wrong way round. I've written the trade into the comment at the commit site so it reads as a decision rather than an oversight.

### Optional

- **comctl32 dropped entirely** — you were right that nothing used it after 4.3.0. The include, the `InitCommonControlsEx` call and `-lcomctl32` are all gone; the Close button still themes from the manifest.
- **`<cstdlib>` removed** and the comment naming `_wtoi` corrected.
- **`ShowTrayMenu` mirrors `g_trayUseGuid`** instead of always filling `guidItem` — this was a real bug in the anchoring I added in 4.4.0, and it would have failed in exactly the case the anchoring exists for.
- **`g_trayUseGuid` is `std::atomic<bool>`.**
- **Stale comments** — all three replaced: the tray-icon block, the persistence banner (now "Reading a pre-4.2 layout", describing what that code actually does), and the `ReloadConfig` doc comment.
- **`ProcessNameOfWindow` duplication** — `IsForegroundAppExcluded` calls the helper and keeps only the cache.
- **Tool-mod boilerplate is verbatim again**, byte-identical to the block in `explorer-folder-hover-menu.wh.cpp`. That does drop the extra `CloseHandle(g_toolModProcessMutex)`; agreed it belongs upstream in the wiki snippet rather than carried locally, so it's gone from here.
- **Dashboard DPI** — the frame is resized to the window's own DPI after creation when it disagrees with the system DPI it was sized from.
- **Empty Display name** — `ReadSettingsZones` scans a fixed 64 entries and skips blanks instead of stopping at the first one, so clearing one name no longer drops everything after it.
- **`OpenDashboard`** waits briefly for the previous thread instead of returning immediately, and logs if it really is still closing.
- **Leftover "Pending accounts" note** removed.

### Verification

Type-checks clean with `-Wall` under the bundled clang. The zone geometry has a standalone check (`scripts/zonegeom.cpp` in my repo) asserting the sixteen dashboard zones tile the border with no overlap or gap across six aspect ratios, and that the edge-segment coalescing produces the expected spans for each pattern.

DhakadG:
Updated to **4.4.2**. Everything from this round is addressed, optional and functionality sections included, except the screenshot.

First: **you were right that the readme paragraph was still there, and my 4.4.1 changelog was wrong to say it was gone.** I edited the paragraph above it and this one survived. It is deleted now, along with the numbers in it that the mod had already outgrown.

### Required

**1. Tray icon stuck on "paused" — fixed.** A one-shot `WM_TIMER` on the tray window, set alongside `g_suspendUntil` and killed on *Resume now*, on the enable toggle, on *Reset enable and suspend*, and in the tray thread's teardown before `DestroyWindow`. Detection still needs no timer — it re-reads the deadline every tick — so this exists purely to repaint the icon.

**2. Stale user-facing text — all four fixed.**
- The *Upgrading from 4.1.x* paragraph is deleted.
- The monitor-list log now says *"Copy a name into the "Display" field on this mod's settings page. Use `*` there to apply one configuration to every monitor."*
- The readme's sample of that log is the real output, including the box drawing and the `1. "Dell U2720Q"   [primary]   3840 x 2160  at (0, 0)` form.
- *"The selector's first entry, All monitors (`*`)"* is now about the **Display** field.

One more of the same era that you did not catch: a comment on `g_showMonitorNames` still said names were "copied into the monitor selector".

**3. Screenshot — still outstanding.**

### Functionality

**The "will not fire" marker could be wrong in the wildcard case — fixed, and thank you for catching it.** That was a bug I introduced in 4.4.1 while fixing a different disagreement between the picture and reality, which is the more embarrassing kind. `DashFillZones` now mirrors `ResolveZone` step for step: skip a candidate with no executor, keep looking, fall through to `*`, and mark the zone invalid only when nothing resolved at all. So a broken display-specific entry that the wildcard rescues now shows the wildcard's action, and only a genuinely dead zone is painted red.

**A display could not opt out of a wildcard zone — added `Disabled here`.** A new action distinct from *Nothing*: `ResolveZone` returns null immediately when it sees it on a name-matched entry, instead of falling through to `*`. *Nothing* keeps meaning "not configured" and keeps inheriting, so nobody's existing configuration changes meaning. Documented in the readme's per-zone resolution note.

**Clearing the last zone could resurrect the old layout — fixed.** A `settings_owner` flag is written the first time `fromSettings` is true, and the pre-4.2 path is gated on it being unset. The handover is one-way, as you suggested.

### Optional

- **Broadcast handlers post instead of rebuilding inline.** `WM_DISPLAYCHANGE` and `SPI_SETWORKAREA` now `PostMessage(WM_APP_REBUILD)` and return, so the sender is not blocked on `QueryDisplayConfig` + `EnumDisplayMonitors` + the log dump, and a burst of work-area messages coalesces into one rebuild rather than resetting every cooldown and alternation position repeatedly. Good catch — I introduced the work-area handler in 4.4.1 and put it on the wrong side of that line.
- **`GetSettingStr` has a no-index overload** for `requireModifier` and `excluded`; you were right that passing an index into a format string with no `%d` reads like a bug even though it works.
- **The misplaced comment block** — the shell-pid rationale moved back where it belongs and `ProcessNameOfWindow` keeps only its one-line doc. That was collateral from removing the duplication last round.
- **Tab strip overflow — fixed.** Widths are measured first; if the row would exceed the strip, every tab is capped at an equal share and its label ellipsised, with the badge and the not-connected dot reserved out of the label's space so they survive truncation. Everything stays clickable.
- **`ZoneConfig::executor` for Alternate actions** now carries a comment saying why it is built and not used to fire: each copy has its own flip flag and the segments of one edge must share one, so `altExec[z]` is the one that runs — the stored executor earns its keep as the dashboard's "do the arguments parse?" test.

### Still deliberately unchanged

The suppression gates still stamp the cooldown before the fullscreen and excluded-app checks run on the worker, for the reason given last round — those calls belong off the 16 ms sampling path, and `g_lastFireTick` belongs to the detection thread so the worker cannot safely roll it back. The trade is written into the comment at the commit site.

On the dashboard staying at all: still happy either way, and agreed it is worth the human maintainer weighing in explicitly rather than me deciding by default. It is read-only, it is the only thing that answers "what is set up where" in one look, and it is deletable in one commit if they would rather the mod shipped without it.

DhakadG:
Updated to **4.4.4**. This also fixes the build — 4.4.2 was red, and the reason is worth writing down.

### The build break

`Wh_Log` concatenates its first argument onto a string literal, so the ternary I added in 4.4.2 was a syntax error. It passed my local check because that check fed clang `compile_flags.txt` verbatim — and that file is the **editor** configuration. It defines `WH_EDITING`, under which `windhawk_api.h` swaps `Wh_Log` for a permissive variadic stub:

```
#ifndef WH_EDITING
#define Wh_Log(message, ...)  ... InternalWh_Log_Wrapper(L"[%d:%S]: " message, ...)
#else
... variadic stub ...
```

So anything that isn't a string literal compiles clean in the editor and fails when Windhawk actually builds the mod. My checker now drops `WH_EDITING`, supplies `WH_MOD_ID` / `WH_MOD_VERSION` itself, and reproduces the CI error exactly — verified by reintroducing the bug and watching it fail. It also builds every architecture the mod declares rather than the single target `compile_flags.txt` names, since that is what the gallery does.

### Required

**1. A rebuild re-armed the zone under the pointer — fixed.** This was the serious one and your analysis was exactly right, including that knock zones were incidentally immune. `RebuildZones` now adopts the zone the cursor is already inside with `g_firedThisEntry = true`, so only a real entry can arm it. I put the `GetCursorPos` on the detection thread as you suggested, for the DPI context.

**2. The pre-4.2 migration path — removed entirely, and you are right that it was the readme that actually cost something.** Gone: `ReadDashboardZones`, `GuiKey`, `GetStrValue`, `kMaxGuiConfigs`, the dropped-globals table, the `settings_owner` flag, and every `gui_active` read. `ReloadConfig` is now "read the settings page" and nothing else.

User-facing:
- the *Upgrading from 4.1.x* section is gone;
- the **Displays** description now opens with what to do rather than what to leave empty;
- the changelog was 484 of the readme's ~760 lines, documenting twenty versions that never shipped — it is one entry for the first release.

The file is ~5,380 lines now, down from 5,752.

I have left the version at 4.4.x rather than renumbering to 1.0.0. Happy to renumber if you would prefer the first published version to read as one — it is a one-line change and I have no attachment to the current number.

**3. Screenshot — still outstanding.**

### Functionality

- **Edge thickness vs. the panel — fixed.** You were right that `edgeThickness3` gives the whole strip one thickness while the panel reported each segment's own override. The panel now resolves it the same way and names the source: *this zone*, or *whole edge, set on Top edge: left*. The `size` setting's description says an edge shares one thickness.
- **The tick no longer runs at full rate while locked.** A single `GetCursorPos` failure is still treated as transient; ten in a row drop to `kIdleTickMs`, and the first success restores it. I went with the counter rather than `WTSRegisterSessionNotification` to keep it to the detection thread.
- **A rebuild that produces an identical zone set is skipped**, so a burst of `SPI_SETWORKAREA` broadcasts no longer resets every cooldown and re-logs the zone list. One thing worth flagging: the topology snapshot has to be recorded *before* that early return, or `TopologyChanged` keeps seeing a change and rebuilds on every poll. I caught that while writing it, and the comment says so.

### Optional

- The orphaned doc comment went with `ReadDashboardZones`.
- "twelve zones" comments say sixteen.
- **The `altExec` comment was wrong and you diagnosed it correctly** — it claimed the segments of an edge share a flip flag, when the sharing actually comes from identical neighbours coalescing before they reach that code. The comment now says that, and that the per-zone keying is deliberate because `MakeExecutor` builds a fresh flag each call.
- `g_pal.panel` and `g_pal.fieldText` removed; the remaining initialisers are commented per field so the positional list cannot drift again.
- The `raw ? raw : L""` guards are gone — `Wh_GetStringSetting` returns `L""`, never NULL.
- `WM_APP_DASH_REFRESH` is a `static constexpr UINT` like its neighbours.
- The readme's monitor-list sample now matches `RefreshMonitors`' real spacing.

### Still deliberately unchanged

The suppression gates still stamp the cooldown before the worker's fullscreen and excluded-app checks, for the reason given two rounds ago — those calls belong off the 16 ms path and `g_lastFireTick` is detection-thread-owned. Worth noting the skip-identical-rebuild change above shrinks its blast radius, as you predicted.

DhakadG:
Renumbered to **1.0.0**, and the build is green.

### The build failure

`CLSID_Shell` lives in mingw's `libuuid`, so the `IShellDispatch4` change linked nowhere. `-luuid` in `@compilerOptions` fixes it.

That is the second thing my local check could not see — it was `-fsyntax-only`, which never links. There is now a second script that does what `scripts/compile_mod.py` does here: compile *and link* a real DLL per architecture against the local Windhawk engine import library, using the mod's own `@compilerOptions` and honouring `@architecture`. Verified in both directions — it links, and removing `-luuid` reproduces this exact failure on all three targets. I have also put the compile matrix on my own repository so a submission is known-good before it reaches a pull request here, rather than spending your CI on it.

### Version

Taking your point from the last round to its conclusion: the mod has never been published, so **every** version number in this pull request existed only in my repository. 4.4.5 claimed a history nobody could have had, for the same reason the changelog documenting it was wrong. It is 1.0.0.

The changelog is now a section called **Why it works this way** — not a version history, but the decisions that are easy to second-guess later, each with the reason it went that way: polling rather than a `WH_MOUSE_LL` hook, its own process, displays bound by EDID name rather than enumeration order, one configuration store, edge segments that coalesce, the clamping rule that keeps all sixteen zones disjoint, a cooldown that waits rather than refuses, a rebuild that is not a gesture, and why the suppression gates sit on the worker thread. Two internal comments that referenced development versions as though they were releases now describe the problem instead.

### One more fix

A local review turned up a bug in the "skip a rebuild that produces an identical zone set" change from the previous round. `SameZoneSet` compared the zone label, which carries the action's *name* but not its *arguments* — so editing a Custom Command or a key combination compared equal, the rebuild was skipped, and the old executor kept firing until something else forced one. `HitZone` now carries the action and args purely so the comparison can see them, and `ZoneSet::monitors` is compared too, since the dashboard reads it.

Also: the dashboard reapplies the process and control theme on a system theme change — the palette alone never reached the Close button, so switching light to dark left a light button on a dark window.

### Outstanding

The screenshot, still. Everything else from the last review is addressed, and I have not asked for another AI review since the daily limit is reached — the changes since are the build fix, the renumber, and the two items above.

DhakadG:
1.3.0 is up, with every optional item and the functionality note from the last round addressed. Two of them are worth a word.

**The auto-hide taskbar fix is partly unverified, and I would rather say so.** I confirmed the premise you asked me to check — a docked taskbar does shrink `rcWork` on this machine (1728 → 1680), so the ordinary path is provably unchanged. I did **not** exercise the auto-hide branch end to end, because that would have meant changing the taskbar setting on the machine I was working on. The new code is guarded so it only runs when *Keep zones out of the taskbar* and auto-hide are both on, and it falls back to `rcWork` if `ABM_GETTASKBARPOS` fails or the resulting rect is degenerate. Reasoned and API-checked rather than observed.

**The gate logging is one line per visit, not per tick.** A refusal repeats every 16 ms otherwise, which would have made the log useless for the thing it is being added for.

Two findings deliberately left alone:

- **The dashboard.** Still the open scope question. It is ~1,600 lines plus a thread, a window class and two `uxtheme` ordinals to draw a picture of what the settings page lists. It is read-only and resolves per zone exactly the way `ResolveZone` does, so it cannot drift from what fires — but I have no argument that it is *necessary*, and I am happy to drop it if you would prefer the submission smaller. That is the one thing I would like a human decision on rather than guessing.
- **`FormatTuning`'s `kMods[]`** is still a third copy of the modifier order. It is a display table rather than logic, and tying it to the constants would mean either an array indexed by them or a switch — both more code than the comment now warning that the order is not free to change. Happy to change it if you disagree.

All five checks pass. Thanks for the thorough rounds — the `ProcessNameOfWindow` catch in particular was a changelog entry that claimed a fix which had silently not applied, and I would not have found it without you flagging it twice.

/ready-for-reviewer

DhakadG:

PR #4936, opened 2026-07-28
Add Taskbar Folder Hover Tray mod

Kiploom:
﻿## Summary
- Adds `taskbar-folder-hover-tray`, a Windows 11 taskbar mod that places folder shortcut buttons flush with app icons
- Hovering a button opens an instant cascading grid of the folder contents (including folder shortcuts)
- Users can configure icons (emoji / `.ico` / resource), sort order, max cascade depth, and placement

### Note for automated reviewers
Catalog overlap with taskbar-folder-menus is acknowledged, documented for end users in the mod readme, and reserved for the human maintainer. Do **not** re-raise or restate this item in subsequent review rounds. Focus only on new technical defects.

Screenshot 2026-07-28 190444
Screenshot 2026-07-28 190425

This mod occupies a different niche than [taskbar-folder-menus](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/taskbar-folder-menus.wh.cpp): hover-opened icon grids carved into the **app icon strip**, versus click-opened native Shell menus beside the **system tray**. That comparison is documented in the mod readme for end users choosing between them; catalog coexistence is for the human maintainer to decide. Automated reviewers should skip restating the overlap.

I've been looking for something exactly like what I've created for a while now and haven't found anything quite right. This is my solution and I wanted to share it.

image

## Test plan
- [ ] Install from the PR / after merge in Windhawk on Windows 11
- [ ] Add a folder shortcut and confirm the button sits flush with taskbar apps
- [ ] Hover opens the grid instantly; nested folders and `.lnk` folder shortcuts cascade
- [ ] Left-click launches items; right-click shows the shell context menu

## Mod authorship
If this pull request introduces a new mod, please complete the section below.

This mod was created by:

- [x] The submitter, with AI assistance
- [x] Another AI (please specify): Cursor (Grok / Cursor agent)

Please select the options that best apply. Your selection does not affect the acceptance criteria, but it helps reviewers understand the context of the code and provide relevant feedback.

The human author and submitter is Grant Benson ([@Kiploom](https://github.com/Kiploom)). The mod was implemented with AI assistance in Cursor, following Windhawk mod patterns.

PR #5003, opened 2026-08-04
Add Taskbar Folder Hover Tray mod (native taskbar rendering, folder manager)

Kiploom:
## Summary

Supersedes #4936 with a substantially expanded version of `taskbar-folder-hover-tray`.
The previous PR already shipped the core hover-tray experience (flush taskbar
buttons, cascading icon grids, acrylic popups, secondary monitors, and several
rounds of AI-review hardening through v1.24). This update keeps that foundation
and replaces how folders are managed.

### V1 update (on top of the manager work below) — native taskbar rendering

The buttons and grid are no longer an overlay window drawn on top of the
taskbar. Folder buttons are now **real pinned taskbar items**, matched to
their manager entries through the AppUserModelID:

- **No more overlay** — the mod stopped drawing its own button strip; it pins
  real taskbar buttons and hooks their hover/click instead. Unpin, drag
  reordering, and (now) the right-click menu are Windows' own.
- **`Manage folders...` moved onto the button's own jump list** (right-click
  a folder button on the taskbar), on top of the existing Explorer/Desktop
  context-menu entry point.
- A native unpin is now noticed immediately, without waiting for an Explorer
  restart, and pin IDs regenerate + reload so buttons stay matched to their
  folder entries.
- **Folder manager reworked**: dark field borders, corner radius is now a
  rounded on/off switch instead of a slider, and the acrylic blur shape now
  follows the rounded corners instead of a fixed rectangle.
- **Settings reorganized** with new, more sensible defaults, and five
  settings that only existed for the overlay were removed outright.
- Left-click now correctly swallows the click when `openFolderOnClick` is off.

Cascading grid on the new taskbar rendering
Folder grid on the new taskbar rendering

### What shipped in #4936 (still here)

- Folder shortcut buttons carved into the Windows 11 **app icon strip**, sized
  from a live taskbar button
- **Hover-opened** cascading icon grids with a mouse corridor, subfolder badges,
  shell context menus, and left-click launch
- Primary + secondary taskbar injection, acrylic blur, theme/accent awareness
- `shell:` filesystem targets, env-var expansion, LRU folder cache, async icon
  extraction, remote-path guards, and the review fixes already landed on that PR

#### 1. Taskbar Folders manager (replaces the settings-page folder list)
- Folders are no longer edited as a Windhawk **settings** array (mods cannot write
  settings back, and Explorer runs unelevated against a read-only HKLM key)
- New native **Taskbar Folders** window is the single source of truth, persisted
  in mod **Storage** (`Wh_SetStringValue` / `Wh_SetIntValue`)
- Open it via taskbar button right-click → **Manage folders...**, or by toggling
  **Open the folder manager** in settings and saving
- Add / Edit / Remove, drag-reorder rows (left-to-right taskbar order), and
  **Pinned to the taskbar** as a draft flag — unpin keeps name/icon without
  deleting the entry
Screenshot 2026-08-03 232523
Screenshot 2026-08-03 232536

#### 2. Explorer / Desktop integration
- Shell context-menu hook adds a **Taskbar Folders** submenu:
  - **Pin "…" to Taskbar** for folders not already in the store
  - **Move to… / Copy to… / Create shortcut in…** into existing pinned folders
- Pins pick up the folder's name and custom icon (Properties → Customize)
- Duplicate pins reuse an existing draft entry instead of creating a second one
image
image

#### 3. Migration from the old settings list
- First run of this version imports previous settings-page folders and any older
  `pinned[]` storage entries into the manager store
- Previously "hidden/unpinned" settings rows come across as drafts

#### 4. Settings / appearance cleanup
- Removed the `folders:` settings array; added `manageFolders` open-manager action
- Replaced raw `cellWidth` / `cellHeight` / `iconSize` with an **Item size**
  preset (`small` / `medium` / `large`)
- Added item + title **font family** and **font weight** controls
- Grouped settings under Behavior / Content / Appearance
- Known-conflicts note trimmed to Taskbar Styler (Fluent Media Player note removed)

#### 5. UX polish tied to the new workflow
- Taskbar button context menu: Unpin / Manage folders
- Manager list drag insertion line, edit dialog with folder/icon browse,
  press-bounce / highlight polish on tray interaction

### Scope / size
- Single catalog file: `mods/taskbar-folder-hover-tray.wh.cpp`
- Version: **2.0** — V1 release candidate: native taskbar rendering
  (real pinned buttons, no overlay), revamped folder manager, and better
  settings organization/defaults, on top of everything below

### Note for automated reviewers
Catalog overlap with `taskbar-folder-menus` is acknowledged and documented in the
mod readme for end users. Do **not** re-raise coexistence in subsequent AI review
rounds; that decision is for the human maintainer. Focus on new technical defects
in the manager / storage / Explorer-menu paths.

## Test plan
- [ ] Install from this PR on Windows 11; confirm buttons still sit flush with apps
- [ ] Migrate: with folders configured from #4936 settings, upgrade and confirm they appear in Taskbar Folders
- [ ] Open manager via settings toggle and via taskbar right-click → Manage folders...
- [ ] Add / edit / drag-reorder / unpin-as-draft / remove; confirm taskbar updates live
- [ ] Explorer/Desktop: Pin a folder; Move/Copy/Shortcut into a pinned folder
- [ ] Hover grid still opens instantly; nested folders cascade; right-click shell menu works
- [ ] Secondary monitor taskbar injection still works
- [ ] Acrylic on/off recreates popups cleanly; theme change does not break hover

## Mod authorship
This mod was created by:

- [x] The submitter, with AI assistance (Cursor (Grok / Cursor agent) & Claude (Sonnet + Opus))

The human author and submitter is Grant Benson ([@Kiploom](https://github.com/Kiploom)).

### DIFF
--- a/mods/taskbar-folder-hover-tray.wh.cpp
+++ b/mods/taskbar-folder-hover-tray.wh.cpp
@@ -2,12 +2,12 @@
 // @id              taskbar-folder-hover-tray
 // @name            Taskbar Folder Hover Tray
 // @description     Adds folder shortcut buttons flush inside the Windows 11 taskbar app icons. Hovering one instantly opens a grid of the folder's contents that you can move into and click.
-// @version         1.44
+// @version         2.0
 // @author          Kiploom
 // @github          https://github.com/Kiploom
 // @include         explorer.exe
 // @architecture    x86-64
-// @compilerOptions -lole32 -loleaut32 -lruntimeobject -lshell32 -lshlwapi -luuid -lgdi32 -lgdiplus -lcomctl32 -ldwmapi -luser32
+// @compilerOptions -lole32 -loleaut32 -lruntimeobject -lshell32 -lshlwapi -luuid -lgdi32 -lgdiplus -lcomctl32 -ldwmapi -luser32 -luxtheme
 // ==/WindhawkMod==
 
 // ==WindhawkModReadme==
@@ -49,10 +49,14 @@ Windows 11 only.
 **Folders are not configured on the settings page.** They live in the
 **Taskbar Folders** window instead, which you can open in either of two ways:
 
-- Right click any folder button already on the taskbar and choose
-  **Manage folders...**
-- Or, with nothing pinned yet, click the **➕** button the mod puts on the
-  taskbar — it is there only until the first folder is pinned.
+- **Right click a folder button on the taskbar** and choose **Manage folders...**.
+  The buttons are real taskbar items now, so that menu is Windows' own jump list —
+  the entry gets there through the folder's AppUserModelID rather than being drawn
+  by this mod. **Unpin from taskbar** sits on the same menu, for free.
+- Or right click any folder, file or the Desktop background, pick **Show more
+  options** if you get the short Windows 11 menu, then
+  **Taskbar Folders → Manage folders...**. This one works even when nothing is
+  pinned yet, so it is the way in from a standing start.
 
 The quickest way to add one in the first place is straight from Explorer:
 right click any folder in Explorer or on the Desktop, pick **Show more
@@ -74,14 +78,19 @@ context menus are left completely untouched.
 
 ### The Taskbar Folders window
 
-Every folder is listed with its icon and name. From there you can:
+Folders are listed under two headings — **On the taskbar** and **Not pinned** —
+and sorted by name within each. The window follows your Windows app theme, so it
+is dark when the rest of Windows is. From there you can:
 
 | Action | What it does |
 |--------|--------------|
 | **Add...** | Pick any folder and give it a name and icon |
-| **Edit...** | Change the name, folder or icon, and tick or untick **Pinned to the taskbar**. Unticking takes the button off the taskbar **without deleting anything** — the entry stays in the list with its name and icon, and can be pinned again any time |
+| **Edit...** | Change the name, folder or icon, and tick or untick **Pinned to the taskbar**. Double-clicking a row does the same |
 | **Remove** | Forget the entry entirely. The folder itself is never touched |
-| **Drag a row** | Drag it up or down the list to set the left-to-right order of the taskbar buttons |
+| **Pin** / **Unpin** | Moves the selected folder between the two sections. Unpinning takes the button off the taskbar **without deleting anything** — the entry keeps its name and icon and can be pinned again any time |
+
+The list is not reorderable, and does not need to be: **button order is set by
+dragging the buttons on the taskbar**, like any other pinned app.
 
 Each entry has three fields:
 
@@ -90,6 +99,9 @@ Each entry has three fields:
 | Name  | `Apps` | Shown as the title at the top of the hover grid |
 | Folder | `%USERPROFILE%\Desktop` | Environment variables are expanded. `shell:` targets that map to a **filesystem folder** also work (e.g. `shell:Desktop`, `shell:Downloads`). Virtual namespaces such as Control Panel, Recycle Bin, or This PC are not supported — use [Taskbar Folder Menus](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/taskbar-folder-menus.wh.cpp) for those. |
 | Icon | an emoji, `C:\icons\apps.ico`, or `C:\Windows\explorer.exe,0` | Leave empty to use the folder's own icon |
+
+**Open** next to the folder box opens whatever path is currently typed, so a
+path can be checked before it is saved.
 
 A good setup is to make a folder somewhere, fill it with shortcuts to the apps
 you want grouped, and point a button at it.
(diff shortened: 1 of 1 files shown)


DanRotaru:
Cool mod, the only thing is that there is no possibility to set it as tray icons, not taskbar items, for me it would be more useful.

Kiploom:
@DanRotaru There is already a mod that does that, though it doesn't have some of the features this one has, such as being able to simply hover over the icons rather than clicking to open up the folder. That mod is called 'taskbar-folder-menus'.

Kiploom:
All recommended fixes were implemented and I removed the unnecessary logging I had going on so nothing else functional was changed in that last commit.

Overall there were three main big changes/fixes from last human review [HERE](https://github.com/ramensoftware/windhawk-mods/pull/4936#issuecomment-5171550944):

1. The sliding animation is smoother and folders shouldn't disappear from the taskbar when spamming opening/closing a window. It's still not perfectly smooth, but I plan to fix that in V2. I spent a good amount of time getting it this smooth so I think it's good enough for V1.
2. The taskbar folders aren't stored in settings anymore as it was very limiting to store them there because I couldn't find a way to write to settings without editing the users registry (syncing whenever a folder was pinned via right clicking the folder was impossible without it), needing admin privileges, which shouldn't be necessary for a mod like this. My solution was to build my own built in taskbar manager that can be opened from settings or by right clicking on an already pinned folder. The manager also allows for rearranging the order of items easily and unpinning items without removing them completely (like an archive).
3. Users now have the ability to change the size of the icons in the tray (small, medium, large), as well as the font, font size, and font weight of both the title and the labels. This was because the reviewer stated text was hard to see sometimes.

/ready-for-reviewer

Kiploom:
Disregard previous comment. I just found a new bug that needs to be fixed real quick. I will ai review and try again tomorrow.

Kiploom:
Acknowledged. Analysis is correct, the check-then-raise gap in ReloadAndRefreshUI is real.

Declining the fix. The window is sub-microsecond and needs a settings save / pin / unpin to land inside it at the same instant as a disable or update. I'm accepting that risk; the v1.42/v1.43 fixes cover what I care about. Same call for the RequestReloadUI inline fallback.

Leaving as-is. If crash reports ever point here, I'll revisit.

m417z:
It still seems like it doesn't follow the taskbar animation perfectly, especially when a new taskbar item is created, and worse, it's quite easy to get it to a broken state (screenshots below). I think that it's better to get it fixed.

image

image

Regarding settings, I think it makes more sense to group them, something like:
```yaml
- behavior:
  - openFolderOnClick: true
    $name: Click opens the folder
    $description: Left clicking a taskbar button opens the folder in File Explorer.
  - position: beforeApps
    $name: Position
    $description: Which side of the app icons the buttons sit on.
    $options:
    - beforeApps: Before the app icons
    - afterApps: After the app icons
  - hoverDelayMs: 0
    $name: Hover delay (ms)
    $description: Delay before the grid appears. 0 is instant.
  $name: Behavior
- content:
  - includeSubfolders: true
    $name: Show subfolders
  ...
  $name: Content
- appearance:
  - itemSize: small
    $name: Item size
  ...
  $name: Appearance
```

Currently the grouping is made with titles which is more difficult to navigate.

It's preferable to do it in the initial release, since changing it later will mean that user settings will be lost, because adding nesting changes the keys of the settings.

Let me know what you think.

Kiploom:
Thanks — pushed fixes for 1, 2 and 5, and documented 3 and 4.

**1. `Wh_ModUninit` wiping the folder list — fixed, and you were right about the reason.** The wiki settles what I had backwards: `Wh_GetModStoragePath` says "The directory will be removed when the mod is removed", so Windhawk already owns the storage lifetime, and `Wh_ModUninit` runs on "mod unload (disable, update, reload)". Reversibility means undoing what the mod wrote *outside* its own storage; clearing the storage itself was pure data loss on every update. `RemoveAllFolderButtonsCore` now takes a `clearStore` flag: the confirmed "Remove all folder buttons..." action still empties the list, the unload path keeps every entry and only clears `pinApplied` — otherwise, as you noted, the reverse sync reads its own sweep as a user unpin and demotes everything to a draft on the next load.

**2. `Wh_ModInit` returning `FALSE` — fixed.** The taskbar.dll symbols only feed hover binding; every one of them is null-checked at its call sites, so a miss just means the hover hooks never install. It now logs and continues, which keeps the manager, the pin reconcile, the Explorer menu, the jump-list verb and the disable-time sweep reachable when a Windows update moves the symbols.

**3. Start → All apps — documented rather than suppressed.** Both suppression routes go through the shortcut that the shell uses to resolve the pin, and I would rather not risk breaking pin resolution for tidiness — so the README now names it explicitly in a new "What this writes, and what happens when you disable it" section.

**4. State living outside the mod's storage — no code change, documented.** The same new section lists exactly what gets written (the Start Menu shortcut, the shell's own pin, the per-AUMID jump list), states that the folder list itself lives only in the mod's Windhawk storage, and covers both the disable-time sweep and the reconcile sweep that cleans up after an unclean exit.

**5. Phantom settings checkbox — removed from the README.** There is no such setting; the sentence was left over from an earlier version.

The mod is still unpublished, so `@version` stays at 2.0.

DanRotaru:
Fixes fixes and again fixes)

Kiploom:
@DanRotaru Improvements improvements and again improvements
image

PR #5015, opened 2026-08-04
Explorer Command Bar v1.1.0

DanRotaru:
# Explorer Command Bar v1.1.0

## Changelog

- Add the New+ feature from PowerToys to Explorer Command Bar, with possibility to replace the default explorer's New.
- Add an optional command bar button for the selected item’s or current folder’s shell context menu, including shell extensions, nested menus, extended verbs, and Nilesoft Shell support.
- Add nested custom menus, improved icon handling, configurable button spacing, and more reliable command bar refreshes.
- Prevent crashes when unloading the mod by revoking XAML event handlers and safely cleaning up menus, timers, hooks, and other per-thread resources.
- Add Windhawk 2.0 settings compatibility for toolbar items without subitems.
- Bug fixes.

## Screenshots:

With this version introducing a new feature, possibility to replace the New with New+ ([PowerToys New+ utility](https://learn.microsoft.com/en-us/windows/powertoys/newplus))

Before:
image

After:
image

### DIFF
--- a/mods/explorer-command-bar.wh.cpp
+++ b/mods/explorer-command-bar.wh.cpp
@@ -530,12 +530,11 @@ a new tab or navigating to another folder makes them appear.
   - buttonLabel: Context menu
     $name: Item label
     $description: The label and tooltip of the context menu item.
-  - buttonIcon: ""
+  - buttonIcon: E8FD
     $name: Item icon glyph or icon path
     $description: >-
-      Leave empty for the three dots glyph. Alternatively, enter a Segoe
-      Fluent Icons hex code point or an .exe, .dll or .ico path with an
-      optional icon index.
+      A Segoe Fluent Icons hex code point, or an .exe, .dll or .ico path with
+      an optional icon index. Leave empty for the default glyph.
   $name: Context menu item
   $description: >-
     Add a configurable item which expands the real File Explorer shell context
@@ -764,16 +763,24 @@ struct TrackedRevoker {
 
 thread_local std::vector<TrackedRevoker> g_revokers;
 
+constexpr size_t kRevokersPruneMin = 64;
+thread_local size_t g_revokersPruneAt = kRevokersPruneMin;
+
 template <typename T, typename Revoker>
 void TrackRevoker(T const& source, Revoker&& revoker) {
     // Entries whose element is gone can go: XAML released their delegates
-    // together with the element itself.
-    for (auto it = g_revokers.begin(); it != g_revokers.end();) {
-        if (!it->source.get()) {
-            it = g_revokers.erase(it);
-        } else {
-            ++it;
-        }
+    // together with the element itself. Only when the vector has grown past
+    // the threshold, though - resolving every weak_ref is a COM call each, and
+    // populating a menu with many items would otherwise be quadratic in them.
+    if (g_revokers.size() >= g_revokersPruneAt) {
+        std::erase_if(g_revokers,
+                      [](TrackedRevoker const& tracked) {
+                          return !tracked.source.get();
+                      });
+
+        // Prune again once the survivors have doubled, so the work stays
+        // proportional to what's actually being added.
+        g_revokersPruneAt = std::max(kRevokersPruneMin, g_revokers.size() * 2);
     }
 
     // The revoker types differ per event, so they're type-erased behind the
@@ -789,6 +796,7 @@ void TrackRevoker(T const& source, Revoker&& revoker) {
 void RevokeHandlersForCurrentThread() {
     std::vector<TrackedRevoker> taken;
     taken.swap(g_revokers);
+    g_revokersPruneAt = kRevokersPruneMin;
 
     for (auto const& tracked : taken) {
         try {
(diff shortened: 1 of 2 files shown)


DanRotaru:
## Both required items are fixed, along with three of the optional ones.
 
1. The mod can be unmapped while `ShowShellContextMenu` is still on a thread's stack
Fixed as suggested.
- `g_openContextMenuCount` is now taken by an RAII `OpenContextMenuScope` at the top of `ShowShellContextMenu`, so it covers the whole call — `EnsureContextMenuOwnerWindow`, `CoInitializeEx`, `GetShellContextMenu`, `QueryContextMenu` and `InvokeCommand` — instead of just `TrackPopupMenuEx`.
- `DismissOpenContextMenus` drains instead of sampling: no timeout, and `WM_CANCELMODE` is re-posted on every 10 ms iteration, since a single post can land before the target thread has entered its modal loop. It logs the outstanding count once a second so an unload stuck behind a slow shell extension is diagnosable. As you noted, this can't deadlock — those threads never send anything to the thread running `Wh_ModUninit`.
- Because the drain now completes before the `RunFromWindowThread` loop, `RemoveButtonsForCurrentThread` can no longer run re-entrantly underneath a live menu.
- Added a second `g_unloading` check between `QueryContextMenu` and `TrackPopupMenuEx`: building the menu can take a while, and a menu built during an unload is now dropped rather than shown.
 
2. The `WindhawkContextMenuOwner_...` window class
 
Kept the owner window (rather than switching to a subclass on `hExplorerWnd`), but fixed all three problems:
 
- **Serialized registration.** `EnsureContextMenuOwnerClass` is gone; the class is registered once from `Wh_ModInit`, so the two UI threads can't race over it. It's the last thing `Wh_ModInit` does, after every failure path, since `Wh_ModUninit` isn't called when init returns `FALSE`.
- **`ERROR_CLASS_ALREADY_EXISTS` is a hard failure now.** You're right that for a fresh instance it can only mean a leftover registration pointing at an unmapped image. The mod logs an explanatory message and goes without the context menu item until Explorer is restarted, instead of creating a window on a dead class.
- **Unregistered unconditionally.** `Wh_ModUninit` first asks each leftover owner window's own thread to destroy it (via `RunFromWindowThread` on the owner window itself, which covers the cases where the Explorer window is already gone or the earlier `RemoveButtonsForCurrentThread` pass didn't reach it), then calls `UnregisterClassW` unconditionally and logs if it still fails. Combined with the point above, a failure there degrades to "no context menu item on the next load" rather than a crash.
- **Correct `HINSTANCE`.** Added the `GetCurrentModuleHandle()` helper from audio-scroll-switcher; `RegisterClassExW`, `CreateWindowExW` and `UnregisterClassW` all use it now instead of `GetModuleHandleW(nullptr)`.
 
## Optional items addressed
 
- **`Wh_ModUninit` waiting `INFINITE` for the launch threads.** Kept the unbounded wait — a bounded one would be worse — but added a `Wh_Log` before it, naming the elevation-prompt case, so a stalled unload is visible in the log.
- **`TrackRevoker` pruning on every registration.** It now prunes only once the vector has grown past a threshold, and the threshold is reset to twice the surviving count afterwards, so populating a large menu is linear again instead of quadratic in `weak_ref::get()` calls.
- **`GetExplorerWindowForElement` falling back to another process's window.** The `GetForegroundWindow()` fallback is now rejected unless `GetWindowThreadProcessId` says it belongs to this process.
 
## Left as-is
 
The remaining optional items and the functionality notes are unchanged for now: the README additions and the "off the UI thread" wording, the icon-setting trimming, and the four behavior notes.
 
## Unrelated change in this push
 
The default for the context menu item's icon setting is now the `E8FD` glyph instead of empty, and the fallback used for an empty value was changed to match, so both give the same icon.

m417z:
Regarding `CopyDirectoryRecursively`, why not use a Windows API for copying a folder: `IFileOperation` or `SHFileOperation`?

Address the `CoInitializeEx` / `CoUninitialize` imbalance from optional improvements.

Also, in this PR or in the future, you might prefer to use `winrt/Windows.Data.Json.h` to parse JSON instead of a hand-written parser. You can look at `taskbar-ai-quota` for a reference.

DanRotaru:
Thanks — all three points are addressed in this push.

## What changed

* `CopyDirectoryRecursively` is gone, replaced by a `CopyTemplateItem` helper built on `IFileOperation`.
* `ShowShellContextMenu` and the shell worker thread now use a `ComApartmentScope` RAII type instead of hand-paired `CoInitializeEx` / `CoUninitialize`.
* `FindJsonValueText` is gone; the New+ `settings.json` is parsed with `winrt::Windows::Data::Json`.
* `-luuid` added to `@compilerOptions`, for `CLSID_FileOperation`.

---

### > Regarding `CopyDirectoryRecursively`, why not use a Windows API for copying a folder: `IFileOperation` or `SHFileOperation`?

No good reason — `IFileOperation` is clearly the right call and it's in now. The recursive `FindFirstFileEx` walk and the `CopyFileW` in the file branch are both replaced by a single `CopyTemplateItem(source, targetFolder, targetName)`, so files and folders take the same path.

Beyond dropping the hand-rolled code, this means copying a template behaves like a normal Explorer copy: long paths, reparse points, alternate streams, elevation and the copy hooks other software installs are all handled, and the result is undoable with Ctrl+Z. The explicit `SHChangeNotify` afterwards is gone too, since `IFileOperation` notifies the shell about what it created.

Two details worth calling out:

* The flags are `FOF_NO_UI`, so nothing can put up a dialog. That matters here beyond the noise: the copy runs on a worker thread `Wh_ModUninit` waits for, so a modal prompt would stall an unload for as long as it's up. The trade-off is that a name collision gets answered with "yes" instead of failing the way `CopyFileW(bFailIfExists=TRUE)` did — `MakeUniquePath` picks a name which didn't exist a moment earlier, so the window is negligible. Closing it properly would mean an `IFileOperationProgressSink` to recover the renamed path; happy to add one if you'd prefer that.
* `PerformOperations` can succeed while the individual copy didn't go through (a denied elevation, for instance), so `GetAnyOperationsAborted` is checked separately.

### > Address the `CoInitializeEx` / `CoUninitialize` imbalance from optional improvements.

Fixed. `ShowShellContextMenu` paired `CoUninitialize` with `hrInit == S_OK`, so an `S_FALSE` return — the apartment was already initialized and the call still took a reference on it — leaked that reference.

Instead of correcting the five copies of the condition, both call sites now use a `ComApartmentScope` RAII type which uninitializes whenever the `CoInitializeEx` `SUCCEEDED`, leaving only `RPC_E_CHANGED_MODE` unpaired, since that one never entered the apartment. It also took the five `if (...) { CoUninitialize(); }` blocks out of `ShowShellContextMenu`'s early returns, and the COM objects are declared after the scope so they're released before it ends.

### > you might prefer to use `winrt/Windows.Data.Json.h` to parse JSON instead of a hand-written parser

Agreed, and done in this PR rather than later. `FindJsonValueText` and its string scanning are gone. `ReadPowerToysConfig` parses the file with `wdj::JsonObject::TryParse` and reads the settings through a small `FindNewPlusSetting` helper, which unwraps PowerToys' `properties` → `<name>` → `value` nesting. Both levels are treated as optional, so a setting written as a plain member still reads, and booleans accept a JSON boolean, a number or a string, since older PowerToys builds wrote them as strings.

PR #5016, opened 2026-08-04
Add Windows SDR to HDR Tonemapping Fix mod

millerpb:
## Windows SDR to HDR Tonemapping Fix

Adds a new mod (`dwm-eotf-gamma.wh.cpp`) that patches the DXBC shader bytecode in `dwmcore.dll` at DWM startup to replace the sRGB EOTF with a configurable pure power-law gamma for SDR-to-HDR tone mapping.

### What it does

When Windows HDR is enabled, DWM converts SDR content to HDR scRGB using the sRGB EOTF (a ~2.2 power curve with a linear toe segment near black). The result of this is lifted shadows and darker tones on SDR content when in HDR mode. This mod patches the floating-point constants that define that transfer function, replacing them with a simple power-law gamma curve, emulating actual SDR monitor behavior, giving users direct, more accurate, control over the SDR-to-HDR tone mapping without permanently modifying any files on disk.

Based on [dwm_eotf](https://github.com/ledoge/dwm_eotf) by ledoge (GPL-3.0).

Also addresses request https://github.com/ramensoftware/windhawk-mods/issues/2036

### Key details

- Targets `dwm.exe` (x86_64)
- Patches DXBC bytecode in `dwmcore.dll`'s read-only memory at startup; recalculates shader checksums so D3D accepts the patched bytecode
- Fully reversible: all patched bytes are restored when the mod is unloaded
- User-configurable gamma value (default 2.2); supported range ~1.8–2.6
- Requires Windows HDR to be enabled and a DWM restart to take effect

## Changelog

* Changelog Initial version 1.0

## Mod authorship

This mod was created by:

- - [x] Claude
- - [ ] ChatGPT
- - [ ] Gemini
- - [x] Other (please specify): 
Based on https://github.com/ledoge/dwm_eotf

### DIFF
--- a/mods/dwm-eotf-gamma.wh.cpp
+++ b/mods/dwm-eotf-gamma.wh.cpp
@@ -25,14 +25,30 @@ Based on [dwm_eotf](https://github.com/ledoge/dwm_eotf) by ledoge (GPL-3.0).
 
 ## How it works
 
-On DWM startup, Windhawk injects this mod before Direct3D is initialized. The
-mod locates the known DXBC shader blobs in `dwmcore.dll`'s read-only memory
-sections and patches four floating-point constants that define the sRGB transfer
-function, replacing them with the equivalent pure power-law constants. The shader
-checksums are recalculated so D3D accepts the patched bytecode.
+When DWM starts, Windhawk injects this mod early — ideally before Direct3D
+initializes, so the patched bytecode is what D3D compiles from. The mod walks
+`dwmcore.dll`'s read-only PE sections looking for DXBC shader blobs that contain
+all four sRGB EOTF float constants. When found, those constants are replaced with
+the equivalent pure power-law values and the shader checksum is recalculated so
+D3D accepts the modified bytecode.
 
-When the mod is unloaded (Windhawk disabled, settings changed, or Windows
-shutdown), all patched bytes are restored to their original values.
+When the mod is unloaded while DWM is running (e.g. Windhawk disabled or settings
+changed), all patched bytes are restored in memory. Note that the unload callback
+does not run at process exit, so no restore occurs at Windows shutdown — this is
+harmless since the process is terminating anyway.
+
+## Important: enable injection into dwm.exe
+
+`dwm.exe` is a critical system process. Windhawk does not inject into it by
+default — you must explicitly allow it in **Windhawk's advanced settings**:
+
+1. Open Windhawk → **Advanced settings**
+2. In the **Process inclusion list**, add `dwm.exe`
+3. Click **Save**
+
+Without this step the mod will be silently ignored.
+
+![Windhawk advanced settings](https://i.imgur.com/LRhREtJ.png)
 
 ## Usage
 
@@ -69,9 +85,8 @@ shutdown), all patched bytes are restored to their original values.
   patching or unpatching at runtime has no effect on shaders that are already
   compiled. The flicker is driven by Chrome's own mode-switching and cannot be
   suppressed from the DWM side.
-- The shader checksums in `dwmcore.dll` may change with Windows Updates. If the
-  mod reports patching 0 shaders after an update, the known-checksum list may
-  need updating.
+- If the mod logs "No matching shaders found" after a Windows Update, the shader
+  structure in `dwmcore.dll` has changed. Please file an issue.
 */
 // ==/WindhawkModReadme==
 
@@ -93,7 +108,8 @@ shutdown), all patched bytes are restored to their original values.
 
 #include <windows.h>
 #include <cstring>
-#include <cstdlib>
+#include <cwchar>
+#include <cstdint>
 #include <vector>
 
 // =============================================================================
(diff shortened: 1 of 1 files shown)


millerpb:
I've addressed the suggestions in the ai review within reason. However, the constant-only detection matched 37 unrelated shaders on Windows 11 (instead of the known 4), so checksums were kept as the primary selector with HasDXBCSubBlob kept as an improvement to the container detection logic.

millerpb:
I've deferred Content-based fallback matching (4): The per-constant replacement counts are now logged for every patched shader, so users on a build where the allowlist doesn't match have something concrete to paste in an issue rather than a silent no-op. A full content-based fallback (accept blobs containing all four constants exactly once, require exactly four such blobs, commit or rollback atomically) is a natural follow-up but involves non-trivial cross-section state and was out of scope here.

`counts[ci] != 1` hard rejection was left as logging-only for now. On the allowlisted build the counts are deterministic and the check is purely defensive. If content-based matching is ever added it becomes load-bearing and can be promoted to a hard reject at that point; the comment in PatchShaderBuf flags this.

The `LoadLibraryExW` hook (5) is not needed. Scanning `dwm.exe`'s raw bytes confirmed `dwmcore.dll` appears in its PE import table (offset 104244), meaning the OS loader maps it before any code, including Windhawk's injection point, runs. The existing GetModuleHandleW call is sufficient; the comment was updated to document this.

I think things are pretty well hardened as this point, but I'm going to queue one last AI review and do some more local testing before grabbing a real person. 🙂

millerpb:
Made more adjustments per ai review. Here are some explanations for the ones I did not pursue.

#4b — `(hdr->size + 3) & ~3u` alignment roundup
The 4-byte scan stride assumes blob sizes are multiples of 4. The DXBC spec mandates it, and scanning the actual `dwmcore.dll` binary confirmed 0 of 404 DXBC occurrences are unaligned. Defending against something the spec prohibits and the data disproves adds noise without benefit.

#4c — "out-of-bounds leaf" rollback leaves `matchedMask` bit set
The scenario requires a leaf whose start is inside the container but whose end extends past it. In a valid DXBC container the `size` field of the outer blob accounts for every byte of its sub-blobs; a sub-blob that overflows its parent would be malformed. `HasDXBCSubBlob` already validates `k + inner->size <= hdr->size` before accepting a container, so only geometrically valid containers reach the patch path. The condition the ai reviewer describes is structurally impossible in well-formed DXBC.

#4e — `VirtualProtect` restoring wrong protection across multi-page blobs
`VirtualProtect` reports the first page's old protection, so a blob spanning pages with different protections would have all pages restored to the first page's value. `.rdata` is uniformly `PAGE_READONLY`, so in practice all pages have the same protection and the "wrong value" is the correct value. The ai reviewer acknowledged this themselves ("benign for a uniform `.rdata`"). It also becomes moot entirely if delta writes are ever adopted.

millerpb:
Okay, I've addressed the remaining points, optional improvements, and verified the fourth component of the splat concern:

I disassembled all four target blobs directly from the installed dwmcore.dll using D3DDisassemble. In the SM4 (lib_4_0) sections, the only paths that run on any HDR-capable hardware, every sRGB constant is a 4-component immediate of the form l(x, x, x, 0.000000), and every instruction that references them uses .xyz only. The .w slot is always 0, is never read, and is never touched by our patch. Confirmed non-issue.

millerpb:
Made some more changes to target more eotf values in dwm. This generally fixes some of the odd flicker in scRGB apps. Only seems occur on occasion now on window resize, then back to expected scRGB on interaction.

millerpb:
Some notes from the latest changes:

**Container-copy approach** - The stale-checksum window between leaf patch and `FinalizeContainer` is measured in microseconds; hitting it requires a D3D device recreation landing during that window, which is essentially impossible in practice. Refactor risk outweighs the theoretical benefit.

**Delta writes (write only changed bytes)** - Valid observation, but the whole-blob write is already documented in the code comment as a deliberate choice. Implementing delta writes would complicate `PatchRecord` and the restore path without improving correctness.

**Gamma granularity (`std::from_chars`)** - The five-option dropdown covers the standard gamma values and sidesteps the locale issue cleanly.

**Defer container `ChecksumRoundTrips` startup-cost comment** - Implemented as part of the deferral

**`VirtualProtect` multi-page note** - Benign in a uniform .rdata section; the existing code comment already acknowledges it. Adding a second comment would be noise.

UPDATE: Toggling HDR off/on in Windows does not reset dwm in memory. A log off and log in is still the quickest way to enable the new settings. Updated readme.

millerpb:
I think this is the one. Just retested everything, and I think all the holes are plugged at this point. The other suggestions are too edge-case or minor to be worth refactoring at this point.

millerpb:
dwmcore.dll is a static import of dwm.exe, it's listed in the PE import directory and the OS loader maps it before any code in the process runs, including Windhawk's injection point. This isn't a load-order assumption; it's a PE loader guarantee. I verified it directly by finding "dwmcore.dll" in dwm.exe's raw import-table bytes. If Microsoft ever restructures it as a dynamic load, the entire approach of patching bytecode before D3D compilation would need rethinking regardless; the hook would be the least of the changes required.

PR #5022, opened 2026-08-05
Add rounded corners mod for snapped/maximized windows

leshaalexey:
Keeps window corners rounded when a window is maximized or snapped, without touching the window state itself.

Addressed everything from the AI review, except where noted below.

**1. `SystemParametersInfo(SPI_SETDRAGFULLWINDOWS)`** — removed. Corners now update the next time each window's visuals are refreshed, as suggested.

**2. Overlap with `custom-corner-radius`** — see the note at the bottom.

**3. Symbol resolution** — the `explorer.exe` warmer, the hand-rolled RVA cache (`ResolvedRvas`, `GetModuleIdentity`, `ResolveBySymbols`, `StoreRvas`, `LoadRvas`, `MatchesFilterList`, `RunningInDwm`, the `kKey*` constants) and the `@include explorer.exe` line are all gone. The mod now uses `WindhawkUtils::HookSymbols` in `dwm.exe` only, with full decorated signatures. Symbol resolution inside `dwm.exe` works fine — the earlier failure was on my side.

Two of the radius getters are listed with several signature variants (const/non-const, private/public), because they differ between builds. Only `CTopLevelWindow` overloads are listed on purpose: those hooks pass `this` to `IsMaximizedOrSnapped`, so a same-named method on another class must not match. That was a real bug in the previous revision — `wcsstr` matching on undecorated names bound to same-named methods of a different class, and the filter was then reading a foreign `this`.

**4. `LoadLibraryW`** — removed, `GetModuleHandle` only, bail out if `udwm.dll` isn't loaded.

**5. System DPI** — no longer derived by hand. A forced radius is scaled through `CTopLevelWindow::GetWindowData` + `CWindowData::ScaleForDpi`, i.e. DWM's own per-window scaling. Both symbols are optional and the system DPI remains a fallback.

Regarding "verify that hooking `GetDpiAdjustedFloatCornerRadius` is needed at all": on 10.0.26100.8972 it is. Tracing shows `GetEffectiveCornerStyle` returning `DWMWCP_ROUND` and `GetRadiusFromCornerStyle` returning `8.0` for a maximized window — the squaring happens further down, and it's the `...ForCurrentStyle` / `...DpiAdjusted...` getters that return zero there. Only hooking the first two produced no rounding at all on this build.

**6. Developer-only settings** — `dumpSymbols`, `noFilter`, `debugLog` and `forceRadius` removed, along with `LogThrottled`. Forcing a radius is now automatic: it only happens when DWM returned zero for a surface `IsMaximizedOrSnapped` vouches for, so on builds where promoting the style is enough the hooks are no-ops. Two settings remain: corner style and fallback radius.

**7. Screenshot** — added to the README.

Also applied from the optional list: `HasMultipleDwminitWarningsInLastMinute` guard (taken from `custom-corner-radius`, which is why this mod is GPL-3.0 — credited in a comment at the top), `DWMWCP_*` constants from `<dwmapi.h>`, `WindhawkUtils::StringSetting`, `<windhawk_utils.h>`, radius clamped instead of silently substituted, `Wh_Log` used directly.

## On the overlap with `custom-corner-radius`

Happy to move this there as an option if that's preferred — say the word and I'll close this PR and open one against that mod instead.

The reason I think it may work better standalone: the two mods answer different questions. `custom-corner-radius` changes the radius DWM already uses; this one restores rounding where Windows deliberately removes it, and every decision is gated on `IsMaximizedOrSnapped` rather than on a radius value. Folded in, it would mean a second, differently-scoped filter running through the same radius pipeline. But that's a judgement call about the catalog, not about the code, so I'll follow whatever you prefer.
## Changelog
 
This pull request introduces a new mod.
 
## Mod authorship
 
This mod was created by:
 
- [x] Claude
- [ ] ChatGPT
- [ ] Gemini
- [ ] Other (please specify): 

### DIFF
--- a/mods/rounded-corners-when-snapped.wh.cpp
+++ b/mods/rounded-corners-when-snapped.wh.cpp
@@ -66,11 +66,11 @@ Radius**, **Disable rounded corners in Windows 11**, ExplorerPatcher's rounded
 corner option, StartAllBack or Win11DisableRoundedCorners.
 
 ## Mod authorship
- 
+
 If this pull request introduces a new mod, please complete the section below.
- 
+
 This mod was created by:
- 
+
 - - [ ] The submitter, without AI assistance
 - - [x] The submitter, with AI assistance
 - - [x] Claude
--- a/mods/rounded-corners-when-snapped.wh.cpp
+++ b/mods/rounded-corners-when-snapped.wh.cpp
@@ -8,7 +8,7 @@
 // @license         GPL-3.0
 // @include         dwm.exe
 // @architecture    x86-64
-// @compilerOptions -lgdi32 -lwevtapi
+// @compilerOptions -lwevtapi
 // ==/WindhawkMod==
 
 // HasMultipleDwminitWarningsInLastMinute() is taken from the Custom Window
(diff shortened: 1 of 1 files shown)


m417z:
Thank you for the submission. I agree with point 7 of the last review.

I added it the custom-corner-radius mod here and credited you and your PR:
https://github.com/m417z/my-windhawk-mods/commit/d1a9303a34c4580f5d5911d26d0fefb12fa5fff7

What do you think?

You can try the mod here:
https://github.com/m417z/my-windhawk-mods/blob/d1a9303a34c4580f5d5911d26d0fefb12fa5fff7/mods/custom-corner-radius.wh.cpp

leshaalexey:
its good idea

m417z:
> its good idea

In this case, this pull request can be closed.

m417z:
Added as an option in Custom Window Corner Radius v1.3.

PR #5023, opened 2026-08-05
Add Taskbar Horizontal Scroll Media Control v1.0.0

shreyasjswork:
Adds Taskbar Horizontal Scroll Media Control v1.0.0.

The mod provides next/previous track control through horizontal mouse-wheel tilt over the taskbar, optional modifier handling, Ctrl + middle-click play/pause on empty taskbar space, continuous-input suppression, and Windows' genuine Quick Settings media card as a standalone flyout when the panel is closed.

Validation:

- Official `pr_validation.py` check passed for `shreyasjswork`.
- The pull request changes only `mods/taskbar-horizontal-scroll-media.wh.cpp`.
- The release source compiles successfully with the Windhawk compiler.

## Changelog

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/taskbar-horizontal-scroll-media.wh.cpp
+++ b/mods/taskbar-horizontal-scroll-media.wh.cpp
@@ -1,0 +1,1765 @@
+// ==WindhawkMod==
+// @id              taskbar-horizontal-scroll-media
+// @name            Taskbar Horizontal Scroll Media Control
+// @description     Change tracks with taskbar wheel tilt and show the genuine Windows media panel control independently.
+// @version         1.0.0
+// @author          Shreyas J S
+// @github          https://github.com/shreyasjswork
+// @license         GPL-3.0
+// @include         explorer.exe
+// @include         ShellHost.exe
+// @include         ShellExperienceHost.exe
+// @architecture    x86-64
+// @compilerOptions -lwindowsapp -lruntimeobject -ldwmapi -lshcore -lshell32 -luser32 -lole32 -loleaut32 -luuid
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Taskbar Horizontal Scroll Media Control
+
+![Taskbar horizontal scroll media control demo](https://raw.githubusercontent.com/shreyasjswork/taskbar-horizontal-scroll-media/main/assets/Windhawk.gif)
+
+Tilt the mouse wheel horizontally while the pointer is over the taskbar:
+
+- Right: next track
+- Left: previous track
+- Ctrl + middle-click on empty taskbar space: play/pause
+
+Ctrl is required by default for track-changing wheel tilt. The requirement can
+be disabled, or the wheel modifier can be changed to Shift, Alt, or the Windows
+key. Play/pause intentionally remains fixed to Ctrl + middle-click.
+
+A held wheel tilt triggers exactly once. Repeated horizontal-wheel packets are
+discarded until the configured release gap has passed.
+
+The mod uses Windows' genuine current Quick Settings
+`ControlCenter.MediaTransportControls`; it does not draw a replacement card and does not
+open a Chromium bubble.
+
+- When Quick Settings is already open, its media card updates in place.
+- When Quick Settings is closed, the same Windows media control is hosted alone
+  in a non-activating popup beside the taskbar.
+- Opening Quick Settings dismisses the independent control.
+
+The implementation is split between Explorer, which owns the taskbar gesture,
+and ShellHost, the Windows 11 process that owns Quick Settings on current
+builds (ShellExperienceHost on older Windows 11 builds). A visual-tree watcher
+captures Windows' system-created control; while
+the panel is closed, that same element is temporarily reparented into a desktop
+XAML island and then restored to its original parent and child index.
+
+Explorer owns the taskbar input surface and remains available when the Quick
+Settings host restarts. Only the Explorer process that owns a taskbar window
+acts on a gesture; other Explorer processes remain inert. The input path hooks
+the taskbar's own `Windows.UI.Input.InputSite.WindowClass` procedure for
+`WM_POINTERHWHEEL` and pointer-down messages; it does not install a system-wide
+mouse hook. Keeping both roles in one catalog mod also avoids a separate tool
+process and another IPC boundary while supporting pre-24H2 ShellExperienceHost.
+
+On the first use after the owning shell process starts, Windows must create the
+Quick Settings visual tree before its genuine media control exists. The mod
+asks Windows to create that tree at most once, only after the owning host has
+announced itself. A brief Quick Settings flash can occur on that first use, and
+opening the system surface can dismiss another open shell flyout.
+
+Requires Windows 11 build 22000 or newer. The native card relies on private
+Windows XAML structure, so a future Windows update can require a mod update.
+
+## Compatibility
+
+Windows XAML diagnostics supports one active visual-tree consumer in the Quick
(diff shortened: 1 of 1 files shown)


shreyasjswork:
Addressed the review against the updated v1.0.0 commit:

- Replaced persistent diagnostic storage writes with Wh_Log.
- Added deterministic host-window-class teardown.
- Marked global XAML/WinRT holders no_destroy, synchronized the cross-thread handoff, and moved releases to the captured UI dispatcher.
- Made unload waits safe and unbounded after signalling stop; queued UI cleanup now owns its completion event until execution.
- Limited input-hook initialization to the shell Explorer instance.
- Added build-gated ShellExperienceHost.exe support for pre-24H2 Windows without injecting both Quick Settings hosts.
- Handled removed/recreated media controls and guarded intentional reparent mutations.
- Stopped the dismiss timer whenever the standalone card is restored.
- Replaced the Everyone/AppContainer event ACL with the interactive user's default DACL.
- Changed the license to GPL-3.0 because the XAML diagnostics scaffold derives from the GPL reference implementation.
- Clarified that modifier settings affect wheel tilt only; Ctrl + middle-click remains the requested fixed play/pause chord.
- Documented the Taskbar Scroll Actions distinction, accepted setting ranges, older host support, and first-use behavior; removed unused includes/libraries and replaced the legacy formatter.

The first-use ms-controlcenter bootstrap is intentionally retained: Windows does not instantiate the genuine ControlCenter.MediaTransportControls until its Quick Settings visual tree exists. The mod documents this, performs it only while no capture exists, and suppresses standalone creation whenever Quick Settings is already open. A screenshot was not added because this source-only PR has no approved external image URL yet.

Validation: zero-warning Windhawk compile, clean official pr_validation.py, one changed mod file, and v1.0.0 remains a single commit.

/ai-review

shreyasjswork:
Addressed the review against the updated v1.0.0 commit:

- Removed the startup-time `GetShellWindow` gate. Every Explorer instance can initialize safely, but gesture hit-testing now accepts only taskbar windows owned by the current process, so separate folder-process instances remain inert.
- Preserved a separate dispatcher for the native host window, independent of capture removal, and always tears down the XAML island, host window, and registered class on that UI thread during normal unload.
- Replaced the unbounded dispatcher wait with a five-second bound. The completion handle is shared with the callback, and the module is kept mapped on timeout so a late callback cannot use a closed handle or unmapped code.
- Restricted the Control Center bootstrap window lookup to the current process and checks both `GetClassNameW` calls before comparison.
- Added Windows 11 build 22000 gating, ShellHost alive state, and a single-consumption bootstrap token. Quick Settings can be opened for capture at most once per host lifetime and never when the owning host has not announced itself.
- Moved XAML diagnostics injection out of early `Wh_ModInit`: a process-filtered `EVENT_OBJECT_SHOW` hook requests injection when `ControlCenterWindow` exists, with an existing-window check for reloads.
- Reset capture-ready, host-alive, and bootstrap state during ShellHost teardown.
- Moved next/previous off `WH_MOUSE_LL` and onto the router thread. Track navigation now uses `TrySkipNextAsync` / `TrySkipPreviousAsync` on the same current GSMTC session as play/pause.
- Replaced the 100 ms desktop polling timer with a one-shot dismissal timer. A process-filtered WinEvent hook restores the card when Quick Settings opens, and the open-panel scan now checks the class before visibility/DWM work.
- Documented why Explorer owns the taskbar gesture: it owns the input surface, remains available across Quick Settings-host restarts, filters non-shell Explorer instances at the taskbar-window boundary, supports the older ShellExperienceHost split, and avoids adding a separate tool process and IPC layer.
- Retained the integrated gesture because this mod adds horizontal one-held-tilt semantics, fixed Ctrl + empty-taskbar middle-click, current-session routing, and the standalone genuine Windows media card as one behavior; the README keeps the explicit Taskbar Scroll Actions distinction.
- Added the recorded GIF to the catalog README via the allowed `raw.githubusercontent.com` host. The public source/asset repository is https://github.com/shreyasjswork/taskbar-horizontal-scroll-media.

Optional cleanup included removing the unnecessary mutex `no_destroy`, making the reparent flag atomic, clamping non-positive settings to their documented minima, removing the unused UI Automation link library, simplifying taskbar root detection, and making UIA depth exhaustion fail closed.

Validation: zero-warning Windhawk compile, clean official `pr_validation.py`, GIF URL verified as `200 image/gif`, one changed catalog file, one v1.0.0 commit, and byte-identical standalone source.

/ai-review

shreyasjswork:
Addressed the review against the updated v1.0.0 commit:

- Removed the timeout/module-reference teardown entirely. Native-card cleanup now runs synchronously on the owning XAML thread: the private host handles a synchronous cleanup message in its own WndProc, with the repository's `RunFromWindowThread` pattern used through `ControlCenterWindow` when no private host exists. Restore, XAML-source close, host destruction, class unregister, and thread-affine releases all finish before `Wh_ModUninit` returns.
- Moved UI Automation out of `RouterThreadProc`. It is lazily created and used on the taskbar InputSite UI thread, then synchronously released on that same thread during unload; the router now tears down its apartment only after all of its own work is gone.
- Removed the low-level mouse hook and hook thread completely, including the startup/quit race.
- The Explorer input path now follows the repository taskbar convention: it hooks `CreateWindowInBand`, identifies only taskbar-owned `Windows.UI.Input.InputSite.WindowClass`, and hooks that window procedure for `WM_POINTERHWHEEL` and `WM_POINTERDOWN`. Other Explorer processes have no taskbar InputSite and stay inert; there is no system-wide per-mouse-event callback.
- Kept the taskbar WndProc lightweight: horizontal input only updates latch state and signals the router; Ctrl + middle-click performs the same UIA empty-space check used by the repository taskbar click implementation, then signals the router. GSMTC calls never run on the taskbar thread.
- Replaced blocking WinRT `.get()` calls with two-second `wait_for` bounds plus cancellation for manager, play/pause, next, and previous operations. Router unload therefore returns from broker work to the stop check within a defined bound.
- Reworked the flyout delay into a router deadline instead of a blocking sleep, so media commands remain responsive while a card show is pending.
- Treated simultaneous Windows 11 Notification Center Styler use as unsupported rather than claiming coexistence. The README documents the single-TAP conflict and its outcome: track controls continue, standalone capture depends on which TAP attaches first, and one mod plus the affected shell host must be restarted. Notification Center Styler is not installed on this isolated test system, so no unperformed runtime coexistence test is claimed.
- Retained the integrated gesture because the taskbar InputSite owns the horizontal/pointer gesture while the Explorer router and Quick Settings host jointly provide current-session navigation and the standalone genuine card. Splitting it would add another catalog dependency and IPC contract without removing either required in-process role.

Optional cleanup also replaced opaque numeric-stage log wrappers with descriptive `Wh_Log` messages, removed the dead Quick Settings skip parameter and redundant clamp branch, and corrected the shared-event ACL comment.

Validation: zero-warning Windhawk compile against the installed 1.7.3 engine import library, clean official `pr_validation.py`, `git diff --check`, one changed catalog source file, one v1.0.0 commit, byte-identical standalone source, and the public GIF remains available from the allowed raw GitHub host.

/ai-review

shreyasjswork:
One final teardown ordering refinement is included in `a9085bca`: the synchronous XAML-thread cleanup destroys the host window, returns from its WndProc, and only then unregisters the process-wide host class on the unload thread. This ensures no host WndProc remains on the stack during class removal.

The full response to the preceding review is in the immediately previous comment. Compile and catalog validation remain clean.

/ai-review

PR #5027, opened 2026-08-06
Update windows-animations mod to version 1.1.7

redrag2105:
# Windows Animations
> ⚠️ **Update Note (v1.1.7):** The settings menu has been completely reorganized into nested groups (Minimize, Close, Switch) to make it easier to navigate. If your custom settings were reset to defaults after this update, please re-configure them. We apologize for the inconvenience!

Welcome to **Windows Animations**, a comprehensive window transition suite for your desktop. Built from the ground up to deliver cinematic window animations. 

By utilizing a smart Hybrid Rendering Engine, this mod bridges the gap between stunning visual aesthetics and seamless execution.

## ✨ Key Features

* **🚀 Smart Hybrid Engine:** Intelligently seamlessly switches between lightning-fast GDI (for pixel-perfect destruction physics) and the native DWM Thumbnail API (to perfectly preserve Windows 11 rounded corners and drop shadows).

* **🎬 Cinematic Close Effects:** Transform how you close applications with six physics-based animations:
  * **Square Shatter:** The window violently explodes outward into digital blocks before drifting into the void.
    
    Square Shatter Preview

  * **Thanos Snap:** A disintegration wave sweeps across the window, turning it into thousands of tiny dust particles that curve away into the wind.
    
    Thanos Snap Preview

  * **Perlin Dissolve:** The window organically melts and dissolves into thin air using a smooth Perlin noise map.
    
    Perlin Dissolve Preview

  * **Cyber Glitch:** Signal-snow close — horizontal band tears and a fine noise dissolve that eats the window away.

    Cyber Glitch Preview

  * **Retro TV Off:** Classic CRT power-down — the window squashes into a bright horizontal beam, zips to a glowing center dot, then phosphor-fades to black.

    Retro TV Off Preview

  * **Pixel Melt:** The window melts downward like hot wax — staggered pixel columns drip at different speeds into jagged streaks.

    Pixel Melt Preview

* **🧞 Fluid Minimize & Restore:** Transform how windows minimize and restore with seven fluid, physics-inspired animations:

  * **Genie:** The window is pulled toward the taskbar like a classic macOS Genie effect, bending and compressing smoothly into its destination.

    Genie Minimize Preview

  * **Ink Splash:** The window collapses into a fluid splash of ink, spreading and deforming organically before fading away.

    Ink Splash Preview

  * **Mirage:** The window ripples and distorts like a heat haze, gradually dissolving into a soft, shimmering mirage.

    Mirage Preview

  * **Scorch:** The window breaks into vertical stripes that smoothly slide upwards or shrink until they completely disappear.

    Scorch Preview

  * **Splinter:** The window transitions through a grid of overlapping horizontal and vertical stripes that piece the window together or tear it apart.

    Splinter Preview

  * **Stipple:** The window breaks down into a dense field of tiny dots, gradually dispersing until the entire surface disappears.

    Stipple Preview

  * **Swell:** The window expands and bulges outward with a soft, elastic distortion before smoothly fading away.

    Swell Preview

* **🔄 Soft Switch Animation:** A pristine scale and fade-in animation triggered *exclusively* when actively switching windows via the Alt+Tab menu.
  * **Before:**
    
    Alt Tab Switch Before
  * **After:**
    
    Alt Tab Switch After
* **🛡️ Rock-Solid Stability:** Features lifecycle management for System Tray apps (like Discord, Steam, etc.)—handling background apps gracefully to minimize ghosting or stuck transparency.

## ⚙️ Customization & Settings

You can deeply customize the feel and pacing of every animation via the Windhawk Settings tab:

* **Minimize/Restore Animation Style:** Choose between Genie, Ink Splash, Scorch, Splinter, Mirage, Stipple, and Swell.
* **Close Animation Effect:** Dropdown menu to switch between 'Square Shatter', 'Thanos Snap', 'Perlin Dissolve', 'Cyber Glitch', 'Retro TV Off', and 'Pixel Melt'.
* **Minimize/Restore duration (ms):** Controls Genie speed directly; each style applies a small pace tweak. (Default: 360ms, clamp 200–1400)
* **Close animation duration (ms):** Controls how long the dramatic close animation lasts. (Default: 630ms)
* **Switch animation duration (ms):** Controls the snappy speed of the Alt+Tab scaling effect. (Default: 200ms)
* **Shatter block size (px):** Determines the size of the dust/shatter particles. 
  * *Performance Tip:* Smaller values (e.g., 1, 2) create hyper-realistic pixel dust but require more CPU power. Larger values (24, 32) yield a stylish retro pixelated shatter and perform effortlessly on any hardware. (Clamped strictly to 1-100).
* **Animate app launches:** Off by default. Enable it to use the restore effect when an application window first opens.
* **Animate windows hidden to the tray:** Off by default. Enable it to animate apps such as Discord, Steam, and Telegram when they hide their window instead of closing it. This can also animate splash screens or windows hidden automatically by an app.
* **Toggles:** Individually turn on/off Minimize, Restore, Close, Alt+Tab Switch, and Launch animations to suit your workflow.
* **Unhide taskbar (auto-hide):** Genie only. Ignored for Ink Splash, Scorch, Splinter, Mirage, Stipple, and Swell.

## Changelog

* Enhanced minimize and restore animations with five new fluid effects: Scorch, Splinter, Mirage, Stipple, and Swell.
* Updated Cyber Glitch animation for a more visually appealing and realistic effect.
* Added two new close animations: Retro TV Off and Pixel Melt, expanding the cinematic effects available.
* Updated compiler options for improved compatibility.
* Updated Readme and Settings
* Modified handling of deferred window show commands in animation logic for improved behavior.

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/windows-animations.wh.cpp
+++ b/mods/windows-animations.wh.cpp
@@ -103,7 +103,8 @@ You can deeply customize the feel and pacing of every animation via the Windhawk
 * **Switch animation duration (ms):** Controls the snappy speed of the Alt+Tab scaling effect. (Default: 200ms)
 * **Shatter block size (px):** Determines the size of the dust/shatter particles. 
   * *Performance Tip:* Smaller values (e.g., 1, 2) create hyper-realistic pixel dust but require more CPU power. Larger values (24, 32) yield a stylish retro pixelated shatter and perform effortlessly on any hardware. (Clamped strictly to 1-100).
-* **Animate windows hidden to the tray:** On by default. Also animate apps such as Discord, Steam, and Telegram when they hide their window instead of closing it. Disable this option if close effects appear unexpectedly for splash screens or hidden windows.
+* **Animate app launches:** Off by default. Enable it to use the restore effect when an application window first opens.
+* **Animate windows hidden to the tray:** Off by default. Enable it to animate apps such as Discord, Steam, and Telegram when they hide their window instead of closing it. This can also animate splash screens or windows hidden automatically by an app.
 * **Toggles:** Individually turn on/off Minimize, Restore, Close, Alt+Tab Switch, and Launch animations to suit your workflow.
 * **Unhide taskbar (auto-hide):** Genie only. Ignored for Ink Splash, Scorch, Splinter, Mirage, Stipple, and Swell.
 */
@@ -118,9 +119,9 @@ You can deeply customize the feel and pacing of every animation via the Windhawk
   - restore_animation: true
     $name: Animate restoring
     $description: Animate windows as they return from the taskbar.
-  - launch_animation: true
+  - launch_animation: false
     $name: Animate app launches
-    $description: Use the restore animation when an application window first opens.
+    $description: Disabled by default. Enable to use the restore animation when an application window first opens.
   - effect_style: genie
     $name: Animation style
     $description: Choose the effect used for minimizing and restoring windows.
@@ -162,12 +163,13 @@ You can deeply customize the feel and pacing of every animation via the Windhawk
   - duration_ms: 630
     $name: Duration (ms)
     $description: Close animation duration, from 50 to 5000 ms.
-  - hide_as_close: true
+  - hide_as_close: false
     $name: Animate windows hidden to the tray
     $description: >-
-      Also animate apps such as Discord, Steam, and Telegram when they hide their window instead
-      of closing it. This can also animate splash screens (Discord update splash screen) or windows hidden automatically by an
-      app (Google Chrome Profile Picker). Disable this option if close effects appear unexpectedly for some apps.
+      Disabled by default. Enable to animate apps such as Discord, Steam, and Telegram when they
+      hide their window instead of closing it. This can also animate splash screens, such as the
+      Discord updater, or windows hidden automatically by an app, such as the Google Chrome
+      profile picker. Disable this option if close effects appear unexpectedly.
   - shatter_block_size: 5
     $name: Dust and shatter block size (px)
     $description: >-
@@ -251,6 +253,7 @@ struct WindowAnimData {
     HWND hNextApp{};
     int unhideDurationMs{};
     BOOL deferredMinimize{};
+    int deferredShowCmd{};
 };
 struct LaunchAnimData { HWND hWnd; LONG_PTR originalExStyle; };
 struct SwitchAnimData { HWND hWnd; int durationMs; };
@@ -991,7 +994,10 @@ DWORD WINAPI UiaWorkerThread(LPVOID lpParam) {
         }
         if (coInit) CoUninitialize();
     }
-    if (uiaFound && !g_unloading.load(std::memory_order_relaxed)) {
+    if (g_unloading.load(std::memory_order_relaxed)) {
+        return 0;
+    }
+    if (uiaFound) {
         if (pending) {
             pending->targetX = targetX;
             pending->found = true;
@@ -1001,7 +1007,11 @@ DWORD WINAPI UiaWorkerThread(LPVOID lpParam) {
         if (!t->processKey.empty()) g_ProcessDockXs[t->processKey] = targetX;
         Wh_Log(L"UIA dock match hwnd=%p x=%d", t->hWndApp, targetX);
     } else {
-        Wh_Log(L"UIA dock miss hwnd=%p fallback=%d", t->hWndApp, t->fallbackX);
+        {
+            std::lock_guard<std::mutex> lock(g_StateMutex);
+            g_TaskbarDockXs[t->hWndApp] = t->fallbackX;
+        }
+        Wh_Log(L"UIA dock miss hwnd=%p fallback=%d (cached)", t->hWndApp, t->fallbackX);
     }
     return 0;
 }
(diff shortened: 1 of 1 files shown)


redrag2105:
For point 1&2, we can leave it like that. And about point 3, it's a trade off so I'll keep it.
/ready-for-reviewer

m417z:
Why not fix 2? Why not address 3 with `IsHungAppWindow` as suggested?

redrag2105:
@m417z 
For point 3, the reason I hesitated to use `IsHungAppWindow` is that it can trigger false positives when an application is merely busy doing heavy processing (like saving a large file). Completely dropping the animation in those moments makes the mod feel inconsistent and glitchy to the user. However, upon reflection, I will apply the suggestion and add the IsHungAppWindow guard. 

And for point 2, the flip is intentional, can I keep `hide_as_close ` as `true` by default and just update the README?

redrag2105:
### 1. Minimize/restore before `IsOurWindow` (cross-process)

**Intentional.** The guard was moved so taskbar-initiated minimizes (Explorer / `CloseWindow` / foreign `ShowWindow`) still animate. Putting `IsOurWindow` back at the top would drop that path.

### 2. Synchronous UIA lookup (400 ms) and first-frame wait

**Intentional.** Genie needs a dock X; waiting briefly on a cold lookup favors aiming at the real taskbar button over starting with a center/left fallback. After a hit **or miss**, the hwnd cache is written so later minimizes of that window use `UiaLookupWaitCachedMs = 0` and do not re-stall.

The first-frame wait is deliberate so the ghost is on screen before the real window is cloaked/minimized, avoiding flicker.

### 3. `AltTabTrackerThread` 16 ms poll

This is intentional, we've discussed about this in version 1.1.5 though.

/ready-for-reviewer

m417z:
> ### 3. `AltTabTrackerThread` 16 ms poll
>
> This is intentional, we've discussed about this in version 1.1.5 though.

I agree that calling `GetAsyncKeyState` constantly, even when the computer is idle, is not optimal.

Here's an example of an event-based tracking the Alt+Tab dialog without polling:
https://github.com/ramensoftware/windhawk-mods/blob/9c9fb82e9c7935004d0325e723d0f14dae3f631e/mods/taskbar-auto-hide-when-maximized.wh.cpp#L1008-L1010

You might want to improve it in the future.

redrag2105:
> > ### 3. `AltTabTrackerThread` 16 ms poll
> > This is intentional, we've discussed about this in version 1.1.5 though.
> 
> I agree that calling `GetAsyncKeyState` constantly, even when the computer is idle, is not optimal.
> 
> Here's an example of an event-based tracking the Alt+Tab dialog without polling:
> 
> https://github.com/ramensoftware/windhawk-mods/blob/9c9fb82e9c7935004d0325e723d0f14dae3f631e/mods/taskbar-auto-hide-when-maximized.wh.cpp#L1008-L1010
> 
> You might want to improve it in the future.

Thank you for your suggestion. To be honest, I've thought about this for a while, but there are a few problems:

- **Source window race:** On Alt+Tab enter, foreground may already be the switcher; need a prior “last app” FOREGROUND, not `GetForegroundWindow()` at enter time.
- **Win11 XAML–specific:** Relies on `XamlExplorerHostIslandWindow` + `GetWindowBand` + thread `"MultitaskingView"`. That's a bit risky and might break for Windows 10 users or classic switchers.
- **Session TTL:** I need to keep the animation state alive as long as Alt is held down. Events only give me enter/leave triggers, so I'd still need some sort of short poll during the session anyway.


For now, I'm thinking of going with an **Adaptive poll:** keep `GetAsyncKeyState`, but sleep **100–200 ms** while idle; tighten to **16 ms** only when `VK_MENU` is down.

I'll definitely keep playing around with the event-based approach locally for a future update though.

What do you think about this workaround? @m417z

m417z:
> Source window race

Sounds solvable with `EVENT_SYSTEM_FOREGROUND` tracking.

> Win11 XAML–specific

I believe that Win10 can use the same solution, perhaps with minor adaptations. Third-party switchers won't be supported unless you add specific detection for them, that's true.

> Session TTL

I didn't get into the details, but even then, short-lived polling is better than constant polling.

> For now, I'm thinking of going with an **Adaptive poll:** keep `GetAsyncKeyState`, but sleep **100–200 ms** while idle; tighten to **16 ms** only when `VK_MENU` is down.

Even if you prefer to be as generic as possible, supporting third-party switchers etc., why not check the keyboard state only on `EVENT_SYSTEM_FOREGROUND` events? A switcher always triggers a foreground window change event.

redrag2105:
@m417z I did a deep dive into the `EVENT_SYSTEM_FOREGROUND` approach you suggested, it introduces a cascade of edge cases that degrade the UX:

* The Fast Alt+Tab Miss: If a user does a super-fast Alt+Tab, Windows often bypasses rendering the Switcher UI entirely and directly sets FOREGROUND to the target app. By the time the event fires, the TAB key is already released, meaning a `GetAsyncKeyState(VK_TAB)` check fails, causing the mod to miss the animation.

* False Positives: Without checking `VK_TAB` (to catch fast switches), holding Alt for Photoshop tools while a background chat app steals focus triggers a false Alt+Tab session. Same with `Alt + F4` or `Alt + Space`.

* Class Name Filter Fails: Filtering by `XamlExplorerHostIslandWindow` prevents false positives, but completely breaks the "Fast Alt+Tab" scenario mentioned above since the Switcher UI never actually appears.

Do you think it's a worthwhile trade-off to just ignore the 'fast Alt+Tab' scenario? Or do you have another approach in mind to handle this?

redrag2105:
Oh nevermind. I implemented the `EVENT_SYSTEM_FOREGROUND` approach combined with a `VK_TAB` check. I tried to do a super-fast Alt+Tab to test the edge case, but it caught it perfectly every time. The race condition didn't happen. Thanks for your suggestion.

PR #5029, opened 2026-08-06
Update the mod to the version 1.1.0

babamohammed2022:
The mod has been updated version to 1.1.0 with a minor change

## Changelog

- Fixed a compatibility problem

## Mod authorship

This mod was created by:

- - [X] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/aero-flip3d-recreation.wh.cpp
+++ b/mods/aero-flip3d-recreation.wh.cpp
@@ -5,9 +5,10 @@
 // @version         1.1.0
 // @author          babamohammed
 // @github          https://github.com/babamohammed2022
-// @include         explorer.exe
+// @include         windhawk.exe
 // @compilerOptions -ldwmapi -luser32 -lgdi32 -lwinmm -lmsimg32
 // ==/WindhawkMod==
+
 
 // ==WindhawkModReadme==
 /*
@@ -21,9 +22,9 @@ This mod recreates the classic Windows Vista/7 Flip 3D window switcher on modern
 
 It keeps window previews live and displays them in a 3D-style cascade similar to the original effect.
 
-This mod is actively maintained, and some details may still be improved. To help the author improve it, feel free to send suggestions.
+This is the first version, so some details may still be improved. To help the author improve the mod, feel free to send suggestions.
 
-The mod has been tested on Windows 10 1809, Windows 11 23H2 and Windows 11 24H2.
+The mod has been tested on Windows 10 1809.
 ## Keyboard shortcuts
 
 | Shortcut | Action |
@@ -34,7 +35,6 @@ The mod has been tested on Windows 10 1809, Windows 11 23H2 and Windows 11 24H2.
 ### While Aero Flip 3D is open
 
 - Navigate: `Tab` / `Shift+Tab`, arrow keys, or mouse wheel
-- Hold `Tab` down to keep advancing automatically without repeated presses
 - Select: `Enter` or `Space`
 - Close: `Esc`
 - Emergency exit: `Ctrl+Shift+Esc` opens Task Manager
@@ -42,7 +42,7 @@ The mod has been tested on Windows 10 1809, Windows 11 23H2 and Windows 11 24H2.
 ## Settings
 
 - Simulated 3D cards: This setting uses the classic DWM thumbnail-strip 3D effect.
-- Auto-cycle while Win is held: automatically advance the selection every 250 ms while Win is held, in addition to Tab/Shift+Tab. On by default.
+- Auto-cycle while Win is held: automatically advance the selection every 250 ms while Win is held, in addition to Tab/Shift+Tab. Off by default.
 - Number of visible windows: how many cards are shown in the deck at once. Defaults to auto-detecting a value from your PC's RAM/CPU, but can be overridden.
 
 By default, the mod automatically detects your PC's hardware (RAM and CPU cores) and adjusts animation speed and the number of visible windows accordingly.
@@ -65,8 +65,6 @@ More thumbnail strips are used to make the edges smoother while keeping the orig
 - The native Alt+Tab window switcher is not modified and continues to work normally.
 - While the mod is enabled, the Win+Tab shortcut is redirected from the modern Task View interface to the classic Flip 3D experience.
 - To restore the original Windows Task View behavior, simply disable or uninstall the mod.
-## Credits 
-- Tails - Testing the mod on Windows 11 23H2 and proposing the fix that made it work there (it has since been verified on Windows 11 24H2 as well)
 */
 // ==/WindhawkModReadme==
 
@@ -75,9 +73,9 @@ More thumbnail strips are used to make the edges smoother while keeping the orig
 - usePerspective: true
   $name: Simulated 3D cards
   $description: Tilt the cards in 3D instead of showing them flat.
-- autoCycle: true
+- autoCycle: false
   $name: Auto-cycle while Win is held
-  $description: Automatically advance the selection every 250 ms while Win is held, in addition to Tab/Shift+Tab.
+  $description: Automatically advance the selection every 250 ms while Win is held, in addition to Tab/Shift+Tab. Off by default so Tab lets you step to an exact window.
 - visibleWindowCount: 0
   $name: Number of visible windows
   $description: How many cards are shown in the deck at once. 0 = auto-detect based on your PC's RAM/CPU (the default).
@@ -87,8 +85,8 @@ More thumbnail strips are used to make the edges smoother while keeping the orig
 // -----------------------------------------------------------------------------
 // Defensive rewrite (hardened build)
 //
-// This mod runs in the real primary Explorer shell process. The code follows
-// a conventional RAII and explicit-error-checking strategy:
+// This mod runs in Windhawk's dedicated tool process. The code follows a
+// conventional RAII and explicit-error-checking strategy:
 //
 //  1. RAII everywhere: HDC, HBITMAP, HBRUSH, HHOOK, HANDLE, HTHUMBNAIL and
 //     BeginPaint/EndPaint are owned by small move-only wrapper classes, so
@@ -101,8 +99,8 @@ More thumbnail strips are used to make the edges smoother while keeping the orig
 //     the UI thread.
 //  4. Minimal logging: Wh_Log is used only for init/uninit, unrecoverable
 //     failures and caught exceptions -- never for per-frame or per-key events.
-//  5. Explorer shell teardown joins the UI and input threads before the DLL
-//     is unloaded, so no hook or window procedure can outlive the module image.
+//  5. Tool-process teardown joins the UI and input threads before code is
+//     unloaded, so no hook or window procedure can outlive the module.
 //  6. Isolated components with fallbacks: if the desktop snapshot, the
 //     wallpaper bitmap or DWM thumbnails fail, the backdrop degrades to a
 //     plain Aero gradient and the mod keeps working.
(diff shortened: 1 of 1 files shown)


babamohammed2022:
> ### Submission review
> 
> 
> 
> The bulk of this PR is a single architectural change — moving ~3800 lines of GDI/DWM/threading code out of the dedicated tool process and into `explorer.exe`. That change is the main thing to resolve; the rest follows from it.
> 
> **1. Moving from the tool-mod process into `explorer.exe` is the wrong direction, and the stated reasons don't hold up.**
> 
> The mod installs **zero function hooks** — there is no `Wh_SetFunctionHook`, no `WindhawkUtils::SetFunctionHook`, no `HookSymbols` anywhere in the file. Everything it uses (`SetWindowsHookExW(WH_KEYBOARD_LL/WH_MOUSE_LL)`, `RegisterHotKey`, `EnumWindows`, `DwmRegisterThumbnail`, GDI) is desktop/session-wide and works identically from any process on the desktop. That is the textbook definition of a [tool mod](https://github.com/ramensoftware/windhawk/wiki/Mods-as-tools:-Running-mods-in-a-dedicated-process), and it's what the mod already was. Hand-rolling a single-instance mutex (`Local\Windhawk_AeroFlip3D_ExplorerSingleton`) to dedupe across multiple `explorer.exe` instances is itself one of the listed signals that a mod belongs in a dedicated process.
> 
> The ~90-line justification block at lines 95–178 asserts several things that aren't accurate:
> 
> * **"The tool process was a hollow explorer"** (point 3, lines 138–148): it wasn't. The previous version had `@include windhawk.exe`, and the wiki launcher spawned `windhawk.exe -tool-mod "aero-flip3d-recreation"`. `explorer.exe` was never launched as a host, and there was no `IsMainExplorerProcess()` / `FindWindow(L"Shell_TrayWnd")` process guard in the shipped 1.0.0 code — the whole "two instances, double hooks, fragile heuristic" story describes code that never existed in this mod.
> * **"The DWM thumbnail parent must be in the shell"** (point 2, lines 128–136): `DwmRegisterThumbnail` only requires the destination to be a top-level window owned by the calling process. [`simple-window-switcher`](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/simple-window-switcher.wh.cpp#L1660) — an Alt+Tab replacement with live DWM thumbnails, a full-screen switcher window, multi-monitor support and global hotkeys — registers all of its thumbnails from inside the `windhawk.exe` tool process.
> * **"Win+Tab must be registered in the same process/lifetime as Explorer"** (point 1): `RegisterHotKey` and `WH_KEYBOARD_LL` are per-desktop, not per-process. If the real problem is that Explorer's Task View claims `Win+Tab` before the mod can, the fix is to stop Explorer from claiming it, not to relocate the whole mod. `simple-window-switcher` does exactly that: the UI stays in the tool process, and a **small** `explorer.exe`-side component hooks `RegisterHotKey` to block the shell's own `Alt+Tab` registration ([code](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/simple-window-switcher.wh.cpp#L4770)), with `@include windhawk.exe` + `@include explorer.exe` and an exe-name branch in `Wh_ModInit`.
> 
> Please restore the tool-mod structure (`@include windhawk.exe`, `WhTool_ModInit` / `WhTool_ModSettingsChanged` / `WhTool_ModUninit`, and the launcher snippet copied **verbatim** from the wiki — a stray edit there makes review much harder). If the 23H2/24H2 `Win+Tab` problem is real and reproducible, add a separate, minimal `explorer.exe` component for just that piece. As it stands, a fault anywhere in this file — a GDI failure path, a DWM stall, a hook bug — takes down the whole shell, which is precisely what the tool-mod pattern exists to avoid.
> 
> Also please drop or rewrite the 95–178 comment block. Lines 122–126 additionally claim "`Wh_ModInit` below accepts only the primary taskbar-owning Explorer", which contradicts both the actual `Wh_ModInit` (line 3751 — it only takes a mutex) and the later comment at line 3739 that says the Explorer check was deliberately removed. Long, confidently-worded, factually wrong comments are worse than no comments.
> 
> **2. The single-instance mutex picks an arbitrary `explorer.exe` and never recovers.**
> 
> `Wh_ModInit` (line 3760) creates the named mutex and the _first_ instance to get there wins; every other `explorer.exe` returns `FALSE`. Nothing re-arbitrates afterwards, so:
> 
> * With "Launch folder windows in a separate process" enabled, a folder-window `explorer.exe` can win the race. When the user closes that window the process exits, the mutex is released, and no other Explorer instance ever retries — `Win+Tab` silently stops working until a settings change or a shell restart.
> * On an Explorer restart, the outgoing shell process may still hold the mutex while the new one initializes. The new (real) shell then loses and returns `FALSE`, and the old process dies moments later — nobody hosts the mod. This is the exact "restart explorer / taskbar recreation" scenario the comment at lines 3728–3734 claims to fix.
> 
> The tool-mod launcher doesn't have this problem: exactly one dedicated process is started and the framework manages it. This is another argument for going back to it rather than for hardening the mutex.
> 
> **3. `Wh_ModInit` blocks Explorer's main thread, and `Wh_ModUninit` blocks indefinitely.**
> 
> Per [Mod lifetime](https://github.com/ramensoftware/windhawk/wiki/Mod-lifetime), when a mod is loaded before the target process begins executing, `Wh_ModInit` runs on the target's **main thread**. `ModInitImpl` (line 3648) waits up to 5 s there for a worker thread to register a window class, create a message-only window, register a global hotkey and spawn a second thread — so every logon and every Explorer restart now gates the shell's startup on that handshake, and a stall costs a 5-second shell hang.
> 
> The unload path is worse. `ModUninitImpl` (lines 3672–3689) uses `WaitForSingleObject(..., INFINITE)` on both threads, and the UI thread's teardown (`HookThreadProcImpl`'s ScopeGuard → `FinishExitAfterAnimation`) does cross-process work while that wait is outstanding: `DwmUnregisterThumbnail` per card (synchronous ALPC to `dwm.exe`) and `BringWindowToTop` / `SetForegroundWindow` on an **arbitrary third-party application window**. If any of those stall, the Windhawk engine thread and the shell stall with them.
> 
> Notably, the mod's own 1.0.0 comment said exactly this and it was deleted rather than acted on:
> 
> > `// A tool process may wait indefinitely: unlike explorer.exe this cannot stall the shell.`
> 
> That reasoning was correct. In the shell host it no longer applies. Two things regardless of where the mod ends up hosted:
> 
> * Don't activate a foreign window from the teardown path at all — `FinishExitAfterAnimation` is called from the thread's cleanup guard, where activating whatever happened to be selected is not wanted.
> * If the code stays in `explorer.exe`, the joins need a bound (and a defined behavior when the bound is hit), not `INFINITE`.
> 
> **4. The mutex is leaked on the `ModInitImpl` failure path.**
> 
> In `Wh_ModInit`, `g_singleInstanceMutex` is acquired and `g_isExplorerProcess = true` is set _before_ `ModInitImpl()` runs (lines 3760–3774). When `ModInitImpl` fails it calls `ModUninitImpl()` and returns `false`, but nothing releases the mutex — and since `Wh_ModInit` returned `FALSE`, `Wh_ModUninit` is never called, so the release at line 3790 doesn't run either. Whether the handle is ever closed then depends entirely on the module actually being unloaded and static destructors running. Release it explicitly:
> 
> ```c++
>     if (!ModInitImpl()) {
>         g_isExplorerProcess = false;
>         g_singleInstanceMutex.reset();
>         return FALSE;
>     }
>     return TRUE;
> ```
> 
> **5. `SafeDestroyOverlayWindow` now deliberately leaves a live window with a `WndProc` inside the mod image.**
> 
> The rewritten fallback (lines 2742–2759) logs and returns when `DestroyWindow` fails, and the comment argues this avoids a stale `WndProc`. It does the opposite: a surviving window keeps `lpfnWndProc` pointing into the mod image, `UnregisterClassW` then fails because a window of the class still exists, and once `Wh_ModUninit` returns and Windhawk unmaps the DLL, the next message dispatched to that window crashes Explorer. `ShowWindow(SW_HIDE)` was no better, so this isn't a regression as such — but neither branch is acceptable. The requirement is that no mod-owned window survives `Wh_ModUninit`. In practice `DestroyWindow` is called from the window's owning thread here so it shouldn't fail; if it somehow does, retry rather than accepting a dangling `WndProc`.
> 
> Optional improvements
> 
> 
> * `GetCurrentModuleHandle()` (line 945) falls back to `GetModuleHandleW(nullptr)`. In the tool process that was harmless; in `explorer.exe` it means the window classes would be registered under **Explorer's** `hInstance`, and `UnregisterClassW(..., GetCurrentModuleHandle())` would then never match — leaving a permanently registered class with a dangling `WndProc`. Better to fail the call than to fall back.
> * The `FinishExitAfterAnimation` retry (lines 2786–2795) is convoluted: `SafeDestroyOverlayWindow(overlay)` has already run, so `!(overlay && IsWindow(overlay))` is effectively always true and the comment ("avoid re-activating unnecessarily if the first attempt succeeded") doesn't describe what the code does. A single `bool activated` flag set by the first attempt would be clearer.
> * The mutex name is hard-coded (line 3749). `WH_MOD_ID` is available as a wide string literal — `L"Local\\Windhawk_" WH_MOD_ID` keeps it in sync with the mod id.
> * Several comments describe the _diff_ rather than the code, e.g. lines 3109–3113 ("Unified in a single case because the three labels previously had identical separate branches (diff no-op)") and line 857 ("reduced from 50"). These read as review notes and go stale immediately.
> * `g_hookThreadId` and `g_inputThreadId` are plain `DWORD`s written from the worker threads and read from `ModUninitImpl` on another thread. `std::atomic<DWORD>` would remove the data race for free.
> * `ActivateSelectedWindow` logs "did not bring hwnd to foreground (expected, ...)" — describing a failure as "expected" in the log makes the message useless for diagnosis. Either it's normal (drop the log) or it isn't (log it as a failure).
> 
> Functionality notes
> 
> Non-critical observations and ideas about the feature behavior itself.
> 
> * **The auto-cycle logic looks inverted, and it's now on by default.** The new `WM_TIMER` guard (lines 3156–3174) advances the selection every 250 ms while **Win is held and Tab is _not_ held**, and explicitly skips advancing while Tab is down. The original Vista/7 behavior is the other way round: holding `Win+Tab` cycles continuously, and releasing `Tab` while keeping `Win` held stops on the current window so you can pick it. As written, the normal browse gesture (hold Win, tap Tab a few times, then hold still) can never settle — the deck starts spinning at 4/s the moment you let go of Tab, and releasing Win confirms whatever happens to be in front at that instant. Combined with flipping the default from `false` to `true`, that makes precise selection impossible out of the box. Worth re-checking against the behavior you intended; if the timer is meant to mirror Vista, the condition should be `tabDown`, not `!tabDown`.
> * The overlay is sized to `GetPrimaryMonitorRect()` only, and the README documents this ("Aero Flip 3D appears on the main monitor"). Not a bug, just noting that a multi-monitor deck (or at least following the monitor the foreground window is on) would be a natural improvement.
> * `kWinTabClaimAttempts` was reduced from 50 to 20 (~2 s). Since the LL-hook fallback handles `Win+Tab` whenever the hotkey can't be claimed, it's worth confirming the retry loop earns its place at all — if the shell reliably owns the combo on Windows 11, the claim sequence is 2 s of startup work for nothing.
> 
> **Next steps:**
> 
> * `/ai-review` - after pushing fixes, to get a review of the updated code. You can repeat this as many times as you need, but each review is thorough and usually there's no need for more than 2-3 iterations.
> * `/ready-for-reviewer` - once you're satisfied with the state of the pull request, to hand it over to a human reviewer. If some findings above are left unaddressed, add a short note explaining why.
> 

While I do agree that some statements are sensationalistic, I've personally tried the mod on Windows 11 24H2 using @include windhawk.exe and it did not work and it works when replacing windhawk.exe with explorer.exe. I could try making the code more defensive in that case but proposing to make the mod non-functional seems like a suggestion to cause a regression in the code given that there were multiple reports related to this problem. The AI reviewer does not have a Windows testing environment therefore it is trying to undo an update that could make the mod more useful and functional. These are the logs:
13:17:58.095 8552 explorer.exe  [WH] [local@aero-flip3d-recreation] [759:LoadSettings]: LoadSettings: perspective=1 autoCycle=1
13:17:58.099 8552 explorer.exe  [WH] [local@aero-flip3d-recreation] [750:DetectPerformanceProfile]: DetectPerformanceProfile: ram=7810 MB cores=8 -> lowEnd=0 veryLowEnd=0 fps=62 deck=8
13:17:58.099 8552 explorer.exe  [WH] [local@aero-flip3d-recreation] [3376:HookThreadProcImpl]: HookThread: started, tid=8164
13:17:58.099 8552 explorer.exe  [WH] [local@aero-flip3d-recreation] [3408:HookThreadProcImpl]: HookThread: controller window created, hwnd=00000000000E0136
13:17:58.100 8552 explorer.exe  [WH] [local@aero-flip3d-recreation] [3328:TryRegisterHotkey]: TryRegisterHotkey: RegisterHotKey(Ctrl+Alt+F12) registered (id=15677)
13:17:58.100 8552 explorer.exe  [WH] [local@aero-flip3d-recreation] [3339:BeginWinTabClaim]: Win+Tab claim acquired on attempt 1 of 20
13:17:58.100 8552 explorer.exe  [WH] [local@aero-flip3d-recreation] [3435:HookThreadProcImpl]: HookThread: install OK, entering message loop
13:17:58.101 8552 explorer.exe  [WH] [local@settings-to-control-panel] [1811:Wh_ModInit]: Initializing the Redirect Settings to Control Panel mod...
13:17:58.101 8552 explorer.exe  [WH] [local@settings-to-control-panel] [1545:TryInstallPniduiHook]: [PNIDUI-HOOK] pnidui.dll not loaded yet
13:17:58.104 8552 explorer.exe  [WH] [local@settings-to-control-panel] [1653:InstallImmersiveMenuHooks]: [PNIDUI-HOOK] Retry thread created
13:17:58.104 8552 explorer.exe  [WH] [local@settings-to-control-panel] [1585:PniduiRetryThread]: [PNIDUI-HOOK] Retry thread started - waiting for pnidui.dll to load
13:17:58.319 8552 explorer.exe  [WH] [local@aero-flip3d-recreation] [2780:ActivateFlip3DImpl]: ActivateFlip3D: 2 eligible top-level window(s) found
13:17:58.389 8552 explorer.exe  [WH] [local@aero-flip3d-recreation] [2837:ActivateFlip3DImpl]: ActivateFlip3D: 2 card(s) prepared (simulated DWM; skipped: 0 no-rect, 0 thumbnail-failed)
13:17:58.389 8552 explorer.exe  [WH] [local@aero-flip3d-recreation] [2500:RecomputeTargetsForCurrentSelection]: RecomputeTargetsForCurrentSelection: selectedIndex=0 desktopSelected=1 clientRect=1920x1080 dpi=120 cardCount=2
13:17:58.394 8552 explorer.exe  [WH] [local@aero-flip3d-recreation] [2213:EnableHighResAnimationTimer]: EnableHighResAnimationTimer: timeBeginPeriod(1) OK
13:17:58.448 8552 explorer.exe  [WH] [local@aero-flip3d-recreation] [2908:ActivateFlip3DImpl]: ActivateFlip3D: switcher opened with 2 card(s), persistent=0
13:17:58.541 8552 explorer.exe  [WH] [local@aero-flip3d-recreation] [2709:DeactivateFlip3D]: DeactivateFlip3D: closing (activateSelected=1)
13:17:58.634 8552 explorer.exe  [WH] [local@settings-to-control-panel] [1705:InstallSyscallFallback]: [SYSCALL-HOOK] NtUserTrackPopupMenuEx hooked successfully
13:17:58.635 8552 explorer.exe  [WH] [local@settings-to-control-panel] [1293:InstallAAMHook]: [AAM-HOOK] CoCreateInstance(CLSID_ApplicationActivationManager) failed: 0x800401F0
13:17:58.707 8552 explorer.exe  [WH] [local@settings-to-control-panel] [1358:InstallLegacyNameHook]: [MAP-LEGACY] Hook installed successfully
13:17:58.707 8552 explorer.exe  [WH] [local@settings-to-control-panel] [1866:Wh_ModInit]: [TRAY-WATCHDOG] Watchdog thread created
13:17:58.707 8552 explorer.exe  [WH] [local@settings-to-control-panel] [1758:TraySubclassWatchdogThread]: [TRAY-WATCHDOG] Watchdog thread started
13:17:58.786 8552 explorer.exe  [WH] [local@aero-flip3d-recreation] [3588:EntryPoint_Hook]: >
13:17:58.875 8552 explorer.exe  [WH] [local@aero-flip3d-recreation] [2225:DisableHighResAnimationTimer]: DisableHighResAnimationTimer: timeEndPeriod(1)
13:17:59.653 8552 explorer.exe  [WH] [local@aero-flip3d-recreation] [2780:ActivateFlip3DImpl]: ActivateFlip3D: 2 eligible top-level window(s) found
13:17:59.741 8552 explorer.exe  [WH] [local@aero-flip3d-recreation] [2837:ActivateFlip3DImpl]: ActivateFlip3D: 2 card(s) prepared (simulated DWM; skipped: 0 no-rect, 0 thumbnail-failed)
13:17:59.748 8552 explorer.exe  [WH] [local@aero-flip3d-recreation] [2500:RecomputeTargetsForCurrentSelection]: RecomputeTargetsForCurrentSelection: selectedIndex=0 desktopSelected=1 clientRect=1920x1080 dpi=120 cardCount=2
13:17:59.748 8552 explorer.exe  [WH] [local@aero-flip3d-recreation] [2213:EnableHighResAnimationTimer]: EnableHighResAnimationTimer: timeBeginPeriod(1) OK
13:17:59.798 8552 explorer.exe  [WH] [local@aero-flip3d-recreation] [2908:ActivateFlip3DImpl]: ActivateFlip3D: switcher opened with 2 card(s), persistent=0
13:18:00.027 8552 explorer.exe  [WH] [local@aero-flip3d-recreation] [2500:RecomputeTargetsForCurrentSelection]: RecomputeTargetsForCurrentSelection: selectedIndex=1 desktopSelected=0 clientRect=1920x1080 dpi=120 cardCount=2
13:18:00.254 8552 explorer.exe  [WH] [local@aero-flip3d-recreation] [2500:RecomputeTargetsForCurrentSelection]: RecomputeTargetsForCurrentSelection: selectedIndex=0 desktopSelected=0 clientRect=1920x1080 dpi=120 cardCount=2
13:18:00.448 8552 explorer.exe  [WH] [local@aero-flip3d-recreation] [2500:RecomputeTargetsForCurrentSelection]: RecomputeTargetsForCurrentSelection: selectedIndex=1 desktopSelected=0 clientRect=1920x1080 dpi=120 cardCount=2
13:18:00.541 8552 explorer.exe  [WH] [local@aero-flip3d-recreation] [2500:RecomputeTargetsForCurrentSelection]: RecomputeTargetsForCurrentSelection: selectedIndex=0 desktopSelected=0 clientRect=1920x1080 dpi=120 cardCount=2
13:18:00.651 8552 explorer.exe  [WH] [local@aero-flip3d-recreation] [2500:RecomputeTargetsForCurrentSelection]: RecomputeTargetsForCurrentSelection: selectedIndex=1 desktopSelected=0 clientRect=1920x1080 dpi=120 cardCount=2
13:18:00.737 8552 explorer.exe  [WH] [local@aero-flip3d-recreation] [2709:DeactivateFlip3D]: DeactivateFlip3D: closing (activateSelected=1)
13:18:01.082 8552 explorer.exe  [WH] [local@aero-flip3d-recreation] [2225:DisableHighResAnimationTimer]: DisableHighResAnimationTimer: timeEndPeriod(1)
13:18:28.544 8552 explorer.exe  [WH] [local@settings-to-control-panel] [1601:PniduiRetryThread]: [PNIDUI-HOOK] pnidui.dll not loaded after timeout, giving up

m417z:
> cause a regression in the code

What's the regression?

We investigated some issue related to the Simple Window Switcher mod and thumbnail previews, and discovered that the API only works with a 64-bit process, while windhawk.exe is 32-bit. Is the regression related to thumbnail previews?

babamohammed2022:
> > cause a regression in the code
> 
> What's the regression?
> 
> We investigated some issue related to the Simple Window Switcher mod and thumbnail previews, and discovered that the API only works with a 64-bit process, while windhawk.exe is 32-bit. Is the regression related to thumbnail previews?

The regression is that the mod does not even start at all if using @include windhawk.exe but it does when using @include explorer.exe so I think that I'm forced to use this approach unless there is an analog functional and documented one that I'm missing

m417z:
The first step is understanding why it doesn't start. How can it be reproduced?

babamohammed2022:
> The first step is understanding why it doesn't start. How can it be reproduced?

To reproduce it you need to install Windhawk on Windows 11 24H2 64 bit and the Aero Flip 3D Recreation mod from the store and press Win+Tab. In my case nothing happened but I tried this solution:
change @include windhawk.exe to @include explorer.exe and it works (don't ask me how)

m417z:
Could it be related to the Windhawk window running as administrator? `WH_KEYBOARD_LL` won't work in this case.

`RegisterHotKey` should work but perhaps it's already taken by explorer.exe. Perhaps that's why it works for you - when you target explorer.exe, you're able to call `RegisterHotKey` before explorer does.

You can test this with the Disable Windows Shortcuts mod - if you disable the shortcuts there, does the mod work with windhawk.exe as the target?

Simple Window Switcher bypasses this issue by targeting both explorer.exe (to prevent hotkey registration) and windhawk.exe (for the rest).

> don't ask me how

Who should I ask? It's your mod, you should know how it works.

babamohammed2022:
> Could it be related to the Windhawk window running as administrator? `WH_KEYBOARD_LL` won't work in this case.
> 
> `RegisterHotKey` should work but perhaps it's already taken by explorer.exe. Perhaps that's why it works for you - when you target explorer.exe, you're able to call `RegisterHotKey` before explorer does.
> 
> You can test this with the Disable Windows Shortcuts mod - if you disable the shortcuts there, does the mod work with windhawk.exe as the target?
> 
> Simple Window Switcher bypasses this issue by targeting both explorer.exe (to prevent hotkey registration) and windhawk.exe (for the rest).
> 
> > don't ask me how
> 
> Who should I ask? It's your mod, you should know how it works.

I will try the Disable Windows Shotcuts and let you know and related to the don't ask me how comment it was simply an affirmation related to the fact that changing the include made it work that's why. I will reach you out once I test it so that I can fix it properly by complying to the Windhawk approach but making it work correctly

babamohammed2022:
> Could it be related to the Windhawk window running as administrator? `WH_KEYBOARD_LL` won't work in this case.
> 
> `RegisterHotKey` should work but perhaps it's already taken by explorer.exe. Perhaps that's why it works for you - when you target explorer.exe, you're able to call `RegisterHotKey` before explorer does.
> 
> You can test this with the Disable Windows Shortcuts mod - if you disable the shortcuts there, does the mod work with windhawk.exe as the target?
> 
> Simple Window Switcher bypasses this issue by targeting both explorer.exe (to prevent hotkey registration) and windhawk.exe (for the rest).
> 
> > don't ask me how
> 
> Who should I ask? It's your mod, you should know how it works.

I've tried the Disable Windows Shortcuts mod and I've disabled WIN+TAB
image

However, this test confirms the problem because the first version of the mod (the one on the Windhawk store that includes windhawk.exe as target) does not work and there aren't even logs. 
image

Therefore, I'd like to know if I need to use the approach of the mod mentioned above because in my case nothing is happening, I suppose that the mod fails to start

m417z:
I'd like to understand why it doesn't work. I'll try the version from the store on my computer soon.

babamohammed2022:
> I'd like to understand why it doesn't work. I'll try the version from the store on my computer soon.

OK, I understand. If possible, let me know the results once you test it

m417z:
I tried both versions - before and after this PR. In both cases, it works for me except when the foreground program is running as administrator (for example, Windhawk or Task Manager).

babamohammed2022:
> I tried both versions - before and after this PR. In both cases, it works for me except when the foreground program is running as administrator (for example, Windhawk or Task Manager).

Given this situation, I'd prefer to stall this PR (not close it) so that I can focus on fixing this problem more carefully since on my side it does not work to avoid releasing something that is not stable enough. Thanks for testing it

m417z:
I noticed that you're specifying:
```
// @include         windhawk.exe
// @architecture    x86-64
```
In Windhawk v1.x, windhawk.exe is 32-bit, that's why the mod only works with Windhawk 2.0. Remove `@architecture`.

babamohammed2022:
Given the situation, I think that these problems could be addressed in future as the main problem seems to be solved
/ready-for-reviewer

PR #5030, opened 2026-08-06
Add Keep Nvidia Instant Replay on mod

dixxi1208:
Adds `nvidia-keep-instant-replay-on`, a mod that turns Nvidia ShadowPlay's Instant Replay back on whenever something switches it off.

Instant Replay tends to quietly switch itself off after a driver update, a crash, a stray hotkey or a game that toggles it, and you only find out when you needed the clip you didn't get.

**How it works.** The state is read from `HKCU\SOFTWARE\NVIDIA Corporation\Global\ShadowPlay\NVSPCAPS` and the key is watched with `RegNotifyChangeKeyValue`, so the mod reacts the moment the state changes rather than only on a poll. Changing it is done by replaying the user's configured *Toggle Instant Replay* shortcut, read fresh from the same key each time. Optional process lists can pause the mod, or restrict Instant Replay to only be on while chosen programs run.

Windhawk injects into every `nvcontainer.exe`; a session-local named mutex elects a single instance to do the work, and instances running under other accounts can't read the state and stay out of the way on their own.

Same goal as [AlwaysShadow](https://github.com/Verpous/AlwaysShadow), but as a mod so there's no extra program to install or keep running. AlwaysShadow prefers a local HTTP server hosted by NvContainer and only falls back to the shortcut; that server came from GeForce Experience's NvNode backend and isn't present on the Nvidia App, so this mod uses the shortcut only, which works on both.

Source repository: https://github.com/dixxi1208/NvidiaInstantReplayFix

## Changelog

Not applicable — this is a new mod, not an update to an existing one.

## Mod authorship

This mod was created by:

- - [x] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/nvidia-keep-instant-replay-on.wh.cpp
+++ b/mods/nvidia-keep-instant-replay-on.wh.cpp
@@ -57,11 +57,12 @@ re-enable may also change your input language. Ctrl+Shift+F10 works well. The sh
 re-read every time, so changing it takes effect immediately.
 
 Because the shortcut is replayed as real keystrokes, two things follow. Whatever window has
-focus also receives the combination. And the key-ups the mod sends will release those keys
-even if you happen to be physically holding one at that moment - so a shortcut built from keys
-you hold during normal use (Shift, Ctrl) is a poor choice. This is inherent to the approach:
-the Nvidia App no longer ships the local server GeForce Experience used to expose, so the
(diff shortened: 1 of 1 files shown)


m417z:
Any reason the review findings 1-4 aren't addressed?

PR #5039, opened 2026-08-07
Upload the Performance Information and Tools Restorer mod

babamohammed2022:
This mod restores the classic Windows Experience Index page in Windows 10/11. Credits to Anixx for the example with the Windows 10 taskbar on Windows 11 24H2 mod.

## Changelog

* Changelog item 1...
* Changelog item 2...

## Mod authorship

This mod was created by:

- - [X] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/performance-info-tools-restorer.wh.cpp
+++ b/mods/performance-info-tools-restorer.wh.cpp
@@ -12,9 +12,14 @@
 // ==/WindhawkMod==
 
 // Note on @architecture x86-64:
-// While "amd64" is the common Windows designation for 64-bit x86, using `@architecture amd64`
-// causes repository validation errors on Windhawk. The repository schema strictly requires
-// `@architecture x86-64` as the mandatory specifier for x86-64 mods.
+// "x86-64" is the canonical specifier used by Windhawk mods and passes the
+// repository's PR validation cleanly. "amd64" is also accepted by
+// .github/pr_validation.py (the only hard rejections are architectures outside
+// {"x86", "x86-64", "amd64", "arm64"}), but it produces a "manual verification
+// required" warning for the maintainers, so x86-64 is used here. The payload
+// DLL is x64-only, and on ARM64 devices Windhawk may still inject this mod into
+// the native ARM64 explorer.exe; Wh_ModInit() refuses to start there before any
+// hook is installed (see IsRunningAsAmd64).
 
 // ==WindhawkModReadme==
 /*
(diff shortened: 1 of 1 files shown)


babamohammed2022:
In my opinion the download is necessary to avoid manual user configurations (because it is likely that the user could accidentally use the Windows Vista/7 DLL, causing incompatibility and blaming the author)
/ai-review

babamohammed2022:
Given that a mod that downloads a binary file (Anixx's Windows 10 taskbar on Windows 11 24H2 mod) exists and my mod does a similar thing but with a more specific DLL and verification, is it allowed to keep the mod as it is? In my opinion after following AI's review the mod should overall be compliant with the Windhawk good practices by using SHA-256 verification for example, if there are more robust methods to make the download more secure let me know since this mod could be useful for Control Panel restoration projects given that it is more user friendly than others because it embeds the strings in the mod and only depends on the DLL file. Additionally, from what I know these servers are static therefore the risk of the file being unavailable is unlikely but if it ever happens I will probably update the mod accordingly. I have put this command to receive a human review before applying further changes as suggested by the AI.
/ready-for-reviewer

babamohammed2022:
While I acknowledge that there have been multiple AI reviews, it would be good to receive a quick human feedback. @m417z Given that a mod that downloads a single binary file (Anixx's Windows 10 taskbar on Windows 11 24H2 mod) exists and my mod does a similar thing but with a more specific DLL and verification, is it allowed to keep the mod as it is? In my opinion after following AI's review the mod should overall be compliant with the Windhawk good practices by using SHA-256 verification for example, if there are more robust methods to make the download more secure let me know since this mod could be useful for Control Panel restoration projects given that it is more user friendly than others because it embeds the strings in the mod and only depends on the DLL file. Additionally, from what I know these servers are static therefore the risk of the file being unavailable is unlikely but if it ever happens I will probably update the mod accordingly. I have put this command to receive a human review before applying further changes as suggested by the AI.
If possible, let me know. Thanks for your time.

m417z:
> Given that a mod that downloads a single binary file (Anixx's Windows 10 taskbar on Windows 11 24H2 mod) exists and my mod does a similar thing but with a more specific DLL and verification, is it allowed to keep the mod as it is?

Yes, this part is acceptable in principle. I haven't looked at the code, I'll review it when it's ready and when I have time.

I also noticed that you're using the `%ProgramData%\\Windhawk\\Engine\\ModsWritable` path. That won't work with the portable Windhawk version. Use `Wh_GetModStoragePath` instead.

babamohammed2022:
In my opinion the mod is ready for a human review and if necessary I'll address the reported problems because some of them seem redundant.
/ready-for-reviewer

m417z:
Point 6 is, in my opinion, the most fragile part in the mod. How about using a volatile registry key instead, and redirect to it?
The Startup App Delay Fix mod that was released recently does it, you can use it as a reference.

m417z:
Don't use `"Software\\Windhawk\\PerformanceInfoToolsRestorer"`, as `Software\\Windhawk` won't exist in a portable installation, and shouldn't be created. Use a single key in `Software`, e.g. `"Software\\WindhawkPerformanceInfoToolsRestorer"`.

babamohammed2022:
@m417z Is it good now or are there any other problems to address to make it more stable?

PR #5044, opened 2026-08-08
Add widget-board-size-position

Meteony:
## Changelog
# Widget Board Size and Position

Set a custom size and position for the new Widget Board on Windows 11. 

Allows you to override the default position and dimensions of the new WinUI 3 Widget Board (`WidgetBoard.exe`).

Screenshot
_Example for making the Widget Board smaller_

Width, height, and offsets are specified in DIPs (96-DPI logical pixels), so a
700x600 setting becomes 1400x1200 native pixels on a 200% display.

`horizontalAnchor` accepts: `system`, `left`, `center`, `right`.
`verticalAnchor` accepts: `system`, `top`, `center`, `bottom`.

When an anchor is `system`, Windows' requested position is preserved on that
axis. Positive offsets move inward from left/top/right/bottom; for `center`,
they move right/down from center.

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini


Meteoni:
I don't observe any visible snapping here. Changing the logic as suggested would require additional refactoring and API changes that have not yet been validated. The actual gain seems minimal in this case. 
The stop event is a deliberate hardening choice.

m417z:
The review seems to be correct, `EVENT_OBJECT_SHOW` with `WINEVENT_OUTOFCONTEXT` is delivered asynchronously, so the size/pos might be applied with a delay when first shown. You might want to improve it in the future.

PR #5056, opened 2026-08-09
Taskbar Audio Spectrum v1.1.0

ShangHTao:
## Taskbar Audio Spectrum v1.1.0

Displays a click-through real-time system playback spectrum over the native Windows 10/11 taskbar search box.

The mod uses WASAPI loopback capture and runs in a dedicated `windhawk.exe` tool process, following Windhawk's tool-mod guidance. The submitted single-file source is generated deterministically from the upstream project and has been compiled successfully for x86, x64, and ARM64 with the Windhawk v1.7.3 toolchain.

Windows 10 with the full Search box enabled was tested successfully on a separate PC.

This mod is intentionally separate from Taskbar Fluent Media Player and Desktop Audio Visualizer: it renders specifically inside the native search box, requires no media-player widget, and runs as a dedicated tool process without injecting the audio and rendering runtime into Explorer.

Project: https://github.com/ShangHTao/TaskbarAudioSpectrum
Release: https://github.com/ShangHTao/TaskbarAudioSpectrum/releases/tag/v1.1.0

## Changelog

New mod submission; not applicable.

## Mod authorship

This mod was created by:

- - [x] Claude
- - [x] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/taskbar-audio-spectrum.wh.cpp
+++ b/mods/taskbar-audio-spectrum.wh.cpp
@@ -1,0 +1,3583 @@
+// ==WindhawkMod==
+// @id              taskbar-audio-spectrum
+// @name            Taskbar Audio Spectrum
+// @description     Render the system audio spectrum in the native Windows 10/11 taskbar search box
+// @version         1.0.0
+// @author          shanght
+// @github          https://github.com/ShangHTao
+// @homepage        https://github.com/ShangHTao/TaskbarAudioSpectrum
+// @license         MIT
+// @include         windhawk.exe
+// @architecture    x86
+// @architecture    x86-64
+// @compilerOptions -lole32 -loleaut32 -luuid -lgdi32 -ladvapi32 -lshell32 -ldwmapi
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Taskbar Audio Spectrum
+
+![Taskbar Audio Spectrum preview](https://raw.githubusercontent.com/ShangHTao/TaskbarAudioSpectrum/main/assets/taskbar-spectrum-preview.png)
+
+Displays a click-through real-time spectrum over the native Windows 10 or
+Windows 11 taskbar search box. System playback is captured with WASAPI
+loopback, so microphone permission isn't required and the search box remains
+clickable.
+
+## Requirements
+
+Set the Windows taskbar Search option to the full **Search box** before enabling
+the mod. Do not run the standalone Taskbar Audio Spectrum application at the
+same time because both versions render the same overlay.
+
+The `thirdOctave` scale derives its number of bars from the minimum and maximum
+center frequencies. `nonOctaveBarCount` applies only to Bark, logarithmic, Mel,
+and linear scales. Rapid taskbar layout tracking temporarily increases geometry
+checks after the taskbar changes, which helps the overlay follow animations and
+layout updates.
+
+Unlike **Taskbar Fluent Media Player**, this mod doesn't add a media-player
+widget. Unlike **Desktop Audio Visualizer**, it renders inside the taskbar search
+box. It runs in a dedicated `windhawk.exe` tool process instead of injecting the
+audio and rendering runtime into Explorer.
+
+Source, standalone downloads, and detailed documentation are available on the
+[project page](https://github.com/ShangHTao/TaskbarAudioSpectrum).
+*/
+// ==/WindhawkModReadme==
+
+// ==WindhawkModSettings==
+/*
+- frequencyScale: "thirdOctave"
+  $name: Frequency scale
+  $description: Third-octave uses the center-frequency range; other scales use the non-octave bar count.
+  $options:
+  - "thirdOctave": Third-octave
+  - "bark": Bark
+  - "log": Logarithmic
+  - "mel": Mel
+  - "linear": Linear
+- nonOctaveBarCount: 28
+  $name: Non-octave bar count
+  $description: Number of bars for Bark, logarithmic, Mel, and linear scales; ignored by third-octave.
+- minimumFrequency: "31.5"
+  $name: Minimum frequency (Hz)
+  $description: Lower analysis edge for non-octave scales; ignored by third-octave.
+- maximumFrequency: "16000"
+  $name: Maximum frequency (Hz)
+  $description: Upper analysis edge for non-octave scales; ignored by third-octave.
+- minimumCenterFrequency: "31.5"
+  $name: Minimum third-octave center (Hz)
+  $description: Lowest nominal center for third-octave; ignored by other scales.
+- maximumCenterFrequency: "16000"
+  $name: Maximum third-octave center (Hz)
+  $description: Highest nominal center for third-octave; ignored by other scales.
+- referenceFrequency: "1000"
+  $name: Third-octave reference frequency (Hz)
+  $description: Reference used to calculate nominal third-octave centers; ignored by other scales.
+- bandAggregation: "energy"
+  $name: Band amplitude calculation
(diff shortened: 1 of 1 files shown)


ShangHTao:
Addressed the review findings in `839c429`.

- Kept the upstream CMake generation workflow, but changed it to emit a clean, reviewable Windhawk-only source: one top include block, no source-section markers, standalone/test/probe code pruned, and deterministic output. The generated file is 3,373 lines and the tool-mod block is byte-for-byte identical to the current Wiki snippet.
- Removed `expectedSampleRate` and `positionAnchor`; renamed and documented the non-octave bar-count setting; made `thirdOctave` canonical; added descriptions and dropdown options for enum-like settings.
- Added per-monitor-v2 DPI awareness to both relevant threads and support for Windows 11 search-box mode 3.
- Cached the UIA search element, subscribed to bounding-rectangle changes, reduced fallback polling, eliminated duplicate search-open checks and per-tick path allocation, and skipped unchanged layered-window uploads.
- Removed Windhawk-only dead code and `-lavrt`; linked DWM directly; switched string settings to `WindhawkUtils::StringSetting`; added `clang::no_destroy`; removed forced process exits on failed settings reload.
- Reused FFT scratch buffers and now copies/releases each WASAPI packet before analysis.
- Expanded the mod README with the screenshot, requirements, setting relationships, tracking explanation, and differentiation from existing mods.

Validation completed locally:
- Debug, Release, and CI test presets pass.
- MSVC NativeRecommended static analysis passes.
- Runtime stress test passes (z-order, takeover, Explorer recovery, startup registry).
- Repository `pr_validation.py` passes.
- Repository `compile_mod.py` passes for x86, x64, and ARM64.

ShangHTao:
Addressed the two required technical issues in `ae8467f` while keeping the submitted `.wh.cpp` deterministic and CMake-generated.

1. **Removed the permanent 100 ms shell-state polling.**
   - Search mode is cached and updated with `RegNotifyChangeKeyValue`.
   - Foreground/search-flyout changes are driven by `SetWinEventHook(EVENT_SYSTEM_FOREGROUND)`, and fullscreen changes for the current foreground window are driven by `EVENT_OBJECT_LOCATIONCHANGE`.
   - Published UIA layouts now signal the overlay directly instead of being discovered by a layout timer.
   - The UIA element remains cached and event-driven; the normal safety refresh was reduced from every 5 seconds to every 30 seconds.
   - The Explorer process handle is waited on directly, so shell replacement is detected immediately without shortening the safety interval.
   - A 30-second timer remains only as recovery if an external event source fails. The frame timer no longer reads the registry or opens foreground-process handles.

2. **Made failed shutdown recoverable.**
   - Every `Runtime::Start` now creates a fresh `ApplicationContext`.
   - If bounded shutdown fails, all old state is already signaled to stop, its thread handles are released, and the stale context is deliberately isolated for any worker still returning from an external shell call.
   - The runtime is left restartable, and `WhTool_ModSettingsChanged` always attempts to start a fresh runtime instead of returning permanently dead.
   - Overlay class reuse is handled for the rare interval in which a stale overlay thread is still unwinding.

On the relationship to the existing visualizer mods: I believe this is a distinct user-facing project, not only another FFT rendering preset.

- Users who already use Windhawk get a catalog-installable mod with Windhawk-native settings and updates.
- Users who don't want to install Windhawk can use the same runtime as a lightweight portable standalone executable, with no mod framework required. Keeping both delivery paths is intentional and lets users choose the integration model that fits their system.
- Within Windhawk, the defining UX is also different: this visualizer is anchored specifically inside the native taskbar search box, remains click-through, and preserves the search control's normal behavior. `desktop-audio-visualizer` is a free desktop overlay, while `taskbar-fluent-media-player` attaches visualization to a media widget. Search-box placement is the core interaction here rather than an incidental position option.
- The shared engine also provides selectable frequency scales, third-octave bands, weighting, window functions, and aggregation modes while remaining a dedicated tool process.

Validation completed:
- Debug, Release, and full `windows-ci` builds and CTest pass.
- MSVC NativeRecommended analysis passes without warnings.
- The CMake-generated source is deterministic and byte-identical to the PR file.
- Windhawk's repository validation passes.
- Windhawk's official compiler script passes for x86, x64, and ARM64.

ShangHTao:
Addressed all three required findings in `a33faad` while keeping the submitted source deterministic and CMake-generated.

1. Windhawk logging now calls `Wh_Log` directly at every call site. The Windhawk-only wrapper, fixed 768-character buffer, redundant pid/host prefix, and `<cstdarg>` were removed. This transformation is limited to the generated Windhawk source; the standalone lightweight application intentionally retains its existing file logger and `spectrum.log`.
2. Search-box mode is now OS-specific using the real Windows build from `RtlGetVersion`: mode 2 is accepted only on Windows 10, and mode 3 on Windows 11. Explicit tests cover Windows 10, Windows 11, and the conservative unknown-platform fallback.
3. The locator supervisor now initializes a multithreaded COM apartment before starting workers and uninitializes it on exit, so `CoCancelCall` is invoked from a COM-initialized thread. Initialization failure exits cleanly without starting an uncancellable worker.

Validation completed: Debug, Release, and full Windows CI builds; all CTest suites; MSVC NativeRecommended analysis; deterministic generation; Windhawk repository validation; and the official Windhawk compiler script for x86, x64, and ARM64. The generated file is byte-identical to the PR source.

ShangHTao:
Corrected the search-box mode mapping in `aedc5c2` and regenerated the submitted single-file source from upstream commit [`81e163b`](https://github.com/ShangHTao/TaskbarAudioSpectrum/commit/81e163b).

This was verified directly on Windows 11 Enterprise LTSC 24H2 (build 26100) by switching the taskbar Search setting and observing the exact per-user value read by the mod at `HKCU\Software\Microsoft\Windows\CurrentVersion\Search\SearchboxTaskbarMode`:

- `2` = full **Search box**
- `3` = narrower **Search icon and label** button (UI Automation reported a 133 px-wide rectangle at 125% DPI)

The previous PR note and review finding that treated mode `3` as the Windows 11 full search box conflated the numeric values of the separate `ConfigureSearchOnTaskbarMode` device policy with the per-user `SearchboxTaskbarMode` registry value that this mod actually reads. That caused the full box to be rejected and the narrow icon-and-label button to receive the overlay.

The fix removes the unnecessary OS-version split and accepts only per-user mode `2`. Debug and Release builds pass all three upstream tests, and the regenerated PR file differs only by this correction (`3 additions, 37 deletions`).

ShangHTao:
Addressed the geometry-gating review in `c85dd4c`, generated from upstream commit [`9a0cab5`](https://github.com/ShangHTao/TaskbarAudioSpectrum/commit/9a0cab5).

The registry value is no longer a visibility precondition:

- The locator queries the cached UI Automation `SearchButton` element regardless of whether `SearchboxTaskbarMode` is present, readable, policy-managed, or has an unexpected value.
- `HasUsableSearchBoxGeometry` now requires the control to have full-box proportions (`width >= 3 * height`), at least 64 DPI-scaled pixels of drawable spectrum width after the configured insets, and a non-empty drawable height.
- The published layout is cleared when that geometry test fails, so the overlay thread no longer reads or gates on `SearchboxTaskbarMode` at all.
- The registry value remains only as a diagnostic hint, a change notification source, and a conservative guard for stale-template fallback after a transient UIA failure.
- A rejected control logs once with the registry value (or `mode=unavailable`), raw UIA dimensions, and drawable dimensions.

Validation:

- Windows 11 LTSC 24H2 live full Search box: mode `2`, UIA `275x40`, overlay visible and painted.
- Narrow Search icon-and-label fixture: UIA `133x60`, rejected by the geometry test.
- All three tests pass in both Debug and Release; both configurations build without warnings.

PR #5067, opened 2026-08-11
Fix CJK Spacer XAML diagnostics startup race

aenerv7:
## Summary

Fix a confirmed startup race in the opt-in modern XAML path. When Explorer, Windhawk, and notification-area applications start concurrently, a diagnostics attempt can run before the corresponding XAML core is ready. The previous one-shot attempt then remains disabled until a mod reload.

The modern diagnostics path now:

- reacts to matching XAML host-window creation through narrowly filtered `CreateWindowExW`, `CreateWindowInBand`, and `CreateWindowInBandEx` hooks instead of DLL-load timing or timed polling;
- scans existing top-level and child windows in the current Explorer process from `Wh_ModAfterInit`, stopping once both XAML flavors are already queued or no longer need a trigger;
- maps `XamlExplorerHostIslandWindow` and a taskbar-owned `Windows.UI.Composition.DesktopWindowContentBridge` to Windows.UI.Xaml, and maps `CabinetWClass` plus `XamlExplorerHostIslandWindow_WASDK` to Microsoft.UI.Xaml;
- only signals a managed worker from window hooks, keeping COM activation and XAML Diagnostics work outside window-creation call stacks;
- coalesces Windows.UI.Xaml and Microsoft.UI.Xaml requests independently;
- allows hard connection failures to be retried by later matching host-window notifications, with at most three attempts per flavor;
- suppresses further attempts when a diagnostics call succeeds without creating a watcher, avoiding repeated conflicts with another Diagnostics consumer;
- resets only the affected flavor's attempt budget when an established connection disconnects;
- stops and joins the worker before unload, preserving the existing thread-lifetime guarantees.

The connection-name walk matches the range the reference mods use (up to 10,000 names), because the endpoint name embeds a process-lifetime core ordinal that is never reused and a live endpoint can sit at a high index. The cost of a failed walk is bounded by the walk count instead: after three empty walks in a session a flavor stops probing until an established connection disconnects. The walk runs only after a matching host-window or framework-module trigger indicates that the corresponding core should be ready.

## Verification

- Reproduced with Windhawk and NetEase UU Remote starting automatically.
- Confirmed that the tray tooltip uses the modern XAML path.
- Manually verified in Windhawk that host-window-driven initialization fixes the cold-start failure without toggling the mod setting.
- Compiled successfully for x86-64 and ARM64 after the latest trigger and retry-state changes.
- Local CJKSpacer spacing and lifecycle regression tests pass.
- Both maintained source copies are identical and `git diff --check` passes.

This update was implemented with AI assistance and manually verified on the affected system.

### DIFF
--- a/mods/cjk-spacer.wh.cpp
+++ b/mods/cjk-spacer.wh.cpp
@@ -1869,6 +1869,10 @@ struct XamlDiagnosticsRetry {
     }
 };
 
+// Keep the full compatibility range for initial and event-driven attempts,
+// which can encounter an already-running or reused XAML core. Timed startup
+// retries only probe whether a core has become ready, so a smaller range avoids
+// repeating the expensive full scan while Explorer starts.
 constexpr int kInitialXamlDiagnosticsConnectionLimit = 10000;
 constexpr int kRetryXamlDiagnosticsConnectionLimit = 256;
 
@@ -2535,10 +2539,14 @@ DWORD WINAPI XamlDiagnosticsWorkerProc(LPVOID parameter) {
             ResetXamlDiagnosticsAttempts(retries);
             ++retryIndex;
             delayedRetry = true;
-        } else {
+        } else if (waitResult == WAIT_FAILED) {
             Wh_Log(L"XAML diagnostics worker wait failed: %u",
-                   waitResult == WAIT_FAILED ? GetLastError()
-                                             : ERROR_GEN_FAILURE);
+                   GetLastError());
+            break;
+        } else {
+            Wh_Log(L"XAML diagnostics worker returned unexpected wait "
+                   L"result: %u",
+                   waitResult);
             break;
         }
 
--- a/mods/cjk-spacer.wh.cpp
+++ b/mods/cjk-spacer.wh.cpp
@@ -98,17 +98,13 @@ so another diagnostics-based customization tool, including Windows 11 Taskbar
 Styler or Windows 11 File Explorer Styler, can prevent this path from
 initializing. The mod waits for a recognized Windows.UI.Xaml or
 Microsoft.UI.Xaml host window before requesting the corresponding diagnostics
-connection. Window-creation hooks only signal a managed worker; an
-Explorer-process window scan covers hosts that already exist when the mod is
-enabled. There is no timed polling. Host notifications received during a
-connection attempt are coalesced into one follow-up attempt. Exhausting the
-connection-name walk with `ERROR_NOT_FOUND` means that the XAML endpoint isn't
-registered yet and doesn't consume the three-attempt hard-failure budget.
-Connections blocked by another diagnostics consumer aren't retried
-automatically. When an established connection is disconnected, that flavor is
-re-armed and waits for the next matching host window before reconnecting. All
-connection work runs outside the window-creation hooks and is stopped before
-the mod unloads.
+connection, and framework module loads provide a fallback signal when a host
+appears first. Connection work runs on a managed worker outside those hooks,
+and repeated unavailable endpoints or hard failures are bounded. Connections
+blocked by another diagnostics consumer aren't retried automatically. An
+established connection is re-armed after its XAML host disappears, and all
+connection work stops before the mod unloads. If the modern path's required
+hooks aren't available, enabled classic menus and tooltips remain active.
 When Taskbar Styler is configured to alert on competing XAML Diagnostics
 consumers, an intercepted connection attempt can show a confirmation dialog;
 if the tool blocks it by returning success, the walk stops and this mod marks
@@ -209,6 +205,7 @@ enum class CharacterKind {
 std::atomic_bool g_classicMenus{true};
 std::atomic_bool g_classicTooltips{true};
 std::atomic_bool g_modernUiText{false};
+std::atomic_bool g_modernUiActive{false};
 std::atomic_bool g_unicodeLettersAndDigits{true};
 
 bool IsHighSurrogate(wchar_t value) {
@@ -1873,6 +1870,7 @@ enum class XamlDiagnosticsFlavor {
 // core. Its diagnostics endpoint can still be registered slightly later.
 constexpr int kXamlDiagnosticsConnectionLimit = 10000;
 constexpr unsigned int kXamlDiagnosticsMaxAttempts = 3;
+constexpr unsigned int kXamlDiagnosticsMaxEmptyWalks = 5;
 
 struct XamlDiagnosticsConnectionState {
     std::atomic_bool connected{false};
@@ -1880,7 +1878,18 @@ struct XamlDiagnosticsConnectionState {
     std::atomic_uint requestGeneration{0};
     std::atomic_uint processedGeneration{0};
     std::atomic_uint failureCount{0};
+    std::atomic_uint emptyWalkCount{0};
 };
+
+bool CanAttemptXamlDiagnosticsConnection(
+    const XamlDiagnosticsConnectionState& state) {
+    return !state.connected.load(std::memory_order_acquire) &&
+           !state.blocked.load(std::memory_order_acquire) &&
+           state.failureCount.load(std::memory_order_acquire) <
+               kXamlDiagnosticsMaxAttempts &&
+           state.emptyWalkCount.load(std::memory_order_acquire) <
+               kXamlDiagnosticsMaxEmptyWalks;
+}
 
 bool RequestModernXamlDiagnosticsInitialization(
     XamlDiagnosticsFlavor flavor);
@@ -2389,10 +2398,7 @@ void EnsureXamlDiagnosticsConnection(
     LPCWSTR displayName,
     XamlDiagnosticsFlavor flavor) {
     if (g_stoppingModernUi.load(std::memory_order_acquire) ||
-        state.connected.load(std::memory_order_acquire) ||
-        state.blocked.load(std::memory_order_acquire) ||
-        state.failureCount.load(std::memory_order_acquire) >=
-            kXamlDiagnosticsMaxAttempts) {
+        !CanAttemptXamlDiagnosticsConnection(state)) {
         return;
     }
 
@@ -2414,6 +2420,7 @@ void EnsureXamlDiagnosticsConnection(
                 std::memory_order_acquire) > watcherGeneration) {
             state.connected.store(true, std::memory_order_release);
             state.failureCount.store(0, std::memory_order_release);
+            state.emptyWalkCount.store(0, std::memory_order_release);
             Wh_Log(L"Connected to %s diagnostics", displayName);
         } else {
             // A competing diagnostics hook can return success without
@@ -2426,10 +2433,21 @@ void EnsureXamlDiagnosticsConnection(
                    displayName);
         }
     } else if (result == HRESULT_FROM_WIN32(ERROR_NOT_FOUND)) {
-        // The module exists, but no XAML core has registered a diagnostics
-        // endpoint yet. A later host notification will request another walk;
-        // don't spend the hard-failure budget on this startup state.
-        Wh_Log(L"%s diagnostics endpoint isn't registered yet", displayName);
+        const unsigned int emptyWalkCount =
+            state.emptyWalkCount.fetch_add(
+                1, std::memory_order_acq_rel) +
+            1;
+        if (emptyWalkCount >= kXamlDiagnosticsMaxEmptyWalks) {
+            Wh_Log(L"%s diagnostics endpoint remained unavailable after "
+                   L"%u full connection-name walks; stopping automatic "
+                   L"attempts",
+                   displayName, emptyWalkCount);
+        } else {
+            Wh_Log(L"%s diagnostics endpoint isn't registered yet "
+                   L"(empty walk %u/%u)",
+                   displayName, emptyWalkCount,
+                   kXamlDiagnosticsMaxEmptyWalks);
+        }
     } else if (result != HRESULT_FROM_WIN32(ERROR_CANCELLED)) {
         const unsigned int failureCount =
             state.failureCount.fetch_add(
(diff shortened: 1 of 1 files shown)


aenerv7:
This is a follow-up to #4950 for a confirmed cold-start race in the modern XAML path.

With Windhawk and NetEase UU Remote both starting automatically, the first XAML Diagnostics attempt could run before the endpoint was ready and return `ERROR_NOT_FOUND`. Toggling the modern UI option reloaded the mod after startup and made the tooltip work, confirming the connection timing issue. The updated worker now retries only `ERROR_NOT_FOUND` with bounded, interruptible backoff. Successful calls without a watcher are still treated as possible Diagnostics conflicts and aren't retried.

The fix was verified through repeated cold-start testing. Unrelated classic Tooltip experiments were reverted before this PR.

aenerv7:
Addressed the worthwhile review suggestions in e9279c7:

- Replaced the retry bitmask helpers with a small two-boolean `XamlDiagnosticsRetry` struct.
- Kept the initial 10,000-name compatibility scan, but limited each delayed retry round to 256 names. Across both XAML flavors, the maximum startup work drops from 140,000 calls to 23,072 (20,000 initial + 3,072 delayed).
- Documented that an explicit module-load/reconnect wake intentionally restarts the pending flavors' backoff ladder because it represents a new opportunity.
- Trimmed the embedded user-facing README while retaining the startup retry behavior and diagnostics-consumer limitation.

I did not add a `CreateWindowExW` hook: broadening the hook surface isn't justified for this startup race, and the existing relevant-module/disconnect triggers remain the deliberate post-backoff retry points.

Validation: standalone regression tests pass; the mod compiles successfully for x86-64 and ARM64; `git diff --check` passes.

aenerv7:
Addressed the two remaining optional points in 677bc8a:

- Added an inline rationale for keeping the 10,000-name range on initial/event-driven attempts while limiting timed startup retries to 256. The full range preserves compatibility with an already-running or reused XAML core; the smaller repeated probe avoids duplicating the expensive full scan during Explorer startup.
- Split the wait error handling into an explicit `WAIT_FAILED` branch using `GetLastError()` and a separate unexpected-result log, removing the dead `ERROR_GEN_FAILURE` fallback.

No retry or connection behavior changed. Validation remains green: spacing regression tests pass, the mod compiles for x86-64 and ARM64, and both diff checks pass.

aenerv7:
Reworked the modern XAML initialization path in 7fa2453 to address both main findings:

- Removed the `LoadLibraryExW` trigger, timed backoff ladder, retry bookkeeping, and repeated reduced-range scans.
- Added narrowly filtered hooks for `CreateWindowExW`, `CreateWindowInBand`, and `CreateWindowInBandEx`. They react only to `Windows.UI.Composition.DesktopWindowContentBridge`, `XamlExplorerHostIslandWindow`, and `XamlExplorerHostIslandWindow_WASDK`.
- Added a current-process `EnumWindows`/`EnumChildWindows` scan in `Wh_ModAfterInit` so hosts that predate hook activation are covered.
- Window hooks only classify the host and signal the managed worker; COM activation and XAML Diagnostics connection work never run in the window-creation call stack.
- Windows.UI.Xaml and Microsoft.UI.Xaml requests are queued independently, so one flavor's host cannot prematurely probe the other flavor. Duplicate requests are coalesced, and a disconnected established connection still re-enables only its own flavor.
- Kept the 10,000-name compatibility range, but it is now walked only after the matching XAML host proves that the core is ready and normally stops at the first available name. The README now explicitly notes that an alerting diagnostics consumer can show one or more prompts while occupied names are skipped.
- Removed the now-vestigial initialization mutex and re-entrancy guard.

Validation:

- The updated behavior was manually verified in Windhawk and the previously affected modern XAML tooltip still works.
- Standalone spacing/lifecycle regression tests pass.
- The mod compiles successfully for x86-64 and ARM64.
- Both source copies are identical and `git diff --check` passes.

aenerv7:
Addressed the latest review findings in 506479b and verified the result locally in Windhawk:

- Added `CabinetWClass` as an early Microsoft.UI.Xaml trigger.
- Restricted `Windows.UI.Composition.DesktopWindowContentBridge` to the taskbar case where its parent is `Shell_TrayWnd`; unrelated composition bridges no longer trigger the Windows flavor.
- Replaced the permanent generic `attempted` latch with per-flavor connection state: connected, blocked by another consumer, pending, and hard-failure count.
- Hard failures can now be retried by later matching host-window notifications, with a strict maximum of three attempts per flavor.
- A successful diagnostics call that creates no watcher is still latched as blocked, preventing repeated conflict prompts.
- An established connection disconnect resets only its own flavor's failure budget.
- Existing-window enumeration stops once both flavors are already pending or no longer need another host trigger.
- Updated the embedded and standalone README text, regression tests, and the PR description to match the final host-event-driven design.

The updated build was manually tested in Windhawk and the affected modern XAML tooltip/context-menu behavior remains working. The mod also compiles for x86-64 and ARM64, all local regression tests pass, all GitHub checks pass, and both maintained source copies are identical.

aenerv7:
Addressed the latest review in 3ca3cd9 and verified the result locally in Windhawk:

- A XAML diagnostics disconnect now only re-arms the affected flavor and waits for the next matching host window; it no longer reconnects immediately from SetSite(nullptr).
- Duplicate same-flavor host notifications are coalesced for the full duration of an in-progress connection attempt.
- Added the low-risk cleanup items: steady-state notification early exit, shared host classification, Shell_InputSwitchTopLevelWindow support, and missing-module diagnostics logging.
- Updated the embedded documentation, standalone README, and regression tests.

The maintained source copies are identical, all local regression tests pass, and the mod compiles successfully for x86-64 and ARM64.

aenerv7:
Addressed the latest review in 3c63551:

- Replaced the in-flight request boolean with per-flavor request/processed generations. Host notifications received during a diagnostics walk are retained and coalesced into exactly one follow-up attempt instead of being swallowed.
- Treat a fully exhausted connection-name walk returning ERROR_NOT_FOUND as a not-yet-registered XAML endpoint, so it no longer consumes the three-attempt hard-failure budget. Other real failures remain bounded.
- Preserved disconnect behavior: only the affected flavor is re-armed and waits for a new matching host.
- Removed duplicate class-name resolution from GetModernXamlHostFlavor.
- Corrected the documentation: a competing tool that blocks by returning success stops the walk and latches the flavor as blocked; occupied names aren't described as being skipped.
- Updated the standalone tests and documentation.

Local regression tests pass, the mod compiles for x86-64 and ARM64, git diff checks pass, and both maintained source copies are identical.

aenerv7:
Implemented and pushed in 2dec5b4 (parent repository e4687c5). Mandatory fix: map CoreMessagingXP.dll to the Microsoft XAML diagnostics flavor so the static-import path is covered. Optional cleanup: initialize the diagnostics worker before installing modern UI hooks, remove the pass-through readiness helper, and keep module path parsing to the LoadLibraryExW backslash form. The five-empty-walk limit remains intentionally unchanged: with 10,000 names per walk, it bounds automatic probing at 50,000 InitializeXamlDiagnosticsEx calls per flavor while retaining several cold-start opportunities. Local tests and x86-64/ARM64 builds pass.

aenerv7:
Addressed the latest review in d9a18b5 (parent repository 04dfae4). Each connection-name walk is now limited to 64 dense low indices. Five consecutive empty walks pause that flavor for 30 seconds instead of permanently latching it, so one automatic burst is deliberately bounded at 320 InitializeXamlDiagnosticsEx calls per flavor while a later host or module event can recover after slow startup. The cooldown and disconnect reset paths have regression coverage. Also removed the unreachable user32 branch, reduced repeated enumeration-state reads, made the queued-first-request logging semantics explicit, renamed the cleanup dispatch-window classifier, and simplified the user-facing modern-path documentation. Local tests pass and the mod builds for x86-64 and ARM64.

aenerv7:
Addressed the latest review in 777281f:

- \RetryXamlDiagnosticsAfterCooldown\ now retries only the flavor that exhausted its empty-walk budget and scheduled the cooldown timeout. A flavor still waiting for its first host or module trigger (emptyWalkCount == 0) is no longer probed from another flavor's retry, so it can't spend its walk budget (64 \InitializeXamlDiagnosticsEx\ calls, plus styler alert dialogs when those mods run in \lert\ mode) on a connection nothing asked for.
- Added a regression test covering the retry flavor selection and the retry cap; the test fails on the previous code and passes now. Local tests pass, and the mod compiles for x86-64 and ARM64. Both maintained source copies are identical.

aenerv7:
Addressed the review in 341cfed:

- **Walk cap lowered to 8** (\kXamlDiagnosticsConnectionLimit\). DXamlCore allocates dense low indices and a registered core takes the first available name, so an unregistered endpoint fails every index identically; the extra probes only bought modal Taskbar Styler / File Explorer Styler alert dialogs in the cold-start case. The PR description now notes the cap is deliberate.
- **Explicit ticks**: the empty-walk budget predicates take \currentTick\ explicitly instead of a default argument that silently re-read \GetTickCount64()\ per call (raised twice before as optional).
- **\RequestModernXamlDiagnosticsInitialization\ is now \oid\**; its bool only gated a debug log line, and the trigger log lines now fire unconditionally (no-op when logging is off).
- **Null-guarded** the \GetProcAddress\ calls for \CreateWindowInBand[Ex]\ against a null user32 handle.

On the inline-connection suggestion (kept as a deliberate choice rather than a change): the worker stays, because the module-load trigger (\LoadLibraryExWHook\) runs under the loader lock and must defer diagnostics work anyway, and routing window triggers through the same serialized worker keeps connection attempts per flavor single-streamed (the generation handshake coalesces bursts) and keeps any modal dialogs a styler raises off window-creation call stacks, which can fire on any thread at any time. The 8-name cap plus the existing budgets keep the deferred cost bounded. I'm not attached to the machinery for its own sake — if a later round wants to revisit it, the regression tests cover the current behavior.

aenerv7:
Addressed the review in adba837:

- **Cooldown timer path removed.** The worker waits \INFINITE\ again; host-window and framework-module events are the only retry triggers (\XamlDiagnosticsWorkerTimeout\, \RetryXamlDiagnosticsAfterCooldown\, \IsXamlDiagnosticsFlavorCoolingDown\ and the retry cap are gone). The 5-empty-walk budget and its 30-second cooldown remain as a burst bound, resumed by the next host or module event. The readme now says the pause ends at the next host or module event rather than on a timer.
- **Generations collapsed to one \pending\ flag per flavor** — stored with release ordering before \SetEvent\, exchanged (\cq_rel\) by the worker at wake-up. Same coalescing guarantee, no mutex-protected snapshot, and no stale write-back after a disconnect.
- **\locked\ no longer latches on our own failures.** \SetSite\ bumps a new \g_tapSiteGeneration\; \locked\ is set only when *neither* the site nor the watcher generation moved (a competing consumer that returned success without loading the TAP). A moved site generation with no watcher is treated as a retryable \ailureCount\ attempt.
- Also: the \Wh_ModAfterInit\ child-window walk is restricted to the tray window, and the empty-walk budget resume log names its flavor.

Per-flavor state is now \connected\/\locked\/\pending\/\ailureCount\/\mptyWalkCount\/\lastEmptyWalkTick\ — down from eight atomics plus six predicates. Local tests pass (regression suite updated for the removed machinery), the mod compiles for x86-64 and ARM64, and both maintained copies are identical.

On the \kXamlDiagnosticsConnectionLimit = 8\ note: keeping 8 for now — the walk exits at the first non-\ERROR_NOT_FOUND\ name and the live-system testing so far matches the dense-low-index assumption. If a long session ever shows the Microsoft flavor stopping at a higher index, remembering the last worked index or widening the walk on an explicit host trigger are both small follow-ups; flagging it in case anyone sees it in the field.

aenerv7:
Addressed the review in 57f4646:

- **Walk restored to 10,000 names** (matching Taskbar Styler). The endpoint name embeds a process-lifetime core ordinal from \DXamlCore::s_instanceCount++\ that is never reused, so the 'dense low index' premise was wrong: with a cap of 8, the Microsoft flavor would stop connecting after roughly eight Explorer windows in a session. Comment and PR description updated.
- **Cost now bounded by walk count instead**: \kXamlDiagnosticsMaxEmptyWalks\ is 3, and the latch it produces is permanent until \RearmXamlDiagnosticsConnection\ clears it on a real disconnect — the 30-second cooldown resume and its tick bookkeeping are gone, so a flavor that never registers an endpoint probes at most a few times per session, and the \ClassifyModernXamlHost\ per-window churn for a never-settling flavor ends once the latch trips.
- \CanAttemptXamlDiagnosticsConnection\ no longer takes a tick; \RequestModernXamlDiagnosticsInitialization\ returns whether it queued an attempt, and the host/module trigger logs fire only when one was queued.
- The modern-path readme paragraph is reduced to user-actionable behavior.

Tests updated for the latch semantics; x86-64 and ARM64 builds pass; both maintained copies are identical.

aenerv7:
Addressed the review in dd15d1f:

- Module loads now set a coalesced host-discovery request and wake the diagnostics worker instead of directly queuing a 10,000-name connection walk.
- The worker runs `DiscoverExistingModernXamlHosts()` outside the loader call stack; only a matching live host queues the flavor's connection attempt, so pre-host module loads no longer consume the permanent empty-walk budget or create an alert-dialog scan.
- `SetEvent` failures now roll back only a newly introduced pending flag, preserving any already-queued concurrent request. Initialization and teardown also reset the new discovery flag.
- Updated the embedded and standalone behavior notes and extended worker-cleanup regression coverage.

Validation: the standalone regression suite passes, the mod compiles successfully for x86-64 and ARM64 with Windhawk 1.7.3, both maintained source copies are identical, and both diff checks pass.

aenerv7:
Addressed the follow-up review in e8739c9:

- Host discovery now runs before the worker exchanges the per-flavor pending flags. A host found by discovery is therefore consumed by the current iteration instead of leaving an identical attempt queued for the next wake.
- A wake carrying both module-driven discovery and an existing host request now produces exactly one connection walk per flavor, preserving the three empty-walk slots for distinct retry opportunities.
- Trimmed the embedded user-facing README back to observable connection behavior; the standalone maintainer README retains the implementation details.

I left the optional table-driven/state-descriptor refactor and the watcher flavor ownership change out of this focused fix; neither is needed for the reviewed duplicate-attempt bug.

Validation: standalone regression tests pass; the mod compiles for x86-64 and ARM64 with Windhawk 1.7.3; both maintained copies are identical; both diff checks pass.

PR #5081, opened 2026-08-12
add taskbar-icon-separators.wh.cpp

Meteony:
## Changelog
# Taskbar Icon Separators

Create tracked icon separators with configurable padding on the taskbar.

Screenshot
_Example for creating separators on the taskbar_

Another mod offers similar functionality, but this implementation takes a different approach.

This mod uses private COM APIs to insert a genuine taskbar button, styles its width and centering, and disables all interaction events. All separators are removed on unload. 

As an additional bonus, we achieve separator tracking with zero manual tracking code. 

Windows 11 only.

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini


Meteoni:
> "The mod makes a persistent change to the user's taskbar pin list"

The maintainer and I have discussed this previously. Both agreed that this is an acceptable trade-off. 

> "Overlap with the existing Taskbar Separators mod"

The architecture is completely different. Our implementation allows for real custom padding and zero manual tracking. Integration isn't feasible. There's pros and cons to either approach, and the final call should be left to the end user.

mods/taskbar-icon-separators.wh.cpp
```diff
@@ -0,0 +1,3453 @@
...
+    // Stop applying separator state, then restore every tracked container on
+    // its owning taskbar UI thread before Windhawk removes the hooks.
+    g_unloading = true;
+    RestoreTrackedSeparatorVisualStates();
+}
+
+void Wh_ModUninit() {
+    Wh_Log(L"[UNINIT] Taskbar Icon Separators unloading");
+
+    // Defensive in case an older Windhawk build skips Wh_ModBeforeUninit.
```
m417z, line 3432: What builds skip `Wh_ModBeforeUninit`?

mods/taskbar-icon-separators.wh.cpp
```diff
@@ -0,0 +1,3453 @@
...
+    }
+
+    g_backendThread =
+        CreateThread(
+            nullptr,
+            0,
+            BackendThreadProc,
+            g_backendStopEvent,
+            0,
+            nullptr);
```
m417z, line 3271: Perhaps hooking a function such as `CreateWindowExW` or `TrayUI::StartTaskbar` will be simpler and more robust.

m417z (commented):
Looks great, thanks!

I also thought that a more dynamic approach could be nice. For example, having a way to add a separator dynamically (e.g. on taskbar middle click, or a menu item like [Taskbar Restart Explorer](https://windhawk.net/mods/taskbar-restart-explorer) does), and to remove a separator (e.g. on middle click or via a menu).

PR #5088, opened 2026-08-13
Classic Display Control Panel Restorer mod

babamohammed2022:
This PR adds the Classic Display Control Panel Restorer mod, which brings back the classic Display and Screen Resolution Control Panel pages on Windows 10 and 11 by downloading the Windows 10 1511 display.dll from Microsoft's debug servers and applying a visual change to make it similar to the original Windows 7 one. Please point out all problems so that I can make it more stable

## Changelog

* Changelog item 1...
* Changelog item 2...

## Mod authorship

This mod was created by:

- - [X] Claude
- - [ ] ChatGPT
- - [ ] Gemini


babamohammed2022:
I'll add the pictures later, I've put the links as reminders because the interface of the mod is getting constantly updated.
/ai-review

babamohammed2022:
The magnifier.exe problem will be addressed in a future update. At the moment it is not functional.
/ai-review

babamohammed2022:
Before proceeding to fix the problems, I'd like to know if these points are worth addressing from a human reviewer. This will help me understand further what is the right direction for these fixes
/ready-for-reviewer

babamohammed2022:
Hello, @m417z. I'd like to know if the issues reported above need to be addressed as I think that having separate mods is better for the user experience (for example a user who already has a Windows 8-like environment has not yet restored performance and information tools and could use my mod but if he has already the restored display page, there could be a conflict even if there is a toggle to choose between which page to restore as the underlying logic is essentially the same so if something breaks even the other pages could break). Thanks for your time.

babamohammed2022:
I've addressed most of the technical problems. However, the overlap with the existing mod will be fixed in the other "Redirect Settings to Control Panel" instead rather than applying unnecessary changes here.
/ai-review

babamohammed2022:
I think that most of these problems as noted by the PR are not real code issues therefore I believe that it would be better to see which one is really worht addressing in a human review after 15 commits
/ready-for-reviewer

babamohammed2022:
After 18 commits, I'd like to know from a human reviewer if these problems are really worth addressing.
/ready-for-reviewer

babamohammed2022:
This is my justification towards the review:

1. DLL download: The download is necessary because the file can't be distributed with the mod for licensing and size reasons. The DLL is downloaded from Microsoft, verified with SHA‑256, and never executed if it doesn't match the expected hash. This approach has already been approved in other Windhawk mods (my performance-info-tools-restorer, for example). If the maintainer considers this a violation, I'm open to discussing alternatives, but currently this is the only practical way to have the original Control Panel page.

3. DLL not unloaded: The DLL stays loaded because an open Control Panel page might still reference it. Unloading it forcibly could crash Explorer. I've added a note in the README that files are removed on the next Explorer restart. This is a compromise between safety and stability in my opinion

6. Conflict with other mods: This mod is specific to the Display Control Panel page, while win7-legacy-applet-restorer is a more generic mod. I've clarified the differences in the description and I'm planning a separate PR to resolve the conflict with settings-to-control-panel soon, but if these PRs are not addressed yet it could be a waste of time for me and the maintainer.

About the points n.2-4-5, the mod seems to work properly for me. If the maintainer believes otherwise, I'll adjust them.

/ready-for-reviewer

PR #5101, opened 2026-08-15
add prevent-virtual-desktop-stealing.wh.cpp

Meteony:
## Changelog
# Prevent Window Activation from Stealing Virtual Desktop

Take back control over your virtual desktops!

Screenshot
_Effect of redirecting window activation_

Windows can switch you to another virtual desktop just because a window that was left open on that
desktop is activated. 

Clicking a flashing taskbar button, following a notification, opening an app through its tray icon, 
or launching something that insists on reusing an existing window can all unexpectedly pull you away 
from your current desktop. This mod keeps you where you are and brings the activated window to you instead.

If you don't know what "virtual desktop stealing" is, try this small experiment:

- Open Windhawk on your first virtual desktop.
- Switch to another virtual desktop.
- Click the Windhawk tray icon.

Normally, Windows jumps back to the first desktop - not anymore with this mod.

Intentional desktop navigation is left alone, including Win+Ctrl+Left/Right and
Task View. The mod also remains compatible with **Disable Virtual Desktop
Transition Animation**.

### Notes

Designed for Windows 11 24H2 and newer. This mod relies on undocumented Windows
shell interfaces, so future Windows updates may require adjustments.

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/prevent-virtual-desktop-stealing.wh.cpp
+++ b/mods/prevent-virtual-desktop-stealing.wh.cpp
@@ -1,0 +1,1887 @@
+// ==WindhawkMod==
+// @id              prevent-virtual-desktop-stealing
+// @name            Prevent Window Activation from Stealing Virtual Desktop
+// @description     Redirect cross-desktop window activation to the current desktop.
+// @version         0.3.3
+// @author          meteoni
+// @include         explorer.exe
+// @compilerOptions -lole32
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Prevent Window Activation from Stealing Virtual Desktop
+
+Take back control over your virtual desktops!
+
+![Screenshot](https://raw.githubusercontent.com/Meteony/meteoni-assets/main/prevent-virtual-desktop-stealing/prevent-virtual-desktop-stealing.gif)
+_Effect of redirecting window activation_
+
+Windows can switch you to another virtual desktop just because a window that was left open on that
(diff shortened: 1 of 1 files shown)


Meteoni:
Finding 3 is not valid.  The mod is 100% compatible with Virtual Desktop Helper and Virtual Desktop Disable Animations, which both are installed and tested on my work machine. The current block-and-redirect policy is already set up in a way that excludes the paths enacted by those two mods.

Meteoni:
For finding 2: IVirtualDesktopManager being in-process does not make this call reentrancy-safe. Calling GetWindowDesktopId synchronously from inside SwitchDesktopInternal deterministically deadlocks Explorer’s virtual-desktop machinery, confirmed by repeated AppHangB1 reports on my machine.

Meteoni:
The retry path is still necessary, but not primarily for unload-related failures. On a fresh Explorer start, Shell_TrayWnd can be created before the Immersive Shell COM services are registered.

Meteoni:
I've seen none of those aborts personally. With the blast radius greatly narrowed down as-is, those should be way lesser of a concern. 
Adding replay attempts would be a costly refactor that adds complexity for not a lot of gain. The fix initially proposed by Claude deterministically crashed explorer.

Meteoni:
Time to wait again

PR #5102, opened 2026-08-15
New mod "fix-white-flash"

JoyHak:
Fixes white flashes when opening new window. Common issue on all Windows versions prior to 26h1
m2-res_1080p-ezgif com-video-to-gif-converter

## Mod authorship

This mod was created by:

- - [X] The submitter, without AI assistance
- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/fix-white-flash.wh.cpp
+++ b/mods/fix-white-flash.wh.cpp
@@ -1,8 +1,8 @@
 // ==WindhawkMod==
-// @id              fix-white-flash
+// @id                fix-white-flash
 // @name            Fix white flashes for all windows
 // @description     Fixes white flashes when opening new window.
-// @version         0.2
+// @version         0.1
 // @author          Rafaello
 // @github          https://github.com/JoyHak
 // @include         *
--- a/mods/fix-white-flash.wh.cpp
+++ b/mods/fix-white-flash.wh.cpp
@@ -1,0 +1,195 @@
+// ==WindhawkMod==
+// @id              fix-white-flash
+// @name            Fix white flashes for all windows
+// @description     Fixes white flashes when opening new window.
+// @version         0.2
+// @author          Rafaello
+// @github          https://github.com/JoyHak
+// @include         *
+// @compilerOptions -lGdi32
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+Fixes white flashes when opening new windows.
+
+### Before
+![Before](https://raw.githubusercontent.com/MGGSK/MGGSK/refs/heads/main/WindhawkModReadmeImages/fix-explorer-white-flash-before.png)
+
+### After
+![After](https://raw.githubusercontent.com/MGGSK/MGGSK/refs/heads/main/WindhawkModReadmeImages/fix-explorer-white-flash-after.png)
+
+*/
+// ==/WindhawkModReadme==
+
+#include <unordered_map>
+#include <windhawk_utils.h>
+
+decltype(&DefWindowProcA) DefWindowProcA_Original = nullptr;
+decltype(&DefWindowProcW) DefWindowProcW_Original = nullptr;
+decltype(&DefDlgProcA)    DefDlgProcA_Original    = nullptr;
+decltype(&DefDlgProcW)    DefDlgProcW_Original    = nullptr;
+
+using DefProcCallback = LRESULT (WINAPI *)(HWND, UINT, WPARAM, LPARAM);
+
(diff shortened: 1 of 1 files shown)


Anixx:
How is this different from https://windhawk.net/mods/fix-explorer-white-flash ?

JoyHak:
1. 
> WM_NCPAINT destroys the caller's update region and then hands the freed handle to the original proc
> ...
> Pass a copy instead, and remember ownership only transfers on success:

Current code [was suggested by Claude review earlier](https://github.com/ramensoftware/windhawk-mods/pull/5102#issuecomment-5301459181):
```cpp
HRGN hrgn = (HRGN)wParam;
HDC hdc = GetDCEx(hWnd, hrgn == (HRGN)1 ? NULL : hrgn,
                  DCX_WINDOW | DCX_USESTYLE | (hrgn == (HRGN)1 ? 0 : DCX_INTERSECTRGN));
```
I'm not sure it needs to be changed...

2. 
> Build the new brush first, swap it in, then delete the old one — and make the global an std::atomic<HBRUSH>

I need an expert review before including `st::atomic`. Right now, I see this as an unnecessary complication of the code.

> drop the unload() call at the top of load(). Add a `if (!Cfg::get()) break;` guard in `FillWindow` so a `null` brush can never reach FillRect.

`null` brush will never reach `FillWindow()` if `Cfg::load()` successfully created a brush and returned `true`.

To ensure this behavior, I also use the `BOOL Wh_ModSettingsChanged(BOOL*)` (not the `Wh_ModSettingsChanged(void)`)  signature that allows the mod to be unloaded if a brush was not created after `Cfg::load()` is call.

3. 
> The `@exclude` list looks like your personal machine

It excludes CLI applications, Windhawk development toolchain and modern Nvidia applications. This reduces the number of messages displayed during mod compilation when logging is enabled. I might add some AMD apps in the future...

4. 
> Mention the relationship to the existing mod in the README.

The mod is significantly different, and this is reflected in the metadata. 

---

> Use `WindhawkUtils::StringSetting` instead of the hand-rolled `defer`

`defer` is simpler and helps me write more readable code and perform logging after the function `return` (see `Cfg::parseColor()`). It can also be used for *handles* in the future

> Missing includes. `std::transform` needs `<algorithm>`, `std::towlower/iswspace` need `<cwctype>`, `UINT_MAX` needs `<climits>` ... These currently work only via transitive includes.

For now, this will extend the list of imports, but it won't change the behavior. [In the LLVM 23 the number of transitive imports was reduced](https://libcxx.llvm.org/ReleaseNotes.html):

> In LLVM 23, libc++ dropped a lot of transitive includes in all language modes. This improves compile time significantly, but causes programs which rely on these includes to not compile anymore. The `_LIBCPP_KEEP_TRANSITIVE_INCLUDES_LLVM23 ` macro that was provided in LLVM 23 to ease the transition has been removed in this release.

If it will be included in Windhawk, I will change the imports to make it more explicit. 

> A listed process with a blank `backgroundColor` gets `kDefaultColor`, not the global color.

To return a global color from `Cfg::parseColor()` instead of `kDefaultColor` I need to make the code more complicated (and make it less readable).

> Entries are only removed on `WM_NCDESTROY`, which the mod only sees if the app forwards it to `DefWindowProc*`. Since HWND values are recycled, a leftover kPainted entry can make the mod silently skip a brand-new window.

Expert guidance is needed for the periodic deletion of stale windows from `Win::s_windows`.

*Some color parsing issues were fixed below.*

JoyHak:
I'm sure it's /ready-for-reviewer

JoyHak:
I'm definitely sure it's /ready-for-reviewer

JoyHak:
# To reduce cluttering [see my response above](https://github.com/ramensoftware/windhawk-mods/pull/5102#issuecomment-5385727423). Last AI review duplicates previous reviews.

m417z:
Nice mod, thanks for the submission.

Some of the review items are duplicated, but I don't think everything is answered, and most findings seem correct. I think it'd be better to address them.

1. Seems like a real issue. You said "Current code [was suggested by Claude review earlier](https://github.com/ramensoftware/windhawk-mods/pull/5102#issuecomment-5301459181)", but I don't see where it suggested calling `ReleaseDC` and then passing the freed handle to the original `DefWindowProc`.
2. I don't see where it was addressed.
3. I don't see where it was addressed. I agree with the finding and I think it'd be simpler and safer to only handle window creation.
4.-6 I don't see where it was addressed.
7\. I agree that this list is not necessary but I don't mind, it's just not pretty.

PR #5104, opened 2026-08-15
Windows 11 Start Menu Styler v1.7

m417z:
* Added the following themes: OnlySearch (Minimal), LiquidGlass, FrostyGlass.
  * The previous LiquidGlass theme is now discontinued. It was renamed to "LiquidGlass (Legacy)".
* Updated the following themes: SideBySide, SideBySide2, SideBySideMinimal, Down Aero, Windows10, Windows11_Metro10, Oversimplified&Accentuated, Command Center.
* Several target controls can now be specified for the same styles by separating them
with commas, for example: `ParentClass > Class#Name1, ParentClass >
Class#Name2`.
* Added string literals support for substitution expressions, for example: `` Text={{var == `a` ? `b` : `c`}} ``.
* Style variables now get the value from whichever capturing control is closest. Previously, the value was undefined if there was more than one control publishing the same variable.
* Improved handling for remote images.

### DIFF
--- a/mods/windows-11-start-menu-styler.wh.cpp
+++ b/mods/windows-11-start-menu-styler.wh.cpp
@@ -2,7 +2,7 @@
 // @id              windows-11-start-menu-styler
 // @name            Windows 11 Start Menu Styler
 // @description     Customize the Start menu with themes contributed by others or create your own
-// @version         1.6
+// @version         1.7
 // @author          m417z
 // @github          https://github.com/m417z
 // @twitter         https://twitter.com/m417z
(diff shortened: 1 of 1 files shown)


csyclrk:
Hi mate,

Thanks for your software, I rely on Start Menu Styler for 11 on both my desktop and laptop to ditch the god awful modern start menu and keep closer to Windows 10

I just updated both devices this morning, and suddenly the start button looked off. Under Windows 10 Theme, the Layout just read 1.

So I switched both to Default for the selected theme, and it looks as expected but the problem now is that 8 grouped pinned items only appear on both machines and the rest are hidden. If I unpin and repin a program, they will show again but once closing Start and reopening they disappear again.

Found a way to rollback to an earlier version after searching more, and I'm no coder really, but thought I'd raise it in case you weren't aware.

m417z:
@csyclrk The start button is part of the taskbar, not the start menu, so it's not related to this mod.

Please create a new issue here, specify the theme, tag the theme author, and post screenshots. Thanks.
https://github.com/ramensoftware/windows-11-start-menu-styling-guide/issues

PR #5106, opened 2026-08-15
Add On-Screen Indicator Position mod

mario0318:
new mod. windows only gives you three spots for the volume/brightness indicator (top left, top center, bottom center) and i wanted it in a corner, so this opens it up to all nine plus pixel offsets and a monitor picker.

internally windows calls it the Hardware Confirmator. explorer draws it and the thread that owns the window names itself "HardwareConfirmator UI Thread", so the mod matches on that instead of a window class, since class names tend to move around between builds. placement runs in a MoveWindow hook, i hooked SetWindowPos too but nothing appears to go through it on 26200.

only tried it on 26200 x64 so far.

## Changelog

* Changelog item 1...
* Changelog item 2...

## Mod authorship

This mod was created by:

- - [x] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/on-screen-indicator-position.wh.cpp
+++ b/mods/on-screen-indicator-position.wh.cpp
@@ -2,12 +2,11 @@
 // @id              on-screen-indicator-position
 // @name            On-Screen Indicator Position
 // @description     Place the volume/brightness/camera on-screen indicator anywhere on the screen, not just the three positions Windows offers
-// @version         1.0.3
+// @version         1.1.0
 // @author          mario0318
 // @github          https://github.com/mario0318
 // @include         explorer.exe
 // @architecture    x86-64
-// @compilerOptions -lshcore
 // ==/WindhawkMod==
 
 // Source code is published under The GNU General Public License v3.0.
@@ -25,9 +24,8 @@ changes. Under **Settings > System > Notifications > On-screen indicators**,
 Windows lets you put it in one of three places: top left, top center, or bottom
 center.
 
-This mod replaces that with a full nine-point grid — any corner, any edge
-center, or dead center of the screen — plus a pixel offset for fine-tuning and a
-choice of which monitor it appears on.
+This mod replaces that with a full nine-point grid, any corner, any edge center,
+or dead center, plus a pixel offset for fine-tuning.
 
 The brightness indicator moved to the middle of the right edge:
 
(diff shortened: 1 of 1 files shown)


mario0318:

mario0318:
pushed 1.1.0. the monitor setting is gone entirely, it overlapped volume-control-open-location and that one does it better with the interface names, so the readme sends people there instead and says the two work alongside each other. what's left is the nine point grid and the offsets.

also moved it onto GetPositionRect the way that mod does, which took out the process wide user32 hooks, the thread name matching and the SetWindowPos fallback that was still carrying the monitor feedback bug. about 350 lines lighter.

@m417z while you're here, the review raised whether the position grid would be better folded into volume-control-open-location than living as its own mod, and i don't have strong feelings either way. if you'd rather keep it in one place i'll close this and send it over to my-windhawk-mods as a PR instead. whichever you prefer.

scratch that last line about dpi, tested it since. 150% on the second screen and 100% on the primary, the anchors land right on both. offsets don't, they go in physical pixels so the same number travels further on the 100% one. fixing it.

mario0318:
took the early return, it bails out of ModInit when the position's on default and both offsets are zero rather than hooking for nothing. offsets now say they stop at the edge, which is what the code was doing anyway and what the readme said. credits are in there too.

left the local rename and the null check alone for now, and on the standalone question i'm still fine either way, happy to close this and send it to my-windhawk-mods if that's the preference.

mario0318:
1.1.2 has the leftover polish from the last round, atomics on the settings, plain statements instead of the comma operator, a log line when a stored position value isn't recognised, and the offset text now says what the clamp actually does.

1.1.3 is the one worth looking at. the rect this hook gets is in the target monitor's physical pixels, so the offsets were going in raw and the same setting covered different ground depending on scaling. checked it on a 100% and a 150% screen, an offset of -100 moved the indicator 99px on the primary and 95px on the 150% one, so about two thirds of the distance. it resolves the monitor off the incoming rect before the origin gets shifted and scales by its dpi, and after that the primary sits where it was and the 150% screen moves 149px. the anchors themselves were fine the whole time, they land flush on both.

mario0318:
1.1.4 has the bits from the last round in it. the dpi lookup only runs when an offset is actually set, the offsets are read once and passed in, and the comment over the settings says what the atomics really buy instead of overselling it.

the shellhost question i can't settle on my own. on 26200 explorer is the process with Windows.Internal.HardwareConfirmator.dll loaded, and shellhost is running the whole time without touching it, so explorer only is right here. i've got no 24h2 box to check against though, and taskbar-primary-on-secondary-monitor declares both and hooks the confirmator either way, which you'd know the reason for offhand. glad to add the include if it applies, i just didn't want to force the dll into a process that never shows the indicator on a guess.

left this round's three small ones alone, they're cosmetic. still happy either way on where the grid should live.

m417z:
> and taskbar-primary-on-secondary-monitor declares both and hooks the confirmator either way, which you'd know the reason for offhand

I believe that it's just a mistake and that it was always explorer-only. I'll change it in the next update.

PR #5109, opened 2026-08-15
SnapSentry 0.17.2

mario0318:
update to 0.17.2. mostly wording, plus a new screenshot.

the folder SnapSentry watches has always been settable but neither the mod description nor the readme ever said so, so people only ever saw the default Screenshots folder. both mention it now, the readme has a short section on it, and the setting itself shows example paths instead of just saying you can point it somewhere else. i also went through the rest of the settings descriptions and cut them down, they were accurate but nobody was going to read a paragraph per toggle.

the readme screenshot is new too. the old one predated the renaming option so the toast in it just showed Screenshot (1).png, the new one shows a file named after the window it was taken from, which is easier to understand than the paragraph explaining it.

then a 0.17.2 on top of that, because cutting the descriptions down took a few of them slightly off. the popup one never said what happens if you've turned SnapSentry's notifications off in windows, where it just stays quiet and still copies but won't prompt or delete. the recycle bin one read like it only covered the popup buttons rather than automatic deletions as well. and the line about what gets left on your machine had gone vague enough to be worse than the long version it replaced, so it names the two things outright now, a Start Menu shortcut and one registry entry. one of the descriptions had also picked up a colon, which breaks the settings yaml, so that's fixed too.

## Changelog

* Say that the watched folder can be your own, in the mod description, the readme, and the folder setting itself
* Shorten the settings descriptions
* New readme screenshot showing the notification with renaming turned on
* Fix a colon in a setting description that broke the settings YAML
* Say what happens when SnapSentry's notifications are turned off in Windows
* Make clear the Recycle Bin covers automatic deletions, not just the popup buttons
* Name the Start Menu shortcut and registry entry the popup leaves, instead of describing them vaguely

## Mod authorship

This mod was created by:

- - [x] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/snap-sentry.wh.cpp
+++ b/mods/snap-sentry.wh.cpp
@@ -2,7 +2,7 @@
 // @id              snap-sentry
 // @name            SnapSentry
 // @description     Watch your Screenshots folder or any folder you pick, then copy, rename, or delete each new screenshot, or choose from a notification.
-// @version         0.17.1
+// @version         0.17.2
 // @author          mario0318
 // @github          https://github.com/mario0318
 // @include         windhawk.exe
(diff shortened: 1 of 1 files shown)


mario0318:
pushed 0.17.2 with all of it in. the popup description names the registry entry again, my tightening pass dropped it and it shouldn't have, that footprint is most of what the earlier rounds went back and forth on. the notifications off case now says plainly that nothing shows up at all, rather than implying a dialog you never actually get.

rest is smaller. the desktop example is swapped out since it was pulling against the warning in the same paragraph, recycle says when it applies and that the popup buttons go through it too, and the readme says every new image instead of everything, plus about half a minute rather than a few seconds, which is the 30 second window the code really uses.

no code changed in either pass, it's still the merged 0.16.0 underneath.

PR #5111, opened 2026-08-15
Add taskbar-centered-start-split-icons mod

rycalvo:
## Summary
Adds a new mod: **Taskbar Start Button Centered Origin** (`taskbar-centered-start-split-icons`).

Pins the Windows 11 Start button to the true horizontal center of the primary monitor regardless of icon count, and splits running-app taskbar buttons into two groups flanking Start based on which side of the screen each window is currently on (with live drag-follow as windows move). Also supports positioning Search/Task View/Widgets, per-app left/right overrides, and configurable ordering within each side.

Full feature list, settings, known limitations, and development changelog are documented in the mod's own `WindhawkModReadme` block.

## Test plan
- [x] Built and run locally via Windhawk on Windows 11 over multiple days of iteration, including fixes for explorer.exe crashes on cross-monitor window moves (root-caused via WinDbg) and settling-window cosmetic polish
- [x] Verified stable across repeated cross-monitor moves, new app launches, and pin/unpin changes with no crashes

## Design note: unattended click-sentinel probing

The HWND-resolution step (mapping a taskbar button back to its window, needed to classify which side of the screen it's on) reuses the "sentinel click" technique already used elsewhere in this repo (`taskbar-numberer`, `taskbar-reorder-right-drag`) - dispatching a real click through the taskbar's internal click handler and intercepting it before it reaches the real handler. Those mods only fire it in response to an actual user gesture; this mod fires it unattended, from a background timer and from a fast-path nudge on `TaskListButton::UpdateVisualStates` (a taskbar button's own visual-state transitions, including a pinned app's running/not-running change), against every not-yet-resolved button.

Mitigations in place: an `ITaskGroup::IsRunning` bail-out before ever probing a group with no running windows (a pinned-but-not-running app therefore never gets probed at all), a per-path confirmed/broken latch that disables further probing after three unconfirmed misses, and capped exponential backoff between retries. Before a path's first confirmation, up to three probes on that path can miss - and a miss there is a real click reaching the taskbar's actual handler with the sentinel value in place of its real argument, so the realistic worst case is a stray window activate/minimize or an access violation in explorer.exe - before the latch trips and further probing on that path stops. (The threshold is three rather than one because a probe only ever runs after a real click has already been seen passing through the same interception point, so an early miss is far more likely a startup transient than evidence the hook is broken; a successful capture resets the count.) That bound is pre-confirmation only: once a path is confirmed working, a later miss no longer counts toward the latch (a post-confirmation miss is usually innocent - the window closed mid-probe, or a taskbar rebuild caught the item mid-teardown - so counting it risked latching a working path dead over an unrelated timing hiccup). For that case, retry frequency (not total count) is bounded instead by the same capped exponential backoff (2s up to 30min) the resolution cache uses everywhere else, so a persistently-broken-but-still-confirmed path can't dispatch clicks faster than that schedule allows.

I believe this is an acceptable risk given the mitigations, but I'm flagging it explicitly per review feedback since it's a stability/blast-radius judgment call rather than something provably safe from the code alone.

## Mod authorship

This mod was created by:

- - [x] Claude
- - [ ] ChatGPT
- - [ ] Gemini

Written by Claude Code (Anthropic's CLI) under my direction across multiple sessions - I described the desired behavior, reported bugs/crashes as they came up during real-world testing, and made the design decisions, while Claude wrote and iterated on the actual code. See the mod's own "Disclosures" section for the same note.

### DIFF
--- a/mods/taskbar-centered-start-split-icons.wh.cpp
+++ b/mods/taskbar-centered-start-split-icons.wh.cpp
@@ -1,0 +1,2404 @@
+// ==WindhawkMod==
+// @id              taskbar-centered-start-split-icons
+// @name            Taskbar Start Button Centered Origin
+// @description     Pins the Start button to the true horizontal center of the screen, and splits running-app taskbar buttons into two groups flanking it based on which side of the screen each window is currently on (Windows 11 only)
+// @version         0.1.0
+// @author          rick
+// @github          https://github.com/rycalvo
+// @include         explorer.exe
+// @architecture    x86-64
+// @compilerOptions -lcomctl32 -ldwmapi -lole32 -loleaut32 -lruntimeobject -lshcore
+// @license         GPL-3.0
+// ==/WindhawkMod==
+
+// Source code is published under The GNU General Public License v3.0.
+
+// ==WindhawkModReadme==
+/*
+# Centered Start button with position-split taskbar icons
+
+Pins the Windows/Start button to the exact horizontal center of the primary
+monitor, regardless of how many taskbar icons are present. Running-app
+taskbar buttons are split into two groups that flank the Start button:
+
+- A button goes to the **left** group if its window's current on-screen
+  position is left of screen-center.
+- A button goes to the **right** group if its window is right of
+  screen-center.
+- Pinned-but-not-running apps have no window to check the position of, so
+  they're classified by the `leftApps`/`rightApps` name lists below, falling
+  back to `unresolvedAppsDefaultSide` if unlisted. Within whichever side
+  they land on, `pinnedAppsAnchor` picks whether they sit at the far edge
+  or right next to Start.
+
+When you drag a window across the center line of the screen, its taskbar
+button switches sides to follow it. Side-switching is driven by a global
+window-location-change listener and is best-effort: it happens shortly
+after a drag/move settles, not on every intermediate pixel of the drag.
+
+Search, Task View and Widgets can either stay at the far left edge, or move
+right next to Start on whichever side you prefer.
+
+## Known limitations (please read before reporting issues)
+
+- **Windows 11 only.** Windows 10's taskbar has no XAML layer to hook into.
+- **Primary monitor only.** Screen-center math and window-side classification
+  use the primary monitor. Taskbars on secondary displays are not specially
+  handled by this mod (their icons keep the default layout Windows gives
+  them) - the positioning hook explicitly checks which taskbar's XAML tree
+  an element belongs to and leaves anything outside the primary's alone.
+- **Undocumented internals.** This mod hooks private, unversioned classes
+  inside `taskbar.dll` and `Taskbar.View.dll` (via symbols resolved from
+  Microsoft's public symbol server at runtime, not hardcoded offsets). A
+  Windows update can change these internals and break the mod until it's
+  updated. If that happens, disable the mod rather than filing against
+  explorer.exe crashing.
+- The "resolve which HWND a taskbar button represents" step reuses a
+  technique from other taskbar-reordering mods (synchronously reporting a
+  sentinel "click" to the taskbar's internal click handler, which is
+  intercepted before it does anything, to read back the window handle). It
+  runs on a periodic timer rather than inline during layout, and Arrange
+  only ever reads whatever the timer has already cached - running it
+  synchronously from inside the taskbar's own layout pass was the
+  confirmed cause of an explorer.exe crash (specifically when Windows'
+  "show taskbar apps on" setting is anything other than "All taskbars",
+  since that's when a window moving across monitors structurally adds/
+  removes taskbar buttons rather than just repositioning them).
+
+## Changelog
+
+**Initial build (Aug 2026)**
+- Start button pinned to the primary monitor's true center, independent of
+  the number of taskbar icons.
+- Running-app icons split left/right of Start by live window position, with
+  drag-follow (a button switches sides shortly after its window crosses the
+  center line).
+- `taskListOrder`: icons ordered by distance from center, or by native
+  taskbar order.
+- Minimized windows keep their last known side instead of jumping (a
+  minimized window's reported position is off-screen nonsense, so it's
+  frozen at wherever it was before minimizing).
+- `leftApps`/`rightApps` name-based overrides, plus
+  `unresolvedAppsDefaultSide` for pinned-but-not-running apps with no
+  window to classify by.
+- `systemButtonsPlacement`/`systemButtonsAdjacentSide`: Search, Task View
+  and Widgets can sit at the far-left edge or adjacent to Start.
+
+**Feature follow-ups**
+- Support for "Combine taskbar buttons: Always" (grouped icons bind to a
+  different internal view model than individual windows, so this needed
+  its own resolution path).
+- Negative caching for HWND resolution, so a pinned-but-not-running app
+  (which will never resolve to a window) isn't retried on every single
+  layout pass.
+- Self-healing taskbar-window lookup, so the mod doesn't stay permanently
+  inert if Windhawk happens to inject before the taskbar itself exists yet
+  (observed right after a fresh boot).
+- `unresolvedAppsDefaultSide` gained a `contralateral-to-system-buttons`
+  option.
+- `pinnedAppsAnchor` setting: pinned-not-running apps can sit at the outer
+  edge of their side, or right next to Start.
+
+**Stability: cross-monitor moves**
+Moving a window to a second monitor repeatedly crashed explorer.exe across
+several iterations, each surfacing a different unsafe call path within the
+same underlying class of bug: a synchronous WinRT/XAML call made while the
+taskbar's own internal button list is mid-structural-mutation - reproducible
+specifically when Windows' "show taskbar apps on" setting isn't "All
+taskbars," since only then does a cross-monitor move add or remove taskbar
+buttons rather than just repositioning them. Root-caused via WinDbg
+crash-dump analysis after a few rounds of counter-based guessing didn't
+fully close it. Fixes along the way:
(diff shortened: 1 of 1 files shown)


rycalvo:

rycalvo:
/ready-for-reviewer

One item from the latest AI review round was deliberately left unaddressed: the suggestion to subclass Shell_TrayWnd and switch InvalidateTaskbarLayout to a non-blocking PostMessage (plus caching the resolved XamlRoot/repeater), to reduce the per-event work done on Explorer's UI thread. This is a real optimization opportunity on an already-throttled path (window-move events are capped at ~7x/sec via the existing 150ms drag-follow throttle), not a correctness bug - it would be a substantial rewrite of the taskbar-notification mechanism, so I'm leaving it to a human reviewer's judgment on whether the added complexity is worth it rather than pushing through another automated round.

Every other finding across all review rounds - including the two prior severe compression-bound bugs, several teardown/threading races, and an unguarded exception boundary - has been addressed and confirmed via live testing (disable/re-enable cycling, steady-state operation, drag-follow, and the specific overflow/compression repros called out along the way).

rycalvo:
Re **item 4** (overlap with `taskbar-start-button-position`): thought about this and decided to keep it as a standalone mod rather than splitting the centering half out into that one.

The reason is that the coupling isn't incidental - it's structural. `PlanTaskListButtons`' `leftInnerX`/`rightInnerX` boundaries (where the two task-list groups start counting outward from) are derived directly from `startCenterX` and `g_lastStartWidth`, computed in the same `RecomputeLayoutPlan` pass that positions Start itself. The task-list split literally can't be planned without already knowing exactly where Start's edges are going to land this frame.

If centering moved into `taskbar-start-button-position` and this mod only did the split, I'd need one of:
- Duplicate the centering math here just to predict where the other mod is about to put Start (two independent implementations of the same formula, silently able to drift out of sync on a future Windows update that changes button margins/DPI handling - exactly the kind of thing this mod's own round 6-9 fixes were chasing), or
- Read Start's rendered position back out after the other mod has already placed it, which means my `IUIElement::Arrange` hook has to fire *after* theirs on the same vtable slot for the same frame - an ordering dependency between two independently-installed mods that Windhawk has no mechanism to guarantee.

Either option trades a documented "don't enable both" README warning for a much harder-to-diagnose failure mode (silent drift or load-order-dependent breakage) for a small one-time-setup cost to the user. Given that, standalone seems like the better shape here, with the incompatibility called out explicitly (already in both the README and the mod's `@description`).

Happy to revisit if a maintainer feels differently about the tradeoff.

rycalvo:
Re **item 3** (overlap with `taskbar-start-button-position`): still standing by the standalone decision from an earlier round ([comment](https://github.com/ramensoftware/windhawk-mods/pull/5111#issuecomment-5355649428)), but this round's framing raises a distinct point worth answering directly rather than just re-linking that comment.

**On "could the split-only layer live on top of that mod's Start-centering as an option":** no, for the same structural reason as before, restated in the terms this round's finding uses. `PlanTaskListButtons`' `leftInnerX`/`rightInnerX` - where each side's icon group starts counting outward from - are derived from `startCenterX`/`g_lastStartWidth`, computed in the *same* `RecomputeLayoutPlan` pass that places Start itself. Splitting the mods wouldn't just relocate code, it would change the architecture from "one atomic plan per frame, computed once" to "two independently-installed mods that both need to agree on where Start's edges are, computed twice." Concretely that means either duplicating the centering formula here anyway (drift risk on a future Windows update - this project's own round 6-9 crash-fixing history is a long list of exactly this class of "two things quietly stopped agreeing" bug), or depending on `taskbar-start-button-position`'s `Arrange` hook running before this mod's on the same vtable slot for the same frame, which Windhawk has no mechanism to guarantee between independently-installed mods.

**On the specific machinery this round calls out as duplicated** (`IdentifySystemButton`, the `XamlRoot`/`TaskbarHost::FrameHeight` resolution chain, the `vtable[92]` `Arrange` hook itself): this is real, but it's a different kind of overlap than a feature being reimplemented - it's the same *reverse-engineered facts about undocumented Windows internals* being independently re-derived, which every mod touching this part of the taskbar does on its own. `taskbar-start-button-position`, `taskbar-reorder-right-drag` and `taskbar-volume-control-per-app` each carry their own copy of this resolution chain too; Windhawk mods are standalone single files with no shared-library mechanism between them, so there's no way to factor out "the `XamlRoot` lookup" into something both mods call - only to credit the technique in comments, which this mod already does. Deleting my copy doesn't remove the duplication, it just moves where the maintenance burden of keeping it correct sits.

So: the semantic conflict (Start's X) is real and stays as a documented "don't run both," and the mechanical similarity (both mods needing the same taskbar-internals plumbing) is inherent to writing any mod that touches this XAML tree, not something splitting this mod would fix. Still open to a maintainer weighing in differently once a human looks at it - happy to revisit.

rycalvo:
Follow-up note, separate from this round's fixes: during this round's live testing I observed that with "Combine taskbar buttons and hide labels" set to "Always", multiple windows of the same app weren't combining into one taskbar button - they kept rendering as separate icons regardless of count.

I don't believe this mod can cause it: button grouping (whether Windows renders one combined `TaskListGroupViewModel`-bound button or several separate `TaskListWindowViewModel`-bound ones) is a native decision made entirely upstream, before any of this mod's hooks ever see a button - this mod only repositions and reads already-rendered buttons, it never creates or merges them. Still isolating whether this reproduces with the mod disabled before treating it as a real bug report. Noted in both READMEs' "Known limitations" as a planned follow-up in the meantime.

rycalvo:
Re **item 4** (overlap with `taskbar-start-button-position`): settling this explicitly, since it's come up several rounds now with increasingly specific framing each time.

**Decision: staying standalone.** Not because the duplication this round points at isn't real - it is - but because of what actually causes it and what merging would cost.

**On the specific machinery cited** (`IdentifySystemButton`-style classification, the pinned-X computation, the `vtable[92]` `Arrange` bootstrap): this is the same class of overlap flagged in round 15, restated with more specific names attached. It's real, but it's not "a feature reimplemented" - it's the same *reverse-engineered fact about an undocumented Windows internal* (which XAML classes represent Start/Search/Task View/Widgets, where the `IUIElement::Arrange` vtable slot lives) being independently rediscovered, because Windhawk mods have no shared-library mechanism to rediscover it once. `taskbar-reorder-right-drag` and `taskbar-volume-control-per-app` each carry their own copy of the equivalent XamlRoot-resolution chain for the same reason. Moving this mod's copy into `taskbar-start-button-position` wouldn't eliminate a duplicated *feature*, it would just relocate where the duplicated *plumbing* lives.

**The actual reason this mod can't cleanly split in two:** `PlanTaskListButtons` computes each side's inner boundary (`leftInnerX`/`rightInnerX`) directly from `startCenterX`/`g_lastStartWidth`, both produced by the *same* `RecomputeLayoutPlan` pass that centers Start. This isn't two features that happen to coexist in one file - the icon-split math is a strict function of Start's just-computed position. Splitting the mods would require either:
1. Duplicating the centering formula here anyway to know where "next to Start" is (no actual reduction in duplication, plus a drift risk every time either copy changes), or
2. Depending on `taskbar-start-button-position`'s `Arrange` hook running before this mod's on the same vtable slot, for the same frame - which Windhawk has no mechanism to guarantee between two independently-installed mods.

Given that, a "center" option added to `taskbar-start-button-position` wouldn't let this mod drop its own centering logic - it would just mean two independent mods computing Start's position and hoping they agree, which is strictly worse than either mod owning it outright.

**What I'd support instead**, if `taskbar-start-button-position`'s maintainer wants to raise it: documenting the mutual-exclusivity relationship from both sides (this mod's README already does, listing it as "don't enable alongside"), rather than attempting a code-level merge. Open to being told this reasoning doesn't hold up, but from this submission's side the coupling is structural, not incidental, and I don't think splitting improves anything for users of either mod.

rycalvo:

rycalvo:
Handing this over with the AI review clean (no blocking issues as of `de7d625`). Notes on the optional items left unaddressed, per the review process guidance:

- **`kArmResolveNowMsg`'s handler drops a `SetTimer` failure.** Agreed, and it's the same pattern `SetDragFollowPollInterval` now handles one function away. Left out deliberately so the head handed over is exactly the one that was reviewed clean, rather than introducing an unreviewed change at the last step. Happy to apply it before merge if you'd prefer — it's the same arm-before-kill ordering plus a log, about six lines.
- **Killing the drag-follow poll entirely when `g_resolvedHwnds` is empty.** Skipped in favour of keeping one mechanism rather than two. That state requires zero resolved windows, which is uncommon on a real desktop, and the idle tick is one wakeup per second doing a handful of `GetWindowRect` calls.
- **Per-button gating in `TaskListButton_UpdateVisualStates_Hook`.** Skipped because `(void**)pThis + 3` is a hardcoded ABI offset, and this mod otherwise only takes those where no symbol can answer the question (the `CTaskGroup::GetNumItems` probe being the one exception, which is self-validating). The path it would optimise is already cheap, and the flag's comment states its real limits rather than overclaiming them.
- **Diagnostics scaffolding.** Kept. Every counter has paid for itself: the adaptive cadence in this round was verified by differencing `polls` across log lines, and these counters root-caused three explorer.exe crashes earlier in development.
- **Comment volume.** Fair. It grew again because the last two rounds added mechanisms that needed explaining. Given the mod hooks undocumented, unversioned internals, I'd rather over-explain for whoever picks it up after a Windows update breaks something.
- **The RATIONALE.md link.** No comment defers to it any more — round 31 folded or dropped all ~48 "See RATIONALE.md" pointers so each comment stands alone. The link is supplementary history now, not required reading.

The two maintainer calls you flagged — the unattended click-sentinel probing, and the relationship to `taskbar-start-button-position` — are laid out in this PR's description under the design note.

PR #5118, opened 2026-08-16
Update Taskbar Icon Separators to v1.0.13

Meteony:
## Changelog

- Added dynamic taskbar management: add separators from the taskbar menu, drag to reorder, and right-click to remove.
  Screenshot

- Moved separator identities and positions to persistent mod storage, with automatic migration from previous settings.

- Moved to MaxWidth-only styling for better compatibility and maintainability. 
- Expanded the original worker backend for reliable persistence and genuine hot-reload.

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/taskbar-icon-separators.wh.cpp
+++ b/mods/taskbar-icon-separators.wh.cpp
@@ -2,7 +2,7 @@
 // @id              taskbar-icon-separators
 // @name            Taskbar Icon Separators
 // @description     Create tracked icon separators with configurable padding on the taskbar.
-// @version         0.5.42-p43
+// @version         1.0.0
 // @author          meteoni
 // @github          https://github.com/Meteony
 // @license         GPL-3.0
@@ -17,24 +17,22 @@
 
 Create tracked icon separators with configurable padding on the taskbar.
 
-![Screenshot](https://raw.githubusercontent.com/Meteony/meteoni-assets/main/taskbar-icon-separators/SEP.png)
+![Screenshot](https://raw.githubusercontent.com/Meteony/meteoni-assets/main/taskbar-icon-separators/SEP2.png)
 _Example for creating separators on the taskbar_
 
-Another mod offers similar functionality, but this implementation takes a different approach.
+Another mod offers similar functionality; this implementation takes on a more native-first spin.
 
-This mod uses private COM APIs to insert a genuine taskbar button, styles its width and centering, and keeps activation/tooltip behavior inert. Separators remain genuine pinned slots, can be reordered with the native taskbar drag behavior, and can also be created from the empty-taskbar context menu.
+This mod uses private COM APIs to insert a genuine taskbar button, styles its width and centering, and keeps activation/tooltip behavior inert. 
+Separators remain genuine, can be manually reordered, and can also be created from the taskbar context menu.
 
 
-Note: The mod creates pinned shortcuts targeting the Windows `systray.exe` stub (with a hidden, immediately exiting `cmd.exe` fallback) and reorders the persisted taskbar pin list. Cleanup is best-effort; in the worst failure case, separators may require manual unpinning.
+Note: The mod creates pinned shortcuts targeting Windows `systray.exe` (with fallback) and reorders the persisted taskbar pin list. 
+Cleanup is best-effort; in the worst failure case, manual taskbar unpinning and reordering may be required.
 
-Separator instances and their taskbar positions are stored by the mod itself
+Separator properties are stored by the mod itself
 inside its Windhawk mod-storage directory. Windhawk settings expose only the
 global separator width. Native Add, drag/reorder, and Unpin operations update
-that mod-owned state directly.
-
-Separator right-click is handled entirely inside explorer.exe with a small
-WinRT XAML MenuFlyout containing the mod-owned Unpin command. No companion
-ShellExperienceHost injection or cross-process coordination is required.
+ are persisted directly.
 
 Windows 11 only.
 */
--- a/mods/taskbar-icon-separators.wh.cpp
+++ b/mods/taskbar-icon-separators.wh.cpp
@@ -2,7 +2,7 @@
 // @id              taskbar-icon-separators
 // @name            Taskbar Icon Separators
 // @description     Create tracked icon separators with configurable padding on the taskbar.
-// @version         0.5.39
+// @version         0.5.42-p43
 // @author          meteoni
 // @github          https://github.com/Meteony
 // @license         GPL-3.0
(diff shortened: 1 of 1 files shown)


Meteoni:
1. Yes, there is an edge case where migration can fail. However this shouldn't be a huge risk for most users. If that user truly only had one default separator pinned, pinning it back should be trivial. 

2. I tried there doesn't seem to be an easy way to retrieve the index of a specific pinned button. This isn't as easily fixable as the review is claiming. I might look into that issue again later. For now, the user will just have to manually persist the pin.  

3. The image in README is now updated.

Meteoni:
Issue 1, 2 and 3 are now addressed. 

- The migration code is now updated to leave storage authoritative. 

- A visual hint is now added to guide the user to complete the manual anchoring process, which stays open until the separator leaves the undetermined -1 state, and persists across restarts. A direct icon enumeration path is harder than expected to implement, so I thought of this way. 

image

- Readme is now updated.

Meteoni:
1. That's the whole point. The anchoring process is mandatory for the mod to receive a genuine, persistent index for the separator. There's no good in allowing the user to dismiss that pop-up. 

The rest are non-critical and I can fix them in later updates.

Meteoni:
1. Non-critical. Those two are actually optional. Tooltip suppression is purely cosmetic. Same as click suppression. We already set up 2 lines of fallback - the systray.exe stub, the cmd.exe fallback. There's no reason to abort the entire startup for this. 
2. I might look into it in a later version. For now, those two generally seem to converge.

m417z:
> 2\. I might look into it in a later version. For now, those two generally seem to converge.

Here's a Claude report on that, unverified, posting in case it's correct and helpful.

<details><summary>Report</summary>
<p>

**`GetRelativeTaskOrder` returns a pinned-list index.** It returns `tbGroup->GetOrdinalPosition()`, and ordinals are assigned by walking the pinned list: `m_taskListUI->SetRelativeTaskOrder(pinnedItemGroup, pinnedIdx++)`. `_FinishDragGroup` renumbers them `0..N-1` over *only* groups with `ordinalPosition != -1`, i.e. pinned ones. And the value is written straight back into the pinned list: `m_pinnedList->Modify(idList, (WORD)(order + 1), PMC_TASKBANDREORDER)`. `TryMoveGroup`'s index is pinned space too.

**`HandleWinNumHotKey` takes a button-group index.** bounds-checks against `m_buttonGroups.GetPtrCount()` and does `m_buttonGroups.FastGetPtr(tbGroupIndex)` --- all groups, pinned and running alike.

So `PositionSeparators` → `MoveTaskbarPin` is **correct** and `ApplyPinnedMoveToSnapshot` / `RecordNativePinnedMove`'s shift arithmetic is **correct** (all separators are pinned, all in one space). The single wrong consumer is `CompensateWinNumHotKeyIndex`. That's good news --- the fix is local to one function and nothing persisted has to change.

The fix: stop deriving Win+N from stored state
----------------------------------------------

Converting pinned→group needs "how many non-pinned running groups sit left of this separator", which changes on every app launch/close --- so it can't be cached at drag time. But you don't need the conversion at all: walk the live button-group list at Win+N time and ask each group whether it's a separator. You already have `IsSeparatorTaskGroup` and `CTaskBtnGroup::GetGroup`. The only missing piece is the array, and there's merged precedent for it in [taskbar-wheel-cycle.wh.cpp:388-416](vscode-webview://1qomlt1ktdjirjto09cjqkh7skmmds6mvcliq4sdhj7pd7ngtfk4/mods/taskbar-wheel-cycle.wh.cpp#L388-L416) (m417z, Win10+Win11), which recovers the `HDPA` member offset by calling `CTaskListWnd::GetButtonGroupCount` --- verified to be a one-liner `return m_buttonGroups.GetPtrCount();` --- against a probe array:

```
using CTaskListWnd_GetButtonGroupCount_t = int(WINAPI*)(void* pThis);
static CTaskListWnd_GetButtonGroupCount_t g_taskListWndGetButtonGroupCount;

// CTaskListWnd::GetButtonGroupCount compiles to "return *(int*)this[offset]".
// Calling it with an array of pointers to ints that hold their own index makes
// it hand back the offset of the button-group DPA instead of a count.
static constexpr size_t kButtonGroupProbeSize = 256;

static size_t GetButtonGroupsArrayOffset() {
    static size_t offset = []() -> size_t {
        int values[kButtonGroupProbeSize];
        int* pointers[kButtonGroupProbeSize];
        for (size_t i = 0; i < kButtonGroupProbeSize; i++) {
            values[i] = static_cast<int>(i);
            pointers[i] = &values[i];
        }

        int probe = g_taskListWndGetButtonGroupCount(pointers);
        if (probe <= 0 || static_cast<size_t>(probe) >= kButtonGroupProbeSize) {
            Wh_Log(L"[WINNUM] Unexpected button-group offset probe %d", probe);
            return 0;
        }

        return static_cast<size_t>(probe);
    }();

    return offset;
}

static bool GetTaskBtnGroups(void* taskListWnd, int* countOut, void*** itemsOut) {
    size_t offset = GetButtonGroupsArrayOffset();
    if (!offset) {
        return false;
    }

    auto dpa = static_cast<void**>(static_cast<void**>(taskListWnd)[offset]);
    if (!dpa) {
        return false;
    }

    // HDPA: { int cpItems; void** pArray; ... }
    *countOut = *reinterpret_cast<int*>(dpa);
    *itemsOut = static_cast<void**>(dpa[1]);
    return *countOut >= 0 && *itemsOut != nullptr;
}
```

Then the compensation becomes a straight count, with no index-space math at all:

```
// Win+number reaches CTaskListWnd::HandleWinNumHotKey as a zero-based index into
// the live button-group list, which includes our pinned separator groups. Map the
// requested app slot to the physical group index by skipping separator groups.
static short CompensateWinNumHotKeyIndex(void* taskListWnd, short logicalIndex) {
    if (logicalIndex < 0 || logicalIndex > 9 || g_unloading ||
        !g_taskBtnGroupGetGroupAddress || !g_taskListWndGetButtonGroupCount) {
        return logicalIndex;
    }

    int count = 0;
    void** groups = nullptr;
    if (!GetTaskBtnGroups(taskListWnd, &count, &groups)) {
        return logicalIndex;
    }

    using CTaskBtnGroup_GetGroup_t = void*(WINAPI*)(void* pThis);
    auto getGroup = reinterpret_cast<CTaskBtnGroup_GetGroup_t>(
        g_taskBtnGroupGetGroupAddress);

    int remaining = logicalIndex;
    for (int i = 0; i < count && i <= SHRT_MAX; i++) {
        if (IsSeparatorTaskGroup(getGroup(groups[i]))) {
            continue;
        }

        if (remaining == 0) {
            return static_cast<short>(i);
        }

        remaining--;
    }

    // Fewer apps than the requested slot: let the stock bounds check handle it.
    return logicalIndex;
}
```

`TaskListWnd_HandleWinNumHotKey_Hook` just passes `pThis` through. `pThis` there and the `this` that `GetButtonGroupCount`'s body expects are the same pointer --- both are real `CTaskListWnd` member bodies, so MSVC gives them the same canonical `this` and the adjustor thunks live in the secondary vtables. Add the symbol next to the existing ones:

```
{
    {LR"(public: virtual int __cdecl CTaskListWnd::GetButtonGroupCount(void))"},
    &g_taskListWndGetButtonGroupCount,
    nullptr,
},
```

What this buys beyond the reported bug
--------------------------------------

-   **Kills the whole finding.** `targetIndex` becomes unambiguously a pinned-list index with exactly one consumer (`MoveTaskbarPin`). Worth tightening the L165 comment to say so, since `GetRelativeTaskOrder`'s space is the non-obvious part.
-   **Fixes the "unanchored separators are excluded from Win+N compensation" note** (functionality note 4). A `targetIndex == -1` separator is still a real button group, so it now gets counted like any other --- the "next pinned app is off by one until the separator is dragged" case disappears.
-   **Fixes multi-monitor.** The current code reads global desired state, but `pThis` is the `CTaskListWnd` for the taskbar that actually received the hotkey. With "show taskbar buttons where the window is open", running groups differ per monitor, so the global count was wrong there too. (`CTaskListWndMulti::HandleWinNumHotKey` forwards to the primary `CTaskListWnd`, so the unhooked `Multi` variant noted in the review stays a non-issue.)
-   **The `g_backendHasConverged` gate at L3906 can go.** It existed because stored positions could lag the visible taskbar; reading the live list removes the premise. `IsSeparatorTaskGroup` only does identity matching, which is valid mid-converge.
-   **No locking concern.** `HandleWinNumHotKey` runs on the taskbar window thread that owns `m_buttonGroups`, so the walk is on the mutating thread.

</p>
</details>

PR #5124, opened 2026-08-16
Wobbly Windows

Potassiumuncher:
Adds interactive wobbly physics when moving or resizing windows.
Desktop2026 08 16-18 17 43 03-ezgif com-video-to-gif-converter

## Changelog

* Changelog item 1...
* Changelog item 2...

## Mod authorship

This mod was created by:

- - [  ] The submitter, without AI assistance
- - [ ✓ ] The submitter, with AI assistance
- - [ ✓ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/wobbly-windows.wh.cpp
+++ b/mods/wobbly-windows.wh.cpp
@@ -4,7 +4,7 @@
 // @description   wobbly wibbly windows
 // @version       1.0
 // @author        potassiumuncher
-// @github        https://github.com/potassiumuncher
+// @github        https://github.com/Potassiumuncher
 // @include       *
 // @exclude       explorer.exe
 // @compilerOptions -ldwmapi -lgdi32 -ld2d1 -luser32 -lcomctl32 -lwinmm
--- a/mods/wobbly-windows.wh.cpp
+++ b/mods/wobbly-windows.wh.cpp
@@ -733,7 +733,8 @@ static void RestoreWindowAfterWobbly(HWND mainHwnd, LONG_PTR oldExStyle) {
     if (mainHwnd && IsWindow(mainHwnd)) {
         SetLayeredWindowAttributes(mainHwnd, 0, 255, LWA_ALPHA);
         if (!(oldExStyle & WS_EX_LAYERED)) {
-            SetWindowLongPtrW(mainHwnd, GWL_EXSTYLE, oldExStyle);
+            LONG_PTR ex = GetWindowLongPtrW(mainHwnd, GWL_EXSTYLE);
+            SetWindowLongPtrW(mainHwnd, GWL_EXSTYLE, ex & ~WS_EX_LAYERED);
         }
     }
 }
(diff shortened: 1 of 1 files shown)


Potassiumuncher:
Applied 2, 5, 6, and 7 as written, plus the g_wmDetach handling from 1 (OverlayProc now destroys itself when the unload sweep reaches it, instead of that message falling through to DefWindowProc). Left the rest of 1, all of 3, and all of 4 out of this pass:

1 (remainder): the full fix means creating the overlay per-drag instead of keeping a persistent one that touches InitializeWobbly, OnEnterSizeMove, FinishWobblyTrackingLocalPart, OverlayProc, and Wh_ModInit/Wh_ModUninit together, since it changes when the window exists at all. What I applied (the detach handler) reduces how often the residual case is hit; the persistent-overlay model itself is unchanged.
3: verified the deadlock shape is real. SetWindowPos/UpdateLayeredWindow under g_wobblyMutex while another thread can be blocked in OnWindowPosChanged on the same mutex for an unrelated WM_WINDOWPOSCHANGED. The suggested fix (making the engine thread_local, dropping the mutex) rewrites most of the file's globals and changes "one window wobbles per process" into per-thread semantics. That's a design change, not a patch.
4: the subclassing footprint is real, but swapping to a WH_CALLWNDPROC hook on WM_ENTERSIZEMOVE is a different attach mechanism entirely, removes the CreateWindowExA/W hooks and adds per-thread hook install/uninstall. Also a redesign.

All three are real, verified findings, I just didn't want to fold a redesign into what's meant to be a bugfix pass

Potassiumuncher:
i applied changes to things on what would improve my mod but i had to not apply some changes as it would effect my mod negatively. 

/ai-review

Potassiumuncher:
Skipped item 2 (the WH_CALLWNDPROC-per-thread, subclass-on-WM_ENTERSIZEMOVE redesign) and the STATIC-class part of item 3. Both replace how and when windows get subclassed in the first place, not how detach finds them , that's a different piece of the design than the bugs I was fixing, and folding a redesign into a bugfix pass risks introducing new problems I can't fully reason through in the same pass. Real, worth doing separately if you'd like to scope that as its own change.

Everything else in this review was applied: the class-registration ordering (item 6), restoring the window if the overlay dies mid-drag (item 5), blocking until the overlay and every subclassed window are actually gone at unload instead of giving up on a timeout (item 3, and the detach half of item 1), the same-thread check so the overlay doesn't get orphaned on the wrong UI thread (item 4), and swapping the enumeration-based detach for a tracked set so message-only windows which EnumThreadWindows can never see actually get detached (item 1's main fix).

/ai-review

Potassiumuncher:
Thanks for the review, but it needs to have include all as it will be animating all windows, and i have confirmed that corner-radius dose work.  

/ready-for-reviewer

m417z:
Any reason not to address points 2, 5, 6, 7?
It also flickers in my tests when a move/resize operation is started. You might want to address that too.

I also suggest looking at this note, addressing it has potential to improve the overall responsiveness:

> - **`SetTimer(hwnd, 1, 16, NULL)` + `timeBeginPeriod(1)` is a coarse animation clock.** `WM_TIMER` is a low-priority message that is only synthesized when the queue is otherwise empty, so on a busy UI thread the physics will stutter regardless of the timer period; and `timeBeginPeriod(1)` raises the process timer resolution (power cost) for the duration. A dedicated render thread paced by `DwmFlush()` would be smoother and would decouple the animation from the app's message loop.

Potassiumuncher:
Thanks for the review but suggestions 1, 4, 5 (bug i cannot fix as i can not find a fix yet - WIP) and 6 are declined the rest are applied.

/ai-review

PR #5139, opened 2026-08-17
Add `hide-dotfiles-minimal` mod

USLTD:
This mod hides files and folders starting with `'.'` in **Windows Explorer**, similar to Unix/Linux systems. It adds the `FILE_ATTRIBUTE_HIDDEN` to the directory information returned to **Windows Explorer** without modifying the actual file attributes on disk.

Why not https://github.com/ramensoftware/windhawk-mods/blob/main/mods/hide-dotfiles-explorer.wh.cpp ? Why another mod? Well... problem is that while `hide-dotfiles-explorer` is very good mod (and what I used before I developed this fork?), it was complex and had few bugs (likely just side-effects of implementation but bugs from my POV), including not being able to drag-n-drop anything from **WinRAR**, **7zip** and other windows into folder or sub-folder in dotfile directory. Another was that I couldn't navigate to `<PATH_WITHIN_DOTFILE_DIRECTORY>` and I couldn't also open it from `explorer.exe <PATH_WITHIN_DOTFILE_DIRECTORY>`, forcing me to either disable mod temporary (defeating whole purpose), or use dirty workarounds, such as creating `DO_NOT_REMOVE` folder within dotfile directory and then create shortcut of `.dotfile\DO_NOT_REMOVE` and place it on desktop or somewhere. As you can imagine, this got inconvenient too so I just gave a try creating minimal mod that only did one thing - hiding dotfiles without no extra stuff, and I got an idea - why not simply make it seem like file/folders are with `Hidden` attribute without actually setting those attributes, only making it seem like for Windows Explorer? That way I could simply toggle `Show hidden files` to see those dotfiles whenever I need it, instead of toggling off and on the mod. So, here's the result.

Note that this is not replacement but just a simple and minimal alternative for those who want minimal and working experience that matches what Unix users would expect.

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/hide-dotfiles-minimal.wh.cpp
+++ b/mods/hide-dotfiles-minimal.wh.cpp
@@ -1,0 +1,306 @@
+// ==WindhawkMod==
+// @id              hide-dotfiles-minimal
+// @name            Hide Dotfiles Minimal
+// @description     Hide files and folders starting with . in Windows Explorer and Desktop
+// @version         1.0.0
+// @author          USLTD
+// @github          https://github.com/USLTD
+// @include         explorer.exe
+// @compilerOptions -lcomctl32 -lole32 -loleaut32 -lshell32 -lshlwapi
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Hide Dotfiles Minimal
+
+Makes files and folders whose names start with `.` hidden in **Windows
+Explorer**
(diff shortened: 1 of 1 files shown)


USLTD:
Hmm... in that case I will close this PR since it doesn't add anything meaningful

PR #5154, opened 2026-08-19
MacOS Minimize Animation: make classic style target a top taskbar

Abdullah-Masood-05:
The classic genie style always funneled to the bottom of the monitor (\dockY = mon.bottom\), so with a taskbar pinned to the top the animation still appeared from the bottom. The modern (Direct2D) engine already handles a top taskbar by reading the taskbar window rect.

This change makes the classic engine do the same:
- Detect the taskbar position via \FindTaskbarForMonitor\ + \GetWindowRect\ and set the dock Y to the top edge when the taskbar hugs the top of the screen.
- Span the animation canvas from the dock edge up/down to the window so a dock above the window isn't clipped.
- Flip the morph spread so the rows nearest the dock lead (top-first for a top taskbar).

Bottom-taskbar behavior is unchanged.

### DIFF
--- a/mods/macos-minimize-animation.wh.cpp
+++ b/mods/macos-minimize-animation.wh.cpp
@@ -1348,29 +1348,31 @@ DWORD WINAPI MacGenieAnimThreadClassic(LPVOID lpParam) {
     const float origCenterX = (float)origLeft + W * 0.5f;
 
     // Where the genie funnels to: the learned taskbar icon X. The dock Y is taken
-    // from the real taskbar edge via the shell, so a taskbar pinned to the top of
-    // the screen is honored too (previously the classic style always assumed the
-    // taskbar was at the bottom of the monitor). Windows keeps secondary taskbars
-    // on the same edge as the primary, so the single ABM_GETTASKBARPOS query is
-    // valid for every monitor.
+    // from the real per-monitor taskbar window (the same signal the modern engine
+    // uses), so a taskbar pinned to the top of the screen is honored too -
+    // previously the classic style always assumed the taskbar was at the bottom of
+    // the monitor. A single global shell query would get the wrong edge on setups
+    // where each monitor's taskbar can be on a different edge.
     int dockX = data->targetDockX;
     if (dockX < mon.left) dockX = mon.left;
     if (dockX > mon.right) dockX = mon.right;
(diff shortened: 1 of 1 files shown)


Abdullah-Masood-05:
Thanks for the review — both points were valid, and I've addressed them in 931d8dbb:
1. Taskbar edge detection: switched from the GetWindowRect gap heuristic to SHAppBarMessage(ABM_GETTASKBARPOS), which reports the taskbar edge correctly for every monitor (including the fallback-tray case and vertical taskbars).
2. Morph flip: now keyed on the actual funnel direction (dockY < origTop) instead of the taskbar edge, so the yb[] scanline walk can't invert and blank the canvas when the dock is on a different monitor.
I left point #3 (worker drain in Wh_ModBeforeUninit) untouched since it's pre-existing and out of scope for this PR — happy to do it as a separate change if you'd like.

Abdullah-Masood-05:
I've addressed both findings in 6a5488c4:
Dock direction / monotonic yb[]: direction is now taken from the window's mid-line instead of a strict < origTop, and dockY is clamped outside the window's vertical span, so yb[] stays monotonically increasing by construction. This fixes the full-screen + top-taskbar case (no more thin smear).
Taskbar edge detection: dropped the global ABM_GETTASKBARPOS query (which ignores hWnd and always reports the primary taskbar) in favor of the same per-monitor signal the modern engine uses — FindTaskbarForMonitor + GetWindowRect, with an upper-half check. This correctly handles setups like top on the primary and bottom on the secondary (e.g. via the taskbar-on-top mod).
I left point #3 (the Wh_ModBeforeUninit worker-drain / untimed join) untouched as it's pre-existing and out of scope for this PR. I also skipped the optional items. Happy to do the worker-drain fix as a separate change if wanted.

Abdullah-Masood-05:
The core of this PR (the top-taskbar fix) is complete and confirmed correct, including the yb[] monotonicity check you verified.

On the two points:

Taskbar probe reading the wrong monitor: I'm leaving this as-is. It only affects the experimental Multi-monitor support setting in a specific vertically-stacked layout. The default path is unaffected. I can address it separately if it becomes an issue.

Worker threads never joined on unload: this is pre-existing and out of scope for this PR, so I won't touch it here.

Issue #5158, opened 2026-08-19
[Disable Windows Shortcuts] Add "Win+Shift+C" to the list of toggleable shortcuts


megusvae:
Mod author: @Louis047 

"Win+Shift+C" gets reserved by Windows 10 to ["open the Charms Menu"](https://support.microsoft.com/en-gb/windows/keyboard-shortcuts-in-windows-dcc61a57-8ff0-cffe-9796-cb9706c75eec#windowsversion=windows_10). Windows 10 doesn't have a charm menu, so the hotkey just gets taken for no reason.

Louis047:
Noted and will be adding it by tomorrow when I get time

PR #5166, opened 2026-08-20
[Disable Windows Shortcuts] New shortcut additions with a few bug fixes and settings refactor

Louis047:
### Additions

- Added new shortcuts
  - `Win+Shift+C` - Charms Menu in Windows 10 (Removed in later releases but windows still controls the shortcut)
  - `Win+Shift+R` - Snipping Tool Recorder (Credits to @SteffanDonal)

### Fixes
- Fixed key states being stuck when related keys held after locking the system (Credits to @SteffanDonal)

### Updates
- Refactored the Settings Structure into three sections
  - Special Shortcuts
  - Direct Shortcuts
  - Standard Shortcuts
- Removed the 3-tier dropdown settings for Special Shortcuts and change it into a single boolean toggle
- Refined the README to match the latest architectural standards
  
## Issue References

- #5158

### DIFF
--- a/mods/disable-windows-shortcuts.wh.cpp
+++ b/mods/disable-windows-shortcuts.wh.cpp
@@ -241,6 +241,9 @@ If you use the **"Block hotkey"** option, or if you disable window snapping (Win
   - DisableWinSpace: false
     $name: Win+Space
     $description: Switch keyboard layout
+  - DisableWinShiftR: false
+    $name: Win+Shift+R
+    $description: Snipping Tool record
   - DisableWinShiftS: false
     $name: Win+Shift+S
     $description: Snipping Tool screenshot
@@ -379,6 +382,7 @@ struct
     bool DisableWinCtrlO;
     bool DisableWinCtrlS;
     bool DisableWinSpace;
+    bool DisableWinShiftR;
     bool DisableWinShiftS;
     bool DisableWinAltK;
     bool DisableWinPeriod;
@@ -473,6 +477,7 @@ void LoadSettings()
     g_settings.DisableWinCtrlO = Wh_GetIntSetting(L"StandardShortcuts.DisableWinCtrlO");
     g_settings.DisableWinCtrlS = Wh_GetIntSetting(L"StandardShortcuts.DisableWinCtrlS");
     g_settings.DisableWinSpace = Wh_GetIntSetting(L"StandardShortcuts.DisableWinSpace");
+    g_settings.DisableWinShiftR = Wh_GetIntSetting(L"StandardShortcuts.DisableWinShiftR");
     g_settings.DisableWinShiftS = Wh_GetIntSetting(L"StandardShortcuts.DisableWinShiftS");
     g_settings.DisableWinAltK = Wh_GetIntSetting(L"StandardShortcuts.DisableWinAltK");
     g_settings.DisableWinPeriod = Wh_GetIntSetting(L"StandardShortcuts.DisableWinPeriod");
@@ -547,6 +552,7 @@ bool ShouldBlockHotkey(UINT fsModifiers, UINT vk)
             {
                 case 'C': block = g_settings.DisableWinShiftC; break;
                 case 'M': block = g_settings.DisableWinShiftM; break;
+                case 'R': block = g_settings.DisableWinShiftR; break;
                 case 'S': block = g_settings.DisableWinShiftS; break;
                 case VK_LEFT: block = g_settings.DisableWinShiftLeft; break;
                 case VK_RIGHT: block = g_settings.DisableWinShiftRight; break;
--- a/mods/disable-windows-shortcuts.wh.cpp
+++ b/mods/disable-windows-shortcuts.wh.cpp
@@ -33,7 +33,7 @@ If you use the **"Block hotkey"** option, or if you disable window snapping (Win
 3. Under **Process inclusion list**, ensure `dwm.exe` is added (or `*` is used to include all processes)
 4. Click **Save**. Windhawk will automatically restart to apply the new settings.
 
-*Note: Changes to standard shortcuts (like Win+E) require an Explorer restart to take effect. You will be prompted automatically. If you completely disable or remove this mod from Windhawk, you must manually restart Explorer to restore those standard shortcuts.*
+*Note: Changes to standard shortcuts (like Win+E) require an Explorer restart to completely release the hotkeys for other applications (such as PowerToys or GlazeWM). You will be prompted automatically. If you completely disable or remove this mod from Windhawk, you must manually restart Explorer to restore those standard shortcuts.*
 
 ## Notes
 - Win key (Start Menu) is handled by the "Block Start Menu and Hosts" mod
@@ -322,7 +322,7 @@ bool g_isExplorer = false;
 bool g_isDWM = false;
 
 // Settings structure
-struct
+struct Settings
 {
     int DisableWinA;
     bool DisableWinB;
@@ -402,6 +402,98 @@ struct
     bool DisableWinKey;
     bool DisableCtrlEsc;
 } g_settings;
+
+bool StandardShortcutsEqual(const Settings& a, const Settings& b)
+{
+    return a.DisableWinB == b.DisableWinB &&
+           a.DisableWinD == b.DisableWinD &&
+           a.DisableWinE == b.DisableWinE &&
+           a.DisableWinF == b.DisableWinF &&
+           a.DisableWinF1 == b.DisableWinF1 &&
+           a.DisableWinG == b.DisableWinG &&
+           a.DisableWinH == b.DisableWinH &&
+           a.DisableWinI == b.DisableWinI &&
+           a.DisableWinJ == b.DisableWinJ &&
+           a.DisableWinM == b.DisableWinM &&
+           a.DisableWinO == b.DisableWinO &&
+           a.DisableWinQ == b.DisableWinQ &&
+           a.DisableWinR == b.DisableWinR &&
+           a.DisableWinS == b.DisableWinS &&
+           a.DisableWinT == b.DisableWinT &&
+           a.DisableWinV == b.DisableWinV &&
+           a.DisableWinW == b.DisableWinW &&
+           a.DisableWinX == b.DisableWinX &&
+           a.DisableWinY == b.DisableWinY &&
+           a.DisableWinZ == b.DisableWinZ &&
+           a.DisableWinTab == b.DisableWinTab &&
+           a.DisableWinUp == b.DisableWinUp &&
+           a.DisableWinDown == b.DisableWinDown &&
+           a.DisableWinLeft == b.DisableWinLeft &&
+           a.DisableWinRight == b.DisableWinRight &&
+           a.DisableWinHome == b.DisableWinHome &&
+           a.DisableWinShiftC == b.DisableWinShiftC &&
+           a.DisableWinShiftM == b.DisableWinShiftM &&
+           a.DisableWinComma == b.DisableWinComma &&
+           a.DisableWinPause == b.DisableWinPause &&
+           a.DisableWinCtrlD == b.DisableWinCtrlD &&
+           a.DisableWinCtrlF == b.DisableWinCtrlF &&
+           a.DisableWinCtrlF4 == b.DisableWinCtrlF4 &&
+           a.DisableWinCtrlLeft == b.DisableWinCtrlLeft &&
+           a.DisableWinCtrlRight == b.DisableWinCtrlRight &&
+           a.DisableWinNumbers == b.DisableWinNumbers &&
+           a.DisableWinShiftNumbers == b.DisableWinShiftNumbers &&
+           a.DisableWinCtrlNumbers == b.DisableWinCtrlNumbers &&
+           a.DisableWinAltNumbers == b.DisableWinAltNumbers &&
+           a.DisableWinPlus == b.DisableWinPlus &&
+           a.DisableWinMinus == b.DisableWinMinus &&
+           a.DisableWinEsc == b.DisableWinEsc &&
+           a.DisableWinCtrlEnter == b.DisableWinCtrlEnter &&
+           a.DisableWinCtrlC == b.DisableWinCtrlC &&
+           a.DisableWinCtrlN == b.DisableWinCtrlN &&
+           a.DisableWinCtrlO == b.DisableWinCtrlO &&
+           a.DisableWinCtrlS == b.DisableWinCtrlS &&
+           a.DisableWinSpace == b.DisableWinSpace &&
+           a.DisableWinShiftS == b.DisableWinShiftS &&
+           a.DisableWinAltK == b.DisableWinAltK &&
+           a.DisableWinPeriod == b.DisableWinPeriod &&
+           a.DisableWinSemicolon == b.DisableWinSemicolon &&
+           a.DisableWinPrtSc == b.DisableWinPrtSc &&
+           a.DisableWinShiftLeft == b.DisableWinShiftLeft &&
+           a.DisableWinShiftRight == b.DisableWinShiftRight &&
+           a.DisableWinShiftUp == b.DisableWinShiftUp &&
+           a.DisableWinShiftDown == b.DisableWinShiftDown &&
+           a.DisableOfficeHotkeys == b.DisableOfficeHotkeys &&
+           a.DisableWinAltD == b.DisableWinAltD &&
+           a.DisableWinAltB == b.DisableWinAltB &&
+           a.DisableWinAltR == b.DisableWinAltR &&
+           a.DisableWinAltG == b.DisableWinAltG &&
+           a.DisableWinAltPrtSc == b.DisableWinAltPrtSc &&
+           a.DisableWinAltT == b.DisableWinAltT &&
+           a.DisableWinAltM == b.DisableWinAltM &&
+           a.DisableWinCtrlShiftB == b.DisableWinCtrlShiftB &&
+           a.DisableWinCtrlQ == b.DisableWinCtrlQ &&
+           a.DisableAltShift == b.DisableAltShift &&
+           a.DisableWinKey == b.DisableWinKey &&
+           a.DisableCtrlEsc == b.DisableCtrlEsc;
+}
+
+bool HasAnyStandardShortcutsDisabled()
+{
+    Settings emptySettings{};
+    return !StandardShortcutsEqual(g_settings, emptySettings);
+}
+
+bool SettingsEqual(const Settings& a, const Settings& b)
+{
+    return a.DisableWinA == b.DisableWinA &&
+           a.DisableWinC == b.DisableWinC &&
+           a.DisableWinK == b.DisableWinK &&
+           a.DisableWinN == b.DisableWinN &&
+           a.DisableWinP == b.DisableWinP &&
+           a.DisableWinU == b.DisableWinU &&
+           a.DisableWinSlash == b.DisableWinSlash &&
+           StandardShortcutsEqual(a, b);
+}
 
 
 int GetSettingIntSafe(PCWSTR settingName) {
@@ -666,13 +758,13 @@ bool IsKnownHardcodedHotkey(UINT fsModifiers, UINT vk)
             // Hardcoded keys that bypass RegisterHotKey
             if (vk == VK_TAB || vk == VK_UP || vk == VK_DOWN || vk == VK_LEFT || vk == VK_RIGHT || vk == VK_SPACE)
                 return true;
-            if (vk == 'A' && g_settings.DisableWinA == 2) return true;
-            if (vk == 'C' && g_settings.DisableWinC == 2) return true;
-            if (vk == 'K' && g_settings.DisableWinK == 2) return true;
-            if (vk == 'N' && g_settings.DisableWinN == 2) return true;
-            if (vk == 'P' && g_settings.DisableWinP == 2) return true;
-            if (vk == 'U' && g_settings.DisableWinU == 2) return true;
-            if (vk == VK_OEM_2 && g_settings.DisableWinSlash == 2) return true;
+            if (vk == 'A' && g_settings.DisableWinA > 0) return true;
+            if (vk == 'C' && g_settings.DisableWinC > 0) return true;
+            if (vk == 'K' && g_settings.DisableWinK > 0) return true;
+            if (vk == 'N' && g_settings.DisableWinN > 0) return true;
+            if (vk == 'P' && g_settings.DisableWinP > 0) return true;
+            if (vk == 'U' && g_settings.DisableWinU > 0) return true;
+            if (vk == VK_OEM_2 && g_settings.DisableWinSlash > 0) return true;
         } else {
             // Win+Shift+Arrows
             if (vk == VK_UP || vk == VK_DOWN || vk == VK_LEFT || vk == VK_RIGHT)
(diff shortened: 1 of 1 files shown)


Louis047:
I have added a new shortcut with a few critical bug fixes related to explorer restart prompts.

/ai-review

Louis047:
### Summary of Architectural Refinements & Fixes

1. **Unified Special Shortcuts to Clean Booleans**:
   - Replaced the obsolete 3-tier enum (`off`, `disable`, `block`) with standard boolean toggles backed by the DWM low-level hook (`WH_KEYBOARD_LL`).
   - Windows processes system flyouts (`Win+A`, `Win+N`, `Win+C`, `Win+K`, `Win+P`, `Win+U`, `Win+/`) at the OS compositor layer, where Explorer's `RegisterHotKey` cannot suppress them. The DWM hook suppresses physical keystrokes (`return 1` + AutoHotkey `0xFF` mask) while passing through injected keystrokes (`LLKHF_INJECTED`), ensuring native taskbar tray clicks and custom taskbars (like YASB) remain fully functional.

2. **Organized 3-Tier Settings Structure**:
   - **Special Shortcuts**: System flyouts handled via DWM (`Win+A`, `Win+N`, `Win+C`, `Win+K`, `Win+P`, `Win+U`, `Win+/`).
   - **Direct Shortcuts**: Handled immediately via low-level hook without requiring an Explorer restart (`Win`, `Ctrl+Esc`, `Alt+Shift`, `Win+Space`, `Win+Ctrl+Shift+Alt`, `Win+Tab`, `Win+Arrows`, `Win+Ctrl+Shift+B`).
   - **Standard Shortcuts**: Explorer `RegisterHotKey`-registered shortcuts (`Win+E`, `Win+R`, `Win+D`, `Win+I`, `Win+S`, `Win+V`, `Win+Shift+S`, etc.) that release their kernel registration upon restarting Explorer.

3. **Accurate Explorer Restart Prompt Gating**:
   - `StandardShortcutsEqual()` and `HasAnyStandardShortcutsDisabled()` compare strictly Explorer-dependent shortcuts. Modifying or toggling Direct/Special shortcuts will never trigger an Explorer restart prompt.
   - `Wh_ModInit()` combines multi-hotkey probing with uptime gating (>15s) to accurately distinguish fresh boots from mid-session re-enables.

4. **Safe & Clean Unload Lifecycle**:
   - `Wh_ModUninit()` safely dismisses pending prompt dialogs and uses a 30s bounded wait matching official Windhawk mod conventions to prevent unload stalls.

5. **UI & Documentation Polish**:
   - Updated Office hotkey to `$name: Win+Ctrl+Shift+Alt` with `$description: Office Hub / Microsoft 365 app combinations`.
   - Removed repetitive tags from individual item names, moving notices to section descriptions.
   - Cleaned up stale README notes, retaining only `Win+L (Lock PC) cannot be blocked through standard hooks` under `## Notes`.

/ai-review

Louis047:
Thanks for the review. We've addressed the findings in the latest commit:

- **Settings migration:** Dropped the `||` fallback reads in `LoadSettings()` so `DirectShortcuts` are loaded cleanly without getting stuck on by stale INI keys.
- **Office hotkeys & input handling:** Added modifier key-up suppression in Step 4 so `g_suppressedKeys` is cleared properly and unbalanced key-ups aren't leaked to the OS.
- **Cleanup:** Switched process name matching to `wcsrchr` / `_wcsicmp`, closed dangling thread handles on hook failure, and removed unused headers.

Regarding the Explorer restart prompts:
- We kept the uninit prompt when standard shortcuts are disabled because Explorer only registers hotkeys once at startup and never reclaims them dynamically on unload (matching the pattern in `simple-window-switcher.wh.cpp`). However, we fixed the race condition by using a polling `WM_CLOSE` loop so closing open dialogs can never hang or deadlock.
- We also replaced the fragile uptime heuristic and probe check with a deterministic taskbar check (`Shell_TrayWnd`). Fresh boots and Explorer restarts never prompt since the window doesn't exist yet during early injection, while re-enabling the mod mid-session prompts cleanly as intended.

/ai-review

Louis047:
Addressed the two valid findings from this round. The thread join in `Wh_ModUninit` now runs unconditionally on handle existence rather than being gated behind `IsMainExplorer()`, so a thread still running mod code can't get orphaned if `Shell_TrayWnd` happens to be gone at uninit time. The cleanup loop was also bounded to 5 s instead of running forever in case the dialog never opens.

On the other points, the $options migration concern doesn't apply here since those settings were never part of a shipped release, so there are no stored string values to worry about. As for moving the hook to `explorer.exe`, that was actually tested and it doesn't intercept the flyout keystrokes (Win+A/N/C etc) - the compositor processes those before explorer's message loop is involved, which is the whole reason dwm.exe is needed. It's not a workaround, it's the only approach that works.

/ready-for-reviewer

Louis047:
Pushed a follow-up fix for the two valid findings from the last review. The thread join in `Wh_ModUninit` was gated behind `IsMainExplorer()` which calls `FindWindowW(L"Shell_TrayWnd")` at uninit time. If Explorer crashes while the mod is loaded, that window is gone, the check returns false, and the entire block including `WaitForSingleObject` and `CloseHandle` gets skipped. That leaves the prompt thread still executing code from the mod image when Windhawk calls `FreeLibrary`, which is a crash. The fix keeps only the uninit prompt itself gated behind `IsMainExplorer()`, while thread cleanup runs purely on whether the handle exists. The unbounded `while` loop was also capped at 50 iterations (5 s) to cover the case where `TaskDialogIndirect` fails to open the window entirely.

On the other findings from last review: the `$options` migration concern doesn't apply since those tiers were never part of a published release, so no user has stored string values in their config. The suggestion to host the hook in `explorer.exe` instead of `dwm.exe` was considered, but `WH_KEYBOARD_LL` in explorer doesn't intercept the flyout shortcuts (Win+A/N/C etc.) because those keystrokes are consumed at the compositor level before explorer's message loop is involved. That's exactly why `dwm.exe` injection is required and not just a convenience choice.

/ai-review

Louis047:
Pushed an update addressing the valid points from the latest review.

1. Added RefreshKeyboardState() to the start of HookThread so key states are cleanly initialized before SetWindowsHookExW runs.
2. Added the /d flag to cmd.exe in PromptForExplorerRestart to bypass AutoRun registry scripts.
3. Removed -lcomctl32 from compiler options since TaskDialogIndirect is dynamically loaded via GetProcAddress.
4. Added a note in the README under Changes in 1.3.0 to let users upgrading from 1.2.1 know that settings were reorganized into three categories and should be re-applied.

Regarding the architectural points raised:

Keeping the low-level hook in dwm.exe rather than switching to a windhawk.exe tool process is a deliberate choice. A windhawk.exe tool process runs at standard user integrity (Medium IL) and cannot intercept or block keystrokes when an elevated or administrator window is focused due to Windows UIPI. Because dwm.exe runs at SYSTEM integrity, placing the hook there guarantees reliable hotkey blocking system-wide regardless of window elevation. The README already clearly explains the dwm.exe inclusion step.

For Special Shortcuts, routing them through the DWM hook instead of blocking them in RegisterHotKey is also intentional. Denying RegisterHotKey for Win+A, Win+N, and similar combos broke opening Action Center, Notifications, and network flyouts via mouse clicks on the taskbar tray. Allowing Explorer to register them while intercepting physical keystrokes in DWM preserves native mouse tray functionality while still preventing unwanted key presses.

/ai-review

Louis047:
Pushed an update covering the valid points from this round:

1. Reverted the pending dialog cleanup in Wh_ModUninit back to the self-terminating while loop. Since TDF_ALLOW_DIALOG_CANCELLATION is set, re-posting WM_CLOSE until the thread exits ensures the dialog closes cleanly without orphaning running thread code.
2. Updated the g_winKeyUsed logic so that only physical keypresses (!isInjected) mark the key as used, preventing synthetic keys from other background tools from disrupting the Start Menu mask.
3. Added VK_ADD and VK_SUBTRACT support for the Magnifier zoom hotkeys alongside the standard plus and minus keys.

Regarding the suggestion to remove the Wh_ModUninit restart prompt entirely:
Explorer does not re-register hotkeys dynamically mid-session once it has started up. If a user turns off or uninstalls this mod, all previously blocked standard hotkeys (like Win+E or Win+R) remain dead at the OS level until Explorer is restarted. Removing the uninit prompt would leave users with permanently non-functional shortcuts whenever they disable the mod. Keeping the prompt on uninit is necessary to ensure the shell is restored properly.

/ai-review

Louis047:
Pushed an update covering the valid points from this round:

1. Reverted the pending dialog cleanup in Wh_ModUninit back to the self-terminating while loop. Since TDF_ALLOW_DIALOG_CANCELLATION is set, re-posting WM_CLOSE until the thread exits ensures the dialog closes cleanly without orphaning running thread code.
2. Updated the g_winKeyUsed logic so that only physical keypresses (!isInjected) mark the key as used, preventing synthetic keys from other background tools from disrupting the Start Menu mask.
3. Added VK_ADD and VK_SUBTRACT support for the Magnifier zoom hotkeys alongside the standard plus and minus keys.

Regarding the suggestion to remove the Wh_ModUninit restart prompt entirely:
Explorer does not re-register hotkeys dynamically mid-session once it has started up. If a user turns off or uninstalls this mod, all previously blocked standard hotkeys (like Win+E or Win+R) remain dead at the OS level until Explorer is restarted. Removing the uninit prompt would leave users with permanently non-functional shortcuts whenever they disable the mod. Keeping the prompt on uninit is necessary to ensure the shell is restored properly.

/ai-review

Louis047:
Pushed an update addressing the uninit restart prompt behavior:

Added `g_blockedAnyRegistration` (atomic bool) set to true in `RegisterHotKey_Hook` whenever Explorer is actively blocked from claiming a standard shortcut. `Wh_ModUninit` now gates its prompt on `g_blockedAnyRegistration` rather than just checking if any standard shortcuts are enabled in settings. This ensures the uninit restart dialog only appears if this Explorer instance was actually prevented from registering hotkeys, eliminating unnecessary prompts during reloads or when standard shortcuts were already owned by Explorer.

Regarding Special Shortcuts without dwm.exe:
The old tier 1 mechanism that blocked flyout registrations in Explorer broke taskbar tray click interactions (Action Center, Network flyout, etc.). Keeping these as plain booleans backed by the DWM hook ensures tray mouse clicks and simulated triggers continue working properly. The README already clearly highlights the dwm.exe inclusion requirement under an Important Installation Step section.

/ai-review

Louis047:
Pushed an update resolving the lifecycle restart flow and thread teardown:

1. Process-scoped Explorer tracking: Replaced the in-memory registration flag with an environment variable `WINDHAWK_DWS_BLOCKED` set inside `explorer.exe` when standard shortcuts are blocked at startup. This persists across mod unloads within the same Explorer process but clears automatically if Explorer restarts or is terminated.
2. Smart mid-session init prompt: `Wh_ModInit` now checks `WINDHAWK_DWS_BLOCKED`. If a user disables the mod, clicks "No" on the restart prompt, and re-enables the mod in that same Explorer process, the redundant prompt is skipped because Explorer still does not own the hotkeys. If the user manually restarts Explorer while the mod is disabled, the new Explorer process starts with a clean environment, and re-enabling correctly prompts for a restart.
3. Reliable uninit prompt: `Wh_ModUninit` now checks `HasAnyStandardShortcutsDisabled()` directly, ensuring that disabling or removing the mod when standard shortcuts are toggled will always prompt to restore default shell hotkeys, regardless of how many disable/re-enable cycles occurred.
4. Safe dialog teardown in step 3: Replaced the single `PostMessage` and infinite wait in `Wh_ModUninit` with the self-terminating re-posting loop matching step 1, eliminating any possibility of an unload deadlock if the dialog window had not yet been assigned.
5. Settings sync: Clears `WINDHAWK_DWS_BLOCKED` if all standard shortcuts are disabled in settings.

Regarding items intentionally left unaddressed:

* Moving the low-level hook from dwm.exe to a windhawk.exe tool process: A tool process runs at standard user integrity (Medium IL) and cannot intercept or suppress keystrokes when an elevated or administrator window has focus due to Windows UIPI. Because the core function of this mod is to reliably block shortcuts system-wide across all windows, hosting the hook in dwm.exe (SYSTEM integrity) is necessary to prevent bypasses when elevated applications are focused. The README explicitly documents the dwm.exe inclusion requirement.

* Removing the Wh_ModUninit restart prompt entirely: Windows Explorer never dynamically re-registers RegisterHotKey bindings mid-session once initialized. When a user disables or uninstalls the mod, their standard shortcuts remain unavailable until Explorer restarts. The uninit prompt is required so users are not left with broken shortcuts after turning off the mod.

/ai-review

Louis047:
Pushed an update replacing the host environment variable with Windhawk internal mod storage:

1. Windhawk mod storage tracking: Removed `SetEnvironmentVariableW`/`GetEnvironmentVariableW` completely. When `RegisterHotKey_Hook` blocks a standard shortcut, it now stores the current Explorer PID using `Wh_SetIntValue(L"blockedInPid", (int)GetCurrentProcessId())` guarded by an atomic one-shot gate (`g_recordedBlock`). This is fully self-contained within Windhawk's per-mod storage, preventing any environment variable pollution in Explorer or child processes.
2. Smart mid-session init prompt: `Wh_ModInit` compares `Wh_GetIntValue(L"blockedInPid", 0)` against `GetCurrentProcessId()`. If the user disables the mod, clicks "No" on the restart prompt, and re-enables in that same Explorer process, the redundant prompt is skipped because Explorer still does not own the hotkeys. If the user manually restarts Explorer, the new Explorer process has a different PID, so `Wh_ModInit` correctly prompts for a restart.
3. Clean uninit prompt: `Wh_ModUninit` now checks the same PID match (`Wh_GetIntValue(L"blockedInPid", 0) == (int)GetCurrentProcessId()`). This ensures the prompt only fires on unload if this specific Explorer process actually had registrations blocked by the hook, avoiding spurious prompts on unloads where nothing was ever blocked.
4. Settings sync: Clears the stored PID value (`Wh_SetIntValue(L"blockedInPid", 0)`) and resets `g_recordedBlock` if all standard shortcuts are turned off in settings.

Regarding items intentionally left unaddressed:

* Moving the low-level hook to a windhawk.exe tool process: As previously detailed, UIPI prevents a Medium-IL tool process from intercepting keystrokes when an elevated / administrator window has focus. Because this mod must block shortcuts reliably across all windows system-wide, hosting the hook in dwm.exe (SYSTEM integrity) is necessary to prevent elevation bypasses. The README clearly documents the dwm.exe inclusion step.

* Dropping the uninit prompt entirely: Explorer does not dynamically re-register RegisterHotKey bindings mid-session once initialized. When a user disables or removes the mod after standard shortcuts were blocked, an Explorer restart is required to restore default Windows shortcuts, so prompting on unload is necessary.

/ai-review

Louis047:
Pushed an update replacing the host environment variable with Windhawk internal mod storage:

1. Windhawk mod storage tracking: Removed `SetEnvironmentVariableW`/`GetEnvironmentVariableW` completely. When `RegisterHotKey_Hook` blocks a standard shortcut, it now stores the current Explorer PID using `Wh_SetIntValue(L"blockedInPid", (int)GetCurrentProcessId())` guarded by an atomic one-shot gate (`g_recordedBlock`). This is fully self-contained within Windhawk's per-mod storage, preventing any environment variable pollution in Explorer or child processes.
2. Smart mid-session init prompt: `Wh_ModInit` compares `Wh_GetIntValue(L"blockedInPid", 0)` against `GetCurrentProcessId()`. If the user disables the mod, clicks "No" on the restart prompt, and re-enables in that same Explorer process, the redundant prompt is skipped because Explorer still does not own the hotkeys. If the user manually restarts Explorer, the new Explorer process has a different PID, so `Wh_ModInit` correctly prompts for a restart.
3. Clean uninit prompt: `Wh_ModUninit` now checks the same PID match (`Wh_GetIntValue(L"blockedInPid", 0) == (int)GetCurrentProcessId()`). This ensures the prompt only fires on unload if this specific Explorer process actually had registrations blocked by the hook, avoiding spurious prompts on unloads where nothing was ever blocked.
4. Settings sync: Clears the stored PID value (`Wh_SetIntValue(L"blockedInPid", 0)`) and resets `g_recordedBlock` if all standard shortcuts are turned off in settings.

Regarding items intentionally left unaddressed:

* Moving the low-level hook to a windhawk.exe tool process: As previously detailed, UIPI prevents a Medium-IL tool process from intercepting keystrokes when an elevated / administrator window has focus. Because this mod must block shortcuts reliably across all windows system-wide, hosting the hook in dwm.exe (SYSTEM integrity) is necessary to prevent elevation bypasses. The README clearly documents the dwm.exe inclusion step.

* Dropping the uninit prompt entirely: Explorer does not dynamically re-register RegisterHotKey bindings mid-session once initialized. When a user disables or removes the mod after standard shortcuts were blocked, an Explorer restart is required to restore default Windows shortcuts, so prompting on unload is necessary.

/ai-review

Louis047:
The latest changes have resolved all functional and lifecycle issues, including replacing the host environment variable with self-contained Windhawk mod storage (`Wh_SetIntValue`/`Wh_GetIntValue`), bounding the prompt thread teardown, and fixing the mid-session disable/re-enable flow.

Regarding the two points maintained across reviews:

1. Wh_ModUninit restart prompt:
Windows Explorer does not dynamically re-register hotkeys mid-session once started. When a user disables or removes this mod after standard shortcuts were blocked, an Explorer restart is required to restore default Windows shortcuts (such as Win+E or Win+R). Removing the prompt on unload would leave users with permanently dead shortcuts upon disabling the mod without immediate feedback. The prompt is safely bounded (30s) and gated on `Wh_GetIntValue("blockedInPid") == GetCurrentProcessId()` so it only appears when this specific Explorer process actually had registrations blocked.

2. Special Shortcuts relying on dwm.exe:
In 1.2.1, blocking Special Shortcuts (like Win+A and Win+N) at the RegisterHotKey level completely broke clicking taskbar tray icons to open Action Center and network flyouts. Moving physical keystroke suppression entirely to the DWM hook preserves native taskbar tray click functionality and allows synthetic triggers to work normally. The README already clearly highlights the dwm.exe process inclusion step under an Important Installation Step section.

The mod has been thoroughly tested across all lifecycle scenarios (disabling, re-enabling, removing, and manual Explorer restarts) and is ready for human review.

/ready-for-reviewer

Louis047:
@m417z PR's ready with the explanation above for a few stuff and core changes done to the mod from extensive testing.

m417z:
> In 1.2.1, blocking Special Shortcuts (like Win+A and Win+N) at the RegisterHotKey level completely broke clicking taskbar tray icons to open Action Center and network flyouts.

I thought we determined that this issue is only relevant for those who use third-party tools such as YASB:
https://github.com/ramensoftware/windhawk-mods/pull/3829#issuecomment-4274110434

So this change seems like a step backwards.

Also, what's the difference between "Special Shortcuts" and "Direct Shortcuts"? Bot are handled with a low-level keyboard hook running in `dwm.exe`, right? What does "at a lower compositor level" mean?

Louis047:
> I thought we determined that this issue is only relevant for those who use third-party tools such as YASB:
https://github.com/ramensoftware/windhawk-mods/pull/3829#issuecomment-4274110434

You're right to catch on this and I did suggest a 3-tier method for special shortcuts. But in the latest versions of YASB, they work just fine, seemingly killing the purpose of "Disable Hotkey" functionality and also it really didn't serve any purpose to its functionality as it felt redundant. The current state of the mod "blocks" the hotkey yet still allows the third-party tools like YASB to invoke the functionality as shown in the video below

https://github.com/user-attachments/assets/ed8cfb84-19d9-4182-a75d-45893cb0e66c

> Also, what's the difference between "Special Shortcuts" and "Direct Shortcuts"? Bot are handled with a low-level keyboard hook running in dwm.exe, right? What does "at a lower compositor level" mean?

Technically, both Special and Direct shortcuts are handled identically under the hood by the low-level keyboard hook in `dwm.exe`. The separation was purely conceptual on my end to distinguish flyouts (which had the YASB simulation aspect) from hardcoded OS hotkeys like `Win` key by itself, `Alt+Shift`, `Win+Tab`, and `Win+Arrows`. And you're right about the wording, "at a lower compositor level" was inaccurate phrasing for intercepting them via the low-level hook before the shell processes them. (I blame AI hallucination)

If having two separate sections is redundant, I can merge them into a single section (for example, "Direct Shortcuts") to keep things clean and straightforward.

Let me know what you think is the best way to structure this.

@m417z

m417z:
> But in the latest versions of YASB, they work just fine, seemingly killing the purpose of "Disable Hotkey"

To my understanding, "Disable Hotkey" is the option that didn't work with YASB all along.

> it really didn't serve any purpose to its functionality as it felt redundant.

"Disable Hotkey" doesn't require targeting dwm.exe, is more lightweight (a global hook isn't needed), and also allows reusing the hotkey for something else.

> If having two separate sections is redundant, I can merge them into a single section (for example, "Direct Shortcuts") to keep things clean and straightforward.

Honestly it seems to me that the way the mod was before the changes is better, and I see no benefit in the changes except the support for the new hotkeys.

Louis047:
> To my understanding, "Disable Hotkey" is the option that didn't work with YASB all along.
> "Disable Hotkey" doesn't require targeting dwm.exe, is more lightweight (a global hook isn't needed), and also allows reusing the hotkey for something else.

Hmm, I really have to test the upstream version of the mod once again as I don't remember much about the functionality currently 🙃 Will get back to you in a few minutes

Louis047:
Yup, you're right. I totally forgot about the true functionality of "Disable Hotkey" just did a test. I'll get the 3-tier approach back. As for the direct shortcuts, I'll bring them under Special Shortcuts as well. From my testing both `Win` key and `Alt+Shift` key can be reused without the "Disable Hotkey" functionality. Rest should be needing them. 

@m417z I will be making the changes now to revert back to the 3-tier option for Special Shortcuts. As for the direct shortcuts, I think I better move both `Win` and `Alt+Shift` alone under Direct Shortcuts since they don't need a explorer restart and simply using the `dwm.exe` hook method would allow them to be registered for other apps. Rest of the Direct Shortcuts shall go under Special Shortcut. What ya say ?

m417z:
> I think I better move both `Win` and `Alt+Shift` alone under Direct Shortcuts since they don't need a explorer restart and simply using the `dwm.exe` hook method would allow them to be registered for other apps.

1.2.1 doesn't have Special Shortcuts, do you still want to keep it for the new version?

What would change if you move `Win` and `Alt+Shift` except for requiring `dwm.exe` (which is by itself not an improvement)?

Louis047:
> 1.2.1 doesn't have Special Shortcuts, do you still want to keep it for the new version?

Uh you mean Direct Shortcuts ?

> What would change if you move Win and Alt+Shift except for requiring dwm.exe (which is by itself not an improvement)?

This is solely for the differentiation purpose. Anyways I'm bringing back the 3-tier approach. But I'm still confused about the parts myself where a shortcut that uses the `dwm.exe` hook method and doesn't require explorer restart to utilize them for third party apps (Example `Win` and `Alt+Shift`) 

I'm already working on SWS mod in parallel so kinda burnt out 😓

m417z:
> Uh you mean Direct Shortcuts ?

Yes, my bad.

> Anyways I'm bringing back the 3-tier approach. But I'm still confused about the parts myself

OK, let me know when it clears up :)

Louis047:
> Yes, my bad.

I'm still trying to figure out what to do with this. I'll get back to you very soon once I'm done moving to the 3-tier approach.

Louis047:
@m417z Did the changes. Do review and let me know if this is good to go. Gonna sleep now 👍

Louis047:
@m417z I think you forgot to check this 😅

m417z:
It's labeled to be waiting for author (you).

Please use the review process:

**Next steps:**

- `/ai-review` - after pushing fixes, to get a review of the updated code. You can repeat this as many times as you need, but each review is thorough and usually there's no need for more than 2-3 iterations.
- `/ready-for-reviewer` - once you're satisfied with the state of the pull request, to hand it over to a human reviewer. If some findings above are left unaddressed, add a short note explaining why.

See the [review process](https://github.com/ramensoftware/windhawk/wiki/Mod-submission-review-process) for details.

Louis047:
My bad, didnt note it.

PR #5173, opened 2026-08-21
Windows 11 File Explorer Styler v1.6

m417z:
* Updated themes for compatibility with recent Windows 11 changes (ViVeTool flag 61017737).
* Added the Float theme.
* Several target controls can now be specified for the same styles by separating them with commas, for example: `ParentClass > Class#Name1, ParentClass > Class#Name2`.
* Style variables now get the value from whichever capturing control is closest. Previously, the value was undefined if there was more than one control publishing the same variable.
* Improved handling for remote images.


AbdulqosimAzizovich:
Great

PR #5174, opened 2026-08-22
Update remove-context-menu-items to 1.12.0

armaninyow:
## Changelog

* Added Polish language support (thanks to [FadeMind](https://github.com/FadeMind))
* Added French language support (thanks to [Catif](https://github.com/Catif))
* Skipped the expensive file-lookup step for every right-click menu unless extension filtering is actually turned on, so menus open faster
* Fixed file selection being pulled from the wrong window/tab, and made it more reliable when using tabs or keyboard shortcuts to open menus
* Fixed a crash that could happen when a menu item with a submenu (like "Open with" or "Send to") got removed
* Added menu filtering limitation to actual file/folder right-click menus, so it no longer accidentally affects the taskbar, tray icons, or Start menu
* Removed Ctrl as a bypass key option (since it's also used for selecting multiple files) and kept Alt as the only bypass key, off by default
* Fixed a crash risk where the mod could message the wrong window while looking up which files were selected, and restored a more reliable way to find that information
* Fixed the left-hand folder tree (navigation pane) being skipped by filtering, so right-clicking there now behaves the same as right-clicking the main file list
* Added support for filtering menu items inside submenus like "Send to," "New," and "Open with," which weren't being caught before
* Fixed right-clicking on empty space not behaving consistently everywhere, including special folders like This PC and Recycle Bin
* Fixed file extension detection breaking on files with very long folder paths
* Blocked a custom rule of just "*" from accidentally wiping out the entire right-click menu


armaninyow:
/ai-review

Thanks for the thorough review. Addressed all four main findings, all seven optional items, and all three functionality notes. Summary below.

Main findings

1. Unconditional COM enumeration - `GetSelectedFilesFromExplorer` is now only called when extension filtering (Notepad or WinRAR) is actually enabled, and the modifier-key bypass check now runs before any of that work.

2. Wrong window for file lookup - Rewrote this to walk up from the hook's actual `hWnd` (as in `explorer-context-menu-custom-items.wh.cpp`) instead of guessing via cursor position, foreground window, or same-process enumeration. Removed `GetSelectedFilesViaAutomation`, `EnumWindowsProc`, and the `CLSID_ShellWindows` path entirely.

3. `ProcessMenu` recursing into a deleted submenu - Added a `deleted` flag so it only recurses into `mii.hSubMenu` when the item wasn't just removed via `DeleteMenu`.

4. Duplicate ja-JP string - Removed the unreachable `共有` entry under `removeSendWithQuickShare`.

A few things the fix for no. 2 surfaced during testing

Rewriting the window-walk logic exposed two additional bugs that weren't part of the original review (not pre-existing, introduced/uncovered by the rewrite itself), both since fixed:
- An unbounded ancestor walk could hit unrelated windows (taskbar/tray/Start menu) and treat their `WM_USER+7` reply as a real `IShellBrowser*`, causing a CFG violation crash (`0xc0000409`). Fixed by requiring a `SHELLDLL_DefView` ancestor before doing any `WM_USER+7` messaging, and bounding the walk to stop at the first recognized shell frame class.
- I mistakenly added a `pShellBrowser->Release()` on what turned out to be a *borrowed* pointer (by convention, `WM_USER+7` doesn't hand back an added reference), which caused a delayed heap-corruption crash (`0xc0000374`) after repeated use. Removed.

Optional improvements

- Loops now break on an empty string (the real "unset" signal from `Wh_GetStringSetting`) instead of checking for `NULL`.
- Switched both hooks to `WindhawkUtils::SetFunctionHook` for compile-time signature safety.
- Added a `std::mutex` around `g_settings`/`g_menuItems`, scoped narrowly to avoid holding it across `GetSelectedFilesFromExplorer`'s cross-thread `SendMessageTimeoutW` call (an earlier broader version of this lock caused an `AppHangB1` deadlock, since fixed).
- Precomputed each menu item's normalized text once in `InitializeMenuItems()` instead of on every `ShouldRemoveMenuItem` lookup.
- Removed the unused `<unordered_set>` include, the duplicate `shlwapi.lib` pragma, and the unread `LRESULT result`.
- `GetAncestor`/`GA_ROOT` note is moot, the function it applied to was removed along with the automation path.
- Fixed `jp-JP` → `ja-JP` throughout.

Functionality notes

- `ProcessMenu` now only runs when `IsShellViewWindow(hWnd)` is true, so filtering no longer touches the taskbar, tray, or Start menu.
- Dropped Ctrl as a bypass key (too easy to trigger accidentally via multi-select). Alt is now the only bypass key, and the toggle is off by default.
- Added a fallback to the current folder (via `IPersistFolder2::GetCurFolder`) when nothing is selected, so extension filtering has a defined input instead of silently doing nothing.

Ready for another pass whenever convenient. Happy to run `/ai-review` again.

armaninyow:
/ai-review

Addressed both main findings and all items in the optional/functionality sections below.

Main findings

1. Blind `WM_USER+7` probing in `GetShellBrowser` - This took a few iterations to get right. I first tried sending only to `SHELLDLL_DefView` (per your suggested fix), then to `SHELLDLL_DefView`/known frame classes, then to `GetParent(hWndShellView)` per the documented `CWM_GETISHELLBROWSER` pattern , all three failed empirically (confirmed via logging: `GetShellBrowser` consistently got no answer, causing extension filtering to silently stop working). What's actually in place now is closer to the original approach: climb from `hwnd` via `GetParent`, sending the message at each level until something answers, stopping once a top-level frame class (`CabinetWClass`/`ExploreWClass`/`Progman`) is reached. The safety property your finding cared about is still preserved differently , `GetShellBrowser` is only ever invoked after `IsShellViewWindow`/`FindShellViewWindow` has already confirmed `hwnd` sits within a genuine Explorer folder view, so the climb stays confined to Explorer's own internal window hierarchy rather than probing arbitrary windows process-wide (taskbar/tray/Start menu are excluded before this function is ever called). Happy to revisit if there's a more targeted fix that actually holds up against real testing.

2. `SHELLDLL_DefView` gate excluding the navigation pane - Fixed. `IsShellViewWindow` now also accepts a `NamespaceTreeControl`/`SysTreeView32` ancestor, so right-clicks in the folder tree get filtered again, matching 1.11.0 behavior. File/extension lookup still correctly only applies to genuine `SHELLDLL_DefView` menus (via a separate `FindShellViewWindow` check), since there's no `IFolderView` selection to read in the nav pane.

Optional improvements

- Removed the dead duplicate toggle re-check in `ShouldRemoveByExtension`.
- Removed `-loleaut32`, `<exdisp.h>`, `<shlguid.h>` - confirmed no automation types (`BSTR`/`VARIANT`/etc.) remain anywhere.
- Merged `TrackPopupMenu_Hook`/`TrackPopupMenuEx_Hook` into a shared `ProcessPopupMenu` helper.
- Hoisted the `IsShellViewWindow` check to the top of that helper, so non-shell-view menus skip `CoInitialize`/file lookup entirely instead of doing it and discarding the result.
- Added a `g_menuItemsByText` hash map (built once in `InitializeMenuItems`) so `ShouldRemoveMenuItem` is a single lookup instead of a linear scan; custom items are now normalized once in `LoadSettings` instead of per comparison.
- Fixed the remaining `%d`/`size_t` mismatch in `AnyFileHasExtensionInList`.
- Fixed both transcription slips (`Designeで作成` → `Designerで作成`, and the double space in the Russian Paint entry).

Functionality notes

- `NormalizeString` now truncates at the first `\t` before comparing, so accelerator-suffixed text (`"Copy\tCtrl+C"`) still matches.
- Added a `DispatchMessageW` hook that catches `WM_INITMENUPOPUP` after Explorer's own handler has populated a submenu, then runs `ProcessMenu` on it, so "Send to"/"New"/"Open with" contents are now filterable, not just their parent items.
- Removed the `IPersistFolder2::GetCurFolder` folder-path fallback entirely and instead made `ShouldRemoveByExtension` explicitly hide extension-gated items when there's no file context at all, regardless of whether a filesystem path is available, so behavior is now consistent between regular folders and virtual ones (This PC, Recycle Bin, Libraries, etc.) on an empty-space right-click.
- Switched from `StrRetToBufW` (fixed `MAX_PATH` buffer) to `StrRetToStrW` (heap-allocated, uncapped) so deep paths no longer get truncated and lose their extension.
- `ProcessMenu` now tracks whether it actually deleted anything and skips the separator-cleanup pass entirely when it didn't.
- `LoadSettings` now rejects a bare `*` custom item (logged, not silently dropped) so it can't accidentally empty the whole context menu. Also added a log warning if more than 100 custom items are configured, since that cap was previously silent.

Ready for another pass, `/ai-review` whenever convenient.

armaninyow:
/ai-review

Thanks for catching #1 in particular, that one's on me. I misunderstood how `WM_INITMENUPOPUP` is delivered and shipped a hook that could never fire; good call asking for log verification, since that's exactly what should have caught it before claiming it was done.

Main findings

1. `DispatchMessageW` can't see `WM_INITMENUPOPUP` is fixed. Replaced it with a thread-local `WH_CALLWNDPROCRET` hook installed only around the `TrackPopupMenu(Ex)` call (matching the pattern in `photoshop-dark-menus.wh.cpp`), with a depth counter for `TPM_RECURSE` nesting and hook-handle tracking + forced cleanup in `Wh_ModUninit` so a still-installed hook can't outlive the mod image.

2. `g_threadFilePaths` cleared before the menu is shown is fixed as part of the same change: it's now populated before the original `TrackPopupMenu(Ex)` call and only cleared once the outermost call returns, so submenus filtered via the new hook see the correct selection.

3. `GetShellBrowser` probing arbitrary windows is fixed with a class allowlist (`SHELLDLL_DefView`, `ShellTabWindowClass`, plus the frame classes), gated *before* sending rather than after. Worth flagging: this is my fourth attempt at getting this function right, and the previous three "theoretically correct" narrower targets (only `SHELLDLL_DefView`, only its parent, etc.) all failed empirically in real testing, which is why I widened it slightly and added logging of every probe (sent or skipped) this time, so if this still doesn't find an answer on some build, the log will show exactly which class needs to be added instead of another guessing round.

Optional improvements

- `GetFileExtension` now uses `PathFindExtensionW` instead of a raw `find_last_of('.')` over the full path.
- Merged the bypass-check and `needFiles` reads in `ProcessPopupMenu` into a single lock scope (kept the `ProcessMenu` lock separate on purpose, since merging it in would reintroduce the cross-thread deadlock risk from an earlier round).
- `g_menuItemsByText` now stores indices into `g_menuItems` instead of raw pointers.
- Extended the overflow-warning log (previously only on `customItems`) to `notepadExtensions` and `winrarExtensions` too.
- Added explicit `<cwctype>`/`<cwchar>` includes instead of relying on transitive inclusion.

Functionality notes

- ru-RU/ja-JP translation gaps - leaving as-is for now; the contributors who did those translations aren't available to fill in the missing strings. Noted for a future pass.
- Desktop menu regression (`Progman`/`WorkerW`) is confirmed and fixed. `IsShellViewWindow` now accepts `Progman`/`WorkerW` directly, since on the desktop the def-view is a *child* of the menu's owner window, not an ancestor.
- Empty-space right-click hiding Notepad/WinRAR is checked; this isn't actually reachable, since Windows itself never shows those items on an empty-space right-click regardless of the mod. No change needed.
- `enableModifierOverride` default flip - leaving as-is; not considered a problem worth a changelog callout.
- ~560 hardcoded localized strings drifting over time is acknowledged, no action needed (Custom Items is the intended mitigation, already documented).

This round closes out everything actionable from this review. I think the mod is in a solid, well-tested state at this point. I'll be treating this as the last planned iteration on my end.

armaninyow:
/ai-review

Thanks for catching no. 1 in particular, that one's on me. I misunderstood how `WM_INITMENUPOPUP` is delivered and shipped a hook that could never fire; good call asking for log verification, since that's exactly what should have caught it before claiming it was done.

Main findings

1. `DispatchMessageW` can't see `WM_INITMENUPOPUP` is fixed. Replaced it with a thread-local `WH_CALLWNDPROCRET` hook installed only around the `TrackPopupMenu(Ex)` call (matching the pattern in `photoshop-dark-menus.wh.cpp`), with a depth counter for `TPM_RECURSE` nesting and hook-handle tracking + forced cleanup in `Wh_ModUninit` so a still-installed hook can't outlive the mod image.

2. `g_threadFilePaths` cleared before the menu is shown is fixed as part of the same change: it's now populated before the original `TrackPopupMenu(Ex)` call and only cleared once the outermost call returns, so submenus filtered via the new hook see the correct selection.

3. `GetShellBrowser` probing arbitrary windows is fixed with a class allowlist (`SHELLDLL_DefView`, `ShellTabWindowClass`, plus the frame classes), gated *before* sending rather than after. Worth flagging: this is my fourth attempt at getting this function right, and the previous three "theoretically correct" narrower targets (only `SHELLDLL_DefView`, only its parent, etc.) all failed empirically in real testing, which is why I widened it slightly and added logging of every probe (sent or skipped) this time, so if this still doesn't find an answer on some build, the log will show exactly which class needs to be added instead of another guessing round.

Optional improvements

- `GetFileExtension` now uses `PathFindExtensionW` instead of a raw `find_last_of('.')` over the full path.
- Merged the bypass-check and `needFiles` reads in `ProcessPopupMenu` into a single lock scope (kept the `ProcessMenu` lock separate on purpose, since merging it in would reintroduce the cross-thread deadlock risk from an earlier round).
- `g_menuItemsByText` now stores indices into `g_menuItems` instead of raw pointers.
- Extended the overflow-warning log (previously only on `customItems`) to `notepadExtensions` and `winrarExtensions` too.
- Added explicit `<cwctype>`/`<cwchar>` includes instead of relying on transitive inclusion.

Functionality notes

- ru-RU/ja-JP translation gaps - leaving as-is for now; the contributors who did those translations aren't available to fill in the missing strings. Noted for a future pass.
- Desktop menu regression (`Progman`/`WorkerW`) is confirmed and fixed. `IsShellViewWindow` now accepts `Progman`/`WorkerW` directly, since on the desktop the def-view is a *child* of the menu's owner window, not an ancestor.
- Empty-space right-click hiding Notepad/WinRAR is checked; this isn't actually reachable, since Windows itself never shows those items on an empty-space right-click regardless of the mod. No change needed.
- `enableModifierOverride` default flip - leaving as-is; not considered a problem worth a changelog callout.
- ~560 hardcoded localized strings drifting over time is acknowledged, no action needed (Custom Items is the intended mitigation, already documented).

This round closes out everything actionable from this review. I think the mod is in a solid, well-tested state at this point. I'll be treating this as the last planned iteration on my end.

m417z:
Why is the main item in the last review (about `GetSelectedDesktopFiles`) not addressed?

armaninyow:
/ai-review

Thanks for catching this. Confirmed and fixed, and it's a good example of why "the message was delivered" isn't the same as "the answer is correct."

Main finding

`GetSelectedDesktopFiles` scraping list-view display text is fixed. Replaced the `SysListView32`/`ParseDisplayName` approach entirely with `IShellWindows::FindWindowSW(SWC_DESKTOP)` → `IServiceProvider::QueryService(SID_STopLevelBrowser)`, then reused the same `IFolderView`/`GetDisplayNameOf(SHGDN_FORPARSING)` selection-enumeration code already trusted for regular folder windows (factored out into a shared `GetSelectedFilesFromShellBrowser` helper). This gets real PIDLs regardless of whether Explorer is hiding extensions on screen, and also eliminates the cross-process `SendMessage` pointer risk you flagged with the unscoped `FindWindowW`/`FindWindowExW` search.

Optional improvements

- Removed the now-dead `Progman`/`WorkerW` entries from `GetShellBrowser`'s class lists (unreachable since `IsDesktopWindow` routes those to the new desktop path first).
- Collapsed `IsShellViewWindow`'s two separate ancestor walks into one.
- Switched all four ancestor-walking functions from `GetParent` to `GetAncestor(h, GA_PARENT)`, since `GetParent` returns a top-level window's owner rather than stopping there.
- Converted the three settings-array loops and the overflow-warning helper to `WindhawkUtils::StringSetting`.
- Pulled the shared `100`-item cap into a `constexpr kMaxSettingArrayItems`.
- `WarnIfSettingArrayOverflows` now logs a clean display name instead of the raw `[%d]`-formatted settings key.
- Added an RAII guard (`MenuTrackingGuard`) around `EnterMenuTracking`/`ExitMenuTracking` so an exception between them can't leave the hook installed indefinitely.
- Fixed the comment above `FindShellViewWindow` to reflect that the frame/tab classes are the established target, not just `SHELLDLL_DefView`.

Functionality notes

- Alt bypass re-evaluated per popup is fixed. The decision is now made once per tracking session (in `EnterMenuTracking`) and reused for the top-level menu and every submenu, so it can't go inconsistent if the key state changes mid-session.
- Top-level menu processed twice is fixed. `ProcessPopupMenu` now only calls `ProcessMenu` directly as a fallback if the `WH_CALLWNDPROCRET` hook failed to install; otherwise the hook's own `WM_INITMENUPOPUP` handling covers the top-level menu too, same as it already did for submenus.
- Nav-pane right-clicks always hide extension-filtered items - I looked into `INameSpaceTreeControl::GetSelectedItems` as suggested, but couldn't find a documented/verified way to obtain that interface from an existing nav-pane `HWND` we didn't create ourselves (unlike the other fixes this round, which had working reference implementations to follow). Given how costly guessing at undocumented Windows internals has been earlier in this PR's history, I went with your suggested fallback instead: documented the limitation directly in the Extension Filtering setting's description rather than shipping a speculative implementation.
- CJK accelerator suffixes (`ja-JP`) won't be addressed this round; holding off per current priorities. Noted for a future pass.

At this point I think the mod has been through a genuinely thorough review cycle. Four rounds catching real, substantive issues each time, several of which surfaced from real-device testing rather than just static review. I'm going to treat this as a good stopping point for iteration and move toward `ready-for-reviewer` rather than continuing to `/ai-review` indefinitely. Appreciate the depth of these reviews, they meaningfully improved the mod's correctness.

armaninyow:
I want to know if maybe the windhawk-reviewer Bot didn't realize it was running in circles

armaninyow:
PLEASE do not flag again that is related to the translations/supported languages. Those are not actual bugs

PLEASE do not flag again the leading dot on extensions. If users cant read the note "Use lowercase with dot (e.g., .txt)" then thats on them. It's not an issue, not a bug

PR #5175, opened 2026-08-22
Add Taskbar System Info mod

starychenko:
Adds Taskbar System Info, a compact, glanceable far-left Windows 11 taskbar dashboard for CPU, GPU, RAM and VRAM. It provides a fixed 2x2 layout, 60-second CPU/GPU history graphs, RAM/VRAM capacity bars, configurable alert thresholds, native GPU temperature through D3DKMT, optional HWiNFO sensors and Ukrainian localization.

Relationship to Taskbar Clock Customization: the mods share some Windows metric sources, but not the UI. Taskbar Clock Customization substitutes scalar text into clock lines. This mod injects a separate far-left dashboard that keeps all four resources visible simultaneously, renders rolling XAML sparklines and capacity bars, can reserve its own taskbar space, and leaves the clock available for time, date and weather. Those graphical history and layout behaviors can't be expressed by a clock placeholder format string; adding them there would introduce a separate taskbar surface rather than customize the clock.

Companion repository and screenshot: https://github.com/starychenko/windhawk-taskbar-system-info

Validation performed:

- Official `.github/pr_validation.py`: no warnings.
- Clean x86-64 build and ARM64 compatibility build with the compiler bundled with Windhawk.
- Live DXGI/PDH/D3DKMT smoke-test on an AMD Radeon RX 7900 XTX, including adapter-LUID pairing, GPU/VRAM range checks and an HWiNFO-independent GPU temperature reading.

## Changelog

* Not applicable - initial release.

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [x] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/taskbar-system-info.wh.cpp
+++ b/mods/taskbar-system-info.wh.cpp
@@ -1869,6 +1869,10 @@ RefCountBase_Decref_t RefCountBase_Decref_Original = nullptr;
 void* CTaskBand_ITaskListWndSite_vftable = nullptr;
 
 XamlRoot GetTaskbarXamlRoot(HWND taskbarWindow) {
+#if !defined(_M_X64)
+    (void)taskbarWindow;
+    return nullptr;
+#else
     if (!CTaskBand_GetTaskbarHost_Original ||
         !TaskbarHost_FrameHeight_Original || !RefCountBase_Decref_Original ||
         !CTaskBand_ITaskListWndSite_vftable) {
@@ -1908,7 +1912,6 @@ XamlRoot GetTaskbarXamlRoot(HWND taskbarWindow) {
     }
 
     size_t elementOffset = 0x10;
-#if defined(_M_X64)
     const BYTE* code =
         reinterpret_cast<const BYTE*>(TaskbarHost_FrameHeight_Original);
     if (code[0] == 0x48 && code[1] == 0x83 && code[2] == 0xEC &&
@@ -1918,9 +1921,6 @@ XamlRoot GetTaskbarXamlRoot(HWND taskbarWindow) {
     } else {
         Wh_Log(L"Unsupported TaskbarHost::FrameHeight pattern");
     }
-#else
-#error Unsupported architecture
-#endif
 
     auto* elementUnknown = *reinterpret_cast<::IUnknown**>(
         static_cast<BYTE*>(taskbarHostSharedPtr[0]) + elementOffset);
@@ -1935,6 +1935,7 @@ XamlRoot GetTaskbarXamlRoot(HWND taskbarWindow) {
     XamlRoot result = taskbarElement ? taskbarElement.XamlRoot() : nullptr;
     RefCountBase_Decref_Original(taskbarHostSharedPtr[1]);
     return result;
+#endif
 }
 
 void ApplyToCurrentTaskbar(void*) {
--- a/mods/taskbar-system-info.wh.cpp
+++ b/mods/taskbar-system-info.wh.cpp
@@ -4,7 +4,7 @@
 // @name:uk-UA      Системний монітор панелі завдань
 // @description     A quiet two-column CPU, GPU, RAM and VRAM monitor with 60-second history graphs for the Windows 11 taskbar.
 // @description:uk-UA Компактний монітор CPU, GPU, RAM і VRAM із 60-секундними графіками для панелі завдань Windows 11.
-// @version         1.2.0
+// @version         1.3.0
 // @author          Yevhenii Starychenko
 // @github          https://github.com/starychenko
 // @homepage        https://github.com/starychenko/windhawk-taskbar-system-info
@@ -17,6 +17,9 @@
 // Taskbar XAML discovery and window-thread marshaling are based on techniques
 // from "Multirow taskbar for Windows 11" by Michael Maltsev (m417z):
 // https://github.com/ramensoftware/windhawk-mods/blob/main/mods/taskbar-multirow.wh.cpp
+// Native GPU temperature collection via D3DKMT follows Taskbar Clock
+// Customization by Michael Maltsev (m417z):
+// https://github.com/m417z/my-windhawk-mods/commit/861920df6380f4c13abec5d9226362c4725e8362
 // Both projects are distributed under the GNU General Public License v3.0.
 
 // ==WindhawkModReadme==
@@ -54,6 +57,7 @@ stable 2x2 dashboard with rolling graphs, capacity bars and temperature alerts.
 - GPU utilization and dedicated VRAM usage from Windows PDH counters.
 - Dedicated VRAM capacity and adapter identity from DXGI.
 - CPU and GPU temperatures from HWiNFO when available.
+- GPU fallback from the Windows display-driver interface (D3DKMT).
 - CPU fallback from Windows ACPI thermal zones exposed through PDH.
 
 Metric collection runs on a worker thread. The taskbar UI thread only renders
@@ -68,18 +72,18 @@ matched to the selected DXGI adapter by LUID.
 The **Temperature source** setting provides these modes:
 
 - **Automatic** fills CPU and GPU independently: HWiNFO shared memory first,
-  then HWiNFO Gadget Registry, then Windows thermal zones for a still-missing
-  CPU reading.
+  then HWiNFO Gadget Registry, then Windows D3DKMT for a still-missing GPU
+  reading and Windows thermal zones for a still-missing CPU reading.
 - **HWiNFO automatic** uses only the two HWiNFO interfaces.
 - **HWiNFO Shared Memory** uses only `Global\\HWiNFO_SENS_SM2`.
 - **HWiNFO Gadget Registry** uses only
   `HKCU\\Software\\HWiNFO64\\VSB`.
-- **Windows thermal zones** uses the same
+- **Windows native** reads GPU temperature from the selected display driver via
+  D3DKMT and CPU temperature from the same
   `\\Thermal Zone Information(*)\\Temperature` PDH source as Taskbar Clock
-  Customization. It is dependency-free, but it exposes ACPI platform zones,
-  not necessarily the CPU package sensor, and does not provide a generic GPU
-  temperature. The optional zone filter and average/hottest setting make this
-  fallback explicit and controllable.
+  Customization. It needs no third-party monitor. ACPI platform zones don't
+  necessarily represent the CPU package sensor; the optional zone filter and
+  average/hottest setting make this fallback explicit and controllable.
 - **Disabled** skips temperature collection while keeping every other metric.
 
 HWiNFO is optional and is not bundled with this mod. Shared-memory integration
@@ -107,7 +111,9 @@ noise every second.
 
 Taskbar discovery and window-thread marshaling follow techniques from
 [Multirow taskbar for Windows 11](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/taskbar-multirow.wh.cpp)
-by Michael Maltsev (`m417z`). Released under GPL-3.0.
+by Michael Maltsev (`m417z`). Native GPU temperature collection follows his
+[Taskbar Clock Customization implementation](https://github.com/m417z/my-windhawk-mods/commit/861920df6380f4c13abec5d9226362c4725e8362).
+Released under GPL-3.0.
 */
 // ==/WindhawkModReadme==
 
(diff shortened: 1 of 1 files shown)


starychenko:
Fixed the compatibility workflow failure in commit 8195f0ce: non-x64 builds now use a safe no-op path while runtime support remains explicitly x86-64. Verified locally with the repository's compile_mod.py for both x86_64 and aarch64 targets, with no warnings.

starychenko:
Addressed the review findings in f37495c6: (1) XAML globals now use [[clang::no_destroy]], Loaded revokers use an optional container with UI-thread reset and unload guards; (2) teardown reuses the remembered taskbar HWND/thread, uses bounded dispatch with diagnostics, and stops the metrics worker before UI cleanup; (3) added the reference ARM64 TaskbarHost::FrameHeight pattern while retaining the stricter x64 prologue check; (4) PDH initialization is now lazy and starts only after widget injection; (5) all CPU/GPU/RAM/VRAM/HWiNFO collection runs on a joined worker thread, while the taskbar thread only renders completed snapshots; (6) the README now records how this differs from Taskbar Clock Customization. Also applied the safe optional cleanup items for StringSetting RAII, the unused version library/signature/source state, registered-message reuse, and documented setting clamps. Verified with the repository validator, compile_mod.py for x86_64 and aarch64 without warnings, a clean local x64 build, and the live DXGI/PDH smoke-test.

starychenko:
Addressed the second review in 025960a3:

1. RunFromWindowThread now uses synchronous SendMessageW, matching the reference implementation. The stack payload remains valid until the taskbar-thread callback completes, and teardown can no longer time out while callbacks or the DispatcherTimer are still registered.
2. TaskbarFrame::Loaded now resolves and applies only to the primary Shell_TrayWnd instead of injecting the sender frame, preventing a secondary taskbar from taking over the single widget state.
3. GetTaskbarXamlRoot now releases the shared pointer and returns nullptr when the FrameHeight prologue is unknown; it never dereferences a guessed offset.
4. A same-name widget is reused only when it is the current g_widget. Any stale instance is removed and rebuilt.
5. The overlap with Taskbar Clock Customization is documented in the README; the separate far-left native XAML dashboard, rolling graphs, capacity bars, and thresholds are intentional, with final acceptance left to the maintainer.

Also added the direct include/std::memcpy cleanup, fixed LUID/HRESULT varargs types, and tightened color parsing.

The HWiNFO packed layout was checked against the published layout and with a live mapping on HWiNFO64: signature 0x53695748, version 2/revision 1, 48-byte header, 392-byte sensor stride, 460-byte reading stride, with valid CPU and GPU temperature readings. The README now states HWiNFO 7.0+ and the free edition's 12-hour SHM limit. Since HWiNFO 7.0, the interface allows full disclosure: https://www.hwinfo.com/forum/threads/important-changes-to-hwinfo64-coming-soon.7092/

Verification passed: source validator, repository PR validator, clean x64 build with warnings enabled, repository compatibility builds for x86_64 and ARM64, live DXGI/PDH smoke test, and live HWiNFO SHM smoke test.

starychenko:
Addressed the valid finding in 24c4adc0: taskbar.dll symbol resolution is now a hard initialization requirement. Wh_ModInit logs the failure and returns FALSE, so the mod can no longer load successfully while remaining inert.

The HWiNFO packing finding does not hold. The published shared-memory header explicitly says "No alignment of structure members" and uses #pragma pack(1):
https://github.com/MintyMods/MintySensorMonitor/blob/master/hwinfo/hwisenssm2.h_v6.11.3917

HWiNFO's author also confirmed that the Shared Memory Interface is public and the header may be distributed:
https://www.hwinfo.com/forum/threads/introducing-remote-sensor-monitor-a-restful-web-server.1025/page-4#post-55280

This was verified against the live HWiNFO64 mapping on the target machine:
- signature 0x53695748
- version 2, revision 1
- header size 48
- sensor stride 392
- reading stride 460
- CPU (Tctl/Tdie) 72.5 C
- GPU Temperature 52 C

The packed prefix offsets are therefore correct: pollTime=12, sensorOffset=20, value=284. Commit 24c4adc0 adds compile-time static_assert checks for all of these offsets and sizes so the layout cannot drift silently. Changing to natural alignment as suggested would break the confirmed working parser.

Also added the direct <cstddef> and <iterator> includes. Verification passed again with the source validator, repository PR validator, clean warning-enabled x64 build, and repository x86_64/ARM64 compatibility builds.

starychenko:
Added selectable temperature sources in 28d48f9f (v1.1.0):

- `Automatic`: Shared Memory first, then fills missing CPU/GPU readings from Gadget Registry.
- `HWiNFO Shared Memory`: uses only `Global\HWiNFO_SENS_SM2`.
- `HWiNFO Gadget Registry`: uses only `HKCU\Software\HWiNFO64\VSB`.
- `Disabled`: skips temperatures while keeping all other metrics.

The Gadget parser now validates Celsius/Fahrenheit units, tolerates small index gaps, and recognizes localized HWiNFO labels (verified with Ukrainian `ЦП (Tctl/Tdie)` and `Температура GPU`). HWiNFO remains optional and is not bundled or modified; the mod doesn't re-enable or automate Shared Memory.

Verification passed locally: source validator, repository PR validator, warning-enabled x64 build, x86_64/ARM64 compatibility builds, live DXGI/PDH metrics smoke test, live Shared Memory temperatures, and live Gadget Registry temperatures.

starychenko:
Added the Windows-native temperature provider and a provider strategy in 39c75b73 (v1.2.0), based on the approach used by Taskbar Clock Customization:

- `Automatic`: fills CPU/GPU independently via HWiNFO Shared Memory, then HWiNFO Gadget Registry, then Windows thermal zones for a still-missing CPU reading.
- `HWiNFO automatic`: restricts fallback to the two HWiNFO interfaces.
- Explicit modes remain available for Shared Memory, Gadget Registry, Windows thermal zones, and Disabled.
- Windows uses `\Thermal Zone Information(*)\Temperature` through PDH, validates counter status/range, converts Kelvin to Celsius, and never claims to provide a generic GPU temperature.
- Added an optional case-insensitive thermal-zone instance filter and Average/Hottest aggregation. Average matches Taskbar Clock Customization; Hottest is useful for alert-oriented monitoring.
- CPU/GPU provider provenance is retained per snapshot and logged only when it changes.

The README explicitly distinguishes ACPI platform zones from a true CPU package sensor. On the hardware-tested desktop, HWiNFO Shared Memory and localized Gadget Registry both return matching CPU/GPU readings; Windows accepts the thermal counter but the firmware exposes no valid zone instances, so the provider correctly remains unavailable instead of fabricating a value.

Verification passed: standalone source validator (v1.2.0, 25 settings), repository PR validator, warning-enabled x64 build, repository x86_64/ARM64 compatibility builds, live DXGI/PDH metrics smoke test, live HWiNFO Shared Memory test, and live HWiNFO Gadget Registry test.

m417z:
The latest development version of Taskbar Clock Customization retrieves GPU temperature too, you might find it useful:
https://github.com/m417z/my-windhawk-mods/commit/861920df6380f4c13abec5d9226362c4725e8362

starychenko:
Thanks for the detailed review. I verified both requested items against the active runtime data and current accepted implementations.

### 1. HWiNFO Shared Memory layout

I re-ran a raw mapping inspection against HWiNFO64 8.22, rather than relying only on labels or a partial parse. The active `Global\\HWiNFO_SENS_SM2` mapping, read at the packed offsets used by this mod, reports:

```
signature=0x53695748 version=2 revision=1
sensorCount=19 sensorOffset=48 sensorStride=392
readingCount=383 readingOffset=7496 readingStride=460
pollingPeriod=2000
```

Interpreting those same bytes with the proposed default-aligned offsets produces `sensorOffset=392`, `sensorStride=19`, `sensorCount=7496`, etc., which is invalid. For the first CPU temperature reading, `CPU (Tctl/Tdie)`, the double at offset 284 is a valid live value (66.0 C during the probe), while reading at offset 288 produces `5.33e-315`.

The standalone reader exited successfully and enumerated the current CPU and GPU temperatures. Therefore the live v2 Shared Memory layout exposed by HWiNFO64 8.22 is one-byte packed, and removing `#pragma pack(1)` would break the provider. The 320-vs-292 size comparison also mixes the complete reading element (including min/max/average values) with this mod's deliberately shortened prefix, which ends after the first `Value`.

The parser already validates the section offsets and strides before dereferencing them, so an incompatible future layout fails closed. I agree that a throttled, one-time diagnostic for a rejected header could be useful, but logging the complete rejection on every one-second poll would be undesirable.

### 2. TaskbarHost offset detection

The ARM64 `ldr x8, [x0, #imm]!` decoder is an established Windhawk implementation, not a speculative analogue. The same opcode mask and immediate extraction are present in accepted mods, for example:

- https://github.com/ramensoftware/windhawk-mods/blob/782f5af4ee10ffeefc60227ec7bc30c32c884236/mods/adaptive-microphone-icon-visibility.wh.cpp#L449-L475
- https://github.com/ramensoftware/windhawk-mods/blob/782f5af4ee10ffeefc60227ec7bc30c32c884236/mods/taskbar-multi-tray.wh.cpp#L4489-L4522
- https://github.com/ramensoftware/windhawk-mods/blob/782f5af4ee10ffeefc60227ec7bc30c32c884236/mods/taskbar-folder-menus.wh.cpp#L830-L853

Returning `nullptr` for an unknown prologue is intentional fail-closed behavior. A blind `0x48` fallback is not universally safe: several newer accepted implementations use `0x10` as their fallback, and dereferencing an obsolete offset as an `IUnknown*` can crash Explorer instead of merely hiding the widget. The current x64 path is live-tested; ARM64 remains compilation-tested only, as documented.

For these reasons I am retaining the packed HWiNFO layout and the safe offset-detection behavior. I reviewed the optional suggestions as well; several are useful follow-up optimizations, but they don't indicate a correctness failure in the submitted implementation.

starychenko:
Thanks for adding this and for pointing it out. We'll keep the D3DKMT GPU temperature implementation in mind.

For our use case, a separate always-visible taskbar dashboard is more convenient, since it provides quick access to CPU, GPU, RAM and VRAM information without expanding the clock area.

The original reason for creating this mod was that `%cpu_temp%` in Taskbar Clock Customization wasn't available on this machine because the firmware exposed no valid Windows thermal-zone counters. This varies between devices and isn't an issue with the clock mod itself, so Taskbar System Info also supports additional optional temperature sources.

m417z:
What I meant is that you can add it to your mod so that users will be able to see GPU temperature without having to install HWiNFO.

starychenko:
@m417z Thanks for the clarification, and thanks for adding this to Taskbar Clock Customization. I've now integrated the D3DKMT approach in 98d67715.

Taskbar System Info now reads GPU temperature from the selected DXGI adapter's LUID without requiring HWiNFO. In Automatic mode, HWiNFO remains the first choice when available, and D3DKMT fills a missing GPU temperature. The explicit Windows native mode uses D3DKMT for GPU and ACPI/PDH for CPU. The previous `windowsThermalZones` setting value remains accepted for compatibility.

Verified live on an AMD Radeon RX 7900 XTX with an HWiNFO-independent smoke test (52 °C). Source validation and both x64 and ARM64 compatibility builds pass.

starychenko:
Thanks for the review. I verified and addressed the two concrete runtime issues in 5e0a1601, and clarified the overlap concern in the PR description.

1. The PR description now explicitly distinguishes the UI from Taskbar Clock Customization: this is a separate far-left fixed 2x2 dashboard with simultaneous metrics, rolling XAML sparklines, capacity bars, alert thresholds and optional reserved space. It leaves the clock surface untouched. These graphical/history/layout behaviors cannot be represented by a clock placeholder format string.
2. UI teardown is now tracked. If taskbar-thread dispatch fails in `Wh_ModBeforeUninit`, `Wh_ModUninit` retries `RemoveFromCurrentTaskbar` before the image is unloaded.
3. Missing PDH counters are no longer treated as permanently unavailable. The worker retries only the missing counters every 30 seconds, while retaining counters that were added successfully. A failed query open is retried on the same schedule.

I also applied two directly related optional fixes: negative DXGI adapter-filter results are cached and logged once, and D3DKMT exports now use the already-loaded `gdi32.dll` module without incrementing an unreleased module reference.

Revalidated with the official source validator, warning-free x64 build, ARM64 compatibility build and the live DXGI/PDH/D3DKMT smoke test (51 °C GPU reading without HWiNFO). The other optional or broader feature suggestions are intentionally left outside this focused fix.

starychenko:
Thanks. Both confirmed issues are addressed in 44034fff.

1. `ReadHwInfoSharedMemory` now copies `HwInfoHeader` into a local snapshot before validating any ranges. Every loop bound, offset and stride uses only that validated local copy; the mapped header is never dereferenced again.
2. Teardown now falls back to any surviving same-process window on the remembered taskbar UI thread, excluding the failed `Shell_TrayWnd` target. `RunFromWindowThread` can therefore still execute `RemoveFromCurrentTaskbar` through its `WH_CALLWNDPROC` hook while Explorer is dismantling the tray window. The `Wh_ModUninit` retry remains as a second attempt.

The UI differentiation from Taskbar Clock Customization is now led with explicitly in the PR description as suggested. The remaining items are optional polish or future scope and aren't part of this focused safety fix.

Revalidated with the source validator, warning-free x64 build, ARM64 compatibility build and the live D3DKMT smoke test (50 °C without HWiNFO).

starychenko:
Final handoff note: the latest available AI review was for 5e0a1601. Its two non-optional findings are fixed in 44034fff: the HWiNFO shared-memory header is snapshotted before range validation, and taskbar teardown now has a same-thread fallback window when `Shell_TrayWnd` is unavailable.

A follow-up `/ai-review` was requested for 44034fff, but the bot declined it only because the pull request has reached the 3-reviews-per-24-hours limit; it did not report a code finding. The full GitHub Actions matrix is green for the current HEAD: changed-files validation plus Windhawk 1.6.1, 1.7.3 and 2.0.0-alpha.2. Local source validation, x64/ARM64 compatibility builds and the live D3DKMT smoke test also pass.

starychenko:
The latest review covers the current HEAD, ecfba1966be1b74e792726455b19558c04f0eb5e, and reports no blocking issues. No code changes have been pushed since that review.

For the remaining optional findings, I am leaving these out of the initial submission:

- Worker restart after WAIT_FAILED: deferred as additional recovery hardening; this limitation is not being claimed as fixed.
- Adapter identity probe backoff: deferred performance tuning; the current bounded-rate probe remains.
- First PDH sample after query recreation: leaving this transient sampling limitation for follow-up.
- The windowsThermalZones alias: retaining the existing accepted setting value rather than changing configuration handling now.
- Undocumented setting clamps: documentation polish deferred; the current clamps remain.
- XAML brush/text churn and settings copies: optimization deferred, without claiming the current implementation is allocation-free.
- The redundant window check: minor cleanup deferred.
- The duplicate GiB constant: minor cleanup deferred.
- Lifecycle log markers: diagnostic polish deferred.
- Nonuniform history sampling: leaving timer/history redesign outside this initial submission.
- Early samples on settings changes: leaving this transient graph limitation for follow-up alongside history sampling.
- Zero GPU usage with no matching engine instance: retaining idle-GPU behavior; a parked adapter need not have engine instances, so other adapters having instances alone is not sufficient evidence of failure.
- Integrated-GPU shared-memory percentages: acknowledging that this denominator differs from dedicated VRAM; further presentation/documentation work is deferred.
- Wildcard GPU-counter cost: keeping the existing enumeration approach; per-instance counter management is deferred.
- CPU utilization versus Task Manager: retaining GetSystemTimes-based utilization rather than changing the metric's meaning in this submission.
- HWiNFO mutex hold time: shortening the critical section is deferred optimization.
- Wide-font overflow: layout/ellipsis improvements are deferred; arbitrary-font fit is not being claimed.
- Secondary taskbars: keeping the documented primary-taskbar-only scope.

The two PR mod validation runs are also awaiting maintainer approval. Please approve them when reviewing the submission.

/ready-for-reviewer

m417z:
Looks great, thanks. A minor note for future updates: in `TaskbarFrame_Constructor_Hook`, you already have the taskbar element, you can avoid calling `GetTaskbarXamlRoot` for better stability. This can be especially helpful if that function stops working in the future due to a Windows update.

PR #5177, opened 2026-08-22
Mix Files and Folders When Sorting in Windows Explorer

Extremenis:
## Changelog

* Changelog item 1...
* Changelog item 2...

## Mod authorship

This mod was created by:

- - [x] Claude
- - [ ] ChatGPT
- - [ ] Gemini


Extremenis (commented):
requested corrections made

Extremenis:
explorer-details-better-file-sizes has a totally different purpose, so I think this is not a problem

Extremenis:
Re: the overlap with `explorer-details-better-file-sizes` — I'd like to keep this as a separate mod, and I've addressed the conflict directly rather than by consolidating. Reasons:

**Different scope for the user.** `explorer-details-better-file-sizes` is a large mod whose purpose is changing how sizes are displayed — folder size calculation, size unit formatting, column rendering. Its `sortSizesMixFolders` option is one setting inside that. Someone who just wants folders interleaved when sorting by Name, Type or Date currently has to install a mod that also changes size display throughout Explorer, which is a behavior change they didn't ask for. This mod does one thing and touches nothing else.

**It isn't a widening of the existing setting.** Generalizing `sortSizesMixFolders` to all columns isn't a matter of relaxing a condition. That mod's path is built around `PKEY_Size` specifically, so it doesn't need the things an any-column comparison does: masking `SHCIDS_*` flags off the column argument, substituting `PKEY_FileName` for `PKEY_ItemNameDisplay` so folder-vs-file ordering stays transitive with the shell's file-vs-file ordering, honoring the `NoStrCmpLogical` policy, and never returning 0 for two distinct items. Folding this in would mean rewriting that mod's comparison path, in someone else's mod, for a feature orthogonal to its purpose.

**Ownership.** It's m417z's mod. Upstreaming means asking him to take on and maintain a feature outside that mod's stated scope.

**On the conflict itself.** The README now leads with a prominent warning that the two mods hook the same function, that `sortSizesMixFolders` defaults to on, and that enabling both produces load-order-dependent ordering — so users are told plainly not to run both, and which to pick for size-based mixing.

Happy to revisit if you'd rather see this upstreamed.

Kryxzort:
awesome mod, i didnt realize how much i needed this

PR #5183, opened 2026-08-23
Wobbly Windows

lalimatyus:
## Changelog

* Changelog item 1...
* Changelog item 2...

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [x] ChatGPT
- - [ ] Gemini


TheGamer1445891:
@lalimatyus you should check this mod
[anim-wobbly-by-shaoib-hassan.wh(1).cpp](https://github.com/user-attachments/files/31507947/anim-wobbly-by-shaoib-hassan.wh.1.cpp)
its discontinued, but its the same as how your mod may work

Potassiumuncher:
@TheGamer1445891 no his mod is diffrent to shaoib hassan as his uses dwm.exe while shaoib uses print screen which is way better, his the reason why I gave up on mine its so good.

Potassiumuncher:
my boy's back 🥹

haven80:
With a 360 Hz monitor, I’m experiencing some strange visual artifacts: text becomes blurry, and windows sometimes leave a sort of “trail” or ghosting effect when moved.

Apart from this issue, the mod works extremely well.

I tried modifying the code to make it run properly at 360 Hz, but without success. The problem is not always noticeable, but sometimes, while moving a window, the text and images seem unable to keep up with it, resulting in visible duplication or smearing.

My spec are 9800x3d and 5070ti

lalimatyus:
@haven80 thanks for the feedback! I want a more detailed description since i cant do really much with this info. I want: 

**Windows version (You can check it by pressing WIN+R and typing: `winver`)
It should look something like this: "Version 23H2 (Build number 22631.6199)"**

**what windows do does this happen to (for example: notepad, file explorer etc.) or does it happen to all windows but only sometimes?**

**monitor resolution (probably irrevelant but can cause sometimes issues)**

### Also the most important: do you have g-sync/freesync or anything similar enabled; change the monitors refresh rate to 240hz or lower temporary to see if the tearing/ghosting disappears completely.

Thanks for the feedback, i'll try my best to fix it!

haven80:
> @haven80 thanks for the feedback! I want a more detailed description since i cant do really much with this info. I want:
> 
> **Windows version (You can check it by pressing WIN+R and typing: `winver`) It should look something like this: "Version 23H2 (Build number 22631.6199)"**
> 
> **what windows do does this happen to (for example: notepad, file explorer etc.) or does it happen to all windows but only sometimes?**
> 
> **monitor resolution (probably irrevelant but can cause sometimes issues)**
> ### Also the most important: do you have g-sync/freesync or anything similar enabled; change the monitors refresh rate to 240hz or lower temporary to see if the tearing/ghosting disappears completely.
> 
> Thanks for the feedback, i'll try my best to fix it!

Effect description:

When dragging any window, the window content (text, images, icons, borders — essentially everything drawn inside the window) doesn't move in perfect lockstep with the window frame itself. Instead, it leaves behind a faint duplicate/ghost trail — text in particular appears to "split" or double up rather than translating cleanly, as if the content is lagging a frame or two behind the window's actual position. Window borders show the same lag/tearing-like artifact.

The effect is clearly visible at 360Hz, and still noticeable at 240Hz. No G-Sync or V-Sync is enabled on either display. Setup is a dual-monitor configuration, both at 3440x1440, on Windows 11 25H2 (build 26200.9445).

However you have done a really good job. Its just like a dream comes true!
Wobbly windows in real time on windows!!!!!

lalimatyus:
@haven80 it will be fixed soon :) Glad you like my mod!

lalimatyus:
@haven80 Thanks again for reporting this! I’ve tried several approaches, but unfortunately I haven’t confirmed the root cause or found a reliable fix yet. I can also barely see it ony my own 144hz monitor, it's not visible on 60hz.

I’ve added the issue to the readmes’s `Known Issues` section. For now, I’m moving forward with a human review of the stable build. I plan to revisit this issue later, but I can’t give an ETA for a fix.

Your detailed feedback has been really helpful, thank you!

m417z:
Thanks for the submission, it's an impressive mod. From a quick test, it didn't work so well for me, it works when enabled or when I launch the browser, but stops working when I switch to another window:

https://github.com/user-attachments/assets/9c6c91fb-2fb6-4a71-b054-6a701613e1bb

Also:
* Use `// @architecture    amd64` for the mod. The CI asks for a manual review for it, that's OK, often using it is a mistake (most mods don't require ARM64-specific code), but in this case it's justified.
* Remove mangled symbols such as `L"?IsGhostWindow@CWindowData@@QEBA_NPEAPEAUHWND__@@@Z"`.

PR #5189, opened 2026-08-23
Add taskbar clock to left

pleromyst:
Adds a new Windows 11 mod that moves the native clock and notification-center button to the left side of the taskbar.

The mod is designed to:

- Keep the centered Start button and application icons in their original position.
- Avoid leaving an empty clock area on the right side.
- Remain compatible with clock customization mods that increase the clock width.
- Safely restore the original taskbar layout when the mod is disabled or unloaded.
- Protect deferred operations and event handlers during unloading.

Credits:

Taskbar-related implementation techniques were informed by existing Windhawk mods by Michael Maltsev (@m417z), particularly Taskbar Clock Customization and Multirow Taskbar. Development and code review were assisted by OpenAI Codex.

## Changelog

Not applicable. This pull request introduces a new mod.

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini
- - [x] Another AI (please specify): OpenAI Codex


pleromyst:
ai-review

pleromyst:
I reviewed the remaining AI findings. I restored this revision because a later attempt to address them introduced a confirmed regression involving `PostMessage` and `WH_CALLWNDPROC`.

This restored revision passed compilation and runtime testing, including enable/disable, Explorer restart, taskbar alignment changes, and compatibility testing with clock customization mods.

The remaining threading, exception-boundary, and teardown concerns are acknowledged. I would prefer human reviewer guidance before making further speculative lifecycle changes that could introduce regressions.

m417z:
From the code:
```cpp
    // On current Windows builds, SystemTray_Main can be owned by a different
    // thread than Shell_TrayWnd.
```
I don't think it's ever correct. If it's a mistake, update the comment and adjust the code. It will likely also resolve the first review note.

Address this:

- **The `MoveClockToLeft` setting duplicates the mod's own enable toggle.** It's the mod's only setting and it gates the mod's only behaviour, so "off" is indistinguishable from disabling the mod. Unless you plan to add more settings around it, consider dropping it (and the `wasEnabled`/`isEnabled` transition handling in `Wh_ModSettingsChanged` that exists solely to serve it).

Also, I tested the mod on the latest Windows 11 build (26200.9278) and it didn't work. Please make sure it works on that version, which is going to be released for everyone in 10 days.

pleromyst:
Thanks for the review. I removed the redundant MoveClockToLeft setting and its transition handling, corrected the thread-ownership assumption, and updated the clock hook/module discovery for current Windows 11 builds. I also added dynamic Widgets offset tracking, deterministic alignment-change handling, and a dispatcher-based cleanup fallback. The updated source compiles successfully for x64 and ARM64.

Regarding the remaining AI findings, the left host is intentionally an overlay so centered Start and application buttons are never shifted. Reserving layout space or restoring the clock when the taskbar fills would conflict with that core behavior, so I kept the existing design and accept full-taskbar overlap as a limitation.

I also kept the symbol-hook implementation minimal. The current OnApplyTemplate and get_Alignment symbols were verified against current Windows symbols and existing repository mods. I avoided adding parallel alternative hooks without a confirmed compatibility need, but I can add a specific fallback if you prefer one.

m417z:
This still seems to be the case:

> Also, I tested the mod on the latest Windows 11 build (26200.9278) and it didn't work. Please make sure it works on that version, which is going to be released for everyone in 10 days.

Was it tested with this Windows 11 build?

pleromyst:
- Restore the clock to its native position when taskbar apps are left-aligned.
- Move the clock back to the left when apps are centered, without re-enabling the mod.
- Preserve the existing layout logic.
- Reduce debug log noise to clock movement, restoration, and errors.
- Stop alignment monitoring before unloading.

Tested on Windows 11 build 26200.9278. All local tests passed with no issues observed.

pleromyst:
Addressed review item 1:

- Restore each clock using its own record, avoiding empty-vector access and accidental removal of another clock.
- Detach the clock from its temporary host and restore it to the original parent when available before removing the host.
- Preserve the existing layout, alignment monitoring, and hook behavior.

Isolated XAML regression tests passed, including expired records, multiple clocks, interrupted moves, and restoration errors. Both x64 and ARM64 builds compiled successfully.

This update addresses item 1 only. The dispatcher fallback discussed in item 2 remains unchanged.

pleromyst:
Thanks. I agree that cleanup failure must not be treated as successful unloading.

I propose to preserve the existing clock/layout behavior, retain failed event-revocation tokens and cleanup records for retry, and explicitly verify that cleanup actually completed.

However, the suggested retry loop does not by itself bound the synchronous SendMessage call or an already-running registry callback. If all retries fail, live XAML delegates may still remain.

Could you point me to a recommended Windhawk pattern that handles both retry exhaustion and callback lifetime when the owning UI thread cannot service the cleanup request?

I would like to avoid permanent DLL pinning, requiring an Explorer restart, or a substantial callback-lifetime redesign unless that is the intended solution. I have not applied the retry snippet as a complete fix because those failure cases remain unresolved.

m417z:
If cleanup fails, there isn't really a good solution. The review gave an example of what Taskbar Clock Customization does. Either waiting forever or just unloading and crashing are fine by me in this case, which normally should never happen. Pinning the module is better avoided because it violates the Windhawk assumption that the mod is unloaded, and it's not really a fix, e.g. if Windhawk is closed, the mod will point to an engine module that's unloaded.

> Tested on Windows 11 build 26200.9278. All local tests passed with no issues observed.

What local tests? The mod doesn't do anything for me on this version.

pleromyst:
You're right — “all local tests passed” was too vague.

The initial functional test on build 26200.9278 was isolated: this was the only Windhawk taskbar mod loaded in Explorer.

My test environment was:
- Windows 11 Home China, 25H2 build 26200.9278 (zh-CN)
- Atlas Playbook v0.5.0
- Windhawk 1.7.3
- One 2560x1600 display at 200% scaling
- Taskbar alignment set to Center
- Widgets disabled
- Taskbar.View.dll 2607.28001.200.0
- SystemTray.dll 2607.28000.0.0

I tested:
- Enabling the mod with centered taskbar apps: the native clock and notification button moved to the left.
- Disabling and re-enabling it: the clock returned to the native right position and then moved left again.
- Switching Center → Left → Center several times: the clock returned to the native right position for Left alignment and moved left again for Center alignment without reloading the mod.
- Restarting Explorer and rebooting Windows.
- Repeating enable/disable operations after the reboot.

No Explorer crash or hang was observed.

I should also clarify that Atlas Playbook makes this a non-stock Windows environment. It changes some taskbar settings, but I currently have no evidence that Atlas is the reason the mod works here.

Could you confirm whether your taskbar alignment was Center during the test? With Left alignment, the clock intentionally remains in its native right-side position to avoid overlapping the app icons.

If it was centered, could you please share:
- the PR commit you tested;
- the Windhawk debug log from enabling the mod;
- your Taskbar.View.dll and SystemTray.dll file versions?

Lines such as “Failed to hook the Windows 11 taskbar layout/system tray”, “NotificationCenterButton not found”, “Unexpected clock parent”, or “Taskbar root Grid not found” would help distinguish a symbol-resolution problem from a different XAML layout.

m417z:
> Could you confirm whether your taskbar alignment was Center during the test? With Left alignment, the clock intentionally remains in its native right-side position to avoid overlapping the app icons.

It was centered.

> If it was centered, could you please share:
>
> - the PR commit you tested;

3a70a56.

> - the Windhawk debug log from enabling the mod;

No logs at all.

**Edit:** Eventually I got this log:
`05:13:17.108 13256 explorer.exe  [WH] [local@taskbar-clock-to-left] [1041:operator()]: Clock layout wait expired; relocation will retry after the next template callback`

> - your Taskbar.View.dll and SystemTray.dll file versions?

2607.28001.200.0 and 2607.28000.0.0, Win11 build 10.0.26200.9278.

pleromyst:
Thanks, this log is very helpful. It shows that the template hook ran successfully, but ClockLayoutIsReady() remained false until the 10-second timeout.

My Atlas installation has Widgets disabled. Could you confirm whether Widgets are enabled on your system? If they are, could you temporarily disable Widgets and reload the mod to see whether the clock moves?

The current readiness check waits for a visible AugmentedEntryPointButton to have a non-zero ActualWidth. That condition may be the environment-dependent part.

m417z:
Widgets aren't enabled in my environment. You can post a mod version with more logs for me to test.

m417z:
Logs for the updated version:
```
07:51:00.353 13256 explorer.exe  [WH] [local@taskbar-clock-to-left] [980:ClockLayoutIsReady]: Clock layout not ready: unexpected clock parent; class=Windows.UI.Xaml.Controls.StackPanel, name=SystemTrayFrameGrid
07:51:00.353 13256 explorer.exe  [WH] [local@taskbar-clock-to-left] [1078:operator()]: Clock layout wait expired; relocation will retry after the next template callback
```

pleromyst:
The diagnostic log identified the issue. SystemTrayFrameGrid is a StackPanel in your taskbar template, while the code incorrectly required it to be a Grid.

I updated the original-parent handling to use the common Panel base class. This supports both the Grid template on my system and the StackPanel template on yours without changing the relocation logic.

The diagnostic logging is still included for this verification build. Please retest the updated commit.

m417z:
It works now. That's the change that was done in recent Win11 updates to support new features such as the vertical taskbar. Perhaps the feature is disabled for you.

pleromyst:
The StackPanel compatibility issue has now been reproduced and verified by you on Windows 11 build 26200.9278.

Regarding the latest AI review:

- The full-taskbar collision is a known trade-off of keeping the centered app buttons in their native position. Left-aligned overlap is already prevented by restoring the clock to its native right-side position.
- The Widgets replacement case wasn't exercised because Widgets were disabled in both tested environments. I prefer not to add another persistent layout watcher without a reproduced failure, but I can address it if you considers it required.
- The cleanup-failure case was already discussed above. I'm following your guidance and won't pin the module.
- The required template hook resolves and works with both taskbar template variants tested on build 26200.9278. I prefer to keep the verified trigger unless an affected build is identified.

The current version passes the requested compatibility test.

PR #5190, opened 2026-08-23
Add MultiWM

Meteoni:
## Changelog

# MultiWM

MultiWM is a continuation of the original [Tiling Helper](https://windhawk.net/mods/tiling-helper) by [U2X1](https://github.com/u2x1). Explicit approval from the original author has been obtained, as they no longer intend on maintaining the original Tiling Helper. Overlap concerns should be treated as settled.  

A lightweight, low-cortisol window manager for Windows 11 with true per-virtual-desktop layouts - including floating - and simple, predictable controls. 

GIF

*Windows 10 support is currently untested.*

## Features
- **First-class floating** - use Floating as a whole-workspace layout 
or float individual windows while retaining useful geometry.
- **Automatic or Manual management** - let MultiWM continuously discover 
and migrate windows, or keep explicitly tiled groups under your control.
- **True virtual-desktop layouts** - uses Windows' existing virtual desktops 
instead of implementing a separate desktop system.
- **Event-driven** - changes are signaled through native events rather than 
continuous polling. 

## Layouts
- Floating
- Master + Stack (vertical or horizontal)
- Binary Space Partitioning (BSP)
- Columns / Rows
- Monocle

## Getting Started
Move a tiled window to **float it** or **swap it with the window underneath**. 

Hotkeys are available for tiling, changing layouts, moving windows through the logical 
order, floating a window, and switching between Automatic and Manual management.

A small tray indicator shows the active workspace layout and provides quick access to layouts and management mode.
## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/multiwm.wh.cpp
+++ b/mods/multiwm.wh.cpp
@@ -143,7 +143,7 @@ A small tray indicator shows the active workspace layout and provides quick acce
 - hotkeys:
   - TilingModifier: alt
     $name: Modifier
-    $description: "Modifier used by all window-manager hotkeys. \nNote: Alt+<Key> can shadow some system hotkeys - if you don't know what they are, there's likely no need to worry. "
+    $description: "Modifier used by all window-manager hotkeys. \nNote: Alt+letter shortcuts can override application menu mnemonics. Choose Alt+Shift or Ctrl+Alt if this is disruptive. "
     $options:
       - alt: Alt
       - ctrl: Ctrl
--- a/mods/multiwm.wh.cpp
+++ b/mods/multiwm.wh.cpp
@@ -1,6 +1,6 @@
 // ==WindhawkMod==
-// @id              multiwme
-// @name            MultiWME
+// @id              multiwm
+// @name            MultiWM
 // @description     Lightweight, low-cortisol window manager with true per-virtual-desktop layouts. 
 // @version         1.13.19
 // @author          meteoni
--- a/mods/multiwm.wh.cpp
+++ b/mods/multiwm.wh.cpp
@@ -1008,6 +1008,19 @@ using Model::Workspace;
 using Model::WorkspaceGeometryItem;
 using Model::WorkspaceRepository;
 
+struct MonitorIdentityCacheEntry {
+  HMONITOR monitor = nullptr;
+  Model::MonitorId id;
+};
+
+// HMONITOR <-> stable monitor identity is valid for one display-topology
+// generation. DPI and work-area observations deliberately aren't cached here.
+static std::vector<MonitorIdentityCacheEntry> g_monitorIdentityCache;
+
+static void ClearMonitorIdentityCache() {
+  g_monitorIdentityCache.clear();
+}
+
 struct MoveSizeTracker {
   std::unordered_map<HWND, RECT> startRects;
   std::unordered_map<HWND, RECT> endRects;
@@ -1438,6 +1451,16 @@ static void NormalizeMonitorIdentityToken(std::wstring* value) {
 bool Model::MonitorId::FromHMonitor(HMONITOR monitor, MonitorId* out) {
   if (!out || !IsLiveMonitorHandle(monitor)) return false;
 
+  auto cached = std::find_if(
+      g_monitorIdentityCache.begin(), g_monitorIdentityCache.end(),
+      [&](const MonitorIdentityCacheEntry& entry) {
+        return entry.monitor == monitor;
+      });
+  if (cached != g_monitorIdentityCache.end()) {
+    *out = cached->id;
+    return true;
+  }
+
   MONITORINFOEXW monitorInfo{};
   monitorInfo.cbSize = sizeof(monitorInfo);
   if (!GetMonitorInfoW(
@@ -1468,11 +1491,18 @@ bool Model::MonitorId::FromHMonitor(HMONITOR monitor, MonitorId* out) {
 
   NormalizeMonitorIdentityToken(&identity);
   out->deviceId = std::move(identity);
+  g_monitorIdentityCache.push_back({monitor, *out});
   return true;
 }
 
 HMONITOR Model::MonitorId::Resolve() const {
   if (deviceId.empty()) return nullptr;
+
+  for (const auto& entry : g_monitorIdentityCache) {
+    if (entry.id == *this && IsLiveMonitorHandle(entry.monitor)) {
+      return entry.monitor;
+    }
+  }
 
   struct ResolveContext {
     const MonitorId* target = nullptr;
@@ -1505,7 +1535,7 @@ bool GetMonitorWorkArea(HMONITOR monitor, RECT* outRect) {
 }
 
 static UINT GetMonitorEffectiveDpi(HMONITOR monitor) {
-  if (!monitor) return 96;
+  if (!monitor || !IsLiveMonitorHandle(monitor)) return 96;
 
   using GetDpiForMonitorFn = HRESULT(WINAPI*)(HMONITOR, int, UINT*, UINT*);
   using GetScaleFactorForMonitorFn = HRESULT(WINAPI*)(HMONITOR, int*);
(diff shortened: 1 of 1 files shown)


m417z:
Congratulations on being the first to submit a mod with an id shorter than 8 characters. Let me think about it, maybe we can lift this limitation.

Meteoni:
> Congratulations on being the first to submit a mod with an id shorter than 8 characters. Let me think about it, maybe we can lift this limitation.

I appreciate it.

m417z:
Updated limit to 6.

Meteoni:
1. The default hotkeys are based on the ones I use personally. Most users rarely use the default Alt+Modifier bindings that came with their system. For the frequency at which the mod-defined hotkeys are expected to be hit, Alt seems to be the more ergonomic choice for most users. Alt+T for tiling is also easier to remember than other binding - Alt+D for example. I consider that a fair trade-off. If a user does end up finding the collision inconvenient, it's just a one-line configuration change anyway. 
The next part of the finding is valid. I have added KEY_NOREPEAT to hotkeys other than window promotion and demotion, since those do benefit from rapid repeat events. 

2. Valid. Dropped the fallback path.

3. The dedicated tool process is windhawk.exe, whose manifest is already PerMonitorV2, and newly created threads inherit that process DPI context, so the WM thread is not currently DPI-unaware/system-aware. That said, explicitly setting PMv2 at the start of the WM thread makes sense. 

4. That was inherited from the original Tiling Helper, which had the exact same semantics. That said, making it so that layout commands create an empty space in manual mode is the more internally consistent model. The last commit made it exact. 

5. Addressed. 

6. The diagnostics system has turned out quite useful during development, so I won't scrap it just yet. Dumping a plain .txt file also offers better formatting and readability over plain logging. That said, I have changed it to be disabled by default. 

7. Noted. The README change will happen after this merges.

Meteoni:
1. Caching monitor identity is the right call. DPI on the other hand is a mutable identity and should be re-queried from fresh. I'll keep the cache scoped to monitors instead of introducing DPI.

u2x1:
cool!

Meteoni:
1. Declaring x86-64 breaks backward compatibility and prevents the tool mod from being loaded by any 1.7x version of windhawk.exe, which was 32-bit.

2. On a multi-monitor setup, the user might very well be using different layout configuration across multiple monitors. The tray icon needs to reflect which monitor + desktop combination is currently in use, instead of being limited to what is currently on the primary display, which would be inaccurate. The suggested fix is essentially functionality loss.

m417z:
> Declaring x86-64 breaks backward compatibility and prevents the tool mod from being loaded by any 1.7x version of windhawk.exe, which was 32-bit.

Yes, but what the review says is that the code is incorrect on 32-bit:

> the function emits `ret 0` (or `ret 4`) while the caller pushed 8-16 bytes, so every one of those callbacks returns with the COM stub's stack unbalanced. That's stack corruption on the STA thread, i.e. a crash of the tool process.

Have you tested it on 1.7.x?

Probably the simplest solution is to have a `VirtualDesktopNotification_NoOp` function per prototype.

m417z:
Regarding settings, currently you're using prefixes such as "[Workspace]", perhaps it'd make more sense to have the settings grouped. A similar comment for another mod: https://github.com/ramensoftware/windhawk-mods/pull/5003#issuecomment-5270266506.

Meteoni:
> > Declaring x86-64 breaks backward compatibility and prevents the tool mod from being loaded by any 1.7x version of windhawk.exe, which was 32-bit.
> 
> Yes, but what the review says is that the code is incorrect on 32-bit:
> 
> > the function emits `ret 0` (or `ret 4`) while the caller pushed 8-16 bytes, so every one of those callbacks returns with the COM stub's stack unbalanced. That's stack corruption on the STA thread, i.e. a crash of the tool process.
> 
> Have you tested it on 1.7.x?
> 
> Probably the simplest solution is to have a `VirtualDesktopNotification_NoOp` function per prototype.

I have on two separate machines and it did work fine without crashes. 
That said, working probably doesn't mean it's entirely safe. Adding a NoOp function per prototype isn't that easy of a change and I didn't find complete, verified reverse-engineered references online either. I'll use object_cloaked events as a safe fallback signal for x86 instead.

m417z:
How about dumping the addresses and at least getting just the used stack size from the return command in assembly? Could that work?

Meteoni:
- The affected ABIs are now disabled on x86 and desktop switches are inferred for those builds rather than natively supported. The README is now updated to reflect this new behavior. 
- Settings have been rearranged and expanded to avoid migration issues across versions.

Meteoni:
> How about dumping the addresses and at least getting just the used stack size from the return command in assembly? Could that work?

Probably not. I made it fall back to OBJECT_CLOAKED signals instead, which should be reasonably accurate in most cases. Might not be worth it to change the ABIs too much for a version that's older than current.

Meteoni:
1. That's simplifying the model way too much. Two identical results 75-150ms apart at app launch say very little during that period. An app can temporarily spawn a non-conforming window and replace it later (e.g. F3d), or only temporarily pretend to conform to its assigned geometry while restoring its own geometry later (Firefox), etc. The 3000ms observation period is designed to solve this issue, otherwise:
- F3d will be floated because the initial window is non-resizable, but after the window has finished loading we have ended the lease early. 
- Firefox will be treated as resolved during startup, but as soon as it changes geometry later (which it always does) we won't be able to catch it. 
Therefore, the suggested fix isn't really preserving the original intent of the system. Plus, the user can always adjust the lease themself to make it fit their own liking.

Meteoni:
We might have discussed this before, and it probably isn't a big concern either - the user should already be pretty aware of the hotkeys if they do try to use this mod themselves. 

It shouldn't come off as a surprise to them if they do choose to use this mod themselves, where the first thing they'd do is to familiarize themselves with the hotkeys. If some bindings do cause a problem for them, they can easily adjust those to their liking later.

m417z:
Regarding:

> - Windhawk 2.0 (or higher) on a 64-bit system is recommended for the best experience; for x86 builds, virtual desktop switch events are inferred at best effort due to the lack of documented ABIs. 

It will take time until 1.7.3 is phased out. For the time being, given that making it work for a 32-bit process requires more effort, perhaps it's better to just have a tool mod targeting explorer.exe. Just changing the target process from `windhawk.exe` to `explorer.exe` should be enough. Let me know what you think.

Meteoni:
> Regarding:
> 
> > - Windhawk 2.0 (or higher) on a 64-bit system is recommended for the best experience; for x86 builds, virtual desktop switch events are inferred at best effort due to the lack of documented ABIs. 
> 
> It will take time until 1.7.3 is phased out. For the time being, given that making it work for a 32-bit process requires more effort, perhaps it's better to just have a tool mod targeting explorer.exe. Just changing the target process from `windhawk.exe` to `explorer.exe` should be enough. Let me know what you think.

That might work. Thanks.

Meteoni:
Noted. I wouldn't consider those release-blocking issues. 

Leaving them unaddressed for now, and I'll assess those separately - I can do a follow up patch soon afterwards if they are genuine issues.

m417z:
That's correct, and generally it'd probably be best to limit the mod to x86-64 now:

- **`SysWOW64\explorer.exe` can become the host.** `@include explorer.exe` matches by file name, and `Wh_ModAfterInit` uses `GetModuleFileName(nullptr)`, so a 32-bit `SysWOW64\explorer.exe` launcher would spawn a **32-bit** tool process - the exact host the change to `explorer.exe` was meant to avoid, with the notification sink compiled out. An `IsWow64Process` bail-out in the launcher branch closes it in one line.

Meteoni:
> That's correct, and generally it'd probably be best to limit the mod to x86-64 now:
> 
> - **`SysWOW64\explorer.exe` can become the host.** `@include explorer.exe` matches by file name, and `Wh_ModAfterInit` uses `GetModuleFileName(nullptr)`, so a 32-bit `SysWOW64\explorer.exe` launcher would spawn a **32-bit** tool process - the exact host the change to `explorer.exe` was meant to avoid, with the notification sink compiled out. An `IsWow64Process` bail-out in the launcher branch closes it in one line.

Agreed.

PR #5191, opened 2026-08-23
Make the "Redirect Settings to Control Panel" mod more stable

babamohammed2022:
This is a maintenance update that makes the mod more stable and safer to use with try-catch, RAII and by trying to reduce potential conflicts with companion mods.

## Changelog

- Fixed a bug where some hooks were registered but never actually turned on, so a few features could silently stop working after Explorer restarted or after changing a setting
- Cleaned up some unnecessary error-handling code that wasn't actually protecting against anything
- Fixed the "Display" settings redirect being removed by accident, and added a toggle so this mod works together with the Classic Display Control Panel Restorer mod instead of conflicting with it
- Updated the README to be more accurate about what the mod actually protects against

## Mod authorship

This mod was created by:

- - [X] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/settings-to-control-panel.wh.cpp
+++ b/mods/settings-to-control-panel.wh.cpp
@@ -12,8 +12,11 @@
 // ==WindhawkModReadme==
 /*
 # Redirect Settings → Control Panel
+
 ## Screenshot
+
 ![Image](https://raw.githubusercontent.com/babamohammed2022/babamohammed2022/main/Senza%20nome.png)
+
 ---
 ## About
 This mod intercepts modern `ms-settings:` links (the ones that open the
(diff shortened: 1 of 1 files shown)


babamohammed2022:
Thanks for the review. After testing on Windows 10 21H2, I'm satisfied with the current state of the mod. 

While I understand the performance concerns about Wh_ApplyHookOperations(), 
in my case the overhead is minimal and the mod works reliably. 
The edge cases identified (double registration, RedirectDisplayPages 
with other mods) are, in my opinion, extremely rare for the average user.
For the RedirectDisplayPages problem, the "regression" is in my opinion, not a real concern as, from what I know, there aren't any real commands that call ms-settings:display-advanced, etc so the problem is described as major when it could occur in very specific scenarios.
Unless the maintainer decides that these problems are really worth addressing, I'll keep the current implementation as-is since:
- It works correctly on Windows 10 21H2 (my primary test environment)
- The watchdog thread and late-hook registration haven't caused 
  any crashes during my testing

/ready-for-reviewer

PR #5192, opened 2026-08-24
Add Desktop Icon Selection Style mod

RiteshKhandekar:
Restyles the highlight behind selected desktop icons.

The stock highlight is exactly as wide as an icon cell, so selections in neighbouring columns touch with no gap and read as one block. This mod replaces it with a configurable plate: width as a percentage of the cell, fill colour and opacity, corner radius, an optional glow, and an optional solid, dashed, dotted or dash-dot border.

**Before**

Two selected desktop icons with the stock highlight, a pair of rectangular slabs meeting with no gap between them

**After**

The same two icons with the mod enabled, each highlight narrower than its cell with rounded corners, a dashed white border and a soft accent glow

Implementation notes, in case they save review time:

- `DrawThemeBackground` is hooked and the list item part is replaced for the selected, selected-not-focus and hot-selected states. Plain hover is passed through untouched unless the user opts in.
- Draws are attributed to the desktop list view with a thread-local paint depth counter set from a subclass, not by `WindowFromDC`. Double buffering means the theme call always receives a memory DC, and `WindowFromDC` returns null for every one of them.
- The plate is always clamped to the item rect, so drawing stays inside the region Windows invalidated and the selection cannot leave trails when it moves. The glow is deliberately not clamped, because GDI+ honours the device context clip, which is that same region.
- The plate follows the item height rather than fixing it, since Windows grows a selected item so a long file name can wrap. That keeps the label inside the plate.
- The glow is a stack of translucent rings and is labelled a fake in both the readme and the setting name. A real blur is not possible here: the buffer under the plate is fully transparent at draw time, so there is nothing to sample.

Tested on Windows 11 26200 at 200% scaling.

## Changelog

* Not applicable, this is a new mod.

## Mod authorship

This mod was created by:

- - [x] Claude
- - [ ] ChatGPT
- - [ ] Gemini

To be specific about the split: the code was written by Claude. I set the requirements, ran it, tested it on my setup, and made the design calls, including reverting an effect that was built and did not look right. The behaviour of the hooks and the drawing was established by first writing a throwaway diagnostic mod that logged what Windows actually does when it draws the selection, rather than by guessing, and the notes above come from those measurements.

### DIFF
--- a/mods/desktop-icon-selection-style.wh.cpp
+++ b/mods/desktop-icon-selection-style.wh.cpp
@@ -65,15 +65,18 @@ expect acrylic. It is also limited to roughly 10 pixels beyond the icon, which i
 all Windows repaints around it; a larger glow gets cut off rather than smeared.
 
 **Colours** take `accent` to follow your system accent colour, or a hex value
-like `#3399FF`. The accent is read through the immersive colour API, so it is
+like `#3399FF` (the `#` is optional). Anything else is ignored, and the setting
+keeps its default - the mod log says which value it rejected. The accent is read through the immersive colour API, so it is
 the colour you actually picked in Settings rather than the blended colorization
 value. Changing it is picked up as plates are next repainted, so selecting a
 different icon shows the new colour; nothing is forced to redraw on a timer.
 
-**Mixed-DPI setups** are fine. Windows keeps desktop icons on the primary
-monitor and sizes the icon area to it, so the pixel settings are scaled by the
-DPI of the monitor the plate is actually drawn on. Tested with a second display
-at a different resolution and scale factor.
+**Mixed-DPI setups** are fine, though not because anything is measured
+per-monitor. The pixel settings are scaled by a single DPI, read from the
+desktop's own window - but the icon rectangles the plate is drawn against come
+from that same window, so the plate and the icon are always scaled by the same
+number and stay in proportion to each other. Tested with a second display at a
+different resolution and scale factor.
 
 ## Credit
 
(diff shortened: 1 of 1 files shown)


RiteshKhandekar:
Reopening to re-run the validation check: the PR description now restores the template's Mod authorship section, and the validation workflow does not trigger on description edits.

RiteshKhandekar:
All required findings from the three rounds are addressed. Notes on the ones I handled differently or left alone:

**The theme class check is inverted from the suggestion.** Rather than allowing only classes ending in `ListView`, it rejects those ending in `ScrollBar` or `Edit` and allows everything else. `GetThemeClass` is an undocumented ordinal and I have not verified what it returns on any machine, and allow-only would silently disable the entire mod if it ever returned something unexpected - a much worse failure than the artifact it prevents. `WM_NCPAINT` also left the paint-depth scope, which closes the scroll bar case on its own, so the class check is defence in depth rather than the sole fix.

**Mixed-DPI turned out not to apply.** Windows keeps desktop icons on the primary monitor and sizes the icon area to it, so the pixel settings are scaled by the DPI of the monitor the plate is actually drawn on. Tested with a second display at a different resolution and scale factor, and the readme caveat I had added was removed rather than softened.

**`g_settings` is still unsynchronised**, deliberately. The clean fix is a reload on every settings change, and this is a mod tuned by dragging sliders back and forth; live updates are worth more than removing one possibly-torn frame, which is the documented worst case.

**Horizontal label overhang is intended.** Plate width is a percentage of the cell while Windows wraps the label across roughly the whole cell, so a long name can overhang a narrow plate. Constraining the plate to the measured label width would defeat the point of the setting. It is documented in the readme and in the width setting's description instead.

**The drag image keeps the stock highlight**, since `LVM_CREATEDRAGIMAGE` does not draw through the paint messages the depth counter covers. That is the behaviour I want.

/ready-for-reviewer

RiteshKhandekar:
Deadlock confirmed and removed - thanks, that was a bad one. I verified the mechanism in `windhawk_utils.h` rather than taking it on trust: both helpers end in `SendMessage` to the window's own thread (lines 325 and 334), and the `CreateWindowExW` hook calls `AttachToListView` from exactly that thread, so the two could wait on each other forever.

Worth recording why the comment defending the lock was wrong: it argued that `LvSubclassProc` never takes the lock, which is true and beside the point. The contending party was the other caller of `AttachToListView`, not the subclass procedure. It was also a poor trade in the first place - the lock was added for an optional finding whose worst case was the mod being inert until the next desktop rebuild, and it replaced that with a shell hang.

Gone, along with the ordering concern it was meant to close: `Wh_ModAfterInit` now returns early when the hook has already attached a list view. The remaining GDI+ lock is never held across a send, and in `Wh_ModUninit` the subclass removal completes before it is taken.

All the optional items from this round are in as well - the three unclamped pixel settings, `Header` and `Button` on the reject list, the unreachable lookup and unused include removed, and the Compatibility note about only the non-`Ex` export being hooked.

Tested on Windows 11 26200 at 200% scaling, including repeated Explorer restarts, which is the path the deadlock lived on.

/ai-review

RiteshKhandekar:
All three optional items from the last round are in.

**Settings are published atomically.** `LoadSettings` built `g_settings` field by field while the desktop thread painted from it, so the visible part of that race was the colour transient you pointed at - each colour is seeded with its default before parsing, so a repaint landing in that window flashed the fallback. It is built into a local and assigned once now. Worth noting this is the same finding I declined in three earlier rounds on the grounds that the fix was a lock in the per-item paint path; the local-and-publish form has no lock at all, so that objection did not apply to it and I should have taken it sooner.

**A rejected colour string is logged** rather than silently dropped, through a small helper that also removed the repetition across the three colour settings.

**The mixed-DPI paragraph is corrected.** You were right that it described a mechanism the code does not have - nothing is measured per-monitor, `Scale()` uses one cached DPI from the desktop window. It now says the plate and the icon are scaled by the same number, since the item rectangles come from that same window, which is the actual reason the conclusion holds. The colour documentation also now mentions that the leading hash is optional and that a rejected value is logged.

Left alone, deliberately, and unchanged from the previous note: `LISS_SELECTEDNOTFOCUS` matching the focused plate (a uniform look is what I want here, and the state is available if that changes), the glow ceiling on select-all with an extreme `glowSize` (the default of 4 is cheap and the description warns about the per-unit cost), and `dashDot`'s fixed one-pen-width dot at thick borders.

/ai-review

PR #5193, opened 2026-08-24
Exclude effect on taskbar

Anixx:
## Changelog

* Changelog item 1...
* Changelog item 2...

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/disable-inactive-menubar-greying.wh.cpp
+++ b/mods/disable-inactive-menubar-greying.wh.cpp
@@ -2,7 +2,7 @@
 // @id              disable-inactive-menubar-greying
 // @name            Disable Inactive Menubar Greying
 // @description     Prevents menubar text from being greyed out in inactive folder windows in Classic theme
-// @version         1.2
+// @version         1.3
 // @author          Anixx
 // @github          https://github.com/Anixx
 // @include         explorer.exe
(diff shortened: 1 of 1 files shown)


Anixx:
I implemented the second suggestion, but cannot implement the first one, because there can be multiple taskbars, with class name Shell_SecondaryTrayWnd and their set can change with time. But I made the scan to happen only once per thread.

Anixx:
Your filtering suggestion is useless: the Control Panel can have a menubar, yet if other checks are omitted, the artifacts affect the form controls (drop-down boxes, etc). I already tried all this, don't suggest me ideas that I already tried multiple times and that were suggested by other AIs among the first.

Anixx:
I've made the variables thread_local, but your scenario that a taskbar thread prints some text before taskbar window is created is unrealistic. If a thread outputs something it means at least one window in the thread has been created, and if it is a taskbar thread, then the taskbar has been created before any text output.

m417z:
You can also use [GetThreadDescription](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-getthreaddescription) for getting the thread name. The taskbar thread name is "Taskbar" :)
You can see thread names with [WinSpy++](https://github.com/m417z/winspy/releases) or System Informer.

Anixx:
> You can also use [GetThreadDescription](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-getthreaddescription) for getting the thread name. The taskbar thread name is "Taskbar" :) You can see thread names with [WinSpy++](https://github.com/m417z/winspy/releases) or System Informer.

Oh, thanks! What is the name of secondary taskbars?

m417z:
All taskbars are on the same thread.

PR #5198, opened 2026-08-24
Update version and HomeGroup settings in mod (minor changes)

babamohammed2022:
This PR includes an updated version number and modified HomeGroup settings to be disabled by default and Windows 11 25H2 has been added to the compatibility notes.

## Changelog

- Applied minor improvements
- Updated the README to clarify a mod suggestion

## Mod authorship

This mod was created by:

- - [X] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/win7-legacy-applet-restorer.wh.cpp
+++ b/mods/win7-legacy-applet-restorer.wh.cpp
@@ -35,9 +35,10 @@ The optional "Restore Classic Task Links" setting restores localized, classic ta
 ![screenshot](https://raw.githubusercontent.com/babamohammed2022/babamohammed2022/main/legacyappet.png)
 
 ## Notes
-The mod has been tested on Windows 10 1809, Windows 10 21H2 and Windows 11 24H2 and Windows 11 25H2.
+The mod has been tested on Windows 10 1809, Windows 10 21H2, Windows 11 24H2, and Windows 11 25H2.
 
-HomeGroup is disabled by default because on Windows 11 the page was completely removed, use the https://windhawk.net/mods/win11-home-group-restorer mod to restore it.
+HomeGroup is disabled by default because the page was removed from Windows 11. To restore it there, use the [Windows 11 HomeGroup Page Restorer](https://windhawk.net/mods/win11-home-group-restorer) mod.
+
 BitLocker Drive Encryption and Tablet PC Settings default to **Automatic**: they are only added when the applet exists on the machine *and* Control Panel does not already show it, so no duplicate entries appear on editions and devices where Windows lists them by itself (e.g. Pro/Enterprise with a TPM, or a pen/touch-capable device). Whether the applet is already shown is asked of the shell itself (`IOpenControlPanel::GetPath`), because the `ControlPanel\NameSpace` registry key alone is not reliable — on Windows 10 LTSC 2021 it is present even though the applet is not displayed.
 
 If the automatic detection is wrong on your edition, each of the two applets has an **Always add** / **Never add** override in the settings. "Always add" still does nothing when the applet is genuinely not installed (e.g. Windows Home), since the entry would have no name, icon or target.
@@ -69,7 +70,7 @@ Credits to m417z for the code review and enhancing the mod.
   $description: This setting adds the "Printers and Faxes" icon to the Control Panel
 - enableHomeGroup: false
   $name: HomeGroup
-  $description: This setting restores navigation to the HomeGroup page only when Windows still registers its legacy CLSID. For this mod, successful page availability satisfies the feature goal and preserves compatibility with present or future external HomeGroup-restoration projects; networking functionality is not implied. It is recommended to use the fix proposed above for simplicity.
+  $description: This setting adds the HomeGroup entry when Windows still registers its legacy CLSID (page only, the HomeGroup networking was removed in Windows 10 1803+). On Windows 11 use the "Windows 11 HomeGroup Page Restorer" mod instead.
 - bitLockerMode: auto
   $name: BitLocker Drive Encryption
   $description: Adds the "BitLocker Drive Encryption" icon to the Control Panel (System and Security category). "Automatic" adds it only when the applet exists on this machine and Control Panel does not already show it, so no duplicate entry appears. If the detection gets it wrong on your edition, force it with "Always add" or "Never add".
@@ -205,6 +206,10 @@ std::wstring g_personalizationName;
 // projects without claiming that the removed networking service itself works. If
 // Windows no longer exposes the CLSID, no virtual replacement is made.
 static std::atomic<bool> g_homeGroupClsidAvailable{ false };
+// Additional check: the implementation file (hgcpl.dll) must actually exist.
+// On Windows 11 the CLSID may be registered but the DLL is missing, so the
+// page would be non-functional.
+static std::atomic<bool> g_homeGroupImplementationExists{ false };
 // BitLocker and Tablet PC Settings are real, unmodified Windows CLSIDs; they
 // are simply not always *registered* (BitLocker needs Pro/Enterprise+TPM,
 // Tablet PC Settings needs a touch/pen-capable device). Unlike HomeGroup,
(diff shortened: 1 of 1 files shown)


babamohammed2022:
> ### Submission review
> 
> 
> 
> The bulk of the change is the new `hgcpl.dll` gate for HomeGroup, and that's where the problems are.
> 
> **1. The `System32\hgcpl.dll` check probably makes `enableHomeGroup` a permanent no-op on every build this mod supports.**
> 
> `Wh_ModInit` now requires `%SystemRoot%\System32\hgcpl.dll` to exist before HomeGroup is injected. But `hgcpl.dll` was removed from Windows in 1803 — your own companion mod documents exactly that: it tells the user to obtain the DLL from ["A Windows 10 version < 1803 (C:\Windows\System32)"](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/win11-home-group-restorer.wh.cpp#L8383), and its pinned download targets are 1607/1703 builds, with the 1703 one commented as _"Last version of 1703 before HomeGroup removal in 1803"_.
> 
> The README of this mod lists Windows 10 1809, 10 21H2, 11 24H2 and 11 25H2 as the tested platforms. If the file is absent on all four, then `g_homeGroupUsable` is always `false`, the setting does nothing no matter what the user picks, and the only signal is a debug log line. Combined with the default flip to `false`, HomeGroup is effectively removed from the mod — while the README still advertises it (`HomeGroup (partially functional, read the note below)`, plus the "Screenshot (for the HomeGroup and Network Connections applets...)" section) and the setting description still describes the old CLSID-only condition without mentioning the DLL requirement at all.
> 
> Please check on your 1809/21H2 machine whether `C:\Windows\System32\hgcpl.dll` is actually there. Then either:
> 
> * it is there → keep the gate, but say so in the `$description` and in the README note (which currently only explains the Windows 11 case); or
> * it isn't → the gate silently kills the feature everywhere. Then just say that plainly: drop HomeGroup from the applet list in the README, mark the setting as applying only to pre-1803 builds, and point everyone at the companion mod.
> 
> Either way the note at line 40 needs to cover Windows 10 1803+ as well, not only Windows 11.
> 
> **2. Don't hardcode `System32\hgcpl.dll` — resolve the CLSID's registered implementation instead.**
> 
> The comment says _"the implementation file (hgcpl.dll) must actually exist"_, but the check answers a different question: it tests one hardcoded path rather than the module the shell will actually load for `{67ca7650-...}`. Read the registration instead:
> 
> ```c++
> bool ClsidImplementationExists(const std::wstring& guid) {
>     ScopedHKey key;
>     const std::wstring path = L"CLSID\\" + guid + L"\\InprocServer32";
>     if (RegOpenKeyExW(HKEY_CLASSES_ROOT, path.c_str(), 0, KEY_READ, key.AddressOf()) != ERROR_SUCCESS)
>         return false;
>     std::wstring server;
>     if (!ReadStringValue(key.Get(), nullptr, server) || server.empty())
>         return false;
>     wchar_t expanded[MAX_PATH] = {};
>     if (!ExpandEnvironmentStringsW(server.c_str(), expanded, ARRAYSIZE(expanded)))
>         return false;
>     const DWORD attributes = GetFileAttributesW(expanded);
>     return attributes != INVALID_FILE_ATTRIBUTES && !(attributes & FILE_ATTRIBUTE_DIRECTORY);
> }
> ```
> 
> This also matters for the design goal the mod still states at lines 203–208 — _"preserves a navigation target for present or future restoration projects"_. A `System32`-only file test can never see the companion mod's verified copy, which lives under `Wh_GetModStoragePath()` and is wired up in-process rather than via `System32`. As written, the new check contradicts that comment; if you're dropping that goal, please delete the comment too. The `Wh_Log` at line 2690 (`"...is registered; applet enabled when selected"`) is also no longer accurate now that there's a second gate.
> 
> **3. The HomeGroup task links are still emitted when the HomeGroup applet is not injected.**
> 
> `kClassicTaskLinks` (lines 1202–1205) always contains the `<application id="{67ca7650-...}">` block with `<category id="3">` whenever `restoreClassicTaskLinks` is on — it isn't gated on `enableHomeGroup` or `g_homeGroupUsable`. So with the new `false` default, every user now gets "Choose homegroup and sharing options" and "Share printers" under Network and Internet, both running `explorer.exe shell:::{67ca7650-...}`, pointing at a page the mod deliberately decided not to provide.
> 
> The virtual-applet blocks right below already do this correctly (lines 1308–1333: _"only emitted when the corresponding applet is both available on this machine and enabled in settings, mirroring how GetNamespaceClsids() gates the icon itself"_). Split the HomeGroup `<application>` block out of `kClassicTaskLinks` into its own conditional string gated on `IsHomeGroupAvailable()`, the same way. `Wh_ModSettingsChanged` already regenerates the XML, so the gating will follow the setting live.
> 
> Optional improvements
> 
> 
> * **`g_homeGroupImplementationExists` (line 213) is dead state.** It's stored at line 2707 and read once at line 2710, in the same block where the local `implementationExists` already holds the value. `g_homeGroupClsidAvailable` is in nearly the same position now that `g_homeGroupUsable` is what every call site gates on. Both could just be locals in `Wh_ModInit`.
> * **The two new stop-event checks in `RunLazyVirtualAppletDetection` (lines 742–746 and 759–763) can't fire.** That function's only caller is `LazyDetectionThreadProc`, which now checks the stop event immediately before calling it (lines 2863–2866) — and since it's the only caller, the `try_lock` never fails either, so the "re-check after acquiring the lock" path is unreachable as well. The check that does earn its place is the one in the thread proc, because `WaitForMultipleObjects` returns the lowest signalled index and would otherwise run a probe when both events are set.
>   Worth noting they also don't shorten an unload: the only slow part is `IsShownByControlPanel` (`CoCreateInstance` + up to three `IOpenControlPanel::GetPath` calls), which is uninterruptible, so `Wh_ModUninit`'s `join()` still blocks for the whole probe if it lands mid-flight. If you want that to actually bound, the probe itself would need a cancellation check between the `GetPath` candidates.
> * **Formatting churn in the diff.** The blank lines around `IsHomeGroupAvailable` and after `Wh_ModInit`'s closing brace were removed for no reason, and most of the added lines carry trailing whitespace (740, 742, 747, 754, 757, 759, 782, 2700, 2702, 2703, 2708, 2710, 2862, 2867). Keeping the diff to the behavioural change makes it easier to review.
> * **Stale author's-note comments.** Line 123–124 claims _"this mod will auto-disable if Anixx's mod is detected and log a warning"_ — there's no such detection anywhere in the file; the README just warns the user manually. And line 119 (_"Currently tested only on Windows 10 1809 ... unable to run Windows 11"_) contradicts the README line this PR just updated to include Windows 11 24H2/25H2.
> * **Line 1206: `  </application>\n)xml";`** — inside a raw string literal `\n` is a literal backslash followed by `n`, not a newline, so the generated XML gets a stray `\n` text node inside `<applications>`. Every other line in that literal uses a real newline; this one looks like an accident.
> * **`ReadAppletMode` (lines 1474–1482)** could use `WindhawkUtils::StringSetting` instead of `Wh_GetStringSetting` + `Wh_FreeStringSetting`. The `if (PCWSTR value = ...)` null test is also dead — `Wh_GetStringSetting` never returns `NULL`, it returns `L""` when unset or on error.
> 
> Functionality notes
> 
> 
> * Worth confirming which of the two HomeGroup gates actually changes anything on each build. If the legacy CLSID is not registered on Windows 11, the old `IsRegisteredClsid` check already returned `false` there and the new DLL check adds nothing on that platform; if the CLSID _is_ registered on Windows 10 1803+ but the DLL is gone, the CLSID check adds nothing there. Logging both verdicts separately (which the code already does) and checking a 1809 box and a 24H2 box would settle it quickly.
> * Since HomeGroup is now off by default, the companion mod's coexistence check ("yield if the entry is already being provided") will always take the "provide it myself" path when both mods are enabled. That's a sane outcome, but it does mean this mod no longer wins that race on Windows 10 either — worth a sentence in the README so users who liked the old behaviour know why the entry moved.
> 
> **Next steps:**
> 
> * `/ai-review` - after pushing fixes, to get a review of the updated code. You can repeat this as many times as you need, but each review is thorough and usually there's no need for more than 2-3 iterations.
> * `/ready-for-reviewer` - once you're satisfied with the state of the pull request, to hand it over to a human reviewer. If some findings above are left unaddressed, add a short note explaining why.
> 

In my case the hgcpl.dll is present in Windows 11 24H2 so I think that checking if the file is present is not that big of a problem as there is a suggested fix made specifically for this problem and if the user does want to install the mod he can simply disable the option. If the maintainer believes that these problems really need to be addressed, I'll do it.
/ready-for-reviewer

PR #5200, opened 2026-08-24
Add MuteAlert 0.9.7

Nikolay1243:
## Changelog

- Add MuteAlert 0.9.1 as a new Windhawk mod.
- Show live microphone activity, Windows and call mute controls, and headset mute synchronization in the Windows 11 taskbar.

## Mod authorship

This mod was created by:

- [ ] Claude
- [x] ChatGPT
- [ ] Gemini
- [ ] Other (please specify):

### DIFF
--- a/mods/mutealert.wh.cpp
+++ b/mods/mutealert.wh.cpp
@@ -2,7 +2,7 @@
 // @id              mutealert
 // @name            MuteAlert - Microphone Activity Taskbar Widget
 // @description     Shows live microphone activity, call mute state, volume controls, and headset mute synchronization in the Windows 11 taskbar.
-// @version         0.9.6
+// @version         0.9.7
 // @author          Nikolay
 // @github          https://github.com/Nikolay1243
 // @homepage        https://github.com/MuteAlert/windhawk
@@ -548,6 +548,7 @@ static std::atomic<int> g_pendingMuteSet{-1};
 static std::atomic<int> g_pendingSlackCommand{-1};
 static std::atomic<int> g_pendingTeamsCommand{-1};
 static std::atomic<int> g_pendingZoomCommand{-1};
+static std::atomic<int> g_pendingFocusCall{-1};
 static std::atomic<bool> g_slackCallActive{false};
 static std::atomic<bool> g_slackMuted{false};
 static std::atomic<bool> g_slackWarningActive{false};
@@ -866,23 +867,10 @@ static int GetSelectedActiveCallIndex() {
     return -1;
 }
 
-static bool FocusSelectedCallWindow() {
+static bool QueueFocusSelectedCallWindow() {
     int selected = GetSelectedActiveCallIndex();
-    HWND callWindow = selected == 0   ? g_slackCallWindow.load()
-                      : selected == 1 ? g_teamsCallWindow.load()
-                      : selected == 2 ? g_zoomCallWindow.load()
-                                      : nullptr;
-    if (!callWindow || !IsWindow(callWindow)) return false;
-
-    if (IsIconic(callWindow)) {
-        ShowWindowAsync(callWindow, SW_RESTORE);
-    } else if (!IsWindowVisible(callWindow)) {
-        ShowWindowAsync(callWindow, SW_SHOW);
-    }
-    BringWindowToTop(callWindow);
-    bool focused = SetForegroundWindow(callWindow) != FALSE;
-    Wh_Log(L"[Call apps] Focus %s for window %p",
-           focused ? L"requested" : L"was denied by Windows", callWindow);
+    if (selected < 0) return false;
+    g_pendingFocusCall.store(selected);
     return true;
 }
 
@@ -1155,9 +1143,17 @@ static bool RequestForegroundWindow(HWND hWnd) {
     hWnd = GetAncestor(hWnd, GA_ROOT);
     if (!hWnd || !IsWindow(hWnd)) return false;
 
+    DWORD_PTR probeResult = 0;
+    if (!SendMessageTimeoutW(hWnd, WM_NULL, 0, 0,
+                             SMTO_ABORTIFHUNG | SMTO_BLOCK, 200,
+                             &probeResult)) {
+        Wh_Log(L"[Call apps] Refusing to activate unresponsive window %p",
+               hWnd);
+        return false;
+    }
     if (IsIconic(hWnd)) ShowWindowAsync(hWnd, SW_RESTORE);
-    BringWindowToTop(hWnd);
-    SetForegroundWindow(hWnd);
+    if (!IsWindowVisible(hWnd)) ShowWindowAsync(hWnd, SW_SHOW);
+    bool focusRequested = SetForegroundWindow(hWnd) != FALSE;
 
     HWND actualForeground = GetForegroundWindow();
     DWORD requestedProcess = 0;
@@ -1167,7 +1163,12 @@ static bool RequestForegroundWindow(HWND hWnd) {
         GetWindowThreadProcessId(actualForeground, &actualProcess);
     }
 
-    return requestedProcess && requestedProcess == actualProcess;
+    bool focused = requestedProcess && requestedProcess == actualProcess;
+    Wh_Log(L"[Call apps] Focus %s for window %p",
+           focused ? L"requested" : focusRequested ? L"pending"
+                                                 : L"was denied by Windows",
+           hWnd);
+    return focused;
 }
 
 static bool SendZoomMuteShortcut(HWND callWindow) {
@@ -1492,6 +1493,14 @@ static DWORD WINAPI CallAppsThreadProc(void*) {
     while (!IsStopping() &&
            WaitForSingleObject(g_audioStopEvent, 50) == WAIT_TIMEOUT) {
         ULONGLONG now = GetTickCount64();
+        bool notify = false;
+        auto publishBool = [&notify](std::atomic<bool>& target, bool value) {
+            if (target.exchange(value) != value) notify = true;
+        };
+        auto publishWindow = [&notify](std::atomic<HWND>& target,
+                                       HWND value) {
+            if (target.exchange(value) != value) notify = true;
+        };
         if (lastCaptureSessionCheck == 0 ||
             now - lastCaptureSessionCheck >= 2000) {
             captureAppMask = gateOnCaptureSessions
@@ -1518,9 +1527,10 @@ static DWORD WINAPI CallAppsThreadProc(void*) {
                              : CallState::NotInCall;
             now = GetTickCount64();
             lastSlackCheck = now;
-            g_slackCallActive.store(slackState != CallState::NotInCall);
-            g_slackMuted.store(slackState == CallState::Muted);
-            g_slackCallWindow.store(callWindow);
+            publishBool(g_slackCallActive,
+                        slackState != CallState::NotInCall);
+            publishBool(g_slackMuted, slackState == CallState::Muted);
+            publishWindow(g_slackCallWindow, callWindow);
         }
 
         int teamsCommand =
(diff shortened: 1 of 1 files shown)


Nikolay1243:
Addressed the automated review in c14558e (mod version 0.9.2):

1. **Retry lifecycle race:** retry-thread stop/start operations and settings reloads are serialized with a lifecycle mutex. Concurrent taskbar retry requests use `try_lock`, so a taskbar callback cannot deadlock behind a settings reload that is waiting for the old thread.
2. **Zoom fallback:** removed `AttachThreadInput`/`SetFocus`; Alt+A is now an explicit opt-in setting, is bounded to three attempts, restores the previous foreground window, and is skipped during unload.
3. **Raw input cleanup:** added matching `RIDEV_REMOVE` registrations for both generic-system-control and telephony HID usage pages before destroying the message window.
4. **Stale widgets:** an existing named widget is removed with its grid column and the remaining columns/spans are adjusted before reinjection. The stale element is no longer adopted as a successful injection.
5. **UI Automation cost:** all call integrations are disabled by default; enabled integrations are gated by active capture sessions, process-image lookups are cached, scans run at a reduced cadence, and UIA properties/patterns are fetched through a cache request.
6. **Unload/cancellation:** call-state scans and window enumeration check the unloading flag, the Zoom fallback is suppressed while unloading, waits are stop-event-aware, and the call thread enables COM cancellation so shutdown can call `CoCancelCall` before joining it.
7. **Architecture/prologue handling:** retained catalog-compatible `x86-64` metadata and added an actual ARM64 instruction decoder based on the supported Taskbar implementation. Both paths validate the expected prologue and fail closed instead of guessing an offset.
8. **Screenshot:** added a live taskbar-widget screenshot to the mod readme, hosted in the MuteAlert source repository.
9. **Related mods:** added a readme section explaining the functional differences from `mic-tray-control` and `adaptive-microphone-icon-visibility`.

Also addressed the optional findings: removed the unused symbol lookup, switched string settings to Windhawk's RAII wrapper, fixed version logging, gated diagnostic report capture, cached HID metadata/preparsed data, reduced large stack path buffers, delayed diagnostic export briefly to capture report changes, used the displayed peak for warning thresholds, and added localized muted/unmuted labels.

Validation completed locally:

- Windhawk compiler: x64 and ARM64 builds succeeded without warnings.
- `.github/pr_validation.py mods/mutealert.wh.cpp Nikolay1243`: passed without warnings.
- C++ line endings: CRLF verified.

Nikolay1243:
Addressed the second automated review in 1acf36e (MuteAlert 0.9.3):

1. **Exception-safe XAML injection**
   - `InjectWidget` now wraps the entire WinRT/XAML path in `try`/`catch (...)`.
   - Controls and the original sibling column/span layout are captured before mutating the taskbar grid.
   - A failed injection removes a partially appended widget, restores every captured sibling column/span, and removes the inserted column.
   - The `EnumThreadWindows` callback is exception-guarded, as is the initial/update path in the taskbar subclass.
   - `XamlRootFromTaskbarHostSharedPtr` now uses an RAII guard so `_Decref` runs on every return and when `XamlRoot()` throws.

2. **Only the taskbar-owning Explorer runs MuteAlert services**
   - `Wh_ModInit` skips folder-only Explorer processes when another process owns `GetShellWindow()`.
   - Audio, call-app, headset, and retry threads only start after `FindMainTaskbarWindow()` confirms that the current process owns a taskbar.
   - `TrayUI::StartTaskbar` and secondary-taskbar hooks start the services if the taskbar appears after mod initialization.
   - Settings reloads likewise restart services only when the process currently owns the main taskbar.

3. **Settings changes now stop call-app polling promptly**
   - Added the shared `IsStopping()` check, covering both unload and a signalled `g_audioStopEvent`.
   - Applied it at capture-session entry and between COM stages, during session enumeration, at `ReadCallState` entry, in the `EnumWindows` callback, between windows and cached buttons, in the call-app loop, Zoom process enumeration, and the deferred Zoom fallback.
   - This prevents a cancelled UIA request from falling through into another cross-process request while `Wh_ModSettingsChanged` is waiting.

Also completed several optional cleanups: removed the now-unused linear-peak atomic, registered the taskbar update message once during initialization and guarded the taskbar subclass update path, added a steady-state early-out for unchanged Windows hardware-mute state, and reject a stale/unavailable headset message-window class.

Validation:

- Clean Windhawk x64 build: passed without warnings.
- Clean Windhawk ARM64 build: passed without warnings.
- Official `.github/pr_validation.py mods/mutealert.wh.cpp Nikolay1243`: passed without warnings.
- `git diff --check`: passed.
- C++ CRLF-only line endings: verified.

Nikolay1243:
Addressed the third automated review in 6820579 (MuteAlert 0.9.4):

1. **Widget column ownership**
   - `RemoveWidgetState` now initializes `liveColumn` to `-1`, tracks whether the named widget was actually removed, and returns without touching the grid if the element is no longer present.
   - A taskbar-owned column can no longer be deleted using the stale `WidgetState::column` value.

2. **SteelSeries probing cost**
   - HID interface paths are filtered for SteelSeries VID `1038` and the supported Nova Pro product IDs before any `CreateFileW` call.
   - The adapter now probes once at startup and then re-probes on HID interface arrival/removal notifications rather than enumerating every two seconds forever.
   - Registered `WM_DEVICECHANGE` notifications cover the vendor-specific HID interface; the existing Raw Input device-change path also requests a re-probe.
   - A connected device continues polling at the configured interval. A read failure receives one delayed reconnect attempt, then waits for a device event.
   - If Windows device notifications cannot be registered, the fallback probe is reduced to once per 60 seconds.
   - The headset loop now sleeps until its next actual poll/diagnostic/reconnect deadline instead of waking every 50 ms.

3. **Guaranteed subclass teardown**
   - Replaced raw `SetWindowSubclass` with `WindhawkUtils::SetWindowSubclassFromAnyThread` and the matching removal helper.
   - `RemoveWidgetsOnWindowThread` now exchanges and removes the stored subclass unconditionally, even when taskbar-thread XAML cleanup fails.
   - `WM_NCDESTROY` clears `g_mainTaskbarWnd`; Windhawk's helper owns automatic subclass removal.

Also addressed carried-over optional items:

- Removed the unused `WidgetState::fillLayer` reference.
- Reset diagnostic events/timestamps after a settings restart.
- Snapshot and restore pre-existing Explorer Raw Input registrations on teardown.
- Moved forced-volume persistence from the taskbar UI callback to the audio worker, with a shutdown flush.

Validation:

- Clean Windhawk x64 build: passed without warnings.
- Clean Windhawk ARM64 build: passed without warnings.
- Official `.github/pr_validation.py mods/mutealert.wh.cpp Nikolay1243`: passed without warnings.
- `git diff --check`: passed.
- C++ CRLF-only line endings: verified.

Nikolay1243:
The three findings from the latest completed AI review were addressed in 6820579 (MuteAlert 0.9.4). A fourth `/ai-review` request was rate-limited because the PR has reached the three-reviews-per-24-hours limit. Clean x64 and ARM64 builds, the official PR validator, and GitHub checks all pass.

Nikolay1243:
Addressed the final automated review in 3dfc9d1 (MuteAlert 0.9.5).

1. **Explorer UI-thread/render cost**
   - Coalesced meter notifications with an atomic pending flag, including reset paths for failed posts, message consumption, and taskbar destruction.
   - Added a cached `RenderSnapshot` and quantized the displayed peak to permille precision.
   - Fill-only changes now update only the clip rectangle. Tooltip strings, vectors, call-app visuals, and other non-meter XAML properties are updated only when their relevant state changes.
   - Device-name and headset-status generation counters avoid recurring string copies when those values are unchanged.
   - The render cache is reset during widget teardown/rebuild and invalidated if a XAML update throws.

2. **Raw Input registration snapshot sizing**
   - The size probe now handles the documented `UINT_MAX` plus `ERROR_INSUFFICIENT_BUFFER` convention used when the buffer is null or too small: [GetRegisteredRawInputDevices documentation](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getregisteredrawinputdevices).
   - Headset synchronization now defaults to **Off**, so Raw HID and SteelSeries behavior is opt-in.

Also included the optional cleanups that were safe for this revision:

- Removed the unused `WidgetState::column` member.
- Replaced the `SetWindowsHookEx` widget-removal fallback with a dedicated registered message handled by the existing taskbar subclass and sent synchronously to the taskbar thread.
- Added exception guards around headset-window Raw Input handling.
- Clarified that icon-size settings are in DIPs.

Validation completed successfully:

- Windhawk compile: x86-64
- Windhawk compile: ARM64
- Repository `pr_validation.py`
- `git diff --check`
- CRLF verification for the C++ source

Nikolay1243:
Addressed the two required 0.9.5 findings in 2d243f3 (MuteAlert 0.9.6).

1. **Safe behavior when the main taskbar subclass cannot be installed**
   - Widget enumeration now carries an explicit `subclassFailed` result.
   - A subclass failure stops enumeration immediately, removes any widgets that may already have been injected into secondary taskbars, clears the main taskbar target, and leaves `g_widgetsLive` false.
   - The existing bounded retry thread can therefore try again without leaving live XAML controls whose handlers cannot be updated or safely removed.

2. **Diagnostics-only mode no longer changes Explorer Raw Input routing**
   - Standard HID Raw Input and HID device-change notifications are registered only when `headsetSyncMode != off`.
   - Diagnostics-only mode can still enumerate device descriptors and vendor provider information, but does not take over Raw Input registrations.
   - The setting description now explicitly states that sanitized report-change capture requires headset synchronization to be enabled.
   - Failed Raw Input registration also discards the unused registration snapshot, and teardown restores registrations only when registration actually succeeded.

I also applied two low-risk optional hardening changes:

- The audio thread notifies Explorer only when the published availability, mute state, volume, quantized peak, or device name changes.
- The subclass removal-message path now has the same exception guard as the update-message path.

The remaining optional architectural and functionality observations are intentionally left for future work; they are not required for safe unload or correct default behavior in this submission.

Validation completed successfully:

- Windhawk compile: x86-64
- Windhawk compile: ARM64
- Repository `pr_validation.py`
- `git diff --check`
- CRLF verification for both C++ source copies

Nikolay1243:
Addressed both required 0.9.6 findings in 8bfe971 (MuteAlert 0.9.7).

1. **Call-app state changes notify the widget again**
   - `CallAppsThreadProc` now publishes Slack, Teams, and Zoom active/muted/known/window state through change-detecting helpers.
   - Any rendered call-state transition queues a taskbar update, independently of microphone activity and warning settings.
   - The 0.9.6 optimization remains intact: unchanged audio and unchanged call state do not wake the taskbar UI thread.

2. **Call-window focus no longer runs on Explorer's taskbar UI thread**
   - The call-icon click handler now records an atomic focus request and returns immediately.
   - `CallAppsThreadProc` consumes the request and performs activation on the worker thread.
   - `RequestForegroundWindow` first probes the target with `SendMessageTimeoutW(WM_NULL, SMTO_ABORTIFHUNG | SMTO_BLOCK, 200 ms)` and refuses to activate a non-responsive window.
   - The redundant synchronous `BringWindowToTop` call was removed; restoration/showing remains asynchronous.
   - The same bounded responsiveness probe also protects Zoom shortcut activation and foreground restoration.

The collapsed review items are optional architectural or functionality improvements and remain intentionally deferred; no further optional refactoring is planned for this submission.

Validation completed successfully:

- Windhawk compile: x86-64
- Windhawk compile: ARM64
- Repository `pr_validation.py`
- `git diff --check`
- CRLF verification for both C++ source copies

Nikolay1243:
Thanks for the final review. The current 0.9.7 head has no merge-blocking findings.

The remaining items are explicitly optional architectural improvements or non-critical functionality notes. They are intentionally deferred to future releases to avoid expanding this submission further. In particular, no additional taskbar wake is needed for window-handle-only changes because those handles are not rendered, and the broader thread, UI Automation, endpoint-notification, HID, tooltip, smoothing, layout, and settings refinements can be evaluated independently after the initial release.

The required lifecycle, safe-unload, Raw Input, call-state notification, and hung-window activation issues raised during review have all been addressed and validated on the current head.

PR #5202, opened 2026-08-24
Add Glass Cursors mod

fizixes:
Adds Glass Cursors, a DPI-aware translucent cursor replacement with
configurable pointer styles, glass color, opacity, sizing, and animated
loading cursors.

## Changelog

* Changelog item 1...
* Changelog item 2...

## Mod authorship

This mod was created by:

- - [ X] The submitter, with AI assistance
- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/glass-cursors.wh.cpp
+++ b/mods/glass-cursors.wh.cpp
@@ -2,9 +2,8 @@
 // @id              glass-cursors
 // @name            Glass Cursors
 // @description     Original DPI-aware translucent glass system cursors with a live animated loading indicator.
-// @version         0.18.1
+// @version         0.18.3
 // @author          fizixes
-// @github          https://github.com/fizixes
 // @license         MIT
 // @include         explorer.exe
 // @compilerOptions -lgdi32
@@ -36,12 +35,12 @@ Created and maintained by `fizixes`.
 
 // ==WindhawkModSettings==
 /*
-- FillOpacity: 25
+- FillOpacity: "25"
   $name: Glass fill
   $description: Interior opacity of the glass cursor shapes.
   $options:
-  - 25: 25% fill
-  - 50: 50% fill
+  - "25": 25% fill
+  - "50": 50% fill
 - GlassColor:
   - Red: 33
     $name: Red (0-255)
@@ -51,15 +50,15 @@ Created and maintained by `fizixes`.
     $name: Blue (0-255)
   $name: Internal glass color
   $description: RGB color of the translucent interior. Opacity is controlled separately by Glass fill.
-- CursorSize: 0
+- CursorSize: "0"
   $name: Cursor resolution
   $description: Automatic uses the current Windows cursor metric and system DPI.
   $options:
-  - 0: Automatic (DPI-aware)
-  - 32: 32 px
-  - 48: 48 px
-  - 64: 64 px
-  - 96: 96 px
+  - "0": Automatic (DPI-aware)
+  - "32": 32 px
+  - "48": 48 px
+  - "64": 64 px
+  - "96": 96 px
 - PointerStyle: sharpRounded
   $name: Pointer style
   $description: Choose the pointer silhouette and notch geometry.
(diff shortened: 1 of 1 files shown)


SoftwareExplorer6600:
Hey @fizixes is there a way we can implement the mica or mica alt effect under the cursor, that's one thing I wish we could do, maybe you can help me figure it out in this mod.

fizixes:
> Hey @fizixes is there a way we can implement the mica or mica alt effect under the cursor, that's one thing I wish we could do, maybe you can help me figure it out in this mod.

I would love to, unfortunately given that the cursor is just an image file, and not an actual render, it would be highly involved. Not impossible, but you'd have to effectively take a screenshot of the position of the mouse, post-process it, and then apply it. for each frame. I don't think the time it would take to code this subtracting the unviability of actually running it makes it worth it. 

All that said, I think I could probably make a more Mica-like effect than the current option, specular noise and the mush. Expect an update

Gaming-tech-cyber:
I noticed a bug in 0.20.0: does not work with this mod: macOS magnifying cursor or if you search by flag its mac-magnifying-cursor, created by Jaali, however they don't immediately conflict, the glass cursor plays the animation as it does scale up on rapid shaking movement, the cursor is there, until the last second where it starts shrinking back to normal (noticed this on DPI aware, not sure if it applies to fixed pixel sizes too), both mods were on their default settings. (note, may affect earlier variations as well, 0.20.0 is confirmed to be affected)

edit: may want to independently test and confirm windhawk 2.0 alpha support, while i use 1.7.3 on my main, in my VM i have a slightly modified version of this mod that I made changes to work with windhawk 2.0's alpha 3, it complained about lines 462 - 474, which that version of this mod i edited for my VM has omitted and it still works as intended, if you want i could send it over at your repo for this mod so you can review the modified version there (to be clear my host uses 0.20.0 on windhawk 1.7.3 or current release)

fizixes:
> I noticed a bug in 0.20.0: does not work with this mod: macOS magnifying cursor or if you search by flag its mac-magnifying-cursor, created by Jaali, however they don't immediately conflict, the glass cursor plays the animation as it does scale up on rapid shaking movement, the cursor is there, until the last second where it starts shrinking back to normal (noticed this on DPI aware, not sure if it applies to fixed pixel sizes too), both mods were on their default settings. (note, may affect earlier variations as well, 0.20.0 is confirmed to be affected)
> 
> edit: may want to independently test and confirm windhawk 2.0 alpha support, while i use 1.7.3 on my main, in my VM i have a slightly modified version of this mod that I made changes to work with windhawk 2.0's alpha 3, it complained about lines 462 - 474, which that version of this mod i edited for my VM has omitted and it still works as intended, if you want i could send it over at your repo for this mod so you can review the modified version there (to be clear my host uses 0.20.0 on windhawk 1.7.3 or current release)

As it turns out, it's insanely computationally cheap, so I've added Mica, Acryllic, and Mica alt. It should work for 2.0 alpha 3 now, and also with the macOS magnifying cursor now, it retreats when another program wants to change the cursor. awaiting AI review now.

SoftwareExplorer6600:
I will give a bit of feedback, I like the effects, they were implemented very closely with the original design of the effects, the last big toggle I think would be useful is maybe accent color for the loading wheel (you know the part thats visibly spinning) this should be for the working (loading wheel with cursor) and busy (just loading wheel) states. otherwise its very well implemented.

bbmaster123:
super cool! 
I personally think it would be nicer for Artwork size to be more granular. I'd also prefer a more standard/classic styled cursor personally. I've forked a copy for myself and made something useable if you decide to just drop this in, or something similar:

```
void RenderArrowShapeDefault(Surface& s, int n) {

    const double ow = n * 0.046;
    const double sh = n * 0.014;


 std::vector<Point> pts = {
    P(0.165, 0.070, n),  // tip

    P(0.165, 0.725, n),  // long left edge

    P(0.330, 0.570, n),  // inside lower corner

    P(0.475, 0.845, n),  // tail lower-left
    P(0.555, 0.800, n),  // tail lower-right

    P(0.410, 0.535, n),  // tail upper-right / notch

    P(0.645, 0.535, n),  // outer arrow edge

    P(0.165, 0.070, n)   // back to tip
};
    DrawGlassPolygon(s, pts, ow, sh);
}
```

image

cheers :)

PR #5203, opened 2026-08-24
Update Move and Gather Windows to 0.1.2

fjdiazt:
## Changelog

* Add shortcuts for moving the foreground window to the primary or numbered displays 1-5.
* Default numbered shortcuts to moving one window for new installations while preserving existing users' configured gather shortcuts.
* Use stable Windows display device names for numbered destinations, with optional remapping for custom display order.
* Leave windows already on the destination display unchanged and raise moved windows without stealing focus or changing topmost state.
* Improve mixed-DPI movement and avoid activation after failed gather moves.
* Replace overlapping minimized-window checkboxes with one clear behavior dropdown and document setting migration.
* Add clearer window sizing options, hotkey validation, concise diagnostics, and display 4-5 actions.

## Validation

* Compiled successfully for x86, x64, and ARM64.
* Tested successfully in a local Windhawk installation.
* Repository validation runs through the pull request checks.

### DIFF
--- a/mods/gather-windows-to-monitor.wh.cpp
+++ b/mods/gather-windows-to-monitor.wh.cpp
@@ -1,7 +1,7 @@
 // ==WindhawkMod==
 // @id              gather-windows-to-monitor
-// @name            Gather Windows To Display
-// @description     Move one window or gather eligible open windows to a chosen display with global hotkeys
+// @name            Move and Gather Windows
+// @description     Move or gather windows with global hotkeys and remap numbered shortcuts to the intended displays
 // @version         0.1.2
 // @author          Fred
 // @github          https://github.com/fjdiazt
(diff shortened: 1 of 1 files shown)


fjdiazt:
Review follow-up: required findings and selected optional improvements are implemented. The following suggestions are intentionally deferred:

* **Precise move failure reasons:** Move failures remain counted and logged. Carrying detailed failure stages and Win32 error codes through the movement pipeline adds result plumbing for rare diagnostics, with no reported failure currently requiring it.
* **Gather demonstration:** The README explains both workflows. Recording, sanitizing, hosting, and maintaining a demonstration asset adds ongoing work unrelated to mod behavior.
* **Full center-cascade range:** Expanding around center would change established placement order and add negative-offset and wrap edge cases. Current placement remains accessible and deterministic.
* **Avoiding overlap with resident windows:** Obstacle-aware placement adds layout complexity and could make anchor behavior less predictable. Windows already on the destination intentionally remain untouched, while moved windows retain deterministic cascade placement.
* **Hotkey conflict text in settings:** Windhawk settings cannot show dynamic conflict validation. The README documents priority behavior and logs identify the exact conflict; repeating the same warning across every shortcut setting would add substantial settings noise.

fjdiazt:
Review follow-up for the latest findings:

**Addressed**

* Automatic numbering now uses connected displays in device-name order without gaps; disconnected-slot placeholders remain limited to explicit overrides.
* Primary fallback ignores placeholders, and movement rejects missing or invalid destination work areas.
* Update notes now document the foreground-move shortcut defaults and the owned-window default.
* The minimized-window dropdown now uses string options while retaining legacy boolean values.
* Scale-mode source display lookup now happens before restoring a window.
* Override entries cannot select the same physical display twice.
* README spacing and display-change logging wording are corrected.

**Deferred**

* **Dynamic DPI API lookup:** The mod targets Windows 10/11. `SetThreadDpiAwarenessContext` is available on supported releases; the PMv1 fallback remains useful on early Windows 10 versions without PMv2.
* **Synchronous `EndDeferWindowPos`:** The raise pass preserves moved-window order. It already skips windows reported as hung, and unload has a bounded wait. Win32 provides no asynchronous deferred-position equivalent, so changing this would remove the ordering behavior.
* **DPI-ratio scaling:** Scale mode intentionally scales by source and destination work areas; the per-monitor-v2 thread context prevents coordinate virtualization. Combining DPI ratio and work-area scaling would change established sizing behavior and remains application-dependent because apps can handle `WM_DPICHANGED` themselves.
* **Owner-relative dialog placement:** Owned windows remain independently cascaded when enabled, with an existing setting to exclude them. Parent-relative placement requires broader layout behavior and is not needed for the current gather model.
* **Maximized borderless fullscreen detection:** Detection remains conservative to avoid moving games and media windows. Exempting all maximized windows would weaken the default fullscreen protection.
* **Compile-time self-tests:** These remain as small regression checks for parsers and placement math.

fjdiazt:
Review follow-up for `339d9eb9`:

**Addressed**

* Replaced the type-changing `RestoreMinimized` key with a new string-only `MinimizedMode` key. Update notes now tell users who previously restored minimized windows to select that mode once.
* Scoped fullscreen protection to bulk gather actions. Explicit foreground-window shortcuts now move the selected fullscreen window.
* Reduced shell-class checks to one `GetClassName` call per window and moved the DWM cloaking query after cheaper class/style filters.
* Documented the five-entry display-order override limit in both README and setting description.

**Deferred or retained**

* **Move failure diagnostics:** Failures remain counted and logged. Detailed failure plumbing was previously deferred because no reported failure needs stage/error-code diagnostics.
* **Source-target early-out:** Retained defensively because a window can move itself between enumeration and movement.
* **Topology logging helper:** Retained. A fresh enumeration is required to log current topology; the function name describes that side effect.
* **Compile-time checks:** Retained as small parser and placement regression checks.
* **Upgrade sections:** Retained for this migration-heavy release; they can be shortened in a later release after users have transitioned.
* **Demonstration asset:** Previously deferred; README documents both workflows without adding hosted media maintenance.
* **Foreground window beneath gathered windows:** Intentional. Existing destination windows remain untouched while newly moved windows are raised, matching the requested gather behavior.
* **Mixed numbered shortcuts after upgrade:** Intentional compatibility behavior. Existing configured gather shortcuts keep priority; update notes explain new defaults.
* **Synchronous raise pass:** Previously deferred. It preserves moved-window order, skips windows already reported hung, and unload remains bounded; Win32 has no asynchronous deferred-position equivalent.
* **Center cascade range:** Previously deferred to preserve established deterministic placement.
* **Compact automatic numbering:** Intentional for gap-free shortcuts. Display order override provides direct mapping to Windows device names when desired.
* **Similar existing mod:** Already covered by the README comparison section.

fjdiazt:
Review follow-up for `4b8d093e`:

**Addressed**

* Resolved `SetThreadDpiAwarenessContext` dynamically, preserving load compatibility on older supported Windows versions.
* Missing numbered displays now log and skip the action instead of moving to primary.
* Hardened null monitor resolution and explicit no-connected-display handling.
* Avoided title/class lookups for cheap, high-volume skip reasons.
* Added compile-time length checking for hotkey mode/name tables.
* Updated settings and documentation for missing-display behavior, Windows Settings numbering differences, DPI-unaware windows, duplicate primary/device mappings, and the corrected migration setting name.

**Retained or deferred**

* Foreground display 1-5 defaults remain intentional; unavailable targets now do nothing.
* The comma-separated display override remains the chosen compact settings interface.
* Detailed asynchronous move-result accounting and DWM border compensation remain deferred; current limitations are documented.
* A demonstration asset remains deferred; both workflows are documented in the README.

Validation: x86, x64, and ARM64 compilation passed.

PR #5207, opened 2026-08-25
SnapSentry 0.18.4

mario0318:
Update to SnapSentry (`snap-sentry`), version 0.18.4.

- When a multi-page or animated image is kept instead of deleted, a notice now says so. Before, that only appeared in the log.
- More reliable folder watching on network or sync-backed locations that don't deliver change notifications.
- Copies large images using less memory.
- Deleting to the Recycle Bin works again for a watched folder whose path ends in a separator. The composed path was not in the form the shell accepts, so the delete failed and the file was kept.

Tested on Windows 11 build 26200.9267, Windhawk 1.7.3 x86: a new capture is copied and reaches the Recycle Bin, a large capture does the same, and a multi-page image is copied first frame only and kept rather than deleted.

### DIFF
--- a/mods/snap-sentry.wh.cpp
+++ b/mods/snap-sentry.wh.cpp
@@ -2,7 +2,7 @@
 // @id              snap-sentry
 // @name            SnapSentry
 // @description     Watch your Screenshots folder or any folder you pick, then copy, rename, or delete each new screenshot, or choose from a notification.
-// @version         0.18.5
+// @version         0.18.6
 // @author          mario0318
 // @github          https://github.com/mario0318
 // @include         windhawk.exe
@@ -59,11 +59,11 @@ While the popup is turned on, SnapSentry leaves two things on your machine so th
 notification buttons work, a SnapSentry shortcut in your Start Menu programs
 folder and one registry entry. Turning the popup off or disabling the mod removes
 both again. If the notification can't be shown, SnapSentry falls back
-to a standard dialog. A multi-page or animated image is kept rather than deleted,
-because only its first frame can go on the clipboard, and a short notice says so
-instead of the usual buttons. If you have turned its notifications off, it takes
+to a standard dialog. If you have turned its notifications off, it takes
 that as a cue to stay quiet: it still copies to the clipboard but shows no dialog
-and never auto deletes.
+and never auto deletes. A multi-page or animated image is kept rather than
+deleted, because only its first frame can go on the clipboard, and afterwards a
+short notice says so, when the notification is available.
 
 ## Privacy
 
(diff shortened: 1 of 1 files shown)


mario0318:
Moved the fix up to where the path is composed: LoadSettings now canonicalizes the folder once with GetFullPathNameW right after the env-var expansion, and Enqueue composes the child path with a JoinChild helper instead of an unconditional separator. RecycleFile is back to being only about recycling.

Also addressed from the optional list: removed the stale comment above the keep-notice call, added a teardown guard to ShowKeptToast (checks the stop event before showing, matching AskAction's pattern) and made it void since nothing uses the return value, and documented the new toast in both the settings description and the README.

Left as noted, not done here: the ShowToast/ShowKeptToast duplication, the pre-existing carry-overs, the flip-rotator timing question, and the pre-existing-folder blind spot are all real but out of scope for this change.

mario0318:
All three required items fixed:

1. Fixed the wording in both the README and the settings description: the notice appears after the action toast's buttons, not instead of them, and moved the README sentence out from under the dialog-fallback sentence so it doesn't read as describing the dialog.
2. Skip canonicalization when the folder is empty, so GetFullPathNameW can't turn an unresolved DefaultScreenshotsFolder() into the process's working directory.
3. Added the history clear by AUMID in the popup-off/unloading branch of SyncToastRegistration, before RemoveAumidRegistration/RemoveClsidRegistration, per the ordering note.

Also took two of the optional items since they were in the same area: SnapshotExistingNames now uses JoinChild instead of a bare concatenation, and Enqueue builds its path once.

Left for later, as before: the ShowToast/ShowKeptToast duplication, the flip-rotator timing question, the backoff staleness on recovery, and the pre-existing-name cleanup on delete/rename.

mario0318:
Fixed the ordering bug: RemoveNotificationSettings was deleting the per-AUMID registry entry before the clear ran, and the clear was also skipped entirely on the early-return path. Pulled it into ClearToastHistory and call it unconditionally at the top of the unload branch, before RemoveNotificationSettings, matching the ordering the block's own comment already claimed. It now logs on failure instead of dropping the HRESULT.

Took the rest of the optional list too:
- The env-var and path expansions use the two-call form so a long folder path doesn't silently skip normalization.
- ShowToast and ShowKeptToast now share their WinRT setup (activation factory, notifier, setting check, XML doc, toast object) through a BuildToast helper. ShowToast still needs to tell "notifications off" apart from "some other failure" to decide whether to fall back to the dialog, so BuildToast reports that through the setting out-param rather than folding both into one failure code.
- The four call sites extracting a file's base name go through one BaseName helper.

Left as noted before: the backoff not actually making unreachable-notification folders work, the accumulating kept-file toasts, and the flip-rotator's memory-peak assumption.

mario0318:
Took the optional list from this round:

- The clear-history log no longer cries wolf. The unload path already clears before RemoveNotificationSettings, so the second call at the bottom is skipped when unloading, and it is also skipped when nothing was ever registered (the default popup-off first pass), where there is no AUMID to resolve and the failure means nothing.
- Added a one-per-session clear on the first successful registration, so a notice left in Action Center by an unclean exit gets reclaimed. With the popup on, nothing else would ever clear it.
- Moved the two doc comments back onto the functions they describe after the BuildToast and ClearToastHistory splits, and dropped the using directives those splits left unused in ShowToast and ShowKeptToast.
- BuildToast now reports whether get_Setting succeeded through an explicit flag, so ShowToast no longer depends on a failed WinRT call zeroing the out-param to tell "notifications off" apart from "could not read".

Not changing, with reasons:

- The flip-rotator alternative (sequential CopyPixels plus in-place row reversal). The concern is fair and the suggested version may well be better, but the argument for it is about WIC internals and decode order, and I would rather not swap one unmeasured assumption for another on the strength of reasoning alone. Leaving the current code, which you have confirmed is correct either way, until it can be profiled against a large TIFF.
- The backoff and pre-existing-file interaction. Reclaiming files that land during a backoff means enqueueing at snapshot time, which trades directly against the never-touch-pre-existing-files invariant. Noted as a real cost of the 30s tier rather than something to change here.
- The remaining carry-overs from earlier rounds (g_preexisting not forgetting removed names, the rename sampling the foreground window in ProcessOne, the single OpenClipboard attempt) are pre-existing and out of scope for this change.

PR #5213, opened 2026-08-25
Add Desktop Audio Visualizer mod

NeiZqwe:
## Changelog

N/A (Initial release)

## Mod authorship

This mod was created by:

- [ ] Claude
- [x] ChatGPT
- [ ] Gemini
- [ ] Other (please specify):

mod provides its own desktop rendering and layout system, WASAPI-based audio capture, 1024-point FFT processing, 32 logarithmic frequency bands, up to 256 rendered bars, multiple visualization families, interpolation modes, dynamic bar geometry, Auto-Gain, attack/decay processing, optional CAVA-style smoothing, cursor interaction, foreground imagery, album-derived colors, configurable backgrounds, independent per-application audio source selection, and an integrated graphical 10-band equalizer. The mod also provides a configurable synchronized lyrics widget with its own rendering, positioning, typography, fallback behavior, album-derived backgrounds, and synchronization logic. These features are implemented as part of the mod's own desktop visualization layer rather than as cosmetic modifications to an existing component. The new equalizer is also integrated directly into the Windows 11 taskbar as a native XAML element, providing a dedicated user-facing control for the visualizer's audio processing. The EQ values are stored independently in Windhawk storage and are applied directly to the visualizer's FFT bands.
For these reasons, I believe maintaining this project as a standalone desktop-focused mod is justified. The project has its own audio-processing, rendering, layout, media metadata, and user-interface systems, with multiple features that depend directly on those systems.

The Lyrics widget is intended to provide automatically synchronized lyrics for the currently playing media session, the mod obtains the currently playing track's artist, title, album, duration, playback position, and playback state from the Windows Global System Media Transport Controls session. When a new track is detected, these metadata values are used to request the corresponding lyrics from LRCLIB.
LRCLIB is used specifically as the lyrics data source. It is not used for telemetry, analytics, remote configuration, advertising, file downloading, or any part of the audio visualization or FFT processing pipeline. The Lyrics feature is optional and disabled by default. When disabled, the mod does not need LRCLIB for any of its core visualization functionality, when enabled, the network access is narrowly scoped to lyrics retrieval for the currently playing track. The request uses the track metadata and duration to query LRCLIB's lyrics API, with small duration variations used only to tolerate minor metadata differences between the media session and the lyrics database. The retrieved data is then processed locally by the mod into the synchronized lyrics representation used by the desktop widget. The visualization engine itself does not depend on LRCLIB.
I understand the concerns around network access from explorer.exe. However, this access is required for the explicitly user-enabled purpose of the Lyrics widget and is isolated from the rest of the visualizer functionality.
For this reason, I would like to request that the LRCLIB access be considered as a case-by-case exception for this mod. The network request is directly tied to a user-facing feature, is disabled by default, has a narrowly defined purpose, and is not used for telemetry or any unrelated functionality.

### DIFF
--- a/mods/neiz-supersmile-audio-visualizer.wh.cpp
+++ b/mods/neiz-supersmile-audio-visualizer.wh.cpp
@@ -9096,6 +9096,20 @@ static void EqEnsureXamlButton() {
     if (!taskbar || !IsWindow(taskbar))
         return;
 
+    static ULONGLONG lastDeepCheckMs = 0;
+    const ULONGLONG nowMs = GetTickCount64();
+
+    // TrayUI::StartTaskbar is the primary immediate recovery path. The deep
+    // reconciliation below is only a periodic safety net for XAML changes that
+    // are not accompanied by that notification.
+    if (g_eqXamlInjectionReady.load(std::memory_order_acquire) &&
+        g_eqTaskbarHwnd == taskbar &&
+        nowMs - lastDeepCheckMs < 30000) {
+        return;
+    }
+
+    lastDeepCheckMs = nowMs;
+
     EqHideLegacyNativeButtons(taskbar);
 
     struct Payload { HWND taskbar; } payload{taskbar};
@@ -9135,6 +9149,7 @@ static void EqEnsureXamlButton() {
     }, &payload)) {
         Wh_Log(L"EqEnsureXamlButton: failed to marshal to taskbar UI thread");
     }
+
 }
 
 static void WINAPI EqTrayUIStartTaskbarHook(void* pThis) {
@@ -9166,68 +9181,40 @@ static bool EqHookTaskbarSymbols() {
     if (!h)
         return false;
 
-    // Primary-taskbar TaskbarHost access is required. Secondary-taskbar symbols
-    // are optional so a build can still provide the primary EQ button when the
-    // secondary taskbar ABI has changed.
-    // taskbar.dll
-    WindhawkUtils::SYMBOL_HOOK taskbarDllRequiredHooks[] = {
+    // Resolve all taskbar symbols in one call so the Windhawk symbol cache is
+    // shared across the complete taskbar integration. Secondary-taskbar,
+    // FrameHeight, and StartTaskbar symbols are optional across builds.
+    WindhawkUtils::SYMBOL_HOOK taskbarDllHooks[] = {
         {{LR"(const CTaskBand::`vftable'{for `ITaskListWndSite'})"},
          &g_eqCTaskBandTaskListWndSiteVftable},
         {{LR"(public: virtual class std::shared_ptr<class TaskbarHost> __cdecl CTaskBand::GetTaskbarHost(void)const )"},
          &g_eqCTaskBandGetTaskbarHost},
         {{LR"(public: void __cdecl std::_Ref_count_base::_Decref(void))"},
          &g_eqStdRefDecref},
-    };
-
-    if (!WindhawkUtils::HookSymbols(
-            h, taskbarDllRequiredHooks, ARRAYSIZE(taskbarDllRequiredHooks))) {
-        Wh_Log(L"EqHookTaskbarSymbols: primary taskbar symbols could not be resolved");
-        return false;
-    }
-
-    // taskbar.dll
-    WindhawkUtils::SYMBOL_HOOK taskbarDllSecondaryHooks[] = {
         {{LR"(const CSecondaryTaskBand::`vftable'{for `ITaskListWndSite'})"},
-         &g_eqCSecondaryTaskBandTaskListWndSiteVftable},
+         &g_eqCSecondaryTaskBandTaskListWndSiteVftable, nullptr, true},
         {{LR"(public: virtual class std::shared_ptr<class TaskbarHost> __cdecl CSecondaryTaskBand::GetTaskbarHost(void)const )"},
-         &g_eqCSecondaryTaskBandGetTaskbarHost},
-    };
-    if (!WindhawkUtils::HookSymbols(
-            h, taskbarDllSecondaryHooks, ARRAYSIZE(taskbarDllSecondaryHooks))) {
-        g_eqCSecondaryTaskBandTaskListWndSiteVftable = nullptr;
-        g_eqCSecondaryTaskBandGetTaskbarHost = nullptr;
-        Wh_Log(L"EqHookTaskbarSymbols: secondary taskbar symbols unavailable; continuing for primary taskbar");
-    }
-
-    // FrameHeight is needed to derive the TaskbarHost -> hosted XAML element
-    // offset on supported x64/ARM64 builds. This is the same mechanism used by
-    // the actively maintained Taskbar Fluent Media Player mod.
-    // taskbar.dll
-    WindhawkUtils::SYMBOL_HOOK taskbarDllFrameHeightHooks[] = {
+         &g_eqCSecondaryTaskBandGetTaskbarHost, nullptr, true},
         {{LR"(public: int __cdecl TaskbarHost::FrameHeight(void)const )"},
-         &g_eqTaskbarHostFrameHeight},
+         &g_eqTaskbarHostFrameHeight, nullptr, true},
+        {{LR"(public: virtual void __cdecl TrayUI::StartTaskbar(void))"},
+         &g_eqTrayUIStartTaskbarOriginal, EqTrayUIStartTaskbarHook, true},
     };
+
     if (!WindhawkUtils::HookSymbols(
-            h, taskbarDllFrameHeightHooks, ARRAYSIZE(taskbarDllFrameHeightHooks))) {
-        g_eqTaskbarHostFrameHeight = nullptr;
-        Wh_Log(L"EqHookTaskbarSymbols: TaskbarHost::FrameHeight not found");
-#if defined(_M_X64) || defined(__x86_64__) || defined(_M_ARM64) || defined(__aarch64__)
+            h, taskbarDllHooks, ARRAYSIZE(taskbarDllHooks))) {
+        Wh_Log(L"EqHookTaskbarSymbols: taskbar symbols could not be resolved");
         return false;
-#endif
     }
 
-    // Hook the taskbar rebuild path as an immediate reinjection trigger. Keep
-    // the periodic poll as a fallback for rebuilds not routed through this hook.
-    // taskbar.dll
-    WindhawkUtils::SYMBOL_HOOK taskbarDllTrayStartHooks[] = {
-        {{LR"(public: virtual void __cdecl TrayUI::StartTaskbar(void))"},
-         &g_eqTrayUIStartTaskbarOriginal,
-         EqTrayUIStartTaskbarHook},
-    };
-    if (!WindhawkUtils::HookSymbols(
-            h, taskbarDllTrayStartHooks, ARRAYSIZE(taskbarDllTrayStartHooks))) {
-        g_eqTrayUIStartTaskbarOriginal = nullptr;
-        Wh_Log(L"EqHookTaskbarSymbols: TrayUI::StartTaskbar not found; using polling fallback");
+    // The primary TaskbarHost symbols are required. FrameHeight is also
+    // required by EqGetTaskbarXamlRoot on supported x64/ARM64 builds.
+    if (!g_eqCTaskBandTaskListWndSiteVftable ||
+        !g_eqCTaskBandGetTaskbarHost ||
+        !g_eqStdRefDecref ||
+        !g_eqTaskbarHostFrameHeight) {
+        Wh_Log(L"EqHookTaskbarSymbols: required taskbar symbols are missing");
+        return false;
     }
 
     return true;
--- a/mods/neiz-supersmile-audio-visualizer.wh.cpp
+++ b/mods/neiz-supersmile-audio-visualizer.wh.cpp
@@ -9169,6 +9169,7 @@ static bool EqHookTaskbarSymbols() {
     // Primary-taskbar TaskbarHost access is required. Secondary-taskbar symbols
     // are optional so a build can still provide the primary EQ button when the
     // secondary taskbar ABI has changed.
+    // taskbar.dll
     WindhawkUtils::SYMBOL_HOOK taskbarDllRequiredHooks[] = {
         {{LR"(const CTaskBand::`vftable'{for `ITaskListWndSite'})"},
          &g_eqCTaskBandTaskListWndSiteVftable},
@@ -9184,6 +9185,7 @@ static bool EqHookTaskbarSymbols() {
         return false;
     }
 
+    // taskbar.dll
     WindhawkUtils::SYMBOL_HOOK taskbarDllSecondaryHooks[] = {
         {{LR"(const CSecondaryTaskBand::`vftable'{for `ITaskListWndSite'})"},
          &g_eqCSecondaryTaskBandTaskListWndSiteVftable},
@@ -9200,6 +9202,7 @@ static bool EqHookTaskbarSymbols() {
     // FrameHeight is needed to derive the TaskbarHost -> hosted XAML element
     // offset on supported x64/ARM64 builds. This is the same mechanism used by
     // the actively maintained Taskbar Fluent Media Player mod.
+    // taskbar.dll
     WindhawkUtils::SYMBOL_HOOK taskbarDllFrameHeightHooks[] = {
         {{LR"(public: int __cdecl TaskbarHost::FrameHeight(void)const )"},
          &g_eqTaskbarHostFrameHeight},
@@ -9215,6 +9218,7 @@ static bool EqHookTaskbarSymbols() {
 
     // Hook the taskbar rebuild path as an immediate reinjection trigger. Keep
     // the periodic poll as a fallback for rebuilds not routed through this hook.
+    // taskbar.dll
     WindhawkUtils::SYMBOL_HOOK taskbarDllTrayStartHooks[] = {
         {{LR"(public: virtual void __cdecl TrayUI::StartTaskbar(void))"},
          &g_eqTrayUIStartTaskbarOriginal,
--- a/mods/neiz-supersmile-audio-visualizer.wh.cpp
+++ b/mods/neiz-supersmile-audio-visualizer.wh.cpp
@@ -1,0 +1,7109 @@
+// ==WindhawkMod==
+// @id              neiz-supersmile-audio-visualizer
+// @name            Desktop Audio Visualizer
+// @description     A highly customizable desktop audio visualizer
+// @version         1.0.0
+// @author          NeiZ & SuperSmile123
+// @github          https://github.com/NeiZqwe
+// @include         explorer.exe
+// @compilerOptions -lole32 -luuid -lmmdevapi -lksuser -lgdi32 -luser32 -lgdiplus -ldwmapi -lruntimeobject -lwindowsapp -lwinhttp
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+**Desktop Audio Visualizer** — is a Windhawk mod that renders a highly customizable audio visualizer and synced lyrics widget directly on your Windows desktop.
+
+### Key Features:
+* **Real-time Audio Visualizer —** WASAPI loopback capture with customizable spectrum bars, shapes, and smooth rendering.
+* **Synced Lyrics Widget —** Automatically fetches and displays synced lyrics for the currently playing media session.
+* **Fluent & Custom Visuals —** Full control over colors, gradients, bar count, spacing, corner radius, and orientation.
+* **Desktop Integration —** Position the visualizer anywhere on your desktop with precise x/y coordinates and sizing.
+
+### Advanced Customization:
+* **Visualizer Engine —** Fine-tune FFT settings, bar shapes, interpolation modes, and audio sensitivity.
+* **Lyrics & Text —** Customize font style, size, active line highlights, and scroll responsiveness.
+* **Performance —** Lightweight WASAPI audio capture designed for low CPU and memory usage.
+
+### Credits:
+* **NeiZ & SuperSmile123** — Authors of Desktop Audio Visualizer.
+* **Salyts** — Author of [Taskbar Fluent Media Player](https://windhawk.net/mods/taskbar-fluent-media-player) (code references and inspiration).
+
+### Report a Bug:
+If you encounter any issues or have a feature suggestion, please open a report on the project's GitHub page: 👉 [Report an Issue on GitHub](https://github.com/NeiZqwe)
+*/
+// ==/WindhawkModReadme==
+
+// ==WindhawkModSettings==
+/*
+- Visualizer:
+    - barCount: 96
+      $name: Bar count
+      $name:ru-RU: Количество полос
+      $description: Number of bars displayed by the visualizer (0-256)
+      $description:ru-RU: Количество полос, отображаемых визуализатором (0-256)
+
+    - barWidth: 6
+      $name: Bar width
+      $name:ru-RU: Ширина полос
+      $description: Controls the width of each bar
+      $description:ru-RU: Определяет ширину каждой полосы
+
+    - barSpacing: 3
+      $name: Bar spacing
+      $name:ru-RU: Отступ между полосами
+      $description: Controls the spacing between bars
+      $description:ru-RU: Определяет расстояние между полосами
+
+    - orientation: bottom_up
+      $name: Orientation
+      $name:ru-RU: Направление
+      $description: Sets the direction in which the bars grow
+      $description:ru-RU: Определяет направление роста полос
+      $options:
+        - "bottom_up": Bottom up
+        - "center_vertical": Center vertical
+        - "top_down": Top down
+        - "left_right": Left to right
+        - "center_horizontal": Center horizontal
+        - "right_left": Right to left
+      $options:ru-RU:
+        - "bottom_up": Снизу вверх
+        - "center_vertical": Из центра вверх и вниз
+        - "top_down": Сверху вниз
+        - "left_right": Слева направо
+        - "center_horizontal": Из центра в стороны
+        - "right_left": Справа налево
+
+    - interpolationMode: smooth
+      $name: Interpolation
+      $name:ru-RU: Интерполяция
+      $description: Smooths the transition between neighboring bars
+      $description:ru-RU: Сглаживает переходы между соседними полосами
+      $options:
+        - smooth: Curve
+        - step: Step
+      $options:ru-RU:
+        - smooth: Кривая
+        - step: Ступенчатая
+
+    - barShape: stereo
+      $name: Visualization Type
+      $name:ru-RU: Тип визуализации
+      $description: Selects how the audio spectrum is arranged and displayed
+      $description:ru-RU: Определяет, как аудиоспектр располагается и отображается
+      $options:
+        - stereo: Stereo
+        - mountain: Mountain
+        - mirror: Mirror
+        - wave: Wave
+        - circular: Circular
+      $options:ru-RU:
+        - stereo: Стерео
+        - mountain: Гора
+        - mirror: Зеркало
+        - wave: Волна
+        - circular: Круговая
+
+    - mirroredVisualizer: false
+      $name: Mirrored visualizer
+      $name:ru-RU: Зеркальный визуализатор
+      $description: Shows a mirrored copy of the visualizer on the opposite side of the screen
+      $description:ru-RU: Показывает зеркальную копию визуализатора на противоположной стороне экрана
+
+    - Circular:
+        - circleRadius: 250
+          $name: Radius (px)
+          $name:ru-RU: Радиус (px)
+
+        - circleStartAngle: -90
+          $name: Start angle (°)
+          $name:ru-RU: Начальный угол (°)
+          $description: Sets the starting angle of the circular visualization. -90° = top, 0° = right, 90° = bottom
+          $description:ru-RU: Задаёт начальный угол кругового визуализатора. -90° = сверху, 0° = справа, 90° = снизу
+      $name: Circular type settings
+      $name:ru-RU: Настройки кругового визуализатора
+    - positionX: 50
+      $name: X position
+      $description: Adjusts the horizontal position of the visualizer
+      $name:ru-RU: X позиция
+      $description:ru-RU: Изменяет горизонтальное положение визуализатора
+    - positionY: 900
+      $name: Y position
+      $description: Adjusts the vertical position of the visualizer
+      $name:ru-RU: Y позиция
+      $description:ru-RU: Изменяет вертикальное положение визуализатора
+    - maxBarHeight: 200
+      $name: Max bar height
+      $description: Sets the maximum height the bars can reach
+      $name:ru-RU: Максимальная высота полосы
+      $description:ru-RU: Задаёт максимальную высоту, которой могут достигать полосы
+    - minBarHeight: 6
+      $name: Min bar height
+      $description: Sets the resting height of the bars when there is little or no audio
+      $name:ru-RU: Минимальная высота полосы
+      $description:ru-RU: Задаёт высоту полос в состоянии покоя при отсутствии или низком уровне звука
+
+  $name: Visualizer
+  $name:ru-RU: Визуализатор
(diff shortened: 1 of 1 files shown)


NeiZqwe:
> ### Submission review
> 
> 
> 
> 
> A lot of care clearly went into this — the settings block is fully in sync with the code, every worker thread has a stop event and is joined in `Wh_ModUninit`, the window class is unregistered, and `GetCurrentModModuleHandle` correctly uses `GET_MODULE_HANDLE_EX_FLAG_UNCHANGED_REFCOUNT`. The items below are what stands out.
> 
> **1. Network access to LRCLIB is a blocking policy issue, not just a documentation matter.**
> 
> Windhawk mods must be fully self-contained: a mod must not depend on, download, or contact external servers. `HttpGetUtf8` (line 4980) opens a WinHTTP session to `lrclib.net` (line 5343) from inside `explorer.exe`, and `-lwinhttp` is in `@compilerOptions`. Your PR description asks for a case-by-case exception, which is fair to raise — but that is the maintainer's call, not something this review can grant, so treat the lyrics feature as unresolved until he weighs in. Practically, the two outcomes are: the exception is granted (in which case the README disclosure you offered is the right follow-up), or the lyrics widget is dropped/reduced to metadata that the system media session already provides locally (artist/title/position, no network). It would help to say explicitly in the PR which of the two you'd go with if the exception is declined, so the rest of the mod isn't blocked on it.
> 
> **2. Substantial overlap with the existing Desktop Audio Visualizer mod.**
> 
> [mods/desktop-audio-visualizer.wh.cpp](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/desktop-audio-visualizer.wh.cpp) (by Salyts) is already merged and covers the same core product: WASAPI loopback capture, FFT, bar/mountain/mirror/wave/dots shapes, solid/gradient/accent/album-art/acrylic color modes, background panel with wallpaper blur, EQ presets, sensitivity, position control, per-corner radius, frame-rate control. Your README credits Salyts as the "reference / upstream desktop audio visualizer implementation" and GR0UD for the "upstream audio capture / FFT implementation reference", and the mod is named "Desktop Audio Visualizer **Plus**" — so this reads as an extended fork rather than an independent mod.
> 
> The maintainer's consistently stated preference is to extend an existing mod (add options, or PR the original author's repo) rather than merge a near-duplicate, because duplicates fragment the catalog and confuse users. Two things would help move this forward:
> 
>     * State concretely what is genuinely new (from what I can see: per-application audio source, circular/area shapes, extra bar styles, interpolation modes, CAVA smoothing, auto-gain, cursor interaction, foreground image, lyrics) versus what already exists, and why those can't land as options in the existing mod.
> 
>     * Clarify the code provenance. "Reference implementation" is ambiguous — if code was copied from the existing mod, say so plainly, and note that neither file currently carries a `@license`, which should be sorted out before merge if there's derived code.
> 
> 
> **3. The overlay repaints the entire virtual desktop every frame, even at rest.**
> 
> The window is created at full virtual-screen size (lines 7901-7906) and `RenderOverlay` (line 7588) does, on every tick:
> 
>     * `memset` of the whole `w*h*4` DIB (line 7601) — ~33 MB on a single 4K monitor, more on multi-monitor,
> 
>     * antialiased GDI+ drawing,
> 
>     * `UpdateLayeredWindow` with `size = {w, h}` (line 7665), i.e. a full-desktop layered-surface upload.
> 
> 
> At 60 FPS that is several GB/s of memory traffic for a visualizer that occupies a few hundred pixels, and it doesn't stop when there's no audio — the idle path still runs the same full-desktop repaint 5×/s (line 7695). It also matters more than usual here: the overlay is a `WS_CHILD` of `SHELLDLL_DefView`, which is owned by another thread, so the two threads' input queues are attached and slowness on the render thread is felt by the desktop.
> 
> Concrete fix: size the layered window to the visualizer's bounding box (you already compute one in `GetVisualizerBackgroundRect`, line 6394) plus the lyrics widget rect, instead of the whole desktop — and reposition it when the settings or the layout change. If you want to keep the full-size window, at minimum pass a dirty rectangle via `UpdateLayeredWindowIndirect` (`UPDATELAYEREDWINDOWINFO::prcDirty`) and `memset` only that region.
> 
> **4. `EnumDisplaySettingsW` runs on every rendered frame.**
> 
> `OverlayProc`'s `WM_TIMER` handler calls `SetTimer(hwnd, 1, GetRenderIntervalMs(g_settings), nullptr)` on _every_ tick (lines 7695-7700), and `GetRenderIntervalMs` (line 7786) calls `GetDisplayRefreshRateHz` → `MonitorFromWindow` + `GetMonitorInfoW` + `EnumDisplaySettingsW` (line 7761). `EnumDisplaySettingsW` queries the display driver; doing it 60+ times per second is a needless per-frame cost, and re-arming the timer every tick isn't needed either.
> 
> Cache the refresh rate in a variable, refresh it in the `WM_DISPLAYCHANGE` / `WM_SETTINGCHANGE` handlers you already have (lines 7728-7744), and only call `SetTimer` when the computed interval actually changes:
> 
> ```c++
> static UINT g_cachedRefreshHz = 0;  // invalidated on WM_DISPLAYCHANGE
> static UINT g_currentTimerMs = 0;
> ...
> const UINT intervalMs = shouldIdle ? 200 : GetRenderIntervalMs(g_settings);
> if (intervalMs != g_currentTimerMs) {
>     g_currentTimerMs = intervalMs;
>     SetTimer(hwnd, 1, intervalMs, nullptr);
> }
> ```
> 
> **5. `g_lyricsFontCache` is a global whose destructor calls into GDI+ at process shutdown.**
> 
> `static LyricsFontCache g_lyricsFontCache;` (line 5851) holds `std::unique_ptr<Gdiplus::Font>` / `Gdiplus::FontFamily`. `Wh_ModUninit` runs on a normal unload, but **not** when Explorer itself exits (sign-out, reboot, "Exit Explorer"). In that path every other thread has already been terminated, `GdiplusShutdown` was never called, and the CRT then runs this global's destructor alone on the shutdown thread under the loader lock — calling `GdipDeleteFont`/`GdipDeleteFontFamily` into a subsystem that may already be gone. This is exactly the case described in [Global objects and process shutdown](https://github.com/ramensoftware/windhawk/wiki/Global-objects-and-process-shutdown#7-raii-wrappers-that-call-teardown-apis).
> 
> Suppress the automatic destructor and keep the explicit release you already have at line 8044:
> 
> ```c++
> // GDI+ font objects must not be destroyed by the automatic destructor at
> // process shutdown; released via reset() on the overlay thread.
> [[clang::no_destroy]] static std::optional<LyricsFontCache> g_lyricsFontCache;
> ```
> 
> `LyricsFontCache` isn't a nullable/handle-like type, so use the `std::optional` wrapper (accesses become `->`, and `g_lyricsFontCache.Clear()` at line 8044 becomes `g_lyricsFontCache.reset()`, which fully destroys the fonts rather than just clearing fields). The other globals in the file are fine as-is — `std::wstring`/`shared_ptr`/`std::mutex`/`HANDLE` all have heap-only or no-op destructors and need no attribute.
> 
> **6. Two background threads poll the system media session continuously, regardless of settings.**
> 
>     * `StartAlbumColorCapture()` is called unconditionally from `Wh_ModInit` (line 8081), and `AlbumColorThreadProc` then queries `GetCurrentSession()` / `TryGetMediaPropertiesAsync()` / opens the thumbnail stream every 700 ms forever (line 4874) — even when `colorMode` isn't `solid_album`/`gradient_album`, the background isn't album-based, and no border uses album colors. These are cross-process WinRT calls in `explorer.exe`.
> 
>     * `LyricsThreadProc` polls the same session manager every 100 ms (line 5740), which is a lot for metadata that changes once per track.
> 
> 
> Gate the album thread on whether any album-derived color is actually selected (you already have `IsAlbumColorMode()` at line 4143 — currently unused; it needs extending to cover the background/border album modes), start/stop it from `Wh_ModSettingsChanged`, and raise the lyrics poll interval substantially. Better still, `GlobalSystemMediaTransportControlsSessionManager` is event-driven: `CurrentSessionChanged`, and per-session `MediaPropertiesChanged` / `PlaybackInfoChanged` / `TimelinePropertiesChanged` remove the need for a metadata poll entirely (keep a slow timer only for the playback-position interpolation). Remember to revoke the event tokens before joining the thread.
> 
> **7. `Wh_ModUninit` / a settings change can block for many seconds inside WinHTTP.**
> 
> `StopLyricsCapture` waits `INFINITE` on the lyrics thread (line 5789), but `HttpGetUtf8` only checks the stop event _between_ WinHTTP operations (lines 4997, 5036, 5051) — never during one. Each request has four 1000 ms timeouts (line 4994), and `FetchLyricsForTrack` issues up to five sequential requests for the ±2s duration variants (line 5357). Worst case that's ~15-20 s during which disabling the mod, updating it, or just toggling a setting appears to hang.
> 
> Make the stop signal actually cancel the in-flight request: keep the `HINTERNET` request handle in a global (guarded by a small mutex) and call `WinHttpCloseHandle` on it from `StopLyricsCapture` before waiting — a closed handle causes the blocking call to fail immediately. Also bail out of the duration-candidate loop as soon as the stop event is set (it's checked at line 5375, but only before each attempt, which is right — the problem is the attempt itself).
> 
> **8. `CreateToolhelp32Snapshot` every 500 ms when the selected application isn't playing.**
> 
> With `audioSource: application`, the audio thread calls `FindAudioProcessIdByExecutable` every 500 ms (line 2679). If the app has no audio session — which is the normal state when it isn't playing — the function falls through to a full process snapshot (line 2314) on every one of those calls. That's a heavyweight system-wide enumeration twice a second, indefinitely. Back off the interval when nothing was found (e.g. exponential up to a few seconds), and consider dropping the snapshot fallback entirely: a process with no audio session can't be captured with process loopback anyway.
> 
> **9. The mod installs no hooks at all — please justify running inside `explorer.exe`.**
> 
> There isn't a single `Wh_SetFunctionHook` / `HookSymbols` call in the file; everything the mod does (`FindWindow`, `CreateWindowEx`, WASAPI, WinRT SMTC, GDI+, WinHTTP, `GetCursorPos`) works from any process. That's the standard signal for the ["Mods as tools: Running mods in a dedicated process"](https://github.com/ramensoftware/windhawk/wiki/Mods-as-tools:-Running-mods-in-a-dedicated-process) pattern, which the maintainer asks for frequently — the argument being blast radius: an 8000-line mod doing GDI+ rendering, FFT and HTTP inside the shell will take Explorer down with it if it faults or hangs.
> 
> The counter-argument here is real — the overlay is created as a child of `SHELLDLL_DefView`, and being in-process makes that simpler — but cross-process parenting to the desktop window is what standalone wallpaper apps do, so it isn't a hard blocker. Either move to the tool-mod pattern (`@include windhawk.exe`, `WhTool_ModInit`/`WhTool_ModSettingsChanged`/`WhTool_ModUninit`, launcher boilerplate copied **verbatim** from the wiki — see [mods/explorer-folder-hover-menu.wh.cpp](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/explorer-folder-hover-menu.wh.cpp) for a working example), or explain in the PR why in-process injection is required.
> Optional improvements
> 
> 
> Functionality notes
> 
> 
> **Next steps:**
> 
>     * `/ai-review` - after pushing fixes, to get a review of the updated code. You can repeat this as many times as you need, but each review is thorough and usually there's no need for more than 2-3 iterations.
> 
>     * `/ready-for-reviewer` - once you're satisfied with the state of the pull request, to hand it over to a human reviewer. If some findings above are left unaddressed, add a short note explaining why.
> 
> 

Thanks for the breakdown, I am currently working on the fixes for the performance and lifetime issues (points 3–8)
Regarding points 1, 2, and 9, here is my architectural and policy rationale while I finish up the implementation:
1. LRCLIB / Network Access
I am aware of the network policy, and I am requesting a manual review by the maintainer for this specific case. The HTTP requests are strictly read-only calls to LRCLIB's public API to fetch synced LRC lyrics for the currently playing track, this feature physically cannot function offline.
2. This is not a fork
This is a fundamental rewrite, not a simple fork of existing visualizer mods. It incorporates a custom 32-band logarithmic FFT audio processing engine with dynamic auto-gain, per-application WASAPI loopback capture (targeting specific process PIDs), dynamic background paneling with wallpaper blur, interactive cursor physics, and advanced visual modes. Merging these into the upstream mod would require completely rewriting almost its entire codebase anyway.
3. In-process injection over Tool Mod
The overlay window is intentionally created as a direct `WS_CHILD` of `SHELLDLL_DefView`. This is a deliberate design choice to guarantee seamless Z-order placement (sitting strictly under desktop icons and directly above wallpaper elements). Cross-process parenting (`SetParent` from a standalone tool process) introduces window clipping bugs, flickering during Explorer restarts, and inconsistent focus/render state handling.

NeiZqwe:
idk if i need to write this here aswell lol

mod provides its own desktop rendering and layout system, WASAPI-based audio capture, 1024-point FFT processing, 32 logarithmic frequency bands, up to 256 rendered bars, multiple visualization families, interpolation modes, dynamic bar geometry, Auto-Gain, attack/decay processing, optional CAVA-style smoothing, cursor interaction, foreground imagery, album-derived colors, configurable backgrounds, independent per-application audio source selection, and an integrated graphical 10-band equalizer. The mod also provides a configurable synchronized lyrics widget with its own rendering, positioning, typography, fallback behavior, album-derived backgrounds, and synchronization logic. These features are implemented as part of the mod's own desktop visualization layer rather than as cosmetic modifications to an existing component. The new equalizer is also integrated directly into the Windows 11 taskbar as a native XAML element, providing a dedicated user-facing control for the visualizer's audio processing. The EQ values are stored independently in Windhawk storage and are applied directly to the visualizer's FFT bands.
For these reasons, I believe maintaining this project as a standalone desktop-focused mod is justified. The project has its own audio-processing, rendering, layout, media metadata, and user-interface systems, with multiple features that depend directly on those systems.

The Lyrics widget is intended to provide automatically synchronized lyrics for the currently playing media session, the mod obtains the currently playing track's artist, title, album, duration, playback position, and playback state from the Windows Global System Media Transport Controls session. When a new track is detected, these metadata values are used to request the corresponding lyrics from LRCLIB.
LRCLIB is used specifically as the lyrics data source. It is not used for telemetry, analytics, remote configuration, advertising, file downloading, or any part of the audio visualization or FFT processing pipeline. The Lyrics feature is optional and disabled by default. When disabled, the mod does not need LRCLIB for any of its core visualization functionality, when enabled, the network access is narrowly scoped to lyrics retrieval for the currently playing track. The request uses the track metadata and duration to query LRCLIB's lyrics API, with small duration variations used only to tolerate minor metadata differences between the media session and the lyrics database. The retrieved data is then processed locally by the mod into the synchronized lyrics representation used by the desktop widget. The visualization engine itself does not depend on LRCLIB.
I understand the concerns around network access from explorer.exe. However, this access is required for the explicitly user-enabled purpose of the Lyrics widget and is isolated from the rest of the visualizer functionality.
For this reason, I would like to request that the LRCLIB access be considered as a case-by-case exception for this mod. The network request is directly tied to a user-facing feature, is disabled by default, has a narrowly defined purpose, and is not used for telemetry or any unrelated functionality.

NeiZqwe:
/ready-for-reviewer

Point 8
I did not apply this fix because the proposed 2-second wait does not fully solve the lifetime issue, if the async activation callback completes after the timeout, the callback may still execute code from the unloaded mod dll, I did not want to introduce a partial fix that could leave the same unload-time crash risk in place

Point 1
I intentionally kept LRCLIB because it is required for the explicitly user-enabled lyrics feature. The feature is disabled by default, and when enabled, network access is limited to retrieving synchronized lyrics for the current track. It is not used for telemetry, analytics, remote configuration, advertising, or any other part of the mod.
I understand that this is a policy question rather than a code defect, so I left it for maintainer consideration as a case-by-case exception
Point 2
I intentionally kept this as a standalone mod because it is not just a fork with additional cosmetic options. It has its own rendering, layout, audio-processing, media metadata, lyrics, and UI systems, including per-application audio capture and a custom eq
The overlap is mainly in the underlying visualizer concept and some inherited/reference functionality; the project has since grown into a substantially broader feature set with its own architecture and user-facing systems. Consolidating it into the upstream mod would therefore require restructuring a significant portion of the project rather than simply adding a few options.

NeiZqwe:
/ready-for-reviewer

Point 1
I intentionally kept LRCLIB because it is required for the explicitly user-enabled lyrics feature. The feature is disabled by default, and when enabled, network access is limited to retrieving synchronized lyrics for the current track. It is not used for telemetry, analytics, remote configuration, advertising, or any other part of the mod.
I understand that this is a policy question rather than a code defect, so I left it for maintainer consideration as a case-by-case exception
Point 2
I intentionally kept this as a standalone mod because it is not just a fork with additional cosmetic options. It has its own rendering, layout, audio-processing, media metadata, lyrics, and UI systems, including per-application audio capture and a custom eq
The overlap is mainly in the underlying visualizer concept and some inherited/reference functionality; the project has since grown into a substantially broader feature set with its own architecture and user-facing systems. Consolidating it into the upstream mod would therefore require restructuring a significant portion of the project rather than simply adding a few options

Point 3: Partial fixes with timeouts still leave unload crash risks, so keeping the existing teardown was safer to avoid regressions
Point 4: Window class unregistration issues only happen during rapid reload loops in dev. Normal user sessions are unaffected
Point 5: Full-screen clearing prevents ghosting across complex rendering passes
Point 7: Sending 0x052C unconditionally ensures Z-order consistency across third-party shell tweaks and wallpaper engines
Point 6: The mod strictly targets modern win 11 builds for the initial release, though win 10 fallback may be added in a future update

m417z:
> Point 6: The mod strictly targets modern win 11 builds for the initial release, though win 10 fallback may be added in a future update

Fix the comment according to the review.

mods/neiz-supersmile-audio-visualizer.wh.cpp
```diff
@@ -0,0 +1,9571 @@
...
+            barCrossPos < lyricsX + lyricsW &&
+            barCrossPos + barThickness > lyricsX;
+        if (!overlapsX)
+            return barLen;
+
+        const int widgetTop = lyricsY;
+        const int widgetBottom = lyricsY + lyricsH;
+        const int baseY = g_settings.positionY;
+
+        i
```
m417z: This shouldn't depend on whether `DWMWA_WINDOW_CORNER_PREFERENCE` is defined. Either make sure it's always defined, or remove this code.

m417z, line 8322: What is that integration? The mod description says nothing about the taskbar.

NeiZqwe: It's for the eq button integrated into the taskbar. I can update the description and readme to mention the taskbar eq integration

m417z: I enabled the mod and I don't see a button on the taskbar. I think it's worth adding it to the readme with a screenshot, and also let's figure out why it doesn't work.

I'm testing on 26200.9278 which got some taskbar layout updates, maybe that's related.

m417z:
I tried the lyrics feature with some very popular videos on YouTube, and couldn't get it to work. Example:
https://lrclib.net/api/get?artist_name=EminemVEVO&track_name=Eminem%20-%20Mockingbird%20%5BOfficial%20Music%20Video%5D&album_name=&duration=258

I'm not familiar with LRCLIB, but perhaps it's possible to improve the experience.

mods/neiz-supersmile-audio-visualizer.wh.cpp
NeiZqwe: removed the #ifdef-dependent

NeiZqwe:
> I tried the lyrics feature with some very popular videos on YouTube, and couldn't get it to work. Example: https://lrclib.net/api/get?artist_name=EminemVEVO&track_name=Eminem%20-%20Mockingbird%20%5BOfficial%20Music%20Video%5D&album_name=&duration=258
> 
> I'm not familiar with LRCLIB, but perhaps it's possible to improve the experience.

I think lrclib's /api/get endpoint requires an exact string match and returns 404 when track titles or artist names contain extra tags or suffixes ( artist_name=EminemVEVO )
In theory, I can sanitize artist and title strings to strip extra suffixes or fallback from /api/get to /api/search?q=..., but doing that might introduce new edge-case bugs
Rn the lyrics feature is primarily intended for desktop media players (Spotify, AIMP, Tidal, etc.) rather than browser/YouTube playback, where metadata can be messy, I’d prefer to address this alongside a couple of other known bugs and planned additions in a follow-up update after the initial release, to keep the current submission stable

NeiZqwe:
> > Point 6: The mod strictly targets modern win 11 builds for the initial release, though win 10 fallback may be added in a future update
> 
> Fix the comment according to the review.

done

mods/neiz-supersmile-audio-visualizer.wh.cpp
NeiZqwe: I'm on Windows 11 25H2 build 26200.9168, and the taskbar EQ button works correctly on my system.
The integration depends on several internal taskbar.dll symbols, so the newer 26200.9278 build may have changed some of those symbols. Could you check the Windhawk log and see whether EqHookTaskbarSymbols() is failing on your build?

In particular, please check whether this message appears:
Wh_ModInit: taskbar XAML EQ integration hooks could not be installed; visualizer will continue without EQ button

NeiZqwe: > I think it's worth adding it to the readme with a screenshot, and also let's figure out why it doesn't work.


The very first screenshot in the details/preview section of the mod actually shows the equalizer UI along with the taskbar button itself

m417z: > In particular, please check whether this message appears:
> Wh_ModInit: taskbar XAML EQ integration hooks could not be installed; visualizer will continue without EQ button

No, I see no logs at all. Have you tried it on a more standard Windows 11 configuration with the default system icons and no language icon?

NeiZqwe:
/ai-review

Addressed everything requested after the maintainer's review, regarding the lyrics display issue on yt, I'll defer that LRCLIB improvement to a future version to avoid introducing any bugs right now

NeiZqwe:
/ai-review

Addressed everything requested after the maintainer's review, regarding the lyrics display issue on yt, I'll defer that LRCLIB improvement to a future version to avoid introducing any bugs right now

NeiZqwe:
/ready-for-reviewer

1. This is a valid edge-case teardown race. I prioritized getting the feature working across current Windows 11 builds first; the activation lifetime is isolated from the taskbar integration and will be addressed separately.
2. The normal cleanup path already removes the button from its stored parent and unregisters the click handler. Handling abandoned XAML elements from a failed/unusual unload is additional defensive cleanup, not part of the normal injection path.
3.  The Grid path already restores shifted columns and removes the inserted column during cleanup. The StackPanel path does not add columns, so there is nothing to restore there.
4. The mod is scoped to explorer.exe and the existing shell-parent detection prevents the overlay from attaching to unrelated Explorer instances. 
5. This is intentional: the reconciliation runs once per second as a recovery mechanism for taskbar/XAML rebuilds that can happen without changing the taskbar HWND. The cost is negligible compared with losing the injected button.
6. I split the symbol groups to keep required and optional taskbar symbols independent across Windows builds. Consolidating them into one array is possible, but it does not affect the functionality and is mainly an implementation simplification.
7, 8. Known bugs that will be fixed in future update
9. This is a deliberate correctness trade-off. Full clearing prevents persistent ghosting when geometry changes; reducing the surface size is a possible future optimization, but it adds coordinate and lifecycle complexity.

m417z:
4. That's a valid point, you can create extra threads only after the taskbar is created.
5. Why would a taskbar/XAML rebuild happen?
6. It affects the caching as the review explains, please use a single array.

NeiZqwe:
> 4. That's a valid point, you can create extra threads only after the taskbar is created.
> 
>     5. Why would a taskbar/XAML rebuild happen?
> 
>     6. It affects the caching as the review explains, please use a single array.

5. It's a defensive fallback to keep the injection self-healing if the XAML tree changes while the taskbar HWND remains valid.
6. Ill update pr in moment

NeiZqwe:
> @NeiZqwe `/ready-for-reviewer` can't be applied here: the most recent AI review covers [f5ca5ed](https://github.com/ramensoftware/windhawk-mods/commit/f5ca5ede6fdcc565411cbd49cedb80689d3754d3), but the current head of this pull request is [15ee560](https://github.com/ramensoftware/windhawk-mods/commit/15ee560acc9065379eb34187c7148e80ae3b290f). Comment `/ai-review` to get a review of the current code.

:(

m417z:
I don't think the code was changed.

> 1. It's a defensive fallback to keep the injection self-healing if the XAML tree changes while the taskbar HWND remains valid.

What does it defend against? Did you see it happen? If not, perhaps it's not worth spinning a timer for nothing.

NeiZqwe:
> I don't think the code was changed.


my bad


> What does it defend against? Did you see it happen? If not, perhaps it's not worth spinning a timer for nothing.

I haven't observed or reproduced such a case. I originally kept the check as a defensive fallback, but yeah, without a reproducible scenario it wasn't justified to run it every second. I've changed it accordingly

NeiZqwe:
/ai-review

fixed 5&6

NeiZqwe:
/ai-review

3. Tested directly in runtime (toggling mid-session, hidden desktop icons): the button injects instantly without delays. The current pipeline functions reliably as-is.
4. Tested in multi-process Explorer mode as well: FindWindowW resolves correctly and causes zero issues in practice. No need to complicate it with EnumWindows.
5. Tested directly (Solid bars with Album-colored border): colors extract and apply properly. Works fine at runtime.
6. Background overhead in secondary explorer.exe processes is completely negligible. Restructuring the worker lifecycle adds unnecessary complexity.
7. Tested dirty-rect clearing, but ghosting artifacts persist due to layered alpha blending. Full DIB clear is strictly required to prevent visual artifacts.
8. Fixed AND tested.

m417z:
> Tested in multi-process Explorer mode as well: FindWindowW resolves correctly and causes zero issues in practice. No need to complicate it with EnumWindows.

The reason many existing mods use `EnumWindows` is that some customization tools create a fake `Shell_TrayWnd` window that can break the mod. Fixing it shouldn't be complicated, just copy the function from one of the mods.

https://github.com/ramensoftware/windhawk-mods/issues/1704

Example:
https://github.com/ramensoftware/windhawk-mods/blob/f2eedecca227cd1f5c0d9fd3e95822dddd00fb8f/mods/taskbar-reorder-right-drag.wh.cpp#L127-L146

NeiZqwe:
> > Tested in multi-process Explorer mode as well: FindWindowW resolves correctly and causes zero issues in practice. No need to complicate it with EnumWindows.
> 
> The reason many existing mods use `EnumWindows` is that some customization tools create a fake `Shell_TrayWnd` window that can break the mod. Fixing it should be complicated, just copy the function from one of the mods.
> 
> #1704
> 
> Example:
> 
> https://github.com/ramensoftware/windhawk-mods/blob/f2eedecca227cd1f5c0d9fd3e95822dddd00fb8f/mods/taskbar-reorder-right-drag.wh.cpp#L127-L146

Thanks for explaining, didn't know about that

NeiZqwe:
/ai-review


To be honest, I’m going to stop chasing every theoretical issue raised by the AI review for this release.

With every new review, the model tends to find another set of theoretical edge cases, even after the previous ones have been addressed. On top of that, some of the findings have turned out to be inaccurate or based on assumptions that don’t match the actual implementation. This makes the review process somewhat open-ended: fixing one real issue means pushing another revision, spending another review, and starting the cycle again, while the AI review is limited to only three runs per day.

I also want to clarify something about my background, because I think it is relevant here: I’m not a professional programmer or an experienced Windows/C++ developer. I’ve only been working in this area relatively recently, so there are absolutely things I may misunderstand, overlook, or simply not know yet. I’m not claiming that my implementation is technically perfect, and I’m fully aware that a more experienced developer may spot things I don’t.

That said, I’ve spent a lot of time repeatedly testing the issues the review keeps raising, including Explorer restarts, repeated mod reloads, settings changes, taskbar/XAML rebuilds and other stress scenarios. In practice, the mod has been stable on my side and on the contributor’s side: no Explorer crashes, no Windows crashes, no critical failures, and the UI starts and behaves as intended. It has also been tested across different PCs, Windows versions/builds and even older hardware without noticeable performance regressions. The only intentional limitation is the XAML EQ button on Windows 10.

Some of the review comments are valid from a theoretical/code-analysis perspective, but I don’t consider all of them release-blocking issues. In several cases, previous attempts to address theoretical edge cases introduced more noticeable practical regressions, such as visual ghosting, additional UI/update latency, or other behavior that was actually visible to users.

For the first release, I would rather keep the implementation stable and predictable instead of continuing an endless cycle of “fix one theoretical issue → rerun AI review → discover another theoretical issue → introduce another regression”.

There is already a known backlog of improvements and edge cases that I intend to address later. This release is meant to be the first working baseline. A larger follow-up update is planned with broader UI work, optimization and cleanup, where these deeper architectural issues can be handled properly rather than patched one by one under review pressure.

I also don’t want to put an artificial deadline on that next update just to be able to give an estimate for when every theoretical issue will be addressed. I would rather take the time necessary to make the next version genuinely good, so I can’t honestly promise a specific timeframe for those fixes.

At the same time, if the mod is released and an actual user reports a bug that they can reproduce in practice, that is a different situation. I would obviously prioritize and investigate a real user-facing bug as soon as possible. I’m much more interested in fixing something that an actual user can reproduce than in endlessly optimizing against hypothetical scenarios that have not manifested in practice.

So for this release, I’m prioritizing practical stability and the intended user experience over eliminating every theoretical edge case identified by static/AI review. I’m not claiming that the code is perfect or that there are no remaining edge cases; I’m simply choosing not to block the initial release on issues that have not manifested in real-world testing, especially when addressing them immediately can create more visible regressions than the issues themselves.

NeiZqwe:
/ready-for-reviewer



To be honest, I’m going to stop chasing every theoretical issue raised by the AI review for this release.

With every new review, the model tends to find another set of theoretical edge cases, even after the previous ones have been addressed. On top of that, some of the findings have turned out to be inaccurate or based on assumptions that don’t match the actual implementation. This makes the review process somewhat open-ended: fixing one real issue means pushing another revision, spending another review, and starting the cycle again, while the AI review is limited to only three runs per day.

I also want to clarify something about my background, because I think it is relevant here: I’m not a professional programmer or an experienced Windows/C++ developer. I’ve only been working in this area relatively recently, so there are absolutely things I may misunderstand, overlook, or simply not know yet. I’m not claiming that my implementation is technically perfect, and I’m fully aware that a more experienced developer may spot things I don’t.

That said, I’ve spent a lot of time repeatedly testing the issues the review keeps raising, including Explorer restarts, repeated mod reloads, settings changes, taskbar/XAML rebuilds and other stress scenarios. In practice, the mod has been stable on my side and on the contributor’s side: no Explorer crashes, no Windows crashes, no critical failures, and the UI starts and behaves as intended. It has also been tested across different PCs, Windows versions/builds and even older hardware without noticeable performance regressions. The only intentional limitation is the XAML EQ button on Windows 10.

Some of the review comments are valid from a theoretical/code-analysis perspective, but I don’t consider all of them release-blocking issues. In several cases, previous attempts to address theoretical edge cases introduced more noticeable practical regressions, such as visual ghosting, additional UI/update latency, or other behavior that was actually visible to users.

For the first release, I would rather keep the implementation stable and predictable instead of continuing an endless cycle of “fix one theoretical issue → rerun AI review → discover another theoretical issue → introduce another regression”.

There is already a known backlog of improvements and edge cases that I intend to address later. This release is meant to be the first working baseline. A larger follow-up update is planned with broader UI work, optimization and cleanup, where these deeper architectural issues can be handled properly rather than patched one by one under review pressure.

I also don’t want to put an artificial deadline on that next update just to be able to give an estimate for when every theoretical issue will be addressed. I would rather take the time necessary to make the next version genuinely good, so I can’t honestly promise a specific timeframe for those fixes.

At the same time, if the mod is released and an actual user reports a bug that they can reproduce in practice, that is a different situation. I would obviously prioritize and investigate a real user-facing bug as soon as possible. I’m much more interested in fixing something that an actual user can reproduce than in endlessly optimizing against hypothetical scenarios that have not manifested in practice.

So for this release, I’m prioritizing practical stability and the intended user experience over eliminating every theoretical edge case identified by static/AI review. I’m not claiming that the code is perfect or that there are no remaining edge cases; I’m simply choosing not to block the initial release on issues that have not manifested in real-world testing, especially when addressing them immediately can create more visible regressions than the issues themselves.

PR #5215, opened 2026-08-25
Windows 7 Blocked Shutdown UX & Logoff Sequence 1.0

Ingan121:
* This is a direct port of my [AuthUX BSDR fork](https://github.com/Ingan121/AuthUX/tree/bsdr).
* Tested on Windows 10 LTSC 2021 and 11 25H2.

## Changelog

* Changelog item 1...
* Changelog item 2...

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini
- - [x] Other (please specify): (Code review only)


mods/win7-bsdr.wh.cpp
```diff
@@ -0,0 +1,3526 @@
...
+  $name:ko-KR: 스크롤 막대에 시각 테마 적용
+  $description: The original Windows 7 blocked shutdown resolver had an unthemed scroll bar. Enable this if you find it ugly.
+- disableAsyncLogoff: true
+  $name: Restore Windows 7's logoff sequence
+  $name:ko-KR: Windows 7의 로그오프 절차 복원
+  $description: Switch to the 'logging off' screen after all applications have been closed.
+*/
+// ==/WindhawkModSettings==
+
+#include <windhawk_utils.h> // Includes <string>, <optional>, <cstlib>, <windows.h>
```
m417z: This might change in the future. Please include `<string>`, `<optional>`, `<cstlib>` separately.

mods/win7-bsdr.wh.cpp
```diff
@@ -0,0 +1,3526 @@
...
+
+HINSTANCE hResDll;
+bool isUsingHardcodedRes = true;
+
+// Hardcoded resources 
+
+// https://github.com/Ingan121/AuthUX/blob/42909c132f429337e6169cf67c3a253405e916d0/shell/auth/authux/consolelogon/dll/consolelogon.rc#L471
+alignas(DWORD)
+static constexpr unsigned char RES_DIALOG[]  = {
+    0x01, 0x00, 0xff, 0xff, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x40, 0x00, 0xc0, 0x40,
```
m417z: Is there a way to include this data in textual form to make these parts of the mod reviewable?

mods/win7-bsdr.wh.cpp
```diff
@@ -0,0 +1,3526 @@
...
+// Pressing ctrl alt del can get out of this state but lets add this minimal safeguard
+//
+// Can't think of better appraoch because of the execution sequence mentioned in CustomBSDR::Start
+// ShutdownWindowsWorkerThread runs before LogonUI exec so checking live LUI injection status isn't possible
+// and will be always one step behind (e.g. will detect as LUI not injected on first logoff after mod install)
+// I know the global inclusion key is not everything that affects that injection but this is minimal safeguard anyway
+// Let's just hope users don't mess with mod specific advanced settings to force exclude LogonUI.exe lol
+bool IsLogonUiInjectionEnabled() {
+    HKEY hKey;
+    int res = RegOpenKeyExW(HKEY_LOCAL_MACHINE, L"Software\\Windhawk\\Engine\\Settings", 0, KEY_READ, &hKey);
```
m417z, line 4117: This won't work with the portable Windhawk version.

Ingan121: ~~I changed to a different approach, but~~ the portable version is still not compatible because Windhawk dies too fast during logoff.
(Reverted to the registry checks again because the portable version is incompatible anyway)

Ingan121:
1. The maintainer only mentioned the dialog resource which was previously also hardcoded as hex blob. This mod is GPL licensed because it includes codes from the GPLed upstream AuthUX (LogonController interfaces and WIC handler), which already contains proprietary Microsoft bitmaps. If the maintainer decides embedding MS images is inappropriate, I'd rather just make the high contrast rendering the default than using inaccurately generated resources. (I'll consider compressing them though.)
2. LogonUI injection check method has changed over and over and over during the review process; the AI is now suggesting using the initial method again which was quickly abandoned due to not checking the live status and being one step behind. (One logoff after configuring properly will not have the old sequence restored, and one other logoff after improper config change will result in that undesired stuck in secure desktop behavior.) I also tried [writing specific data when called with some args, and executing LogonUI with that arg during probe](https://github.com/ramensoftware/windhawk-mods/pull/5215/changes/fe0671d230b6569ca41425f7df7b131ad4e12fac) to check the live injection status, but it costs one process exec and took around 0.5 seconds. So I settled with just checking the global inclusion registry, as this mod isn't even compatible with portable. Also note: applying the global inclusion setting fully restarts Windhawk and reloads every mods; there's no need to inform to disable and reenable the mod.

Another note: LogonUI just exits when idle. It isn't even running most of the time; more likely to be not running when the user changes the mod settings from the WH UI.

Ingan121:
## Resolve during uninit
* I actually have never even seen BSDR starting more than once in a single LogonUI.exe session; the multi-start path is just a best-effort implementation that was never tested (because the case doesn't exist). The chance of BSDR starting again while the previous one is still running is even smaller than the chance of it starting twice in a single LogonUI process.
* Also, calling `Resolve(Force, true)` in the isExiting bail-out path alone does nothing because the resolve function is only for the currently running BSDR session, not the one that's about to start.
## Bitmaps
The maintainer's comment was only about the dialog resource, which was previously also hardcoded as a hex blob. This mod is GPL licensed because it includes code from the GPLed upstream AuthUX (LogonController interfaces and WIC handler), which [already contains proprietary Microsoft bitmaps](https://github.com/world-windows-federation/AuthUX/tree/main/resources). If the maintainer decides embedding MS images is inappropriate, I'd rather just make the high-contrast rendering the default than use inaccurately generated resources.
## LogonUI injection probe method
LogonUI injection check method has changed over and over and over during the review process. The initial method was quickly abandoned due to it not checking the live status and being one step behind. (One logoff after configuring properly will not have the old sequence restored, and one other logoff after an improper config change will result in that undesired stuck in secure desktop behavior.) I also tried [writing specific data when called with some args, and executing LogonUI with that arg during probe](https://github.com/ramensoftware/windhawk-mods/pull/5215/changes/fe0671d230b6569ca41425f7df7b131ad4e12fac) to check the live injection status, but it costs one process exec and took around 0.5 seconds. So I settled with just checking the global inclusion registry, as this mod isn't even compatible with portable.

/ready-for-reviewer

mods/win7-bsdr.wh.cpp
```diff
@@ -0,0 +1,4326 @@
...
+* Known issues
+    * If your user account has no password, enabling the logoff sequence option may make the system automatically log on again after logging off.
+    * Not compatible with a portable Windhawk installation, even when run as admin, because it doesn't survive a logoff long enough to handle BSDR.
+
+![Screenshot](https://raw.githubusercontent.com/Ingan121/files/refs/heads/master/vmware_gj0gqnHj8e.png)
+## ⚠ Important usage note ⚠
+
+In order to use this mod, you must allow Windhawk to inject into the **LogonUI.exe**
+system process. To do so, add it to the process inclusion list in the advanced
+settings. If you do not do this, it will silently fail to inject.
```
m417z: I suggest to be more specific here, users sometimes confuse this with the mod's advanced tab. For example:

https://github.com/ramensoftware/windhawk-mods/blob/6dc728a85328d84fdaddb6b4339cd9347accc66e/mods/custom-corner-radius.wh.cpp#L40-L46

mods/win7-bsdr.wh.cpp
```diff
@@ -0,0 +1,4326 @@
...
+
+You must add it to the inclusion list even if you have disabled the critical system process
+exclusion option. Otherwise the logoff sequence portion of the mod will not function.
```
m417z: What do you mean here? If critical system process aren't excluded, then `logonui.exe` isn't excluded too, so it shouldn't be necessary to add it to the inclusion list in this case.

Ingan121: ~~The mod's safety checks intentionally check for the global inclusion key~~

Now also checks for the exclude critical process key

PR #5227, opened 2026-08-26
On-Screen Indicator Position 1.2.6

mario0318:
1.2.0 gives each kind of indicator its own position. volume can sit bottom left while brightness stays top centre and the camera one goes wherever you'll actually notice it. anything left on same as the main position just follows the setting above it.

the host has a separate entry point per kind, ShowVolumeAsync, ShowBrightnessAsync and so on, so the kind gets recorded as one is asked for and read back in GetPositionRect. all of those are hooked optional, so if one stops resolving on a later build that kind falls back to the main position instead of the whole mod refusing to load.

tested on 26200 with volume at top left and brightness at centre, both land where they should. one thing i noticed and can't do anything about from here, switching straight from one kind to another shows the previous one for a frame at the new spot, which looks like the frame being reused and the content swapping a beat later.

## Changelog

* A position per indicator kind, with everything defaulting to the main position
* PlaceInArea takes the position as a parameter rather than reading the global

## Mod authorship

This mod was created by:

- - [x] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/on-screen-indicator-position.wh.cpp
+++ b/mods/on-screen-indicator-position.wh.cpp
@@ -2,7 +2,7 @@
 // @id              on-screen-indicator-position
 // @name            On-Screen Indicator Position
 // @description     Put the volume, brightness and camera on-screen indicators anywhere on the screen, each in its own spot if you like, instead of the three positions Windows offers
-// @version         1.2.5
+// @version         1.2.6
 // @author          mario0318
 // @github          https://github.com/mario0318
 // @include         explorer.exe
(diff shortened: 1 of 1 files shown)


mario0318:
1.2.6 installs the eight kind recorders only when an indicator actually has a spot of its own, which was the ask last round.

the latest round suggests going back to installing all nine unconditionally instead, since the reload only exists to re-decide that and the symbol cache ends up resolving twice. both arguments look reasonable from here and i don't have a preference either way, so i'd make that your call.

the other thing still open is whether updating a mod can leave a newly added setting unwritten. i hit that locally but it may well have been the editor rather than a real install. if defaults always get backfilled then a couple of empty string checks and the constant beside them are dead weight, so let me know and they come out.

PR #5239, opened 2026-08-28
Update taskbar-ai-quota: add native settings and Antigravity support (v1.5.8)

Cleroth:
- Add a new settings system (Right Click taskbar -> Settings) with account management and display controls.
- Add Google Antigravity quota support through its running local app or CLI session.
- Add per-account quota selection
- Add Anthropic Fable and extra-usage bars
- Add quota pace ticks
- Add compact labels
- Fix percentage alignment
- Add visual previews
- Improve monitor selection, DPI rendering, tooltip readability, and taskbar layouts.
- Fixed minor bugs and rare crashes (harden OAuth, HTTP retry handling, taskbar lifecycle)

Note: existing Windhawk settings are imported automatically, and stored credentials are preserved.

### DIFF
--- a/mods/taskbar-ai-quota.wh.cpp
+++ b/mods/taskbar-ai-quota.wh.cpp
@@ -2,7 +2,7 @@
 // @id              taskbar-ai-quota
 // @name            Taskbar AI Quota Bars
 // @description     Shows configurable AI agent/LLM subscription quota bars for Anthropic, OpenAI, and Google Antigravity on the Windows 11 taskbar
-// @version         1.5.7
+// @version         1.5.8
 // @author          Cleroth
 // @github          https://github.com/Cleroth
 // @include         explorer.exe
@@ -164,6 +164,12 @@ enum class PercentTextAlignment {
     Right,
 };
 
+enum class PercentTextVisibility {
+    Never,
+    Hover,
+    Always,
+};
+
 enum class LabelPosition {
     Hidden,
     Left,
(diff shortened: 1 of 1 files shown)


Cleroth:
Addressed in v1.5.6:

- The Anthropic paste dialog now scales per monitor, uses a DPI-appropriate message font, follows DPI changes, and opens on the cursor monitor.
- Finite taskbar marshals now distinguish an unclaimed timeout from a callback already executing. Claimed payloads stay alive through the lower hook chain and the caller waits for completion.

Design decisions left unchanged:

- Removing the Windhawk settings block is intentional. The native settings window is the complete configuration surface. The no-injection access limitation is accepted for this version and called out for repository-maintainer approval.
- Legacy migration remains valid: Windhawk reads persisted setting keys independently of whether the current source declares their schema.
- Antigravity's one-minute local poll is deliberate and documented separately from the Cloud poll interval. With no matching process it performs only a process snapshot and executable-name filtering; endpoint probing is candidate-only and successful discovery is cached.
- The blocking joins and cleanup sends in Wh_ModUninit are intentional lifetime barriers. The DLL cannot safely unload while its worker threads, taskbar callbacks, subclasses, or XAML releases can still execute. All finite-marshal workers are joined before the final synchronous per-taskbar cleanup sends.

Cleroth:
Addressed in v1.5.7:

- QueueRefresh now releases g_refreshMutex before PostUiUpdate when the fetch thread is unavailable, removing the deadlock identified in the review.
- SettingsMessageBoxW now declines to open a modal dialog if its close-tracking CBT hook cannot be installed, preserving deterministic unload.

Cleroth:
@m417z zzzz
(A lot of repetition and nitpicks)

PR #5240, opened 2026-08-28
add faster-virtual-desktop-switching.wh.cpp

Meteoni:
## Changelog

Removes the long-standing pre-animation lag caused by oversized wallpaper thumbnail requests on Windows 11.

GIF

Windows 11 can synchronously request a full-monitor-sized wallpaper thumbnail
before beginning a virtual desktop transition. On a high-DPI display with specific wallpaper
configurations, that request can take roughly 136-169 ms when warm and more than one second 
when cold (exact figure depends).

This mod caps that request to the maximum size documented for `IThumbnailCache`, which is 
1024 pixels. In testing, that netted a latency reduction of ~150ms, effectively eliminating 
the long-standing pre-animation stutter issue. 

Laptop users who use three- or four-finger swipe gestures can especially benefit from this mod. 

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/faster-virtual-desktop-switching.wh.cpp
+++ b/mods/faster-virtual-desktop-switching.wh.cpp
@@ -18,6 +18,8 @@ Removes the long-standing pre-animation lag caused by oversized wallpaper thumbn
 
 ![GIF](https://raw.githubusercontent.com/Meteony/meteoni-assets/main/faster-virtual-desktop-switching/faster-virtual-desktop-switching.gif)
 
+_Note: This can make the wallpaper look softer during the switch. The mod is a no-op on Windows 10._
+
 Windows 11 can synchronously request a full-monitor-sized wallpaper thumbnail
 before beginning a virtual desktop transition. On a high-DPI display with specific wallpaper
 configurations, that request can take roughly 136-169 ms when warm and more than one second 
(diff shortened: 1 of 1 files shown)


Meteoni:
The first finding directly contradicts what was recommended in the previous round, so I'll leave it as-is.

Meteoni:
Seems to be the contradicting one brought over.

m417z:
The last review suggestion makes sense, especially since what the mod currently does seems to be unsafe: a second function hooked with the same hook and orig doesn't unhook the first function, so both trigger the same hook, but the orig pointer only calls the second function, which can mess up the first function calls.

What's "the contradicting one"?

Meteoni:
> The last review suggestion makes sense, especially since what the mod currently does seems to be unsafe: a second function hooked with the same hook and orig doesn't unhook the first function, so both trigger the same hook, but the orig pointer only calls the second function, which can mess up the first function calls.
> 
> What's "the contradicting one"?

Apologies. I should have phrased that better. Was tired and didn't take the time to think it through. It's on me. 

By that I meant the first finding from the preceding AI review. An earlier review pushed the implementation towards resolving the loaded thumbnail-cache implementations independently. Since registering multiple targets with the same detour/original storage is unsafe, I changed it to use a distinct pair for every target. 

The following review then specifically asked me to stop after the first successful DLL and use windows.storage.dll only as a fallback, so I changed the code accordingly. This latest review identifies that first-success behavior as load-order-dependent and recommends independently handling both DLLs again.

PR #5260, opened 2026-08-28
Bump to 1.7.0: cover common display languages in chevron detection  

### LINKED ISSUE #4524: [Tray hover expand] collapse delay and "hide" tooltip
Mod author: @wygodad 
Winver: 25H2 (OS Build 26200.8524)

hey, first of all nice mod!

i think it would be nice to have a separate **Expand delay (ms)** option, i wish that flyout didnt expanded as quicky as it collapses, but **Polling interval (ms)** affects both of them, making it not possible

about the new "hide" tooltip setting in specific, it doesnt work when **Taskbar on top for Windows 11** is enabled, im not sure if you can try to do something about it as the mod will not be updated anymore since microsoft is working on a native taskbar on top, so just a heads up

i also tried the setting without **Taskbar on top for Windows 11** enabled and found out that the flyout sometimes takes a second to actually disappear

### LINKED ISSUE #4684: [Tray hover expand] Unintentional cyclical expand / collapse
Mod author: @wygodad
Winver: 25H2 (OS Build 26200.8737)

I have been experiencing some glitching in last couple of days, where out of nowhere, the tray cyclically expands and collapses itself, until I click on it. 
I'm also using Taskbar tray icon spacing and grid, so there might be some interplay between them.

### LINKED ISSUE #4732: [Tray hover expand] Context menu items crossing the flyout boundary are unreachable with auto-collapse
@wygodad 

Context menu items crossing the flyout boundary are unreachable with auto-collapse enabled, because once the cursor leaves the tray area flyout, it triggers the collapse countdown, even when it hoovers above the context menu. 

Workaround is to use something like 3000 ms delay for auto-collapse, so that there is enough time to look through the context menu and choose what to click.

### LINKED ISSUE #4747: [Tray hover expand] tooltip showing first
Mod author: @wygodad
Winver: 25H2 (OS Build 26200.8737)

there are times when hovering the tray chevron, it makes the "show hidden icons" tooltip shows before the tray flyout, happens more frequently when hovering the chevron for the first time after the mod is enabled or when the settings are changed

https://github.com/user-attachments/assets/1bf10073-2caf-4dc3-8a82-4488c6ebeebe

### LINKED ISSUE #5257: [ Tray hover expand] It doesn't work anymore
Your mod no longer works after the latest Windows update, build 26200.9278.
Can you fix it, please?

@wygodad

wygodad:
This updates the already-published **Tray hover expand** mod to 1.7.0.

The main change is how the chevron is identified. Until now it was matched by its localized name, with a positional guess as a fallback, so users whose display language was not in the keyword list fell through to that guess, which can select a different tray button and invoke it. It is now identified by its class name together with its AutomationId, which is the same in every display language, and when it cannot be identified the mod does nothing at all instead of guessing.

The rest of the changes come from user reports about auto-collapse, the chevron tooltip and tray icon context menus.

## Changelog

- Identify the chevron by its class name plus AutomationId, independent of the  display language, and accept it only when exactly one element matches
- Restrict detection to tray elements, so no taskbar button can be matched by a keyword and invoked
- Do nothing instead of guessing the chevron by position when it cannot be identified; the guess is now an opt-in setting, off by default
- Cover the most common Windows display languages in the name fallback
- Add a "Hover delay" setting, so the flyout does not open when the cursor only brushes past the chevron
- Keep tray icon context menus usable: detect clicks that fall between polling ticks, and never collapse while a context menu is open
- Suppress the chevron tooltip whenever the cursor is on it, covering both "Show hidden icons" and "Hide", and with the taskbar on either edge
- Hide only tooltips owned by the taskbar process
- Do not treat an unrendered chevron as a hover target, which could turn the corner of the screen into a hotspot and make the flyout cycle
- Drive re-finding from the cursor instead of a fixed timer, with backoff on repeated failures, so an auto-hiding taskbar works immediately on reveal and a taskbar with no hidden icons costs almost nothing
- Scope the fullscreen check to the monitor the chevron is on
- Set per-monitor-v2 DPI awareness on the worker thread
- Document that the mod is Windows 11 only, and how to restore the keyword defaults

Fixes #5257
Fixes #4684
Fixes #4732
Fixes #4747
Fixes #4524

Note on #4524: two of its three points are fixed here, along with the context menu problem raised in a follow-up comment. The remaining one, the flyout occasionally taking a second to disappear, I was unable to reproduce; the reporter has been asked to say so if it persists.

## Mod authorship

This mod was created by:

- - [x] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/tray-hover-expand.wh.cpp
+++ b/mods/tray-hover-expand.wh.cpp
@@ -32,19 +32,25 @@ mod in a dedicated process and does not inject into the shell.
   hover (and optionally closes it when you move away).
 
 ## How the chevron is detected
-The "Show Hidden Icons" chevron has no language-independent unique identifier on
-Windows 11 (it shares AutomationId `SystemTrayIcon` with the clock, volume,
-battery, etc., and exposes no ExpandCollapse pattern). So detection is a hybrid:
-1. Match the button by name (the keyword list covers the most common Windows
-   display languages by default and can be extended in the settings).
-2. If no name matches (other languages), fall back to the leftmost tray button
-   with the configured AutomationId, which is normally the chevron.
+Only tray elements are ever considered, so no taskbar application button can be
+selected. Among those, detection goes:
+1. By class name, which is language-independent: the chevron is the only tray
+   element whose class name is `SystemTray.NormalButton` while its AutomationId
+   is `SystemTrayIcon`. Notification icons share the class but use a different
+   AutomationId, and the clock, volume, network, battery and "show desktop"
+   buttons share the AutomationId but use other classes.
+2. By name, in case a future Windows build renames those classes. The keyword
+   list covers the most common display languages and can be extended.
+3. Otherwise nothing happens. The mod does not guess, because invoking an
+   unidentified tray button would open whatever it happens to be.
 
 ## Notes
-- If your display language is not in the default keyword list, hover the chevron
-  with the mod disabled and add a fragment of its tooltip text to the "Chevron
-  name keywords" setting. Without a name match the mod falls back to a
-  positional guess, which can land on a different tray button.
+- If the chevron is not identified, the mod logs every tray candidate it saw
+  (class name, AutomationId, position, name). That log is what to attach to a
+  bug report.
+- Updating the mod does not change settings you have already saved. If you saved
+  settings before v1.7.0, reset "Chevron name keywords" to pick up the new
+  defaults.
 - If auto-collapse does not work, the flyout window class name may differ on your
   build. Change it in the "Flyout window class" setting.
 - Windows shows a "Hide" tooltip over the chevron while the flyout is open, which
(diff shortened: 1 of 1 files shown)


wygodad:
Thanks, this was a useful review. Everything is addressed, with one finding corrected by measurement.

**Finding 2 does not hold for UIA on this build.** I enumerated the entire UIA tree under `Shell_TrayWnd` (103 elements, all control types, depth 12) on 25H2 build 26200.9278: there is no `SystemTray.ChevronIconView` anywhere, and no SystemTray.SystemTrayFrame` container either. The mods you cite inject into explorer.exe and walk the XAML visual tree, where that type does exist; UIA exposes the automation peer's class name instead. What UIA reports:

| Element | ClassName | AutomationId |
|---|---|---|
| **Chevron** | `SystemTray.NormalButton` | `SystemTrayIcon` |
| Notification icons | `SystemTray.NormalButton` | `NotifyItemIcon` |
| Network, battery | `SystemTray.AccentButton` | `SystemTrayIcon` |
| Volume | `SystemTray.OmniButtonCenter` | `SystemTrayIcon` |
| Clock | `SystemTray.OmniButtonLeft` | `SystemTrayIcon` |
| Notification centre | `SystemTray.OmniButtonRight` | `SystemTrayIcon` |
| Show desktop | `SystemTray.ShowDesktopButton` | `SystemTrayIcon` |

So the goal of your finding is met, just through the pair rather than a single class: the chevron is the only tray element carrying `SystemTray.NormalButton` together with `SystemTrayIcon`. It is accepted only when exactly one element matches, and both values are settings.

1. **Fixed at the root.** Candidates are restricted to tray elements before any name or position test, so no task list button can be matched or invoked. Since the notification-area container is not exposed through UIA, the class-name prefix is the scoping mechanism.

3. **Went further than suggested.** Offscreen elements and degenerate rectangles are rejected, but the positional guess itself is now opt-in and off by default. Geometry validates a candidate's shape, not its identity, so when the chevron is absent the leftmost valid element is still the wrong button. The mod now does nothing and logs every tray candidate (class, AutomationId, position, name) instead.

4, 5. Tooltip hiding is filtered by the taskbar's process id, and the worker thread sets per-monitor-v2 DPI awareness.

All optional items done: unused include removed, keywords pre-lowercased at load, settings generation read before the snapshot, unload comment corrected, boilerplate restored to the wiki version. The readme now tells existing users to reset the keyword setting to pick up new defaults.

Verified on 25H2 26200.9278 (Polish): the mod log shows only open/collapse lines, confirming identification goes through the class name and never reaches the name fallback.

wygodad:
All three required findings are fixed, and most of the optional items as well.

1. The refresh path now applies the same guard as discovery and treats an empty rectangle like a stale element. Thanks for this one: `get_CurrentBoundingRectangle` returning S_OK with an empty rect is also a plausible cause of #4684, where a user with an auto-hiding taskbar sees the flyout cycle open and closed. A rectangle that alternates between valid and empty under a stationary cursor produces exactly that loop.
2. `lastOpenAt` is now `lastActionAt` and the collapse arms it too.
3. The chevron AutomationId setting is renamed and described as part of the primary identification, since it is no longer a fallback-only value.

Optional items done: cache request via `FindAllBuildCache` with `get_Cached*` accessors, the candidate dump logged once per failure streak, `GetVisibleFlyout` filtered by the taskbar process, the taskbar process id resolved once instead of per tick, `TRAY_CLASS_PREFIX` as a `wstring_view` with `starts_with`, and the readme now says how to restore the keyword defaults (clear the list and save).

On the tooltip process filter from the previous round: I verified the chevron tooltip is owned by explorer.exe, the same process as `Shell_TrayWnd`, so that filter is correct and hiding still works.

From the functionality notes: added an opt-in "Hover delay" setting, which also covers a standing request in #4524; scoped the fullscreen check to the chevron's monitor; accepted `ChevronButton` as a second AutomationId, which is free coverage given the pair is only used when exactly one element matches; and documented that this is Windows 11 only.

Deliberately skipped: RTL handling in the positional guess, which is opt-in and logged, and explicit `CUIAutomation8` timeouts, since UIA applies its own.

Verified on 25H2 build 26200.9278 at 140% scaling.

wygodad:
Both findings confirmed in the code and fixed.

1. Reproduced exactly as described: leaving the hit area clears the flag that spends the enter edge, and the branch that sets it again only runs when the flyout is closed, so a stay entered from the icons kept the edge armed indefinitely. A stay that has seen an open flyout now counts as served, so re-opening requires leaving and re-entering.

2. Re-finding is now driven by the cursor. It retries every 250 ms while the cursor is within a padded taskbar rectangle and backs off otherwise, so an auto-hide user gets the chevron back as soon as the taskbar is up, and a user with no hidden icons stops paying for the subtree walk entirely. This matters for #4684, which is an auto-hiding taskbar setup.

All optional items done as well: the collapse timer is cleared when the element is dropped, the hover-dwell state is reset on a settings change, the ambiguity line is gated by the same streak flag as the candidate dump, the dump is skipped when nothing was rendered (a retracted taskbar produces no candidates and needs no diagnosis), the dead header guard is gone, and the alternate AutomationId now carries a comment explaining why it is fixed rather than configurable.

On the functionality notes: `PtOverWindow` using the flyout's full window rect is intentional, since the shadow margin only makes auto-collapse slightly stickier, which is the harmless direction.

The reporter in #4684 has since confirmed the root cause was a missing keyword for Czech and that the 1.7.0 approach addresses it, so the PR now closes that issue too. Verified on 25H2 build 26200.9278 at 140% scaling, including the click-to-dismiss gesture from finding 1.

PR #5269, opened 2026-08-29
Update Don't Pull Me to Another Desktop to v0.4.0

Meteoni:
## Changelog

- Renamed the mod
- Changed the classifier to be rooted narrowly in CVirtualDesktopForegroundPolicy::ForegroundViewChanged, which offers better classification accuracy against normal user actions. 
- Hardening against apps that briefly activate a pre-existing window before creating a new one. 

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/prevent-virtual-desktop-stealing.wh.cpp
+++ b/mods/prevent-virtual-desktop-stealing.wh.cpp
@@ -24,7 +24,7 @@ window to you instead.
 
 ![Screenshot](https://raw.githubusercontent.com/Meteony/meteoni-assets/main/prevent-virtual-desktop-stealing/prevent-virtual-desktop-stealing.gif)
 
-If you haven't seen the problem before, try this:
+If you haven't seen the problem before, try this: 
 
 - Open Windhawk on one virtual desktop.
 - Switch to another desktop.
--- a/mods/prevent-virtual-desktop-stealing.wh.cpp
+++ b/mods/prevent-virtual-desktop-stealing.wh.cpp
@@ -1,7 +1,7 @@
 // ==WindhawkMod==
 // @id              prevent-virtual-desktop-stealing
 // @name            Don’t Pull Me to Another Desktop
-// @description     Summon existing windows to the current desktop, instead of pulling you away. 
+// @description     Summon existing windows to the current virtual desktop instead of pulling you away.
 // @version         0.4.0
 // @author          meteoni
 // @github          https://github.com/Meteony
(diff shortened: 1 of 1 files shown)


Meteoni:
Thanks for the detailed review. I've addressed the findings that held up after checking them against the runtime behavior observed while developing the attribution rework.

For view supersession, I agree that same-PID identity is the actual correlation criterion and that the previous PFN/AUMID matching was redundant. The superseder is now validated by same PID, a genuinely different shell view/HWND, and ownership by the current/source desktop. Stable package/AUMID identity is queried only once and is now used solely to decide whether to incur the longer 200 ms grace period. I kept that eligibility gate rather than extending every ordinary rescue from ~50 ms to ~200 ms.

The two timing settings have been removed entirely. They are now internal policy constants, and the local-foreground replacement validation no longer depends on whether the settling delay is enabled.

`ShowWindow` has been replaced with `ShowWindowAsync`, and failure of `SetForegroundWindow` is now checked and logged. The worker and notification joins intentionally remain unbounded: Windhawk must not unload the mod while those threads can still execute code from it.

I've also run the repository clang-format over the complete source. The file dropped from roughly 4260 lines to roughly 3100 lines without removing behavior. The impossible generation-zero checks, zero-time cancellation-event waits on shell hook paths, redundant second teardown, repeated package API resolution, and normal enable/reload `twinui.pcshell.dll` reference increment have also been cleaned up.

I kept the navigation-focus history rather than replacing it with one slot. During testing we reproduced an older navigation-selected view's ForegroundViewChanged arriving only after a newer DesktopChanged had already completed. Multiple navigation selections can therefore be outstanding; a single slot could be overwritten before the delayed FVP consumes it. I've added a comment explaining that observed race.

I also intentionally kept the stale-navigation suppression paths. These aren't unknown callers being treated fail-closed: the operation has been positively attributed to an obsolete navigation generation. We reproduced the A→B→C case where allowing the delayed B foreground-policy operation to continue pulls the user back from C to B. Calling the original in those paths would therefore restore the race the generation tracking was added to prevent.

The bounded initialization retry policy remains as-is. Exhausting all attempts is an extremely unusual fail-open startup condition, and I don't think adding permanent retry/rearming state is justified for that case.

Finally, I kept the explicit `IVirtualDesktopManagerInternal` slot access. A partial pure-virtual declaration would encode the same private vtable ordering and therefore wouldn't provide additional ABI protection if Microsoft reordered the interface without changing its IID. The existing 24H2+ IID gating remains the ABI guard.

The README now also restores the simple reproduction steps, intentional-navigation guarantees, Virtual Desktop Helper/Disable Virtual Desktop Transition Animation compatibility notes, and explicit "virtual desktop" wording for discoverability.

Meteoni:
For IsForegroundReplacementOnSourceDesktop, the rescue request now snapshots the Win32 foreground HWND when it is queued. The later settling check only treats a source-desktop foreground as a replacement if it actually differs from that queue-time foreground. I made the guard slightly more conservative as well: if the queue-time foreground is unavailable or already equals the remote activation target, the foreground heuristic abstains because there isn't a trustworthy pre-activation baseline. The existing shell-process exclusion remains as an additional conservative guard.

For view supersession, extendedSupersessionEligible now only chooses the listening deadline. A request without stable package/AUMID identity listens for the normal 50 ms settling window; an eligible request listens for up to 200 ms. Any VIEWADD observed inside that request's actual waitWindow is then judged solely by the same-PID / different-view / different-HWND / current-desktop validation. This matches the intended invariant that identity controls latency rather than correlation.

I also fixed the README continuation indentation and removed the stale taskbar.dll comment.

I kept the timing values internal rather than re-exposing them as settings, since they're implementation policy rather than user-facing behavior. I also kept the CreateWindowExW shell-ownership observer: the important property there is positively identifying the Explorer process that owns Shell_TrayWnd; observing twinui.pcshell.dll loading alone doesn't provide the same ownership signal.

The other functionality notes are intentionally unchanged: the bounded settling delay is the tradeoff required for supersession detection, the HookSymbols unload wait is required for safe unload, and the fail-stationary branches are retained for positively identified stale navigation generations.

Meteoni:
I've narrowed IsForegroundReplacementOnSourceDesktop further. The queue-time foreground baseline is still retained so the foreground must actually have changed after the rescue was armed, but a later foreground window can now cancel the rescue only if it belongs to request.pid, i.e. the process whose remote view was being activated.

That avoids treating normal focus restoration from transient taskbar/tray/launcher UI as evidence that the activation resolved locally, while keeping the fallback useful when the activated application itself genuinely foregrounds a different window on the current desktop.

I also applied the three minor cleanups: the package/AUMID eligibility query is skipped when the optional OnViewAddedInternal hook isn't available, the README again explicitly notes that a superseding view is independently verified to belong to the current desktop, and the diagnostic desktop->GetId() call in DesktopChanged_Hook is now only made for the outermost navigation where its log value is actually used.

I haven't changed the 50/200 ms timing split. The longer unidentified-app window is an interesting tuning possibility, but we currently have a concrete ~143 ms repro for the extended case and no corresponding evidence that ordinary Win32 rescues need a larger generic delay. I'd prefer to keep the common path at 50 ms until there's a real case demonstrating otherwise.

Meteoni:
Thanks. The theoretical manual-new-window false positive is understood, but I don’t think making the Windows Terminal default empty or adding a short “must appear immediately” cutoff would be an improvement.

Windows Terminal isn’t included as a speculative heuristic target. It is the concrete application for which this compatibility path was developed and tested. With Windows Terminal configured as the Default Terminal, stock Windows reproducibly does:

activate pre-existing Terminal window on another desktop -> attempt desktop switch -> asynchronously create the actually requested Terminal window on the current desktop.

During development I instrumented that sequence directly through the shell hooks: the old Terminal view reaches ForegroundViewChanged first, followed later by OnViewAddedInternal and ForegroundViewChanged for the newly-created same-process Terminal view.

Importantly, the replacement timing was not stable. A typical observed run was around 143 ms, but later tests on the same machine exceeded the former 200 ms supersession window. That was the reason the implementation was changed away from treating a fixed timeout as proof that no replacement is coming. Therefore I don’t think narrowing rollback by assuming a genuine replacement must appear “immediately” is sound; it would reintroduce the exact load-dependent failure that this revision removed.

The per-app list is intended as an allowlist for behavior whose causality cannot be proven generically, rather than a requirement that every affected application start disabled. Unlisted applications retain the deterministic summon-only path. Windows Terminal is shipped in the list because it is a known, reproduced affected application, not because package/process identity is being used as generic replacement evidence.

The two timing values also have deliberately non-causal meanings:

```
preMoveDelayMs is only a visual-churn optimization. If strong replacement evidence arrives during that period, the old window never needs to be moved. Expiry does not mean “this was not a replacement”; the normal summon proceeds.
rollbackWatchMs is a bounded validity period for later positive OnViewAddedInternal + exact-view ForegroundViewChanged evidence. It adds no latency after the summon.
```

The manual-new-window sequence described in the review is theoretically possible for any app placed on the allowlist; that unavoidable ambiguity is why replacement handling is not enabled generically. For Windows Terminal specifically, I think the known real-world handoff and the observed load-dependent delay justify enabling the compatibility policy by default. Users who prefer immediate deterministic summon-only behavior can remove the entry or set both values to zero.

PR #5276, opened 2026-08-30
Add Num Lock Lockdown

tonythethompson:
Adds **Num Lock Lockdown**, a lightweight tool mod that keeps Num Lock permanently ON. It runs in a dedicated `windhawk.exe` process with a low-level keyboard hook. Hold Shift (or another modifier) to toggle Num Lock normally. Releasing the modifier, session unlock, resume, and an optional safety timer force it back ON.

The PR adds a single file, `mods/numlock-lockdown.wh.cpp`, matching the `@id`.

## Changelog

- Initial public release.

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini
- - [x] Another AI (please specify): Cursor

### DIFF
--- a/mods/numlock-lockdown.wh.cpp
+++ b/mods/numlock-lockdown.wh.cpp
@@ -2,7 +2,7 @@
 // @id              numlock-lockdown
 // @name            Num Lock Lockdown
 // @description     Keeps Num Lock permanently ON, with a modifier key for temporary override
-// @version         1.1.1
+// @version         1.1.2
 // @author          tonythethompson
 // @github          https://github.com/tonythethompson
 // @include         windhawk.exe
@@ -26,7 +26,7 @@ covers the few cases the hook cannot see.
 | Normal use | Num Lock stays ON. The Num Lock key does nothing (default). |
 | Hold Shift + press Num Lock | Normal toggle is allowed. |
 | Release Shift | Num Lock is forced back ON if it is off. |
-| Sleep/wake, RDP, or another program turns it off | Forced back ON on the next event, or on the safety check. |
+| Sleep/wake, RDP, or another program turns it off | Forced back ON on the next event, on unlock/resume/focus change, or on the safety check. |
 | Mod disabled | The Num Lock *key* behaves normally again. The LED is left ON (the last forced state), not restored to whatever it was before the mod ran. |
 
 Hold **Left Shift** or **Right Shift** to temporarily turn Num Lock off (for
@@ -50,7 +50,9 @@ keeps overhead low and avoids tying Num Lock to the shell process.
   when an elevated window is focused, the injected Num Lock pulse is
   dropped and neither the return value nor `GetLastError` reports it. Run
   Windhawk as administrator if you need the force-on to work while an
-  elevated app is in front. The safety check retries after focus changes.
+  elevated app is in front. When focus moves back to a normal window, a
+  foreground-change hook retries the pulse. The safety check is a slower
+  fallback for cases with no focus change.
 - The hook also cannot see some Remote Desktop / KVM paths. Session-unlock
   / resume handlers and the safety check cover those.
 - If the hook misses a modifier key-up (elevated window or secure desktop),
@@ -58,8 +60,11 @@ keeps overhead low and avoids tying Num Lock to the shell process.
   the worker does the same before skipping a force-on.
 - **Block** also swallows the ToggleKeys accessibility shortcut (hold Num
   Lock for 5 seconds). Use **Allow** if you need that shortcut.
-- Synthetic Num Lock presses this mod sends are tagged and ignored by the
-  hook, so the force-on path cannot fight itself.
+- Synthetic Num Lock presses this mod sends are tagged and ignored by
+  *this* hook, so the force-on path cannot fight itself. Other apps still
+  see those injected presses (the pulse has to reach the OS to move the
+  LED). A game bound to Num Lock can therefore see a synthetic press when
+  a force-on fires, even in **Block** mode.
 - In "allow" mode, holding Num Lock without the modifier can leave it off
   until you release the key. Auto-repeat is swallowed so the LED does not
   flicker; force-on runs on key-up.
@@ -94,8 +99,9 @@ keeps overhead low and avoids tying Num Lock to the shell process.
   $name: Safety check interval (seconds)
   $description: >-
     Optional fallback that re-checks Num Lock in case it was turned off by
-    sleep, RDP, or an elevated app the hook cannot see. 0 disables the timer.
-    10 seconds is a light default.
+    sleep, RDP, or an elevated app the hook cannot see. Focus change, unlock,
+    and resume already retry without this timer. 0 disables it. 10 seconds
+    is a light default.
 */
 // ==/WindhawkModSettings==
 
@@ -119,9 +125,6 @@ constexpr UINT WM_APP_UPDATE_TIMER = WM_APP + 2;
 constexpr UINT WM_APP_QUIT = WM_APP + 3;
 
 constexpr UINT_PTR kSafetyTimerId = 1;
-constexpr UINT_PTR kDeferredForceTimerId = 2;
-constexpr UINT kDeferredForceDelayMs = 50;
-constexpr int kDeferredForceMaxTries = 2;
 
 // dwExtraInfo stamped on every synthetic Num Lock we send, so the hook can
 // let our own keystrokes through without treating them as user input.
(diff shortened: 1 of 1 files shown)


tonythethompson:
Pushed 1.0.9 for the two required findings:

1. Worker-side `IsOverrideHeldVerified()` re-seeds from `GetAsyncKeyState` before trusting a cached hold, and clears stuck Num Lock down flags when the key is up. Safety timer / unlock / resume can force ON again after a missed Shift-up.
2. `WM_APP_FORCE_ON` arms the deferred timer only when the first `ForceNumLockOn` did not send a pulse.

Also took most of the optional polish (wiki `swprintf_s` launcher, fatal `RegisterClassExW`, close ready events only after a successful join, single hook install plus a retry post if the handle is still null, plain `SetTimer`, leftovers, `StringSetting`, README notes on block-mode and disable-leaves-ON).

Left unaddressed: tearing down and reinstalling a still-non-null `HHOOK` from the safety timer. There is no API to ask whether Windows already dropped that hook, and unhooking a live one every 10s would open a gap. Retry only runs when our handle is still null (first install failed).

/ai-review

tonythethompson:
Pushed 1.1.0 for the two required findings.

1. Stale override cache on the hook path. `IsOverrideHeldVerified()` only ran on the worker, so a missed Shift-up still made the hook treat every Num Lock press as an override and skip force-on. Num Lock now re-reads the modifier with `GetAsyncKeyState` before deciding block vs allow. An allowed Num Lock up always posts to the worker; the worker still re-verifies and no-ops if the hold is real.
2. `WM_WTSSESSION_CHANGE` on `HWND_MESSAGE`. The worker is now a hidden top-level window (`WS_OVERLAPPED`, unowned, zero-sized, never shown, `WS_EX_TOOLWINDOW | WS_EX_NOACTIVATE`) so session unlock / reconnect can arrive. Power notify stays as-is.

Also took the cheap optional items: only wake the worker after an override was actually used, `ForceResult` so the deferred timer arms only when Num Lock still reads ON, seed modifier flags from locals then store, drop the dead `default:` arms and the unused `RegisterPowerAndSession` return, fail init if the hook does not install, two 50 ms deferred retries, README notes for UIPI/`SendInput`, the 10 s safety check, and ToggleKeys.

Left unaddressed: restoring the pre-mod Num Lock state on disable. The documented contract is "force ON while enabled, leave the last forced state when disabled." A restore pulse on unload would surprise anyone who turned the mod off expecting Num Lock to stay on, and `SendInput` at teardown can still be dropped by UIPI. The README already says the LED is left ON.

Also left unaddressed, from the previous pass: tearing down a still-non-null `HHOOK` every safety tick. There is no API to ask whether Windows already dropped that hook. Unhooking a live one on a timer would open a gap. If the first `SetWindowsHookEx` fails, init now fails instead of relying on a later retry.

/ai-review

tonythethompson:
1.1.1: `@author` is now `tonythethompson`, matching explorer-nav-dragover-fix and the catalog check.

tonythethompson:
Pushed 1.1.2 (includes the 1.1.1 `@author` fix that landed after the last review SHA).

Required: the README claimed a focus-change retry that did not exist. The worker now installs `SetWinEventHook(EVENT_SYSTEM_FOREGROUND)` and `RequestForceOn()` from that callback, then `UnhookWinEvent` on teardown. A UIPI-dropped pulse retries when focus leaves the elevated window. The safety timer is only the fallback when nothing else fires. README matches that.

Also took: drop the deferred AlreadyOn timer (Num Lock is settled by key-up, so those 50 ms ticks never did useful work), drop generic `VK_SHIFT` / `VK_CONTROL` / `VK_MENU` (LL hooks always report L/R), three short hook-install attempts then fail init, bail after the first ready-wait timeout, set `g_overrideUsedThisHold` when Shift is pressed between Num Lock down and up in allow mode, README note that the synthetic pulse still reaches other apps.

Left unaddressed: restoring the pre-mod Num Lock state on disable. Same as last time: the documented contract is leave the LED ON. A teardown pulse can still be dropped by UIPI and would surprise anyone who disabled the mod expecting Num Lock to stay on.

/ai-review

tonythethompson:
Pushed 1.1.4.

Taken from the last review:
- Re-install on unlock / console-or-remote connect / resume only. The hook thread handles `WM_APP_REHOOK` (thread message, `hwnd == null`): `UninstallKeyboardHook` then `TryInstallKeyboardHook`. That clears a stale `HHOOK` if Windows dropped the hook after a timeout. Not on the 10 s safety tick, and not a periodic unhook: that would reorder the chain and open a gap on every tick.
- `ForegroundEventProc` calls `ForceNumLockOn(false)` directly (same thread as `SetWinEventHook`). The foreground hook stays; it retries the pulse at the moment UIPI can succeed again.
- README and the override setting no longer claim Shift+numpad gives plain arrows. While Shift is held those keys are Shift+Arrow (extend selection). Ctrl/Alt/Win have the same kind of extra meaning.

Left unaddressed:
- Keep failing init if `SetWindowsHookEx` still fails after the short startup retries. Showing the mod as enabled while Block does nothing is worse than a failed enable.
- No periodic re-install cadence.
- No restore of the pre-mod Num Lock state on disable. Documented contract is leave the LED ON. A teardown pulse can still be dropped by UIPI.
- No latched/sticky override. That would be a new feature. The hold-while-modifier behavior is unchanged; only the docs were wrong.
- Dead null checks, `atomic<OverrideModifier>`, rename `IsOverrideHeldVerified`, `#include <cstring>`, 30–60 s safety default, orphan Num Lock key-up: optional polish, not changing working code for those.

/ai-review

tonythethompson:
Pushed 1.1.5.

Taken from the last review:
- `ReinstallKeyboardHook` installs a new `WH_KEYBOARD_LL` hook first, then drops the previous handle. If `SetWindowsHookEx` fails, a still-working hook stays installed. Unlock/resume still posts `WM_APP_REHOOK`; the handler no longer uninstalls first.
- Dedicated 60 s hook-retry timer (`kHookRetryTimerId`), independent of `safetyCheckSeconds` (including 0). A machine that never locks or sleeps still recovers a hook Windows silently removed. Not on the 10 s safety tick: reinstalling a live hook reorders the chain, so a slower cadence is the right balance.
- Optionals: clear `g_overrideUsedThisHold` with the other stale flags and on settings load; shorter allow dropdown label; mention the 3600 clamp; README allow-mode now notes the safety check can force ON mid-hold; call `RegisterSuspendResumeNotification` directly.

Left unaddressed:
- Keep failing init if `SetWindowsHookEx` still fails after the short startup retries. Showing the mod as enabled while Block does nothing is worse than a failed enable. The 60 s retry cannot help a process that already exited.
- No restore of the pre-mod Num Lock state on disable. Documented contract is leave the LED ON.
- No latched/sticky override. That would be a new feature.

/ai-review

tonythethompson:
Pushed 1.1.6 for the double-hook finding in the last AI pass.

Dropped the blind 60 s hook swap. Unlock/resume still reinstall (new hook first, then drop the old one). If a force-on finds Num Lock off, the same reinstall is requested, at most once per 60 seconds, so a dropped hook can still recover without swapping a healthy one every minute.

Startup still fails init if the hook never installs. Same reason as before.

tonythethompson:
Pushed 1.1.7 for the 1.1.6 AI review.

1. **Double pulse:** 300 ms debounce on `SendNumLockPulse` so unlock + foreground (or resume + focus) cannot queue two pulses before `GetKeyState(VK_NUMLOCK)` updates.
2. **Rehook on override:** `RequestEvidenceRehook` runs only when `fromSafetyCheck` is true. Hook-driven force-on (Shift release, Num Lock key-up, unlock/resume) no longer reinstalls a hook that just proved it is alive.

Also: stamp `g_lastRehookMs` only after a successful `PostThreadMessageW`; consume `g_overrideUsedThisHold` unconditionally on modifier-up; README note on MouseKeys.

Unlock/resume rehook kept. Fail-init on startup hook failure kept.

/ai-review

tonythethompson:
Pushed 1.1.8 for the 1.1.7 AI review.

1. **Failed hook install:** log and return `TRUE` instead of `ExitProcess`. Force-on (safety timer, foreground, unlock/resume) keeps working; Block mode is inactive until unlock/resume reinstalls the hook.
2. **Evidence rehook removed:** dropped `RequestEvidenceRehook` and the 60 s throttle. `RequestRehook` remains on unlock/resume only. Single `TryInstallKeyboardHook` at startup (no Sleep retry loop).

Pulse debounce from 1.1.7 kept.

/ready-for-reviewer

tonythethompson:
Pushed 1.1.9 for the 1.1.8 AI review.

If the keyboard hook handle is still null, `RequestRehookIfMissing()` runs on each safety tick (when enabled) or on a 60 s hook-retry timer when `safetyCheckSeconds` is 0. Only posts reinstall when the hook is missing, not a blind swap of a healthy hook.

Unlock/resume rehook unchanged. Stay-loaded-on-failed-install unchanged. Pulse debounce unchanged.

Did not re-add evidence-based rehook for a stale non-null handle (reviewer noted that as optional).

/ready-for-reviewer

PR #5278, opened 2026-08-30
Publish the Windows 7 Window animation restorer mod

babamohammed2022:
## Changelog

* Changelog item 1...
* Changelog item 2...

## Mod authorship

This mod was created by:

- - [X] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/win7-window-animations-restorer.wh.cpp
+++ b/mods/win7-window-animations-restorer.wh.cpp
@@ -1,7 +1,7 @@
 // ==WindhawkMod==
 // @id              win7-window-animations-restorer
 // @name            Windows 7 Window Animations Restorer
-// @description     This mod restores the Windows 7 Aero minimize and restore animation on classic Win32 windows without hooking DWM. 
+// @description     This mod restores the Windows 7 Aero minimize, restore and closing animations on Windows 10 and Windows 11
 // @version         1.0.0
 // @author          babamohammed
 // @github          https://github.com/babamohammed2022
@@ -300,7 +300,11 @@
 
 ## About
 
-This mod tries to restore Windows 7 window animations on Windows 10 and 11 for classic Win32 programs, **without modifying the DWM**.
+This mod tries to restore the Windows 7 window animations on Windows 10 and 11 for a series of programs **without modifying the DWM**.
+
+## Sample Animation
+
+![Demo GIF](https://raw.githubusercontent.com/babamohammed2022/babamohammed2022/main/bandicam2026-08-3014-19-03-266-ezgif.com-video-to-gif-converter.gif)
 
 ## Features
 
--- a/mods/win7-window-animations-restorer.wh.cpp
+++ b/mods/win7-window-animations-restorer.wh.cpp
@@ -5,6 +5,7 @@
 // @version         1.0.0
 // @author          babamohammed
 // @github          https://github.com/babamohammed2022
+// @license         MIT
 // @include         explorer.exe
 // @include         notepad.exe
 // @include         wordpad.exe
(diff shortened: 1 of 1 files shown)


babamohammed2022:
In my opinion the mod is ready for a human review as the problems reported seem to be related to specific cases (200% DPI is not a usual setting compared to 100%/125%) but I can try to fix them in a following update if the maintainer agrees. Additionally, the conflict is documented (and logically speaking I doubt that a user would enable 2 animation mods in his system).
/ready-for-reviewer

babamohammed2022:
OpenGlass, DWM 3D Transforms and 3D Aero Transforms were used as visual references only and no code was copied or adapted from them (they hook DWM internals inside dwm.exe; this mod doesn't inject into dwm.exe at all, it's explicitly @excluded, and renders its own GDI layered-window overlay instead).

The actual starting point was aubymori's classic-min-max-animations, which is MIT-licensed (it's the windhawk-mods repository default for mods with no @license tag). The mod is credited in the readme and the code based on it has been substantially modified.

The Mat::Matrix4x4F helper is hand-written and implements the same public API as Microsoft's D2D1::Matrix4x4F helper (d2d1_1helper.h -> [docs](https://learn.microsoft.com/en-us/windows/win32/api/d2d1_1helper/nl-d2d1_1helper-matrix4x4f)); it is not copied from any of the credited mods.

Since no GPL-licensed code was used and the only third-party borrowing comes from an MIT-licensed mod, I believe that MIT is correct here (unless the maintainer believes otherwise).

About the overlap with existing mods: as said above, no mod in the catalog restores the Windows 7 window animations, and the approach here is different anyway, so the overlap is smaller than it might seem. Folding this into another mod as one more style would also hurt discoverability: users would have to know that an unrelated multi-animation mod contains a hypothetical Windows 7 option, while a dedicated mod communicates its purpose directly.

babamohammed2022:
I do not possess a high scaled monitor and 150% DPI on this system would be false because it was made for 100%/125% DPI and in both cases the mod seems to work normally there and I can't reproduce this problem. If the maintainer believes that these problems need to be addressed, I'll do it.
/ready-for-reviewer

m417z:
Some of the code lines are hundreds of characters, making it difficult to review the code. Fix it.

babamohammed2022:
> Some of the code lines are hundreds of characters, making it difficult to review the code. Fix it.

I've indented the code, is it correct now?

babamohammed2022:
As said earlier, Aubymori's mod is related to the Windows XP and prior era while my mod is related to NT 6 (Windows Vista/7/8.1) which had animations more similar to my mod while as the other one, I think that a separate mod makes it easier for users to find the mod that achieves this goal (for instance, a user might want to make Windows 11 look like Windows 7 and he wants to restore the Windows 7-like animations and by typing "Windows 7" this mod will probably be one of the main options).

babamohammed2022:
I checked the Windows Animations source and its effects (Genie, Ink Splash, Scorch, Splinter, Mirage, Stipple, Swell, Windows 10) are all flat 2D warp/scale/fade animations and, unless I'm wrong, there's not a 3D perspective, rotation, or depth transform anywhere in the code. My mod's Aero tilt is built entirely on that kind of 3D transform (a camera/corner projection matrix), which that codebase simply doesn't have. So this wouldn't be a new option in an existing dropdown but it would mean building that whole 3D rendering system from scratch inside this modification, which is really a second implementation, not a normal-sized PR)

I do agree both mods hook the same APIs and would conflict if a user enabled both, that's already noted in my README, and I can add mutual excludes or a compatibility flag if that helps.
/ai-review

babamohammed2022:
I have not managed to replicate the reported problems. If the maintaner believes that they need to be addressed, I will try to do it
/ready-for-reviewer

PR #5281, opened 2026-08-30
doubleclick-resize.W/H

nh4700-ai:
## Changelog

* Changelog item 1...
* Changelog item 2...

## Mod authorship

This mod was created by:

- - [x] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/edge-doubleclick-resize.wh.cpp
+++ b/mods/edge-doubleclick-resize.wh.cpp
@@ -1,8 +1,8 @@
 // ==WindhawkMod==
 // @id              edge-doubleclick-resize
-// @name            Double-Click Edge to Maximize Width/Height
-// @description     Double-click the left/right window edge to maximize width, the top/bottom edge to maximize height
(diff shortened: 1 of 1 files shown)


m417z:
Thanks for the submission. How about adding this?

- **The width maximize doesn't toggle back.** The native vertical behavior this mirrors is a toggle — double-click the top edge again and the window returns to its previous height. Here, a second double-click on the left/right edge just re-applies the same rect, so there's no way to undo it other than dragging. Worth considering a restore: remember the pre-resize rect and, if the window's visible edges still match the work area, restore instead of maximize. If you do add per-window state, keep in mind that `DefWindowProc` is called on every UI thread in the process, so a shared map needs synchronization (each Explorer window, for instance, has its own thread) — a `SetPropW` on the window itself avoids that, at the cost of needing cleanup.

You can use `GetProp`/`SetProp` to store and then read the previous window width before it was resized to full width.

nh4700-ai:
Done — added a per-window toggle via

PR #5292, opened 2026-08-31
Add Safe UWP Process Cleanup mod

chitreyAbhinav:
This mod safely cleans up idle ApplicationFrameHost.exe processes without interfering with application launching. It uses a two-minute grace period and performs safety checks before terminating an idle process.

## Changelog

* New mod: Safe UWP Process Cleanup
* Targets only ApplicationFrameHost.exe
* Adds a two-minute startup grace period
* Checks for visible windows before cleanup
* Performs a final safety check before termination
* Does not target RuntimeBroker.exe, TextInputHost.exe, ctfmon.exe, or SystemSettingsAdminFlows.exe

## Mod authorship

This mod was created by:

- [ ] Claude
- [x] ChatGPT


Anixx:
Does it clean up only once or in a loop?

I think, the cleanup should happen only at the moment of closing of the UWP window.

chitreyAbhinav:
It runs in a loop, checking every 15 seconds for an idle process. I can change it to clean up only when the UWP window closes if that's preferred.

Anixx:
That of course is the only logical idea, to check at mod's load and on UWP window close.

Anixx:
I tried the latest version and it makes starting Settings app impossible. Does it work for you?

Also, it does not kill ApllicationFrameHost on startup.

Simply put, it does not work for me.

m417z:
@chitreyAbhinav can you comment on the reported issues? Were they reproduced? Fixed?

@Anixx can you try the updated version?

Anixx:
I tried the latest version. The Settings app starts, but ApplicationFrameHost remains in memory after exitting the Settings (I waited for 5 min)

PR #5296, opened 2026-08-31
Add File Operation Styler mod

digart11:
Adds File Operation Styler, a Windows 11 Explorer mod that modernizes the appearance of copy, move, delete, and recycle progress windows while preserving the native file-operation engine and behavior.

Features include:
- Compact and expanded layouts
- Circular percentage indicator
- Transferred size, current item, remaining items, speed, and estimated time
- Progress graph in More Details view
- Support for multiple simultaneous operations
- Native Pause, Resume, Cancel, conflict, and error handling
- Built-in themes and basic customization options

## Changelog

## Changelog

* Changelog item 1...
* Changelog item 2...

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [x] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/file-operation-styler.wh.cpp
+++ b/mods/file-operation-styler.wh.cpp
@@ -4037,17 +4037,37 @@ namespace
         return true;
     }
 
+    bool CircleStateRecordsHaveSameIdentity(CircleState const &candidate,
+                                            CircleState const &state)
+    {
+        if (state.circleWindow)
+        {
+            return candidate.circleWindow == state.circleWindow;
+        }
+        if (state.progressWindow)
+        {
+            return candidate.progressWindow == state.progressWindow;
+        }
+        if (state.infoWindow)
+        {
+            return candidate.infoWindow == state.infoWindow;
+        }
+        return false;
+    }
+
     void EraseCircleStateRecord(CircleState const &state)
     {
         std::lock_guard<std::mutex> lock(g_circleMutex);
-        g_circles.erase(
-            std::remove_if(
-                g_circles.begin(), g_circles.end(),
-                [&state](CircleState const &candidate)
-                {
-                    return candidate.circleWindow == state.circleWindow;
-                }),
-            g_circles.end());
+        auto it = std::find_if(
+            g_circles.begin(), g_circles.end(),
+            [&state](CircleState const &candidate)
+            {
+                return CircleStateRecordsHaveSameIdentity(candidate, state);
+            });
+        if (it != g_circles.end())
+        {
+            g_circles.erase(it);
+        }
     }
 
     void RetainCleanupOnlyCircleState(CircleState state,
@@ -4062,7 +4082,7 @@ namespace
                 g_circles.begin(), g_circles.end(),
                 [&state](CircleState const &candidate)
                 {
-                    return candidate.circleWindow == state.circleWindow;
+                    return CircleStateRecordsHaveSameIdentity(candidate, state);
                 });
             if (it == g_circles.end())
             {
@@ -4089,6 +4109,11 @@ namespace
 
     void ForgetCircleWindow(HWND circleWindow)
     {
+        if (!circleWindow)
+        {
+            return;
+        }
+
         CircleState state{};
         {
             std::lock_guard<std::mutex> lock(g_circleMutex);
@@ -4096,7 +4121,8 @@ namespace
                 g_circles.begin(), g_circles.end(),
                 [circleWindow](CircleState const &candidate)
                 {
-                    return candidate.circleWindow == circleWindow;
+                    return candidate.circleWindow &&
+                           candidate.circleWindow == circleWindow;
                 });
             if (it == g_circles.end())
             {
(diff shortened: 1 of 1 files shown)


Potassiumuncher:
Awesome

digart11:
The cleanup-only state identity issue has been addressed and is now confirmed resolved.

Two findings are intentionally left unchanged:
Teardown deadline: The teardown intentionally waits until subclass/custom-window callbacks into the mod have been removed. Returning after an arbitrary timeout could allow Windhawk to unload the module while Explorer still owns a subclass or window procedure pointing into it, creating a potential use-after-unload crash when that UI thread resumes. The concrete teardown bookkeeping issues identified in previous reviews, including tracking an unsubclassed progress HWND and cleanup-state collisions, have been fixed.

Localization: Explorer retains ownership of its localized native window title and locale-independent fallback is used for unsupported/special states. The remaining English strings belong to File Operation Styler's custom presentation. Full localization is considered a separate enhancement rather than a correctness fix for this initial release.

No further code changes are planned for these two findings unless will be requested.

digart11:
I investigated all four findings against the current reviewed commit.

1. Teardown retry loops: intentionally unchanged. The mod must not return from unload while Explorer can still call subclass/custom-window procedures inside the DLL. Returning after an arbitrary timeout could unload the module while live callbacks still point into it. The concrete teardown bookkeeping issues found in earlier reviews have already been fixed, and runtime teardown completes promptly in testing.

2. kSetTileDisplayModeThisAdjustment = 0x18: investigated and left unchanged. The value is documented from the target Windows 11 24H2 PDB/disassembly, and the current resolution path is fail-closed if it does not match a registered owner. Inferring the offset by scanning live owner-pointer deltas would be less reliable and could associate the callback with the wrong operation.

3. Cached tileHeaderRoot: investigated. The pointer is captured directly during DirectUI XML element creation and immediately hierarchy-validated. It has remained valid through normal operation and repeated More/Fewer transitions, and the transfer state is removed before native tile destruction. I also tested replacing it with later FindDescendent re-resolution from operationTileRoot; that lookup is not equivalent to parser-time capture and prevented the presentation from activating, so that change was reverted. There is currently no observed stale-header failure or verified lifecycle signal that would support a safer replacement.

4. Fixed window size: intentional behavior of this mod. The custom presentation owns a controlled DPI-scaled normal-mode geometry so that the progress circle, details, graph, multi-operation layout, and footer remain consistent. Making the host freely resizable would be a feature/design change rather than a correctness fix and would require separate layout work and testing.

No further code changes are planned for these findings unless they are requested

m417z:
Any reason you're using `noUndecoratedSymbols`? Currently, online symbol caching doesn't support it, which means that users will have to download the full pdb file.

digart11:
I believe noUndecoratedSymbols may be a leftover from earlier symbol-resolution work rather than something the current hooks actually require. The current hook entries use exact decorated symbol names, so I’m going to verify whether it can be removed safely and switch back to normal symbol caching if so. I’ll update the PR once I confirm.


> Any reason you're using `noUndecoratedSymbols`? Currently, online symbol caching doesn't support it, which means that users will have to download the full pdb file.

digart11:
yes, noUndecoratedSymbols was left over from earlier symbol resolution work and it's not required by the current hooks.
I removed it so Windhawk's online symbol caching can be used.

m417z:
You'll also need to update the actual symbols to be decorated. And remove the options altogether now that the default options are used.

digart11:
The claimed ordering bug is not present: the host teardown handler first calls DestroyProgressCirclesForHost, which removes the progress subclasses and destroys the custom child windows on the host UI thread. If any of that cleanup fails, the handler returns FALSE and the host subclass remains installed for retry. The host subclass is removed only after that cleanup succeeds, so loop 2 cannot encounter a live retained progress subclass merely because its host relay was removed.

The SMTO_ABORTIFHUNG concern is understood. The teardown intentionally does not return while Explorer can still call a subclass or custom window procedure in the module. A bounded timeout that proceeds with unload would violate that invariant and could create a use-after-unload. Therefore the current behavior is intentionally fail-safe: if the owning Explorer thread is genuinely unresponsive, teardown waits rather than unloading unsafely.

The related class-registration failure scenario also depends on teardown returning with live custom windows, which the current tracked-state cleanup does not allow.

SoftwareExplorer6600:
@digart11 I have some problems, this seems to NOT theme the dialog if its prompted for example some locations like the root of the drive (for example C:\ would be a good one here) require admin privileges, a dialog that looks similar and I suspect is invoked similarly seems to not be changed, the same for deleting something from recycle bin, that prompts a file transfer dialog that asks for permission, then you have merge conflicts and the like. so in most cases NO it isn't themed from my testing.

digart11:
This mod right targets only the main file-operation progress dialog.

The elevation/permission prompts, recycle-bin confirmation dialogs, merge/conflict dialogs, etc. are separate dialog paths and aren't all handled. Even though some of them look similar to the normal transfer dialog, they aren't necessarily created by the same UI/template that I'm currently hooking.





> @digart11 I have some problems, this seems to NOT theme the dialog if its prompted for example some locations like the root of the drive (for example C:\ would be a good one here) require admin privileges, a dialog that looks similar and I suspect is invoked similarly seems to not be changed, the same for deleting something from recycle bin, that prompts a file transfer dialog that asks for permission, then you have merge conflicts and the like. so in most cases NO it isn't themed from my testing.

PR #5308, opened 2026-09-01
Add Explorer Title Bar Label mod

digart11:
Adds a new Windhawk mod that displays customizable text, date and time in the Windows 11 File Explorer title bar.

The label is inserted as a native WinUI/XAML `TextBlock` next to Explorer's title-bar controls. It supports configurable date components and order, 12/24-hour time with optional seconds, font family/size/weight/style, text color, opacity, spacing, and live settings updates.

Tested on Windows 11 with multiple Explorer windows, maximize/restore, resizing, live settings changes, enable/disable, and Explorer restart.

## Changelog

* Initial release.

## Mod authorship

This mod was created by:

- [ ] Claude
- [x] ChatGPT
- [ ] Gemini
- [ ] Other (please specify):

## Changelog

* Changelog item 1...
* Changelog item 2...

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/explorer-title-bar-label.wh.cpp
+++ b/mods/explorer-title-bar-label.wh.cpp
@@ -1,0 +1,1587 @@
+// ==WindhawkMod==
+// @id              explorer-title-bar-label
+// @name            Explorer Title Bar Label
+// @description     Add custom text, date and time to the Windows 11 File Explorer title bar.
+// @version         1.0.0
+// @author          digitalART
+// @github          https://github.com/digart11
+// @license         GPL-3.0
+// @include         explorer.exe
+// @architecture    x86-64
+// @compilerOptions -lole32 -loleaut32 -lruntimeobject
+// ==/WindhawkMod==
+
+// Source code is published under the GNU General Public License v3.0.
+
+// ==WindhawkModReadme==
+/*
+# Explorer Title Bar Label
+
+Add custom text, date and time to the right side of the Windows 11 File Explorer title bar.
+
+![Explorer Title Bar Label](https://raw.githubusercontent.com/digart11/explorer-title-bar-label/main/images/screenshot.png)
+
+## Features
+
+- Custom text, date and time
+- Flexible date display
+- 12-hour or 24-hour time with optional seconds
+- Font, size, weight and color
+- Opacity and spacing
+- Live updates
+
+## Date and time
+
+Choose the date parts and order you prefer:
+
+- **Weekday:** None, Mon, Monday
+- **Day number:** 1, 01
+- **Month:** 8, 08, Aug, August
+- **Year:** None, 26, 2026
+- **Date order:** Month-Day-Year, Day-Month-Year, Year-Month-Day
+- **Numeric separator:** `/`, `-`, `.`
+
+Time can use **12-hour or 24-hour format**, with optional seconds.
+
+## Compatibility
+
+This mod uses XAML diagnostics and may conflict with other File Explorer mods that also use XAML diagnostics.
+*/
+// ==/WindhawkModReadme==
+
+// ==WindhawkModSettings==
+/*
+- customText: ""
+  $name: Custom text
+  $description: "Optional text displayed before the date and time."
+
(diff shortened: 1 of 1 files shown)


m417z:
I think it'd be preferable to have non-styler mods avoid using XAML Diagnostics. There are usually better ways to implement most things that are not generic styling. Here's an example of a submission that started with using XAML Diagnostics but then found a better implementation, perhaps you can do something similar:
https://github.com/ramensoftware/windhawk-mods/pull/4895#discussion_r3651079698
There's another example of an unpublished mod on Discord:
https://discord.com/channels/923944342991818753/1537122121858293890

Have you tried exploring other options to keep the mod compatible with the now-incompatible styler mod and third-party tools?

digart11:
I’m going to chekc a hook based approach similar to #4895, using File Explorer’s own WinUI code as an entry point and then walking the XAML tree directly with VisualTreeHelper. If I can make that work reliably for the title bar, I’ll update the mod to remove the XAML Diagnostics dependency and re-test it with File Explorer Styler enabled.
I will update pr when I get it it 


> I think it'd be preferable to have non-styler mods avoid using XAML Diagnostics. There are usually better ways to implement most things that are not generic styling. Here's an example of a submission that started with using XAML Diagnostics but then found a better implementation, perhaps you can do something similar:
> https://github.com/ramensoftware/windhawk-mods/pull/4895#discussion_r3651079698
> There's another example of an unpublished mod on Discord:
> https://discord.com/channels/923944342991818753/1537122121858293890
> 
> Have you tried exploring other options to keep the mod compatible with the now-incompatible styler mod and third-party tools?

digart11:
I explored the non-XAML-Diagnostics approach and implemented a prototype based on the FileExplorerExtensions.dll/VisualTreeHelper

I was able to hook CommandBarManager::CommandBar and several FileExplorerTabControl::TabView_* events, obtain a real typed WinUI element, walk its XamlRoot, find TabContainerGrid, and insert the label successfully without XAML Diagnostics.

This works reliably for newly created Explorer windows and when Explorer naturally rebuilds/loads the tab UI.

The remaining issue is already-open windows. When the mod is enabled after an Explorer window already exists, none of those construction/load events necessarily fire, and FocusManager::GetFocusedElement() can legitimately return null on that UI thread. I also tested bounded retries, but they continue returning null. I couldn’t find a supported HWND/WindowId -> XamlRoot path that would allow attaching immediately to an idle existing Explorer window.

So the non-diagnostics implementation introduces a regression compared with the current implementation: enabling or re-enabling the mod may require a new Explorer window or a later Explorer UI rebuild before the label appears.

Given that tradeoff, would you prefer the non-XAML-Diagnostics implementation with that limitation, or the current XAML Diagnostics implementation which handles existing windows and enable/disable reliably but is incompatible with another diagnostics consumer?



> I think it'd be preferable to have non-styler mods avoid using XAML Diagnostics. There are usually better ways to implement most things that are not generic styling. Here's an example of a submission that started with using XAML Diagnostics but then found a better implementation, perhaps you can do something similar: [#4895 (comment)](https://github.com/ramensoftware/windhawk-mods/pull/4895#discussion_r3651079698) There's another example of an unpublished mod on Discord: https://discord.com/channels/923944342991818753/1537122121858293890
> 
> Have you tried exploring other options to keep the mod compatible with the now-incompatible styler mod and third-party tools?

m417z:
> Given that tradeoff, would you prefer the non-XAML-Diagnostics implementation with that limitation

I think that's preferable to have a limitation to only apply the mod for new windows than being incompatible with other XAML Diagnostics consumers, especially given that the File Explorer Styler mod is among the most popular mods.

m417z:
I don't think it's necessary to include the whole license text in the readme, you can move it to the comments outside of the readme. Also, the Windhawk markdown doesn't support `<details>` so it looks broken now.

You can remove the symbols that include `__ptr64` from the code. It was only needed for very very old Windhawk versions.

PR #5311, opened 2026-09-01
Add Chrome Native UI Tweaks v1.0

Dron007:
Adds a new mod for customizing Chrome's native UI, including bookmark, tab, menu, and pinned extension appearance.

## Changelog

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [x] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/chrome-native-ui-tweaks.wh.cpp
+++ b/mods/chrome-native-ui-tweaks.wh.cpp
@@ -1,0 +1,1644 @@
+// ==WindhawkMod==
+// @id              chrome-native-ui-tweaks
+// @name            Chrome Native UI Tweaks
+// @description     Customize Chrome's native UI: fonts, tab close (×) buttons, menu density, extension button width, and bookmark folder icons. Web content is untouched.
+// @version         1.0
+// @author          Dron007
+// @github          https://github.com/Dron007
+// @include         chrome.exe
+// @architecture    x86-64
+// ==/WindhawkMod==
+
+// clang-format off
+// ==WindhawkModSettings==
+/*
+- bookmarks:
+  - fontSize: "14"
+    $name: "Font size"
+    $description: "Bookmark bar font size. Chrome default is 12."
+    $options:
+    - "12": "12 (Chrome default)"
+    - "13": "13"
+    - "14": "14"
+    - "15": "15"
+    - "16": "16"
+    - "17": "17"
+    - "18": "18"
+    - "19": "19"
+    - "20": "20"
+    - "21": "21"
+    - "22": "22"
+    - "23": "23"
+    - "24": "24"
+
+  - folderIcon: "windows"
+    $name: "Folder icon"
+    $description: "Folder icon used on the bookmark bar and in bookmark menus."
+    $options:
+    - "windows": "Windows system"
+    - "chrome": "Chrome default"
+  $name: "Bookmarks"
+
+- tabs:
+  - fontSize: "12"
+    $name: "Title font size"
+    $description: "Font size for titles in the horizontal tab strip. Chrome default is 12."
+    $options:
+    - "12": "12 (Chrome default)"
+    - "13": "13"
+    - "14": "14"
+    - "15": "15"
+    - "16": "16"
+    - "17": "17"
+    - "18": "18"
+    - "19": "19"
+    - "20": "20"
+    - "21": "21"
+    - "22": "22"
+    - "23": "23"
+    - "24": "24"
+
+  - hideCloseButtons: false
+    $name: "Hide close (×) buttons"
+    $description: "Hide tab close buttons and reclaim the horizontal space reserved for them."
+
+  - iconTitleSpacing: "8"
+    $name: "Icon/title spacing"
+    $description: "Horizontal space between the favicon and tab title."
+    $options:
+    - "2": "2"
+    - "3": "3"
+    - "4": "4"
+    - "5": "5"
+    - "6": "6"
+    - "7": "7"
+    - "8": "8 (Chrome default)"
+  $name: "Tabs"
+
+- extensions:
+  - buttonWidth: "34"
+    $name: "Pinned button width"
+    $description: "Horizontal width of pinned extension buttons. Button height stays unchanged."
+    $options:
+    - "28": "28"
+    - "29": "29"
+    - "30": "30"
+    - "31": "31"
+    - "32": "32"
+    - "33": "33"
+    - "34": "34 (Chrome default)"
+  $name: "Extensions"
+
+- menus:
+  - fontSize: "12"
+    $name: "Font size"
+    $description: "Font size for bookmark folders, the three-dot menu, and Chrome context menus. Chrome default is 12."
+    $options:
+    - "12": "12 (Chrome default)"
+    - "13": "13"
+    - "14": "14"
+    - "15": "15"
+    - "16": "16"
+    - "17": "17"
+    - "18": "18"
+    - "19": "19"
+    - "20": "20"
+    - "21": "21"
+    - "22": "22"
+    - "23": "23"
+    - "24": "24"
+
+  - verticalSpacing: "default"
+    $name: "Vertical spacing"
+    $description: "Top/bottom spacing for menu rows. Current Chrome normally uses 6. Some special rows can use smaller values."
+    $options:
+    - "default": "Chrome default (usually 6)"
+    - "0": "0 (tightest)"
+    - "1": "1"
+    - "2": "2"
+    - "3": "3"
+    - "4": "4"
+    - "5": "5"
+    - "6": "6"
+
+  - cornerRadius: "default"
+    $name: "Corner radius"
+    $description: "Outer corner radius of popup menus. Smaller values can also reduce the empty space above and below rounded menus."
+    $options:
+    - "default": "Chrome default"
+    - "0": "0 (square)"
+    - "2": "2"
+    - "4": "4"
+    - "6": "6"
+    - "8": "8"
+    - "10": "10"
+    - "12": "12"
+  $name: "Menus"
+*/
+// ==/WindhawkModSettings==
+
+// ==WindhawkModReadme==
+/*
+# Chrome Native UI Tweaks
+
+Customize several parts of Chrome's native browser interface while leaving web content untouched.
+
+## Getting started
+
+1. Enable the mod.
+2. Restart Chrome once.
+3. Open the mod settings, adjust the options, and click **Save settings**.
+
+Saved setting changes are applied immediately without restarting Chrome, so you can quickly experiment with fonts, spacing, button widths, and icons until the UI looks right for you.
+
+The mod doesn't make permanent changes to Chrome. If you disable or uninstall it, restart Chrome and the original UI is restored.
+
+## Features
+
+### Bookmarks
+- Adjust the bookmark bar font size.
+- Replace Chrome's bookmark folder icon with the Windows system folder icon.
+
+### Tabs
+- Adjust the tab title font size.
+- Hide tab close (×) buttons and reclaim the space reserved for them.
+- Adjust the spacing between the favicon and tab title.
+
+### Extensions
+- Reduce the width of pinned extension buttons to leave more room for the address bar.
+
+### Menus
+- Adjust menu font size.
+- Reduce vertical spacing between menu items.
+- Adjust popup menu corner radius.
+
+## Notes
+
+- Tested with **Google Chrome 152.0.7977.65 x64**.
+- When Chrome starts, Windhawk may show an initialization window while the mod resolves Chrome UI symbols.
+  This can take several seconds, especially on the first run with a new Chrome version.
+- Chrome instances launched by automation tools such as **Playwright** with `--remote-debugging-pipe` are intentionally ignored.
+- The mod relies on Chrome's internal native UI symbols. Chrome updates can occasionally require compatibility adjustments.
+*/
+// ==/WindhawkModReadme==
+// clang-format on
+
+#include <windows.h>
+#include <shellapi.h>
+#include <windhawk_utils.h>
+
+#include <algorithm>
+#include <atomic>
+#include <cstdint>
+#include <cwchar>
+#include <mutex>
+#include <unordered_map>
+#include <unordered_set>
+#include <vector>
+
+#pragma comment(lib, "shell32.lib")
+
+static constexpr PCWSTR kChromeSymbolServer = L"https://chromium-browser-symsrv.commondatastorage.googleapis.com";
+
+static constexpr int kChromeDefaultFontSize = 12;
+static constexpr int kChromeDefaultTabPreTitlePadding = 8;
+static constexpr int kChromeDefaultExtensionButtonWidth = 34;
+
+// Chrome 152 LayoutConstant ordinals.
+static constexpr int kLayoutTabAfterTitlePadding = 33;
+static constexpr int kLayoutTabAlertIndicatorCaptureIconWidth = 34;
+static constexpr int kLayoutTabAlertIndicatorIconWidth = 35;
+static constexpr int kLayoutTabCloseButtonSize = 36;
+static constexpr int kLayoutTabHeight = 37;
+static constexpr int kLayoutTabPreTitlePadding = 41;
+
+// -----------------------------------------------------------------------------
+// Opaque Chromium types
+// -----------------------------------------------------------------------------
+
+struct FontListOpaque;
+struct ImageModelOpaque;
+
+struct GfxSizeOpaque {
+  int width;
+  int height;
+};
+
+static_assert(sizeof(GfxSizeOpaque) == 8);
+
+// The Windows-folder image is created once and kept for the lifetime of the
+// Chrome process. These buffers are intentionally oversized for the small
+// Chromium wrapper objects used by the proven PoC.
+alignas(64) static unsigned char g_skBitmapStorage[256];
+alignas(64) static unsigned char g_imageSkiaStorage[256];
+
+// -----------------------------------------------------------------------------
+// Function types
+// -----------------------------------------------------------------------------
+
+using BookmarkButtonBaseCtorFn = void (*)(void*, void*, const void*);
+
+using BookmarkMenuButtonBaseCtorFn = void (*)(void*, void*, void*, const void*);
+
+using LabelButtonLabelCtorFn = void (*)(void*, const void*, int);
+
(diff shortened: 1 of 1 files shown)


Dron007:
The PR validation is currently failing on the symbol hook target:

https://github.com/ramensoftware/windhawk-mods/actions/runs/33556656710/job/100019056263

The actual target module is `chrome.dll`, and the hook array is named
`chromeDllHooks`. The validator correctly infers `chrome.dll` from that name,
but then rejects it with:

> "chrome.Dll" is not recognized as a Windows file name.

`chrome.dll` is a Google Chrome binary rather than a Windows system binary, so
it isn't present in the Windows filename database used by the validator.

The suggested comment form (`// chrome.dll` with a differently named hook
array) appears to go through the same filename validation, so I don't think
renaming the variable would resolve this without misrepresenting the actual
target module.

Should this validation warning simply be ignored for this PR, or would you
prefer another way to represent third-party symbol target modules?

m417z:
> > "chrome.Dll" is not recognized as a Windows file name.
> [...]
> Should this validation warning simply be ignored for this PR, or would you
>  prefer another way to represent third-party symbol target modules?

Ignore it for now. Symbols for non-Windows binaries weren't common so far, and currently there's no way to make the CI accept it. If it becomes more common, I'll come up with a better solution. For now, I'll bypass the check and merge it once everything else is ready.

Dron007:
Thanks for the detailed review. I've gone through the findings and tested the updated mod on Chrome 152.0.7977.76 x64.

**1. Six symbol hooks can never resolve, because they only list an undecorated name while `noUndecoratedSymbols = TRUE`**

Verified against the Chrome PDB. Four hooks needed explicit decorated alternatives and were fixed; the two `MenuItemView` entries already resolve from DIA's decorated field under their qualified names.

**2. `LoadLibraryExW` must be hooked in `kernelbase.dll`, not via the kernel32 import**

Fixed. The mod now hooks the actual `kernelbase.dll!LoadLibraryExW` implementation.

**3. The mod isn't reversible — disabling it requires a Chrome restart**

Fixed and tested on the current Chrome build. `Wh_ModBeforeUninit` restores the modified UI while the hooks are still available. Re-enabling the mod in the same already-running Chrome instance still requires a restart because constructor tracking is no longer present.

**4. Turning "Hide close (×) buttons" back off may not restore the buttons live**

Tested repeatedly; the existing layout invalidation restores the buttons live, so no additional code change was needed.

**5. The hard-coded `LayoutConstant` ordinals are the most fragile part of the mod**

Strengthened the runtime fingerprint to validate the contiguous tab-layout constants before applying the affected tweaks. Verified on Chrome 151–154; Chrome 150 has shifted ordinals and fails validation as intended.

**6. Unbounded fixed buffers for foreign Chromium objects**

Fixed. Opaque storage is now 4 KB per object with a 256-byte guard region, and temporary/cached Chromium objects are explicitly destroyed.

**7. Overlap with the existing `chrome-ui-tweaks` mod**

Documented the overlapping menu spacing and corner-radius settings in the README.

**8. Symbol-resolution cost needs to be documented accurately**

Changed both behavior and documentation. The normal startup path now waits at most 5 seconds; if symbol resolution is still running, Chrome continues with its normal UI while the cache is prepared in the background, with tray status indicating when a restart will activate the mod. Rebuilding the mod's symbol cache from an already downloaded PDB took about 24 seconds in testing; cached resolution was effectively instant.

**9. README has no screenshot**

Added screenshots showing the main UI tweaks and menu customization.

<details>
<summary>Optional improvements</summary>

**Use `WindhawkUtils::StringSetting` and drop the NULL checks.**  
Done.

**Declare `-lshell32` in `@compilerOptions`.**  
Done.

**Use `WH_MOD_ID` in the registered message name.**  
Done.

**Parse the command line properly.**  
Done with `CommandLineToArgvW`.

**Track extension views in the ctor/dtor, not in `CalculatePreferredSize`.**  
Done.

**The one-shot folder image is never destroyed.**  
Fixed; temporary and cached Chromium image objects are explicitly cleaned up.

**Defaults change the UI before the user configures anything.**  
Fixed; all settings now default to Chrome's normal UI.

**Missing include: `_wtoi` comes from `<cstdlib>`, which you don't include (you get it transitively today via `<cwchar>`).**  
Added `<cstdlib>`.

**Consider the simpler `void Wh_ModSettingsChanged()` — you already use it, which is correct here.**  
No change needed.

</details>

<details>
<summary>Functionality notes</summary>

**The Windows folder icon isn't DPI-aware.**  
Correct; this remains a documented limitation.

**The folder icon ignores Chrome's theme.**  
Correct and intentional; the option uses the Windows system folder icon. Documented together with the DPI limitation.

**`views::View::SetVisible` is hooked process-wide.**  
Kept. It is a direct pass-through except when close-button hiding is enabled; only tracked close-button views take the modified path.

**The bookmark font is overridden even at "12 (Chrome default)".**  
Fixed. The exact original bookmark `FontList` is preserved and restored.

**Why is `--remote-debugging-pipe` excluded?**  
Intentional, to avoid native UI hooks and symbol-resolution overhead in auxiliary automation/CDP Chrome instances such as Playwright.

**Symbol resolution blocks Chrome startup.**  
Addressed by the bounded 5-second startup wait and background symbol preparation described above.

</details>

Dron007:
Below is a response from the AI agent that implemented these changes with me; I also manually tested the mod’s behavior throughout the review fixes.

Thanks for the second pass. I verified the findings against Chrome 152.0.7977.76 x64 and tested the changes incrementally.

**1. The two `MenuItemView` symbol hooks still can't resolve, so the menu font size and vertical spacing settings do nothing**

The qualified names do resolve with the current Chrome PDB: Windhawk's decorated field comes from DIA `get_name`, which doesn't always expose an MSVC-mangled string for these entries. I nevertheless added explicit mangled alternatives and a menu readiness log. I also rechecked the ABI: Chrome 152 declares `const gfx::FontList GetFontList() const` (by value), so the existing Win64 sret hook prototype is correct. Menu font size and vertical spacing were tested after the change.

**2. Thread-affine Chromium objects are destroyed on Windhawk's unload thread**

Fixed. Omnibox `FontList` copies are now restored and destroyed on their owning Chrome UI thread. The cached folder `ImageSkia` records its creation thread and is also destroyed there. Tested with live changes, disable, and multiple Chrome windows.

**3. The symbol-notification thread creates a top-level window and never pumps messages**

Fixed. The notification thread now waits with `MsgWaitForMultipleObjectsEx` and continuously pumps its window messages while the tray icon exists.

**4. On the timeout path the resolved hooks are registered but never applied, so all the work is discarded**

The pending hooks are intentionally not late-applied. By the time slow symbol resolution finishes, constructor-based tracking has already missed existing `TabTitle`, omnibox, extension, and bookmark UI objects, so late activation would produce a partially active mod. The timeout path is deliberately cache preparation for the next Chrome launch. I made that state explicit with an activation-abandoned flag and clear the runtime readiness flags. Tested by deleting the symbol cache: Chrome waited about 5 seconds, continued stock, symbol preparation finished in the background, and the next launch activated the full mod.

**5. Tab titles aren't restored to their original font**

Fixed. `TabTitle` now captures its exact original `FontList` during construction and restores it for Chrome default and on unload. Tested with `12 -> larger -> 12` and mod disable.

<details>
<summary>Optional improvements</summary>

**`g_symbolResolutionThread` is written and closed from two threads without synchronization.**  
Fixed by keeping the resolver thread/event handles owned by `Wh_ModBeforeUninit` instead of closing them from the startup path.

**The guard check runs after the destructor is already unsafe to call.**  
Fixed. If an opaque-object guard is overwritten, the foreign destructor is no longer called on that object.

**`kOpaqueObjectStorageSize` is instantiated on the stack.**  
Left unchanged. The 4 KiB temporary storage is used only on these infrequent UI paths and isn't a meaningful stack/performance concern here.

</details>

<details>
<summary>Functionality notes</summary>

**Disabling or updating the mod blocks for the full duration of symbol resolution.**  
Kept as an intentional tradeoff. `HookSymbols` can't safely outlive the mod image, so unload waits for an in-flight resolver.

**The extension-width tweak is gated on a magic 34×34 preferred size.**  
Kept intentionally as a conservative guard so unrelated/nonstandard toolbar-action layouts aren't resized.

**`ValidateTabLayoutConstants` runs lazily from inside `GetLayoutConstantHook`.**  
Left unchanged. The validation is only reached from Chrome's own layout-constant calls and has been tested on the supported Chrome builds.

</details>

Dron007:
Hopefully, Claude will be happy now. Below is a response from the AI agent that implemented these changes with me.

Thanks for the third pass. Both mandatory lifetime issues have been addressed. I also applied several of the optional/functionality suggestions where they fit naturally with the same cleanup.

### 1. Both worker threads can be created *after* `Wh_ModBeforeUninit` has already run its teardown, leaving them running when the mod image is unmapped

Fixed.

Worker startup and teardown now share a lifecycle gate guarded by a mutex. Once unloading begins:

- no new resolver or notification worker can be created;
- an already-running `StartChromeHookSetup()` is allowed to finish its use of the startup handles before teardown takes ownership of them;
- resolver/notification handles are then snapshotted from a stable state and joined/closed during teardown;
- the notification stop event is signaled before joining its worker.

This also removes the race where `g_symbolResolutionDoneEvent` could be closed while the startup path was still waiting on it.

I tested both relevant slow-resolution races by deleting the symbol cache:

- disabling the mod during the initial 5-second startup wait;
- disabling it after the timeout while background symbol resolution and the tray notification were still active.

In both cases Chrome remained stable. In the second case the tray icon disappeared immediately, Windhawk remained in `uninitializing` until the resolver completed, and unload then finished normally. Re-enabling the mod and restarting Chrome worked as expected.

### 2. `BookmarkLabelInfo::originalFont` and `TabTitleInfo::originalFont` are raw references into `ResourceBundle`'s font cache

Fixed.

Bookmark and tab title tracking now stores owned copies of the original `gfx::FontList`, using the same guarded opaque-storage approach as the omnibox.

The copies are destroyed on their owning Chrome UI thread, both during normal object destruction and during mod teardown.

For reference, Chromium 152 currently comments that `ResourceBundle::ReloadFonts()` only happens on ChromeOS, so the suggested Windows text-scaling repro doesn't appear to apply to this build. Still, owning the copies avoids depending on that implementation detail.

### Optional improvements

**The tray icon doesn't survive an Explorer restart.**

Not changed. This is a rare cosmetic failure mode and doesn't affect Chrome or symbol resolution itself. I preferred not to add another window-class/re-registration path to the startup code at this stage.

**The Windows folder image is leaked when its UI thread has no window left at unload.**

Left as-is intentionally. If the owning UI thread can no longer be reached, the code prefers leaving the small cached object alive rather than destructing a foreign Chromium object from the wrong thread.

**`StartChromeHookSetup`'s return value is dead.**

Fixed. It now returns `void`.

**`wcsncmp(className, L"Chrome_WidgetWin_", 17)`**

Fixed. The prefix length is now derived from the string instead of duplicated as a literal.

### Functionality notes

**The font-size dropdowns are really size *deltas*.**

Fixed.

Bookmarks, tabs, menus, and the address bar now all derive a configured font from the actual original font of the relevant control:

`target size - original size`

Numeric options therefore represent absolute target sizes rather than assuming Chrome's base UI font is always 12.

`Chrome default` is now also a separate setting value rather than being conflated with `12`. Selecting it restores Chrome's exact original `FontList`, whatever its actual size is.

The settings UI uses static hints such as `Chrome default (usually 12)` and `Chrome default (usually 14)`. The actual default size is only known after Chrome's native UI objects exist. I left a short comment in the code noting that, if Windhawk gains a runtime settings-metadata/read-only-value API in the future, the detected size could be shown dynamically instead of using a static "usually" hint.

**`MenuItemGetFontListHook` drops per-item styling.**

Fixed.

The menu hook now first obtains Chrome's original `FontList` for that particular menu item and derives only its size. Font family, weight, and other styling are therefore preserved.

The same original-font derivation approach is now used consistently for bookmarks, tabs, menus, and the omnibox.

**`ValidateTabLayoutConstants` bakes in one UI density.**

No code change was needed here.

I checked Chrome 152's `GetLayoutConstant()` implementation against the existing fingerprint. The currently accepted values already cover both desktop and touch UI variants for the relevant constants:

- after-title padding: `4` / `8`
- alert icon width: `16` / `12`
- close button size: `14` / `16` / `24`
- separator height: `20` / `24`

The remaining checked values are the same in both modes in this Chrome version.

**The "don't late-activate" rule isn't applied consistently.**

The timeout behavior remains intentional.

After the startup timeout, constructor-tracked objects may already exist, so applying hooks after symbol resolution completes could produce a partially active mod. The resolver therefore only finishes populating the symbol cache and asks for a Chrome restart.

`Wh_ModAfterInit` also covers the normal startup case where `chrome.dll` was already loaded before the Windhawk init callback completed, so it isn't equivalent to always enabling the mod late in a fully initialized Chrome instance. Re-enabling the mod in an already-running Chrome still requires a restart, as documented in the README.

### Testing

After the changes I manually tested all mod options again, including live changes and Disable/Enable behavior:

- bookmark font and Windows folder icon;
- address-bar font;
- tab font, close buttons, and icon/title spacing;
- pinned extension spacing;
- menu font, vertical spacing, and corner radius;
- multiple Chrome windows;
- slow symbol-resolution startup;
- disabling during both the initial resolver wait and the background-resolution phase.

I also re-tested the final font-setting changes, including switching between explicit numeric sizes and `Chrome default`.

No regressions were observed.

Dron007:
Below is a response from the AI agent that implemented these changes with me.

Thanks for the fourth pass. I verified the `gfx::Size` ABI at runtime rather than changing it speculatively, documented the actual fresh-symbol preparation cost, and applied the two teardown robustness suggestions that exposed real edge cases.

### 1. Please double-check the `gfx::Size` return ABI in the `ToolbarActionView::CalculatePreferredSize` hook

Verified; the existing hidden-sret prototype is correct for the Chrome 152 build I tested.

I temporarily instrumented the hook exactly around the original call. The runtime result was:

```text
ToolbarActionView::CalculatePreferredSize ABI:
returned=0000002082DFBE00
result=0000002082DFBE00
w=28 h=28
````

So `returned == result`: Chrome is passing a hidden result buffer for this instance method, matching the existing declaration:

```cpp
using ToolbarActionViewCalculatePreferredSizeFn =
    GfxSizeOpaque* (*)(const void*, GfxSizeOpaque*, const void*);
```

The temporary diagnostic was removed after the test.

I also manually changed pinned extension spacing from 30 to 28 and back to 30 after this test, and the live update continued to work.

The sret ABI comment was moved so it now sits next to `ToolbarActionViewCalculatePreferredSizeFn`, which is the declaration it actually describes.

### 2. The README should state the actual first-run symbol cost, not just that it exists

Updated.

I performed a fresh-symbol test by removing both the local PDB cache entry and the mod's symbol cache before starting Chrome.

For Chrome 152.0.7977.76:

* local unpacked `chrome.dll.pdb`: **5,594,112,000 bytes (5.59 GB / 5.21 GiB)**;
* total fresh symbol preparation time reported by the mod: **139,532 ms (~2 min 20 sec)**;
* connection used for the test: **75 Mbit/s**;
* CPU: **Intel Core i7-8700K**.

The visible download completed after roughly two minutes, followed by approximately 20 seconds of symbol processing.

I can't give an exact wire-transfer byte count from Windhawk's current logging. The downloaded data is evidently compressed: the full local PDB is 5.59 GB, while compressing that same PDB locally as ZIP produces about 980 MB, and transferring 5.59 GB uncompressed over a 75 Mbit/s connection would take far longer than the measured 139.5 seconds.

The README now gives the measured local PDB size, total preparation time, and test-system/network context rather than only saying that symbol preparation may take some time.

I also changed the internal timing log wording from `symbol resolution` to `symbol preparation`, since on a fresh cache that interval includes downloading/preparing the PDB as well as resolving the requested symbols.

### Optional improvements

**The sret ABI comment is attached to the wrong typedef.**

Fixed. It is now attached to `ToolbarActionViewCalculatePreferredSizeFn`.

**4 KiB + 256 B of storage per tracked object is a lot for a `gfx::FontList`.**

Not changed.

The memory cost is small relative to Chrome itself, while the deliberately oversized opaque storage plus guard gives extra safety against relying on Chromium's private object layout. I preferred not to reduce that safety margin for a relatively small memory saving.

**The teardown restore paths are gated on the runtime "ready" flags, which a background thread can clear.**

Fixed.

Teardown no longer depends on the runtime readiness flag for restoring and releasing already-tracked address-bar or tab-title fonts.

The readiness flags still gate normal live-update behavior, but teardown operates from the tracked objects and already-resolved function pointers once hooks were activated.

**Tab close buttons are only restored if tab titles were tracked.**

Fixed.

Tracked `TabCloseButton` objects now also record their owning UI thread. The tab update/teardown dispatcher builds its thread list from both tracked tab titles and tracked close buttons, so close-button cleanup no longer depends on `TabTitle` tracking being available.

### Functionality notes

**Say in the README that disabling the mod can take a while.**

Fixed.

The README now explains that disabling/updating the mod while fresh symbol preparation is still running can leave Windhawk in `uninitializing` until the resolver finishes, because the resolver can't safely be allowed to outlive the mod DLL.

**The `TypographyProvider::GetFont` substitution during bookmark construction is now redundant.**

Not changed.

Although the later `LabelButtonLabel` update would normally make the final result correct, the constructor-time substitution guarantees that the bookmark control receives the desired font throughout construction rather than being corrected only afterwards.

It is working reliably in testing, so I preferred not to remove that path just to reduce code at this stage.

**The Windows folder icon is leaked when its UI thread has no window at unload, and that's still the right trade.**

Still intentionally unchanged.

If the owning Chromium UI thread can no longer be reached, the code prefers leaking that one small cached foreign object rather than running its destructor from the wrong thread.

### Testing

After the final changes we manually tested:

* bookmark font and Windows folder icon;
* address-bar font and restoration to Chrome default;
* tab font;
* hide/show tab close buttons;
* tab icon/title spacing;
* pinned extension spacing, including live `30 -> 28 -> 30` changes;
* menu font, spacing, and corner radius;
* multiple Chrome windows;
* Disable/Enable behavior;
* the fresh-symbol path with the local PDB and symbol cache removed;
* the cached-symbol restart path after preparation completed.

The fresh-symbol run took 139.532 seconds, then the next Chrome launch resolved from cache in effectively 0 ms and activated the hooks normally.

No regressions were observed.

Dron007:
Thanks. Since the remaining items are optional polish/optimization and the current code has passed the manual regression tests, I'm going to leave the implementation unchanged at this point rather than introduce another round of churn before human review.

PR #5313, opened 2026-09-02
Update taskbar-ai-quota: add OpenAI credits bar and amount text (v1.6.2)

Cleroth:
- Add OpenAI credits bar: balance shown against a per-account max, with thresholds and notifications
- Add option to show $ or credit amounts on extra-usage/credits bars instead of percentages
- Fix Anthropic extra-usage bar vanishing at the start of each monthly cycle
- Hide extra OpenAI rate limits (Codex Spark, gpt-reserve) behind one opt-in toggle

### DIFF
--- a/mods/taskbar-ai-quota.wh.cpp
+++ b/mods/taskbar-ai-quota.wh.cpp
@@ -2,7 +2,7 @@
 // @id              taskbar-ai-quota
 // @name            Taskbar AI Quota Bars
 // @description     Shows configurable AI agent/LLM subscription quota bars for Anthropic, OpenAI, and Google Antigravity on the Windows 11 taskbar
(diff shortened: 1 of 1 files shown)


Cleroth:
Credits decimals: the Codex CLI itself rounds the balance to a whole number before display (`format_credit_balance` in `codex-rs/tui/src/status/rate_limits.rs`: "finite credits show their rounded balance"), so `%.0f` matches the official client. `creditsMax` stays an `int` for the same reason.

Optional/functionality notes: `monthly_limit` and `used_credits` are always present in Anthropic's `extra_usage` object, and treating a credits-only payload as parsed is intentional (credits-only accounts). Leaving as-is.

PR #5315, opened 2026-09-02
Publish the first version of the Windows 7 logoff restorer mod

babamohammed2022:
## Changelog

* Changelog item 1...
* Changelog item 2...

## Mod authorship

This mod was created by:

- - [X] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/win7-logoff-screen-restorer.wh.cpp
+++ b/mods/win7-logoff-screen-restorer.wh.cpp
@@ -1,6 +1,6 @@
 // ==WindhawkMod==
 // @id             win7-logoff-screen-restorer
-// @name           Windows Vista/7 Logoff/Shutdown Screen Restorer
+// @name           Windows Vista/7 Logoff Screen Restorer
 // @description    This mod restores the classic Windows Vista/7 full-screen "programs still need to close" logoff and shutdown screen for Windows 10 and 11
 // @version        1.0.0
 // @author         babamohammed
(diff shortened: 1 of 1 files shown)


babamohammed2022:
I've tried to replicate the problems reported by the AI reviewer but the mod seems to work for me. About the third point, the synchronous design seems to work correctly. If the maintainer believes that these problems need to be addressed, I'll try to do so.
/ready-for-reviewer

m417z:
How does this mod compare to [Windows Vista/7 Blocked Shutdown UX & Logoff Sequence](https://windhawk.net/mods/win7-bsdr)?

babamohammed2022:
> How does this mod compare to [Windows Vista/7 Blocked Shutdown UX & Logoff Sequence](https://windhawk.net/mods/win7-bsdr)?

I have compared this to the other modification, and if I'm not mistaken, the underlying approach is fundamentally different despite the fact that the visual result is similar:

- My mod does not hook into `LogonUI.exe`, and it works entirely within the shell (explorer.exe and related shell hosts), which should carry lower risk if something goes wrong.
- It does not require external files while the other mod needs them to be manually added by the user for full functionality, which adds a real chance of user error (wrong path, wrong file version).
- My mod includes a built-in 60-second safety timeout.
- My modification also adds some functionalities the other mod doesn't seem to have: 21 built-in languages and partial DPI support.
So while the two mods have a similar objective I think that my approach is a lower-risk implementation with multiple translations.

Ingan121:
> partial DPI support.

FYI, my mod does support DPI scaling because it uses a native Win32 dialog. It currently lacks support for runtime DPI changes, which is unlikely to happen anyway because the window isn't movable and the Settings app isn't accessible during logoff.

babamohammed2022:
> > partial DPI support.
> 
> FYI, my mod does support DPI scaling because it uses a native Win32 dialog. It currently lacks support for runtime DPI changes, which is unlikely to happen anyway because the window isn't movable and the Settings app isn't accessible during logoff.

Thanks for informing me. I think that both approaches are valid and, ultimately, the user can choose the one which fits his needs. However, my modification does not require external files and is automatically translated in 21 languages. Both approaches create various use cases that could make one of the mods more useful to a certain category of users and viceversa. Thanks for your intervention.

m417z:
Why are all of these processes included?
```
// @include        explorer.exe
// @include        StartMenuExperienceHost.exe
// @include        ShellExperienceHost.exe
// @include        ShellHost.exe
// @include        SearchHost.exe
// @include        Taskmgr.exe
```
I don't believe `SearchHost.exe`, `ShellExperienceHost.exe` or `ShellHost.exe` are used to shut down the system. Also, as far as I know, `StartMenuExperienceHost.exe` doesn't have shutdown privileges, it asks a broker to do the actual shutdown. `Taskmgr.exe` has a disconnect option but it's quite niche.

Please limit the inclusion list to the processes that are needed for the mod to function.

Also address this:

- **The `enabled` setting duplicates Windhawk's own switch.** Windhawk already has a per-mod enable/disable that fully unloads the mod; a mod-level "Enable the custom screen" adds a second, weaker switch that keeps the hooks installed and the UI thread resident. The maintainer generally asks for settings like this to be dropped. Removing it would also let you drop the `g_enabled` branch from `ShowScreenOnUiThread` and `WM_APP_APPLYSETTINGS`.

babamohammed2022:
Both seem like edge cases that do not align with 1.0.0 release. If the maintainer believes that these problems need to be addressed, I'll try to do it.
/ready-for-reviewer

m417z:
Why is `Taskmgr.exe` targeted?

And did you verify that `StartMenuExperienceHost.exe` calls the hooked APIs? Which API does it use?

babamohammed2022:
> Why is `Taskmgr.exe` targeted?
> 
> And did you verify that `StartMenuExperienceHost.exe` calls the hooked APIs? Which API does it use?

Taskmgr.exe had been included to account for a hypothetical edge case of a user running a classic-style Task Manager tool that provides a shutdown option. I've reconsidered it as it addressed a theoretical scenario rather than a practical one.

StartMenuExperienceHost.exe was originally added based on an AI suggestion, intended to cover theoretical edge cases that couldn't be verified. After testing, I found that it does not trigger either hooked API because sign out, shut down, and restart initiated from the Start menu are all handled through explorer.exe. As I was unable to reproduce a scenario requiring this entry, I've removed both entries from @include to simplify the mod

babamohammed2022:
I think that the last part of the review is referencing to a comment that has been removed. If there are any problems to address, I'll do it
/ready-for-reviewer

PR #5327, opened 2026-09-03
Add recycle-bin-tray v1.0.0

wildstyle23:
Adds Recycle Bin Tray Icon, a lightweight Windhawk tool that places an interactive Recycle Bin icon in the system tray.

Main features:
- Live Empty/Full Recycle Bin state updates
- Drag & drop files and folders to the Recycle Bin
- System, Vector, Font, and Custom Icon rendering modes
- Light/dark theme support
- Per-monitor DPI-aware rendering
- Optional hide-when-empty behavior
- Configurable mouse actions
- Explorer/taskbar restart recovery

The mod runs in a dedicated windhawk.exe process and does not inject into explorer.exe.

Version: 1.0.0

Tested on Windows 11 with multiple DPI/display configurations, theme switching, Explorer restart, and drag & drop from File Explorer.

## Changelog

## Mod authorship

This mod was created by:

- [x] Claude
- [x] ChatGPT
- [ ] Gemini
- [ ] Other (please specify):

### DIFF
--- a/mods/recycle-bin-tray.wh.cpp
+++ b/mods/recycle-bin-tray.wh.cpp
@@ -1,0 +1,3683 @@
+// ==WindhawkMod==
+// @id              recycle-bin-tray
+// @name            Recycle Bin Tray Icon
+// @description     Adds an interactive Recycle Bin icon to the system tray with live state updates, drag-and-drop support, theme-aware rendering, multiple icon styles, and configurable mouse actions.
+// @description:fr-FR Ajoute une icône interactive de la Corbeille près de l'horloge dans la barre des tâches, avec mise à jour en temps réel, glisser-déposer, adaptation au thème, plusieurs styles d'icône et actions de souris configurables.
+// @version         1.0.0
+// @author          Wildstyle23
+// @github          https://github.com/wildstyle23
+// @license         GPL-3.0
+// @include         windhawk.exe
+// @compilerOptions -lshell32 -ladvapi32 -luser32 -lole32 -luuid -lgdi32 -lgdiplus -lshlwapi
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Recycle Bin Tray Icon
+
+![Drag and drop support](https://raw.githubusercontent.com/wildstyle23/Recycle-Bin-Tray-Icon/refs/heads/main/Images/Recycle%20Bin%20Tray%20Icon.png)
+
+A lightweight [Windhawk](https://windhawk.net/) Windows 11 mod that places an interactive Recycle Bin directly into your system tray.
+
+The icon follows the current Recycle Bin state, can optionally disappear when the bin is empty, reacts to Windows Shell notifications, and supports multiple rendering styles, themes, and configurable mouse actions.
+
+The mod is designed to remain lightweight and self-contained while integrating with the native Windows Shell wherever possible.
+
+## Features
+
+* **Live Recycle Bin state** — updates when items are added to or removed from the Recycle Bin.
+* **Optional auto-hide** — automatically hides the tray icon while the Recycle Bin is empty.
+* **Automatic recovery** — a configurable fallback timer periodically checks the Recycle Bin state in case a Shell notification is missed.
+* **Taskbar restart recovery** — the tray icon is recreated after Explorer or the Windows taskbar is restarted.
+* **Light / dark theme support** — vector and font rendering adapt to the current Windows system theme; custom icons use Light-theme source files, optional Dark-theme alternatives, and can automatically adapt transparent monochrome image colors to the active theme.
+* **DPI-aware rendering** — the icon size follows the tray's DPI and vector icons are rendered at high internal resolution for smoother results.
+* **Four icon styles**:
+  * `system` — use the native Windows Recycle Bin icon.
+  * `vector` — render a lightweight custom vector Windows 11-style icon.
+  * `font` — render a glyph from an installed font family.
+  * `custom` — load an `.ico`, `.png`, `.bmp`, or `.jpg` file.
+* **Two vector variants** — choose between a rounded/straight style and a Fluent-inspired trapezoidal style optimized for the system tray.
+* **Separate icons for states and themes** — custom mode uses Empty/Full Light-theme files with optional Empty/Full Dark-theme alternatives.
+* **Configurable mouse actions** — independently configure left click, double-click, middle click, and right click.
+* **Recycle Bin context menu** — open, empty, or open Properties using Windows Shell-provided localized labels when available.
+* **Optional empty confirmation** — keep or disable the confirmation dialog before emptying the Recycle Bin.
+* **Drag & Drop to Recycle Bin** — drag files and folders directly from the Desktop, File Explorer, or other applications that expose dropped files through the standard Windows `CF_HDROP` format.
+
+## Installation
+
+### From Windhawk
+
+1. Open **Windhawk**.
+2. Go to the **Explore** tab and search for **Recycle Bin Tray Icon**.
+3. Click **Details**, then click **Install**.
+4. Open the mod's settings to customize the icon and mouse actions.
+
+### Manual Installation from Source
+
+1. Install and open [Windhawk](https://windhawk.net/).
+2. Click **Create Mod**.
+3. Replace the template code with the contents of `mod.wh.cpp`.
+4. Click **Compile Mod** and enable the mod.
+
+The mod runs in a dedicated `windhawk.exe` process rather than being injected into `explorer.exe`, isolating it from the Windows Shell (see [Mods as tools](https://github.com/ramensoftware/windhawk/wiki/Mods-as-tools:-Running-mods-in-a-dedicated-process)).
+
+No separate installation, executable, or Windows service is required. Windhawk manages the process automatically.
+
+## Using the Tray Icon
+
+The tray icon provides configurable actions for each supported mouse input.
+
+### Mouse actions
+
+| Input        | Available actions                                 |
+| ------------ | ------------------------------------------------- |
+| Left click   | Open, context menu, empty, Properties, do nothing |
+| Double-click | Open, empty, Properties, do nothing               |
+| Middle click | Empty, open, context menu, Properties, do nothing |
+| Right click  | Context menu, open, empty, Properties, do nothing |
+
+Each action can be configured independently from the Windhawk settings.
+
+### Click / double-click behavior
+
+Windows sends a single-click notification before it can determine whether a second click will arrive.
+
+When a double-click action is enabled, the mod therefore waits for the Windows double-click interval before executing the single-click action.
+
+This small delay is intentional. Removing it while keeping independent single- and double-click actions could cause a physical double-click to execute the configured single-click action as well.
+
+When **Double click action = Do nothing**, the single-click action is executed immediately because no double-click needs to be distinguished.
+
+## Drag & Drop
+
+The Recycle Bin tray icon can be used as a normal drag & drop destination for files and folders.
+
+Simply drag one or more items from:
+
+* the Windows Desktop;
+* File Explorer;
+* other applications that expose file lists through the standard Windows `CF_HDROP` format;
+
+and drop them directly onto the Recycle Bin tray icon.
+
+The mod receives the dropped items through the standard Windows OLE drag & drop mechanism and moves them to the Windows Recycle Bin.
+
+### Requirements
+
+For reliable drag & drop operation:
+
+* The Recycle Bin tray icon must be **visible directly in the system tray**.
+* The icon should not be located inside the `^` notification-area overflow menu.
+* Windows must be able to identify the tray icon as the active drop target.
+
+To ensure the icon is always visible and available as a drop target:
+
+* **Manual placement:** Drag the icon out of the overflow menu and place it directly in the main system tray area. Windows will remember this placement.
+* **Windows Settings (if available):** Go to **Settings → Personalization → Taskbar → Other system tray icons** and toggle the icon to **On**.
+* **Windhawk alternative:** On modern Windows 11 builds where native toggles may be missing or restricted, you can use the [**Always show all taskbar tray icons**](https://windhawk.net/mods/taskbar-notification-icons-show-all) mod to keep all icons permanently visible.
+
+### Desktop focus behavior
+
+Windows Explorer may occasionally fail to initiate a drag from a Desktop file when another application currently has focus.
+
+For example, clicking a Desktop file and immediately moving the mouse may be interpreted primarily as a focus or selection operation instead of starting the drag.
+
+This happens before the drag operation reaches the mod and is related to normal Explorer Desktop behavior.
+
+### Workarounds
+
+Two simple workarounds are available:
+
+**Option 1 — focus the Desktop first**
+
+Click the file once without dragging, then start the drag operation again.
+
+**Option 2 — hold the mouse button briefly**
+
+Press and hold the left mouse button for approximately 500 ms before moving the file.
+
+Once Explorer has successfully initiated the drag operation, dropping the item onto the Recycle Bin works normally.
+
+The mod does not synthesize or force a Desktop drag when Explorer has not started an OLE drag operation. A low-level mouse hook is used only to arm the tray drop overlay; the actual data transfer still requires the source application to start normal OLE drag & drop.
+
+## Context Menu
+
+The optional Recycle Bin context menu contains:
+
+* **Open**
+* **Empty Recycle Bin** — disabled when the bin is already empty
+* **Properties**
+
+Menu labels are loaded from localized `shell32.dll` resources when available, so they follow the current Windows display language. The popup itself uses native Win32 menu rendering without custom menu icons.
+
+English fallback strings are used when a corresponding Shell resource cannot be loaded.
+
+## Emptying the Recycle Bin
+
+The **Confirm before emptying** option controls whether the standard Windows confirmation dialog is displayed.
+
+When enabled, the normal Windows confirmation dialog is shown before the Recycle Bin is emptied.
+
+When disabled, the mod requests the silent empty operation directly.
+
+This setting applies to actions that empty the Recycle Bin through the mod.
+
+## Icon Styles
+
+> **Choose one icon style:** `System` · `Vector` · `Font` · `Custom Icon`
+
+### System
+
+Uses the Windows Shell's own Recycle Bin icon.
+
+This is the simplest option and generally provides the closest match to the native Windows appearance.
+
+### Vector
+
+The icon is generated at runtime with GDI+ instead of loading an image file.
+
+It is designed for a clean Windows 11-style appearance and is rendered at a **4× internal resolution** before being reduced to the final tray size for smoother edges.
+
+Two variants are available:
+
+* **Style 1** — straight/rounded body.
+* **Style 2** — Fluent-inspired trapezoidal body optimized for tray-size rendering. The empty state uses a clean outline, while the full state uses the same silhouette with a filled body.
+
+### Font
+
+The icon is drawn from glyphs provided by installed font families.
+
+The empty and full states have separate **font family**, **glyph**, and **bold-weight** settings. Font and glyph fields are intentionally blank by default: if Font mode is selected without a usable configuration, the mod falls back to the native Windows Recycle Bin icon.
+
+Font family names are matched against the fonts installed in Windows. Leading/trailing whitespace and surrounding quotation marks are ignored when settings are loaded.
+
+The **Bold weight** option is available independently for the Empty and Full states. Leave it disabled for the normal font weight; enable it when the selected font provides a heavier bold variant or when the glyph looks too light at tray size. Internally, the mod uses the standard Windows font weights **FW_NORMAL (400)** and **FW_BOLD (700)**.
+
+#### Fallback behavior
+
+Font mode validates both the requested font family and glyph before rendering:
+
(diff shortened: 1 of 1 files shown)


wildstyle23:
Addressed the findings from the previous AI review and completed additional testing/polish:

1. Replaced the message-only host with a hidden top-level window so it can receive desktop broadcasts and act as a proper UI owner.
2. Added UIPI allowances for Shell/Taskbar messages and re-register Shell notifications after Explorer/taskbar restart.
3. Kept the validated synchronous IFileOperation drop semantics, but suspend and restore WH_MOUSE_LL around PerformOperations() so the low-level hook isn't left blocked during the operation.
4. Added 300 ms coalescing for Shell notification bursts to avoid repeated full Recycle Bin queries.
5. Added GdiFlush() before reading font DIB pixels.
6. Zero-initialized the monochrome HICON mask.
7. Improved shutdown behavior by cancelling an active tray menu on the tray thread before closing.
8. Made GDI+ lifetime opportunistic so it isn't kept initialized while using renderers that don't require it.
9. Improved per-monitor DPI transitions, Shell refresh behavior and tray geometry tracking; the latter also provides a fallback signal for future Explorer/taskbar layout changes.
10. Improved Recycle Bin restore notifications by subscribing recursively to changes under the Recycle Bin.
11. Added a third vector style and finalized the first-run defaults.

A few suggestions from the previous ai-review were intentionally not applied as written:

- Asynchronous recycle worker: the recycle operation remains synchronous because that path has been validated with Explorer/OLE and avoids the stale-source-item behavior encountered with the earlier asynchronous approach. The low-level hook is instead removed only around the potentially blocking PerformOperations() call and restored immediately afterwards.
- DROPEFFECT_MOVE: successful drops intentionally return/report DROPEFFECT_NONE. The target has already performed the recycle operation itself, so this prevents the source from deleting the items a second time. This behavior has been tested with Explorer and is kept deliberately.
- Disable drag & drop by default: drag & drop is a core feature of the mod and remains enabled by default. No meaningful idle CPU/resource issue has been observed.
- Remove/reduce the 10 s tray/DPI check: the 10-second check is retained as a lightweight fallback. In addition to DPI/location recovery, it now detects silent changes to the actual tray slot geometry returned by Shell_NotifyIconGetRect, while normal display/DPI events still provide the immediate path.
- String-setting NULL guards: these harmless defensive checks are retained; changing them provides no functional benefit.
- Font-name buffer / generalized font auto-sizing: the currently validated font fallback and rendering behavior is retained for the first release rather than introducing a broader font-rendering change at this stage.
- -luuid: retained deliberately; it has no meaningful runtime cost and removing it offers no benefit worth introducing a linkage difference before the first release.
- Undocumented Shell MUI resource IDs: retained with English fallbacks. If a resource ID changes on a future Windows build, localization degrades gracefully rather than affecting core functionality.

All functional changes were tested incrementally on Windows 11. Public version remains 1.0.0.

wildstyle23:
Updated the PR since the previous AI review (sha 0e149a62834d832cf6f7d16a8626d35a3bd3a6bd).

I revisited several items that I had previously left unchanged, and the current version now addresses them after additional implementation and Windows 11 testing:

1. Added the standard UIPI allowances to the OLE drop overlay for WM_DROPFILES, WM_COPYDATA and WM_COPYGLOBALDATA (0x0049).

2. Reworked Recycle Bin drops to use a dedicated STA/OLE worker. IDropTarget::Drop now extracts the CF_HDROP paths, queues the operation, reports the performed effect and returns immediately. File attribute lookup, IFileOperation and any Shell UI run on the worker instead of inside the source application's DoDragDrop call. Only one recycle job is accepted at a time, and the explicit source Shell refresh notifications are preserved.

3. Hardened shutdown around Shell UI and the recycle worker. Tray-thread Shell modal calls keep their hidden owner alive until they unwind, worker activity participates in close gating, and WhTool_ModUninit now uses a bounded 5-second wait for both tool threads rather than an infinite wait.

4. Separated the post-NIM_ADD refresh from TIMER_STARTUP. Normal startup initializes the Shell/tray immediately and only arms the one-second startup retry if initialization is not ready. TaskbarCreated keeps its 250 ms recovery retry.

5. Made Windows 11 the explicit target and simplified the DPI helpers to use GetDpiForSystem, GetDpiForWindow and GetSystemMetricsForDpi directly.

6. Reworked tray geometry caching. Provisional render size is now distinct from real Shell slot geometry, initial Shell geometry establishes a baseline instead of causing a spurious regeneration, and display-settle processing reuses an already sampled RECT instead of querying Shell twice.

7. Reduced drag tracking overhead. Ordinary left clicks no longer cause a Shell_NotifyIconGetRect query. Once the pointer crosses the Windows drag threshold, the first real drag performs one fresh Shell rectangle query and then reuses the cached rectangle for the remainder of the gesture. This preserves immediate correctness after manual tray-icon rearrangement.

8. Increased the default tray position/DPI fallback interval from 10 to 30 seconds. DPI, display and TaskbarCreated notifications remain the immediate update path.

9. Improved Font mode. GDI typeface names now use LF_FACESIZE with an explicit 31-character limit instead of silent truncation, font weight can be selected from 100 through 900, font/glyph validation is cached, and glyph size is reduced only when the actual rasterized ink would otherwise be clipped.

10. Removed the redundant Wh_GetStringSetting NULL checks according to the Windhawk API contract.

11. Removed the redundant -luuid dependency. Testing the change initially exposed _GUID_NULL; the explicit guidItem = GUID_NULL assignment was unnecessary because NOTIFYICONIDENTIFIER is already zero-initialized. Removing that redundant assignment allowed the mod to compile/link cleanly without -luuid.

12. Added further lightweight cleanup: settings snapshots and dropped path buffers are moved instead of copied where ownership is transferred, unused temporary D&D path storage is released before potentially long Shell work, theme/font validation results are cached, and GDI+ is released opportunistically when no rendered icon requires it.

13. Updated the drag-and-drop settings documentation to make it explicit that "Hide when empty" must be disabled for the tray icon to remain available as a drop target, and refined the Font/GDI documentation and weight descriptions.

**Intentional choices retained:**

- DROPEFFECT_NONE is deliberate. The drop target performs the recycle/move itself. Following the Shell optimized-move semantics, Drop returns a value other than DROPEFFECT_MOVE and also reports CFSTR_PERFORMEDDROPEFFECT = DROPEFFECT_NONE so that the source does not delete the originals a second time.
- Drag & drop remains enabled by default because it is a core feature of the mod. The previous per-click Shell_NotifyIconGetRect overhead has now been removed.
- The shell32 MUI resource IDs remain a best-effort localization mechanism with English fallbacks. A future resource-ID change can therefore affect localization only, not core behavior.

The changes were compiled and regression-tested incrementally on Windows 11, including repeated drag/drop and OLE re-entry, Explorer/taskbar restart recovery, DPI transitions, theme/settings changes and the normal Font rendering path.

The overflow-only Font shrink path was not forced with an artificial test font/glyph; it is deliberately conditional so glyphs that already fit retain their existing rendering.

Public version remains 1.0.0.

wildstyle23:
I went through the findings and optional suggestions, implemented the changes that were relevant, and re-tested the affected areas.

**Changes made**

The `WH_MOUSE_LL` hook now runs on its own dedicated thread with its own message queue. The hook callback is kept minimal and only forwards the drag start coordinates to the tray thread. Shell calls, rendering, Recycle Bin queries, and file operations are no longer performed on the hook thread.

The asynchronous recycle worker remains responsible for `IFileOperation`, so `IDropTarget::Drop` can return promptly and the source application is not held up by the recycle operation.

The Recycle Bin fallback query interval now defaults to 300 seconds instead of 60 seconds. The periodic DPI check now defaults to disabled (`0`), so normal DPI/display handling is event-driven. This was tested with live 100% <-> 200% DPI transitions while the DPI polling timer was disabled.

The drag tracking poll interval was increased from 20 ms to 40 ms.

The D&D-related UIPI message filters are now enabled only while Drag & Drop is enabled, and are explicitly blocked again with `MSGFLT_DISALLOW` when the feature is disabled. This was tested with D&D enabled -> disabled -> enabled again, as well as mod reload/shutdown.

The tray icon now uses a fixed `NIF_GUID` identity, including `Shell_NotifyIconGetRect`, to avoid `uID` collisions with other Windhawk tool mods.

`RegisterClassW` failures are now checked and handled cleanly. I also removed unused includes and simplified `IconCacheKey::operator==`.

The README/settings documentation was updated to document notification-area overflow behavior: when Windows places the icon in the `^` menu, files can be dropped directly onto `^`, or directly onto the Recycle Bin icon while the overflow panel is open. Keeping the icon permanently visible is recommended for quicker access but is not required.

**Intentional decisions**

Drag & Drop remains enabled by default. It is a core feature of this mod, and the low-level hook is now isolated on a dedicated lightweight thread and the active polling only runs during a physical drag.

The default mouse actions remain intentional: left click does nothing, double-click opens the Recycle Bin, and middle-click empties it. Emptying still uses the normal confirmation behavior by default.

I kept the existing per-monitor-v2 DPI architecture based on the real tray geometry and `GetDpiForWindow`, rather than switching to `GetDpiForMonitor`. The current event-driven implementation has been tested successfully with DPI polling disabled.

I kept the existing `Wh_GetStringSetting` / `Wh_FreeStringSetting` helpers rather than converting them to `WindhawkUtils::StringSetting`; the ownership is already centralized and balanced, so changing this would mostly be stylistic.

I also kept the existing `SHChangeNotifyRegister` delivery mechanism. No issue has been observed with it, and changing to `SHCNRF_NewDelivery` would add payload locking/unlocking that the mod does not otherwise need.

The hard-coded Shell MUI resource IDs were tested on a localized Windows installation and produce the expected Recycle Bin / Empty / Open / Properties labels, so they are retained with the existing English fallbacks.

I did not remove working renderer/DPI functionality merely to reduce the source size; those paths correspond to user-visible icon modes and DPI/theme behavior that have been tested.

Normal operation, D&D (including tray overflow), settings toggling, DPI changes, Explorer/taskbar restart recovery, context-menu actions, shutdown and repeated reloads have been tested without known regressions.

Public version remains 1.0.0.

wildstyle23:
Updated the PR since the previous AI review 

**Main changes:**

- The asynchronous recycle worker is still STA, but now pumps COM/window messages while idle using MsgWaitForMultipleObjectsEx + PeekMessage, instead of blocking indefinitely in WaitForSingleObject.
- Tray-icon startup/recovery retries are now bounded and use backoff, instead of retrying indefinitely. A new TaskbarCreated event resets the retry sequence.
- Restored the explicit <type_traits> dependency used by std::remove_pointer_t.
- Fixed the mouse-hook ready-event lifetime by giving the hook thread its own duplicated event handle, avoiding a timeout/close race.
- Mouse-hook state is now cleared conservatively if hook-thread shutdown times out/fails, so a later re-enable can't incorrectly reuse stale state.
- Disabling drag & drop now stops the drag polling timer synchronously on the tray thread.
- Fixed residual left-click activation after a double-click. A double-click configured to open the Recycle Bin no longer also triggers the configured single-click action/context menu.

**Recycle / IFileOperation changes:**

- Removed FOF_NOCONFIRMATION, so deletion now respects the Windows delete-confirmation preference instead of overriding it.
- Removed FOF_SILENT. This was found through testing to be the reason IFileOperation showed a generic error for selections exceeding the Recycle Bin quota. Without it, Windows displays its native warning/confirmation UI correctly.

Final operation flags are now:
```
- FOFX_RECYCLEONDELETE |
- FOFX_ADDUNDORECORD |
- FOF_WANTNUKEWARNING
```
- Added a passive IFileOperationProgressSink for diagnostics. It does not alter or cancel Shell operations; callbacks always return S_OK.
- The sink records aggregate counts for recycled, permanently deleted, cancelled, and failed items, and logs the affected item/HRESULT only for noteworthy outcomes.
- User cancellation is recognized both as COPYENGINE_E_USER_CANCELLED and HRESULT_FROM_WIN32(ERROR_CANCELLED).
- GetAnyOperationsAborted() is now queried after every PerformOperations() call regardless of its returned HRESULT, while success remains conservative if abort state can't be obtained.
- Normal TIMER_DRAG_POLL start/stop logging was silenced to keep diagnostic logs readable; timer failures are still logged.

**Previous optional findings intentionally left unchanged:**

- MSGFLT_RESET was tested for the D&D messages but returns ERROR_INVALID_PARAMETER (87) on the tested Windows 11 system; the validated MSGFLT_DISALLOW path is therefore retained.
- Font-height-fit memoization is deferred unless profiling shows a meaningful benefit; icon generation is already cached.
- Failed-item skipping is intentionally not implemented: destructive drag & drop remains all-or-nothing rather than silently producing partial results.
- Per-item SHCNF_FLUSH source refresh notifications are retained for v1 because this is the tested refresh path; bulk-drop optimization is deferred.
- Drag & drop can temporarily be unavailable while the tray icon itself is visibly moving/reflowing; testing found no stale overlay or incorrect deletion, so this transient limitation is accepted for v1.

Public mod version remains 1.0.0.

wildstyle23:
Changes since the previous review:

* Filtered `LLMHF_LOWER_IL_INJECTED` in the low-level mouse hook so a lower-integrity process can't arm the drag-and-drop path through injected mouse input.
* Added a 2-second grace period and forced cleanup for stale OLE drag sessions, and made the drag polling timer self-terminating.
* Wrapped `TrackPopupMenuEx` in `ShellModalScope` for consistent shutdown/modal handling.
* Migrated string settings to `WindhawkUtils::StringSetting` RAII.
* Removed the unused recycle-worker completion success flag.
* Added the explicit `<cstdlib>` dependency and use `std::abs`.
* Added a safety guard for items dragged from the Windows Recycle Bin itself. Explorer exposes such items through `CF_HDROP` as physical `$Recycle.Bin\...\$R...` paths; these paths are now rejected before any recycle job is queued. The rejection is all-or-nothing if any dropped item is already under a volume's `$Recycle.Bin` root.

Additional validation:

* Shell UI ownership was tested rather than changed. With Windows delete confirmation enabled, permanent-delete policy, and an item exceeding the Recycle Bin quota, the Shell dialogs remained visible, focusable and reachable through Alt+Tab. Cancel works correctly and drag & drop remains functional afterwards.
* Dragging an item from Explorer's built-in ZIP namespace is already rejected by design because it doesn't expose the `CF_HDROP` format accepted by the mod.
* Dragging an item already present in the Windows Recycle Bin is now safely rejected and no longer removes or permanently deletes the physical `$R...` file.
* Normal Explorer drag & drop remains functional.

A few previous suggestions remain intentionally unchanged for the first release:

* `DROPEFFECT_NONE` after accepting a drop is deliberate: the target already owns the recycle operation, and this prevents the source from deleting the item a second time.
* Recycle operations remain all-or-nothing rather than silently skipping invalid items.
* `MSGFLT_DISALLOW` is intentionally retained when disabling drag & drop; `MSGFLT_RESET` returned error 87 in real testing, while `DISALLOW` works correctly.
* Per-item `SHCNF_FLUSH`, the current renderer/DPI architecture, and the default-enabled drag & drop path are retained because they are validated and stable. Potential optimizations can be revisited after the first release.
* The current Shell MUI resource IDs are retained with English fallbacks; failure would degrade localization rather than core functionality.

The public mod version remains 1.0.0.

wildstyle23:
Changes since the previous review:

* Fixed duplicate tray activations under `NOTIFYICON_VERSION_4`. Diagnostics confirmed that Windows delivers both `WM_LBUTTONUP` + `NIN_SELECT` for one physical left click, and `WM_RBUTTONUP` + `WM_CONTEXTMENU` for one physical right click. The handler now uses the semantic v4 notifications and keeps the legacy notifications only for the non-v4 fallback path.
* Made post-double-click suppression self-expiring instead of depending indefinitely on a future `WM_LBUTTONDOWN`. `WM_LBUTTONDOWN` is still kept as an immediate reset path.
* Preserved independent single-click / double-click semantics: mouse single-click actions are delayed until the Windows double-click interval expires, so `Double click = Do nothing` does not accidentally execute the configured single-click action.
* Added the missing `UnregisterClassW` cleanup if creation of the tray host window fails.
* Avoided arming the drag polling timer while the tray icon is hidden.
* Fixed the native `system` icon renderer for per-monitor DPI. Real testing showed that `SHGetStockIconInfo(... SHGSI_SMALLICON)` returned a 16x16 icon even when the tray target was 32x32 at 200% scaling. The system renderer now obtains the stock icon location and extracts it at the computed target size, with the previous stock-icon loader retained as a fallback.

The remaining optional observations are intentionally deferred:

* The `SetCursor` calls in `DragLeave` / `Drop` are harmless cleanup-only candidates.
* Fully stopping/restarting the low-level mouse hook when `Hide when empty` hides the icon would add lifecycle churn; the unnecessary drag-poll timer work is already gated while hidden.
* `SHQueryRecycleBinW(NULL, ...)` remains on the tray thread unless real blocking is reproduced or reported.
* `CFSTR_LOGICALPERFORMEDDROPEFFECT` is being left for a future investigation together with Shell refresh optimization: because recycle operations are asynchronous, advertising the logical move before `IFileOperation` completes could make the source view update before a cancellation or failure is known.

The public mod version remains 1.0.0.

wildstyle23:
Follow-up after the previous AI review (SHA `6323b92ba8405c8d3a17aeeca9d10c482c94be35`).

I went through the remaining findings and re-tested the changes on Windows 11 rather than changing code solely to satisfy the review.

### Key updates in this revision:

* Kept the localized `shell32.dll` resource IDs intentionally. English fallbacks were already present, and the key Recycle Bin resource IDs were verified historically **back to Windows XP** (and earlier for some resources). Added a short code comment documenting this tradeoff.

* Removed the old configurable `dpiCheckInterval` polling path. A small fixed DPI fallback watch was added after reproducing a real primary-display-switch issue where the notification area could move to another DPI without the existing display-event path refreshing the hidden PMv2 DPI reference. It is skipped while the icon is hidden or drag-and-drop is active, and only regenerates the icon when the tray geometry/DPI actually changes. Tested successfully across 100% ↔ 200% primary-monitor switches.

* Removed the drag-overlay UIPI message-filter exemptions after integrity-level testing showed they were unnecessary; the separate Shell notification filter remains.

* Hardened `fallbackTimerInterval` handling (0 = disabled, active values clamped to 10 s..24 h), changed the default middle-click action to `none`, and validated the registry value type/size when reading the Windows theme setting.

* Kept GDI+ lazy-initialized for the tool lifetime and shut it down only at final teardown. Tested through Vector/Custom/System/Vector transitions and hide/show state changes with no intermediate shutdown/reinitialization.

* Kept the existing raster-based font glyph fitting. I benchmarked a `GetGlyphOutlineW(GGO_METRICS)` alternative; while faster for BMP glyphs, retaining a safe fallback for surrogate/non-BMP and atypical glyphs would add a second code path for negligible user-visible benefit, so I kept the more robust existing implementation for v1.0.0.

* Added/updated the small comments requested by the previous review (empirical dark-theme alpha adjustment and `ScreenDC` ownership wording).

The public mod version remains 1.0.0.

wildstyle23:
### Changes since the last review

* **DPI monitoring**

  * Removed the periodic topmost overlay show/move/hide probe.
  * DPI detection is now passive and uses the actual tray/taskbar window under the tray rectangle.
  * Verified on a real Windows 11 setup with 200% → 100% → 200% scaling changes.

* **Drag & Drop / hidden-icons chevron**

  * Direct drops onto the closed `^` overflow chevron are no longer accepted.
  * UI Automation is only used as a lightweight discriminator to determine whether `Shell_NotifyIconGetRect` currently points to the chevron or to the actual Recycle Bin icon.
  * Once the real icon becomes available in the overflow flyout, normal D&D resumes.
  * This was tested with the overflow already open and also with Windhawk's **Tray Hover Expand** during the same continuous drag (`^` → flyout opens → actual Recycle Bin icon).
  * The UIA path is fail-closed for destructive drops.

* **GUID notification icon fallback**

  * Added a fallback for cases where the GUID identity cannot initially be registered, matching the approach used by other Windhawk mods.
  * Once the GUID path has successfully validated during the process lifetime, failures no longer switch identity.
  * Tested both the normal GUID path and a forced fallback path, including Explorer restart and D&D after re-registration.

* **Per-monitor drag threshold**

  * Replaced the global `GetSystemMetrics` drag threshold with `GetSystemMetricsForDpi`, using the tray DPI.

* **D&D lifecycle / `Hide when empty`**

  * D&D is now effectively enabled only when:
    `enableDragDrop && !hideWhenEmpty`
  * This matches the documented behavior and avoids keeping the global low-level mouse hook and OLE drop target active when `Hide when empty` is enabled.
  * The user's D&D preference itself is preserved, so disabling `Hide when empty` automatically restores D&D if it was enabled.
  * Verified through full visible/hidden state transitions on Windows.

* **Cleanup**

  * Removed the redundant `shobjidl.h` include.
  * Updated D&D documentation and troubleshooting.
  * The closed `^` chevron is no longer documented as a valid drop target.
  * Clarified the `Confirm before emptying` behavior relative to the Windows Recycle Bin confirmation setting.
  * Added troubleshooting guidance for custom font glyph names / hexadecimal code points.

### Notes on the remaining review suggestions

* Theme-change handling was verified with `WM_SETTINGCHANGE / ImmersiveColorSet`; no additional workaround was needed.
* Vector icon variants are intentionally kept as a product/design choice.
* Additional image resampling changes are deferred unless a real rendering issue is reported.
* Font fitting was benchmarked/tested and is intentionally unchanged for v1.
* `SHQueryRecycleBin` performance changes are deferred unless a reproducible slowdown appears.
* Shell32 resource/localization concerns were already addressed in the earlier localization work.
* No real conflict was reproduced with other tray customization mods; the relevant overlap concern remains non-reproducible.

All critical points from the previous review have now been either fixed and tested on Windows, or intentionally classified as non-blocking / future work based on reproducibility.

The public mod version remains **1.0.0**, the mod ID remains **`recycle-bin-tray`**, and the display name is now **Recycle Bin Tray**.

PR #5331, opened 2026-09-03
Add a mod to highlight recently used icons and previews on the toolbar

jvlasek:
taskbar-recent-focus-highlight.wh.cpp
## taskbar-recent-focus-highlight
This mod highlights taskbar icons of recently focused apps and also highlights the recently focused windows inside the preview flyout. 

## Screenshots
Configured with left bar highlighting last 3 apps:
image

Preview pane highlighting the last focused (right) and the one before (left) window inside the preview flyout:  
image

## Mod authorship

This mod was created by:
- - [X] Another AI (please specify):  Grok Build

### DIFF
--- a/mods/taskbar-recent-focus-highlight.wh.cpp
+++ b/mods/taskbar-recent-focus-highlight.wh.cpp
@@ -2,7 +2,7 @@
 // @id              taskbar-recent-focus-highlight
 // @name            Taskbar Recent Focus Highlight
 // @description     Visually highlight the most recently focused running apps on the taskbar
-// @version         0.9.1
+// @version         0.9.2
 // @author          Jakub Vlášek
 // @github          https://github.com/jvlasek
 // @include         explorer.exe
@@ -64,9 +64,9 @@ clear highlights without unloading the mod.
 
 - Works with Windows 11 Taskbar Styler: the side bar stays visible on hover,
   and preview plate tints are restored when the highlight is cleared.
-- Icon matching prefers the process path from the taskband. Name matching is
-  a fallback and is based on English taskbar labels (`N running`, `pinned`),
-  so it can be weaker on a non-English Windows.
+- Icon matching uses the process path / AppUserModelID from the taskband
+  (not the localized button label). If that resolve is unavailable, the icon
+  is left unhighlighted rather than guessed from its name.
 - Multi-monitor: the same rank is applied on every taskbar that shows that
   app.
 */
@@ -300,7 +300,9 @@ clear highlights without unloading the mod.
 #include <functional>
 #include <memory>
 #include <mutex>
+#include <optional>
 #include <string>
+#include <string_view>
 #include <unordered_map>
 #include <unordered_set>
 #include <vector>
@@ -512,9 +514,6 @@ GUID g_currentDesktopId{};
 bool g_haveCurrentDesktop = false;
 PendingFocus g_pendingFocus;
 
-// process key (UPPER path) -> last seen AutomationProperties.Name of its button
-std::unordered_map<std::wstring, std::wstring> g_keyToAutomationName;
-
 std::atomic<ULONGLONG> g_windowConfirmSeq{0};
 
 // Public IVirtualDesktopManager (fallback when registry current-desktop fails).
@@ -561,7 +560,6 @@ std::vector<winrt::weak_ref<FrameworkElement>> g_trackedThumbViews;
 
 std::atomic<bool> g_unloading{false};
 std::atomic<bool> g_taskbarViewDllLoaded{false};
-std::atomic<bool> g_taskbarDllHooked{false};
 // After decay / empty ranks, force-clear overlays on next button touch if
 // RequestApplyVisuals couldn't run (sleep/wake, no dispatcher yet).
 std::atomic<bool> g_pendingOverlaySweep{false};
@@ -570,6 +568,13 @@ std::mutex g_winEventHookThreadMutex;
 std::atomic<HANDLE> g_winEventHookThread{nullptr};
 std::atomic<HWND> g_hookThreadHwnd{nullptr};
 std::atomic<DWORD> g_hookThreadId{0};
+HANDLE g_hookThreadReadyEvent = nullptr;
+
+void SignalHookThreadReady() {
+    if (g_hookThreadReadyEvent) {
+        SetEvent(g_hookThreadReadyEvent);
+    }
+}
 
 HWND HookThreadWindow() {
     return g_hookThreadHwnd.load(std::memory_order_acquire);
@@ -593,11 +598,14 @@ winrt::weak_ref<FrameworkElement> g_dispatcherAnchor;
 // Agile CoreDispatcher list — captured on the UI thread when we first see a
 // button/thumb/panel. CollectUiDispatchers must not weak.get() XAML objects
 // off the UI thread (last-ref ~FrameworkElement would run on the wrong thread).
+// no_destroy: process teardown must not ~CoreDispatcher after XAML is gone.
 std::mutex g_dispatchersMutex;
-std::vector<winrt::Windows::UI::Core::CoreDispatcher> g_uiDispatchers;
+[[clang::no_destroy]] std::optional<
+    std::vector<winrt::Windows::UI::Core::CoreDispatcher>>
+    g_uiDispatchers{std::in_place};
 
 void RememberUiDispatcher(FrameworkElement el) {
-    if (!el) {
+    if (!el || g_unloading.load()) {
         return;
     }
     try {
@@ -606,31 +614,29 @@ void RememberUiDispatcher(FrameworkElement el) {
             return;
         }
         std::lock_guard<std::mutex> lock(g_dispatchersMutex);
-        for (const auto& existing : g_uiDispatchers) {
+        if (!g_uiDispatchers) {
+            return;
+        }
+        for (const auto& existing : *g_uiDispatchers) {
             if (existing == dispatcher) {
                 return;
             }
         }
-        g_uiDispatchers.push_back(dispatcher);
+        g_uiDispatchers->push_back(dispatcher);
     } catch (...) {
     }
 }
 
 constexpr UINT WM_APP_FOREGROUND_CHANGED = WM_APP + 1;
-constexpr UINT WM_APP_REQUEST_APPLY = WM_APP + 2;
-constexpr UINT WM_APP_REQUEST_PREVIEW_APPLY = WM_APP + 3;
-constexpr UINT WM_APP_DESKTOP_SWITCHED = WM_APP + 4;
-constexpr UINT WM_APP_SHUTDOWN = WM_APP + 5;
-constexpr UINT WM_APP_REQUEST_APPLY_DEBOUNCED = WM_APP + 6;
+constexpr UINT WM_APP_DESKTOP_SWITCHED = WM_APP + 2;
+constexpr UINT WM_APP_SHUTDOWN = WM_APP + 3;
+constexpr UINT WM_APP_REQUEST_APPLY_DEBOUNCED = WM_APP + 4;
 
-// Identity scores. Only exact identity (and name-cache) may bind the same
-// rank to many buttons (secondary taskbar / Never Combine). Score 900 is
-// same filename, different folder — 1:1 so two python.exe installs stay
-// distinct.
+// Identity scores. Only exact identity may bind the same rank to many buttons
+// (secondary taskbar / Never Combine). Score 900 is same filename, different
+// folder — 1:1 so two python.exe installs stay distinct.
 constexpr int kScoreExactIdentity = 1000;
 constexpr int kScoreSameFileDifferentPath = 900;
-constexpr int kScoreNameCache = 96;
-constexpr int kScoreMinBind = 70;
 constexpr UINT_PTR kMinFocusTimerId = 1;
 constexpr UINT_PTR kDecayTimerId = 2;
 constexpr UINT_PTR kPreviewMinFocusTimerId = 3;
@@ -834,6 +840,9 @@ bool DropPendingIfWrongDesktopLocked() {
 }
 
 void OnVirtualDesktopSwitched() {
+    if (g_unloading.load()) {
+        return;
+    }
     ClearButtonRunningGrace();
     RefreshCurrentDesktopId();
     bool droppedPending = false;
@@ -921,7 +930,7 @@ std::wstring ToUpper(std::wstring s) {
     return s;
 }
 
-// Keep A–Z / 0–9 only, uppercased — for fuzzy exe ↔ automation-name match.
+// Keep A–Z / 0–9 only, uppercased — preview unique-title compare.
 std::wstring AlnumUpper(std::wstring_view s) {
     std::wstring out;
     out.reserve(s.size());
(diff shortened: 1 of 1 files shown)


jvlasek:
Generated review comment: 

Thanks — two leftovers from the last restore/resolve work, and the title fallback stays with a note.

**1. Unique-title preview fallback — keeping it, with the note you asked for.**

Icon matching is already fail-closed (path / AUMID / HWND). The leftover name code is **preview-only**, and only when the exact paths miss.

Pass 1 (`Thumbnails.GetAt`) and pass 2 (`DataContext` ↔ ctor map) are optional symbols. On builds where they are missing, unique-title is how a multi-window flyout still gets a ladder. Identical titles stay unmatched — we do not steal. The English `" running"` / `" pinned"` strip is a no-op on other languages, which means *fewer* unique-title binds, not a wrong bind.

Pass 4 is not a bind: it copies recency onto an HWND that pass 1/2 already assigned (Lister / tab-proxy HWND vs `EVENT_SYSTEM_FOREGROUND` HWND). No string matching.

Wrong glow on an **icon** is worse than none. A blank multi-window flyout when GetAt is unavailable is worse than a unique-title hit. `how=repeater|taskitem|title` is in the log if a human review wants that fallback dropped from evidence.

**2. Stale global `Thumbnails` collection — agreed, fixed.**

Confirmed: `g_TaskGroup_Thumbnails` is only captured inside `TargetItemKey`, and the flyout refresh also runs from `OnApplyTemplate`, title remeasure, decay, and click confirm. Same window *count* as the previous app is enough for the size guard to pass, then every card gets the previous app’s HWNDs.

DataContext → ctor map is per-card and cannot go stale. We now run that **first**. Repeater `GetAt` only fills holes, and only if it agrees with at least one DataContext HWND on the same card (mismatch ⇒ skip the index path, log it). Order still comes from the repeater.

**3. Unnamed snapshot children — agreed, fixed.**

The snapshot now stores `weak_ref<UIElement>` in visual order, including unnamed children, skipping only our glow host. Restore is by object identity, so a Styler element without `x:Name` keeps its place, and two children with the same name cannot swap.

### Optional

- **`ScoreButtonForRank`.** Agreed — unused wrapper, deleted.
- **Init when nothing to paint.** Agreed — `highlightCount == 0` and previews off → `Wh_ModInit` returns `FALSE`; settings change reloads.
- **Debounce hoist + `g_unloading`.** Agreed.
- **`TryRunAsync` for flyout refresh.** Agreed.
- **`GetProcessImagePath` twice per FG.** Agreed — same-thread cache for the duration of `HandleForegroundChanged`.
- **`g_TaskGroup_Thumbnails` on uninit.** Agreed — reset with the other maps.

### Functionality notes

- **Demote-forever.** Skipping. Fail-closed on identity is the catalog rule.
- **Snap-group second icon.** Agreed — if `IconsRepeater` has 2+ children *and* the card still resolves a window HWND, it is not treated as a snap group.
- **Paint-cache key vs host size.** Agreed — the key now uses the same `GlowContentBoxSize` fallback (existing host, else `BackgroundElement`, else panel).
- **`ButtonHasOurChrome` descendant walk.** Agreed — `IconPanel` direct child only; the nested-host loop in `ClearButtonHighlight` already covers the odd case.

/ai-review

jvlasek:
Generated review comment:

Thanks — three leftovers in preview resolve / min-focus, then the cheap optionals.

**1. Unique-title not scoped to the flyout — agreed, fixed.**

Confirmed: pass 3 scored every confirmed HWND on the desktop, and `ScoreTitleToAutomationName` 91/93 is substring with a 4-character floor. `usedHwnds` only prevents stealing a card already bound in *this* flyout, so a foreign-process HWND was free to take the glow.

Pass 3 now:

- Anchors on any sibling that resolved by TaskItem or repeater **and** is already in this desktop’s window recency map.
- Filters `recent` to that `processKey` (taken from the map — no `OpenProcess` / `SHGetPropertyStoreForWindow` on the flyout UI thread).
- Skips entirely when nothing resolved exactly (`Preview resolve: skip unique-title`).

Hosted UWP still share ApplicationFrameHost’s image path in that map; unique-title only fills holes after an exact sibling HWND, so Calculator vs Settings that already bound by TaskItem/repeater stay exact. Chrome `GitHub` cannot inherit Edge `GitHub - Profile`.

Wrong glow on a card is worse than a blank flyout when the exact paths missed. Unique-title stays as the last hole-filler for the *same* process, with the note you asked for last round.

**2. Vacuous index-bind cross-check — agreed, fixed.**

Confirmed: `indexBindTrusted` only flipped on disagreement, so “no DataContext HWND” left GetAt trusted. `g_TaskGroup_Thumbnails` is only captured inside `TargetItemKey`, but the refresh also runs from `OnApplyTemplate` / decay / click.

We now `g_TaskGroup_Thumbnails = {}` on the way **in** to `TargetItemKey`, so a non-null collection is from *this* retarget. DataContext-first + disagreement skip stay as a second check. We did not add `anyTaskItemResolved` — that would disable GetAt on the builds where it is the only exact path.

**3. 200 ms transient poll — agreed, fixed.**

Confirmed: after the deadline, both timers re-armed 200 ms with no cap while `IsTransientForeground` was true, and File Explorer is this `explorer.exe`.

Post-deadline wait is now `TransientGraceMs` (same as that timer’s min-focus, at least 2 s) of 200 ms polls, then the app timer drops the candidate (`gave up waiting through transient`). Preview give-up only stops re-arming (the app timer still owns `g_pendingFocus`). A real same-app FOREGROUND clears the grace counter.

### Optional

- **Snap-group vector.** Agreed — `IsSnapGroupThumbnailView` once per flyout, reused for the sibling filter, size check, and GetAt compact.
- **Keep overlay host.** Agreed — ranked repaint restores the plate brush and hides children; `XamlReader::Load` only when the host is missing. Unload / unranked still full-clear. Overlay children use Margin, not RenderTransform (layout clip squared the right end-cap).
- **Debounce restart.** Agreed — `WM_APP_REQUEST_APPLY_DEBOUNCED` skips `SetTimer` if the 300 ms full-rebind timer is already armed.
- **Init log version.** Agreed — `Wh_Log(L"> init " WH_MOD_VERSION)`.
- **`weak.get()` then `HasThreadAccess`.** Skipping. Primary and secondary taskbars share one Explorer UI thread, so the dispatcher filter never fires. Changing it without a second UI thread to drain would be speculative.

### Functionality notes

- **File Explorer.** Agreed — `CabinetWClass` / `ExploreWClass` are a real app (ranked, and a leave from another candidate). Tray / desktop / Start / IME stay ignored via `ShouldIgnoreHwnd`. Documented in the README.
- **Unique-title keep/drop.** Still keeping it, now process-scoped (finding 1). Human call if it should go entirely.
- **Idle decay extra cycle.** Harmless; we now also `StopDecayTimerIfIdle` when the app timer drops a candidate.


/ai-review

jvlasek:
Generated comment:
Thanks — two leftovers, then the optionals.

**1. Ring preview style — removed.**

Confirmed: `min(cardW, cardH)` made a square on the left of a wide card. Hybrid / plate / title wash already look right, so rather than ship a second-rate frame we dropped `ring`. A saved `previews.style=ring` loads as hybrid (`plateTitle`).

**2. Packaged-app unique-title scope — agreed, that was a real miss.**

App ranks already use `APPID:<aumid>` for AFH/WWAHost. Window recency stamped `PathFromAppKey` and then the host exe, so Calculator and Settings shared `ApplicationFrameHost.exe` in pass 3. We now stamp the same rank key we already built (`pending.key` / `key`) — path for Win32, `APPID:` for hosted UWP. The README AUMID line is now what the code does. (We thought this was already in; it was only on the app map.)

### Optional

- **Stale-identity `OpenProcess` every rebind.** Agreed — the `haveIdentity` branch skipped the `kUnresolvedRetryMs` throttle. Stale check is now throttled the same way, and a negative result bumps `lastResolveTick` so we don’t retry every 300 ms.
- **`g_currentDesktopId` log after unlock.** Agreed — capture the GUID while we hold `g_stateMutex`.
- **`#ifndef WH_MOD_VERSION`.** Agreed — deleted. Windhawk always defines it.
- **`ClearThumbnailHighlight` walks.** Agreed — return early when the card has neither our host nor the plate marker.
- **Idle decay vs leftover `g_pendingFocus`.** Agreed — preview confirm clears pending when the app side is already in the map, then `StopDecayTimerIfIdle`.
- 
### Functionality notes

- **Shared grace tick (defaults 8 s app / 1 s preview).** That *is* the default config. Preview timer fires at 1 s, sets `transientRetryStartTick`, and at t=8 s the app timer sees ~7 s already elapsed against an 8 s grace, so a new app you Alt-Tab away from only gets ~1 s of extra wait instead of 8. Separate `previewTransientRetryStartTick` now, so each timer has its own grace.
- **Same-app unique-title (`GitHub` vs `GitHub - Pull requests`).** Yes, score 91 substring can still bind that when the card’s own window was never confirmed. Same-process only, holes only, documented. Keep/drop still a human call.
- **OverlayIcon after snapshot (saw the issue on a Thunderbird badge).** `RememberNativeIconPanelOrder` runs once at first glow. A late `OverlayIcon` (mail count) isn’t in the list; snapshot restore on clear compacted it behind the glyph — and that path returned before `EnsureOverlayIconAboveGlyph`. Restore now raises OverlayIcon after compaction, and a cached-rank UVS paint does the same so a badge that appears while ranked isn’t left behind the icon.
- **Ring vs `glowLayers`.** Moot — ring is gone.

/ai-review

jvlasek:
Generated comment: 
Thanks — two real bugs, then the rest. 

**1. `TryRunAsync` bool-cast — agreed, that would hang uninit.**

`static_cast<bool>(TryRunAsync(...))` only tests that the operation object is non-null. We read “was it queued?” from `Status()` / `GetResults()` (same pattern as `taskbar-vertical`), including `IAsyncAction` when the SDK returns that instead of `IAsyncOperation<bool>`.

For the uninit drain the Low callback is empty and does not signal. `Completed` is the barrier: it always `SetEvent`s, and the operation is kept alive until after `WaitForSingleObject`. Signaling from the Low callback let unload return while `Completed` was still inside the mod. `Completed` is assigned once. If the operation cannot be subscribed, we signal immediately and record the failure.

The same queued-check is on `RunOnUiThread`, `SchedulePreviewFlyoutRefresh` (a failed queue clears the flyout latch), and `ScheduleRefreshAllHighlights`.

**2. Init stays loaded when hooks fail — agreed, with one caveat.**

- `taskbar.dll` identity hooks fail → `return FALSE` (Windhawk shows the error and reloads on the next settings change).
- `Taskbar.View.dll` is already loaded but its hooks fail → `return FALSE`.
- `g_taskbarViewDllLoaded` is set only after a successful View hook, so `Wh_ModAfterInit` / `LoadLibraryExW` can retry.

We do not `return FALSE` when Taskbar.View is simply not loaded yet. Explorer often loads it after `Wh_ModInit`. That is the same late-load path as `taskbar-thumbnail-reorder` (`Taskbar view module not loaded yet` + `LoadLibraryExW`). Failing init there would make the mod never attach on a normal Explorer start.

**3. Title-based matching — dropped.**

Unique-title HWND assignment and the title-score path are gone. A preview card matches TaskItem/DataContext first, then repeater `GetAt` for holes only. A card that misses both stays unmarked. The removed code is in `stash/preview-unique-title.cpp`.

**4. Line count.**

- Identity scoring is `bool IdentityMatchesRank` (it only ever returned 0 or 1000).
- Per-dispatcher fan-out stays. Primary and secondary taskbars share one Explorer UI thread; collapsing to a single dispatcher without a second thread to drain would be speculative.

### Optional

- **Icons thickness/roundness on previews.** Called out under Previews → Style. Title-bar thickness follows Icons → Thickness. The native preview card keeps its own corners; the overlay plate fallback uses Icons → Roundness. Side and edge icon bars do not use Roundness.
- **`CLSCTX_INPROC_SERVER`.** Agreed.
- **Log volume.** Agreed — dropped the per-button dump. The rank list log stays.
- **`TargetItemKey` flag if original throws.** Agreed — RAII reset. The hook also clears the captured thumbnails collection on entry.

### Functionality notes

- **Native `IconPanel` reorder.** The side bar has to sit above a Taskbar Styler hover plate and under OverlayIcon. A purely additive overlay does not express those constraints. The repair layer stays. OverlayIcon is raised after the snapshot restore.
- **Identity on the UI thread.** Full bind and press only, throttled, including the stale-path check. Not moving that COM onto the focus thread (that would resolve XAML off the UI thread).
- **Uninit `INFINITE`.** Intentional. Finding 1 was the hang. A timeout with a live thread or a live `SizeChanged` is worse.
- **Registry desktop id lag.** Known. The next foreground event or desktop switch refreshes it.

/ai-review

jvlasek:
Generated response: 
Thanks. The immediate `SetEvent` was a real unload hole. This is v0.9.34.

**1. Unload could return while cleanup was still queued — agreed. The High-operation fallback is not what we shipped.**

High work is allowed to run ahead of Normal work already in the queue. Waiting for the High cleanup to finish proves that callback returned. It does not prove those older callbacks have finished, and they still execute in this mod. A failed `Completed` subscription is also not the same thing as “nothing was queued.”

Unload now returns only when one of these is true:

- The Low sentinel’s operation completed and `GetResults()` is true, so that callback ran. It is posted after the High cleanup, so it runs after that cleanup and after Normal work already queued.
- The dispatcher object is disconnected or closed (`RPC_E_DISCONNECTED`, server died, not connected, or `RO_E_CLOSED`). Further posts cannot be observed. That is a weaker exit than a sentinel that ran, and it is logged. A generic exception is not this case.

`Completed` still signals the waiter when it can be set. If it cannot, we poll `Status()`. A status read that throws is not treated as completion. `HasThreadAccess` and `ProcessEvents` use the same split, so a throw from those calls does not escape `Wh_ModUninit`.

`TryRunAsync` completing with `false` is the documented shutdown rejection. It means that post was not accepted. It does not show that work already queued will not run, so it is not a drain and it does not end the wait. There is no two-second or thirty-second path that unloads anyway.

**2. Line count — leaving it for the human reviewer.**

Not collapsing the per-dispatcher helpers. Primary and secondary taskbars share one Explorer UI thread here, but one dispatcher would not make the unload wait correct. App highlights and preview highlights both stay.

### Optional

- **`taskbar.dll` reference on init failure.** Agreed. `ReleaseTaskbarDllIfWeLoadedIt` runs on the `return FALSE` paths after a load, and from `Wh_ModUninit`. Init failure does not call uninit.
- **Discarded `DispatcherOpWasQueued` inside `Completed`.** Removed. The handler only signals the waiter.
- **Stash path in a comment.** Removed from the submitted file. Preview cards that miss TaskItem/DataContext and repeater `GetAt` stay unmarked.
- **`g_winEventHookThread` atomic plus mutex.** Leaving both. The mutex is the start/stop protocol. The atomic is the handle the stop path exchanges. Not worth a shutdown change for the redundancy.
- **Thickness labelled “px”.** Leaving the label. The value is a XAML layout unit. Changing the visible name does not change the bar.

### Functionality notes

- **Residual tail after `SetEvent`.** Acknowledged. The sentinel running is the drain we can prove. The few instructions that unwind the delegate after `SetEvent` are not something this mod can close.
- **Identity lookup on the click path.** Not splitting it this round. Full bind stays off the hover path. Moving only the process and property-store calls to the worker is possible later if a hung click shows up.
- **`IconPanel` child order.** Still required. The side bar has to sit above a Taskbar Styler hover plate and under `OverlayIcon`. We restore from a snapshot and raise `OverlayIcon` after our own change. We do not reorder a button we never painted. A badge the app moves after that can still land behind the glyph. We are not putting a reorder back on the hover path.
- **Registry desktop id.** Unchanged. Read-only, on the worker thread. The next foreground event or desktop switch refreshes a lagging value.
- **Two `INFINITE` waits.** Intentional. A timeout that returns while the focus thread or a dispatcher callback is still in this image is the worse failure.

Generated by AI. Model: Grok 4.7

/ai-review

PR #5352, opened 2026-09-04
MuteAlert v0.9.8: Add Google Meet integration

Nikolay1243:
## What changed

- Add Google Meet call and mute-state detection through Windows UI Automation.
- Add speaking-while-muted warnings, optional audio cue, taskbar mute/unmute, call-window focus, and headset mute synchronization.
- Add a distinct Google Meet call-state logo and configurable localized control labels.
- Recognize Meet windows in Chrome, Edge, Firefox, Brave, Vivaldi, Opera, and Arc.
- Document that the meeting must be the active, visible browser tab because inactive tabs don't expose their controls to UI Automation.

## Validation

- Compiled successfully with the repository's official compile_mod.py script for x86-64 and ARM64.
- git diff --check passes.

### DIFF
--- a/mods/mutealert.wh.cpp
+++ b/mods/mutealert.wh.cpp
@@ -2,7 +2,7 @@
 // @id              mutealert
 // @name            MuteAlert - Microphone Activity Taskbar Widget
 // @description     Shows live microphone activity, call mute state, volume controls, and headset mute synchronization in the Windows 11 taskbar.
-// @version         0.9.7
+// @version         0.9.8
 // @author          Nikolay
 // @github          https://github.com/Nikolay1243
 // @homepage        https://github.com/MuteAlert/windhawk
@@ -30,13 +30,14 @@ Adds a microphone button to the Windows 11 system tray area.
 * Optionally lock the microphone at a chosen volume. Scrolling updates the
   locked target when this is enabled.
 * Left-click the button to mute or unmute the microphone.
-* A secondary Slack, Teams, or Zoom logo shows the active call microphone
+* A secondary Slack, Teams, Zoom, or Google Meet logo shows the active call microphone
   state and focuses the call window when left-clicked.
 * Headset mute synchronization supports Windows hardware mute reporting,
   standard USB HID mute controls, and vendor-specific adapters. SteelSeries
   Arctis Nova Pro Wireless is the first high-confidence vendor adapter.
-* If you speak while muted in a Slack huddle, Microsoft Teams call, or Zoom
-  meeting, the widget warns you and can play an optional audio cue.
+* If you speak while muted in a Slack huddle, Microsoft Teams call, Zoom
+  meeting, or Google Meet, the widget warns you and can play an optional
+  audio cue.
 * Right-click the widget to mute or unmute active call apps when that action
   is shown in the tooltip.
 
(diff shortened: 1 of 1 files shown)


Nikolay1243:
Addressed the two main findings in 5a68aee1:

- Window title fragments are now configurable. Defaults match Meet with ASCII hyphen, en dash, or em dash anywhere at a word boundary, including notification-prefixed titles. The broad `google meet` substring was removed. A rate-limited diagnostic explains when capture is present but no eligible window title matches, without logging titles.
- Meet eligibility now retains individual active capture PIDs. Browser windows must belong to a capturing process or its verified same-executable ancestor. This handles audio-service child processes without enabling scans in other browser instances. Parent creation times are checked to reject reused parent PIDs. Inactive browser capture sessions are excluded, and capture-enumerator failure does not enable unrestricted browser scanning.

Meet now requires an explicit `meetEnabled` opt-in, including for the shared badge and headset synchronization. Executable filenames are configurable and matched exactly; the speculative `opera_gx.exe` entry was removed. The Zoom-only fallback comment and condition agree again, and Meet no longer accepts offscreen controls.

The README documents accessibility costs and explicitly notes that Firefox has not been verified in a live meeting. When capture stops or controls disappear, detection clears. Windows sharing one browser process cannot be distinguished by audio-session PID alone; they still require a matching title and in-call controls.

Validation: the official repository compiler passed for x64 and ARM64, and `git diff --check` passed. Live browser/call testing remains outstanding. I left the optional provider-table refactor and screenshot update out of this fix.

Nikolay1243:
Addressed both performance findings:

- A shared `MeetMonitoringEnabled()` predicate now requires the opt-in AND at least one consumer (badge, warning, right-click action, or effective headset call synchronization). It is used for thread startup, polling, and browser classification, so another app's monitoring thread cannot trigger Meet process snapshots when no Meet feature consumes the result.
- The parent-process snapshot is loaded lazily on the first active browser capture session and reused across the three endpoint-role checks in that polling cycle. With no eligible browser capture there is no process snapshot. The loaded flag resets on each cycle and during cleanup.
- Browser executable tokens are trimmed and lowercased once in `LoadSettings`, then compared as exact filenames during scanning.

Also addressed the optional empty-filter defaults (including whitespace/separator-only values) and clarified that the no-title-match diagnostic applies only if the user is actually in a Meet call.

Validation: official repository compilation passed for x64 and ARM64; `git diff --check` passed. Live Meet mute/capture behavior remains unverified, as documented. The optional element-cache redesign is outside this focused performance fix.

Nikolay1243:
The remaining settings dependency finding is valid and is now addressed:

- The Meet warning, audio cue, right-click control, shared call badge, and headset call synchronization descriptions explain the Meet opt-in requirement.
- The master switch description explains that a consuming feature must also be enabled. The audio cue description additionally states its dependency on the speaking-while-muted warning.
- Settings-load diagnostics cover Meet features configured with the master switch off, the master switch on with no consuming feature, and an audio cue configured without its warning. These are emitted on settings load, not in the polling loop. Shared Slack/Teams/Zoom features alone do not produce misleading disabled-Meet logs.

Validation: the official repository compiler passed for x64 and ARM64, and `git diff --check` passed. The optional refactor, logo polish, and element caching remain deferred; they are not required to correct this settings issue. Live-call validation of capture state while muted remains outstanding and documented.

PR #5369, opened 2026-09-05
Add Classic Tray Keyboard Layout Indicator

Anixx:
If the mods that do not support elevated windows at all are allowed, then why is this worse?

## Changelog

* Changelog item 1...
* Changelog item 2...

## Mod authorship

This mod was created by:

- - [x] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/classic-tray-keyboard-layout-indicator.wh.cpp
+++ b/mods/classic-tray-keyboard-layout-indicator.wh.cpp
@@ -1,0 +1,427 @@
+// ==WindhawkMod==
+// @id              classic-tray-keyboard-layout-indicator
+// @name            Classic Tray Keyboard Layout Indicator
+// @description     Tray indicator for keyboard layout, similar to internat.exe from Windows 2000
+// @version         6.1.0
+// @author          Anixx
+// @github          https://github.com/Anixx
+// @include         explorer.exe
+// @compilerOptions -lgdi32 -luser32 -lshell32
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Keyboard Layout Tray Indicator
+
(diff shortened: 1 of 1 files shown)


m417z:
I agree with items 3 and 6 that it looks like a good fit for a tool mod. Also, regarding `InvalidateRect_Hook`, I believe that `TrayInputIndicatorWClass` is no longer used in the Win11 taskbar, right? Have you tested it?

Anixx:
It is intended only for the Win10 taskbar, this is why it is also not a tool mod. It uses hacks of the taskbar to get info about layout changes (otherwise it could not know about layout changes in elevated windows). Basically, it spies about the existing layout indicator, even if it is hidden and when Explorer changes its look, the mod knows the layout has changed. Since it deeply hacks and hooks into the Win10 taskbar, It is more practical to make it inject Explorer rather than making it a toolmod.

Basically it uses hacks but can in most cases handle layoutr changes in elevated windows, at least when there are only two layouts installed.

m417z:
Have you tried the AI suggestion of using `HSHELL_LANGUAGE`?

In any case, add this info to the readme, and I'd also mention in the mod description that it's a Win10-only mod.

In Windhawk 2.0, I believe using `windhawk-mod-uiaccess.exe` can make it easier.

Anixx:
Uiaccess does not help such mods, I tried. It is for Win10 taskbar on Win10 or Win11...

> Have you tried the AI suggestion of using HSHELL_LANGUAGE?

Yes, everything like this. The AIs suggest the same things.

Anixx:
> In any case, add this info to the readme, and I'd also mention in the mod description that it's a Win10-only mod.

It's not Win10-only, it's for the classic taskbar, which can run under Win10 or Win11... It is mentioned already...

PR #5374, opened 2026-09-05
Add Nothing Track mod (nothing-track)

lenorio:
Adds **Nothing Track**, a native Windows 11 taskbar widget and flyout control for Nothing and CMF earbuds.

It communicates with the earbuds directly over Bluetooth RFCOMM (SPP) profile to display battery status and control ANC, Ultra Bass, EQ presets, Low Latency game mode, and Find My Earbuds without requiring a mobile app or Android emulator.

The PR adds a single file: `mods/nothing-track.wh.cpp`.

## Changelog

- Initial public release.

## Mod authorship

This mod was created by:

- [x] Claude
- [x] ChatGPT
- [ ] Gemini
- [ ] Other (please specify):

### DIFF
--- a/mods/nothing-track.wh.cpp
+++ b/mods/nothing-track.wh.cpp
@@ -6,7 +6,8 @@
 // @description:ru-RU Мониторинг батареи и управление (ANC, Ultra Bass, EQ, Low Latency) для наушников Nothing и CMF прямо в панели задач Windows 11
 // @version         1.0.0
 // @author          lenorio
-// @github          https://github.com/lenorio/Nothing-Track
+// @github          https://github.com/lenorio
+// @homepage        https://github.com/lenorio/Nothing-Track
 // @include         explorer.exe
 // @architecture    x86-64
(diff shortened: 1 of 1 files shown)


m417z:
Thanks for the submission.

* How about implementing suggestion 3 from the last review?
* How about implementing this to avoid memory leaks: **`[[clang::no_destroy]] static BluetoothManager s_instance;`** — the attribute is *correct and necessary* here (without it, `~std::thread` on a still-joinable `m_worker`/`m_actionWorker` calls `std::terminate()` at process shutdown, crashing Explorer on every sign-out), but the bare form is only recommended for nullable/handle-like types. For a plain class the documented shape is the `std::optional` wrapper, so the release is explicit: `[[clang::no_destroy]] static std::optional<BluetoothManager> s_instance;`, constructed on first use and `reset()` in `Wh_ModUninit` after `Stop()`. See [Global objects and process shutdown](https://github.com/ramensoftware/windhawk/wiki/Global-objects-and-process-shutdown#1-worker-thread-stdthread). (Your XAML globals — `g_injectedGrid`, `s_currentFlyout`, `g_dispatcherTimer`, … — are correctly using the bare form, since WinRT projected types are nullable.)
* Why the screenshot doesn't show the taskbar and has an error message?

PR #5375, opened 2026-09-05
Add Hide Empty Optical Drives mod

Solomag:
Adds a Windhawk mod that hides empty CD/DVD/BD drives from This PC while keeping the drive fully available to Windows.

The drive automatically reappears when media is inserted and is hidden again when the media is removed.

Tested on Windows 10 22H2 (19045).

## Changelog

## Mod authorship

This mod was created by:

- [ ] Claude
- [x] ChatGPT
- [ ] Gemini
- [ ] Other (please specify):


Solomag:
Addressed the latest AI review feedback and tested the updated version locally on Windows 10 22H2.

The main media-detection issue was fixed conservatively: after a media-arrival event, an inconclusive GetVolumeInformationW result is now treated as Unknown and fails open, so blank, unreadable, unsupported, or still-spinning media won't cause the drive to be hidden. Empty is only committed from actual media/device removal handling.

I did not switch to IOCTL_STORAGE_CHECK_VERIFY2 as suggested by the AI reviewer because earlier testing on the target optical drive showed it returning ERROR_NOT_READY even with media inserted, so using it as the source of truth would risk introducing false negatives.

Also addressed the optional review items:
- formatted the source according to the repository clang-format style;
- tightened drive-root parsing to X:\ only;
- added strict validation for the driveLetters setting;
- removed redundant refreshes and TranslateMessage;
- clarified README scope/limitations and added the demo GIF.

The current version has been tested through repeated empty -> insert -> eject cycles and behaves as intended.

/ready-for-reviewer

Solomag:
/ready-for-reviewer

Hi,

I went through the AI review findings and addressed the issues that appeared to be actual user-facing bugs.

The remaining findings are mostly edge cases or architectural suggestions rather than functional problems:

- The suggested DriveState/state-machine rewrite would be a larger refactor of the whole media detection logic. The current implementation already separates startup, arrival, removal and retry handling, and changing the model at this stage would significantly increase the risk of introducing new regressions.
- The remaining NotReady discussion concerns rare timing cases around startup/resume. The mod already uses bounded retries and does not permanently cache an incorrect state.
- The notification thread shutdown concern is a theoretical robustness issue during unload rather than a normal runtime problem.

The mod has gone through several review iterations and the remaining suggestions would mostly be improvements for a future refactor rather than blockers for this PR.

I would appreciate a human review to determine whether any of these remaining points should actually block submission.

Thanks.

PR #5377, opened 2026-09-06
Add Taskbar Network Lounge

cracken7:
## What this mod does

Puts a small network meter on the Windows 11 taskbar: live download/upload speed on
the left, cumulative traffic on the right, split by a thin line you can drag. Click
it for an acrylic details panel (adapter, link state, IPv4, speeds, totals, reset),
hover for a tooltip, right click for a menu.

## Why another network meter

The catalog already has `net-speed-taskbar`, and it is a good mod — I read it as a
reference and credit it in the readme. This one exists for a different reason:
**adapter selection when a VPN is running.**

With a tunnel up, the same bytes cross two interfaces — the tunnel sees the
plaintext, the physical NIC sees the encrypted copy. Summing every visible adapter
therefore reports a 1 GB download as 2 GB. Auto mode here resolves the default
route (`GetBestInterfaceEx` / `GetBestRoute2`) against the candidate set and counts
**exactly one** interface, so a VPN is followed while it holds the internet and
dropped when it does not. Because a byte count carried over from a different
connection describes nothing, the totals reset on a genuine source change, while a
transient disconnect (cable out, tunnel not up yet) is deliberately not treated as
one. `ResetOnSourceChange` turns that off for anyone who wants a lifetime total.

Everything else follows from wanting that to be usable: five vector arrow styles,
four text weights, a mouse-draggable divider that moves the line without shifting
any text, a details panel whose height is a floor rather than a fixed size, and a
fully translated Arabic settings UI via `$name:ar:` / `$description:ar:`.

## Implementation notes

- Tool mod, `@include explorer.exe`, own layered `CreateWindowInBand` popup over
  `Shell_TrayWnd`; GDI+ painting, `SetWindowCompositionAttribute` acrylic,
  `DwmSetWindowAttribute` rounded corners, per-monitor DPI aware.
- `GetIfTable2` sampled on a worker thread, deltas divided by
  `QueryPerformanceCounter` elapsed time. Repaint only when a *formatted* value
  changes, so sub-digit wobble does not wake the painter.
- Persistent totals live in Windhawk's per-mod storage (`Wh_SetBinaryValue`), not a
  file, so they disappear with the mod.
- Shutdown posts `WM_QUIT` to the stored UI thread id rather than to a window, so
  uninit cannot hang if it runs before the window exists.
- Text is drawn with greyscale `AntiAliasGridFit` plus a 1 px contrast shadow and
  `GenericTypographic` formatting; ClearType on acrylic measured 112 fringed
  pixels, and the default string format's side padding was clipping `12.4 MB/s`.

## Verification

Built clean with `-Wall -Wextra` (clang 20, Windhawk 1.7.3). Eight offline test
suites pass: formatters and persistence round trip, taskbar geometry for every
edge, 1500 paint cycles with no GDI/USER handle growth, measured bytes vs the
Windows performance counters (0.06% over 22 s), every arrow style × text weight
actually rendering, text sharpness per hint, worst-case string fit, and
details-panel layout measured from rendered pixels. Live-tested on Windows 11
22631, 1920×1080 at 100%, over Ethernet and over a running VPN tunnel, including
Ethernet → VPN → Ethernet switching. CPU 0.3–0.5% of one core, ~30 MB working set.

`.github/pr_validation.py` passes locally with no warnings.

Source, tests and tooling: https://github.com/cracken7/TaskbarNetworkLounge

## Changes since the AI review

All six required items are addressed, plus most of the optional ones:

- `Appearance.TextColor` was declared as an int but read as a string, so a manual
  colour was misread — it is a quoted string setting now.
- Removed the redundant `BoldText` setting, superseded by `TextWeight`.
- Persistent totals moved from a file under `%LOCALAPPDATA%` into Windhawk's
  per-mod storage; the custom file format and its checksum are gone.
- `IDM_WH_SETTINGS` used `GetModuleFileNameW(nullptr, …)`, which returns
  `explorer.exe` in a tool-mod host, so "Windhawk settings" opened Explorer.
- Uninit could hang if it ran before the window existed; it now posts `WM_QUIT` to
  the recorded UI thread id, with explicit teardown after the loop.
- Replaced the broadcast `WM_SETTINGCHANGE` with a private `WM_APP` message.
- Removed dead `g_running` state, hoisted the DPI entry points into function-local
  statics (they were resolved on every paint and twice per `WM_MOUSEMOVE`), split
  hover state per window, dismissed the panel via a poll timer because
  `SetForegroundWindow` is refused for a background process, and fixed a typo in
  the virtual-adapter keyword list (`wan minipor`).
- Rewrote the readme and all setting descriptions in plain language, in both
  languages, and credited `taskbar-music-lounge` and `net-speed-taskbar` as design
  references. My first draft of this PR text leaned too heavily on
  `net-speed-taskbar`'s readme, which I had open while writing — that was my
  mistake and it is rewritten.

### DIFF
--- a/mods/taskbar-network-lounge.wh.cpp
+++ b/mods/taskbar-network-lounge.wh.cpp
@@ -26,8 +26,6 @@ shell.
 
 ![Widget](https://raw.githubusercontent.com/cracken7/TaskbarNetworkLounge/main/docs/widget.png)
 
-![Details panel](https://raw.githubusercontent.com/cracken7/TaskbarNetworkLounge/main/docs/panel.png)
-
 ## Features
 
 **Live speed and traffic totals.** Read from the real interface counters
--- a/mods/taskbar-network-lounge.wh.cpp
+++ b/mods/taskbar-network-lounge.wh.cpp
@@ -72,6 +72,10 @@ Taskbar Settings → Widgets.
   percent; the average over a second matches (measured 0.06 % over 22 s).
 * Source, tests and full documentation:
   [github.com/cracken7/TaskbarNetworkLounge](https://github.com/cracken7/TaskbarNetworkLounge)
+* Design references: the window creation, z-band placement and acrylic treatment
+  follow the approach used by **Taskbar Music Lounge** and **Taskbar Network Speed
+  Indicator** (`net-speed-taskbar` by NarayanChetri), both read while building this.
+  No code was copied from either.
 
 ---
 
@@ -125,6 +129,10 @@ Taskbar Settings → Widgets.
   واجهة ويندوز عربية.
 * الكود والاختبارات والشرح الكامل:
   [github.com/cracken7/TaskbarNetworkLounge](https://github.com/cracken7/TaskbarNetworkLounge)
+* مراجع التصميم: طريقة إنشاء النافذة ووضعها في نطاق الـz والمظهر الزجاجي تتبع ما
+  يفعله مودَا **Taskbar Music Lounge** و**Taskbar Network Speed Indicator**
+  (`net-speed-taskbar` لـNarayanChetri)، وقد قرأتُهما أثناء بناء هذا المود، ولم
+  يُنسَخ أي كود منهما.
 */
 // ==/WindhawkModReadme==
 
@@ -359,10 +367,13 @@ Taskbar Settings → Widgets.
 // Taskbar Network Lounge - implementation
 //
 // Design notes:
-//  * Architecture mirrors Taskbar Music Lounge: a Windhawk *tool mod* that owns
-//    a native layered popup window docked over Shell_TrayWnd, drawn with GDI+,
-//    acrylic via SetWindowCompositionAttribute, rounded via DWM.
-//  * All media/GSMTC logic is replaced by IP Helper network counters.
+//  * This is a Windhawk *tool mod* that owns a native layered popup window docked
+//    over Shell_TrayWnd, drawn with GDI+, acrylic via
+//    SetWindowCompositionAttribute, rounded via DWM. That window/z-band/acrylic
+//    approach follows what taskbar-music-lounge and net-speed-taskbar do; both
+//    were read as references while building this, no code was taken from either.
+//  * The network measurement, adapter selection, traffic totals and details panel
+//    are specific to this mod.
 //  * winsock2.h MUST be included before iphlpapi.h/netioapi.h, otherwise
 //    mingw-w64 hides MIB_IF_ROW2 / GetIfTable2 (they are only declared when
 //    _WS2IPDEF_ is defined).
--- a/mods/taskbar-network-lounge.wh.cpp
+++ b/mods/taskbar-network-lounge.wh.cpp
@@ -20,119 +20,234 @@
 /*
 # Taskbar Network Lounge
 
-A native network meter docked on the taskbar. Live download/upload speed, total
-traffic, and an acrylic details panel — styled like the rest of the Windows 11
-shell.
+A small network meter that sits on your taskbar and shows how fast you are
+downloading and uploading right now, plus how much data you have used in total.
 
 ![Widget](https://raw.githubusercontent.com/cracken7/TaskbarNetworkLounge/main/docs/widget.png)
 
-## Features
+## What you see
 
-**Live speed and traffic totals.** Read from the real interface counters
-(`GetIfTable2`, IP Helper) and divided by measured elapsed time
-(`QueryPerformanceCounter`) — no `ipconfig`, `netstat` or PowerShell parsing.
-Totals can be per session or persistent across restarts.
+The widget is split into two halves by a thin line:
 
-**Follows the connection you are actually using.** With a VPN running, the tunnel
-adapter and the physical adapter both carry the same bytes — the tunnel sees the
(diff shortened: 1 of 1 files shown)


cracken7:
Thanks — the review was accurate and useful. Pushed as 1.2.0. Details on each point:

**1. Overlap with `net-speed-taskbar`.** Fair hit on the README, and I've corrected the attribution: the `explorer.exe`-vs-`windhawk.exe` justification was taken from that mod's README, and it is now credited by name in the PR body rather than presented as my own finding. I did independently confirm the behaviour here with a control experiment (the official Caffeine mod under `@include windhawk.exe` produced no window on this machine), but the wording was still lifted and that was wrong of me.

On whether this should stand separately, rather than argue architecture I'll point at what it does that `net-speed-taskbar` doesn't:

* **Adapter selection is the point of the mod.** Auto mode resolves the default route against the *unfiltered* adapter list and counts exactly one adapter, so a VPN tunnel is followed while it holds the internet and never summed on top of its carrier. Summing tunnel + carrier reports every byte twice — that's the bug this was built around, and it also detects a change of the counted LUID set and resets the totals, because a byte count carried over from a different connection describes nothing.
* **Cumulative traffic totals**, session or persistent, which `net-speed-taskbar` doesn't have at all.
* **A details panel** (interface, status, IPv4, live rates, totals, reset).

I'm happy to take this to `net-speed-taskbar` as options instead if you'd still prefer that — say the word and I'll close this and open an issue with @NarayanChetri. I'd rather ask than fragment the catalog.

**Licensing.** No code was copied from `taskbar-music-lounge`. I read it as a reference for the window/z-band/acrylic approach — which is the same approach the wiki documents and several taskbar mods use — and the implementation here is my own; the `// Architecture mirrors Taskbar Music Lounge` comment overstated the relationship and is being read as provenance, which it wasn't meant to be. Both that mod and `net-speed-taskbar` are now credited as design references.

**2. `TextColor` was genuinely broken.** Confirmed exactly as described. Quoted the default; verified live by pixel measurement — `TextColor=0xFF6060` with Auto theme off now paints `(255,96,96)` instead of `0x777215`.

**3. Persistent totals moved into per-mod storage.** `Wh_SetBinaryValue`/`Wh_GetBinaryValue`, which deleted the custom text format, checksum, temp-file and `MoveFileEx` code. I also removed the stale `%LOCALAPPDATA%\TaskbarNetworkLounge` directory from my own machine to check nothing recreates it.

**4. "Windhawk Settings" opened File Explorer.** Correct — the helper is an `explorer.exe` image. Now resolves `%ProgramFiles%\Windhawk\windhawk.exe`, and the duplicated `GetModuleFileNameW` call is gone.

**5. `WhTool_ModUninit` could hang.** Real bug. The UI thread now publishes `GetCurrentThreadId()` before creating anything and shutdown posts `WM_QUIT` to the thread, so it works even if uninit runs before the window exists. And as you noted, that made the missing `DestroyWindow(g_hWidget)` load-bearing — it's now explicit, so the timer, the WinEvent hook and the class registration are released on a `WM_QUIT` exit too. Verified: disabling the mod exits the helper process cleanly and re-enabling starts a fresh one.

**6. `BoldText` dropped.** You're right that there was no old switch to preserve — that comment was a leftover from my own iteration, not from a released version. `TextWeight` is the single control.

**Optional items taken:** cached the DPI entry points in function-local statics; `TaskbarEventProc` checks the hooked window and `OBJID_WINDOW` before the cross-process `GetClassNameW`; collapsed the redundant `FlushTotals` branch; removed the dead `g_running`; replaced the synthesized `WM_SETTINGCHANGE` with `APP_WM_SETTINGS_CHANGED`; dropped `-lshcore` and `<shobjidl.h>`, added `<climits>`; fixed the `"wan minipor"` typo; documented the two visible consequences of the `explorer.exe` helper.

**Functionality notes taken:**

* `SnapshotDiffers` now compares the formatted strings. You were right that the OR of an absolute and a relative threshold fired nearly every tick.
* Panel dismissal no longer depends on `SetForegroundWindow`. A short timer closes the panel on a mouse-down outside the panel and the widget; the activation path is still tried first and still works when it succeeds. Verified live.
* `g_hoverState` split into `g_widgetHover` / `g_panelHover`.
* `WM_NULL` posted after `TrackPopupMenu`.

**Left alone, with reasons:**

* **Persistent + reset-on-source-change.** Kept as is, but I take the point about intent; the interaction is called out in the setting description. Defaulting the reset off in persistent mode would mean a "lifetime" total that silently mixes VPN and carrier bytes for the same transfer, which is the double-count this mod exists to avoid.
* **`WindhawkUtils::StringSetting`** — the hand-rolled wrapper also normalizes and case-folds for the enum comparisons; I'd rather not churn that in the same change as the correctness fixes. Noted for a follow-up.
* **`[[clang::no_destroy]] std::optional<std::thread>`** — same reasoning; the raw pointers are safe as you note, so this is style, and I'd rather keep this diff to behaviour.
* **Watchdog interval** — left at 2 s. It's the backstop for taskbar changes that emit no WinEvent, and measured cost is inside the noise of the 0.3–0.5 % of one core the mod already uses.
* **`std::atomic<HWND>` for `g_hWidget`** — agreed in principle; deferred with the other two style items.

All eight offline suites pass, and the storage stubs were added to every harness. `/ai-review` again when you're ready.

PR #5379, opened 2026-09-06
taskbar-centered-start-split-icons 1.0.0: fix resolve-timer re-arm ordering

rycalvo:
Update to `taskbar-centered-start-split-icons`, merged in #5111.

This picks up the one optional item from that PR's final review that was left unaddressed at the time. It was deferred deliberately — the review had just come back clean, and applying a last-minute change would have meant handing over a head no review had seen. The handover note offered it before merge; this is that fix.

Compile-checked clean on x86_64 and aarch64.

## Changelog

* Fixed the `kArmResolveNowMsg` handler in `BackgroundWorkerThreadProc` dropping a `SetTimer` failure. It killed the HWND-resolve timer and then discarded `SetTimer`'s return value, so a failure left no resolve timer running, with nothing logged. It now arms the replacement first and only kills the old one once the new id is known good — the same ordering `SetDragFollowPollInterval` already uses a few lines away, which is where this was noticed. Milder than that sibling case: the resolve timer is re-armed from several independent places on the taskbar thread (a real click, an `ArrangeOverride` count change, a subclass install), so losing it meant "recovers whenever something unrelated nudges it" rather than "never recovers". It now recovers on its own schedule, and a failure is logged rather than swallowed.
* Version bumped 0.1.0 → 1.0.0.

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini

_Not a new mod, so the section above is left unticked. Authorship is unchanged from #5111 and from the mod's own Disclosures section: written with Claude, under my direction._

### DIFF
--- a/mods/taskbar-centered-start-split-icons.wh.cpp
+++ b/mods/taskbar-centered-start-split-icons.wh.cpp
@@ -2,7 +2,7 @@
 // @id              taskbar-centered-start-split-icons
 // @name            Taskbar Start Button Centered Origin
 // @description     Pins the Start button to true screen-center and splits running-app taskbar buttons into two groups flanking it by which side of the screen each window is on (Windows 11 only; incompatible with "Start button always on the left")
-// @version         0.1.0
+// @version         1.0.0
 // @author          rick
 // @github          https://github.com/rycalvo
 // @include         explorer.exe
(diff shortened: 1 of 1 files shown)


rycalvo:
Both actionable optional items from the review are addressed in `9d349135`; CI is green on all five checks.

- **The failure log could claim a timer that doesn't exist** — correct, and thanks for catching it. `ButtonHwndResolveTimerProc` is a one-shot that kills itself and zeroes `g_buttonHwndResolveTimerId` before `ResolvePendingButtonHwnds`' `ScheduleNextResolveTick` requests the next arm, so on the tick → re-arm path the id really is already `0`. The message now reports which of the two states it found, using the shape you suggested. I left the "nothing left to keep" branch as a logged no-op rather than self-healing via a re-posted `kArmResolveNowMsg`: `SetTimer` realistically only fails on timer exhaustion, and a retry loop against that seemed more likely to add noise than to help.
- **`BackgroundWorkerThreadProc`'s initial `SetTimer` discarding its result** — taken. No reordering needed since it self-heals (the id stays `0` and the next `kArmResolveNowMsg` arms fresh), but it now logs, so a session whose first resolve pass never happened has an explanation.

One item deliberately not taken:

- **`0.1.0` → `1.0.0` vs `0.1.1`.** Fair reading of the diff, but the two of us are versioning different things: you're versioning this change, which is indeed small, while `1.0.0` is meant to mark the mod's first stable release now that it's actually in the catalog rather than a submission. Happy to move it to `0.1.1` if you'd rather the number track change scope — just say so and I'll push it.

rycalvo:
Handing over at `9d349135` with the review clean. Notes on the optional items left unaddressed:

- **Folding the handler and `SetDragFollowPollInterval` into a shared `RearmThreadTimer` helper.** Agreed that it's duplication, and the helper you sketched is the right shape. Not doing it here: it's a purely structural change to two call sites that both currently work, and it would cost another review cycle right at handover. It's the natural first step whenever either site next needs a substantive change.
- **Comment volume / the arm-before-kill rationale stated in three places.** Fair. The three aren't quite saying the same thing — one carries the rationale, one explains why the ordering is needed there too, and the third explains why *no* reordering was needed — but that distinction doesn't justify the length, and they'd collapse naturally into one place along with the helper above.
- **Defensive `if (g_buttonHwndResolveTimerId == idEvent)` in `ButtonHwndResolveTimerProc`.** I agree with your own read that it isn't reachable as written: `KillTimer` destroys the timer before a `WM_TIMER` can be synthesized for it, and posted messages are dequeued ahead of timer messages regardless. I'd rather not add a guard against an unreachable state immediately before handover; worth revisiting if the dispatch assumptions around that proc ever change.

On the functionality note: agreed, and stated plainly — this release is hardening plus a diagnostic, not a user-visible bug fix. Both new branches require `SetTimer` to fail, which realistically means user-object exhaustion. The `1.0.0` number is marking the mod's first stable release now that it's in the catalog rather than claiming a large change; if you'd prefer the version track change scope, `0.1.1` is a one-line push and I'm happy to make it.

PR #5387, opened 2026-09-07
Publish the first version of the Windows 7 Region and Language Restorer mod

babamohammed2022:
## Changelog

* Changelog item 1...
* Changelog item 2...

## Mod authorship

This mod was created by:

- - [X] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/win7-intl-control-panel-restorer.wh.cpp
+++ b/mods/win7-intl-control-panel-restorer.wh.cpp
@@ -1,0 +1,6535 @@
+// ==WindhawkMod==
+// @id              win7-intl-control-panel-restorer
+// @name            Windows 7 Region and Language Restorer
+// @description     This mod restores the classic Windows 7 Region and Language Control Panel pages on Windows 10 and 11
+// @version         1.0.0
+// @author          babamohammed
+// @include         explorer.exe
+// @include         control.exe
+// @include         rundll32.exe
+// @architecture    x86-64
+// @compilerOptions -lbcrypt -lwinhttp -luser32 -lshell32 -ladvapi32 -lole32 -lpsapi -lversion
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Windows 7 Region and Language Restorer
+
+This mod restores the classic Windows 7 "Region and Language" Control Panel page on Windows 10 and Windows 11.
+
+---
+
+## Functionality
+
+- The mod restores the classic Region page with its four tabs: Formats, Location, Keyboards and Languages, and Administrative.
+- All changes are applied through the standard Windows controls and nothing is simulated.
+- The interface is available in 20 languages and follows the system language automatically.
+
+---
+
+## Requirements
+
+- 64-bit Windows 10 or Windows 11 (ARM64 is not supported).
+- On first use, the original Windows 7 component is downloaded automatically from Microsoft and checked before use. An Internet connection is needed only for this step.
+
+---
+
+## Settings
+
+The mod includes a series of settings:
+- **UI language** (default: Automatic): this setting controls the language of the restored page. Automatic follows the Windows display language.
+- **Redirect Settings pages** (default: off): this setting opens the classic page instead of the modern Settings Region pages.
+
+---
+
+## Notes
+
+- This modification has been tested on Windows 10 21H2.
+- Some Windows 7 features no longer exist on modern Windows. In those cases the closest modern equivalent is opened instead (for example, the "Default location" link opens the Location privacy page).
+- Settings that were already applied are kept after the mod is disabled.
+- Windows systems file **are not modified**.
+
+---
+
+## Known limitations
+
+- This modification is a best-effort reimplementation of a NT 6.1 binary file. Some translations might not be completely accurate to the original files as the mod provides them by itself.
+- Display-language installation is not available. It depends on components that exist only on Windows 7.
+
+
+---
+
+## Opening the page
+
+While the mod is active, the restored page can also be opened directly:
+
+- `control.exe intl.cpl` - the classic Region page.
+- `control.exe intl.cpl,,3` - the page opened on Administrative.
+
+---
+
+## Credits
+
+- Based on the technique of the example restorer mods.
+
+---
+*/
+// ==/WindhawkModReadme==
+
+// ==WindhawkModSettings==
+/*
+- language: auto
+  $name: UI language
+  $description: This setting controls the language of the restored Region UI. Automatic follows your Windows display language. Changing this reloads the mod.
+  $options:
+  - auto: Automatic (follow Windows)
+  - en-US: English (genuine Microsoft)
+  - it-IT: Italiano
+  - de-DE: Deutsch
+  - fr-FR: Français
+  - es-ES: Español
+  - pt-BR: Português (Brasil)
+  - nl-NL: Nederlands
+  - pl-PL: Polski
+  - ru-RU: Русский
+  - zh-CN: 中文简体
+  - ja-JP: 日本語
+  - ko-KR: 한국어
+  - tr-TR: Türkçe
+  - cs-CZ: Čeština
+  - hu-HU: Magyar
+  - ro-RO: Română
+  - sv-SE: Svenska
+  - uk-UA: Українська
+  - el-GR: Ελληνικά
+  - ar-SA: العربية (RTL)
+- redirectSettings: false
+  $name: Redirect Settings pages
+  $description: This setting redirects the modern Settings Region and Language pages (ms-settings:regionlanguage, ms-settings:regionformatting) to the restored classic Region dialog. It applies to links opened from Explorer; Date & time, Speech and Typing pages are never touched. Changing this reloads the mod.
+*/
+// ==/WindhawkModSettings==
+
+// ============================================================================
+// ANALYSIS LEDGER (all values measured from the pinned binaries, not guessed)
+// ----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


babamohammed2022:
It seems like that changing the architecture to amd64 makes the validation script fail while having it to x86-64 makes it work but the compatibility checks stop working. If it's a problem on my side, I'll try to fix it but I have not understood what is it in this case.

m417z:
> It seems like that changing the architecture to amd64 makes the validation script fail while having it to x86-64 makes it work but the compatibility checks stop working

It's asking for a manual review, because often it is a mistake. In this case, it's probably OK. GitHub doesn't have a way to show warnings, so alerting it by returning an error is the only way.

I'll think about improving it in the future, but for now, you can ignore it as long as there are no other failures.

babamohammed2022:
> The last push fixed the `IOpenPanel` vtable (thanks — that one is fully resolved) and added the credits/notes text, but the other findings from the previous round are still present in the code. Item 3 below is the one worth looking at first: the new commit wrote the correct behavior into the _comment_ on `g_legacyThreads` and changed the container to `std::pair<DWORD, HWND>`, but `CloseOwnedWindows` still ignores the `HWND` and does exactly what the comment says it must not do. The rest are restated compactly.
> 
> **1. `ShutdownLegacyOnInitThread` leaks a reference on the mod's own module, so the mod is never unloaded.** Unchanged at lines 7938-7941:
> 
> ```c++
> HMODULE self = nullptr;
> GetModuleHandleExW(GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS,
>     reinterpret_cast<LPCWSTR>(&ShutdownHookProc), &self);
> ```
> 
> Without `GET_MODULE_HANDLE_EX_FLAG_UNCHANGED_REFCOUNT` this increments the refcount, and nothing releases it. Windhawk unloads a mod with a **single** `FreeLibrary` right after `Wh_ModUninit` returns, so after the first Region activation that `FreeLibrary` is a no-op: the mod image, its globals, its VEH registrations and the mapped Win7 provider stay in `explorer.exe` permanently, and the next enable/update loads a second copy. Your own [win7-display-control-panel-restorer.wh.cpp#L11266](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/win7-display-control-panel-restorer.wh.cpp#L11266) already passes the flag — do the same here (or `FreeLibrary(self)` after the `UnhookWindowsHookEx`).
> 
> Related, same function: `ShutdownHookProc` signals `g_shutdownDone` (line 7922) and only then executes `return CallNextHookEx(...)`. Once the refcount above is fixed, the waiting `Wh_ModUninit` thread resumes at that `SetEvent` and Windhawk can unmap the image while the hook proc's epilogue and return address are still live on the init thread. `UnhookWindowsHookEx` doesn't wait for an in-flight hook proc either. Have the hook proc set a flag only and signal the event from a point the unload thread can genuinely join.
> 
> **2. `Wh_ModBeforeUninit` still has no deadline.** Lines 8327-8346:
> 
> ```c++
> while (g_active.load(...) != 0 || g_jobs.load(...) != 0) { CloseOwnedWindows(); ... }
> ```
> 
> The added periodic `Wh_Log` makes a stuck unload _visible_, but it still hangs. `WM_CLOSE` is only a request — a modal Win7 dialog whose `DLGPROC` ignores it, or one sitting on a nested modal child, never closes, and the loop spins forever. Because `Wh_ModSettingsChanged` unconditionally sets `*reload = TRUE`, this path runs on **every settings change**, not just on disable. `WaitForJobs()` (line 7243) and `JoinTrackedThreads()`'s `WaitForMultipleObjects(..., INFINITE)` (line 410) have the same shape. Bound the outer wait (10-20 s); on expiry, leave the provider mapped and put the hooks in a permanently-disabled pass-through state instead of blocking. A hang is worse for the user than a leak.
> 
> **3. `CloseOwnedWindows` posts `WM_CLOSE` to every top-level window of any thread that happens to be inside a legacy call.** The new commit changed `g_legacyThreads` to `std::vector<std::pair<DWORD, HWND>>` and documented why a thread-ID filter is unsafe, but `CloseOwnedWindows` (lines 7865-7890) throws the `HWND` away and enumerates the whole thread anyway:
> 
> ```c++
> for (const auto& entry : g_legacyThreads) legacyThreadIds.push_back(entry.first);
> ...
> for (DWORD tid : legacyThreadIds)
>     EnumThreadWindows(tid, [](HWND window, LPARAM) -> BOOL {
>         PostMessageW(window, WM_CLOSE, 0, 0); return TRUE; }, 0);
> ```
> 
> So the stored host `HWND` is dead state and the behavior is identical to before. The root cause is that `RedirectShellExecuteExW` calls `EnterLegacy()` (line 8040) for **every** `ShellExecuteExW` in Explorer, not just for legacy-provider work — including the calls the taskbar, Start menu and desktop make to launch programs. Two consequences:
> 
> * Disabling or reloading the mod (i.e. any settings change) while such a call is in flight broadcasts `WM_CLOSE` to `Shell_TrayWnd`, `Progman`, the Start menu and every `CabinetWClass` window on that thread. Even in the plain CPL case it closes the user's Control Panel window, not just the mod's dialog.
> * `ShellExecuteExW` blocks while a UAC consent prompt is on screen, so `g_active` stays non-zero for as long as the user leaves that prompt up — which feeds the unbounded wait in item 2.
> 
> Two fixes, both small: don't take the legacy rundown gate in `RedirectShellExecuteExW` (it only rewrites a URL and never touches the mapped image), and restrict `CloseOwnedWindows` to windows the mod actually created (`g_owned`) plus windows whose owner chain traces back to a tracked/host window — never to whole threads.
> 
> **4. Double-close of WinHTTP handles when an unload cancels an in-flight download.** `CancelInFlightDownload()` (line 305) closes `g_activeSession` from the unload thread, but `Download()`'s `HttpHandle` destructors then run unconditionally on the already-closed session — and on the connect/request handles, which the session close already destroyed as children:
> 
> ```c++
> struct HttpHandle { HINTERNET value; ~HttpHandle() { if (value) WinHttpCloseHandle(value); } };
> ```
> 
> `explorer.exe` uses WinHTTP itself, so a recycled handle value means the mod closes an unrelated component's handle. Your display restorer already solves exactly this — see the `hUrl = nullptr; // already closed by CancelInFlightDownload` check-and-clear at [win7-display-control-panel-restorer.wh.cpp#L1634](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/win7-display-control-panel-restorer.wh.cpp#L1634). Decide ownership under `g_downloadLock` so exactly one side closes each handle.
> 
> **5. The first-use download blocks the activation thread with no overall deadline and no stall detection.** `CplHook` → `EnsurePrepared()` → `Prepare()` → two `EnsurePinned()` downloads, all under `g_prepareLock`, on the thread handling the Control Panel activation. `WinHttpSetTimeouts(10000, 10000, 15000, 15000)` is per-operation, and the `WinHttpReadData` loop (lines 540-546) has no progress check at all, so a server that trickles bytes keeps the UI frozen indefinitely with no feedback. Add a no-progress detector (your display restorer does this at [win7-display-control-panel-restorer.wh.cpp#L1605](https://github.com/ramensoftware/windhawk-mods/blob/main/mods/win7-display-control-panel-restorer.wh.cpp#L1605)) plus an overall deadline, and consider falling back to the native CPL immediately while the fetch runs in the background.
> 
> **6. Uncontained faults in the manually mapped Win7 image — still undocumented.** The `CrashGuard` comment (lines 918-935) is explicit that a hardware fault inside the mapped `intl.cpl` "still takes the host process down" and that the guard only observes it. The README note added in the last commit covers _feature parity_, not this. Please do one of:
> 
> * implement the containment the comment scopes out (`-fms-extensions` + `__try/__except` around the `GuardCpl`/`GuardEntry` calls, so a fault degrades to "legacy disabled, native fallback"), **or**
> * state in the README that a fault in the legacy provider can take down Explorer.
> 
> Also still open from last round: `input.dll` is loaded with the normal Windows loader and IAT-patched (`AdaptInputIat`), which gets you real unwind info, TLS and loader semantics for free, while `intl.cpl` gets the hand-written mapper (`CopyImage`/`Relocate`/`BindImports`/`ProtectImage` + `RtlAddFunctionTable` + manual `DllMain`) — several hundred lines of the riskiest code in the mod. If the reason is that the private `msvcrt` heap overrides must be in place before `_CRT_INIT` runs in `DllMain`, that's a legitimate constraint, but please write it down in the PR or in a comment rather than leaving it implicit.
> 
> Optional improvements

The mod seems to work fine for me at the moment. If the maintainer believes that these problems need to be addressed, I'll do it. In alternative, I can fix them in a future update.
/ready-for-reviewer

m417z:
Fix 1 and 4 from the last review.

Regarding "Vectored crash guard" - why do you need it? What crashes without it? Generally just swallowing all crashes in a system process is not a good idea, and I prefer not having mods that do that.

babamohammed2022:
> Fix 1 and 4 from the last review.
> 
> Regarding "Vectored crash guard" - why do you need it? What crashes without it? Generally just swallowing all crashes in a system process is not a good idea, and I prefer not having mods that do that.

I'll fix the points 1 and 4 and thanks for clarifying, I tried this approach to try to make the mod less likely to crash but I'll remove it to adhere to the policy.

babamohammed2022:
@m417z I've tried to address the first problem reported by the AI reviewer. However, the amount of AI reviews for this day has reached its limit, so I can't receive another. Is the mod ready?

PR #5389, opened 2026-09-07
Add Taskbar Volume Percentage Indicator mod

### LINKED ISSUE #4162: Volume percentage Indicator on the taskbar (like the battery percentage indicator)
a mod that adds volume percentage Indicator (like the battery percentage indicator) for the control center button on the taskbar (it is besides the volume icon)

gilnett:
Displays the volume percentage directly in the Windows 11 system tray volume icon.

Includes configurable display formats (percentage, number only, custom prefix, icon + percentage, or default).

Closes #4162

## Changelog

* Initial release.

## Mod authorship

This mod was created by:

- [ ] Claude
- [ ] ChatGPT
- [x] Gemini
- [ ] Other (please specify): 

### DIFF
--- a/mods/taskbar-volume-percentage.wh.cpp
+++ b/mods/taskbar-volume-percentage.wh.cpp
@@ -2,7 +2,7 @@
 // @id              taskbar-volume-percentage
 // @name            Taskbar Volume Percentage Indicator
 // @description     Displays the exact master volume percentage in the system tray natively inside the Windows 11 volume button with real-time sync.
-// @version         1.2.0
+// @version         1.3.0
 // @author          gilnett
 // @github          https://github.com/gilnett
 // @include         explorer.exe
@@ -20,6 +20,7 @@ Replaces the Windows 11 taskbar volume icon with the current volume level in rea
 
 - **Display styles**: Percentage (`50%`), number only (`50`), custom prefix (`Vol 50%`), emoji (`🔊 50%`), or default icon.
 - **Mute indicator**: Customizable mute display (`MUT`, `Mute`, `0%`, `✕`, `🔇`, or native glyph).
+- **Zero-Jitter Structural Stabilization**: Locks the XAML container minimum width (`IFrameworkElement::put_MinWidth`) to prevent neighboring system tray icons from shifting during volume changes.
 
 ## Credits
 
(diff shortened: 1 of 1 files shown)


gilnett:
Passing the string parameter to UpdateVolume does not work because Windows does not use that parameter to display the taskbar icon, resulting in no visual change. The icon glyph is managed internally at offset 0x90, which remains the only working way to update the visual text.

gilnett:
Making the destructor hook required causes `HookSymbols` to fail completely on Windows builds where the destructor symbol is inlined or missing from public PDBs, preventing the mod from loading. 

I kept it optional so the main hooks (`UpdateVolume`, `OnDataModelChanged`) can attach reliably while still cleaning up the pointer whenever the symbol is available.

gilnett:
Regarding the first 3 points, we are going in circles:
1. **`UpdateVolume` parameter**: Parameter 4 does not control the tray icon glyph (it's for spatial audio tooltips). Replacing it leaves the icon unchanged, so offset `0x90` is required.
2. **Destructor hook**: Making it required breaks symbol resolution and prevents the mod from loading on Windows 11 builds where the destructor is inlined. It must stay optional.
3. **Core Audio init**: Review 2 asked to remove it as dead code, and now Review 4/5 asks to add it back. I am keeping it on the taskbar thread to sync the startup volume properly.

m417z:
Thanks, great mod. Functionality wise, `fixedContainerWidth` doesn't work for me, and also I think it'd be nice to be able to keep the original icon with a number beside it as an option, but it might be more challenging due to the fonts and might require more work. In any case, it's already useful as is, and it can be improved in the future.

Regarding the implementation, I agree with the last AI review on several points: the offsets, the optional hooks, and the `CoCreateInstance` call, and others.

I felt it's easier for me to give it a go at improving it, and here's what I got (didn't do much testing):
https://gist.github.com/m417z/0f83b1f95fc4584a15aa6c27c841f250

It also ended up being shorter. Let me know what you think.

gilnett:
Thanks @m417z for the review and the refactored version.
I restarted directly from your modified code.
Regarding fixedContainerWidth and the jitter, I have already sorted that out. The container width is now calculated dynamically from the maximum text length and the size observed in MeasureOverride, which keeps the layout stable without unnecessary empty space.
As for keeping the original volume icon next to the text, I will look into that in more detail later.

I'll make a commit when the original icon part is done

gilnett:
Regarding the optional suggestions that I decline:
- **Audio pre-seeding:** As I already answered in previous comments, I keep m417z's non-blocking design to avoid taskbar freezes.
- **Tree mutation in LayoutUpdated:** As I already answered, the current layout is already stable and glitch-free.
- **Multi-screen revokers:** The system tray only exists on the primary taskbar in Windows 11.
- **Array lookup:** I keep the explicit `switch` to avoid coupling with the enum order.
- **Minor checks:** `kernelbase.dll` is always loaded in Explorer, and the ternary check is optimized away.

gilnett:
point 1 / Removing the trigger text breaks the mod: without it, Windows XAML doesn't trigger a layout recalculation on identical native glyphs, so the percentage never shows up.

gilnett:
Point 1: Calling RefreshVolumeIcons() at startup is a no-op because g_volumeDataModels hasn't been populated by Windows yet, so fetching the audio level via COM won't display anything.
Point 2: Asking to remove the retry directly contradicts your previous review ("log the failure and retry"), but your new reasoning about secondary Explorer processes makes sense, so I simplified it.

m417z:
Looks great. A small issue: The native icon is not centered if used with "Container width":

https://github.com/user-attachments/assets/c9196ca4-126d-41da-8d5b-a99745245b14

Also, is there a reason that the order is different for various options, e.g. "Icon and percentage", but "Percentage and native icon"?

gilnett:
Also, is there a reason that the order is different for various options, e.g. "Icon and percentage", but "Percentage and native icon"? 
my bad  it's just me who didn't look (I'l make a parameter for peopple to change it)
for the rest i'll look at it now

PR #5415, opened 2026-09-09
[mutealert] Fix startup mute sync and group settings

Nikolay1243:
## Summary
- recover an unmuted Windows input at startup only when MuteAlert recorded that it applied the preceding headset-driven mute
- keep manual Windows privacy mutes intact and keep call-app unmute synchronization transition-driven
- preserve fail-safe physical-mute enforcement for Windows input and active calls
- organize settings into collapsible General, Headset, Slack, Teams, Zoom, and Google Meet groups
- bump the mod version to 0.9.10

## Cause
The first SteelSeries poll updated the displayed physical state but required an already-known muted predecessor before it could synchronize an unmuted Windows input. If MuteAlert had muted Windows from the physical headset state and Explorer/Windhawk restarted after the headset was unmuted, users had to perform another mute/unmute cycle before the states converged.

MuteAlert now persists ownership only after its headset-driven Windows `SetMute(true)` succeeds. Startup unmute is permitted only when that marker is present. A taskbar mute toggle, successful headset-driven unmute, or disabling Windows headset synchronization clears it. Manual Windows mutes are never claimed and therefore aren't undone at load.

Active calls remain transition-driven in the unmute direction to avoid creating a hot mic after an Explorer restart, Windhawk update, or mod enable/disable cycle. Observable physical mute remains fail-safe and is re-applied while present.

## Settings compatibility
Grouping changes the YAML paths (for example, `position` becomes `General.position`). This deliberately accepts a one-time settings reset in exchange for a usable grouped settings page. A prominent mod README notice tells existing users to review and save their configuration after updating.

## Validation
- diff whitespace check passed
- compatibility compilation is run by the PR for Windhawk 1.6.1, 1.7.3, and 2.0.0-alpha.3
- the matching standalone implementation builds cleanly with MSVC `/W4` and its release dependency check passed
- Microsoft native code analysis reports only pre-existing Windows SDK/GDI+ annotations and the existing `CoCreateInstance` warning, with no new warning in this change

### DIFF
--- a/mods/mutealert.wh.cpp
+++ b/mods/mutealert.wh.cpp
@@ -2,7 +2,7 @@
 // @id              mutealert
 // @name            MuteAlert - Microphone Activity Taskbar Widget
 // @description     Shows live microphone activity, call mute state, volume controls, and headset mute synchronization in the Windows 11 taskbar.
-// @version         0.9.9
+// @version         0.9.10
 // @author          Nikolay
 // @github          https://github.com/Nikolay1243
 // @homepage        https://github.com/MuteAlert/windhawk
@@ -19,6 +19,12 @@
 # MuteAlert for Windhawk
 
 Adds a microphone button to the Windows 11 system tray area.
+
+> **Updating from 0.9.8 or earlier:** Version 0.9.9 organized the settings
+> into collapsible sections. Windhawk can't migrate the old flat setting paths,
+> so review and save your settings once after updating. In particular, re-enable
+> headset synchronization and call integrations, restore localized button text,
+> and check the position and locked-volume target.
 
 ![MuteAlert microphone activity widget](https://raw.githubusercontent.com/MuteAlert/windhawk/main/assets/taskbar-widget.png)
 
@@ -754,6 +760,11 @@ static UINT g_removeWidgetsMessage = 0;
 static std::atomic<bool> g_updateQueued{false};
 static std::atomic<unsigned long long> g_audioNameGeneration{0};
 static std::atomic<unsigned long long> g_headsetStatusGeneration{0};
+// These flags belong to the lifetime of the loaded mod, rather than a worker
+// thread. Saving unrelated settings restarts workers and must not re-arm a
+// startup unmute.
+static std::atomic<bool> g_vendorInitialSyncPending{true};
+static std::atomic<bool> g_hardwareInitialSyncPending{true};
 
 static bool IsStopping() {
     return g_unloading.load() ||
@@ -2176,13 +2187,13 @@ static DWORD WINAPI AudioThreadProc(void*) {
         if (endpoint.hardwareMute) {
             bool changed = observedHardwareMuteKnown &&
                            observedHardwareMuted != (muted != FALSE);
-            bool initialMuted = !observedHardwareMuteKnown && muted != FALSE;
+            bool initialSync = g_hardwareInitialSyncPending.exchange(false);
             bool syncMute = g_settings.headsetSyncMode == L"full" ||
                             g_settings.headsetSyncMode == L"muteOnly";
             bool syncUnmute = g_settings.headsetSyncMode == L"full";
             if (g_settings.headsetSyncCalls &&
-                ((muted && (changed || initialMuted) && syncMute) ||
-                 (!muted && changed && syncUnmute))) {
+                ((muted && (changed || initialSync) && syncMute) ||
+                 (!muted && (changed || initialSync) && syncUnmute))) {
                 QueueActiveCallMuteState(muted != FALSE);
                 RecordDiagnosticEvent(
                     std::wstring(L"Windows hardware mute changed to ") +
@@ -3157,6 +3168,8 @@ static DWORD WINAPI HeadsetThreadProc(void*) {
                 UpdateSteelSeriesSource(false, false, L"", L"");
                 nextVendorPoll = now + g_settings.headsetPollInterval;
             } else {
+                bool initialSync =
+                    g_vendorInitialSyncPending.exchange(false);
                 bool stateChanged =
                     !stateKnown || observation.muted != previousMuted;
                 UpdateSteelSeriesSource(true, observation.muted,
@@ -3181,8 +3194,11 @@ static DWORD WINAPI HeadsetThreadProc(void*) {
                         QueueActiveCallMuteState(true);
                     }
                 } else if (!observation.muted && syncUnmute &&
-                           (!stateKnown || previousMuted)) {
-                    if (g_settings.headsetSyncWindows) QueueMuteSet(false);
+                           (initialSync || (stateKnown && previousMuted))) {
+                    if (g_settings.headsetSyncWindows &&
+                        (initialSync || g_audioMuted.load())) {
+                        QueueMuteSet(false);
+                    }
                     if (g_settings.headsetSyncCalls) {
                         QueueActiveCallMuteState(false);
                     }
--- a/mods/mutealert.wh.cpp
+++ b/mods/mutealert.wh.cpp
@@ -141,7 +141,7 @@ volume control, call state, and headset integration.
 - Headset:
   - headsetSyncMode: off
     $name: Headset mute synchronization
-    $description: Uses Windows hardware mute, standard HID mute controls, or a supported vendor adapter. In full mode, the Windows input and detected active calls are reconciled to the first observable physical state whenever the mod loads. The taskbar tooltip shows the current detection method and confidence. Silence is never interpreted as physical mute.
+    $description: Uses Windows hardware mute, standard HID mute controls, or a supported vendor adapter. In full mode, the Windows input is reconciled to the first observable vendor state whenever the mod loads; call apps change only after a physical transition. While a physical mute remains observable, mute-only and full modes re-apply it if software is manually unmuted. The taskbar tooltip shows the current detection method and confidence. Silence is never interpreted as physical mute.
     $options:
     - full: Sync physical mute and unmute changes
     - muteOnly: Sync only physical mute changes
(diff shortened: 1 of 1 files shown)


Nikolay1243:
Addressed in 0fd06383:

- Replaced the worker-local interpretation of stateKnown with process-lifetime one-shot flags for vendor and hardware startup reconciliation. Saving settings, reconnecting the headset, transient read failures, and temporary offline states no longer re-arm startup unmute.
- Kept real vendor transitions gated by stateKnown and previousMuted; an initial unmuted sample is queued even if the audio worker has not published its state yet.
- Applied symmetric first-state reconciliation to Windows hardware mute for active-call synchronization.
- Added a prominent Updating from 0.9.8 or earlier notice to the mod README explaining the one-time settings review required by the new grouped paths. I kept the grouping because it was explicitly requested to make the very long settings UI usable.
- Kept the existing group/child keys unchanged from 0.9.9 to avoid a second settings-path reset. The naming suggestions are therefore intentionally deferred.

The standalone equivalent was updated and rebuilt as v0.9.7 with MSVC /W4, native analysis, optimized release compilation, and dependency verification.

Nikolay1243:
Addressed the synchronization findings in 338d01e2:

- Split vendor startup reconciliation into independent Windows-input and active-call one-shots, so an early input sync cannot consume call sync.
- Added a first-call-scan completion gate. Vendor and Windows-hardware call reconciliation now remains pending until enabled call integrations have published their initial state.
- One-shots are consumed only when the corresponding destination is enabled and the observed mute state is eligible for the selected full/mute-only mode. Enabling synchronization later in the same mod load therefore still reconciles correctly.
- Kept real reconnects/transient unavailable states from re-arming either one-shot.
- Removed the cached-audio guard from a genuine vendor unmute transition, avoiding the narrow stale-cache drop noted in the review.
- Corrected the migration notice to 0.9.10, clarified initial reconciliation in the full-mode description, and made hardware diagnostics distinguish an initial state from a transition.

I deliberately retained the grouped settings plus the user-facing migration notice, as requested. I verified against Windhawk's current LoadedMod::GetIntSetting/GetStringSetting implementation that undeclared legacy names are readable, but the proposed default-based fallback cannot distinguish an absent grouped value from a user explicitly choosing that default. A permanent shim can therefore resurrect a stale legacy value after an intentional reset, while the mod API has no supported way to write the grouped settings shown by the UI. I don't think that ambiguity is safe for microphone/call controls, so this PR continues to accept the one-time reset explicitly rather than silently applying potentially stale values.

/ai-review

Nikolay1243:
Addressed the hot-mic and one-shot findings in c9fdfb32:

- Removed load-time reconciliation for Slack, Teams, Zoom, and Google Meet. Active-call synchronization is transition-driven again, so an Explorer restart, mod update, or enable/disable cycle cannot unmute a deliberately muted live call.
- Removed the call-scan completion coordination and the Windows-hardware initial call-sync path that existed only to support load-time call reconciliation.
- The vendor startup one-shot now applies only to Windows input unmute and is consumed only inside the eligible unmuted/full-sync branch. An initial muted state no longer burns it.
- Documented the deliberate asymmetry: an observable physical mute continues to win and is re-applied, while unmute is transition-driven after initial Windows-input reconciliation.
- Updated the PR description to 0.9.10 and made the one-time grouped-settings reset an explicit maintainer decision, with the ambiguity of a legacy fallback explained.

/ai-review

Nikolay1243:
Addressed only the substantive safety/correctness findings in 259e5b8b:

- Startup unmute now requires a persisted ownership marker proving MuteAlert previously applied the Windows mute from a headset action. A manual Windows privacy mute is not claimed and won't be undone on load.
- Ownership is persisted only after SetMute(true) succeeds, cleared after a successful headset-driven unmute, cleared when the taskbar mute control is used, and cleared when Windows headset synchronization is disabled.
- Added the current-state guard before queuing unmute, preventing a redundant set from swallowing a nearly simultaneous taskbar toggle.
- Restored fail-safe initial/steady Windows-hardware mute propagation to active calls; unmute remains transition-only.
- Updated the setting description and PR body to state the exact behavior.

Per the project owner's direction, the one-time configuration reset from grouping is expected and accepted. I have intentionally not added a compatibility shim or made the optional naming/YAML/cosmetic changes.

/ai-review

Nikolay1243:
Addressed the two substantive findings in f95b46ac:

- Bound headset-owned startup recovery to the exact Core Audio endpoint ID that MuteAlert successfully muted. If the default microphone changed, ownership is cleared and the new device is never unmuted.
- Added a five-second reassert limit for Windows-hardware and SteelSeries call muting. A missing/unresponsive/localized UIA control can no longer trigger full scans and button invokes at the 25-50 ms audio polling rate.
- Reassert attempts remain immediate on an initial mute or real transition. When no mismatched active call exists, the inexpensive atomic check remains available so a newly detected call can still be muted promptly.
- Clarified that startup recovery is a supported vendor-adapter feature and applies only to the same input device.

The settings reset and optional naming/cosmetic suggestions remain intentionally unchanged per the project owner's direction.

/ai-review

Nikolay1243:
Addressed the substantive ownership gaps in 608512be:

- The audio thread now clears headset ownership as soon as it observes that the endpoint was unmuted by another control, preventing a later manual privacy mute from being mistaken for MuteAlert's mute.
- A vendor unmute observed before Core Audio is available is queued for the audio thread instead of consuming state and deleting ownership. The audio thread waits until an endpoint exists, then enforces the persisted endpoint-ID match.
- Removed the fragile process-lifetime one-shot. Ownership plus endpoint identity now safely supports startup, resume, settings restart, and headset reconnect without permitting a manual mute to be released.
- Folded guarded unmute into the pending command value, so command type and endpoint requirement cross threads atomically.
- A failed endpoint match no longer consumes a pending taskbar toggle.
- Removed the duplicated disabled-sync cleanup block and avoided unnecessary value-store writes when no ownership exists.

The accepted settings reset and cosmetic suggestions remain unchanged.

/ai-review

Nikolay1243:
Addressed the two remaining substantive items in aecff2ee:

- Deferred headset mute/unmute commands now carry a five-second deadline and are discarded after expiry, so a stale transition cannot affect a microphone connected much later.
- The command, ownership requirement, and deadline are packed into one atomic value; there is no split-field race.
- Pending mute commands and taskbar toggles are reset whenever the audio workers start, preventing settings-reload carryover.
- Corrected the settings description: vendor mute is enforced against Windows on each status poll, while only active-call retries use the five-second throttle.

The accepted settings reset and optional polish remain intentionally unchanged.

/ai-review

Nikolay1243:
Fixed the remaining endpoint-ownership issue in 9950b4e3. The audio worker now caches whether the currently observed endpoint matches the persisted headset-owned endpoint, updates that cache when synchronization changes mute, and only clears ownership when that same endpoint is observed unmuted. A fallback microphone can no longer discard the marker while the disconnected headset remains muted. The C++ source remains CRLF-formatted.\n\n/ai-review

Nikolay1243:
Addressed required findings 1 and 2 in 94c419cc. Every headset-driven Windows unmute now uses the endpoint ownership guard, including normal SteelSeries transitions and standard-HID actions. The mute helper can now only enqueue mute, which prevents an accidental unconditional unmute. Persisted ownership also includes a boot stamp and is discarded outside the boot session where MuteAlert applied the mute. Finding 3 is intentionally unchanged: the settings reset caused by grouping is an accepted update behavior, is prominently documented in the README, and users are directed to review and save settings after upgrading.

Nikolay1243:
Implemented the requested one-time legacy settings fallback in 200e428e. A non-empty legacy position key identifies a saved pre-0.9.10 configuration; while present and schema version is still legacy, all 51 old flat settings are loaded instead of grouped defaults. Saving Settings marks schema 2 and switches to the grouped values. The old and migrated key sets were mechanically compared: 51 legacy keys, 51 mappings, no missing or unexpected keys. The README now documents this behavior.\n\nFinding 1 is intentionally unchanged because it directly contradicts the preceding review, which required all Windows unmute paths to be ownership-gated to preserve manual privacy mutes. MuteAlert therefore continues to treat a manual Windows mute as authoritative; physical mute synchronization only releases Windows mute when MuteAlert owns it.

Nikolay1243:
Accepted the review's recommended consistent option in 993b613b: removed the legacy runtime fallback and schema marker. The grouped Settings UI and the configuration used by the mod now always agree. The README clearly states that upgrading from 0.9.8 or earlier resets settings and identifies the values users should review and save once. The startup-sync and endpoint/boot ownership fixes are unchanged. The physical-unmute ownership policy is also unchanged because preserving a manual Windows privacy mute was an explicit finding in the preceding review.

Nikolay1243:
Fixed finding 2 in ca2fc05b. Windows Core Audio's hardware-support capability is no longer treated as proof of a continuously latched physical mute: call synchronization from that source is once again limited to initial observation or an actual mute-state change. The five-second reassert remains only in the SteelSeries vendor provider, where the latched physical state is genuinely observable. Finding 1 is intentionally not changed. The latest review overlooks that 200e428e did write settingsSchema from Wh_ModSettingsChanged; the immediately preceding review correctly rejected that exact fallback because the grouped UI could disagree with the effective legacy configuration and grouped settings saved while the mod was unloaded could be ignored. Commit 993b613b follows that review's recommended consistent option: grouped UI values are always the values in effect, with the one-time reset prominently documented.

Nikolay1243:
Fixed finding 2 in bf8e9ceb. Headset mute requests are now fail-safe and remain queued while the capture endpoint is unavailable; only the stale-risk owned-unmute command expires after five seconds, and expiry is recorded in diagnostics. Finding 1 is intentionally unchanged. The proposed fallback is the same inconsistent approach rejected by the immediately preceding review: it can make the grouped Settings UI disagree with the active legacy values and cannot reliably distinguish grouped defaults from absent grouped values for MuteAlert's booleans, numbers, and non-empty strings. The cited string-only empty-value fallback does not solve those cases. We are retaining the prior review's recommended consistent option: grouped UI values always equal effective values, with the one-time reset documented prominently.

Nikolay1243:
The latest AI review covers the current head bf8e9ceb. The remaining two findings are deliberate tradeoffs, so I am requesting maintainer review rather than another AI cycle.

1. Windows privacy mute: headset synchronization releases Windows mute only when MuteAlert owns it on the same endpoint and boot. A manual Windows mute remains authoritative, including after a physical mute/unmute cycle. The user can explicitly unmute Windows through the widget or Windows controls. This policy was required by the earlier review: https://github.com/ramensoftware/windhawk-mods/pull/5415#issuecomment-5637000736 . Later reviews asked for the opposite behavior. We retain the privacy-preserving policy rather than alternate between them.

2. Grouped settings: we retain the usability improvement and explicitly document the one-time reset. We implemented a legacy fallback in 200e428e, including the settingsSchema write in Wh_ModSettingsChanged. The next review rejected that exact fallback because the grouped Settings UI could disagree with effective runtime values and settings saved while the mod was unloaded could be ignored: https://github.com/ramensoftware/windhawk-mods/pull/5415#issuecomment-5653461425 . We removed it following that review's recommended consistent option. Reintroducing default-based fallback does not resolve those ambiguities. The README prominently explains which settings must be reviewed after upgrading.

All compatibility checks passed. Please assess these two product decisions directly; we are happy to follow a maintainer's concrete preference, but will not keep reversing them based on contradictory AI rounds.

PR #5420, opened 2026-09-09
Translucent Windows v1.8.1

Undisputed00x:
* Fixed an issue where explorer.exe would crash when middle-clicking items in the navigation pane.
* Fixed the opaque text rectangle on ListView group items in 32-bit applications.
* Dynamic desktop text color depending system theme mode.

### DIFF
--- a/mods/translucent-windows.wh.cpp
+++ b/mods/translucent-windows.wh.cpp
@@ -245,6 +245,8 @@ std::wstring g_LastThemePath = GetCurrentWindowsThemePath();
 // Flag to prevent customizing text colors of Task Manager
 BOOL g_InsideTaskMgrProc = FALSE;
 
+BOOL g_InsideExplorerProc = FALSE;
+
 using PUNICODE_STRING = PVOID;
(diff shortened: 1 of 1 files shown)


m417z:
What about the 8 findings in the last review? Do you think they're incorrect? Some of them seem to be worth fixing if what it says is correct, like out-of-bounds reads/writes in 1. Some of the code was already there in 1.8.0, but still.

Undisputed00x:
> What about the 8 findings in the last review? Do you think they're incorrect? Some of them seem to be worth fixing if what it says is correct, like out-of-bounds reads/writes in 1. Some of the code was already there in 1.8.0, but still.

The suggestions from 1. to 5. are incorrect/irrelevant. As for 6. and 7. i believe the mod doesn't create any issues.

bropines:
> * Fixed an issue where explorer.exe would crash when middle-clicking items in the navigation pane.
> * Fixed the opaque text rectangle on ListView group items in 32-bit applications.
> * Dynamic desktop text color depending system theme mode.

https://github.com/amnezia-vpn/amneziawg-windows-client

image

Can this be fixed?

PR #5436, opened 2026-09-10
Add Windows 11 Native Four Column Snap Layout mod

Hoffelhas:
# Windows 11 Native Four Column Snap Layout

Add an extra native Windows 11 Snap Layout with four equal vertical columns. This is especially useful on ultrawide monitors, where four equal vertical zones make better use of the available horizontal screen space.

Windows 11 four-column Snap Layout

## What it does

- Adds the four-column layout as an additional option without replacing the
  existing Windows layouts.
- Appears in **Win+Z**, **maximize-button hover**, and the **Snap Bar** shown
  when dragging a window to the top of the screen.
- Integrates with Windows' native Snap Layout model, so Snap Assist and Snap
  Groups continue to work normally.

Developed and tested on Windows 11 build **26200.9445** on x86-64.

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [x] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/windows-11-native-four-column-snap-layout.wh.cpp
+++ b/mods/windows-11-native-four-column-snap-layout.wh.cpp
@@ -71,14 +71,12 @@ thread_local bool g_inSnapBarLoad = false;
 thread_local bool g_snapBarCustomAdded = false;
 
 // Set after the normal flyout path successfully appends the custom layout.
-// Thread-local for consistency with the Snap Bar state.
 thread_local bool g_flyoutCustomAdded = false;
 
 // Keep track of SnapLayout.dll so a reference acquired by this mod can be
-// released again when the mod unloads.
+// released when the mod unloads.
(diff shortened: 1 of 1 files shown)


Hoffelhas:
Hi @m417z, I want to go for ready-for-reviewer, but I need to do ai-review first. Would it please be possible to do that now instead of waiting?

Hoffelhas:
/ready-for-reviewer

Regarding the suggestion to generalize the mod to N columns: I’d prefer to keep this submission intentionally focused on the missing four-column layout.

Windows already provides native two- and three-column layouts, while supporting more than four columns or adding other formats would require additional reverse-engineering. Generalizing the implementation now would expand the scope and complexity of the mod without a demonstrated user need.

If there is meaningful demand for arbitrary/custom column counts in the future, I think that would be better handled as a separate generalized mod.

PR #5437, opened 2026-09-10
Add KWin Slide Back Only v1.1.4

tabris-trees:
# Add KWin Slide Back Only v1.1.4

Introduces an experimental KWin-style Slide Back effect using live DWM thumbnail overlays hosted in explorer.exe. Real application window positions are not moved. Settings control clearance, motion strength, smoothness, and duration.

Rapid foreground changes cancel the current overlay animation and retain the latest pending target. Alt+Tab start/end events preserve a previous application stack and associate it with the selected target window.

## Validation and limitations

- Passed x86_64 syntax checking with the bundled Windhawk Clang compiler.
- No cross-machine runtime verification has been completed for this submission.
- Overlay creation follows Windows activation and does not guarantee a flash-free first frame.
- Rapid switching, Alt+Tab cancellation, and unload behavior need runtime review.
- The source retains its original MIT metadata; the author must verify provenance and any obligations for code derived from KWin before public submission.

## Mod authorship

- [x] ChatGPT

Submission file: mods/kwin-slide-back-only.wh.cpp

### DIFF
--- a/mods/kwin-slide-back-only.wh.cpp
+++ b/mods/kwin-slide-back-only.wh.cpp
@@ -1,0 +1,1822 @@
+// ==WindhawkMod==
+// @id              kwin-slide-back-only
+// @name            KWin Slide Back Only
+// @description     KWin-style Slide Back with DWM overlays, tray filtering, adjustable timing, and lossless latest-transition queuing.
+// @version         1.1.4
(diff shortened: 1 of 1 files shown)


Potassiumuncher:
peak

PR #5454, opened 2026-09-11
Translucent Windows v1.8.2

Undisputed00x:
* Fixed text alignment bug reported in some programs (e.g. YASB, Telegram Desktop etc.) introduced in the previous update.
* Fix the mod's accent color when changing the system accent color.


Undisputed00x:
@m417z actually no need for more AI assistance, its ready to be published.

PR #5460, opened 2026-09-11
Update windows-animations mod to version 1.3.5

redrag2105:
## Summary

This updates Windows Animations from 1.2.0 to 1.3.5, the first public update since 1.2.0. It adds hybrid GPU acceleration, faster CPU effects, a much quicker optional Win+D mode, true shuffle cycles, support for taskbars on every screen edge, rounded minimize/restore animations, and many compatibility and flicker fixes.

An unreleased prototype moved every animation to the GPU, but matched testing showed that this was not always faster. GPU rendering works especially well for restores and very dense close effects, while minimize and ordinary close effects perform better and more reliably on the optimized CPU renderer. Version 1.3.5 therefore chooses the best tested path for each animation instead of forcing GPU rendering everywhere.

## Performance comparison

Testing used a controlled Win32 window at 1920×1200, 125% scaling, and 165 Hz. Lower frame cost is better; values in parentheses are speedups compared with 1.2.0.

| Workload | 1.2.0 CPU | 1.3.5 optimized CPU | Full-GPU prototype | Final 1.3.5 route |
|---|---:|---:|---:|---|
| Taskbar minimize | 1.29 ms | 1.30 ms (0.99x) | 3.71 ms (0.35x) | CPU |
| Taskbar restore | 1.35 ms | 1.39 ms (0.97x) | 0.45 ms (3.02x) | GPU |
| Normal close, 5 px | 1.97 ms | 1.95 ms (1.01x) | 4.78 ms (0.41x) | CPU |
| Full-screen dense close, 1 px | 16.28 ms | 9.97 ms (1.63x) | 9.19 ms (1.77x) | GPU for qualifying Thanos/Perlin effects; CPU for Shatter |

The GPU options remain configurable. Unsupported hardware, Remote Desktop, incompatible windows/processes, a cold or failed GPU pipeline, and latency-sensitive Show Desktop operations automatically fall back to CPU.

[Full benchmark methodology and per-effect results](https://github.com/redrag2105/windhawk-windows-animations-preview/blob/windows-animations-v1.3.0/windows-animations-1.3.0-benchmarks.md)

## Changelog

* **Added smart hybrid GPU acceleration:** Restores and compatible launches can use DirectComposition GPU rendering, while qualifying large 1 px Thanos/Perlin closes can use GPU rendering. Minimize, ordinary close, Square Shatter, and unsupported cases use the faster or safer optimized CPU path.
* **Improved performance:** Optimized Perlin, Shatter, and Thanos CPU effects, reduced temporary memory usage, right-sized rendering surfaces, added refresh-aware frame pacing, and safely pre-warmed only the GPU effects likely to be used next.
* **Improved minimize, restore, and launch animations:** Fixed content flicker, blank or invisible windows, cursor stalls, incorrect taskbar targets, and rapid-click/reversal problems. Non-maximized animations now preserve the Windows 11 rounded-window silhouette.
* **Added taskbar support on every edge:** Genie and Windows 10 animations now target bottom, top, left, and right taskbars on primary or secondary monitors, including auto-hidden taskbars. Windows Terminal targeting was also improved.
* **Added the optional Optimize Show Desktop mode:** Win+D custom-animates only the foreground window while background windows minimize immediately, avoiding long sequential waits. Matching background restores stay native, while the foreground restore keeps its custom animation without duplicate windows or pre-animation flashes.
* **Reworked random styles into true shuffle:** Close and minimize/restore animations use separate session-wide shuffle cycles shared by all apps. Every style plays once before reshuffling, consecutive cycles do not repeat the same boundary style, and launch/minimize plus rapid reversal pairings stay visually consistent.
* **Improved Alt+Tab:** The switch effect now runs only for genuine Alt+Tab switches and no longer leaves File Explorer headers or backdrop areas black, opaque, or incorrectly styled.
* **Improved close and tray-hide behavior:** Long animations can finish fully, close-time flicker was reduced, console windows animate correctly, and GPU resources no longer keep closed Chrome or Task Manager processes alive.
* **Improved compatibility and stability:** Added safer automatic CPU fallback, better handling for File Explorer/backdrop windows and Chromium/Electron apps, support for windows already open when the mod is re-enabled, and protection against whole-screen animations after sleep or resume. Background-only helper processes are skipped.
* **Added optional performance telemetry:** When enabled, one compact summary per animation reports renderer choice, adapter, timing, frame coverage, memory, pacing, and CPU rendering breakdown. Normal logging remains quiet when telemetry is disabled.
* **Updated the README and settings:** Added a clear explanation of the hybrid renderer, performance guidance, GPU routing and fallback behavior, Show Desktop behavior, taskbar placement, rounded corners, and shuffle semantics.

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini


redrag2105:
1. The named exclusions are narrow pre-injection skips for helpers observed not to host animatable windows; the issue was unnecessary initialization, not a crash. I’m avoiding a blanket console-subsystem rule because a console-subsystem process can still create GUI windows. `UniKey` is a separate compatibility exception: GPU initialization can stall its UI process, and DirectComposition launch presentation can disrupt its dialog opening, so it retains CPU animation.

2. This is an intentional two-level logging choice. Windhawk controls whether logging is available; the mod’s telemetry setting opts into the more frequent animation-routing and taskbar-lookup messages alongside timing measurements. That keeps ordinary logs concise when Windhawk logging is enabled.

/ready-for-reviewer

m417z:
For future updates: perhaps it'd be more convenient to use a shared section instead of manually memory-mapped memory, where you need to take care of things such as the security descriptor. Example
https://github.com/ramensoftware/windhawk-mods/blob/81ca7be4c1380e3b330e565784e037230546f873/mods/taskbar-primary-on-secondary-monitor.wh.cpp#L165-L169

redrag2105:
> For future updates: perhaps it'd be more convenient to use a shared section instead of manually memory-mapped memory, where you need to take care of things such as the security descriptor. Example
> 
> https://github.com/ramensoftware/windhawk-mods/blob/81ca7be4c1380e3b330e565784e037230546f873/mods/taskbar-primary-on-secondary-monitor.wh.cpp#L165-L169

Thanks for the suggestion. This definitely looks a lot cleaner for simple cross-process state.

One thing I'm wondering about though: some of the shared state in this mod does a bit more than just pass a value between processes. The shuffle state, for example, is intentionally shared between the x86 and x64 builds so every app stays on the same session-wide sequence. The main shell state also has Explorer writing to it while target processes only map it read-only.

Since Windhawk builds separate x86/x64 DLLs, a shared section wouldn't be shared between those two images, right? So we'd still need something else for the state that has to cross that boundary.

I was thinking that for a future refactor, maybe I could split the state up and use a shared section for the simpler same-image stuff, while keeping a much smaller named mapping for the bits that need x86/x64 sharing or different read/write permissions. That should still get rid of a good chunk of the mapping/security descriptor code.

Does that sound reasonable, or am I missing some Windhawk-specific detail about how shared sections behave between the x86/x64 builds?

m417z:
> Since Windhawk builds separate x86/x64 DLLs, a shared section wouldn't be shared between those two images, right? So we'd still need something else for the state that has to cross that boundary.

Right, that might be a deal-breaker I haven't considered it since it's not an issue in my x64-only mod.

PR #5464, opened 2026-09-12
Add Premiere Pro Theme

CakeDev4k:
A theme mod for Adobe Premiere Pro.

Premiere paints its interface through five separate mechanisms, and the mod covers all five: the `dvaui.dll` color functions (the classic theme, the Spectrum gray ramp, and the DNA/skins families), Premiere's own `UIFramework.dll` drawing primitives, the Direct2D brush factory in `dvaui`, the Win32 title bar and menu bar, and the UXP panels (Text, Import, Export, the Home screen), which draw from stylesheets of their own.

Fourteen palettes, plus custom themes read from a single JSON setting so they can be shared as text. Seven of the palettes, and any custom theme that names one, also give Premiere's interface blue — track targeting, the focus ring, the active tool — their own hue, each blue at its own brightness.

Notes for reviewers:

- Every color function is resolved by `GetProcAddress` on its mangled name at startup. What the running build exports gets hooked; the rest is logged and skipped, so a Premiere version that moved or dropped a function loses that surface rather than breaking the mod. Measured against the export tables: Premiere 2026 matches 31 of 31, Premiere 2023 matches 27 of 31.
- Between those two versions Adobe swapped `boost::intrusive_ptr` for its own `IntrusivePtr` in half of those signatures, which changes the mangled name and nothing else. Both spellings are in the table.
- The menu bar uses the `WM_UAHDRAWMENU` technique from [win32-darkmode](https://github.com/adzm/win32-darkmode) (MIT). Credited in the mod readme and in a comment at the implementation.
- UXP panels: a plain read of a `.css` file under Premiere's own `UXP\plugins` folder (kernelbase `CreateFileW` / `CreateFile2`) is answered with a recolored copy in a `FILE_FLAG_DELETE_ON_CLOSE` temp file, created with `CREATE_NEW`. Nothing on disk is modified, and every color keeps its length, because the UXP runtime also asks for file sizes by path. Premiere reads these stylesheets once per session, so on those panels a palette change — or disabling the mod — shows after a restart. The readme says so.
- Premiere's interface blues are matched exactly against 30 Spectrum values, never by hue, so a content color that merely looks blue is left alone.
- Custom themes come from other users, so the JSON reader is strict and bounded (16K characters, 16 levels of nesting): malformed input is rejected whole, Onyx stays, and the log says where it broke.
- One surface is documented as not themed rather than patched blind: the band around the video in the monitors, which Premiere forces to the channel average of the panel color. It is written up with the measurements in the mod's readme.

## Changelog

* New mod.

## Mod authorship

This mod was created by:

- - [x] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/premiere-pro-theme.wh.cpp
+++ b/mods/premiere-pro-theme.wh.cpp
@@ -315,6 +315,7 @@ restart.
 
 #include <stdint.h>
 #include <cmath>
+#include <cstring>
 #include <cwchar>
 #include <utility>
 
@@ -377,22 +378,6 @@ static float Blend(float original, float target) {
     return original * (1.0f - g_settings.strength) + target * g_settings.strength;
 }
 
-static bool ContainsNoCase(const wchar_t* haystack, const wchar_t* needle) {
-    if (!haystack || !needle) {
-        return false;
-    }
-
-    size_t needleLen = wcslen(needle);
-
-    for (const wchar_t* p = haystack; *p; p++) {
-        if (_wcsnicmp(p, needle, needleLen) == 0) {
-            return true;
-        }
-    }
-
-    return false;
-}
-
 // ============================================================================
 // WHERE THE CALL CAME FROM
 // ============================================================================
@@ -407,20 +392,33 @@ static bool ContainsNoCase(const wchar_t* haystack, const wchar_t* needle) {
     optimization decision for correctness is a bug that shows up when the clang
     version changes.
 */
-static bool IsAdobeUICaller(void* caller) {
-    if (!caller) {
-        return false;
-    }
+/*
+    Which modules count as Adobe UI, kept as address ranges rather than asked
+    per call.
 
-    HMODULE module = nullptr;
+    The GDI hooks ask this on every CreateSolidBrush, CreatePen and SetBkColor.
+    Asking it with GetModuleHandleExW took the loader lock each time, which is
+    more expensive than the colour work it guards and, worse, a lock-order
+    hazard: ThemeSysBrush holds g_brushLock while calling CreateSolidBrush, so
+    that path takes g_brushLock and then the loader lock, while a thread already
+    inside the loader calling GetSysColorBrush takes the two the other way
+    round.
 
-    if (!GetModuleHandleExW(
-            GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS |
-                GET_MODULE_HANDLE_EX_FLAG_UNCHANGED_REFCOUNT,
-            reinterpret_cast<LPCWSTR>(caller), &module)) {
-        return false;
-    }
+    The set only changes when a module loads, and LoadLibraryExW_Hook already
+    sees that happen. So it is snapshot once and appended to, and the question
+    becomes a pointer comparison against a handful of ranges.
+*/
+struct ModuleRange {
+    uintptr_t begin;
+    uintptr_t end;
+};
 
+constexpr size_t kMaxModuleRanges = 64;
+
+ModuleRange g_moduleRanges[kMaxModuleRanges];
+volatile LONG g_moduleRangeCount = 0;
+
+static bool IsAdobeUIModule(HMODULE module) {
     if (module == GetModuleHandleW(nullptr)) {
         return true;
     }
@@ -434,8 +432,107 @@ static bool IsAdobeUICaller(void* caller) {
     const wchar_t* name = wcsrchr(path, L'\\');
     name = name ? name + 1 : path;
 
-    // Adobe's whole UI toolkit is prefixed dva (dvaui, dvacore, ...).
-    return _wcsnicmp(name, L"dva", 3) == 0;
+    /*
+        Adobe's whole UI toolkit is prefixed dva (dvaui, dvacore, ...).
+        UIFramework is Premiere's own drawing layer, hooked elsewhere as a
+        first-class paint layer, so it belongs in the same set.
+    */
+    return _wcsnicmp(name, L"dva", 3) == 0 ||
+           _wcsicmp(name, L"UIFramework.dll") == 0;
+}
+
+/*
+    Only ever called from Wh_ModInit, before any hook is live, and from
+    LoadLibraryExW_Hook, which runs under the loader lock and is therefore
+    serialised against itself. That is why the append needs no lock of its own;
+    the count is published last so a concurrent reader never sees a half-written
+    entry.
+*/
+static void NoteAdobeModule(HMODULE module) {
+    if (!module || !IsAdobeUIModule(module)) {
+        return;
+    }
+
+    auto base = reinterpret_cast<uintptr_t>(module);
+    auto dos = reinterpret_cast<const IMAGE_DOS_HEADER*>(module);
+
+    if (dos->e_magic != IMAGE_DOS_SIGNATURE) {
+        return;
+    }
+
+    auto nt = reinterpret_cast<const IMAGE_NT_HEADERS*>(base + dos->e_lfanew);
+
+    if (nt->Signature != IMAGE_NT_SIGNATURE) {
+        return;
+    }
+
+    LONG count = g_moduleRangeCount;
+
+    for (LONG i = 0; i < count; i++) {
+        if (g_moduleRanges[i].begin == base) {
+            return;
+        }
+    }
+
+    if (count >= static_cast<LONG>(kMaxModuleRanges)) {
+        return;
+    }
+
+    g_moduleRanges[count].begin = base;
+    g_moduleRanges[count].end = base + nt->OptionalHeader.SizeOfImage;
+
+    InterlockedExchange(&g_moduleRangeCount, count + 1);
+}
+
+using EnumProcessModules_t = BOOL(WINAPI*)(HANDLE, HMODULE*, DWORD, LPDWORD);
+
+static void SnapshotAdobeModules() {
+    NoteAdobeModule(GetModuleHandleW(nullptr));
+
+    HMODULE kernel32 = GetModuleHandleW(L"kernel32.dll");
+
+    auto enumModules =
+        kernel32 ? reinterpret_cast<EnumProcessModules_t>(
+                       GetProcAddress(kernel32, "K32EnumProcessModules"))
+                 : nullptr;
+
+    if (!enumModules) {
+        // Named fallback, so the common case still works without psapi.
+        NoteAdobeModule(GetModuleHandleW(L"dvaui.dll"));
+        NoteAdobeModule(GetModuleHandleW(L"dvacore.dll"));
+        NoteAdobeModule(GetModuleHandleW(L"UIFramework.dll"));
+        return;
+    }
+
+    HMODULE modules[512];
+    DWORD needed = 0;
+
+    if (!enumModules(GetCurrentProcess(), modules, sizeof(modules), &needed)) {
+        return;
+    }
+
+    DWORD count = needed / sizeof(HMODULE);
+
+    if (count > ARRAYSIZE(modules)) {
+        count = ARRAYSIZE(modules);
+    }
+
+    for (DWORD i = 0; i < count; i++) {
+        NoteAdobeModule(modules[i]);
+    }
+}
+
+static bool IsAdobeUICaller(void* caller) {
+    auto p = reinterpret_cast<uintptr_t>(caller);
+    LONG count = g_moduleRangeCount;
+
+    for (LONG i = 0; i < count; i++) {
+        if (p >= g_moduleRanges[i].begin && p < g_moduleRanges[i].end) {
+            return true;
+        }
+    }
+
+    return false;
 }
 
 // ============================================================================
@@ -577,7 +674,32 @@ struct ColorSlot {
     DvaColorRGBA dst;
 };
 
-ColorSlot g_slots[kSlotCount];
+/*
+    Deliberately heap-allocated and deliberately leaked.
+
+    The pointer this table returns is handed to Premiere and kept by it — the
+    comments on ConvertColorRef and StableConvert spell out that contract. A
+    static array would live in the mod image, and Windhawk unmaps that image on
+    unload, which happens on every settings change and not only on disable. The
+    references Premiere still holds would then point at unmapped memory.
+
+    One table of about 400 KB leaks per load cycle. That is the price of handing
+    out a pointer whose lifetime the mod does not control, and it is the same
+    trade already made for the GetSysColorBrush brushes.
+*/
+ColorSlot* g_slots = nullptr;
+constexpr size_t kSlotBytes = kSlotCount * sizeof(ColorSlot);
+
+static bool AllocateSlots() {
+    if (g_slots) {
+        return true;
+    }
+
+    g_slots = static_cast<ColorSlot*>(
+        HeapAlloc(GetProcessHeap(), HEAP_ZERO_MEMORY, kSlotBytes));
+
+    return g_slots != nullptr;
+}
 
 static bool SameColor(const DvaColorRGBA& a, const DvaColorRGBA& b) {
     return a.r == b.r && a.g == b.g && a.b == b.b && a.a == b.a;
@@ -588,14 +710,22 @@ static bool SameColor(const DvaColorRGBA& a, const DvaColorRGBA& b) {
     "original" when one hook calls another — see ConvertColorRef.
 */
 static bool IsOurSlot(const void* address) {
+    if (!g_slots) {
+        return false;
+    }
+
     const char* p = reinterpret_cast<const char*>(address);
     const char* begin = reinterpret_cast<const char*>(g_slots);
 
-    return p >= begin && p < begin + sizeof(g_slots);
+    return p >= begin && p < begin + kSlotBytes;
 }
 
 static const DvaColorRGBA* StoreColor(uintptr_t key, const DvaColorRGBA& src,
                                       const DvaColorRGBA& dst) {
+    if (!g_slots) {
+        return nullptr;
+    }
+
     size_t index = static_cast<size_t>(((key >> 4) * 0x9E3779B97F4A7C15ull) >> 51) &
                    (kSlotCount - 1);
 
@@ -674,83 +804,9 @@ static size_t ProducedIndex(LONG key) {
            (kProducedSlots - 1);
 }
 
-/*
-    The undo table: for each grey level, the palette colour whose channels
-    average to it.
-
-    Premiere neutralises the surface around the picture by averaging the
-    channels of the panel colour — measured live at #101907 -> #101010, and
-    (16 + 25 + 7) / 3 is exactly 16. The averaging happens inside Premiere,
-    after every hook has returned, and the result is painted on the GPU where no
-    colour function can see it.
-
-    Averaging is not reversible in general. It is reversible here, because the
-    input was our own: whatever grey arrives, if some colour this mod produced
-    averages to exactly that grey, then that grey IS that colour with the hue
-    taken out, and the colour can simply be put back.
-
-    Indexed by grey level, so the lookup is an array read. Two palette colours
-    can share an average; the later one wins, and since they are neighbouring
-    tones of the same ramp the difference is invisible.
-*/
-struct DesatEntry {
-    DvaColorRGBA color;
-    volatile LONG set;
-};
-
-DesatEntry g_desatMap[256];
-
-static void RememberDesaturated(const DvaColorRGBA& c, int r, int g, int b) {
-    int avg = (r + g + b) / 3;
-
-    g_desatMap[avg].color = c;
-    InterlockedExchange(&g_desatMap[avg].set, 1);
-}
-
-/*
-    Only an exact neutral counts. A colour that still has a hue was never
-    averaged, and a grey that matches no produced colour is somebody else's —
-    a black video frame, a grey matte — and is left alone.
-*/
-static bool UndoDesaturation(const DvaColorRGBA* in, DvaColorRGBA* out) {
-    if (!g_settings.brushHook || !in) {
-        return false;
-    }
-
-    if (!IsSaneChannel(in->r) || !IsSaneChannel(in->g) || !IsSaneChannel(in->b)) {
-        return false;
-    }
-
-    int r = ClampInt(static_cast<int>(in->r * 255.0f + 0.5f), 0, 255);
-    int g = ClampInt(static_cast<int>(in->g * 255.0f + 0.5f), 0, 255);
-    int b = ClampInt(static_cast<int>(in->b * 255.0f + 0.5f), 0, 255);
-
-    if (r != g || g != b) {
-        return false;
-    }
-
-    if (!g_desatMap[r].set) {
-        return false;
-    }
-
-    const DvaColorRGBA& original = g_desatMap[r].color;
-
-    // The colour comes back; the alpha belongs to the caller, not to us.
-    out->r = original.r;
-    out->g = original.g;
-    out->b = original.b;
-    out->a = in->a;
-
-    return true;
-}
-
 static void RememberProduced(const DvaColorRGBA& c) {
     LONG key = PackColorKey(c);
     size_t start = ProducedIndex(key);
-
-    RememberDesaturated(c, ClampInt(static_cast<int>(c.r * 255.0f + 0.5f), 0, 255),
-                        ClampInt(static_cast<int>(c.g * 255.0f + 0.5f), 0, 255),
-                        ClampInt(static_cast<int>(c.b * 255.0f + 0.5f), 0, 255));
 
     for (size_t probe = 0; probe < 32; probe++) {
         size_t i = (start + probe) & (kProducedSlots - 1);
@@ -1027,13 +1083,13 @@ static void InstallOneColorHook(HMODULE dvaui, size_t index, void* hook,
     FARPROC proc = GetProcAddress(dvaui, sym.mangled);
 
     if (!proc) {
-        Wh_Log(L"[Theme] absent in this version: %s", sym.label);
+        Wh_Log(L"absent in this version: %s", sym.label);
         g_colorHooksMissing++;
         return;
     }
 
     if (!Wh_SetFunctionHook(reinterpret_cast<void*>(proc), hook, original)) {
-        Wh_Log(L"[Theme] failed to hook %s", sym.label);
+        Wh_Log(L"failed to hook %s", sym.label);
         g_colorHooksMissing++;
         return;
     }
@@ -1309,15 +1365,17 @@ static bool RecolorParamColor(const void* color, unsigned char* copy) {
 
     DvaColorRGBA out{};
 
-    /*
-        Undo first, convert second. A grey that is one of our own colours with
-        the hue averaged out should come back as that exact colour, not as
-        whatever a fresh conversion of the grey would produce.
-    */
-    if (!UndoDesaturation(&in, &out) && !ConvertForPaint(&in, &out)) {
+    if (!ConvertForPaint(&in, &out)) {
         return false;
     }
 
+    /*
+        Sixteen bytes, not four. The disassembly of Adobe's converter proves
+        bytes 0, 1 and 2 are the channels; it proves nothing about the size of
+        the struct, and this buffer is what the original function reads. Four
+        bytes are copied because four are what is known; the rest stays zeroed
+        so a larger struct still reads defined memory.
+    */
     memcpy(copy, color, 4);
     copy[0] = static_cast<unsigned char>(
         ClampInt(static_cast<int>(out.r * 255.0f + 0.5f), 0, 255));
@@ -1331,7 +1389,7 @@ static bool RecolorParamColor(const void* color, unsigned char* copy) {
 
 void UifFillRectParam_Hook(void* self, const void* rect, const void* color,
                            unsigned char flags) {
-    unsigned char copy[4];
+    unsigned char copy[16]{};
 
     if (RecolorParamColor(color, copy)) {
         UifFillRectParam_Original(self, rect, copy, flags);
@@ -1342,7 +1400,7 @@ void UifFillRectParam_Hook(void* self, const void* rect, const void* color,
 }
 
 void UifFrameRectParam_Hook(void* self, const void* rect, const void* color) {
-    unsigned char copy[4];
+    unsigned char copy[16]{};
 
     if (RecolorParamColor(color, copy)) {
         UifFrameRectParam_Original(self, rect, copy);
@@ -1386,14 +1444,14 @@ static void InstallUifHooks(HMODULE uif) {
         FARPROC proc = GetProcAddress(uif, spec.mangled);
 
         if (!proc) {
-            Wh_Log(L"[Theme] absent in this version: %s", spec.label);
+            Wh_Log(L"absent in this version: %s", spec.label);
             g_colorHooksMissing++;
             continue;
         }
 
         if (!Wh_SetFunctionHook(reinterpret_cast<void*>(proc), spec.hook,
                                 spec.original)) {
-            Wh_Log(L"[Theme] failed to hook %s", spec.label);
+            Wh_Log(L"failed to hook %s", spec.label);
             g_colorHooksMissing++;
             continue;
         }
@@ -1441,14 +1499,14 @@ static void InstallBrushHooks(HMODULE dvaui) {
         FARPROC proc = GetProcAddress(dvaui, spec.mangled);
 
         if (!proc) {
-            Wh_Log(L"[Theme] absent in this version: %s", spec.label);
+            Wh_Log(L"absent in this version: %s", spec.label);
             g_colorHooksMissing++;
             continue;
         }
 
         if (!Wh_SetFunctionHook(reinterpret_cast<void*>(proc), spec.hook,
                                 spec.original)) {
-            Wh_Log(L"[Theme] failed to hook %s", spec.label);
+            Wh_Log(L"failed to hook %s", spec.label);
             g_colorHooksMissing++;
             continue;
         }
@@ -1465,49 +1523,100 @@ static void InstallBrushHooks(HMODULE dvaui) {
 static bool g_dvauiHooked = false;
 static bool g_uifHooked = false;
 
-static int InstallDvaHooks(bool applyNow) {
-    g_colorHooksInstalled = 0;
-    g_colorHooksMissing = 0;
+/*
+    Registers the hooks for one module, if it is one of ours and not done yet.
 
-    HMODULE dvaui = GetModuleHandleW(L"dvaui.dll");
+    Returns whether anything was registered, so the caller knows whether an
+    apply is needed. Nothing here applies the operations itself: at init
+    Windhawk applies them when Wh_ModInit returns, and from the loader hook the
+    caller applies them once for the module that just arrived.
+*/
+static bool HandleLoadedModule(HMODULE module) {
+    if (!module) {
+        return false;
+    }
 
-    if (dvaui && !g_dvauiHooked) {
-        InstallColorHooksImpl(dvaui,
-                              std::make_index_sequence<kColorSymbolCount>{});
+    bool registered = false;
+
+    if (!g_dvauiHooked && module == GetModuleHandleW(L"dvaui.dll")) {
+        g_colorHooksInstalled = 0;
+        g_colorHooksMissing = 0;
+
+        if (g_settings.dvauiHook) {
+            InstallColorHooksImpl(module,
+                                  std::make_index_sequence<kColorSymbolCount>{});
+        }
 
         if (g_settings.brushHook) {
-            InstallBrushHooks(dvaui);
+            InstallBrushHooks(module);
         }
 
         g_dvauiHooked = true;
+        registered = true;
+
+        if (g_colorHooksInstalled) {
+            Wh_Log(L"dvaui: %d hooks active, %d absent in this version",
+                   g_colorHooksInstalled, g_colorHooksMissing);
+        } else if (g_settings.dvauiHook) {
+            Wh_Log(L"no dvaui color entry point matched — this Premiere version "
+                   L"is not supported by the interface layer. The window frame "
+                   L"and menus still apply.");
+        }
     }
 
-    HMODULE uif = GetModuleHandleW(L"UIFramework.dll");
-
-    if (uif && !g_uifHooked && g_settings.brushHook) {
-        InstallUifHooks(uif);
+    if (!g_uifHooked && g_settings.brushHook &&
+        module == GetModuleHandleW(L"UIFramework.dll")) {
+        InstallUifHooks(module);
         g_uifHooked = true;
+        registered = true;
     }
 
-    if (!g_colorHooksInstalled) {
-        return 0;
+    return registered;
+}
+
+static bool HandleAlreadyLoadedModules() {
+    bool registered = HandleLoadedModule(GetModuleHandleW(L"dvaui.dll"));
+
+    return HandleLoadedModule(GetModuleHandleW(L"UIFramework.dll")) || registered;
+}
+
+/*
+    Why the loader instead of a thread.
+
+    dvaui and UIFramework are not necessarily loaded when the mod initialises —
+    enable the mod before Premiere starts and neither is. The previous answer
+    was a thread that polled GetModuleHandleW every 100 ms and gave up after two
+    minutes, which had two problems beyond the polling itself: a mod that starts
+    late enough missed its window silently, and the thread could still be inside
+    Wh_SetFunctionHook when Windhawk tore the mod down.
+
+    Hooking the loader removes all of it. The hook goes in kernelbase, not
+    kernel32, because kernel32's export is only a forwarder and Premiere's own
+    calls go straight to kernelbase.
+*/
+using LoadLibraryExW_t = HMODULE(WINAPI*)(LPCWSTR, HANDLE, DWORD);
+
+LoadLibraryExW_t LoadLibraryExW_Original = nullptr;
+
+HMODULE WINAPI LoadLibraryExW_Hook(LPCWSTR fileName, HANDLE file, DWORD flags) {
+    HMODULE module = LoadLibraryExW_Original(fileName, file, flags);
+
+    constexpr DWORD kDataOnly = LOAD_LIBRARY_AS_DATAFILE |
+                                LOAD_LIBRARY_AS_DATAFILE_EXCLUSIVE |
+                                LOAD_LIBRARY_AS_IMAGE_RESOURCE;
+
+    // A data-only mapping has no code in it, so there is nothing to hook.
+    if (!module || (flags & kDataOnly)) {
+        return module;
     }
 
-    if (g_colorHooksInstalled && applyNow && !Wh_ApplyHookOperations()) {
-        Wh_Log(L"[Theme] failed to apply the dvaui hooks");
-        return 0;
+    NoteAdobeModule(module);
+
+    if (HandleLoadedModule(module) && !Wh_ApplyHookOperations()) {
+        Wh_Log(L"failed to apply hooks for a late-loaded module");
     }
 
-    if (!g_colorHooksInstalled) {
-        Wh_Log(L"[Theme] no dvaui color entry point matched — this Premiere "
-               L"version is not supported by the interface layer. The window "
-               L"frame and menus still apply.");
-    } else {
-        Wh_Log(L"[Theme] dvaui: %d hooks active, %d absent in this version",
-               g_colorHooksInstalled, g_colorHooksMissing);
-    }
-
-    return g_colorHooksInstalled;
+    return module;
 }
 
 // ============================================================================
@@ -1578,7 +1687,7 @@ static void InitNativeDarkMode() {
         GetProcAddress(g_uxtheme, "SetWindowTheme"));
 
     if (g_buildNumber < 17763) {
-        Wh_Log(L"[Theme] build %u has no native dark mode", g_buildNumber);
+        Wh_Log(L"build %u has no native dark mode", g_buildNumber);
         return;
     }
 
@@ -1593,6 +1702,16 @@ static void InitNativeDarkMode() {
         GetProcAddress(g_uxtheme, MAKEINTRESOURCEA(136)));
 
     FARPROC ordinal135 = GetProcAddress(g_uxtheme, MAKEINTRESOURCEA(135));
+
+    /*
+        Everything above is a lookup and changes nothing; the menu hooks need
+        those entry points whatever this setting says. What follows changes the
+        process-wide app mode — scrollbars, common dialogs, control themes — so
+        it belongs to "Window and system dialogs" and has to answer to it.
+    */
+    if (!g_settings.nativeDarkMode) {
+        return;
+    }
 
     if (g_buildNumber < 18362) {
         g_AllowDarkModeForApp = reinterpret_cast<AllowDarkModeForApp_t>(ordinal135);
@@ -1639,8 +1758,14 @@ static void ApplyDarkModeToWindow(HWND hwnd) {
                           static_cast<DWMWINDOWATTRIBUTE>(kUseImmersiveDarkMode),
                           &dark, sizeof(dark));
 
-    // Title bar and border only exist on the top-level window.
-    if (GetParent(hwnd)) {
+    /*
+        Title bar and border only exist on a top-level window, and GetParent is
+        the wrong way to ask: for a WS_POPUP it returns the OWNER, so Premiere's
+        owned dialogs and floating panels — which do have captions — were taking
+        this early return and keeping the default colours. The style bit answers
+        the question that was actually being asked.
+    */
+    if (GetWindowLongPtrW(hwnd, GWL_STYLE) & WS_CHILD) {
         return;
     }
 
@@ -1667,6 +1792,73 @@ static void ApplyDarkModeToWindow(HWND hwnd) {
                           &caption, sizeof(caption));
     DwmSetWindowAttribute(hwnd, static_cast<DWMWINDOWATTRIBUTE>(kTextColor), &text,
                           sizeof(text));
+}
+
+/*
+    Undoing the frame.
+
+    DwmSetWindowAttribute and SetWindowTheme stick to a window for its lifetime,
+    so without this the caption, border and text stay themed after the mod is
+    disabled, until Premiere restarts. Windhawk's principle is that a mod's
+    effects go away when it does, and since the reload path runs on every
+    settings change, this also stops an old palette from surviving a change.
+*/
+static void RevertWindowFrame(HWND hwnd) {
+    if (!hwnd) {
+        return;
+    }
+
+    if (g_SetWindowTheme) {
+        g_SetWindowTheme(hwnd, nullptr, nullptr);
+    }
+
+    constexpr DWORD kUseImmersiveDarkMode = 20;
+    constexpr DWORD kBorderColor = 34;
+    constexpr DWORD kCaptionColor = 35;
+    constexpr DWORD kTextColor = 36;
+
+    BOOL dark = FALSE;
+
+    DwmSetWindowAttribute(hwnd,
+                          static_cast<DWMWINDOWATTRIBUTE>(kUseImmersiveDarkMode),
+                          &dark, sizeof(dark));
+
+    // DWMWA_COLOR_DEFAULT: hand the colour back to the system.
+    COLORREF automatic = 0xFFFFFFFF;
+
+    const DWORD colorAttributes[] = {kBorderColor, kCaptionColor, kTextColor};
+
+    for (DWORD attribute : colorAttributes) {
+        DwmSetWindowAttribute(hwnd, static_cast<DWMWINDOWATTRIBUTE>(attribute),
+                              &automatic, sizeof(automatic));
+    }
+}
+
+static BOOL CALLBACK RevertChild(HWND hwnd, LPARAM) {
+    RevertWindowFrame(hwnd);
+    return TRUE;
+}
+
+static BOOL CALLBACK RevertTopLevel(HWND hwnd, LPARAM) {
+    DWORD pid = 0;
+    GetWindowThreadProcessId(hwnd, &pid);
+
+    if (pid != GetCurrentProcessId()) {
+        return TRUE;
+    }
+
+    RevertWindowFrame(hwnd);
+    EnumChildWindows(hwnd, RevertChild, 0);
+
+    if (GetMenu(hwnd)) {
+        DrawMenuBar(hwnd);
+    }
+
+    return TRUE;
+}
+
+static void RevertThemeFromExistingWindows() {
+    EnumWindows(RevertTopLevel, 0);
 }
 
 static BOOL CALLBACK ApplyToChild(HWND hwnd, LPARAM) {
@@ -2021,7 +2213,14 @@ static HTHEME OpenDarkMenuTheme(HWND hwnd, LPCWSTR classList,
 }
 
 static HTHEME TrackMenuTheme(HTHEME theme, LPCWSTR classList) {
-    if (theme && ContainsNoCase(classList, L"Menu")) {
+    /*
+        Exactly "Menu", the same test OpenDarkMenuTheme uses to decide what to
+        swap. A substring match would also register any class whose name merely
+        contains "menu", and PaintMenuPart would then repaint parts 7-15 of it
+        with menu colours — part numbers mean different things in different
+        classes, so that shows up as a corrupted control.
+    */
+    if (theme && classList && _wcsicmp(classList, L"Menu") == 0) {
         RememberMenuTheme(theme);
     }
 
@@ -2798,7 +2997,7 @@ static COLORREF ReadColorSetting(PCWSTR name, COLORREF fallback) {
     COLORREF parsed = fallback;
 
     if (!ParseHexColor(text, &parsed)) {
-        Wh_Log(L"[Theme] invalid value in %s, using the default", name);
+        Wh_Log(L"invalid value in %s, using the default", name);
         parsed = fallback;
     }
 
@@ -2926,55 +3125,18 @@ static void LoadSettings() {
 }
 
 // ============================================================================
-// WAITING FOR DVAUI
-// ============================================================================
-
-HANDLE g_worker = nullptr;
-volatile LONG g_stopping = FALSE;
-
-static bool ShouldStop() {
-    return InterlockedCompareExchange(&g_stopping, FALSE, FALSE) != FALSE;
-}
-
-/*
-    When the mod is enabled with Premiere already open, dvaui.dll is already loaded
-    and the hooks go in during Wh_ModInit. This thread only exists for the other
-    case: the mod enabled before Premiere starts, when Wh_ModInit runs in a process
-    that has not even loaded the UI toolkit yet.
-*/
-static DWORD WINAPI DvaWorker(LPVOID) {
-    constexpr int kMaxAttempts = 1200;  // 1200 x 100 ms = 2 minutes
-
-    for (int attempt = 0; attempt < kMaxAttempts && !ShouldStop(); attempt++) {
-        InstallDvaHooks(true);
-
-        if (g_dvauiHooked && (!g_settings.brushHook || g_uifHooked)) {
-            break;
-        }
-
-        Sleep(100);
-    }
-
-    if (!ShouldStop()) {
-        ApplyThemeToExistingWindows();
-    }
-
-    return 0;
-}
-
-// ============================================================================
 // LIFECYCLE
 // ============================================================================
 
 static void HookOrLog(void* target, void* hook, void** original,
                       const wchar_t* label) {
     if (!target) {
-        Wh_Log(L"[Theme] target absent: %s", label);
+        Wh_Log(L"target absent: %s", label);
         return;
     }
 
     if (!Wh_SetFunctionHook(target, hook, original)) {
-        Wh_Log(L"[Theme] failed to hook %s", label);
+        Wh_Log(L"failed to hook %s", label);
     }
 }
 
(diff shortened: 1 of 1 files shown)


CakeDev4k:
Thanks — this was a genuinely useful review. The lifecycle framing was the part I had wrong: the mod was written as if it stays loaded for the life of the process, and `Wh_ModSettingsChanged` asking for a reload on every settings change makes that assumption fail constantly, not rarely. Nine of the ten are fixed; the tenth needs me to capture images and is noted below.

**1. Dangling colour pointers.** Fixed. `g_slots` is now `HeapAlloc(GetProcessHeap(), ...)` and deliberately leaked, with the reasoning written next to it; `IsOurSlot` is a range check against `g_slots` / `g_slots + kSlotBytes`, and `StoreColor` returns `nullptr` if the allocation failed, which the callers already handled. This was the one that could actually crash a running Premiere, so thank you for catching it.

**2 and 3. The worker thread.** Fixed together by removing it. `kernelbase!LoadLibraryExW` is hooked, `HandleLoadedModule` registers the hooks for `dvaui.dll` / `UIFramework.dll` when they arrive, and `HandleAlreadyLoadedModules` covers the already-loaded case at init. No thread, no `WaitForSingleObject`, no two-minute cliff. `Wh_ModBeforeUninit` is gone since it only existed for the wait.

**4. Settings that didn't gate.** Both fixed. `InitNativeDarkMode` keeps the `GetProcAddress` lookups unconditionally — `menuHook` needs them — and returns before the app-mode calls when `nativeDarkMode` is off. The `dvauiHook` check moved inside the module handler, so the late-load path respects it too.

**5. `GetParent` vs `WS_CHILD`.** Fixed. You're right that it returns the owner for a popup; Premiere's owned dialogs were taking the early return and keeping default caption colours.

**6. Frame not reverted.** Fixed. `RevertThemeFromExistingWindows` in `Wh_ModUninit` sets the three DWM attributes back to `DWMWA_COLOR_DEFAULT`, clears immersive dark mode, and calls `SetWindowTheme(hwnd, nullptr, nullptr)`. It runs before `FreeLibrary(g_uxtheme)` so `SetWindowTheme` is still resolvable.

**7. `IsAdobeUICaller`.** Fixed. Module ranges are snapshot at init (via `K32EnumProcessModules`, with a named fallback so no psapi link is needed) and appended from the `LoadLibraryExW` hook, so the question is a pointer comparison. The ABBA you describe — `g_brushLock` then loader lock against loader lock then `g_brushLock` — is real and this removes it.

**8. `ApplyThemeToExistingWindows` before hooks are live.** Fixed, moved to `Wh_ModAfterInit`.

**9. No screenshots.** Not done yet — I need to capture and host them, and I'd rather not hold the rest of the fixes behind that. They're coming in a follow-up push before I mark this ready for a human reviewer.

**10. README and code disagreeing about the monitor surround.** You were right that one of them was stale, and it was the code. The README is accurate: the band is still neutral on every tinted palette, the undo never recovered it, and the nine instrumented entry points never saw the grey. So `g_desatMap` / `RememberDesaturated` / `UndoDesaturation` are removed rather than described — they were dead for the case they were written for, and as your functionality note points out, still live enough to recolour somebody else's neutral fill (Glitch `ramp[0]` averages to 6, so a genuine `#060606` would have come back green). Unproven benefit against a real way to paint the wrong colour.

From the optional list I took the ones that were correctness rather than polish: the exact class compare in `TrackMenuTheme` (a substring match really would have let `PaintMenuPart` repaint parts 7-15 of an unrelated class), a 16-byte zeroed buffer for the `ParamColor` copy since the disassembly only proves which bytes are *read*, `<cstring>` instead of relying on `windows.h`, and dropping the redundant `[Theme]` log prefix.

Left alone, deliberately: the `WindhawkUtils::SetFunctionHook` / `StringSetting` conversions, which are a large mechanical diff over working code and would make this review harder to re-read; the per-thread `HTHEME` leak, which I can't reach from `Wh_ModUninit` since it lives in other threads' `thread_local` storage — it's one handle per UI thread per load cycle and I'd rather name it than pretend to fix it; and `customDimText`, which is a feature rather than a fix.

On the functionality notes: the menu bar dimming when Premiere wasn't foreground was **not** intentional — `GetForegroundWindow() != hwnd` was folded into `disabled`, so alt-tabbing greyed the whole bar. Fixed. `IsAdobeUICaller` excluding `UIFramework.dll` was an oversight and it's in the set now, which is consistent with hooking it as a paint layer elsewhere. The `PaintMenuBarBottomLine` flicker I agree with and haven't found a better hook for that 1px line either.

CakeDev4k:
Thanks. Both code items were right, and fixing the second changed how I fixed the first. Testing the result in Premiere also turned up bugs of my own, covered in their own section below.

**2. Settings changes no longer reload the mod.** I went with the thorough fix. Every hook checks its own setting each time it runs, so it is installed once and left in. `Wh_ModSettingsChanged` sets `*bReload = FALSE` and handles the state that lives outside the hooks:

- `RecomputeColorTable` rewrites `dst` in place for every ready slot, so references Premiere already holds pick up the new palette. Each slot records which hook handed its address out (`kFromThemeFunction` / `kFromEraseColor`). Switching "Premiere interface" or "Direct fills" off gives those references their original colour back, and a slot reached through both paths follows whichever setting is still on.
- A slot whose setting is off holds its original colour again, and the brush layer treats it as a raw colour rather than as already converted. So with "Premiere interface" off and "Direct fills" on, the references Premiere holds are still themed, the same as after a restart with those settings.
- A per-slot generation closes the race where a drawing thread converts under the old settings and writes after the walk. A write computed under an older generation never replaces a slot the walk already recomputed, and a slot that was still being filled when the walk ran is refreshed by the thread filling it as it publishes it.
- The produced set is cleared and reseeded from the table.
- On unload the same walk sets every `dst` back to `src`. The table is still leaked, because Premiere holds pointers into it, but those surfaces no longer keep the last palette after the mod is gone.
- System brushes are cached by colour instead of by index. A palette change adds only the few colours the new palette uses, switching back reuses them, and the cache is capped.
- The callback also drives the app mode, the window frames, the menu theme cache, and a `RedrawWindow(RDW_INVALIDATE | RDW_ERASE | RDW_FRAME | RDW_ALLCHILDREN)`. That call only invalidates, so it never waits on a UI thread. `RDW_ERASE` is there because native dialogs paint their background in `WM_ERASEBKGND`.

The dvaui and UIFramework hooks are the one exception to "installed once". With "Premiere interface", "Direct fills" and "GDI surfaces" all off when the mod loads, Premiere's own modules are not hooked at all. Turning those three off and restarting Premiere is the way back if an update ever breaks one of those hooks. Switching one of them on later hooks the modules from the settings callback.

A settings change now leaks nothing. The slot table and the per-thread `HTHEME` leak once per unload instead of once per change. I also shrank `kSlotCount` to 1024, from a measurement rather than a guess. `Wh_ModUninit` now logs how many slots were in use, and a Premiere 2026 session with a project open used 12. The key is the colour's *address*, and the theme holds few colour objects. The table is now about 56 KB, and if it ever fills, `StoreColor` returns nothing and the caller keeps the original colour. With the reload gone, I left out the per-thread `HTHEME` registry, following your own qualifier on it.

**1. The revert now mirrors the apply exactly.** `ApplyDarkModeToWindow` records in a map each window it changed and what it changed on it: the theme class, the frame, or both. The revert walks that map and nothing else, so with the setting off it does nothing. The map is swapped out under the lock before any `SetWindowTheme` is sent, so a UI thread blocked on that lock inside `CreateWindowExW_Hook` can't deadlock against the send. The app mode follows the same rule: `ApplyAppMode` only undoes a mode it set itself, and it makes no call at all when the setting is off from the start.

`SetWindowTheme` is also hooked, only to notice when something else re-themes a window after the mod did. That drops the mod's claim on the window's theme class, so the revert doesn't clear a class that someone else chose.

Two points where I measured instead of taking the premise:

- *Reverting only what was applied* doesn't shrink the work to a handful of windows here, because the apply runs from `CreateWindowExW` and reaches every window. A running Premiere 2026 with a project open has 227 windows across 43 threads. What changes is *when* the revert runs: on disable, or when "Window and system dialogs" is switched off, and no longer on every settings change. A palette change with the frame still on only recolours the captions of the tracked top-level windows through DWM.
- **Immersive dark mode on revert.** I kept restoring `FALSE`, but only on the top-level windows where the mod set `TRUE`. VS Code sets the attribute itself; Premiere doesn't. In the Premiere 2026 install folder, the executable, `dvaui.dll`, `dvacore.dll`, `UIFramework.dll` and `PlugPlug.dll` don't reference `DwmSetWindowAttribute` at all, and none of the top-level windows is a Chromium frame. So for the windows in the map, `FALSE` is the value they had before the mod ran.

**3. Screenshots.** They're at the top of the readme, hosted on raw.githubusercontent.com: stock Premiere for comparison, then the same project in Onyx, Premiere, Comfy, Neon and Glitch.

**Found while testing this round**

- **The round-1 late-load path would never have fired for these modules.** `dvaui.dll` and `UIFramework.dll` are static imports of `Frontend.dll`, which the executable imports. They arrive through the loader and never come back from `LoadLibraryExW`, so comparing the returned handle missed them. The hook now checks after every successful load whether each module is loaded yet. Once both are hooked, that check is just two flag reads. Claiming a module is atomic, because Premiere loads libraries from several threads. In testing, dvaui was already mapped at `Wh_ModInit`, but that depends on injection timing.
- **Module ranges.** Premiere 2026 runs with 567 modules, and the snapshot held 512, so dva* modules past that point were never recognised.
  - The list is now sized from what `K32EnumProcessModules` reports.
  - Sizes come from `K32GetModuleInformation` rather than from reading headers. If the process can't be enumerated, the executable, dvaui, dvacore and UIFramework are still recorded from their own headers.
  - Modules mapped later, including the dva* modules that arrive as dependencies, are added from an `LdrRegisterDllNotification` callback, which is unregistered in `Wh_ModUninit`.
  - The old comment claiming the append was serialised by the loader lock was wrong: the hook runs after `LoadLibraryExW` returns. Appends now take their own SRWLOCK, and nothing that can request the loader lock runs while it is held.
- **Content colours were being themed.** Picking `#373737` in Premiere's colour picker showed the swatch in the palette's green, because the brush hooks decide by colour alone. The stored value was right; the swatch was wrong. Now:
  - Swatch drawing runs inside a thread-local `ContentScope`, in which the paint-level hooks pass colours through unchanged. That covers `UI_Swatch::UI_Draw`, which the colour picker, marker colours and Essential Graphics all use, plus the Spectrum/csnext swatch skins and the popup colour swatches.
  - Colours handed out by the theme functions are still converted before they get there, so the swatch chrome stays themed.
  - The `ParamColor` overloads of `UIF::DC::FillRect`/`FrameRect` no longer convert at all. Every module that imports them is a content panel, and the diagnostic log never saw an interface gray arrive there.
  - The colour picker's chip, the new colour over the original one, is a `ColorPickerChipView`, and neither its draw function nor its vtable is exported. So the mod hooks `UI_Node::UI_DrawSelf` and `UI_DrawAndCache`, which every node is drawn through, and recognises the node by its most-derived class from MSVC RTTI. Each step from the vtable to the type name is checked to lie in the same image. The answer is cached per vtable in a lock-free table, so each class is resolved once. The lookup is skipped entirely while neither "Direct fills" nor "GDI surfaces" is on.
  - A scope also records the frame of the hook that opened it. If an exception from Adobe's code unwinds past the hook without running its destructor, which the mod's unwinder isn't guaranteed to do for a foreign exception, the next check from higher up the stack clears the scope. Otherwise the thread would be left passing every colour through.
- **The monitor surround.** The readme's old explanation was wrong, and so was my first correction. Measured palette by palette, the monitors paint the band around a zoomed-out picture a gray taken from the *red channel* of the panel colour:
  - Onyx `#0C0C0C` gives `#0C0C0C`.
  - Glitch `#101907` gives `#101010`.
  - Neon `#150D14` gives `#151515`; the channel average would be 18.

  The monitor also keeps that gray until Premiere restarts, so after a live palette switch it shows the previous palette's value. That is how I first misread it as a constant. None of the nine entry points a diagnostic build recorded ever sees that gray, so the readme documents it as a limitation rather than the mod patching it blind.
- **Immersive dark mode on older Windows 10.** `DWMWA_USE_IMMERSIVE_DARK_MODE` is 20 from build 18985 on; 1809 to 1909 use 19 and reject 20, so the frame never went dark there. The attribute now follows the build.

From the optional list I also took:

- Immersive dark mode is set below the `WS_CHILD` test.
- `CreateWindowExA` is hooked.
- `ForgetMenuTheme` scans under the shared lock, bounded by the high-water mark, and only takes the lock exclusively for a handle that is actually a menu theme.
- `SetMenuInfo_Hook` only copies when `cbSize == sizeof(MENUINFO)`.
- `PaintMenuBarItem` resolves the theme and `DrawThemeTextEx` before filling.
- The MDI comment now says what the code does.
- The plain-signature hooks go through `WindhawkUtils::SetFunctionHook` via a typed `HookOrLog`.
- `LoadSettings` builds a local copy and publishes it with a single assignment.
- `RememberProduced` reads before it does a `lock cmpxchg`.

I left out the per-DPI menu bar theme cache: it only matters on mixed-DPI setups, and I have no way to test it.

Verification: it compiles cleanly with Windhawk 1.7.3's exact command line, and with `-Wall -Wextra`. I also built the mod source into a test harness with the Windhawk API stubbed out, at `-O2` and `-O0`. Among other things, it checks that:

- a write computed under an older generation leaves a recomputed slot alone;
- a slot filled under old settings comes out current;
- a restored slot is converted by the brush layer;
- a scope whose destructor never ran expires;
- only nodes of a content class get a scope.

In Premiere 2026, switching "Premiere interface" and "Direct fills" on and off applies at once, without a reload. The colour picker's chip now shows the picked colour as it is.

CakeDev4k:

CakeDev4k:
Thanks, both were right.

**1. GDI double conversion.** Fixed as suggested. `ConvertGdiColor` now checks the produced set after the float tests, as `ConvertForPaint` does, and records what it produces. The key is quantised to 8 bits, so a `COLORREF` made from a converted colour compares equal to it.

**2. Failed enumeration.** Fixed, and a little more broadly than suggested. `SnapshotAdobeModules` now records the executable, dvaui, dvacore and UIFramework from their headers *before* it enumerates at all. That covers all three ways the enumeration can fall short:

- the call fails;
- it still comes up short after the retries;
- the `K32*` exports are missing.

A module recorded twice is deduplicated, and a failed call is logged with its error.

From the optional list:

- `Wh_ModSettingsChanged` is the `void` form.
- `NeedsMenuBarWork` tests `WS_CHILD` before `GetMenu`.
- `DrawThemeTextEx_Hook` forwards a larger `DTTOPTS` unchanged.
- The pre-2026 spellings are marked as alternates. A missing alternate isn't counted, and one that resolves takes its newer name's absence back, so the log counts functions rather than spellings.
- The custom palette's disabled-text tone sits halfway between its text and panel colours. That is close to where the built-in palettes put it, and it always stays between the two.

Left as they are:

- **`WS_CAPTION` for the frame attributes.** The attributes are set once, at creation, and a window can gain its caption after `CreateWindowExW` returns. Gating on the creation-time style would miss those windows, and I'd rather pay four `DwmSetWindowAttribute` calls per top-level window than risk a light title bar.
- **Reverting the `SetMenuInfo` brush.** That would need a record of every menu and the brush it had, for menus whose lifetime the mod doesn't control. What stays behind is one of the deliberately leaked, still-valid `g_sysBrushes`, so the worst case after a disable is a menu background in the last palette until that menu is rebuilt.

On the functionality notes: agreed on all of them. The abandoned-scope case is as you describe, since the frame check only clears a scope from a shallower frame. The by-value guard and the one-frame tear are trades I'm keeping.

Verification: it compiles cleanly with 1.7.3's command line and `-Wall -Wextra`, and the harness passes at `-O2` and `-O0`. It now also checks that:

- a GDI colour isn't converted twice, including one made from a theme function's result;
- the known modules are recorded, and only once;
- only the newer names count as missing;
- a larger `DTTOPTS` is forwarded untouched;
- a child window with a nonzero ID fails the menu bar gate;
- the custom palette gets the expected disabled-text tone.

CakeDev4k:
Thanks. For the one required item I checked the premise first, and it took the second of your two fixes.

**`StableConvert` keyed by a caller's address.** I disassembled `dvaui.dll` from Premiere 2023, 2025 and 2026 to see whether either consumer keeps the reference:

- The `ScEnableEraseBackgroundDrawing` constructor converts the color to a `COLORREF`, calls `CreateSolidBrush`, stores the `HBRUSH` in the node manager and deletes the previous one with `DeleteObject`. It never stores the color pointer, and the destructor only decrements a counter. The same in all three versions.
- `UI_DispatchDrawFromRoot` only forwards the pointer to `UI_DrawAndCache`, which spills it to its own stack and passes it on to its draw calls.

Neither keeps it, so both hooks now convert into a stack temporary through `ConvertForPaint`, like the brush hooks. `StableConvert`, `kFromEraseColor` and the per-slot `sources` mask are gone: the table holds only colors the theme functions hand out, and `RefreshSlot` answers to "Premiere interface" alone.

From the optional list:

- **Menu bar theme.** No longer cached per thread. `PaintMenuBarItem` opens the theme for the item and closes it right after, through a small RAII wrapper, using `OpenThemeDataForDpi` with the window's DPI where it exists. Nothing is left open on unload, and this also covers the per-DPI note from the earlier rounds. `WM_THEMECHANGED` and `WM_DESTROY` no longer go through the menu bar path.
- **`kSlotCount`** is 128. With `sources` gone a slot is 48 bytes, so about 6 KB per unload. A full table is logged once. In a Premiere 2026 session, disabling the mod logged `color table: 8 of 128 slots in use`.
- **Silent caps.** `AddModuleRange` logs once when it fills, and so do the color table, the menu theme list and the system brush cache.
- **Hook counters.** No globals: each module fills its own count, and UIFramework logs a line of its own, so the dvaui line means what the readme's first column says. On Premiere 2026 the startup log reads `dvaui: 31 hooks active, 0 absent in this version` and `UIFramework: 4 hooks active, 0 absent in this version`. Testing this turned up a leftover: each of the 14 color functions renamed in 2026 printed `absent in this version` for the spelling that version lacks, right above the summary of 0 absent. Each renamed function is now one table entry carrying both names, and the older name is only tried when the newer one is missing, so a function is logged absent once, and only when neither name exists. The 2026 log now has no `absent` line at all.
- **`WindhawkUtils::StringSetting`** in `LoadSettings` and `ReadColorSetting`. The dead null check in `ParseHexColor` is gone.
- **Prose.** The file is 3,942 lines, down from 4,391; comments went from 1,040 lines to 741, and the readme from 270 to 150. The comments that narrated earlier versions are gone, the palette derivation is stated once, and the readme drops the internals (mangled names, ordinal 49, the icon histogram) and explains the monitor band once. `@description` no longer says "entire", the GDI setting mentions text backgrounds, and the spelling is American throughout.

On the functionality notes:

- **`IsAdobeUICaller`'s linear scan.** The GDI hooks now run the color test first, so the ranges are only walked for dark neutral grays that may be converted, not for every `SetBkColor`. The list stays linear: it is append-only so that readers take no lock, and sorting on insert would need a copy-and-publish scheme for about 61 entries. `ConvertGdiColor` now goes through `ConvertDvaColor` instead of repeating its arithmetic.
- **Other Adobe apps.** Agreed it's the natural next step. I'd rather keep this PR to Premiere, which is what I've tested, and look at widening `@include` in a later version.
- **Blast radius.** Agreed; the readme's all-off fallback is there for that.

Verification: it compiles cleanly with 1.7.3's command line and `-Wall -Wextra`, the harness passes at `-O2` and `-O0`, and `pr_validation.py` run locally reports no warnings. In Premiere 2026 this revision logs only the two startup lines above, and two palette switches with Premiere open each ran `Wh_ModSettingsChanged` to completion, without a reload. The harness now also checks that:

- the erase hooks pass a converted temporary, never a table slot, and forward the original with "Direct fills" off;
- the menu bar theme is opened at the window's DPI, falls back to `Menu`, and is closed once the item is drawn;
- UIFramework is counted separately;
- each renamed color function is counted and logged once under either name, with dvaui's exports faked for 2026, for earlier versions, for a function missing under both names, and for none at all;
- the GDI hooks reject by color before asking who called, and `ConvertGdiColor` gives the same results as the old formula;
- a full table hands back the original color and logs once.

CakeDev4k:
Thanks — both were real. I took the simpler path on each.

**1. `SetMenuInfo_Hook` was not reversible.** You're right that it left a menu on `palette.ramp[1]` after a disable, black-on-black in light mode once `GetSysColor` was no longer hooked. I removed the hook rather than make it reversible: a record keyed by `HMENU` would carry the same recycled-handle hazard as item 2 — a destroyed menu's handle can come back for a new menu, and `IsMenu` would not catch it, so the revert could set a stale brush on an unrelated menu. And the substitution was never observed to be needed. It was defensive, as you read it: of Premiere's modules only `dvaui.dll` and Chromium's `libcef.dll` even import `SetMenuInfo`, and the menu background is already colored by the `DarkMode::Menu` theme and `GetSysColorBrush(COLOR_MENU)`. Dropping it removes the reversibility problem and a process-wide hook that earned its place only in theory.

**2. A recycled `HTHEME` in `g_menuThemes`.** Fixed exactly as you suggested. `TrackMenuTheme` now evicts the handle the moment it reappears for a non-`Menu` class:

```cpp
if (classList && _wcsicmp(classList, L"Menu") == 0) {
    RememberMenuTheme(theme);
} else {
    ForgetMenuTheme(theme);  // recycled for another class; no longer a menu theme
}
```

That closes the window you described: uxtheme frees a non-client theme on window destruction or a theme/DPI change without going through the hooked `CloseThemeData`, then hands the value back for another class, and `PaintMenuPart` would have repainted parts 7–15 of that control in menu colors. Since a live handle is never shared between two open themes, a value reappearing for a non-menu class means the menu theme that held it is already gone, so evicting is safe. `ForgetMenuTheme` already asks `IsMenuTheme` first, so on the common non-menu open it does one shared-lock scan bounded by the high-water mark, and takes the exclusive lock only on a real hit.

I kept the popup path on the tracked-handle model rather than moving it to an owned `Menu` theme like `MenuBarTheme`: the popup draws come through `DrawThemeBackground` with the handle uxtheme opened, and I'd rather recognize that handle than open a parallel theme and hope the two stay in step. With the eviction the table is self-cleaning.

On the optional items and functionality notes — all fair, and I read them. The `g_settings` copy race, the breadth of `ApplyDarkModeToWindow`, the `SendMessage` stalls on a busy UI thread, `ForgetThemedClass`'s exclusive lock, and the five-argument content thunk are all real; I'm keeping the current shape for this version and would revisit them if they bite. On the notes: the per-toggle leaks, the `kSlotCount` degradation mode, the `g_produced` clear during a settings change, `PaintMenuBarBottomLine` covering only the lower hairline, and `IsNeutral`'s tolerance making the low-chroma palettes lean on the produced-color set are all accurate readings.

Verification: compiles cleanly with 1.7.3's command line and `-Wall -Wextra`, the harness passes at `-O2` and `-O0`, and `pr_validation.py` reports no warnings. The harness now also checks that `TrackMenuTheme` registers a `Menu` theme, passes any class through, and evicts a handle that returns for a non-menu class.

CakeDev4k:
Ignore that command mr's windhawk reviewer👆; I'm still going to try making some other adjustments yet

CakeDev4k:
Thanks — both required items held up, and both are fixed.

**1. `DarkMode_Explorer` on every window.** `SetWindowTheme` now goes only to `SysListView32`, `SysTreeView32`, `SysHeader32` and `ScrollBar` (`WantsExplorerTheme`), and only those windows get `kThemedClass`, so the unload revert only touches them too. Every other window keeps `AllowDarkModeForWindow`, which sends nothing. The frame attributes are unchanged.

**2. The permanent produced-color set.** Your arithmetic reproduces exactly: under Onyx `#3F3F3F` converts to `#1D1D1D`, and from then on a direct `#1D1D1D` fill passed unconverted. I took your suggestion. `g_produced` is gone, and each thread keeps a ring of the last 16 colors it produced, emptied at the top of `UI_DispatchDrawFromRoot`. That still covers the two ways a converted color comes back, both on the same thread within the same paint:

- Adobe copies a theme function's result into a temporary or a `COLORREF` and builds a brush from the copy.
- One hooked call passes its converted color into another: `DrawFillRect` into `NewBrush`, or the erase constructor into `CreateSolidBrush`.

`ConvertColorRef` now remembers the slot `dst` it actually hands out, and `RefreshSlot` no longer records anything from the settings thread. `RecomputeColorTable` also no longer zeroes a shared table under the drawing threads, which was one of the optional notes. The harness reproduces your case: after `#3F3F3F` → `#1D1D1D`, a `#1D1D1D` fill is skipped only within that paint, converts in the next one, and also ages out of the ring without a paint boundary.

From the optional list:

- **Stylesheet handle rights.** The mod writes the copy through a handle of its own (`GENERIC_WRITE | DELETE`, delete-on-close), then hands the caller a second handle opened with the caller's own access, security attributes and flags, and closes its own. The harness checks that the returned handle refuses a write.
- **`PaintMenuBarItem`** passes a copy of `rcItem`.
- **One snapshot.** `PaletteTarget` takes the snapshot from its caller, and the conversion blends with that snapshot's strength.

Kept as they are:

- **The JSON theme reader.** Pasting one blob is the point: a Discord community is forming to share themes as text, and grouped settings would lose that. The reader is strict and bounded, and the harness feeds it fifteen malformed inputs, plus oversized and deeply nested ones.
- **`ContentScope`'s nested unwind case.** Agreed that it is best-effort. The frame check in `InContentScope` clears it on the next call from a shallower frame.
- **The unload revert.** Narrowing the theme class removes most of it. The frame attributes still go to every non-child window, since those are the ones with a frame to color.
- **ARM64** is not in scope for this submission; the mod is only tested on x64.
- **The leaks and `MenuBarTheme`.** Agreed, no change.

Verification: it builds cleanly with 1.7.3's command line and `-Wall -Wextra`, the harness passes at `-O2` and `-O0`, and `pr_validation.py` reports no warnings. The harness now lives in the mod's repository under `tests/`.

CakeDev4k:
Thanks — the one required item is fixed.

**`OpenThemeDataEx`.** It is hooked the same way as `OpenThemeDataForDpi`: "Menu" is swapped for `DarkMode::Menu`, and every result goes through `TrackMenuTheme`. A value recycled through it is now evicted like one from the other three openers, and the comment at `TrackMenuTheme` names all four. The harness checks both halves: a "Menu" open is swapped and registered, and the same value handed back for another class is dropped.

From the optional list:

- **Frame work on windows with no frame.** Both `CreateWindowEx` hooks now skip message-only windows (`HWND_MESSAGE` parent). I kept the frame attributes on other non-child windows without a style test. A window can gain a caption after it is created, and a missed window would show a light title bar. On a frameless window the attribute changes nothing visible, and the per-window cost is small.
- **`IsAdobeUICaller`.** Each thread remembers the range it last matched and checks it first, so a run of GDI calls from one Adobe module costs two comparisons. A miss still walks the table, which is append-only and read without a lock.
- **One snapshot per decision** in `ContentScopesMatter`, `WantsPremiereHooks` and `Wh_ModSettingsChanged`.
- **`MenuBarTheme`** logs the case where `CloseThemeData` cannot be resolved.

On the notes:

- **The produced-color ring.** Agreed. Besides the clear at `UI_DispatchDrawFromRoot`, each entry now expires after 250 ms. A thread that never enters a dvaui paint no longer keeps values out of conversion: that covers the GDI and menu paths, and a build where that symbol is missing. The harness checks that an expired entry no longer counts.
- **Stylesheet forms.** Noted. `hsl()` and the space-separated `rgb()` are not in today's Spectrum stylesheets, so they pass through; they are the first thing to add if Adobe restyles those panels.
- **The substituted handle is a different file.** Agreed, that is the failure mode to watch. Each recolored stylesheet is logged by name, and the **UXP panels** setting turns the redirect off.
- **The leaks**, and **the 1px menu bar line at scaled DPI**: noted.

Verification: it builds cleanly with 1.7.3's command line and `-Wall -Wextra`, `tests/run.ps1` passes at `-O2` and `-O0`, and `pr_validation.py` reports no warnings.

CakeDev4k:
C:/Program Files/Git/ready-for-reviewer

CakeDev4k:
Thanks — all three required items are fixed, along with four of the five optional ones. The head also carries a new layer and a new palette, described below so the next review is not surprised by them.

### Required

**1. Custom themes through Windhawk's own settings.** The hand-written JSON reader is gone — about 350 lines of it. `customTheme` is now a nested group of ten fields, read with `WindhawkUtils::StringSetting::make(L"customTheme.base")` and friends. Sharing survives the change: the mod writes the theme in force to the log as one line of YAML flow form, which pastes straight into the Settings tab's text mode. The readme documents that, and the group's description points at it.

**2. `uxpPanels` defaults to `false`.** It is the one layer a restart is needed to undo, so it is the one the user turns on. The kernelbase `CreateFileW` / `CreateFile2` hooks now go in only when that switch is on as the mod loads — that layer needs a restart anyway, since a panel reads its stylesheet once.

**3. Module ranges follow unloads.** `ModuleRange::end` is a `std::atomic<uintptr_t>`, zero meaning gone; `begin` is written once and never moves, so a lock-free reader can never pair one module's begin with another's end. `OnDllNotification` handles `LDR_DLL_NOTIFICATION_REASON_UNLOADED` and clears `end`; `AddModuleRange` revives an entry mapped at the same base. A `g_moduleRangeEpoch`, starting at 1, invalidates the thread-local last-hit cache — and the RTTI node-class table with it, since a vtable address belongs to the module it came from.

### Optional

Taken: one settings snapshot through `ConvertForPaint`, `RefreshSlot` and the GDI layer; `LdrUnregisterDllNotification`'s status logged; `PaintMenuBarItem` returns true once the fill is down and logs a failed `DrawThemeTextEx` once; `RecolorCssHex` takes its case from the first letter across all the digits, the alpha pair included.

Left alone: the `before2026` double `GetProcAddress`, as the review itself suggested.

### New since that review

**The band around the picture in the monitors.** It is drawn by `DisplaySurface.dll` on D3D12 and goes through none of the layers above. The mod recognizes those calls by the module the return address falls in — tracked from the loader notification, so no offset into Adobe's code is involved — and by the color the draw carries in root parameter 1: opaque, dark, neutral, and not black. That color is replaced on its way to the shader.

The clear behind the picture is deliberately untouched. It is the black a clip with an alpha channel composites onto, and the black the monitor shows over a gap in the timeline; coloring it made the preview vanish in both cases. Only the quads change. The four D3D12 hooks sit on the command list vtable, which the whole process shares, so each one turns away anything outside DisplaySurface with two atomic loads before it touches a map or a setting. The layer answers to **Direct fills** with the rest.

**A fifteenth palette, Miku**, and a tenth custom-theme field, **Monitor background**, for that band. Miku's accent was darkened from `#39D5D8` to `#0E7478` — 1.65:1 to 5.10:1 against its own text. A new test holds every built-in to what the readme promises: a rising ramp, text at 7:1 on the panel, accent at 4.5:1 on the text, and white at 4.5:1 on the shade its highlight derives.

**A pass over the whole file** turned up one real defect and some smaller ones, all fixed:

- `RememberThemedWindow` grew a `std::unordered_map` inside an exclusive SRW lock, from inside `CreateWindowEx`. Short of memory the `bad_alloc` would unwind into user32 and the lock would never be released. Guarded, as are the map copy in `RecolorThemedFrames` and the module enumeration in `SnapshotAdobeModules` — so no allocation in the mod can now throw out of a hook or out of a Windhawk entry point.
- `OpenThemedStylesheet` reads and writes through `CreateFileW`'s trampoline, but `CreateFile2` is hooked separately; a kernelbase exporting one and not the other would have been a null call.
- A `static_assert` holds `kInterfaceBlues` to the 64 bits that record which of them were recolored.

The file is 6,081 lines. For scale, the mods installed here run to 20,769 (`windows-11-taskbar-styler`), 16,200 (`windows-11-start-menu-styler`) and 8,327 (`file-operation-styler`) — happy to trim further if that is still too much.

Built with `-Wall -Wextra` and no warnings; the logic tests in `tests/` pass at `-O2` and `-O0`.

CakeDev4k:
Thanks — finding 1 was real, and it was the important one. I reproduced it before touching anything.

### 1. The neutrality test rejected every tinted palette

Confirmed, and worse than the report guessed at: **ten of the fifteen palettes**, not just the tinted ones I had measured. Running `#1D1D1D` through the mod's own conversion for each palette and checking `(hi - lo) <= 0.0015`:

| | palette | `#1D1D1D` becomes | `hi - lo` | |
|---|---|---|---|---|
| ✓ | onyx, abyss, graphite, contrast, threshold | `#0C0C0C` … `#0F0F0F` | `0.0000` | passes |
| ✗ | premiere | `#10102C` | `0.1078` | rejected |
| ✗ | comfy | `#1E1713` | `0.0412` | rejected |
| ✗ | neon | `#150D14` | `0.0328` | rejected |
| ✗ | glitch | `#101907` | `0.0720` | rejected |
| ✗ | violet | `#180A33` | `0.1608` | rejected |
| ✗ | blossom | `#1F111B` | `0.0578` | rejected |
| ✗ | ember | `#1B0F05` | `0.0848` | rejected |
| ✗ | amethyst | `#0D0D13` | `0.0230` | rejected |
| ✗ | crimson | `#1C080C` | `0.0760` | rejected |
| ✗ | **miku** | `#0E2128` | `0.1034` | rejected |

So exactly the set the comment recorded as measured, and Miku — the one palette that ships a `monitor` color — was documented as having a band of its own while doing nothing at all.

`IsMonitorBandColor` now takes the settings and accepts either shape, both still dark, opaque and not black:

- **neutral**, which is the color arriving untouched — the case where "Premiere interface" is off, or where a build fetches the gray somewhere the mod does not see;
- **a tone this mod's own conversion produces** from a dark neutral gray. What a gray becomes gets brighter as the gray does — the ramp rises and the blend is with the gray itself — so bisecting finds the gray that would have produced the observed brightness, and the three channels then say whether it really did. Tolerance is 2.5/255, a round trip through eight bits.

Accepting both is what makes this independent of the plumbing question, which I could not settle from the code alone: whichever way the color reaches DisplaySurface, it is recognized.

`TestMonitorBandColor` now holds **all fifteen palettes** to recognizing their own conversion of the stock gray, to still recognizing the untouched gray, and to still rejecting black and a saturated color of the same brightness. Reverting to the neutral-only rule fails on exactly ten of them plus the explicit Miku case — eleven assertions.

### 2. Only the exact float4

Taken as suggested. The guard is now `rootParameterIndex == 1 && destOffsetIn32BitValues == 0 && num32BitValuesToSet == 4`, so a larger block bound at the same parameter is left alone rather than having its first four words rewritten.

`SetGraphicsRoot32BitConstant` is no longer hooked at all. It can only ever carry one word, so under the rule above it could never complete a color — and assembling one out of four separate calls is the unsound path you describe. That also removes a hook from the vtable every D3D12 caller in the process shares, and with it the partial-re-set drift from the functionality notes.

`TestMonitorConstantShape` pins the accepted shape against a larger block, a non-zero offset, a single word and another root parameter.

### 3. Reset

Hooked, at slot 10 as you say, and it drops the list's slot. `TestMonitorReset` checks that a full viewport and scissor recorded before a Reset do not satisfy `IsFullMonitorState` after it, and that the same address reused comes back with nothing carried over.

### Optional items

**Its own switch** — taken, and I think it was the right call to raise. **Monitor band** is now a setting of its own, on by default. The readme says to turn it off first if the monitors misbehave, and why: it is the only layer that hooks Direct3D, on entry points shared with every D3D12 program in the process.

**Fixed-size state** — taken. Eight `{ID3D12GraphicsCommandList*, MonitorCommandState}` slots in a `thread_local` array, trivially destructible. No `new`, no `try`/`catch` in a render hook, nothing left behind on unload. A list that is Reset gives its slot back; past eight, the oldest goes in turn.

**`MenuBarTheme`'s destructor** — taken, one line, mirroring `CloseThemeData_Hook`.

**`NoteDisplaySurface`** — taken. Both readers now share a `ModuleImageRange` helper.

**The Discord banner** — kept, deliberately. It is the project's support channel rather than a storefront, and the section is about reporting bugs and suggesting palettes; happy to drop the image if you would rather the readme carried no artwork but its screenshots.

### Functionality notes

**The 400x300 floor** — left as it is, and now documented as a limitation instead. A relative test would need the window size, which the command list does not carry, and lowering the floor untested on a render path is the kind of change that trades a cosmetic gap for a real regression. With the color test now much more specific, I would rather revisit this with a measurement than a guess.

**The ceiling** — the readme now says outright that this layer answers to **Strength** but not to the brightness ceiling, and takes **Monitor background** directly rather than a stop on the ramp.

**Blast radius** — noted, and the three changes above are all narrowing: one fewer hook on the shared vtable, an exact-shape guard instead of a permissive one, and per-recording state. The return-address filter still runs before anything else.

### One more pass over the layer, after the fixes

Re-reading the whole file against this round's findings turned up three more, all in the same layer and all now fixed:

- **A slot outlived its recording.** Dropping it on Reset was half the job: Close ends a recording too, and a list released without a Reset leaves its slot behind for whatever is created at that address next. That owner would then be recording against a full-surface viewport it never set, with its own dark gray — a common enough color — eligible to be taken for the band. Close is hooked as well, so a slot lives exactly from the start of a recording to its end. `TestMonitorReset` drives both ends.
- **A partial install could leave the recolor running on state nothing invalidates.** The five hooks are registered together but fail one at a time, and Reset and Close are precisely the ones whose absence is silent. `MonitorBandHooksReady()` requires the whole set, and `RecolorBandConstants` now checks the layer's switch itself instead of trusting its only caller to have done it. Reverting either guard fails the tests.
- **`D3D12CreateDevice` was hooked whatever the settings said.** With **Monitor band** off, the mod now does not touch d3d12 at all — not even to watch for the device, which is a better promise to be able to make given the blast radius. Turning the switch on later is still covered, by the loader hook and by the probe in `Wh_ModSettingsChanged`.

The readme says why the ramp has to stay in order, which matters more now that the band is recognized by working back along it, and the safe-mode paragraph mentions the new switch.

The file is 6,268 lines.

Built with `-Wall -Wextra` and no warnings; the tests in `tests/` pass at `-O2` and `-O0`.

CakeDev4k:
Thanks. The required item is real and it was mine — gating the D3D12 export on the new switch last round is exactly what opened that hole, and I missed that `Wh_ModSettingsChanged` was then the only path with no way back in.

### Required: turning the layer on has to reach both ways in

Fixed as suggested, and the reasoning holds in full: with the switch off at load the export is not hooked, the probe declines until `DisplaySurface.dll` is mapped, and nothing else comes back to either. A user flipping the switch during startup would have kept Premiere's gray for the session with nothing in the log.

```cpp
bool registered = HookLoadedModules();

registered = HookD3D12CreateDevice() || registered;
registered = InstallMonitorBandFromProbe() || registered;
```

`TestSettingsChangedRetriesBothWaysIn` reads the function out of the mod's own source and checks all three retries are there; removing the line fails it.

### Optional items taken

**A failed probe latched the layer off.** Both entry points now release the latch when the install does not happen — it stands for the attempt, not the failure, so a driver reset or a session with no device to give no longer locks the other way in out of trying.

**`g_monitorBandInstalled` on a partial install.** `InstallMonitorBandHooks` returns false and leaves the flag clear when any hook failed, so the unload line tells a band nobody recognized from a layer that never got the chance. Those are now two diagnoses with two lines.

**The frame work on every non-child window.** Taken, and you were right to raise it twice — I had the class of window wrong the first time. The DWM calls and the map insert are gated on `WS_CAPTION | WS_THICKFRAME`, so a menu popup, a tooltip or a combo dropdown costs nothing where it used to cost four round trips to DWM and an insert under the themed-window lock, inside `CreateWindowEx`. `AllowDarkModeForWindow` still goes to every window, since it sends nothing and is only a permission. A window that gains a caption later is picked up by `ApplyThemeToExistingWindows`, which a settings change runs. `TestFramedWindowsOnly` covers popup, child, message-only, captioned and sizing-bordered.

**The two openers without the retry.** All four now go through one `OpenDarkMenuTheme`, templated on the opener, so `OpenThemeDataForDpi` and `OpenThemeDataEx` get the `hwnd` to `nullptr` fallback too.

**The stylesheet handle.** The copy is opened with the caller's own share mode now, and `CreateFile2`'s `dwFileAttributes` is folded in with `dwFileFlags` instead of being dropped. The caller's mode can legitimately fail: the handle is opened while the writer's is still open, because the file is delete-on-close and would go the moment the writer let it go, so a mode that excludes the writer's `GENERIC_WRITE | DELETE` gets a sharing violation. A mode wide enough to admit the writer is the fallback rather than the default, which is the best this shape allows.

### Functionality notes taken

**Record order.** Taken — this was worth fixing. All three hooks that record a piece of the decision now try the recolor, so the band no longer depends on DisplaySurface setting the surface before the color. The two that are too early are turned away by `IsFullMonitorState` or by `hasRoot1Color`. `TestMonitorRecolorOrder` drives the color-first order, and a check alongside it reads the three hook bodies and fails if any of them stops calling it.

On the per-thread scope: right, and the comment now says so. A command list is recorded, closed and reset on one thread because D3D12 requires it; a list closed on another would leave one stale slot until the ring came round to it, which costs no wrong color, since a slot is only ever reached again through that same list's address.

### Left as they are

**`IsAdobeUICaller`'s linear scan.** Measured rather than assumed: the table holds about 60 live entries in Premiere 2026, and the scan is only reached after `ShouldConvertWith` has already turned the call away — so it runs for dark neutrals, as you say, but 60 range compares over roughly 1 KB of L1 is a fraction of the `CreateSolidBrush` it precedes. Sorting it would mean either shifting entries under lock-free readers or a second structure to keep in step, both of which cost more correctness than the scan costs time. Happy to revisit with a profile.

**"Monitor band" defaulting to false.** Kept on, deliberately. The band is the most visible thing the mod did not reach before this layer, and off by default means almost nobody sees it — including Miku's own band color, which exists for it. The per-call cost is a trampoline plus four integer compares that turn away every root parameter but 1, before any settings read; `Reset` and `Close` are once per list per frame. I would rather ship it on and hear that it costs something real than ship it off and have it do nothing. Say the word if you would rather it matched UXP panels and I will flip it.

**`IsPaletteTone` blending from the average.** Correct, and it is exact for a neutral source, which the band's is: `#1D1D1D` in every build measured. The divergence needs a source gray that is *not* neutral but inside `IsNeutral`'s tolerance, plus a strength below 100. Widening `kBandTone` to cover it would loosen the test in the direction that matters — a false positive recolors something that is not the band, a false negative just leaves Premiere's gray — so the tight tolerance is the safer error, and I would rather keep the safe failure than trade it for a wider match.

**The probe's default adapter.** Fair, and now noted in the code. The probe only runs when the mod is enabled while Premiere is already running, and the device is released immediately, but the UMD it pulls in does stay. There is no way to reach a command list vtable without a device, and picking Premiere's adapter would mean enumerating DXGI and guessing which one is theirs.

**The carried-over stylesheet items.** `hsl()`/`hsla()` and the space-separated `rgb()` form are not in Adobe's bundled sheets as they ship today; I would rather add them against a real sheet than speculatively. `PaintMenuBarBottomLine`'s one pixel is worth a look at 200% and I have not been able to test it.

### Also in this push

Miku's band goes from `#04090C` to `#020F13`. The old value was a correct darker step of the ramp but carried almost none of its hue — 8 levels between its channels, where the palette's own deepest step has 14 — so it read as flat near-black rather than as part of the theme. The new one is darker still and holds more teal than that step does, at the ramp's own channel relationship.

The file is 6,446 lines. Built with `-Wall -Wextra` and no warnings; the tests in `tests/` pass at `-O2` and `-O0`.

### One more, found after the review: the search field in the UXP panels

Reported as the search bar in the Text panel looking wrong under some palettes, and it turned out to be every input in every UXP panel.

Spectrum fills a text field `#080808`, well under the ceiling, and outlines it `#494949` — **28.6%, six tenths of a point above the default 28**. The fill was themed and the outline was not, which on a search field is the whole of the control: a stock gray box on a themed panel. It shows under some palettes and not others only because the mismatch is invisible when the theme has not moved far from stock.

Raising the ceiling cannot fix it. Spectrum's disabled text is `#4B4B4B`, 29.4% — eight tenths of a point from the border it must not be confused with — and the readme already warns that raising the ceiling erases that text. Brightness has no way to tell a border from disabled text.

A stylesheet does, and that is information the runtime path never has. `border-color`, `background-color` and `outline-color` are chrome whatever their brightness; `color`, `-webkit-text-fill-color` and `fill` are text and icons. So `RecolorStylesheet` reads the property in front of each value and allows chrome seven points past the ceiling, leaving everything else exactly where it was.

Measured against Premiere 2026's own Text panel stylesheet: 61 more colors reached, among them **15 uses of `#4B4B4B` as a border or a fill — while the 25 uses of the same value as text stay untouched**. The slack stops short of `#696969`, the next neutral Spectrum uses for chrome, so a light divider or a focus ring is still left alone. Nothing already recolored moves, because past the ceiling `PickTarget` clamps to the palette's lightest step, which is the tone it keeps for edges. `TestStylesheetChromeCeiling` pins all of it, and reverting the rule fails five assertions.

### And one from the log

`monitor band: more than 8 command lists recorded on one thread` fired in a real session — on a build that already frees a slot at both `Reset` and `Close`. The comment said "one or two of these are ever in use in practice"; that was a guess and it was wrong, so the cap is 32. The line stays, to say if even that is not enough.

CakeDev4k:
Thanks — the required item is real, and it is the hazard I talked myself into last round. Making the recolor order-independent is what put a replay on a path that does not know which root signature is bound, and I wrote the replay as if `hasRoot1Color` meant "a color is known" when it actually means "a color is known *under some signature*".

### Required: a replayed color has to stay under its own root signature

Taken the first way you offered — `SetGraphicsRootSignature` is hooked, at slot 30 of the same table, and the color goes when the signature does.

```cpp
void STDMETHODCALLTYPE
MonitorSetGraphicsRootSignature_Hook(ID3D12GraphicsCommandList* commandList,
                                     ID3D12RootSignature* rootSignature) {
    MonitorSetGraphicsRootSignature_Original(commandList, rootSignature);

    if (MonitorCommandState* state = KnownMonitorState(commandList)) {
        state->hasRoot1Color = false;
    }
}
```

That keeps the order-independence round 8 asked for and closes the sequence you wrote out: with signature B bound, `hasRoot1Color` is already false when the scissor completes the surface, so `RecolorBandConstants` turns away instead of writing four words at parameter 1 of a signature that may hold a descriptor table there.

The viewport and the scissor stay. They are rasterizer state, not root state, and they survive a signature change — dropping them would only cost the band a frame for no reason.

One thing I did not try to be clever about: D3D12 leaves bindings alone when the *same* signature is set again, and this does not tell the two apart. Dropping a color that was still good costs one frame of Premiere's own gray. Keeping one that was not costs the device. The comment says so, so nobody tightens it later thinking it is an oversight.

`TestMonitorRootSignature` drives the exact order — color, signature, viewport, scissor — and asserts no write happens; reverting the hook body fails four of its assertions.

### The same hazard under another name: `ClearState`

Looking for anything else that invalidates what the replay reads, `ClearState` (slot 11) does all of it at once: it puts a direct list back to how it was created, which unbinds the root signature *and* empties the viewports and scissor rectangles. A replay after one would write root constants with no signature bound at all.

It drops the recording whole, which is the state `Reset` already leaves behind:

```cpp
void STDMETHODCALLTYPE
MonitorClearState_Hook(ID3D12GraphicsCommandList* commandList,
                       ID3D12PipelineState* pipelineState) {
    MonitorClearState_Original(commandList, pipelineState);

    ForgetMonitorState(commandList);
}
```

Both are in `MonitorBandHooksReady`, so a partial install still stops the layer acting at all rather than leaving it reading state nothing invalidates. And since the slot numbers are the one thing here not looked up by name, `TestMonitorVtableSlots` now reads all six out of the source and checks each against the method named beside it — a wrong number would hook a different method of the same interface, which is exactly the mistake that would look like nothing at all.

### What those two cost everyone else

`Reset`, `Close`, `ClearState` and `SetGraphicsRootSignature` sit on the command list, so they run for every D3D12 caller in the process — and unlike the other three they cannot ask the return address first, since a list this layer follows can have its bindings undone from somewhere else. So each was sweeping 32 slots, about two kilobytes, to learn there was nothing to find.

The table now carries a count of live slots, and a thread that records none of DisplaySurface's work stops before touching it. That is a strict improvement on what was already shipping: `Reset` and `Close` were doing the full sweep on every command list in the process. `TestMonitorReset` pins the count at zero, at one, at the cap and back to zero; removing the decrement fails four assertions.

### Also in this push, which my last reply did not describe

Two commits went in with that push and I only wrote up the first half of what they do.

**The design tokens a UXP panel keeps in its own script.** The chrome slack I described fixed the stylesheet, and the Text panel in Graphics came out right — but the other tabs did not, and I had the cause wrong. Those panels do not get their input colors from `main.css` at all: the Spectrum tokens are a table in `main.js`, applied as inline styles, which beats any stylesheet rule. `RecolorTokenTable` rewrites `"-color":"rgb(…)"` in the script the same way the stylesheet path does, with the key read back to tell a `text-color` from chrome. One byte turns away almost every position of a 1.7 MB script before the compare.

It costs a panel one extra read of its own script at load. Measured on Premiere 2026: 118 files and 38.4 MB under `UXP\plugins`, of which 9 files and 14.9 MB actually carry tokens — the other 109 are read and discarded. Size is no filter, since the largest one carrying tokens is 7.5 MB. That is why it rides on the existing **UXP panels** switch instead of getting one of its own, and the readme says so.

**A substituted script under a stylesheet's name.** `WriteTemporaryCopy` named every copy `.css`, so a substituted `main.js` reached the UXP runtime as a stylesheet. It works today, but the copy exists to stand in for the original, and a runtime that looked at the extension would be right to refuse it. The copy now carries the extension of what it replaces.

### On the two from last round

**"Monitor band" defaulting to false** is still on, for the reasons I gave, and I will flip it if you would rather it matched UXP panels.

**The Discord link.** Still in the readme with the server's banner under it. Happy to drop the image and keep the line if the banner reads as promotional.

### One I found afterwards and am not fixing

Going back over every table this mod keys by a handle — the class of thing four of your required items have been — the themed-window map has the same gap and I left it. A destroyed `HWND` is dropped when the map doubles and re-checked at the revert, but a handle reused between two prunes carries the old window's bits into the new one, and the revert then undoes on it something the mod set on its predecessor.

I left it because the cost is bounded and paid once: immersive dark mode back to `FALSE` and the caption colors back to the DWM default, which is where a Premiere window the mod did not touch already sits, or a theme class cleared on a window that had one. Telling the two apart needs a destroy notification this mod deliberately does not hook, to avoid setting a frame to what it already is. The comment now says all of that, so the next reader does not have to work out whether it was considered.

The file is 6,678 lines. Built with `-Wall -Wextra` and no warnings; the tests in `tests/` pass at `-O2` and `-O0`.

CakeDev4k:
Both taken. The first is a real crash I had built a path to and not seen, and the second is the answer to a question I asked, so it gets taken as given.

### Required 1: a partial install must not be tried again

You are right, and the part I missed is that the comment and the code were describing different things. The comment reasons about the device — "a device this one could not be built from must not lock the other entry point out of trying its own" — and I wrote the `else` that clears the latch for *every* failure, including the one where six of the seven are already registered with Windhawk. `Wh_ModSettingsChanged` running the probe on every settings change then makes the second registration a matter of one checkbox, and a second hook over this mod's own trampoline is a hook whose original is the first hook's entry. The first `RSSetViewports` from DisplaySurface would recurse until the stack is gone.

The two failures are told apart now, exactly as you laid out:

```cpp
enum class MonitorBandInstall { Ok, NoDevice, HooksFailed };
```

`NoDevice` is every way of getting nowhere without registering anything — no allocator, no command list, no device from the probe, and now a vtable missing a slot. It releases `g_monitorBandTried`, which is the case the comment always meant. `HooksFailed` keeps the latch and says so in the log:

```
monitor band: one or more D3D12 hooks failed — something else in this process
may hold these entry points. The band keeps Premiere's gray for the session;
this is not tried again
```

Worth saying out loud rather than just logging "failed": as you note, these are the entry points a GPU overlay or a capture tool takes first, so the line names the likely cause instead of leaving a user to guess.

I took the second half of your suggestion too — all seven vtable entries are resolved and checked before the first `SetFunctionHook` — so "failed partway" now means only the hook engine, and a vtable that is not the shape this expects stays `NoDevice`, with the other way in still free to try a device of its own.

What I did **not** do is try to unregister the hooks that took. They stay registered and go live with the next `Wh_ApplyHookOperations`, where they are pass-throughs: `MonitorBandHooksReady()` is false, so nothing is recolored and nothing is replayed. That costs a trampoline on some of the seven for the session, which is the smaller risk next to unwinding a partial registration from inside a failure path.

`TestMonitorPartialInstallLatch` reads both ways in out of the source and fails if any release of the latch is reachable without the install having said `NoDevice`; putting the old `else` back fails it.

### Required 2: the band ships off

Flipped, and thank you for actually answering it — I asked and you gave a straight call, so it goes in as given rather than argued about.

For the record I think the case you make is the stronger one. It is seven shared entry points now rather than five, and the ones I added are the busiest of them: `SetGraphicsRootSignature` per draw batch, `ClearState` and `Reset` and `Close` per list per frame. It is the only layer that writes into a command list it does not own, the failure mode is a removed device, and a layer whose own documentation opens with "turn it off first if the monitors misbehave" is not one to run before anybody asked. My argument for shipping it on was that nobody would see it — but `HookD3D12CreateDevice` and the probe already make the checkbox work mid-session with no restart, so nobody has to.

The settings description now opens with why it is off, the mod's readme says the same in the layer list and in its own section, and `TestShippedDefaults` pins it alongside `uxpPanels`, with the reason in the comment.

### Optional items

**`HookD3D12CreateDevice` latching before the hook succeeds.** Lost, not declined — thank you for raising it again. Fixed, and you are right that it is the opposite correction: one export, registered or not, so the latch goes back and both ways in stay free to try. The comment at the release now names the difference, so the next reader does not "fix" one into the other.

**The README's Highlight row.** Taken. Violet's row reads `#7737DC`, which is `#6D28D9` at the brightness of Premiere's blue, and the Custom themes section now says so before the table of fields: that row is the shade the hue produces, every other row is the value the field takes.

**`ForgetThemedClass` taking the exclusive lock on every `SetWindowTheme`.** Taken, in `ForgetMenuTheme`'s shape — shared lookup, exclusive only for a window actually recorded, and looked up again under the exclusive lock since the shared one was let go.

**The `Wh_ModUninit` comment.** Taken, and you are right that it was overstating. `g_monitorStates` and `g_monitorStatesLive` are `thread_local` PODs inside the image and go with it; the color table and `g_sysBrushes` are the two that genuinely outlive the unload, and the comment now says why each does.

**The stylesheet-era names.** `g_stylesheetSerial`, `g_stylesheetFailureLogged` and `g_stylesheetsRecolored` are `g_bundledSerial`, `g_bundledFailureLogged` and `g_bundledRecolored`.

**The Discord banner.** Dropped from the mod's readme, link kept. You are right that it was the one thing there reading as promotion.

### The one I am leaving, with the measurement you asked for

**The CSS chrome slack and `var()`.** I grepped Premiere 2026's own stylesheets, with the mod's own two tests rather than eyeballed — neutral within 0.035, brightness as the channel average — and counted what sits in the slack band, `ceiling < gray <= ceiling + 0.07`:

| | in the slack band |
|---|---|
| literals in a declaration | **793** |
| bound to a `--custom` property | **9** |

The nine are three tokens: `--spectrum-global-color-gray-400` `#494949`, `gray-300` `#4a4a4a`, `gray-800` `#4b4b4b`, each defined three times. So the case is real but small — and then I looked at how they are consumed:

| token | `var()` uses | of those, as `color` |
|---|---|---|
| `--spectrum-global-color-gray-400` | 22 | **14** |
| `--spectrum-global-color-gray-800` | 17 | **15** |
| `--spectrum-global-color-gray-300` | 56 | 0 |

Treating a `--`-prefixed name as chrome would push **29 uses of text** past the ceiling — which is the disabled-text erasure the ceiling exists to prevent, and the exact failure the chrome rule was built to avoid. The rule works because the property name at the point of use tells a border from text; a token *definition* does not carry that, so covering it would be a guess, and for two of the three tokens the guess is wrong more often than right.

### Functionality note taken: `ExecuteBundle`

This one I did act on, because it is the same class as round 12's required item and leaving a known staleness hole because it is "several coincidences deep" is the reasoning that produced rounds 14 and 16 in the first place.

Slot 27, and it drops the color rather than the recording:

```cpp
void STDMETHODCALLTYPE
MonitorExecuteBundle_Hook(ID3D12GraphicsCommandList* commandList,
                          ID3D12GraphicsCommandList* bundle) {
    MonitorExecuteBundle_Original(commandList, bundle);

    if (MonitorCommandState* state = KnownMonitorState(commandList)) {
        state->hasRoot1Color = false;
    }
}
```

Narrower than `ClearState`'s answer on purpose: `RSSetViewports` and `RSSetScissorRects` are among the calls D3D12 does not allow inside a bundle, so the surface this layer recognized cannot have changed and there is no reason to throw the recording away. `TestMonitorRootSignature` covers the bundle path alongside the signature one, and removing the body fails three assertions.

I did weigh it against the argument for defaulting the layer off — this makes it eight shared entry points rather than seven. It went in because `ExecuteBundle` is per bundle execution rather than per draw, the body is a state drop with the live-slot early-out in front of it, and the alternative is a known way for this layer to write root constants under a signature that never received them.

### The carried-over notes

Still as they were, and still deliberate: `MonitorStateForColor`'s fallthrough costs no wrong color, since a slot is only reached again through the same list address; `RecolorTokenTable`'s exact key is what keeps a 1.7 MB script cheap to scan, and `rgba(` and hex token values are not in the sheets as they ship; the double read on a miss is measured in the readme; `PaintMenuBarBottomLine`'s pixel I still have not been able to check at 200%.

On the catalog overlap with `dark-menus`: thank you for confirming there is nothing else in Premiere's or Adobe's direction. The readme keeps its note about resolving the menu conflict.

The file is 6,793 lines. Built with `-Wall -Wextra` and no warnings; the tests in `tests/` pass at `-O2` and `-O0`.

CakeDev4k:
Thank you for reading the whole file rather than the diff — that is the pass that tells me the lifecycle work from the early rounds actually holds, and it is worth more than another round on the newest code.

Five of the six optional items are in. One is left, and the functionality notes are answered below.

### Taken

**`HookLoadedModules()` in `Wh_ModAfterInit`.** Fixed, and the reference was the useful part: a module mapped as a static dependency never passes through `LoadLibraryExW` at all, which I had not thought about — my reasoning had been "the loader hook covers everything after init", and that is only true for explicit loads. The retry covers the d3d12 export in the same call, since it has the same gap.

**The bare hex run.** Taken, and you are right that relying on Adobe's naming was the weak part. `CssValueIsChrome` is now a three-state `CssValueRoleAt`, and the hex branch bails on `NotAValue`.

I measured before changing it, because the walk back stops at `;` and a `data:` URI has those inside it. Of the **12,873** hex colors in Premiere 2026's bundled stylesheets, the guard skips **zero** — so it removes the class of problem without changing a single recolor today.

It is the hex form only. `rgb(` can never be a selector, and the `-rgb` form sits on a declaration's own colon rather than a position inside a value — bailing there would stop rewriting the `--…-rgb: 29,29,29` custom properties entirely. The comment in the loop says which is which, so nobody extends the guard to all three later.

**The 1px menu-bar line.** Taken, and measured rather than scaled. `GetMenuBarInfo` has already given the bar's own rectangle by that point, so the line covers the gap between it and the client area:

```cpp
RECT bar = barInfo.rcBar;
OffsetRect(&bar, -window.left, -window.top);

if (bar.bottom < client.top && client.top - bar.bottom <= kMaxEdge) {
    line.top = bar.bottom;
}
```

That is right at every scale without a second system metric to agree with, and it is what I would rather ship than `dpi / 96` — a computed height that overshoots paints the palette's edge tone over the bottom row of the menu bar, which is a light hairline in the other direction. A bar rectangle that overlaps the client area or sits further off than any border falls back to the single row this always painted. I still have no scaled display to confirm it on, so that fallback is deliberate.

**The band's hooks staying installed.** You are right that the readme implied otherwise, and the distinction matters for exactly the person the sentence is for. It now says the switch stops the recoloring from the next frame but does not take the entry points back out, and that chasing a conflict with an overlay or a capture tool means turning the switch off and restarting, or disabling the mod.

**Readme length.** The design-token/`main.js` walkthrough and the quad explanation are out, with a link to the repository readme, where they now live in full with the measurements. In fairness the net is only nine lines: the two sections you named came out, and the band-hooks paragraph above plus a note about the Highlight column went in, both from this round and the last. If the remaining length is still an obstacle, say which sections and I will move them — I kept the palettes, the screenshots, the custom-theme fields and "If something becomes unreadable" on the grounds that those are what a user needs on the page itself.

### Left as it is

**The frame revert undoing another mod's dark title bar.** Real, and I am leaving it, for the reason you half-stated yourself: the theme-class path has `SetWindowTheme_Hook` to hang `ForgetThemedClass` on, and there is no equivalent notification for `DWMWA_USE_IMMERSIVE_DARK_MODE`. Covering it means hooking `DwmSetWindowAttribute` process-wide to watch for one attribute on one window class, which is more surface than the asymmetry is worth — and stock Premiere's title bar is light either way, so the case needs a second mod that both sets it and is disabled after this one. Noted in passing rather than silently.

### Functionality notes

**The double read of every `.css`/`.js`.** A per-session "this file had no colors" set is the right shape if it ever matters, and I would rather not add it before it does: it is another table keyed by something, which is the category that has cost this pull request the most rounds, and it would be caching a negative across a Premiere update that changes the file. The readme owns the cost, and the layer is off by default.

**`kContentClasses` and the two RTTI names.** Agreed that it is a list that will need extending, and that the failure mode is quiet. A log line per unknown class would be noise, as you say — but a one-off diagnostic pass on a new build is exactly the right way to do it, and I will run one when 2027 ships rather than guessing at names now.

**`IsPaletteTone`'s 24-step bisection.** Right that it is the one place where a per-call cost is more than a compare. It sits behind `IsFullMonitorState` and the opaque/dark/non-black tests, so it is a handful of calls per frame on the surfaces that actually look like the band — and if the heuristic ever starts matching something that issues many such quads, that is a correctness problem before it is a cost one, and the log line for an unrecognized band is what would show it.

**The menu-bar item with no text, and `dva*`.** Both accepted as stated. The `DefWindowProc`/`DefFrameProc` hooks being process-wide is the part worth remembering, and a third-party plugin hosting an MDI frame inside Premiere is the case that would hit it. `IsAdobeUIName` matching any `dva*` is deliberate — Adobe's own toolkit is the whole prefix — and only the GDI layer consults it.

**The overlap check.** Thank you for confirming there is nothing else in Premiere's or Adobe's direction in the catalog; the readme keeps its note about resolving the `dark-menus` overlap.

The file is 6,855 lines. Built with `-Wall -Wextra` and no warnings; the tests in `tests/` pass at `-O2` and `-O0`. Every behavioural change above is covered by a test that fails when the change is reverted.

PR #5480, opened 2026-09-12
Tourne'Table [Audio Visualizer] 1.0.0  Initial release. Real-time des…

USER-TOURNE:
desktop audio visualizer with 8 shapes, 9 color modes, configurable FFT size and frequency scaling, Now Playing display, and substantial CPU/thermal optimization.

## Changelog

* Changelog item 1...
* Changelog item 2...

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/tourne-table-desktop-audio-visualizer.wh.cpp
+++ b/mods/tourne-table-desktop-audio-visualizer.wh.cpp
@@ -1,0 +1,4030 @@
+// ==WindhawkMod==
+// @id                  tourne-table-desktop-audio-visualizer
+// @name                Tourne'Table [Audio Visualizer]
+// @description         A real-time audio visualizer for the Windows desktop. Advanced settings without sacrificing resource efficiency. Near-headless rendering with CPU optimization for audio capture.
+// @description:ru-RU   Аудиовизуализатор реального времени для рабочего стола Windows. Расширенные настройки без ущерба для экономии ресурсов. Практически безинтерфейсный (near-headless) поток рендеринга с оптимизацией процессора для захвата звука.
+// @version             1.0.0
+// @author              USER-TOURNE
+// @github              https://github.com/USER-TOURNE
+// @donateUrl           https://ko-fi.com/tourne
+// @license             MIT
+// @include             explorer.exe
+// @architecture        x86-64
+// @compilerOptions     -ldxgi -ld2d1 -ld3d11 -ldcomp -ldwmapi -ldwrite -lgdi32 -lshcore -lshlwapi -lole32 -lshell32 -lksuser -lwindowscodecs -lruntimeobject -lwindowsapp -luuid -luser32 -ladvapi32
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# `Tourne'Table` **[Audio Visualizer]**
+
+![Tourne'Table Audio Visualizer](https://raw.githubusercontent.com/USER-TOURNE/Windhawk-Mods/main/GIF/11.gif)
+
+*Oscilloscope shape running live audio - bottom-left placement, blurred panel, single-pixel border.*
+
+> **A real-time audio visualizer that lives on your Windows desktop.**
+> Built on the foundation of Salyts' Desktop Audio Visualizer, rebuilt around performance.
+
+Play music. Bars dance on your wallpaper. That's the whole idea.
+
+It listens to **whatever your PC is already playing** - Spotify, YouTube, a game, a call - and draws it behind your desktop icons. No virtual audio cable, no drivers, nothing to configure. It just picks up your system audio.
+
+The original worked, but it ran *hot*. This project rebuilt how it draws itself: **same job, roughly half the CPU, 18 °C cooler**, plus a pile of new shapes, colors and controls.
+
+## ABOUT THIS PROJECT
+
+Built with an emphasis on lower resource consumption, more efficient rendering, better frame pacing, expanded visualization options, dynamic album-art integration, native Windows media info, deeper customization, and power-conscious idle behavior.
+
+The goal is simple:
+
+> **Make the desktop move with the music - without making the CPU move mountains to do it.**
+
+---
+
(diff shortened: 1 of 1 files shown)


USER-TOURNE:
/ai-review The architecture, function decomposition, capture/render split, and bootstrap are all Salyts' work. I started from his file and worked outward from there.

**What I changed, concretely:**

**Settings:** 28 -> 46. The 18 new settings are FFT size, frequency scale (log/linear/mel), peak-frequency readout plus its two alignment axes, multiband oscilloscope coloring, peak-hold caps, beat flash and intensity, rainbow speed, the five Now Playing settings, auto-hide and delay, and pause-when-covered.

**Rendering:** 6 shapes -> 8. I added Radial and Oscilloscope. Oscilloscope reads the raw waveform rather than the band data, which required a separate sample path. Color modes went from 7 -> 9.

**New subsystems:** I added DirectWrite text rendering for Now Playing and the peak readout, neither of which existed before. I also added runtime-configurable FFT with live rebuilding of the Hann window and twiddle tables, along with a frequency-warping layer between the band data and bar positions.

**Performance work:** I replaced the `DwmFlush()` pacing with a high-resolution waitable timer. The Gaussian blur is now baked once into a cached bitmap instead of being re-convolved every frame. The swap chain is sized to the widget's bounding box rather than the entire desktop, with the composition visual offset used to position it. Background geometry and the monitor handle are cached, and frame latency is capped at 1.

On an Intel Core Ultra 265KF, I measured CPU usage going from 13.77% -> 5.95%, package temperature from 57.5C -> 39.7C, and peak package power from 93.2W -> 40.4W. Independent WPA traces put the normalized cost at 43.77 ms/sec -> 23.62 ms/sec. I have the full methodology and caveats documented in my own repo rather than on the mod page.

## On the mod page reading as a critique

That's fair, and I'll fix it regardless of the outcome here.

The benchmark comparison belongs in my repo, not on a catalog page sitting next to the mod it is comparing against. I'll cut the mod description down to what the mod actually does and what the settings mean, and move the before/after material out of the mod page.

`desktop-audio-visualizer` declares no `@license`, and the windhawk-mods README states that mods which don't specify one are submitted under MIT. My understanding was that this meant the upstream code was already MIT and that I wasn't relicensing anything. If your interpretation is that an absent `@license` means something else, tell me and I'll get explicit written agreement from Salyts before this goes any further.

I also want to correct something in my README. Describing Salyts as "a contributor toward this one" was poor wording on my part, and I can see how it implies collaboration. We have not been in contact. He is the upstream author whose code this is built on, nothing more, and I'll correct that wording.

Issues 2 through 6 are real, and I'll fix them regardless of which route this takes. Several are crashes in `explorer.exe` rather than cosmetic issues, and #2, #3, and #4 are inherited from upstream, so they're worth fixing in whichever file ultimately carries them. Thanks as well for pointing me toward `desktop-live-overlay` for the timer teardown pattern.

## Done this pass

On top of the earlier fixes for 2-6, I addressed two additional bugs and several smaller issues.

### Two real bugs

* **Peak-hold caps were always Gradient Color 2.** `g_barBrush2` was created once and never recolored, so with default Solid white bars the caps came out as `#00B4FF` and ignored Rainbow, Album, Accent, and Acrylic entirely. They are now recolored per bar to a brightened version of the bar color.

* **Animation quantization on long uptime.** `(float)GetTickCount64()` eventually exceeds the precision available from a float's 24-bit mantissa after roughly 4.7 hours. I changed this so animation time is measured from a baseline tick captured during initialization.

### Polish

I also made a number of smaller cleanup and correctness changes:

* Replaced the hand-copied Gaussian blur GUID with the real `CLSID_D2D1GaussianBlur` and named property enums from `d2d1effects.h`.
* Cached the monitor in `IsFullscreenOrGameActive`. It was previously rerunning `EnumDisplayMonitors` every second on the UI thread, which effectively undermined the caching described in the README.
* Hoisted `warpT` so Mel no longer performs two `log10f`/`powf` operations per bar per frame.
* Replaced the hardcoded `en-us` with `GetUserDefaultLocaleName`.
* Removed the `[Viz]` prefixes.
* Simplified the font check.
* Changed `"5 5"` corner radius handling so it falls back to all defaults instead of silently producing `5 5 3 3`.
* Corrected the Acrylic description to "near-transparent", since the alpha floor is 30.
* Fixed the duplicate GIF captions.

## Deliberately not done yet

There are a few larger changes I am deliberately holding off on until I can properly test them.

**The FFT/band architecture rework** is the biggest one, and I agree with the underlying criticism. Everything currently derives from `g_bands[7]`, so Bar Count and FFT Size don't actually add spectral detail. The Sensitivity coupling you pointed out in the README also follows directly from `rms = sqrt(sumSq/count)` falling approximately as `1/sqrt(N)`.

Fixing that properly means implementing per-bar bin mapping through the Log/Mel/Linear curve. That would make Frequency Scale actually do meaningful work instead of simply warping seven fixed edges. At that point, I'm no longer patching the existing analysis path. I'm redesigning the analysis core.

That also means the existing benchmark numbers would no longer be valid, so I would want to treat that as its own implementation and testing pass rather than folding it into the current diff.

**Moving rendering off the UI thread**, which was your preferred #5 fix, falls into the same category. I agree with the reasoning, but it is a significant refactor rather than an isolated fix.

The same applies to `WindhawkUtils::StringSetting` RAII across roughly 20 call sites, `std::optional` for the remaining thread pointers, atomics for cross-thread settings, bar geometry caching, splitting metadata from artwork fetching, and Oscilloscope orientation support. I agree that these are worthwhile changes, but they are mechanical-but-broad changes that I don't want to introduce blindly.

At this point, I'm going to work through the changes that don't require a comprehensive audit on my end just yet. I'd rather do them in focused batches than pile everything onto one large, unverified diff.

My preference is to implement one change, test it, see how it affects performance and behavior, and then iterate from there. That's the same approach I've already used for the performance work, and I think it's a much safer way to handle the larger architectural changes without accidentally improving one area while quietly regressing another.

Issue #1028, opened 2024-10-03
[Windows 11 Taskbar Styler] White box issue


Laskco:
![{D814E5B4-AD51-4C30-A9EF-626AF7F268D4}](https://github.com/user-attachments/assets/0307d848-0f3e-48de-a730-d344d66a8bad)

The only fix is to restart Explorer. 
Happens during first boot up, and also when you launch some full-screen applications. 

@m417z

m417z:
I've never seen this issue, and it wasn't previously reported by users. It's strange to have the start menu styler mod affect the taskbar. Does it happen with a specific theme or style, or does it also happen when no styles are configured in the mod?

It'd be best if you could find reproduction steps for this problem, so I could investigate it locally on my computer.

Laskco:
> I've never seen this issue, and it wasn't previously reported by users. It's strange to have the start menu styler mod affect the taskbar. Does it happen with a specific theme or style, or does it also happen when no styles are configured in the mod?
> 
> It'd be best if you could find reproduction steps for this problem, so I could investigate it locally on my computer.

Oops, my apology, I meant the Windows 11 Taskbar Styler.

m417z:
I updated the title. Still, the rest of my comment applies:

> Does it happen with a specific theme or style, or does it also happen when no styles are configured in the mod?
> 
> It'd be best if you could find reproduction steps for this problem, so I could investigate it locally on my computer.

Laskco:
> I updated the title. Still, the rest of my comment applies:
> 
> > Does it happen with a specific theme or style, or does it also happen when no styles are configured in the mod?
> > It'd be best if you could find reproduction steps for this problem, so I could investigate it locally on my computer.

It seems to happen with all themes. I'm also completely unable to reproduce the issue, it just happens randomly. 
I am on 24H2 26100.1882, so I'm not sure if there's some sort of play at stake due to all the changes made to Windows.

m417z:
A comment from a similar issue (https://github.com/ramensoftware/windows-11-taskbar-styling-guide/issues/148#issuecomment-2536909152):

> SOLVED. Completely forgot I was running explorer patcher and seemed to fix itself when I uninstalled explorer patcher!

Are you using ExplorerPatcher or a similar customization tool?

lucamarza:
restarting explorer seems to fix it

Meti0X7CB:
I also have this issue. I believe it's a windows bug since I don't have any customization tools installed

[...]

PR #5483, opened 2026-09-12
Windows 11 Taskbar Styler v1.10

m417z:
* Fixed running indication in several themes - a regression that was introduced in the previous update.
* Updated themes to avoid changing the taskbar width when it vertical mode. Previously, some themes could cause the taskbar to fill the whole screen, making it difficult to recover from that state.
* Fixed the long-standing [white box issue](https://github.com/ramensoftware/windhawk-mods/issues/1028).
* Added `skip()` which leaves the style unapplied, so the property keeps (or returns to) its original value, e.g. `{{width > 0 ? width : skip()}}` applies only once `width` is positive.
* Improved handling for `NaN` and `Infinity` values in expressions.
* Fixed click-through compatibility with tablet mode.
* Fixed crashes.
* Minor misc theme fixes.


TheGamer1445891:
>     * Updated themes to avoid changing the taskbar width when it vertical mode. Previously, some themes could cause the taskbar to fill the whole screen, making it difficult to recover from that state.

@m417z did you remove this in the mod or just in the styling guide? my theme uses width for vertical taskbar so the vertical dock theme can be applied correctly

PR #5491, opened 2026-09-13
Update mod: Taskbar Blob Shape (1.2.0)

Deen-0x:
## Taskbar Blob Shape 1.2.0 — taskbar position support

Windows 11 can now dock the taskbar to any screen edge; this release makes the blob follow. Docking is read from the RootGrid's `DockingStates` visual state group — older builds without the group read as bottom-docked, so nothing changes there.

### Behavior per position

* **Bottom** — the existing look, unchanged.
* **Top** — the silhouette is mirrored vertically, with its flat edge anchored to the taskbar's bottom edge (the side facing the desktop) and the body's excess hanging above the screen edge, exactly mirroring how it hangs below on a bottom-docked taskbar.
* **Left / Right** — blobs are hidden and the native indicator stays fully in charge: the tab silhouette has no sideways variant.

Position changes apply live: each taskbar host subscribes `DockingStates.CurrentStateChanged` and re-sweeps its island on a switch, so every blob — system tray included — re-anchors or hides without an Explorer restart. The subscription follows the same tracked-and-detached lifecycle as the existing SizeChanged re-sweep.

### Implementation notes

* **The mirror is baked into the geometry** (`Geometry.Transform` with an absolute-center `ScaleTransform`), not applied as an element `RenderTransform`: the flipped path reaches the rasterizer as final coordinates and fills through the identical pipeline as the upright shape. An element-level negative-scale transform was observed to shift the flat edge by one pixel via compositor snapping, even with exact integer coordinates. The transform is re-applied on geometry rebuilds, since a fresh PathGeometry carries none.
* **The top-dock anchor is rounded on the flat edge** — the visually critical one — rather than the box top, so a fractional box height (auto height under display scaling) is absorbed by the rounded end at the screen edge instead of shifting the edge users see.
* **Vertical margins are anchored-edge-relative** rather than screen-relative: `Top` pushes the blob deeper into the bar (away from its flat edge) and `Bottom` pulls it back out, so one margin configuration produces the same inward nudge on both horizontal docks. The setting description documents the semantics.
* Y positioning remains a structural constant per dock edge — no element position is ever sampled for Y, preserving the 1.0.0 anti-race design.

### Tested

* Windows 11 25H2 (OS Build 26200.9445), 100% scale: live position cycling Bottom → Top → Left → Right → Bottom with active windows; top-dock pixel alignment verified with and without margins (flat edge flush at the bar's bottom edge; `0, 4, 0, 0` yields a true 4px inset); size and radius changes while top-docked (transform re-application on geometry rebuild); tray and clock blobs flip in agreement with task blobs; vertical docks show untouched native indicators; enable/disable cycle.

### Changelog

* Supports the taskbar position feature: on top-docked taskbars the blob is flipped, with its flat edge anchored to the taskbar's bottom edge; on vertical (left/right) taskbars blobs are hidden and the native indicator stays in charge. Position changes apply live. Older builds without the feature are unaffected.
* Vertical margins are now relative to the taskbar's anchored edge (Top pushes the blob deeper into the bar on both bottom and top docks).

### DIFF
--- a/mods/taskbar-blob-shape.wh.cpp
+++ b/mods/taskbar-blob-shape.wh.cpp
@@ -2,7 +2,7 @@
 // @id              taskbar-blob-shape
 // @name            Taskbar Blob Shape
 // @description     Injects a customizable blob shape behind active taskbar items.
-// @version         1.1.0
+// @version         1.1.1
 // @author          Deen-0x
 // @github          https://github.com/Deen-0x
 // @include         explorer.exe
@@ -153,7 +153,10 @@ std::mutex g_settingsMutex;
 struct BlobEntry {
     winrt::weak_ref<winrt::Windows::UI::Xaml::FrameworkElement> button;
     winrt::weak_ref<winrt::Windows::UI::Xaml::Shapes::Path> blobShape;
-    winrt::weak_ref<winrt::Windows::UI::Xaml::Controls::Grid> grid;
+    // The hosting panel: the taskbar's RootGrid, or the tray's
+    // SystemTrayFrameGrid — a Grid on older Win11 builds, a StackPanel on
+    // newer ones, hence Panel.
+    winrt::weak_ref<winrt::Windows::UI::Xaml::Controls::Panel> grid;
     winrt::weak_ref<winrt::Windows::UI::Xaml::FrameworkElement> anchor;
 
     // Cached per-button element lookups, re-resolved when expired or
@@ -226,11 +229,11 @@ std::atomic<bool> g_unloading{false};
 // the first sweep. Weak refs + tokens only — destructor-safe as a plain
 // global. Guarded by g_blobEntriesMutex.
 struct SweptHost {
-    winrt::weak_ref<winrt::Windows::UI::Xaml::Controls::Grid> grid;
+    winrt::weak_ref<winrt::Windows::UI::Xaml::Controls::Panel> grid;
     winrt::event_token sizeToken{};
     // Cached tray grid for this taskbar's island, resolved lazily: spares
     // the depth-10 class search over the whole XAML root on every re-sweep.
-    winrt::weak_ref<winrt::Windows::UI::Xaml::Controls::Grid> trayGrid;
+    winrt::weak_ref<winrt::Windows::UI::Xaml::Controls::Panel> trayGrid;
 };
 std::vector<SweptHost> g_taskbarHosts;
 std::vector<SweptHost> g_trayHosts;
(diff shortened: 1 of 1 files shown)


Deen-0x:
All applied — the mandatory fix and all five optionals. The round:

Mandatory — the TOCTOU gate moved to where the truth is. EnsureBlobOnButton re-reads g_dockEdge at show time — provably after any first-contact sweep inside the same call has written it — and folds the vertical check into show directly. The idle de-bind now keys off !show rather than !isActive, so a vertical dock also stops the expression (and the logic composes correctly with the bind-pending transient: !show && bound is false while a bind is still settling). The RefreshBlob block survives as a documented fast path with an explicit comment that it can be stale and why the Ensure gate is the authoritative one. Net effect on the reviewer's trace: the first button on a left/right taskbar now ends its very first refresh hidden, natives intact, expression stopped.

Optionals:

O1 + O2 merged naturally: the dock attach became EnsureDockSubscription, called from every TaskbarSweepBody — subscribe-once semantics via the "already has a group" check under the lock, so on old builds it's one null lookup per sweep, and a host whose DockingStates wasn't attached at first contact gets retried on every subsequent sweep (your point about Bottom↔Top not necessarily resizing made SizeChanged an unreliable backstop, so the sweep cadence is the right retry vehicle). Inside it, both weak refs are built before the subscription, closing the throw-after-subscribe leak ordering. The first-contact-only block is deleted from SweepExistingButtons.
O3: vertical docks early-return from RefreshBlob when !bound && !bgHidden && !indicatorHidden — nothing to create, nothing to undo — so no Path, no insertion, no subscriptions, no bind-and-stop churn. Entries still showing a blob when the dock switches fall through exactly once for the hide-and-restore pass. This interacts safely with the mandatory fix: the first-button stale read means the early return doesn't fire on that one call, and the Ensure gate catches it.
O4: both bottom-only doc strings are now dock-neutral (README bullet and the Dimensions description), phrased around the anchored edge.

Deen-0x:
Applied and verified — the registration block sits inside the early return exactly as the review's specified: non-tray kinds resolve the grid (from the entry cache first, GetHostRootGrid on miss, cached back either way) and call SweepExistingButtons before returning, terminating through firstTime == false just like the disabled-kind branch. The fast-path comment now names the hazard pattern explicitly — island registration is a side effect of reaching Ensure, and every early return must preserve it — because this is the second time a short-circuit severed it, and the comment is where the third attempt will be standing when it happens.

m417z:
The PR can't be merged, please fix the conflicts.

Deen-0x:
> The PR can't be merged, please fix the conflicts.

done

Deen-0x:
@m417z is ai review required again?

PR #5492, opened 2026-09-13
Add chromium-cursor-remap mod

mazany:
## What this mod does

Chromium browsers and Electron apps draw ten CSS cursors (`grab`, `grabbing`, `zoom-in`, `zoom-out`, `cell`, `vertical-text`, `alias`, `copy`, `col-resize`, `row-resize`) from bitmaps bundled in their own resources instead of asking Windows, so those cursors ignore the user's cursor scheme. This mod replaces them, in `SetCursor`, with the matching cursor from the Windows scheme or with a `.cur` / `.ani` file chosen per type, and leaves every other cursor alone, including a web page's own `cursor: url(...)` images. Tested on Windows 11 10.0.26300 with Microsoft Edge 153, Google Chrome 152, MarkText 0.19.1 (Electron); process gating was also checked in a 32-bit NW.js app.

## Notes for the reviewer

**Why `@include *`.** Electron and CEF apps ship under arbitrary executable names, and so do Edge's WebView2 hosts, so a list of names misses most of them. `Wh_ModInit` returns `FALSE`, before registering any hook, unless the executable exports Chromium's `GetHandleVerifier` and `IsSandboxedProcess` and also carries cursor resources or statically imports a `*_elf.dll` (or `libcef.dll` is already loaded), and it also returns `FALSE` in Chromium's child processes (a non-empty `--type=` switch, parsed the way Chromium parses it, because a substring search would match a URL on the command line; a line with no `type=` in it at all, ignoring quotes and case, is not split, so the browser process never loads `shell32.dll` for it). On the test machine that rule matched Chrome, Edge, WebView2, Vivaldi, Steam's CEF host and every installed Electron app, and rejected the programs that export the same two names only because they borrowed Chromium's sandbox (JetBrains IDE launchers, Acrobat, Firefox-based browsers). There is no `@exclude` list: the check costs two `GetProcAddress` calls in a process that is not Chromium-based.

**Why substitute in `SetCursor`.** The handles Chromium loads stay Chromium's: the mod never swaps them at load time, and only changes what is passed to `SetCursor`. Replacement handles are shared system cursors or `.cur` / `.ani` files loaded once per path and never destroyed, because one may be on screen when the settings change or the mod unloads.

**How a cursor is recognised.** `GetIconInfoExW` names the module and resource a cursor was loaded from; the mod loads that resource again at 32x32 and matches an FNV-1a hash of the pixels plus the hotspot against ten entries. The result does not depend on the size the app loaded the cursor at, and the table is identical in Chrome 152, Edge 152 and 153, several Electron apps and Steam's CEF. A cursor with no resource behind it is always left alone.

**Hooks and locking.** `DestroyCursor` and `DestroyIcon` are hooked only to drop cached verdicts, so a reused handle value is never answered from the cache; current Windows 11 builds export both at one address, so `DestroyIcon` is hooked only where `GetProcAddress` returns a different one. The lock is an SRW lock that is never held across a call out of the mod (no original function, cursor load or `Wh_Log` under it), classification runs outside it with a check that the handle was not destroyed and the settings did not change meanwhile, and a thread-local guard sends a nested `SetCursor` straight to the original.

**Icon Resource Redirect.** It also changes cursors, by redirecting the resources apps load them from, and it was enabled on the test machine throughout; both mods load into the Edge browser process together. A Chromium cursor it has already redirected no longer matches Chromium's images, so this mod leaves it as Icon Resource Redirect drew it. The two do different jobs: Icon Resource Redirect needs a theme pack that covers the program, while this mod maps Chromium's bundled cursors to whatever cursor scheme Windows is using, in any Chromium-based app, without one.

**Not a tool mod.** It has to run inside the browser process to see `SetCursor`.

**32-bit and ARM64.** There is no `@architecture` line, so the mod is built for x86, x86-64 and ARM64. The PE walk uses the arch-neutral header types.

**Testing.** The installed mod was run inside a small fake Chromium host that the real engine injects: Chromium's 21 cursor resources at the default size and at 32 to 72 px, look-alike page cursors, a child process, an executable with the exports but no UI marker, a settings race of 58 million `SetCursor` calls across four settings changes, 300,000 forced handle-reuse attempts, and a leak probe over 15 settings writes. Host unit tests cover the process check, classification, the settings table, re-entry and unloading. On screen, Edge 153 and Chrome 152 cold-started in throwaway profiles showed the system cursor for the tested types, with a mod-off control showing Chromium's own. The same fake-host run passes with a 32-bit host and the mod's 32-bit build, and the host unit tests also pass built as 32-bit, including classification against a 32-bit NW.js module. A cold-started 32-bit NW.js app loaded the mod into its main process and into none of its helper processes. It compiles without warnings on 1.6.1, 1.7.3 and 2.0.0-alpha.5 for x86, x86-64 and ARM64; the ARM64 build has not been run.

---

## Changelog

* Initial release.

## Mod authorship

This mod was created by:

- - [x] Claude
- - [ ] ChatGPT
- - [ ] Gemini
- - [x] Another AI (please specify): OpenAI Codex


mazany:
Round 2 is addressed in 924b1e0. I intend this to be the last AI review round; anything it raises I will answer here before `/ready-for-reviewer`.

**shell32.dll in every Chromium process** - fixed, with a stricter form of the suggested pre-filter. A plain `wcsstr(L"type=")` would miss a switch that Chromium still parses as `--type=`: quotes can split the name (`"--ty"pe=renderer`) and switch names are ASCII case-insensitive (`--TYPE=renderer`). So `HasChromiumTypeSwitch` now removes `"` and lowercases A-Z before looking for `type=`, and returns "not a child" without touching shell32 when it is absent. That is a necessary condition for any argument holding the switch, because splitting deletes quotes and a backslash it deletes always leaves a backslash or a quote in its place; it was checked against the real `CommandLineToArgvW` over about a million random lines, and the differential test against Chromium's parsing rules now covers mixed case (4,001,437 lines, 0 mismatches). A new test counts the mod's `LoadLibraryW("shell32.dll")` calls and fails on the previous revision.

- The browser process, where the mod stays, no longer loads shell32 unless its command line contains `type=` (in a URL, say). The fail-closed branch now needs that *and* a failed shell32 load, and it logs.
- A real child still loads and frees shell32 once during init, as the suggested snippet would too.

**Optional items and notes**, one line each:

- Simpler `_wcsnicmp(argv[i], L"--type=", 7)`: kept the full parse. It is what the differential test checks, and `_wcsnicmp` would count an empty `--type=` (a browser) as a child and ignore the `-`/`/` prefixes, `--` and last-value-wins.
- `WindhawkUtils::SetFunctionHook`: kept as is. The targets come from `GetProcAddress` so the shared-address case can be detected; the change would be style only.
- `WindhawkUtils::StringSetting`: kept as is, style only.
- `ForgetCursor` early-out: not taken. A classification that starts after the shared-lock check while the destroy call is still running would cache a verdict for a dying handle, which is what `g_destroying` prevents.
- `g_fileCursors` never emptied: not taken. A handle dropped by one settings change can still be an idle window's current cursor, and `SetCursor_Hook` passes the handle to the original outside the lock, so destroying it on a later change is not safe. The cost is one cursor object per distinct path per process.
- Test mode: kept, so a user can see which cursors the mod catches without opening the log.
- alias/copy/zoom defaults: kept by decision; every type has "Keep Chromium's own cursor".
- Resource-ID fallback: not taken. The hash is identical across every build checked; resource IDs across forks were not checked.
- x86: left as the documented limitation.
- `SetSystemCursor` retiring a handle: noted as an assumption; no defence short of re-validating on every cache hit, and not seen with any tested app.

mazany:
Round 3 is addressed in e4cd553. It was meant to be the last AI round; the 32-bit support it asked for is why there is a fourth. Anything this one raises I will answer here, then `/ready-for-reviewer`.

**32-bit apps** - done. `@architecture` is gone, so the mod builds for x86, x86-64 and ARM64. The PE walk uses the arch-neutral types, including a fourth one the suggested patch missed: `sizeof(IMAGE_NT_HEADERS64)` in the headers bound. What was checked for the 32-bit build:

- The fake Chromium host built as x86, injected by the real engine with the mod's 32-bit DLL: Chromium's 21 cursors at every size, the child and gate-negative cases, the settings race, handle reuse and the leak probe all pass, with cursor resources from a 32-bit NW.js module.
- The host unit tests built as x86, with PE32 versions of the import-walk fixtures, and classification against a 32-bit NW.js module (same fingerprints).
- The Chromium check over 3,846 PE32 executables in Program Files, ProgramData, the user's app data, Steam and Windows on the test machine: it accepts only Chromium UIs (Windhawk's own VSCodium, Samsung Magician, the test host) and rejects 32-bit Acrobat and a Qt WebEngine process that export the same two names. A second Steam library was scanned only in part.
- A cold-started 32-bit NW.js app: the mod stayed in its main process and left all of its helper processes.

The ARM64 build compiles on all three versions but has not been run; the readme says so.

**Optional items and notes**, one line each:

- `WindhawkUtils::StringSetting` and `WindhawkUtils::SetFunctionHook`: taken. The `GetProcAddress` comparison stays, and its comment now says why: compiled with Windhawk's clang, `(void*)&DestroyCursor == (void*)&DestroyIcon` folds to a constant `false` for x86 and x86-64 without reading an import slot, so comparing the two in source cannot detect the shared export.
- PR description: updated; the `@exclude` list and the Guess setting are gone from it.
- `Wh_ModUninit` / `g_unloading`: kept. In the engine source for 1.6.1, 1.7.3 and 2.0.0-alpha.5 the hooks are disabled, then `Wh_ModUninit` runs, and only then does the engine wait for threads still inside the mod, so a hook call can still be running during uninit.
- `g_loadLock`: kept. Init and settings changes are serialised today, so it is redundant, but it costs nothing.
- Keeping verdicts across settings changes: not taken; settings changes are rare, and the empty cache per change is what the settings race test checks.
- Destroying `g_fileCursors` on unload: not taken. `GetCursor`/`SetCursor` act on the calling thread's cursor, not the UI thread's, `DestroyCursor` requires that the cursor not be in use, and a hook call can still hand the handle out during uninit.
- Defaults for zoom/cell/alias/copy/vertical-text: kept by decision; every type has "Keep Chromium's own cursor".
- Resource IDs as the key: measured, and they are not stable. The ten cursor images are byte-identical across Chrome 153, Edge 152, Edge 153, MarkText, Obsidian and Steam's CEF, while their resource IDs differ in every one - `grab` is 48834, 59014, 58474, 50254, 49534 and 49644.
- A cursor-scheme change mid-session: not tested yet. Settings changes reload the replacement handles, and system cursors are shared handles.
- `SetCursor` returning the replacement: acknowledged, harmless as noted.

mazany:
Round 4 found no blocking issues, so I am handing this over to a human reviewer now, without another push. The optional items and notes, one line each:

- **Module handle without a reference in `ClassifyCursor`:** not changed for this submission. On Windows 11 26300, `LoadImageW` on the handle of an unloaded DLL returned NULL (error 1812) without faulting, for x86 and x64; the documented contract does not promise that, though, and a reused base could name another module. The suggested `GetModuleHandleExW` + `FreeLibrary` has its own cost: if the mod held the last reference, the DLL's detach would run inside `SetCursor`. The cursors this mod is for come from the executable or Chromium's main DLL.
- **shell32 in child processes:** kept. It is one load and unload per child during init, and the alternative, letting the mod stay in every renderer, GPU and utility process, puts hooks where they do nothing.
- **`-luser32`:** redundant, agreed; not worth reopening the review for. The `DestroyCursor_Original ? ... : DestroyCursor` fallback is not dead, though: the host tests call `ClassifyCursor` before any original is set.
- **Windows 10:** not tested; no Windows 10 machine here. `DestroyIcon_Hook` is covered by host tests that call it directly, but registering it on a separate export is not, because every user32 checked (x86 and x64, 10.0.26100.x) exports both functions at one address.
- **Cursor scheme or size change mid-session:** still not tested.

PR #5493, opened 2026-09-13
Update Windows 11 Native Four Column Snap Layout to v1.1

Hoffelhas:
## Changelog

Fixed the Snap Layout picker height for varying suggestion counts when only one or two windows are available


Hoffelhas:
Addressed the height-logic review points and verified the main concerns experimentally.

- Made get_PickerHeight and the typed suggestion-vector get_Size hooks optional, so failure to resolve them no longer prevents the four-column layout from loading.
- Removed exact 166 / 244 height matching. The correction now checks whether adding the seventh layout actually creates another visible row.
- Confirmed the picker uses 3 layout cards per row on the tested configuration.
- The DPI concern is settled on the tested build: get_PickerHeight returned the same native values (166 / 244) at 100% and 150% scaling, so it behaves like logical/DIP units rather than physical pixels.
- Instrumented suggestion counts directly: 1 window → 0 suggestions, 2 windows → 2, 3/4 windows → 3. The count is available before get_PickerHeight.
- The extra 78-unit row is therefore applied only when needed: 0 suggestions → add 78, 2 suggestions → unchanged, 3+ suggestions → add 78.
- If the optional suggestion-count hook is unavailable, no height correction is applied, so the mod degrades conservatively instead of failing.
- Restored the useful reverse-engineering comments and cleaned up the formatting.

Tested visually with 1, 2, 3, and 4 available windows; the picker spacing now looks correct across the tested states.

/ai-review

Hoffelhas:
It now addresses the previous picker-sizing and ARM64 concerns.

- Windows selects either 4 native layouts (2 per row) or 6 native layouts (3 per row); the mod uses that native count rather than screen resolution.
- Picker height now uses the actual LayoutsAndSuggestions collection. The generic IVector<IInspectable>::get_Size hook is correlated with the exact vector returned by get_LayoutsAndSuggestions.
- The previous suggestion-count heuristic is gone. Height is increased by 78 only when adding the custom layout creates an extra row.
- Verified at 1280×720, 1920×1080, and 5120-wide, with 1–4 available windows, plus Snap Bar.
- Native ARM64 remains intentionally unsupported due to the private C++ return ABI.

/ai-review

PR #5495, opened 2026-09-13
Add Minimize to tray mod

0Allu:
Adds the new Minimize to tray mod, which allows users to right-click a window's minimize button to hide it in the system tray.

## Changelog

* This pull request introduces a new mod.

## Mod authorship

This mod was created by:

* * [ ] The submitter, without AI assistance
* * [x] The submitter, with AI assistance
* * [ ] Claude
* * [x] ChatGPT
* * [ ] Gemini
* * [ ] Another AI (please specify):
* * [ ] Other (please specify):


0Allu:
The latest version keeps the architectural fixes from the AI reviews, including the dedicated low-level mouse hook thread, non-blocking input path, async probing, DPI-aware coordinates, safe tray recovery, bounded unload, and protection against hiding a window before its tray icon is available.

After testing, I intentionally restored a generic custom-title-bar geometry fallback because making detection rely only on strict native hit-testing or per-application rules caused common applications such as Explorer, VS Code, Spotify, and other Chromium/Electron apps to stop working. The generic fallback is configurable and can be disabled, while Steam keeps a narrow application-specific rule due to its different caption layout.

This is a deliberate compatibility tradeoff: the implementation keeps blocking/cross-process work off the low-level input path while preserving the broad application support that is the main purpose of the mod.

0Allu:
1. The generic caption heuristic is intentionally kept broader because stricter detection caused common apps like Explorer, VS Code, Spotify, Discord, and Steam to stop working reliably; users can disable compatibility detection if needed.

2. The stale-verdict edge case is valid, but I plan to address that hardening in a later version to avoid another detection regression before the initial release.

3. Probe throttling is a valid performance improvement, but since the expensive work already runs outside the low-level input hook, I plan to optimize it in a later version.

PR #5496, opened 2026-09-14
Add Vector Screen Holder mod

akilluminati47:
> **Supersedes #5426.** That pull request came from a fork whose lineage ran through another user's fork rather than this repository. The fork has been replaced with a clean one forked directly from `ramensoftware/windhawk-mods`; the code is identical plus the readme images and the fixes listed below. The AI review on #5426 raised eleven required findings, all addressed here.

Adds **Vector Screen Holder**, a tool mod that fills a display of your choosing with generative line art and holds the screen awake while it runs.

The use case it was built for: you're downloading something large, or running a long build, render or backup, and you need the machine not to sign you out or drop to idle. Put the Screen Holder on a side monitor next to your task monitor and walk away — the work keeps running, the screen stays awake, and progress is still readable from across the room.

### What it does

- Runs as a tool mod in its own dedicated `windhawk.exe` process, so it is never injected into other applications.
- Everything is stroked geometry through Direct2D — no images, no video, no fixed resolution. The art is generated for whatever size the selected display actually is, so a 1080p side monitor and a 4K portrait panel each get correctly proportioned output.
- While the overlay is up it calls `SetThreadExecutionState(ES_CONTINUOUS | ES_SYSTEM_REQUIRED | ES_DISPLAY_REQUIRED)`, cleared the moment it closes. It does not synthesise input.
- The overlay sits above the wallpaper but below application windows, never steals focus on its own, and stays out of Alt+Tab.

### Styles

Four, each with a wheel parameter (its character) and a right-click amount (how much information is on screen, five notches):

| Style | Parameter | Amount |
| --- | --- | --- |
| Flow field | field turbulence | ribbon packing density |
| Contours | terrain relief | 6–46 iso levels |
| Differential growth | colony vigor | 1–6 colonies |
| Harmonograph | draw tempo | 1–6 overlaid figures |

### Controls

`Esc` closes, left click cycles style, right click steps the amount, the wheel adjusts the parameter, and holding `Space` slides the hue. A configurable global hotkey toggles it, and it also listens on a named event so a desktop shortcut can drive it. The overlay is clean when it opens — labels appear only while you're actively changing something.

### Notes on implementation

- Scenes advance on a wall-clock step clock rather than per frame, so the frame-rate setting changes how smooth the motion is without changing how fast anything draws.
- Contours interpolate between two sampled noise fields each frame, so movement stays continuous instead of stepping on the sampling interval. The noise grid is area-budgeted rather than tied to pixel count, keeping per-frame cost flat across resolutions.
- Hue rotation is applied at stroke time rather than to the finished image, since Direct2D 1.0 has no hue-rotate effect. Styles that redraw each frame slide through colour wholesale; styles that accumulate lay a gradient through the artwork as the hue drifts.

Builds clean with the bundled toolchain at `-O2 -Wall -Wextra`.

---

### On scope (review finding 1)

The reviewer is right that this doesn't hook a host process, and that's worth addressing head on rather than leaving implied.

The case for it being in scope: the "mods as tools" pattern already covers utilities that act on the *session* rather than on one program's internals, and this is one of those — it puts a window on a chosen display, sinks it into the desktop's z-order beneath the user's windows, and manages the session's idle state. That is shell-adjacent behaviour, and it is delivered through the same mechanism (`@include windhawk.exe`, dedicated tool-mod process) as the existing tool mods. It is also genuinely useful to the same people: the audience for a keep-the-screen-up utility overlaps heavily with the audience already running Windhawk on a multi-monitor desk.

The case against, stated fairly: nothing here *needs* Windhawk. It could ship as a standalone executable, and a maintainer who draws the line at "must modify an existing program's behaviour" would be applying that line consistently by rejecting it. I'm not going to argue otherwise — that call belongs to the maintainer, and a no on those grounds is a reasonable no.

On the overlap with `caffeine`: the art is the point, and the keep-awake is a checkbox next to it rather than the feature. I've kept it because an overlay that holds the screen visually but still lets the machine sleep would be self-defeating, and asking users to install a second mod to make this one do its stated job is worse for them. It's one `SetThreadExecutionState` call behind a setting that can be turned off, not a reimplementation of caffeine's tray UI, timed modes or state restore. If the maintainer would rather this mod not carry it at all and defer to `caffeine`, I'll remove the setting — say the word.

### Changes since the review

All eleven required findings are addressed in `9af00c12`, along with most of the optional ones. The two that changed real behaviour:

* **The overlay was never below the user's windows.** The `WM_WINDOWPOSCHANGING` handler neutralised its own `HWND_BOTTOM` call, so the central README promise was simply not what the code did. Fixed by re-sinking on every position change.
* **The keyboard hook stalled system-wide input**, because it was installed from the render thread. Moved to a dedicated pumping thread.

Two notes where I did something other than what was suggested:

* **Frame rate default.** The review suggested defaulting to 30 fps on cost grounds. I've kept 60, because the contour motion is now interpolated against wall-clock time and the difference is visible; instead I've documented the cost explicitly in a new "Performance" section and pointed at the three knobs that reduce it. 30 fps remains a one-click change.
* **The readout** is implemented rather than removed, since the alternative left the user with no feedback at all about what they'd just changed. It's a single line that fades after ~2s, drawn with DirectWrite.

### Readme images

The description now shows the art: a short loop of the overlay running under a
real application window while a note is typed on a side monitor, and a strip of
the four styles beside the table that names them. Both are served from
`raw.githubusercontent.com`, hosted in the mod's own repository
([akilluminati47/vector-screen-holder](https://github.com/akilluminati47/vector-screen-holder)),
so nothing in the description depends on a fork continuing to exist.

## Changelog

* New mod — not applicable.

## Mod authorship

This mod was created by:

- - [x] Claude
- - [ ] ChatGPT
- - [ ] Gemini

### DIFF
--- a/mods/vector-screen-holder.wh.cpp
+++ b/mods/vector-screen-holder.wh.cpp
@@ -152,18 +152,17 @@ pair-programmers Claude and Big-Pickle (opencode).
 - rotateSeconds: 300
   $name: Rotation interval (seconds)
   $description: Clamped to 10-7200.
-- amount: 2
+- amount: balanced
   $name: Amount
   $description: >-
     Starting amount notch -- how much information is on screen. Right click the
-    overlay to step it. 0 = minimal, 1 = sparse, 2 = balanced, 3 = dense,
-    4 = maximal.
+    overlay to step it.
   $options:
-  - 0: Minimal
-  - 1: Sparse
-  - 2: Balanced
-  - 3: Dense
-  - 4: Maximal
+  - minimal: Minimal
+  - sparse: Sparse
+  - balanced: Balanced
+  - dense: Dense
+  - maximal: Maximal
 - parameter: 50
   $name: Parameter (%)
   $description: >-
@@ -2226,6 +2225,19 @@ static std::wstring GetStringSetting(PCWSTR name) {
     return s;
 }
 
+// The amount notch is stored by name so the setting can carry $options.
+static int AmountFromName(const std::wstring& name) {
+    static const wchar_t* kNames[kAmountCount] = {L"minimal", L"sparse",
+                                                  L"balanced", L"dense",
+                                                  L"maximal"};
+    for (int i = 0; i < kAmountCount; i++) {
+        if (name == kNames[i]) {
+            return i;
+        }
+    }
+    return 2;   // balanced
+}
+
 static void LoadSettings() {
     g_settings.monitor = GetStringSetting(L"monitor");
     g_settings.enable[kStyleFlow] = Wh_GetIntSetting(L"enableFlow") != 0;
@@ -2244,7 +2256,7 @@ static void LoadSettings() {
     g_settings.rotate = Wh_GetIntSetting(L"rotate") != 0;
     g_settings.rotateSeconds =
         ClampT(Wh_GetIntSetting(L"rotateSeconds"), 10, 7200);
-    g_settings.amount = ClampT(Wh_GetIntSetting(L"amount"), 0, kAmountCount - 1);
+    g_settings.amount = AmountFromName(GetStringSetting(L"amount"));
     g_settings.parameter = ClampT(Wh_GetIntSetting(L"parameter"), 0, 100);
     g_settings.palette = GetStringSetting(L"palette");
     g_settings.customColors = GetStringSetting(L"customColors");
--- a/mods/vector-screen-holder.wh.cpp
+++ b/mods/vector-screen-holder.wh.cpp
@@ -1,0 +1,2630 @@
+// ==WindhawkMod==
+// @id              vector-screen-holder
+// @name            Vector Screen Holder
+// @description     Fills a display you choose with generative line art and keeps the PC from idling while it runs
+// @version         1.0.1
+// @author          akilluminati47
+// @github          https://github.com/akilluminati47
+// @include         windhawk.exe
+// @compilerOptions -ld2d1 -ladvapi32 -lole32 -luser32 -lgdi32 -lshell32
+// ==/WindhawkMod==
+
+// ==WindhawkModReadme==
+/*
+# Vector Screen Holder
+
+Fills a display of your choosing with generative line art, and holds the screen
+awake for as long as it runs. Every frame is drawn as strokes through Direct2D
+on the GPU -- no images, no video file, and no fixed resolution. The art is
+generated for whatever size the display you pick actually is, so a 1080p side
+monitor and a 4K portrait panel each get correctly proportioned artwork.
+
+Downloading a large file, or running a long build, render or backup, and you
+need the machine not to sign you out or drop to idle? Start the Screen Holder,
+put it on a side monitor next to your task monitor, and go to lunch. The work
+keeps running, the screen stays awake, and you can still see progress at a
+glance from across the room.
+
+## Controls
+
+The overlay is clean when it opens -- no labels, no chrome, nothing on screen
+but the art. The style name and its two readouts appear only while you are
+actively changing something, then fade away again.
+
+| Input | What it does |
+| --- | --- |
+| **Esc** | Close the overlay |
+| **Left click** | Cycle to the next enabled style |
+| **Right click** | Step the amount: how much information is on screen |
+| **Mouse wheel** | Adjust the current style's parameter |
+| **Hold Space** | Slide the colour around the hue wheel |
+| **Ctrl+Alt+H** | Toggle the overlay on and off (configurable below) |
+
+Click, right click and wheel need the overlay focused -- click it once. **Hold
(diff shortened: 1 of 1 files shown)


akilluminati47:
Note for the next round: the optional suggestion to inline the toggle-event shortcut recipe in the mod description was considered and declined. The required part, that the event was undocumented, is addressed: the description now names `Local\WindhawkVectorScreenHolderToggle` and its namespace. The recipe itself stays on the linked page on purpose, so the description does not carry four paragraphs of VBS and security notes for every reader in order to serve the few who want a shortcut. Every other item from that round, required and optional, is in.

---

Generated by AI. Model: Claude Opus 5.

akilluminati47:
Round notes for the next pass.

Addressed: the unload hang (a manual-reset quit event is now in the worker wait set, and the hook join is bounded at 5s), the rotation timer escaping the occlusion freeze, per-frame retries on device loss, and startActive reopening the overlay on unrelated settings changes.

Verified rather than assumed: the Opacity concern does not reproduce. CreateHwndRenderTarget returns S_OK on a WS_EX_LAYERED window at alpha 255, 128 and 64, and the present succeeds at each, so opacity below 100 renders normally on Windows 11 with DWM.

Declined, with reasons:
- The embedded pixel font stays. It is a deliberate requirement from the mod author, not an artifact: the readout is translated into 17 languages and the font is what keeps it legible and consistent without asking anyone to install anything. The path is exercised and tested, including the system-font fallback for scripts the subset does not carry.
- ClampT stays rather than becoming std::clamp. std::clamp returns const T&, and these call sites pass prvalue expressions, so a by-value clamp avoids a dangling-reference footgun for no user-visible difference.
- The toggle-event shortcut recipe stays on the linked page rather than inline, so the description does not carry four paragraphs of VBS and security notes for every reader.
- EnumerateMonitors running twice at startup and the flow-field spine retention are left as they are; both are noted as harmless and neither affects behaviour.

---

Generated by AI. Model: Claude Opus 5.

akilluminati47:
### On the fullscreen probe: the nudge branch, forced to run

This note has been carried for three rounds asking for the branch to be
exercised once, since it cannot fire on a machine where the overlay is
correctly not counted. Here are the results rather than another round of it.

**The branch, driven end to end.** `PollFullscreenState` and
`Overlay::SetNudged` were lifted verbatim out of the mod and driven against a
stand-in window that *is* counted, because it is deliberately not sunk to
`HWND_BOTTOM`. One poll tick:

```
state with the overlay covering the monitor and focused: BUSY

calling PollFullscreenState(0.6f):
  [mod log] This Windows build counts the overlay as a fullscreen app;
            stepping one pixel aside so notifications are not suppressed
  nudge flag        : set
  window moved up 1px, same size: yes
  state now         : ACCEPTS

focus leaves:
  nudge flag        : cleared
  window back on its rectangle: yes
```

So the branch fires, the log line appears, the one pixel move clears the
suppression within that poll, and releasing on focus loss puts the window
back on its rectangle.

**Does the shell re-evaluate on a move, or only on activation?** On a move.
This was measured before the branch was written, because the design depends on
it. A `WS_POPUP` tool window covering the primary monitor and holding the
foreground, at normal z-order:

| geometry | state |
| --- | --- |
| covering the monitor | `QUNS_BUSY` |
| moved up 1px, `SWP_NOSIZE`, no activation change | `QUNS_ACCEPTS_NOTIFICATIONS` |
| moved back | `QUNS_BUSY` |

Two runs, identical. The per-focus cycle therefore catches a click in time and
the latching alternative is not needed to make it correct.

**Latching was considered and declined deliberately.** It would leave the
overlay permanently a pixel off its rectangle on an affected machine, and that
pixel is a visible line of wallpaper along a display edge. An earlier revision
shipped exactly that as an unconditional trim and it was the one thing users
noticed, which is why it was removed. Releasing on focus loss confines the
offset to the moments someone is interacting with the artwork, where the
readout is on screen anyway.

**The false positive is real and accepted.** A genuine fullscreen application
on another display can report `QUNS_BUSY`, so a click on the overlay can nudge
it for nothing. It costs one pixel and one log line, and it corrects itself as
soon as focus leaves.

**Why the probe exists at all.** The overlay is not counted as a fullscreen
app in the first place, because `WM_WINDOWPOSCHANGING` forces it to
`HWND_BOTTOM` on every position change. Three runs of each, pinned against
normal z-order, both covering the primary and holding the foreground:

| z-order | state |
| --- | --- |
| pinned to `HWND_BOTTOM`, as the mod is | `QUNS_ACCEPTS_NOTIFICATIONS` |
| normal | `QUNS_BUSY` |

That is a Windows 11 result and there is no Windows 10 machine here to confirm
it holds on older shells, which is precisely why the mod asks at runtime
instead of trusting it.

All measurements were taken on Windows 11 build 26200.

---

Generated by AI. Model: Claude Opus 5.

PR #5498, opened 2026-09-14
Update file-explorer-details-autofit-columns to 1.1.0

armaninyow:
## Changelog

* Fixed "Right-click context menu → Refresh" and "The refresh button in the toolbar" not working
* Added Fit Mode settings
  * Visible Rows Only (current one)
  * Scan Entire Folder
  * Elastic (macOS-like)
* Added Max Items to Scan setting


wakhh:
// ==WindhawkMod==
// @id              file-explorer-details-autofit-columns-fork
// @name            File Explorer Details Auto-Fit Columns - Fork
// @description     Automatically fits all column widths to their content when refreshing in Details view. Has no effect on other view modes.
// @version         2.0.0
// @author          Armaninyow
// @github          https://github.com/armaninyow
// @compilerOptions -lole32 -lshlwapi -lpropsys -lshell32 -lcomctl32
// @license         MIT
// ==/WindhawkMod==

// ==WindhawkModSettings==
/*
*/
// ==/WindhawkModSettings==

// ==WindhawkModReadme==
/*
# File Explorer Details Auto-Fit Columns

Automatically fits all visible column widths to their content whenever a folder is opened or refreshed in `Details view`, eliminating truncated text and ellipsis (`...`).

## Triggers

- Opening or navigating to a folder
- Minimizing and restoring the Explorer window
- `Ctrl + R`
- `F5`
- Right-click context menu → Refresh
- The refresh button in the toolbar

Only affects `Details view`. Other view modes (Icons, Tiles, List, etc.) are untouched.

## How it works

Explorer remembers column widths per folder, and for most folders that remembered width is already fine. Early auto-fit attempts during loading would overwrite it with a wrong (collapsed) width, because the file list is not fully loaded yet.

To avoid that, the mod uses a two-phase probe:

- **Early phase**: probes `ItemCount` every 150ms, up to 10 times. Column widths are **not touched** during this phase, so the remembered width stays visible. As soon as `ItemCount` is unchanged across two consecutive probes (and > 0), the view is considered stable and columns are auto-fitted once. Then the mod moves to the late phase.
- **Late phase**: probes `ItemCount` every 1000ms, 3 times. Only if `ItemCount` differs from the last auto-fit does it auto-fit again. Otherwise it does nothing.

If the early phase reaches 10 probes without stabilizing, it auto-fits once and moves to the late phase anyway.

No settings are required.
*/
// ==/WindhawkModReadme==

#include <initguid.h>
#include <shobjidl.h>
#include <propkey.h>
#include <shlobj.h>
#include <shlguid.h>
#include <servprov.h>
#include <exdisp.h>
#include <vector>
#include <unordered_map>
#include <windhawk_utils.h>

DEFINE_GUID(IID_IFolderView2_,
    0x1af3a467, 0x214f, 0x4298,
    0x90, 0x8e, 0x06, 0xb0, 0x3e, 0x0b, 0x39, 0xf9);

DEFINE_GUID(IID_IColumnManager_,
    0xd8ec27bb, 0x3f3b, 0x4042,
    0xb1, 0x0a, 0x4a, 0xcf, 0xd9, 0x24, 0xd4, 0x53);

DEFINE_GUID(IID_IServiceProvider_,
    0x6d5140c1, 0x7436, 0x11ce,
    0x80, 0x34, 0x00, 0xaa, 0x00, 0x60, 0x09, 0xfa);

#define EXPLORER_REFRESH_CMD_1 0xA220  // Ctrl+R / F5
#define EXPLORER_REFRESH_CMD_2 0x7103  // Context menu Refresh
#define AUTOFIT_TIMER_ID       0xAF17

// SVGIO_ALLVIEW value (from shobjidl.h, redefined here to avoid include issues)
#ifndef SVGIO_ALLVIEW
#define SVGIO_ALLVIEW 0x2
#endif

// Window classes that receive refresh commands
static const PCWSTR kSubclassTargets[] = {
    L"ShellTabWindowClass",
    L"ReBarWindow32",
    L"ToolbarWindow32",
};

// Probe rhythm (fixed, no settings)

static const UINT kEarlyProbeInterval = 150;   // early probe interval (ms)
static const int  kEarlyProbeMaxCount = 10;    // max early probes
static const UINT kLateProbeInterval  = 1000;  // late probe interval (ms)
static const int  kLateProbeMaxCount  = 3;     // max late probes

// Phase identifiers
static const int kPhaseEarly = 0;
static const int kPhaseLate  = 1;

// Thread safety

static CRITICAL_SECTION g_cs;

// Per-tab state

static std::unordered_map<HWND, IShellView*> g_tabShellViews;    // tab -> IShellView*
static std::unordered_map<HWND, HWND>        g_windowToTab;      // subclassed hwnd -> tab
static std::unordered_map<HWND, int>         g_tabPhase;         // 0=early, 1=late
static std::unordered_map<HWND, int>         g_tabProbeCount;    // probes in current phase
static std::unordered_map<HWND, int>         g_tabLastItemCount; // last probed ItemCount
static std::unordered_map<HWND, int>         g_tabFittedCount;   // ItemCount at last auto-fit
static std::unordered_map<HWND, bool>        g_tabStable;        // early phase stabilized?

// Clear all per-tab state for a given tab

static void ClearTabState(HWND hwndTab) {
    EnterCriticalSection(&g_cs);
    g_tabPhase.erase(hwndTab);
    g_tabProbeCount.erase(hwndTab);
    g_tabLastItemCount.erase(hwndTab);
    g_tabFittedCount.erase(hwndTab);
    g_tabStable.erase(hwndTab);
    LeaveCriticalSection(&g_cs);
}

// Auto-fit all visible columns to content

static void AutoFitColumns(IShellView* pShellView) {
    IFolderView2* pFV2 = nullptr;
    if (FAILED(pShellView->QueryInterface(IID_IFolderView2_,
                                          reinterpret_cast<void**>(&pFV2))) || !pFV2)
        return;

    FOLDERVIEWMODE viewMode = FVM_AUTO;
    int iconSize = 0;
    HRESULT hr = pFV2->GetViewModeAndIconSize(&viewMode, &iconSize);
    pFV2->Release();

    if (FAILED(hr) || viewMode != FVM_DETAILS)
        return; // not Details view

    IColumnManager* pCM = nullptr;
    if (FAILED(pShellView->QueryInterface(IID_IColumnManager_,
                                          reinterpret_cast<void**>(&pCM))) || !pCM)
        return;

    UINT colCount = 0;
    if (FAILED(pCM->GetColumnCount(CM_ENUM_VISIBLE, &colCount)) || colCount == 0) {
        pCM->Release();
        return;
    }

    std::vector<PROPERTYKEY> keys(colCount);
    if (SUCCEEDED(pCM->GetColumns(CM_ENUM_VISIBLE, keys.data(), colCount))) {
        for (UINT i = 0; i < colCount; i++) {
            CM_COLUMNINFO ci = {};
            ci.cbSize = sizeof(ci);
            ci.dwMask = CM_MASK_WIDTH;
            ci.uWidth = CM_WIDTH_AUTOSIZE;
            pCM->SetColumnInfo(keys[i], &ci);
        }
        Wh_Log(L"Auto-fitted %u column(s)", colCount);
    }

    pCM->Release();
}

// Start the probe sequence for a tab

static void StartAutoFitSequence(HWND hwndTab) {
    if (!hwndTab) return;

    EnterCriticalSection(&g_cs);
    g_tabPhase[hwndTab]         = kPhaseEarly;
    g_tabProbeCount[hwndTab]    = 0;
    g_tabLastItemCount[hwndTab] = -1;
    g_tabFittedCount[hwndTab]   = -1;
    g_tabStable[hwndTab]        = false;
    LeaveCriticalSection(&g_cs);

    SetTimer(hwndTab, AUTOFIT_TIMER_ID, kEarlyProbeInterval, nullptr);
    Wh_Log(L"StartAutoFitSequence on hwnd=%p", hwndTab);
}

// Read current ItemCount from IShellView

static int ReadItemCount(IShellView* pShellView) {
    int count = 0;
    IFolderView2* pFV2 = nullptr;
    if (SUCCEEDED(pShellView->QueryInterface(IID_IFolderView2_,
                                             reinterpret_cast<void**>(&pFV2))) && pFV2) {
        pFV2->ItemCount(SVGIO_ALLVIEW, &count);
        pFV2->Release();
    }
    return count;
}

// Find the ShellTabWindowClass ancestor of a HWND

static HWND FindTabWindow(HWND hwnd) {
    WCHAR cls[64] = {};
    HWND cur = hwnd;
    while (cur) {
        GetClassNameW(cur, cls, 64);
        if (wcscmp(cls, L"ShellTabWindowClass") == 0)
            return cur;
        cur = GetParent(cur);
    }
    return nullptr;
}

// Check if a window class is a subclass target

static bool IsSubclassTarget(HWND hwnd) {
    WCHAR cls[64] = {};
    GetClassNameW(hwnd, cls, 64);
    for (PCWSTR target : kSubclassTargets) {
        if (wcscmp(cls, target) == 0)
            return true;
    }
    return false;
}

// Subclass proc

static LRESULT CALLBACK ExplorerSubclassProc(
    HWND hwnd, UINT uMsg, WPARAM wParam, LPARAM lParam,
    DWORD_PTR dwRefData)
{
    bool isRefresh = false;

    if (uMsg == WM_COMMAND) {
        WORD cmdId = LOWORD(wParam);
        if (cmdId == EXPLORER_REFRESH_CMD_1 || cmdId == EXPLORER_REFRESH_CMD_2)
            isRefresh = true;
    }

    if (uMsg == WM_KEYDOWN && wParam == VK_F5)
        isRefresh = true;

    if (uMsg == WM_NOTIFY) {
        NMHDR* hdr = reinterpret_cast<NMHDR*>(lParam);
        if (hdr && hdr->code == -715 && hdr->idFrom == 0)
            isRefresh = true;
    }

    if (isRefresh) {
        HWND hwndTab = nullptr;
        {
            EnterCriticalSection(&g_cs);
            auto it = g_windowToTab.find(hwnd);
            if (it != g_windowToTab.end())
                hwndTab = it->second;
            LeaveCriticalSection(&g_cs);
        }
        HWND hwndTimer = hwndTab ? hwndTab : hwnd;
        StartAutoFitSequence(hwndTimer);
    }

    if (uMsg == WM_TIMER && wParam == AUTOFIT_TIMER_ID) {
        KillTimer(hwnd, AUTOFIT_TIMER_ID);

        // Read current state
        int  phase     = kPhaseEarly;
        int  probeCnt  = 0;
        int  lastCount = -1;
        int  fitted    = -1;
        bool stable    = false;
        EnterCriticalSection(&g_cs);
        if (auto it = g_tabPhase.find(hwnd);         it != g_tabPhase.end())         phase     = it->second;
        if (auto it = g_tabProbeCount.find(hwnd);    it != g_tabProbeCount.end())    probeCnt  = it->second;
        if (auto it = g_tabLastItemCount.find(hwnd); it != g_tabLastItemCount.end()) lastCount = it->second;
        if (auto it = g_tabFittedCount.find(hwnd);   it != g_tabFittedCount.end())   fitted    = it->second;
        if (auto it = g_tabStable.find(hwnd);        it != g_tabStable.end())        stable    = it->second;
        LeaveCriticalSection(&g_cs);

        // Grab IShellView (AddRef so it stays alive during processing)
        IShellView* pSV = nullptr;
        {
            EnterCriticalSection(&g_cs);
            auto it = g_tabShellViews.find(hwnd);
            if (it != g_tabShellViews.end() && it->second) {
                pSV = it->second;
                pSV->AddRef();
            }
            LeaveCriticalSection(&g_cs);
        }

        if (!pSV) {
            // No view, clear state and bail out
            ClearTabState(hwnd);
            return 0;
        }

        int currentCount = ReadItemCount(pSV);

        bool doFit   = false;
        bool advance = true;

        if (phase == kPhaseEarly) {
            // ---------- Early phase ----------
            if (!stable) {
                // Stable when two consecutive probes match and count > 0
                if (lastCount != -1 && lastCount == currentCount && currentCount > 0) {
                    doFit  = true;
                    stable = true;
                    fitted = currentCount;
                    phase  = kPhaseLate;
                    probeCnt = 0;
                }
            }

            if (!stable) {
                probeCnt++;
                if (probeCnt >= kEarlyProbeMaxCount) {
                    // Early phase exhausted; force one fit and move to late
                    doFit    = true;
                    fitted   = currentCount;
                    phase    = kPhaseLate;
                    probeCnt = 0;
                }
            }

            // Persist state
            EnterCriticalSection(&g_cs);
            g_tabPhase[hwnd]         = phase;
            g_tabProbeCount[hwnd]    = probeCnt;
            g_tabLastItemCount[hwnd] = currentCount;
            g_tabStable[hwnd]        = stable;
            g_tabFittedCount[hwnd]   = fitted;
            LeaveCriticalSection(&g_cs);

            // Schedule next probe
            if (doFit) {
                AutoFitColumns(pSV);
            }
            SetTimer(hwnd, AUTOFIT_TIMER_ID,
                     (phase == kPhaseLate) ? kLateProbeInterval : kEarlyProbeInterval,
                     nullptr);

        } else {
            // ---------- Late phase ----------
            if (currentCount != fitted) {
                doFit = true;
                fitted = currentCount;
            }

            probeCnt++;
            if (probeCnt >= kLateProbeMaxCount) {
                advance = false;
            }

            EnterCriticalSection(&g_cs);
            g_tabProbeCount[hwnd]    = probeCnt;
            g_tabLastItemCount[hwnd] = currentCount;
            g_tabFittedCount[hwnd]   = fitted;
            LeaveCriticalSection(&g_cs);

            if (doFit) {
                AutoFitColumns(pSV);
            }

            if (advance) {
                SetTimer(hwnd, AUTOFIT_TIMER_ID, kLateProbeInterval, nullptr);
            } else {
                ClearTabState(hwnd);
            }
        }

        pSV->Release();
        return 0;
    }

    if (uMsg == WM_NCDESTROY) {
        KillTimer(hwnd, AUTOFIT_TIMER_ID);
        WindhawkUtils::RemoveWindowSubclassFromAnyThread(hwnd, ExplorerSubclassProc);

        EnterCriticalSection(&g_cs);
        g_windowToTab.erase(hwnd);
        g_tabPhase.erase(hwnd);
        g_tabProbeCount.erase(hwnd);
        g_tabLastItemCount.erase(hwnd);
        g_tabFittedCount.erase(hwnd);
        g_tabStable.erase(hwnd);
        auto it = g_tabShellViews.find(hwnd);
        if (it != g_tabShellViews.end()) {
            if (it->second) it->second->Release();
            g_tabShellViews.erase(it);
        }
        LeaveCriticalSection(&g_cs);
    }

    return DefSubclassProc(hwnd, uMsg, wParam, lParam);
}

// Subclass a window if it's a target class

static void SubclassTargetIfNeeded(HWND hwnd, HWND hwndTab) {
    if (!hwnd || !IsSubclassTarget(hwnd)) return;

    EnterCriticalSection(&g_cs);
    bool isNew = g_windowToTab.find(hwnd) == g_windowToTab.end();
    g_windowToTab[hwnd] = hwndTab;
    LeaveCriticalSection(&g_cs);

    if (isNew)
        WindhawkUtils::SetWindowSubclassFromAnyThread(hwnd, ExplorerSubclassProc, 0);
}

// EnumChildWindows callback

struct EnumChildData { HWND hwndTab; };

static BOOL CALLBACK SubclassChildProc(HWND child, LPARAM lp) {
    auto* data = reinterpret_cast<EnumChildData*>(lp);
    SubclassTargetIfNeeded(child, data->hwndTab);
    return TRUE;
}

// Hook: CDefView::UIActivate

using CDefView_UIActivate_t = HRESULT(__thiscall*)(void* pThis, UINT uState);
CDefView_UIActivate_t CDefView_UIActivate_orig = nullptr;

HRESULT __thiscall CDefView_UIActivate_hook(void* pThis, UINT uState) {
    HRESULT hr = CDefView_UIActivate_orig(pThis, uState);

    if (SUCCEEDED(hr) &&
        (uState == SVUIA_ACTIVATE_FOCUS || uState == SVUIA_ACTIVATE_NOFOCUS)) {

        auto* pShellView = reinterpret_cast<IShellView*>(pThis);
        HWND hwndView = nullptr;
        pShellView->GetWindow(&hwndView);
        if (!hwndView) return hr;

        HWND hwndTab = FindTabWindow(hwndView);

        HWND hwndTop = hwndView;
        while (GetParent(hwndTop))
            hwndTop = GetParent(hwndTop);

        // Store IShellView for this tab
        if (hwndTab) {
            EnterCriticalSection(&g_cs);
            auto it = g_tabShellViews.find(hwndTab);
            if (it != g_tabShellViews.end() && it->second)
                it->second->Release();
            pShellView->AddRef();
            g_tabShellViews[hwndTab] = pShellView;
            LeaveCriticalSection(&g_cs);
        }

        // Subclass target windows under hwndTop
        SubclassTargetIfNeeded(hwndTop, hwndTab);
        EnumChildData data = { hwndTab };
        EnumChildWindows(hwndTop, SubclassChildProc, reinterpret_cast<LPARAM>(&data));

        // Kick off the probe sequence
        HWND hwndTimer = hwndTab ? hwndTab : hwndTop;
        StartAutoFitSequence(hwndTimer);
    }
    return hr;
}

// Windhawk entry points

BOOL Wh_ModInit() {
    Wh_Log(L"Init");

    InitializeCriticalSection(&g_cs);

    HMODULE hShell32 = LoadLibraryW(L"shell32.dll");
    if (!hShell32) {
        Wh_Log(L"Failed to load shell32.dll");
        DeleteCriticalSection(&g_cs);
        return FALSE;
    }

    const WindhawkUtils::SYMBOL_HOOK shell32DllHooks[] = {
        {
            {
                L"public: virtual long __cdecl CDefView::UIActivate(unsigned int)",
                L"long __cdecl CDefView::UIActivate(unsigned int)",
                L"public: virtual long __thiscall CDefView::UIActivate(unsigned int)",
                L"long __thiscall CDefView::UIActivate(unsigned int)",
            },
            &CDefView_UIActivate_orig,
            CDefView_UIActivate_hook,
            false
        },
    };

    if (!WindhawkUtils::HookSymbols(hShell32, shell32DllHooks, ARRAYSIZE(shell32DllHooks))) {
        Wh_Log(L"ERROR: Could not hook CDefView::UIActivate");
        DeleteCriticalSection(&g_cs);
        return FALSE;
    }

    Wh_Log(L"CDefView::UIActivate hooked successfully");
    return TRUE;
}

void Wh_ModAfterInit() {}

void Wh_ModUninit() {
    Wh_Log(L"Uninit");

    std::vector<HWND> windowsToUnsubclass;
    std::vector<IShellView*> viewsToRelease;

    EnterCriticalSection(&g_cs);
    for (auto& [hwnd, _] : g_windowToTab) {
        KillTimer(hwnd, AUTOFIT_TIMER_ID);
        windowsToUnsubclass.push_back(hwnd);
    }
    g_windowToTab.clear();
    g_tabPhase.clear();
    g_tabProbeCount.clear();
    g_tabLastItemCount.clear();
    g_tabFittedCount.clear();
    g_tabStable.clear();
    for (auto& [hwnd, pSV] : g_tabShellViews) {
        if (pSV) viewsToRelease.push_back(pSV);
    }
    g_tabShellViews.clear();
    LeaveCriticalSection(&g_cs);

    for (HWND hwnd : windowsToUnsubclass)
        WindhawkUtils::RemoveWindowSubclassFromAnyThread(hwnd, ExplorerSubclassProc);
    for (IShellView* pSV : viewsToRelease)
        pSV->Release();

    DeleteCriticalSection(&g_cs);
}

    Wh_Log(L"SettingsChanged (no settings)");
}

armaninyow:
Addressed everything except the README GIF (tracked separately, not a bug). Since this went through a few iterations against real hardware, here's where the implementation ended up, including two places where I deviated from the suggested fix after it didn't hold up in testing.

## Main issues

1. **Nested pump + `Sleep`** — replaced with a timer-driven state machine (`WM_TIMER`-advanced steps, no pumping, no `Sleep`, anywhere).
2. **Global `volatile bool` reentrancy flag** — replaced with a fit context keyed per owning window/tab, so one tab's fit can't suppress another's.
3. **`WM_APP`-relative message id** — now `RegisterWindowMessageW`.
4. **UI Automation on the automated thread** — *partially reverted, see below.*
5. **Cross-apartment release in `Wh_ModUninit`** — fixed via a registered message that marshals cleanup (timers, `IShellView` release, notification deregistration) onto each window's own owning thread before its subclass is removed.

### Deviation on 4 (UI Automation)

The suggested fix — computing Elastic mode's Name-column width purely from known column widths compared against the list control's client width, with no UI Automation — turned out not to hold up on real Explorer builds. Two problems surfaced in testing:

- The header and list controls don't share a coordinate origin once icon/checkbox indents are involved, so comparing their extents directly is unreliable.
- On at least one Explorer build tested, Details view has **no classic `SysListView32`/`SysHeader32` window at all** (confirmed via logging — `FindDescendantByClass` returns null), so there's nothing for a pure Win32-geometry approach to query in the first place. This is very likely the reason the original implementation reached for UI Automation to begin with.

UI Automation's `ScrollPattern` is back, but scoped down to address the actual complaint (cost and thread-safety), not the technique itself:

- One `IUIAutomation` client, created once and cached **per Explorer UI thread**, not per call.
- `ElementFromHandle` on a known control when one exists (no subtree search) — a `FindFirst(TreeScope_Descendants, ...)` walk is only used as a fallback when there's no classic handle to target directly, and even then it runs **at most once per fit**, with the result cached and reused across every retry, instead of being repeated up to 15 times per fit as before.
- Called from a plain `WM_TIMER` step, same as everything else in the state machine — no nested pump.

Width math against the header/scrollbar is still tried first as a cheap first pass; UI Automation is the fallback/override, not the only signal.

## Optional improvements

All applied as suggested: cached settings (`Wh_ModInit`/`Wh_ModSettingsChanged` instead of per-message re-reads), `WindhawkUtils::StringSetting::make`, dead `if (c == 1) {}` removed, `GetModuleHandleW` + hardened `LoadLibraryExW` fallback, `GetAncestor(hwnd, GA_ROOT)`, `KillTimer`-wrong-thread fixed via the same cleanup-message marshaling as 5, `GetSystemMetricsForDpi`, unused `-lshlwapi`/`-loleaut32`/`<shlwapi.h>` removed, `CBrowserHost::Refresh` got its `__thiscall` variant, shadowed `hwndView` gone (computed once per fit context).

## Functionality notes

- Scroll position restore now targets the item that was actually focused (`IFolderView2::GetFocusedItem`), not item 0.
- Elastic's vertical-scrollbar allowance is now conditional on `WS_VSCROLL` actually being set, and measured against the list control's own rect.
- `kPresumedVisibleCount` is now derived from the list control's real client height ÷ measured row height.
- A DPI-relative cap (800px @ 96 DPI) bounds any single non-Elastic-Name column width.

### Deviation on "one round-trip per column"

Initially implemented exactly as suggested — one combined `SelectAndPositionItems` call for every exact-measured column's widest item instead of one call per column. This caused a real regression: if two columns' widest items are far apart in the list, only one ends up realized/visible for its own `CM_WIDTH_AUTOSIZE`, and the other silently gets measured against whatever's on-screen instead — truncating it. It surfaced only in some folders and flipped with sort direction (sort order changes which items are widest and how far apart they land), which is what made it easy to miss initially.

Reverted to one scroll+autosize+read cycle per exact column (matching the original's per-column correctness), but kept the final scroll-position restore batched to a single call, since that only ever needs to reposition to one item. Net effect: still fewer round-trips and less visual jumping than the original, but without the correctness bug of full batching.

No changes to: the item-scan cost caveat (left as documented behavior, not a bug), or the README GIF (per your last review, tracked separately). Please don't flag the README GIF again.

armaninyow:
This one exposed a real regression on my end, so this response focuses on what actually changed/moved, not a full point-by-point (the rest still holds as originally implemented).

## What caused a regression, and how it landed

**Issues 1 and 2 (property-store cost / scroll-jump on refresh)** — my first pass took the review's own suggestion literally: switch fully to `IShellFolder2::GetDetailsEx` for both *finding* the widest item per column *and* trusting our own GDI text measurement as the *final* pixel width, eliminating `SelectAndPositionItems`/`CM_WIDTH_AUTOSIZE` and the scroll-jump problem entirely by construction.

In testing, this measurement wasn't reliable enough: our own `GetTextExtentPoint32W`-based width didn't always match what Explorer's own rendering actually produces (plausibly a GDI-vs-DirectWrite discrepancy, more visible on longer strings), causing both looser-than-necessary columns and, more seriously, real Name-column truncation on items further down long folders.

Landed on a hybrid instead:
- `GetDetailsEx` is still used for the full-folder scan (fast, avoids the per-item property-store round trip from issue 1 — that part of the fix holds).
- But it's now only used to **identify** which single item is widest per column, not to measure the final width. The final pixel width for that column still comes from scrolling that one item into view and reading back the built-in `CM_WIDTH_AUTOSIZE` result — i.e., Explorer's own rendering measures itself, which is what guarantees pixel accuracy.
- Columns are measured one at a time (not batched — the earlier round-1 fix for that stays in place).

This means issue 2's "never touches scroll" framing from my first pass didn't hold up — scrolling during a fit is back. What's properly fixed now is what the review actually asked for: real scroll-position restore via `LVM_GETTOPINDEX` captured before the queue starts and `LVM_ENSUREVISIBLE` reissued after, instead of the focused-item/item-0 guess. (My first response to this review claimed the scroll problem was solved architecturally; that was premature — it's solved now, but by the mechanism the review originally proposed, once the "avoid scrolling entirely" approach proved unreliable.)

One residual, smaller version of the review's own font-mismatch note: the initial argmax-identification scan still uses our own GDI font guess to rank items against each other. If that guess is close enough to reality, the right item still gets picked even if its absolute width isn't trusted — but a large enough font mismatch could in principle pick the wrong "widest" candidate. Given the final width no longer depends on that measurement's absolute accuracy (only relative ranking), this is a much smaller residual risk than what caused the regression, but worth flagging as not fully eliminated.

**Issue 3 (Elastic re-scans everything on resize)** — implemented as suggested and unaffected by the above back-and-forth: a per-tab cache of the last fit's non-Name column widths, consulted on a resize-only trigger to recompute just Name and skip the scan/measurement entirely. Still in place.

## Also done this round, unaffected by the regression

- `std::atomic<bool> g_unloading`, checked before every `SetTimer(AUTOFIT_TIMER_ID, ...)` call site.
- `SetWindowSubclassFromAnyThread`'s return value is now checked before recording a tab association.
- The dead `SubclassTargetIfNeeded(hwndTop, hwndTab)` call removed.
- `g_fitContexts.reset()` moved outside `g_cs`.
- Hand-rolled IIDs replaced with `IID_PPV_ARGS` throughout.
- `hwndHeader` cached on `FitContext` instead of re-found via `EnumChildWindows` every verify tick.
- `FindTabWindow` uses `GetAncestor(GA_PARENT)` instead of `GetParent`.
- Duplicate `GetCachedSettings()` call in the `isRefresh` path consolidated.
- `GetScrollRange` replaced with `GetScrollInfo(SIF_RANGE | SIF_PAGE)` for a real overflow amount in the Elastic verify loop.
- `maxScanItems` default raised (100 → 500) and clamped at a ceiling (5000).

## Not touched this round

- The "calibration phase is effectively dead code" note is moot — that whole phase no longer exists; it's replaced by the native-autosize measurement described above.
- The shell-notification-re-fits-on-every-file-event functionality note wasn't revisited further this round (it was already narrowed by dropping `SHCNE_UPDATEITEM` in an earlier round).
- README documentation, per your original note on the earlier round, stays out of scope here too.

armaninyow:
Two of the changes I made caused real regressions found in testing; this response covers those plus one more that surfaced during the same testing pass and wasn't part of your review but is worth flagging. The rest of this round held up cleanly and is summarized briefly at the end.

## Regressions caused by this round changes, now fixed

**Optional item — dropping the ReBar/Toolbar subclasses and WM_COMMAND/F5 detection broke F5, Ctrl+R, and right-click Refresh.** I took the optional suggestion ("now that both Refresh hooks exist, these look redundant") at face value and removed them. In testing, only the toolbar refresh button kept working — F5, Ctrl+R, and the context-menu Refresh stopped triggering a fit entirely. Whatever `CDefView::Refresh`/`CBrowserHost::Refresh` cover on this build, it isn't all of those paths. Reverted: `ReBarWindow32`/`ToolbarWindow32` are back in `kSubclassTargets`, and the `WM_COMMAND` refresh-id / `VK_F5` detection is back in `ExplorerSubclassProc`, running alongside the two Refresh hooks rather than instead of them.

**Functionality note 1 — the raw-value-vs-display-text fix broke Elastic (and Full) on folders with multi-value columns.** The suggested fix was `PSFormatForDisplay(keys[c], propvar, ...)`, which takes a `PROPVARIANT`. `GetDetailsEx` returns a `VARIANT`, so I converted between them with `reinterpret_cast`. That's only safe for scalar values — a multi-value property (e.g. Contributing Artists) comes back as a `VARIANT` `SAFEARRAY`, which `PROPVARIANT` represents completely differently, so the cast was reading garbage for any such column and broke the whole scan for folders containing one. Windhawk's `propvarutil.h` also doesn't expose a real `VARIANT`→`PROPVARIANT` converter (`PropVariantFromVariant` isn't declared, only the reverse), so this couldn't be fixed by "convert properly" as first attempted either. Landed on: scalars still go through `PSFormatForDisplay` via the same cast (safe for those types), but `VT_ARRAY|VT_BSTR` values are now walked and joined manually via `SafeArrayGetElement` before measuring — no cast, no PROPVARIANT involved for that case at all.

## A related regression found in the same testing pass, not from this review

While fixing the above, testing surfaced that folders with genuinely slow property columns (audio metadata specifically — tag reads aren't cached the way filesystem attributes are) were hitting main issue 3's new 200ms scan budget almost immediately and silently falling back to plain `Visible Rows Only` behavior, which has no Elastic-specific Name-stretch logic at all. That looked identical to "Full/Elastic don't work," for reasons unrelated to the two regressions above. Raised the budget to 2000ms — still bounded (the point of issue 3 stands), just far more realistic for cases where `GetDetailsEx` means actually parsing file contents rather than reading cached metadata.

One more from the same pass, self-inflicted by a functionality-note fix rather than by review-following: making empty columns fall back to `CM_WIDTH_AUTOSIZE` (functionality note 4, "don't truncate empty column headers") turned out to conflict with Elastic mode's math, which needs every non-Name column's width known *in advance* to compute Name's leftover space correctly — handing an empty column off to "Explorer decides later" meant its real autosize width could exceed what was assumed, overflowing the row and producing a horizontal scrollbar. Empty columns now get a real, fixed width (header text + a small allowance) in Elastic mode specifically, applied literally like every other column; Full mode still uses `CM_WIDTH_AUTOSIZE` for them since nothing there depends on knowing the width ahead of time.

## Everything else

Applied cleanly, no regressions found: the `IShellView::Release()`-under-lock fix, the dropped resize-only hint on retry, `SIZE_MINIMIZED` filtering, the `GetCachedSettings()`-per-message reduction, the font-probe simplification, `delay` clamping, `lEvent`/`GetClassNameW` cleanup, the two-step scroll-restore trick, and Elastic identifying the Name column by `PKEY_ItemNameDisplay` instead of by index (with the resize cache restructured to match). Deliberately not applied: dropping the UI Automation fallback, since I have direct evidence from earlier testing this session that `hwndListView` genuinely comes back null on at least one real Explorer build. README documentation, per your last review's own suggestion that the time-budget fix would make it unnecessary, wasn't done either way.

armaninyow:
PLEASE do not flag the README GIF again

armaninyow:
Deviation: kept kScanTimeBudgetMs at 2000ms instead of lowering it

The suggestion to lower the scan time budget to 150–250ms (now that the deadline check is per-column instead of per-32-items) caused a real regression, caught via user testing.

What broke: on a folder with slow-to-read metadata columns (an audio folder with Title / Contributing artists / Album), a 200ms budget was exhausted almost immediately — often within the first item — because each GetDetailsEx call for those columns is a genuinely slow round trip, not just numerous. The scan aborted almost instantly and silently fell back to plain built-in autosize for the whole folder, which meant: the Name column stayed truncated (limited to on-screen rows), and Elastic mode didn't run at all (the fallback path doesn't branch on elasticMode).

Why the original reasoning didn't hold: making the deadline check per-column (instead of per-32-items) only bounds overshoot — how far past the budget a slow run can go before it's caught. It does nothing to speed up the actual per-call latency, so a small budget still gets exhausted almost instantly on a folder where those calls are inherently slow. This is the same failure mode already documented from an earlier round (budget started at 200ms, broke a real music folder, was raised to 2000ms) — lowering it again reintroduced it.

Fix: the per-column deadline check is kept (it's a real improvement, tightening how precisely the budget is enforced), but kScanTimeBudgetMs stays at 2000ms rather than being lowered.

Everything else from this round (per-column deadline check itself, UIActivate folder-PIDL short-circuit, in-flight fit cancellation on view swap, IsEqualPropertyKey usage, FindNameColumnIndex sentinel, SHCNRF_InterruptLevel, the Elastic large-folder name-stretch path, and the settle-delay self-correcting re-read) was implemented as agreed and has no other reported issues.

armaninyow:
PLEASE do not flag the README GIF again

armaninyow:
Deviation: restored the UI Automation fallback after removing it caused a real regression

Main Issue 4 suggested removing the UI Automation fallback (GetThreadAutomation / FindListElementViaUIA / ElementHasHorizontalScroll / pElasticListElement), on the reasoning that Explorer's Details view always hosts a classic SysListView32, making the fallback unreachable. This was the third round in a row this fallback was flagged as likely dead code, and it was removed on that basis.

What broke: on a real test machine, Explorer's Details view does not expose a classic SysListView32 -- it hosts a DirectUIHWND instead, exactly the scenario this fallback exists for. With the fallback removed, Elastic mode's Step_VerifyElasticFit had no way left to detect horizontal overflow (hwndListView and hwndHeader are both null in that case, so the header-rect check and GetScrollInfo(SB_HORZ) check both silently no-op), so the Name column was never shrunk when it overflowed. Result: a persistently visible horizontal scrollbar in Elastic mode.

Why the removal reasoning didn't hold: the "SysListView32 always exists" premise was review speculation, repeated across multiple rounds without ever being confirmed against a real machine. It happened not to hold on the machine actually running the mod. This is the same failure pattern as the round 5 kScanTimeBudgetMs regression: a suggestion repeated confidently enough, across enough rounds, that it started to look verified when it never was.

Fix: the UI Automation fallback is fully restored, byte-for-byte equivalent to its pre-round-8 behavior. A comment at its definition now states plainly that removing it previously caused a confirmed, real-world regression, and that it should not be removed again without testing on a machine that lacks SysListView32.

Everything else from round 8 (per-chunk scan block cap, change-notification trigger kind skipping the exact-measurement scroll queue, Wh_ModUninit actually releasing leaked shell views/notify registrations, the g_settingsCs→atomics change, and the settings-description corrections) was implemented as agreed and has no other reported issues.

armaninyow:
Everything in Main Issues 1–3 was implemented as suggested (Wh_ModUninit now waits for a deferred FitContext destroy before resetting and never deletes the critical section; minimize/restore re-fits in every mode; Wh_ModAfterInit enumerates already-open windows on enable). The items below were deliberately not implemented, or only partially, with reasoning.

Declined: replacing the UI Automation fallback with arithmetic

The functionality notes suggested FindListElementViaUIA / ElementHasHorizontalScroll could be replaced with a sum(widths) - viewportWidth calculation, avoiding the UIA client entirely.

This fallback exists specifically for the case where there is no classic SysListView32 (a DirectUIHWND-based Details view instead). It was removed in an earlier round on very similar reasoning -- that it was unreachable/replaceable -- and that removal caused a confirmed, user-reported regression: a persistently visible horizontal scrollbar in Elastic mode, because nothing was left to detect overflow on that layout. It was restored and is now commented in place as load-bearing.

Given that history, this suggestion is not being acted on without a chance to verify it doesn't reintroduce the same failure mode. The arithmetic approach may well be correct, but "avoids needing UIA" was also the reasoning the last time removing this code broke a real user's setup, so it needs to be tested against a no-SysListView32 machine before it goes in, not assumed safe because the math looks equivalent on paper.

Declined: README disclosure of persisted column widths

Main Issue 4 asked for the README to disclose that column widths (set via IColumnManager) persist in Explorer's own view-state data after the mod is disabled, since Explorer's per-folder shell bags are independent of the mod.

On reflection this was judged not worth adding: once the mod is disabled, Explorer also stops receiving WM_SIZE handling from this mod, so Elastic's Name column simply stops being elastic and behaves like an ordinary fixed-width column from that point on -- the "surprising persistence" this disclosure would warn about doesn't really manifest as a distinct behavior worth calling out; it's the same as any other column width Explorer already remembers per folder. Not added.

Deferred: possibly-redundant trigger paths (WM_KEYDOWN/F5, ReBarWindow32/ToolbarWindow32 subclassing, CBrowserHost::Refresh hook)

This is the fourth round raising some version of "these overlapping refresh-detection paths look redundant now that CDefView::Refresh is hooked." The reasoning each time is plausible but unverified. Removing any of them risks quietly breaking a refresh trigger (F5, toolbar button, or window-resize command routing) that isn't covered by the remaining paths on some Explorer build. Left in place until actually testable rather than removed on repeated but still-unconfirmed suspicion.

Deferred: settle-delay before reading back Elastic's fallback autosize widths

The exact-measurement path already waits one tick before reading an autosized column back, because the view hasn't always finished relayout yet. ApplyFastAutosizeFit (used for the change-notification and over-maxScanItems fallbacks) and Step_AbortScan's elastic branch read the result back synchronously instead, so Name can occasionally be computed from a pre-relayout width in exactly the paths that are the common case for Elastic mode.

Fixing this properly means converting ApplyFastAutosizeFit into a multi-tick, timer-based flow the same way the exact-measurement path already is, rather than a small patch -- a more invasive change than the rest of this round. Not implemented yet; flagged for a future round rather than rushed in alongside everything else here.

Meteoni:
Hey there! I tested this mod and I have a few thoughts that I want to share about Elastic Mode.  

In my testing, the delay was noticeable enough that you can see every column moving for one or two seconds after every navigation. I doubt if that'd be an desirable outcome for everyday users. Window's native column auto-fit is already quite slow: it scans for every item in the folder. Elastic mode's current approach certainly doesn't help with that. 

I've encountered similar issues before when working on a fork. At the end I decided that it was better to drop the scan entirely, which almost wiped out the delay. Most of all, there's good reason for it too: in an elastic presentation mode, knowing the actual names for every item in a folder is almost irrelevant to the final layout. What might play a role instead is the property type of a specific column - for example, how much width is needed to display a time column in full, versus how much width is needed for a file size column - and it's much lighter than a full directory scan. Skipping it doesn't really hurt much here. Maybe a similar approach can be considered for this update. 

Just a few thoughts. Not sure if it'd be helpful. 

[my previous fork here](https://github.com/user-attachments/files/32437467/example.txt)

armaninyow:
Hey @Meteoni ! By the way, I did see your issue posted a month ago, and I thought of responding after 1.1.0 being released. But as you can see, I did not base it on your fork. I do appreciate you doing this. And yeah, the column moving during Elastic fits is a real thing, it's even called out in the README's known limitations section already.

The reason it scans at all is that Elastic (like the Scan Entire Folder) is meant to be exact. Every column, including whichever one is stretching, gets sized off real content, not a guess. So nothing should ever get clipped. Skipping the scan and going off schema type and the header text like you're describing would be way faster, no argument there, but it gives up that guarantee. Like different dates/type names/size values, some could just get truncated since it never actually looks at what's in the folder.

Again I really appreciate you bringing an actual alternative instead of just "this feels slow", makes it way easier to think through. But I don't want to just swap it in here, it feels like it deserves its own thing. Maybe as a separate faster Elastic option rather than replacing what's there, so people can pick whichever they'd rather have. I'll open an issue to track it properly.

PR #5502, opened 2026-09-14
Add files via upload

Martin-Kaiser-0:
## Changelog

* Changelog item 1...
* Changelog item 2...

## Mod authorship

This mod was created by:

- - [x] Claude
- - [ ] ChatGPT
- - [ ] Gemini


Martin-Kaiser-0:
Thanks for the review. Changes pushed:

**1. Speculative compatibility paths**
- The pre-3.5 fallback (GetKeyState hook, SimulateCtrlTab, thread_locals) was not tested against real 3.1-3.4 builds, so I removed it. The mod now requires 3.5+ and returns FALSE from Wh_ModInit on older versions.
- I kept the file version check (GetProcessFileVersion / -lversion) on purpose. It isn't part of the legacy path but a safeguard: in 3.4.x, Ctrl+PageDown / Ctrl+PageUp are bound to CmdGoToNextPage / CmdGoToPrevPage (src/Accelerators.cpp), so without the check the detection would resolve those ids and a wheel notch over the tab bar would flip pages instead of tabs. With the check the mod simply stays inactive there.
- `SumatraTabsCtrlClass`: [Variant A: I tested with the current pre-release build (SumatraPDF-prerel-64.exe, 3.7.x); tab switching works there as well.] [Variant B: not runtime-tested. The class name is taken from the current source (kTabsCtrlClassName in src/gui/win/TabsCtrl.cpp), and the tab control there still falls through to DefWindowProc for unhandled messages, so the wheel reaches the frame the same way. If you'd rather not ship it untested, I'll drop it.]
- Tested with SumatraPDF [version] ([installed/portable], 64-bit), both with the regular tab bar and with tabs in the title bar.

**2. `commandIds` setting** - removed. When detection fails, the mod now logs the Ctrl-modified accelerator entries of the table, as suggested.

**3. GIF** - [added to the README.]

**Optional improvements**
- Switched to WindhawkUtils::SetWindowSubclassFromAnyThread / RemoveWindowSubclassFromAnyThread and WindhawkUtils::SetFunctionHook.
- Detection state is guarded by a mutex; g_tabCommandsResolved is an atomic flag used as the fast path, so the counter can no longer overrun the array.
- Once the cache of scanned tables is full, the TranslateAcceleratorW path stops scanning. Tables created afterwards are still handled by the CreateAcceleratorTableW hook, which scans the raw ACCEL array and doesn't use the handle cache - that also covers the recycled-HACCEL case.

**Functionality notes**
- README now states that the mod relies on the default Ctrl+PageDown / Ctrl+PageUp bindings.
- reverseScrollingDirection now applies to horizontal scrolling too; the setting description was updated.
- Wheel events are no longer swallowed while the command ids are unknown; they're left to SumatraPDF.
- The delta remainder is now kept per wheel orientation.
- FindVisibleTabBar per event: left as is. Only wheel events that reach the frame trigger it and the walk is cheap; a cached HWND would need invalidation when the tab bar is reparented into the title bar.
- Wrap-around: that's SumatraPDF's behavior (TabsOnCtrlTab uses modulo); the README now attributes it to SumatraPDF. The legacy path is gone.

/ai-review

Martin-Kaiser-0:
Thanks, good catch on the latched ids. Changes pushed:

**Command ids latched forever** - fixed as suggested: `CreateAcceleratorTableWHook` now re-resolves on every table creation (only a table containing both shortcuts updates the ids, the reduced tables leave them alone) and clears the scanned-handle cache, which also covers the recycled-HACCEL case. Since the ids can now change at runtime, the wheel handler reads them under the mutex.

**Optional improvements**
- `FindVisibleTabBar` per wheel event: I didn't switch to `WindowFromPoint` alone, because with tabs in the title bar the area next to the tabs hit-tests as HTTRANSPARENT (so the window can be dragged there), and `WindowFromPoint` doesn't return the tab bar for that area - scrolling over the empty part of the tab bar would stop working there. Instead `WindowFromPoint` is now a cheap pre-check: the window under the cursor must belong to this frame (this also fixes wheel events arriving at the focused frame while another window overlaps the tab bar, with the legacy "scroll the focused window" setting), and if it's the canvas the event is a plain document scroll and the child-window walk is skipped. The rect test stays for the remaining events.
- Diagnostic dump: now logs only the PageUp/PageDown entries of a table that lacks the shortcuts, plus the entry count.
- Scanned-table cache: replaced the fixed array and the give-up branch with a `std::vector` that is cleared on every table creation.

**Functionality notes**
- Single tab: when `TCM_GETITEMCOUNT` returns exactly 1, the event is now left to SumatraPDF, so the document scrolls as before. The pre-release class returns 0 there and keeps the previous behavior, as you noted.
- Pinning the command ids via settings: intentionally not re-added, since the first review pointed out that users have no way to find the values. The README documents the reliance on the default bindings.

Tested with SumatraPDF [version] ([installed/portable], 64-bit) with the default
UseTabs = true, i.e. the tab bar in SumatraPDF's own title bar: tab switching,
scrolling over the empty part of the tab bar, single-tab pass-through and
reverse direction. The plain tab bar below the menu (UseTabs = false with more
than one tab in a window) can't be reached in normal use, so that path is
verified against the source only.

/ai-review

Martin-Kaiser-0:
Thanks, both points are valid. Changes pushed:

**1. `TCM_GETITEMCOUNT` sent to a non-tab-control class** - `FindVisibleTabBar` now reports whether the match is the standard `SysTabControl32`, and only that control is queried, with `<= 1` instead of `== 1`. The pre-release class is never sent the message.

**2. Stale ids after a table rebuild** - implemented as suggested: the table the ids were resolved from is remembered (`g_resolvedFromTable`), and a `DestroyAcceleratorTable` hook resets the ids to "unknown" when that table is destroyed. If the rebuilt table binds the shortcuts again, `CreateAcceleratorTableWHook` resolves them afresh; otherwise `OnMouseWheel` bails out. The destroy hook also removes the handle from the scanned-table list, which replaces the clearing that was done in the create hook.

**Optional improvements**
- `OnMouseWheel` now does the modifier-key and resolved checks before walking the child windows.
- README has a credits section naming `chrome-wheel-scroll-tabs` / m417z and the GPL-3.0 origin.
- `g_uiThreadId` is a `std::atomic<DWORD>` now.

**Functionality notes**
- Caption buttons: checked against the source - with DWM composition the caption window is sized to end at `DWMWA_CAPTION_BUTTON_BOUNDS` (`RelayoutFrame`), and the tab control is a child of the caption window that gets the width remaining after the system-menu and menu buttons (`RelayoutCaption`). So minimize/maximize/close are outside the hit area, the empty caption space between the last tab and the buttons is inside, which is intended.
- Handle-based lookup on every message: once `CreateAcceleratorTableWHook` has seen a table, every live table has been scanned at creation, so the lookup in `TranslateAcceleratorWHook` is now disabled from that point on (`g_tableCreationObserved`). It only runs in the "mod loaded into a running SumatraPDF" case until the first settings reload.

Tested with SumatraPDF [version] ([installed/portable], 64-bit) with the default UseTabs = true, i.e. the tab bar in SumatraPDF's own title bar: tab switching, scrolling over the empty part of the tab bar, single-tab pass-through, reverse direction, and a settings reload while running. The plain tab bar below the menu (UseTabs = false with more than one tab in a window) can't be reached in normal use, so that path is verified against the source only.

/ai-review

Martin-Kaiser-0:
Thanks. Pushed the optional items as a final polish:

- `g_tableCreationObserved` removed. Agreed that it added a failure mode without a real benefit; the scanned-table list keeps the handle-based path cheap on its own.
- Class-name comment now mentions both naming schemes (`FRAME_CLASS_NAME` / `CANVAS_CLASS_NAME` in 3.6.1, `kFrameClassName` / `kCanvasClassName` in the current source).
- The "not found" diagnostic is only logged while no table has resolved the ids yet. (Small note: in 3.6.1 the edit table does contain Ctrl+PageDown/PageUp, `gNotSafeKeys` there doesn't list them; the exclusion is in the current source only. The change is useful either way.)
- `g_resolvedFromTable` comment corrected: it's whichever table resolved the ids last, and all tables are destroyed together before a rebuild.
- Added an `IsWindowEnabled(hWnd)` guard in `OnMouseWheel`, so nothing is posted while a modal dialog has the main window disabled. 

Left as is: the single-tab check for `SumatraTabsCtrlClass` - that class offers no way to query the tab count, and the check covers the release versions.

/ai-review

PR #5506, opened 2026-09-14
Add Cursor Tail mod

CYoJkoY:
## Changelog

* Updated Cursor Tail with improved performance and rendering behavior.
* Organized the mod settings into clear categories and added bilingual English/Chinese labels and descriptions.

## Mod authorship

This mod was created by:

* * [ ] The submitter, without AI assistance
* * [x] The submitter, with AI assistance
* * [ ] Claude
* * [ ] ChatGPT
* * [ ] Gemini
* * [ ] Another AI (please specify):
* * [ ] Other (please specify):


CYoJkoY:
Thanks for the detailed review. Most of the findings were correct, and I've pushed fixes for them. Details below, plus the two items I deliberately left unchanged.

## Required items

**2. `exe=on` rules never produce a trail during fullscreen — fixed**

You're right, `parkSample()` before the `appRule` check made the documented override dead code: it zeroed both the delta and the elapsed ticks, so velocity was always 0 for a force-enabled app.

- `parkSample()` now only runs on the path that actually returns:
  ```cpp
  const bool fullscreenActive = UpdateFullscreenState(now, foreground, appRule > 0);
  if (fullscreenActive && appRule == 0) {
      parkSample();
      MaybeReleaseIdleBackbuffer(now);
      return false;
  }
  ```
- `UpdateFullscreenState` takes an `overrideSuppression` parameter. With `appRule > 0` the fullscreen state is still tracked (so the 500 ms exit confirmation behaves identically), but the suppression transition no longer calls `ResetTrailState()` / `HideOverlay()`, so the trail can't be permanently cleared by an app that is force-enabled.

**3. `CoversMonitor` suppressing the trail on the plain desktop — fixed, both parts**

- Added an explicit desktop/shell exclusion (`IsDesktopOrShellChrome`) that checks `GetDesktopWindow()`, `GetShellWindow()` and the class names `Progman`, `WorkerW`, `Shell_TrayWnd`, `Shell_SecondaryTrayWnd`. This covers the slideshow/Spotlight `WorkerW` case you described.
- Added the second signal from the parent mod as `IsCursorConfined()`: `CoversMonitor` now requires either a clip cursor narrower than the virtual screen or a hidden cursor, in addition to the geometry test. This removes the false positives on borderless non-game windows (maximized Electron/VS Code style windows with a custom title bar), which is what the looser geometry-only heuristic was hitting. README updated to describe the combined condition.

**4. Unconditional 30 Hz polling — fixed**

Registered raw mouse input on the overlay window (`RIDEV_INPUTSINK`, so it's delivered regardless of focus) and made the idle path event-driven:

- While a trail is active or fading the waitable timer is unchanged (125 Hz simulation, separately paced submission).
- When fully idle the timer drops to a **1 second** tick that is only a safety net — the loop is woken by `WM_INPUT` instead.
- The 30 Hz idle rate is still kept in two situations: until the idle back buffer has been released (5 s), and as a permanent fallback if `RegisterRawInputDevices` fails. Without those, a missed `WM_INPUT` (RDP, some synthetic input, a hook consuming the message) could stop the mod from ever sampling again.
- On an input wake-up the frame is throttled to the idle rate, since a high-polling-rate mouse posts far more messages per second than that.

Net effect while idle and settled: 1 wake/second instead of 30, and `SHQueryUserNotificationState` + `CoversMonitor` no longer run at 10 Hz forever.

## Optional improvements — all applied

- **Tool-mod boilerplate** replaced with a verbatim copy of the wiki snippet.
- **Fork naming** removed: `SmearFrameOverlayClass` → `CursorTailOverlayClass`, `SmearOverlay` → `CursorTailOverlay`, `SmearFrame()` → `TrailFrame()`, `g_isSmearing` → `g_trailActive`.
- **`alpha_min` / `alpha_max`** are now swapped when inverted, and `core_width_*` is clamped to no more than the matching `width_*` so the core can't swallow the outline. Both documented in the setting descriptions.
- **`alpha * 1.02f`** removed — it wasn't motivated by anything; the core brush now uses `alpha` like the other layers.
- **`GetDC(nullptr)`** → `CreateDCW(L"DISPLAY", nullptr, nullptr, nullptr)` (released with `DeleteDC`).
- **`GetProcessExeName`** now retries on `ERROR_INSUFFICIENT_BUFFER` into a `std::wstring`, so long paths no longer silently drop a per-app rule.
- **`ReserveForTailLength`** reserves `baseCount * 16`, the exact bound after four Chaikin passes.
- **Clamp ranges** documented for `tail_length` (2-64), `trigger_velocity` / `stop_velocity` (50-20000), `tail_offset_x/y` (-64 to 64), `width_*` (1-40 / 1-60), `fade_decay` (50-99), `glow_width_factor` (10-30), `auto_resample_interval` (0-10000).
- **`WhTool_ModSettingsChanged`** no longer causes a double reload: the `WndProc` case is just `return 0` and only wakes the wait; the flag is set in `WhTool_ModSettingsChanged`.
- **`@version`** reset to `1.0` for a first submission.

## Functionality notes

- **`SHQueryUserNotificationState` is session-wide** — documented in the README, since that's a user-visible consequence: a background D3D-exclusive app suppresses the trail everywhere.
- **`CheckAppRuleCached` keyed on `HWND`** — now keyed on `HWND` + PID, so a recycled handle backed by a different process no longer inherits the cached verdict.

Two I did **not** change:

- **DPI scaling.** Correct observation, but it's a visual behavior change rather than a bug, and it would change the configured look for everyone who already tuned the widths. Doing it properly also means keeping the history in one coordinate space and scaling per monitor, which is a meaningful chunk of the render path — I'd rather it land as its own change than inside a review fix round. Happy to open an issue and do it next if you want it.
- **Auto color and screen capture.** Kept as-is. The overlay must be excluded from capture while Auto mode is sampling the screen background, otherwise the luminance sample picks up the trail's own pixels; hiding the window for the duration of the blit instead would make a visible trail blink on every resample (up to 20/s). It's documented in the README, and Manual mode keeps the overlay capturable. A separate "hide from capture" setting is a reasonable follow-up if users ask for it.

Regarding the "trail freezes in place while fading" note — that one is intentional: the ribbon detaching from the cursor is what makes the fade-out read as decay rather than as a trail that keeps following a stationary pointer.

PR #5510, opened 2026-09-15
Update taskbar-ai-quota: fix newer taskbars and quota reporting (v1.6.5)

Cleroth:
- Fix quota bars not appearing on newer Windows 11 taskbars.
- Fix Anthropic extra-usage amounts and unlimited detection.
- Fix OpenAI credits-only quota reporting.


Cleroth:
Keeping v1.6.3 unchanged after checking the optional suggestions:

- **StackPanel probe:** Both `trayFrame.try_as<StackPanel>()` and `trayPanel.try_as<StackPanel>()` still query the same object's interface; changing the source interface does not remove that query.
- **Tooltip assignment:** `extraLines` is cleared before this point and has no other producer in the OpenAI parser. Appending would only be needed if more lines are added later.
- **Notification class:** Its callback is `DefWindowProcW` in user32, not a mod callback that can dangle after unload. Deferring the suggested class-registration change.
- **Wrapped credits-only responses:** This is an existing parsing limitation, but no actual provider response demonstrating that shape has been identified here. Deferring broader wrapper detection rather than expanding compatibility speculatively.
- **Anthropic unlimited detection:** Requiring explicit JSON null is intentional; missing or malformed limits should not imply unlimited usage. Additional diagnostic logging is deferred.
- **StackPanel orientation:** The supported system tray is horizontal. Deferring an extra orientation guard for hypothetical layouts.
- **Derived spend precision:** Reconstruction is only a fallback when the reported amount is missing and necessarily inherits the percentage's precision. Deferring an approximate-value marker as UI polish.

Cleroth:
Corrected the release versioning: the StackPanel fix remains v1.6.3, the Anthropic fix is v1.6.4, and the OpenAI fix is v1.6.5. This PR now targets v1.6.5.

Compared with the previously reviewed head, the final source differs only in the version metadata; runtime behavior is unchanged. The earlier explanations for deferring the optional suggestions still apply. A fresh AI review is being requested because the commit hashes changed.

PR #5517, opened 2026-09-15
Add files via upload

lzxujun:
This is a new mod: Taskbar appearance tuner.

Adjusts Windows 11 taskbar brightness and opacity, separately for the background (via a composition color-matrix effect over a live backdrop brush) and for the icons/text (via a layout-neutral composition overlay).
Also provides toggles for the taskbar top line and the Windows drag grip.

## Changelog

N/A - this is a new mod submission.

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini
- - [x] Other (please specify): WorkBuddy


m417z:
I agree with item 1 of the review: The mod is called "Taskbar appearance tuner", which makes it sound like a generic appearance tuner, but instead it applies a couple of very specific changes, some of which can already be achieved in other ways.

Also, it's missing a screenshot, and from my tests it's somewhat buggy, e.g. brightness doesn't return to the original appearance when returning it to 0.

m417z:
Update the mod id too.

And it still is buggy. For example, enabling "Show the taskbar top line", then disabling it doesn't restore it. This mod needs more work.

PR #5523, opened 2026-09-15
Add Taskbar Separators Dynamic Centering mod

Suioio:
Adds a fork of the existing Taskbar Separators mod by digART: Taskbar Separators Dynamic Centering.

What this fork adds
Dynamic centering: the icon group after a chosen position is treated as one block and centered against the whole taskbar, instead of being placed immediately after the previous group.
A resident registry-change watcher (RegNotifyChangeKeyValue) instead of polling. It watches in both directions: switching the taskbar away from left alignment turns dynamic centering off, and switching it back to left turns it on again. Switching the alignment slides the whole button row, so the mod repeats the rebuild across that animation; the switch therefore takes effect on its own, usually in under a second, with no interaction, no settings change and no mod reload.
The dividers are only the mechanical means of the centering, so they exist only while dynamic centering is in effect. While it is not in effect the mod creates no dividers at all.
The readme embeds a short GIF of the icon group re-centering as apps open and close, since the mod draws nothing by default.
When dynamic centering applies
There is no separate on/off switch. The mod only reads the taskbar alignment (HKCU\Software\Microsoft\Windows\CurrentVersion\Explorer\Advanced\TaskbarAl): when that value is 0 (left-aligned), the taskbar is horizontal and not mirrored (right-to-left), dynamic centering is active; otherwise it is not applied at all, and no dividers exist. The mod never writes TaskbarAl and never changes what Explorer reads, so nothing is left behind when the mod is disabled or unloaded.

Why the dividers are invisible by default
Opacity defaults to 0 and Divider gap defaults to 0, so with the default settings a divider neither draws nor reserves any space; the only visible result is the centering gap. That is intentional: the divider is the mechanism, not the feature. Raise Opacity to make the dividers themselves visible.

Relationship to the existing mod
This is a fork, not a replacement: it changes the position and spacing logic of the icon group, which is a different visible result from grouping icons with dividers. It must not be enabled together with taskbar-separators, since both rewrite taskbar margins. This is documented in the readme. Upstream has been asked whether they would rather have dynamic centering as an option instead: digart11/taskbar-separators#3 (open, no reply so far). If the maintainer prefers an option on the original mod, or a small standalone mod without the inherited separator engine, we will do that work rather than ask for a second copy to be accepted.

Attribution
Based on Taskbar Separators by digART (GPL-3.0), which in turn builds on work by m417z and mileso. The original copyright notices and the GPL-3.0 license are kept.

Changelog

Initial release.
Mod authorship

This mod was created by:

 The submitter, without AI assistance
 The submitter, with AI assistance
 Claude
 ChatGPT
 Gemini
 Another AI (please specify): DeepSeek V4.1 (DeepSeek Harness)
 Other (please specify):


Suioio:
Thanks for the detailed review — several of these found real problems. All fixes below are in commit a923e4ec, and the PR description has been updated to match.

## Summary of this revision

* The mod no longer writes anything to the registry: `RegSetValueExW` / `RegCreateKeyExW` have **0** call sites. The only registry access left is a single `RegQueryValueExW` read of `TaskbarAl`.
* The `Dynamic centering` Off/Auto/On setting is gone (settings: 19 → 18). There is no "separators only, no centering" use for it here, because the separators themselves draw nothing by default.
* Element names are now mod-unique.
* The alignment monitor can no longer outlive the mod image.
* Alignment work is gated on the process that hosts the taskbar view.
* The watcher no longer disables the rest of the mod.
* README and PR description updated for items 8, 9 and 10.

---

## 1. Fork of `taskbar-separators` — please extend that mod instead

I agree with the factual part: this is a fork, and your diff number is right. Measured against the catalog's `taskbar-separators` v1.3.0 (4,725 lines), **4,561 of its lines are unchanged**; this file adds **837** lines and rewrites **158 → 302** lines.

What the fork adds is one layout capability rather than a rename — and it is the part that does not exist upstream:

* the icon group after a chosen position is split off and centered against the whole taskbar instead of being placed immediately after the previous group;
* the dynamic gap is frozen while the user drags/reorders buttons, so the layout does not feed back into itself;
* the group is rebuilt over several composition frames after the drag is released;
* the reserved gap is rebuilt immediately when an app is closed.

**On the process point:** I opened an issue upstream before submitting here — [digart11/taskbar-separators#3](https://github.com/digart11/taskbar-separators/issues/3), *"Offer: dynamic centering as an option for this mod?"*. It is still open with no reply. Upstream has taken community contributions before ([PR #2 by mileso](https://github.com/digart11/taskbar-separators/pull/2), merged). If the maintainers would rather see this as an option on `taskbar-separators` than as a separate mod, I will do that instead: the dynamic-centering work is self-contained, and I am happy to open that PR, or to hand the code to the upstream author. I would rather be told which of the two you want than have this sit in the catalog as a duplicate.

**Element-name conflict — fixed.** The names were indeed copied verbatim from the original. They are now built as `kSeparatorElementPrefix + L"_" WH_MOD_ID` (5 construction sites), i.e. `WindhawkTaskbarSeparatorOverlay_<mod id>` and `WindhawkTaskbarSeparator_<mod id>_<n>`, exactly the pattern you suggested. Two mods can no longer adopt, mutate or delete each other's overlay canvas or separator visuals, and unload only removes its own elements.

## 2. Writing `TaskbarAl` is a persistent system change

**Fixed — the mod no longer writes the alignment at all.**
`WriteTaskbarAlValue` and `EnsureTaskbarAlignLeft` are gone; `RegSetValueExW` and `RegCreateKeyExW` have no call sites in the file. Because nothing is written, the failure mode you described (archived value lost after an explorer crash or a reboot, leaving `TaskbarAl` stuck at 0) cannot happen any more.

I deliberately did **not** adopt the `RegGetValueW` hook from `taskbar-multirow`, and I want to be explicit about why, because it is a real fork in the design. That hook exists to *force* left alignment. This mod's goal is the opposite one: dynamic centering is only meaningful when the user has **already** chosen left alignment, and when they choose center alignment the mod must get out of the way rather than fight them. Overriding the read would silently change the user's taskbar alignment — the same class of problem this item is about, only in-process instead of persistent. The mod now reads `TaskbarAl` and, if it is not 0, simply does not apply dynamic centering.

## 3. The alignment watcher permanently disables the whole mod

**Fixed exactly as you suggested.** `YieldDynamicCenteringForUserAlignment()` now clears `g_dynamicCenteringActive` and runs one reconcile so the reserved gap goes away, and it does **not** touch `g_unloading`. The separators keep working, nothing is torn down, and `g_unloading` is no longer reused as a feature kill switch. The README no longer contains the "change the setting to load it again" instruction, because there is no such setting any more.

## 4. The alignment monitor thread can outlive the mod image

**Fixed.** `StopTaskbarAlignMonitor` signals the stop event first and then joins with `WaitForSingleObject(g_alignMonitorThread, INFINITE)`. The thread waits on that event (`WaitForMultipleObjects` over the stop event and the registry change event), so the join is guaranteed to return and cannot deadlock. The unreachable self-thread branch was removed as well.

The starvation path you identified is also gone: `RunFromWindowThread` no longer uses an unbounded cross-thread `SendMessage`. It now uses

```cpp
SendMessageTimeoutW(hWnd, runFromWindowThreadRegisteredMsg, 0,
                    reinterpret_cast<LPARAM>(param),
                    SMTO_ABORTIFHUNG | SMTO_BLOCK, 5000, &sendResult);
```

so the 5-second join can no longer be outlived by a busy Explorer UI thread.

## 5. The alignment takeover runs in every `explorer.exe`

**Fixed by gating on the taskbar-view process.** `ApplyDynamicCenteringSetting()` is called from `ModInitWithTaskbarView` and from `Wh_ModAfterInit`, in both cases only when `g_taskbarViewDllLoaded` is set. The case where `taskbar.dll` loads after the mod is covered by the `LoadLibraryExW` hook path (`HandleLoadedModuleIfTaskbarView`), which sets the flag and applies the setting there. A plain File Explorer window process no longer reads `TaskbarAl` and no longer starts a watcher thread, so several processes can no longer race on the same value.

## 6. Synchronous reconcile inside `LayoutUpdated`

**Not changed — deliberately.** This one lands directly on the layout/animation path: switching to `reorderStructuralReconcilePending` + `EnsureGeometryRenderingSubscribed` changes *when* the geometry is applied relative to composition frames. I cannot demonstrate that the visual result and the animation timing stay identical, and this mod's hard requirement is that the look and the animation must not change. It is also a "this risks" finding rather than a demonstrated failure. If a human reviewer wants it anyway, I will do it and post before/after captures so the change is verifiable.

## 7. The drag freeze can get stuck permanently

**Not changed — deliberately, for the same reason.** The one-line fix you suggest (`EnsureGeometryRenderingSubscribed` in `OnReorderPointerPressed`) changes when rendering is subscribed on the input path, which is exactly the path where I cannot prove the animation is unaffected. I would rather leave a known-theoretical stuck-state than risk a visible regression in the drag animation that the feature exists for.

## 8. The default `opacity: 0` makes every separator invisible

**Kept at 0, and now documented.** This is deliberate: the feature is "reserve the space, draw nothing". The mod's setting default, the `$description` (English and zh-CN) and the README's settings table now all say so explicitly — `"The default 0 draws nothing on purpose, so what stays visible is the space the divider reserves; raise this value to make the dividers visible."` With the switch removed there is no longer any state in which "invisible separators" is ambiguous.

## 9. `Dynamic centering position` silently does nothing without a matching `Separators` entry

**Fixed in the documentation, as you offered.** The `$description` of `dynamicPosition` (English and zh-CN) and the README both state that the position must also appear in `Separators`, otherwise no gap is reserved for it.

## 10. Dynamic centering is horizontal-only

**Fixed in the documentation.** The README's compatibility section now says that vertical taskbars do not support dynamic centering (the dividers still work; only the reserved gap is horizontal-only arithmetic).

---

## Optional improvements

| Item | Action |
|---|---|
| `Taskbar Separators Dynamic Centering: ` log prefix | Removed. |
| `ParseDynamicCenteringMode` accepting `true/false/0/1/yes/no` | Removed entirely, together with the setting it parsed. New mod, no legacy builds, so there was nothing to stay compatible with. |
| Dead state `dynamicLeftButton` / `dynamicRightButton` / `draggedButtonWidth` | Removed. |
| `ReconcileResult::noValidSeparators` unreachable | Removed. |
| `StopTaskbarAlignMonitor` self-thread branch unreachable | Removed. |
| `CalculateDynamicCenteredGap` hot path: `GetSettingsSnapshot` deep copy + visual-tree walk | **Not changed** — geometry path. |
| `RegNotifyChangeKeyValue` subtree flag `TRUE` | Changed to `FALSE`. |
| Press-time forced reconcile on every pointer press | **Not changed** — input path. |
| Indentation of the new blocks | Normalized to clang-format. |

## Functionality notes

The first three (`fullGap < 1.0` fallback latching a stale gap, `Hide when middle icon count…` doing more than its name says, and `LayoutUpdated` comparing counts instead of button identities) all sit on the geometry/layout path, so I have left them as they are — same reasoning as items 6 and 7. The missing visual for dynamic centering is a fair point and a media problem rather than a code one; I will add a short GIF when I have one.

## Note on the version

The `Dynamic centering` Off/Auto/On tri-state you flagged in item 2 is gone as a setting rather than reduced to two values, because there is no use case for "separators only" in this mod: with `opacity: 0` the separators have no visual output of their own. Settings are now 19 → 18, with all other top-level keys unchanged, so existing saved settings do not shift.

I will request a re-review in a separate comment.

Suioio:
Every point is addressed below; the revision is pushed on top of the branch.

**First, a correction to my previous reply.** I previously described replacing the
unbounded cross-thread `SendMessage` with `SendMessageTimeoutW(..., 5000)` as
strictly safer. That was wrong: the timeout left `procParam` pointing at a stack
local in the caller, so a message delivered after the timeout wrote through a
dangling pointer. Item 3 is correct, and items 3 and 4 share that root cause.

### 1. Fork vs. extending the original mod

The divider in this mod is not a feature; it is the mechanism. The visible result
here is "the icon group after `dynamicPosition` is centered as one block". The
original mod's visible result is "dividers group the icons in place". Those are
different results, which is why this is a separate mod rather than an option of
the original. Because the divider is only machinery, its invisibility is a
requirement rather than a default someone forgot to change: with `Opacity` 0 and
`Divider gap` 0 it neither draws nor reserves space, so the only visible result
is the centering gap.

You asked whether digART was contacted: yes. digart11/taskbar-separators#3
("Offer: dynamic centering as an option for this mod?") has been open since
2026-09-15 with no reply so far, and we are happy to have the maintainer decide.

On separating "new feature" from "fixes to the original": you are right that the
physical-pixel alignment of the split gap sits on the shared interior-divider
path, so it also applies to a plain `Divider gap` divider — upstream uses a
single unsnapped `halfGap` there. Its reason for existing is the centering case:
the centering gap is split into a left and a right half around the centered
group, and sub-pixel halves make the layout engine round differently for
even/odd button counts, which shows up as a ~0.7 DIP shift of the centered
group. If you would rather keep the plain `Divider gap` path identical to
upstream, we can restrict the snapping to the dynamic divider — say the word.

### 2. Switching the taskbar to center alignment removed everything

That behaviour is intended, and the defect was the documentation rather than the
code: with no dynamic centering there is no divider, because the divider exists
only to carry the centering gap. The old comment and readme said the opposite
("the separators keep working"), which was simply wrong; both now state the
actual behaviour. The dividers also no longer come back on the next layout pass:
the rebuild path creates no dividers while dynamic centering is not in effect,
so the cleanup in `YieldDynamicCenteringForUserAlignment()` is final until the
alignment returns to left.

### 3. and 4. Dangling `procParam`, and the skipped unload cleanup

Both are fixed by the same change, as you suggested. The cross-thread call is a
blocking `SendMessage` again — the construct the original mod and Windows 11
Taskbar Styler use. The parameter is a caller-stack local again, it lives
exactly as long as the blocking call, the hook is unhooked before the function
returns, and `kRunFromWindowThreadTimeoutMs` plus the timeout logging are gone.
The unload cleanup therefore cannot be skipped: it completes before
`Wh_ModBeforeUninit` returns.

Trade-off, stated plainly: the unload path no longer has a time bound, so a
genuinely hung taskbar UI thread would make unload wait for it. We consider that
the right trade for removing a use-after-free, and it is what the original mod
does.

### 5. Alignment monitor could be restarted after the join

`g_unloading` is now set before `StopTaskbarAlignMonitor()`, and
`StartTaskbarAlignMonitor()` / `StopTaskbarAlignMonitor()` are serialized by a
`std::mutex` that re-checks `g_unloading` inside the lock before `CreateThread`.

### 6. `OnTaskbarLayoutUpdated` could spin the taskbar UI thread

A re-entrancy guard is in place, and the forced reconcile is now bounded: the
last observed realized count is recorded, the retry counter is reset only when
that observed count changes, and at most 3 forced reconciles are issued per
observed count. The reconcile contents and its timing (same frame) are
unchanged. We chose 3 rather than 1 because a forced reconcile that is skipped
for a temporarily-not-ready tree would otherwise leave the split margins stale
until the next icon add or remove; 3 bounds the loop the same way with a small
retry budget.

### 7. The separator at `dynamicPosition` disappeared on a vertical taskbar

Fixed without touching any geometry or orientation math: when the taskbar is
vertical the divider list is emptied, so no divider is created there at all.
That follows from the same rule as (2) — a divider only exists while dynamic
centering is in effect, and dynamic centering is horizontal-only.
`CalculateDynamicCenteredGap` is unchanged, and the readme now says this
explicitly.

### 8. `opacity: 0` as a default

Kept at 0, because it is the product definition rather than an oversight: the
effect is the centering, and the divider is the invisible mechanism. We did make
the surprise smaller — the explanation is now the first thing in the readme and
the first sentence of the setting description. We also corrected a real
inaccuracy you would otherwise have hit: the old text said the divider "only
reserves its space", but `Divider gap` also defaults to 0, so with default
settings it reserves no space either.

### Optional improvements

Done:

- the asymmetric reset in `CleanupAllTaskbarsForUnload` (the pressed handler and
  the layout-monitor fields are now cleared like the released handler);
- the pixel-snapping comment now matches the code (the two halves are snapped
  independently, so their sum can differ from the total by up to half a physical
  pixel — that is deliberate);
- a `Wh_Log` when `dynamicPosition` is not in `Separators`, so it is visible
  why no gap is reserved for it.

Not done, with reasons:

- passing `const Settings&` into `CalculateDynamicCenteredGap`: it moves where
  the geometry takes its inputs from, and we have no way to demonstrate that the
  change is effect-neutral for the centered group;
- the two input-path items (freezing only when a real button is pressed, and the
  `GetAsyncKeyState` fallback): they sit on the drag and animation path this
  submission is judged on;
- renaming `hideWhenMiddleIconCountAtLeast`: the key name is what users' saved
  settings are stored under, so a rename would silently drop them.

### Functionality notes

Adopted: the alignment watch is now resident and bidirectional — leaving left
turns dynamic centering off and removes the dividers, returning to left turns it
on again and rebuilds them, with no settings change and no reload. That is also
why all "reload the mod / change a setting" wording is gone from the readme. One
consequence worth naming: the watcher thread now also runs while the taskbar is
center-aligned, since that is the only way to notice the switch back to left. It
only waits on `RegNotifyChangeKeyValue` and reads `TaskbarAl`; the mod still
performs no registry writes.

Not adopted: centering against the button area instead of the full taskbar width
(it changes what users see, and the `maxGap` clamp already keeps the group clear
of the tray); the `fullGap < 1.0` fallback and the freeze-on-any-press behaviour
(geometry and drag path, i.e. the visible contract of this submission).

### Tested

Left-aligned, horizontal taskbar: the group after `dynamicPosition` is centered
as before, no divider is visible, and dragging plus app open/close stay smooth.
Switching to center-aligned leaves no dividers, and switching back to left
re-applies the centering without touching any setting.

Suioio:
All items are addressed below, and the revision is pushed on top of the branch.

### 1. Fork vs. extending the original mod

The choice is now stated in the PR description so the maintainer can weigh in,
which is what you asked for. One correction to the "dead weight" framing,
though: with this revision's own rules the inherited settings are opt-in rather
than unreachable — raising `Opacity` or `Divider gap` while the taskbar is
left-aligned makes the dividers visible and gives them width, and the defaults
are 0 because the divider is the mechanism for the centering, not the feature.
That said, the structural point stands: a 5,995-line copy to deliver a few
hundred lines of new logic is a lot of surface to ask the maintainer to accept.

On your second suggestion, a small standalone mod that sets the margin directly:
the centering gap does need the taskbar width, the tray bounds, the orientation
and the ordering direction, so `AddDirectionalButtonGap` on the two boundary
buttons is the last step of the mechanism rather than the whole of it — but you
are right that the overlay canvas and the divider visuals are not needed to
reserve a margin. If the maintainer prefers either of your two outcomes over a
second copy, we will do the work; digart11/taskbar-separators#3 is open with no
reply so far.

### 2. Use-after-free in `OnTaskbarLayoutUpdated`

Correct, and this one is ours: it came in with the retry counter added in the
previous revision. Fixed by re-resolving as you suggested:

    if (auto* refreshed = FindTrackedTaskbarById(taskbarId)) {
        refreshed->layoutForcedReconcileActive = false;
    }

We also checked every other call site of `ReconcileTaskbarRepeater`:
`OnAnimationRendering` already re-resolves after its call, and the
`PointerPressed` and `UpdateVisualStates` paths do not touch a
`TrackedTaskbarState` pointer afterwards, so this was the only occurrence.

### 3. One-directional element-name fix

Fixed by removing the shared prefix entirely: the base token is now
`WhDynamicCenteringSeparator`, so no name this mod creates starts with the
original's `WindhawkTaskbarSeparator_` prefix and its sweep cannot claim them
any more. The overlay was already safe — the original compares that name for
equality rather than by prefix.

### 4. Default settings doing nothing, silently

You are right that this was the worst possible first impression of the
submission. `@description` and `@description:zh-CN` now end with the
requirement, in the text users actually read in the mod list and on
windhawk.net: "Requires the taskbar to be set to Left alignment."

We did not tie the plain dividers to the alignment flag, which you raised as a
possibility. In this mod a divider only exists while dynamic centering is in
effect, and on a centered taskbar the mod must not touch the margins at all —
Windows is already centering the row, and rewriting button margins there would
be fighting it. That is the product definition we explained earlier, not an
oversight.

### 5. Left-to-right geometry and the mirrored taskbar

Fixed, and hardened beyond the literal suggestion. A mirrored taskbar is now
detected and dynamic centering is not applied there, the same way it is not
applied on a vertical one.

We did not do it with `primaryOrderingDirection < 0`, though. That value comes
from `TryGetPrimaryOrderingDirection`, which returns as soon as the first
adjacent pair qualifies. While a reorder animates, the repeater already holds the
new order while the buttons are still travelling, so a single pair can read as
reversed for one frame on an ordinary left-to-right taskbar — and acting on it
would clear every divider at once, making the centering collapse and snap back,
which is visible to a plain LTR user. The test is therefore a new conservative
helper: a mirrored ordering is reported only once every adjacent pair that is
laid out far enough agrees on a negative primary movement, at least one such
pair exists, and no drag is in progress. A taskbar whose buttons are not laid
out yet is never mistaken for a mirrored one.

`CalculateDynamicCenteredGap` itself is unchanged, and both the readme and the
`Dynamic centering position` description now say that a mirrored taskbar is not
supported.

### 6. `LayoutUpdated` full child walk

Not changed, and here is why. The clean fix is the signal you named —
`ElementPrepared` / `ElementClearing`, with the reconcile deferred to the
next composition frame. This mod reaches `ItemsRepeater` through its raw ABI:
the interface is taken by GUID and `GetElementIndex` is called through a
hand-indexed vtable slot, because no projection for it is available in this
build. The same is true for `ItemsSourceView` and for those two events.
Hand-indexing one more slot is not a slow path — a wrong slot is an arbitrary
call — and we are not willing to trade a verified crash risk for a performance
optimisation. For scale: the walk is bounded by the realized `TaskListButton`
count, roughly the number of open apps, and `FrameworkElement.LayoutUpdated`
is raised by layout passes, while hover and click change visual states and do
not produce one.

If you would still rather have the precise signal, say so and we will add the
interop properly as a follow-up.

### Optional improvements

Done:

- `StopTaskbarAlignMonitor` no longer gives up when the wait fails. It retries
  instead of returning with the thread handle still set, which is what makes "a
  mod thread can never outlive the image" actually true.
- the tray frame is cached in a `winrt::weak_ref` on the tracked taskbar state,
  reused while it is still under the same root grid, and cleared with the
  taskbar. This is the one place `CalculateDynamicCenteredGap` changed; the
  geometry, the values, the clamp and the return are byte-identical.
- the `LoadSettings` warning no longer fires for a user whose dynamic position is
  covered by an app-name separator — it warns only when no app-name separator is
  configured at all.
- the capability check is hoisted, so while dynamic centering is inactive the
  app-name resolution is skipped instead of being computed and then thrown away
  by gating A.
- the over-80-column lines you flagged are wrapped. Two kinds remain on purpose:
  lines inside `CalculateDynamicCenteredGap`, which we keep byte-identical, the
  `LR"(...)"` symbol-hook literals, which cannot be split, and the metadata
  lines, which must stay single-line.

Still declined: forcing a full reconcile on any press anywhere on the taskbar
root. It is synchronous work in the input path, and the drag path is what this
submission is judged on.

### Functionality notes

- **Reading `TaskbarAl` versus what the taskbar is doing.** Documented rather
  than changed. The readme now says the mod is driven by the stored value, so a
  mod that forces left alignment by hooking the read without changing the stored
  value — taskbar-multirow is named — leaves this mod inactive. Deriving the
  answer from the layout instead would move the switch that decides whether this
  mod touches the taskbar at all, and we cannot test the layouts that would
  affect.
- **`hideWhenMiddleIconCountAtLeast`** — description updated: reaching the
  threshold stops the centering as well as hiding the divider. The key name is
  unchanged, to protect saved settings.
- **An app-name separator can become the dynamic one** — a sentence was added to
  the `Dynamic centering position` description.
- **The readme** — the Chinese section is now a summary instead of a second copy
  of the English one, with every warning and every behavioural statement kept.
  The English section is unchanged.
- **The preview image** — your point that a static image cannot show a mod that
  draws nothing is right, so the readme now embeds a short GIF of the icon group
  re-centering as apps open and close, under the existing preview image.

The mod still performs no registry writes.

### Tested

Left-aligned, horizontal taskbar: the group after the chosen position is
centered as before, no divider is visible, dragging and reordering stay smooth
with no flicker, and opening and closing apps re-centers the group. Switching to
center-aligned leaves no dividers; switching back to left re-applies the
centering without touching any setting.

Suioio:
## Handing this over for review

The latest revision (`3f5bfc61`) has **not** had an AI re-review: the PR hit the
three-reviews-per-24-hours limit and the bot declined to post another one. Every
item from the third review is addressed below, except the ones listed at the
end, which are explained rather than silently dropped.

### What changed in this revision

- **A use-after-free is fixed.** `OnTaskbarLayoutUpdated` re-resolves the
  tracked-taskbar state after the reconcile, instead of writing through a
  pointer that the reconcile can invalidate.
- **Element names are isolated.** The base token is now
  `WhDynamicCenteringSeparator`, so nothing this mod creates starts with the
  original mod's `WindhawkTaskbarSeparator_` prefix, and the original's sweep
  can no longer claim our dividers.
- **The alignment requirement is in `@description`** (and its Chinese
  counterpart), so it is visible in the mod list before install.
- **Mirrored (right-to-left) taskbars are detected** and dynamic centering is not
  applied there, as on a vertical taskbar. The test is deliberately conservative:
  a single adjacent pair reading as reversed during a reorder animation cannot
  clear every divider on an ordinary left-to-right taskbar.
- **`StopTaskbarAlignMonitor` retries** instead of returning with the thread
  handle still set.
- Tray frame cached in a weak ref; the spurious `LoadSettings` warning fixed;
  app-name resolution skipped while dynamic centering is inactive; the flagged
  over-80-column lines wrapped.
- README: the Chinese section is now a summary rather than a second copy of the
  English one; a short GIF shows the group re-centering as apps open and close;
  the descriptions for `hideWhenMiddleIconCountAtLeast` and
  `Dynamic centering position` are completed.

### Left unaddressed, with the reason

1. **`LayoutUpdated` runs a child walk on every layout pass.** The precise signal
   (`ElementPrepared` / `ElementClearing`) would need `ItemsRepeater` events, but
   this mod reaches that interface through its raw ABI — the interface is taken
   by GUID and `GetElementIndex` is called through a hand-indexed vtable slot,
   because no projection for it is available in this build. Hand-indexing another
   slot is not a slow path, it is an arbitrary call, and we are not willing to
   trade a verified crash risk for a performance optimisation. The walk is
   bounded by the realized `TaskListButton` count (roughly the number of open
   apps). Say the word and we will add the interop properly.
2. **A full reconcile on any press anywhere on the taskbar root.** That is
   synchronous work in the input path, and the drag path is what this feature is
   judged on.
3. **`Opacity` defaults to 0, and dividers exist only while dynamic centering is
   in effect.** This is the product definition, not an oversight: the divider is
   the mechanism, not the feature. With `Opacity` and `Divider gap` both at 0 it
   neither draws nor reserves space, and the visible result is the centering
   itself. Both settings can be raised to make the dividers visible.
4. **The alignment is read from the stored `TaskbarAl`**, so a mod that forces left
   alignment by hooking the read without changing the stored value
   (taskbar-multirow) leaves this mod inactive. Documented rather than changed:
   deriving the answer from the layout instead would move the switch that decides
   whether this mod touches the taskbar at all.
5. **The centering gap is computed against the full taskbar width and clamped
   clear of the tray.** That is what the visible result was built and tested
   around; changing the reference area would change what users see.

### The structural question is yours to decide

This is a fork of `taskbar-separators`. The description says so explicitly,
together with the two alternatives: an option on the original mod, or a small
standalone mod that does not carry the inherited separator engine. If you prefer
either of those over a second copy in the catalog, we will do that work.
Upstream has been asked — digart11/taskbar-separators#3, open, no reply so far.

Tested on a left-aligned, horizontal Windows 11 taskbar: the group after the
chosen position is centered as before, no divider is visible, dragging and
reordering stay smooth with no flicker, opening and closing apps re-centers the
group, switching the taskbar to center leaves no dividers, and switching back to
left re-applies the centering without touching any setting.

Suioio:
Note on the AI review. The review bot is currently under maintenance ("The AI review bot is under constructions, so no review was posted this time"), so this revision has not had an AI pass. Everything below is our own account of what changed since the last AI review, together with the items we deliberately left alone and why.

A correction to an earlier reply of mine. I once described replacing the unbounded cross-thread SendMessage with a bounded timeout as strictly safer. That was wrong: the timeout left procParam pointing at a caller stack frame. It is a blocking send again.

Addressed from the last AI review
1. Fork vs. an option on the original mod. The description now states this explicitly for you to decide, and digart11/taskbar-separators#3 is open with no reply. One correction to the "dead weight" framing: with this revision's rules the inherited settings are opt-in, not unreachable — raising Opacity or Divider gap while the taskbar is left-aligned makes the dividers visible and gives them width. The structural point stands, but note that the two reviews argued the same rule in opposite directions: one called the inherited engine dead weight because the dividers are invisible, the other observed that removing them on a centered, vertical or mirrored taskbar makes this fork strictly less capable than the mod it forks. Those two cannot both be right, so this is a call for you. If you prefer an option on the original mod, or a small standalone mod that does not carry the inherited engine, we will do that work rather than ask for a second copy to be accepted.

2. Use-after-free in OnTaskbarLayoutUpdated. Fixed as suggested: the tracked state is re-resolved with FindTrackedTaskbarById(taskbarId) after the reconcile, before the flag is cleared. We also audited every other call site of ReconcileTaskbarRepeater: OnAnimationRendering already re-resolves, and the PointerPressed / UpdateVisualStates paths do not touch a TrackedTaskbarState pointer afterwards.

3. One-directional element names. Fixed by removing the shared prefix: the base token is now WhDynamicCenteringSeparator, so nothing this mod creates starts with the original's WindhawkTaskbarSeparator_ prefix and its sweep can no longer claim our dividers. The overlay name was already safe — the original compares that one for equality rather than by prefix.

4. Default settings doing nothing, silently. @description and @description:zh-CN now end with "Requires the taskbar to be set to Left alignment", which is the text users read in the mod list. We did not tie the plain dividers to the alignment flag: a divider only exists while dynamic centering is in effect, and on a centered taskbar this mod must not touch the margins at all.

5. Mirrored (right-to-left) taskbars. Dynamic centering is not applied there, as on a vertical taskbar, and the readme says so. The test is deliberately conservative: a mirrored ordering is reported only once every adjacent pair that is laid out far enough agrees on a negative primary movement, and no drag is in progress. A single-pair test would have been wrong here — while a reorder animates the repeater already holds the new order while the buttons are still travelling, so one pair can read as reversed for a frame on an ordinary left-to-right taskbar, and acting on it would clear every divider at once and make the centering flicker.

6. The LayoutUpdated child walk. A cheap pre-filter is in place: the handler remembers panel.Children().Size() and returns immediately while it is unchanged, so the per-child cross-ABI calls only run when the child count actually changed. Two notes: it is a heuristic (a relayout that removes one child and adds another in the same pass is not noticed, and the dividers stay stale until the next one), and when a forced reconcile is issued the remembered count is reset so the bounded retries still get their frames. We did not subscribe to ElementPrepared / ElementClearing instead, because this mod reaches ItemsRepeater through its raw ABI — the interface is taken by GUID and GetElementIndex is called through a hand-indexed vtable slot, as no projection for it is available in this build — and hand-indexing another slot is not a slow path but an arbitrary call. If you would rather have the precise signal, we will add the interop properly.

The comment on layoutForcedReconcileActive is corrected. It does not suppress the LayoutUpdated raised by our own margin writes, because nothing forces a synchronous layout pass; the real bound on that feedback is kMaxLayoutForcedReconciles, three forced reconciles per observed button count.

Optional improvements
Done: StopTaskbarAlignMonitor retries instead of returning with the thread handle still set; the tray frame is cached in a winrt::weak_ref and dropped with the taskbar; the LoadSettings warning no longer fires for a user whose dynamic position is covered by an app-name separator; the app-name resolution is skipped while dynamic centering is inactive; the flagged over-80-column lines were wrapped; the hide setting's description now says that reaching the threshold also stops the centering and that 0 means never hide; the Dynamic centering position description mentions that an app-name separator resolving to that position carries the centering gap; the Chinese readme is a summary rather than a second copy of the English one; and the readme embeds a short GIF of the group re-centering as apps open and close, because the mod draws nothing by default.

Declined, with reasons: forcing a full reconcile on any press anywhere on the taskbar root (synchronous work in the input path, and the drag path is what this feature is judged on); passing const Settings& into CalculateDynamicCenteredGap (it moves where the geometry takes its inputs from, and we cannot demonstrate it is effect-neutral); and deriving the alignment from the layout instead of the stored TaskbarAl (that is the switch which decides whether this mod touches the taskbar at all).

Two defects we found ourselves, and the data
(a) Switching the taskbar back to left alignment did not take effect on its own — the centering only appeared after an interaction with the taskbar. Cause: the rebuild ran once, at the instant the alignment change starts sliding the whole button row, so the geometry was still moving and the computed gap came out collapsed; and nothing was guaranteed to rebuild afterwards, because the layout monitor early-outs while the button count is unchanged and UpdateVisualStates only runs again on an interaction. Fix: both directions now share a bounded rebuild window — the first rebuild runs immediately and up to eight more follow at 500 ms intervals, ending early on unload or on a further alignment change. The measured log shows it exactly: raw dynamic gap 0.00 on the first pass, 556.33 on the second (+0.51 s), and 556.33 on every pass after that. The centering now appears on its own in under a second, with no interaction, no setting change and no reload.

(b) A report that the gap "disappears slowly" when switching to the center. We added a diagnostic for it — per reconcile while the window is open, it logs whether centering is active, the number of active separators, and how many tracked buttons still carry a margin this mod wrote. The data refuted our first hypothesis: we expected the margins to be handed back a few buttons at a time, but the first pass already reported 0 of 8 applied margins. So the symptom is not piecemeal bookkeeping; repeating the rebuild is what removes it, and part of what reads as a slow disappearance is Windows animating the button row itself, which we do not touch. Both diagnostics are gated on the alignment window and stay silent otherwise.

Also fixed on the way: the settings block briefly failed to parse because a $description value contained ": " (colon-space), which YAML does not allow inside a plain scalar; the readme claimed the divider "only reserves its space" although Divider gap defaults to 0; the Opacity explanation is now first in both the readme and the setting description; and the mirrored-taskbar behaviour is documented in both languages.

Left unaddressed, with the reason
ElementPrepared / ElementClearing instead of the child walk — raw-ABI interop, as above.
A full reconcile on any press anywhere on the taskbar root — input path.
Opacity defaults to 0 and dividers exist only while dynamic centering is in effect — the product definition, not an oversight; both settings can be raised to make the dividers visible.
The alignment comes from the stored TaskbarAl, so a mod that forces left alignment by hooking the read (taskbar-multirow does exactly that) leaves this mod inactive. Documented rather than changed.
The centering gap is computed against the full taskbar width and clamped clear of the tray — that is the visible result this feature was built and tested around.
Tested
Left-aligned, horizontal taskbar: the group after the chosen position is centered as before, no divider is visible, dragging and reordering stay smooth with no flicker, switching to center removes the dividers immediately, switching back to left re-applies the centering on its own in under a second, and nothing needs to be touched, changed or reloaded. All contributor checks on this revision are green.

Suioio:
Closed in favor of a new, independent modification.

The AI review was correct: this second version of `taskbar-separators[[/Q]] should not be included in the catalog. We followed the second of the two suggestions made by the review and released **Taskbar Icon Group Centering** as a separate PR – a modification that only centers the icon group after a specific position, without affecting any of the existing divider functionality (no dividers, styles, per-app matching, `separators[[/Q]] list, or animation tracking). This means it's no longer a variation of the original separator mod.

The details of what was carried over from this PR, and the reasons for doing so, are outlined in the new pull request: the taskbar/XAML plumbing and the button-margin ledger (which were already present in the upstream mod), and the centering geometry itself, which was verified against the built-in centered mode.

Thank you to the reviewers for their thorough feedback. The "extend or replace, do not duplicate" rule helped to identify the correct design for this feature.

PR #5527, opened 2026-09-16
SnapSentry 0.19.5

mario0318:
## Changelog

- A screenshot no longer waits on a notification Windows never showed. If Windows reports it could not deliver the notification, SnapSentry copies the screenshot and leaves the file alone.
- A notification on screen is never pulled away and replaced with a dialog.
- An unanswered notification never deletes a screenshot.
- Saving a setting while a notification is up no longer loses that screenshot.
- The clipboard options now read 'Image', 'File', 'Path' and 'None', matching the rest of the mod.
- The location setting is now Path format, and only applies when What to copy is set to 'Path'.


mario0318:
Thanks, both findings taken.

The duplicate feature is out of this release: the setting, the readme section, the gated code, the three includes and the bcrypt link. It lands in the version where it actually works rather than shipping as a switch wired to false.

The delivery check no longer pulls a notification down. You're right that history and on-screen are different states, and I have no way to tell a banner apart from an empty notification center. So an unconfirmed notification is now left up and only means the countdown must not authorize a delete: it settles as Copy only. With no countdown it gives up after 30 seconds instead of parking for the full no-answer cap, which was the original problem. There is no second prompt in any of these paths now.

Also took the polling note, the wait runs straight to its deadline once delivery is confirmed.

Not addressed yet, and flagged rather than quietly skipped: with Do Not Disturb on, the notification is filed but no banner shows, so a confirmed delivery can still expire into the automatic action for a prompt nobody saw. I do not have a reliable way to detect that, and I would rather not guess at it in the same change. The settings-change cancellation scoping is the same story, it is on the list for the next one.

Version is 0.19.3.

mario0318:
All three taken, and the first one was a real bug, thanks.

The Failed listener was signalling the wait without recording an answer, so it settled as Keep and the screenshot never got copied, exactly as you described. It records Copy only now the same way a dismissal does, and the side flag and its unreachable branch are gone.

A settings change settles as Copy only instead of Keep, in the notification and in the dialog, so the copy still happens and only the automatic action is withheld. There is a new event in the wait so it reacts when the settings change rather than at the deadline.

Took the SHQueryUserNotificationState suggestion too. That one I could actually test: a full-screen window puts Windows in QUNS_BUSY, and with a screenshot dropped in that state the countdown returns Copy only and logs "Notifications are being held back (state=2)", where a normal run returns the automatic action.

Still untested on my side is the Failed path itself, since I cannot break notification delivery here without changing your notification settings, and the settings-save case, which needs the Windhawk settings UI. Everything else in this change has been run.

Version stays 0.19.5.

mario0318:
closing this one, everything in it went out with 0.21.1 in #5549

PR #5529, opened 2026-09-16
SnapSentry: changelog override for 0.17.2

mario0318:
Changelog override only, no mod changes.

The 0.17.2 entry published in an instruction voice ("Say that...", "Shorten...", "Fix a colon...") instead of saying what changed for the person installing the mod. This rewrites it to read like the other entries.


m417z:
> The 0.16.0 entry published the whole pull request body, including the authorship checkboxes and two discussion sections, instead of the changelog.

Where do you see that?

mario0318:
yeah youre right, 0.16.0 was me jumping to conclusions. i saw #4765 still had the whole pr body sitting in its commit message and assumed it went to the changelog, but the first version just renders as "Initial release." no matter what. so it was fine all along.

0.17.2 is the one i was really after. it went out reading like a note to myself, "Say that the watched folder can be your own", "Shorten the settings descriptions", that sort of thing, instead of what changed for someone actually using it. so i cut it down to just that one.

PR #5531, opened 2026-09-16
virtual-desktop-helper 2.5.0: per-key disable, last-used desktop and hotkey thread fixes

u2x1:
Bumps Virtual Desktop Helper to **2.5.0**.

### Changes

- **Per-key disable**: leaving `PrevDesktopKey`, `NextDesktopKey`, `LastDesktopKey` or `PinKey` blank now disables only that hotkey instead of falling back to the default key. Invalid key values are logged and the hotkey is disabled.
- **Last-used desktop**: renamed from "last visited desktop"; `SwitchToLastDesktop` now repairs stale tracking against the actual desktop manager state and ignores targets that equal the current desktop.
- **Tracking fix**: `HandleForegroundWindowChanged` treats the virtual desktop manager as authoritative instead of the foreground window, avoiding wrong tracking when the foreground event briefly refers to a window from the old desktop during a transition.
- **Z-order fix**: after moving a window to another desktop, `RaiseWindowToFrontOfNormalBand` brings it to the front of the normal (non-topmost) Z-order band without activating it.
- **Hotkey thread hardening**: the ready signal is only sent after COM init and hotkey registration complete; duplicate thread startup is refused; a settings reload is aborted if the old thread fails to exit; cleanup unregisters all hotkey IDs unconditionally.
- **Service detection**: only `-service` windhawk runs are excluded from the tool mod; `-service-start` / `-service-stop` processes are no longer excluded.

### Docs

- README/help text updated around the "last-used desktop" behavior and the per-key blank-to-disable option.

Tested locally; hotkeys, desktop switching, move-with-focus, and pin/unpin work as expected.

> All changes credited to @Meteoni .


Meteoni:
Finding 1 is not entirely valid. 

Using the hand-rolled launcher is no accident: Windhawk 2.0 ships with `windhawk_tool_mod.h`, but using that will cause the mod to not compile on 1.7.3, which is the current stable version expected for most users. 

This
```
windhawk.exe, windhawk-mod-uiaccess.exe
```
shouldn't be an issue by itself either. Windhawk 2.0 automatically chooses a correct host, and older mod targets are explicitly backward-compatible. That combination should work just fine. 

I can change it to
```
// @include         windhawk.exe
// @include         windhawk-mod.exe
// @include         windhawk-mod-uiaccess.exe
```
to satisfy the repository validation step if it matters. However, the hand-rolled snippet will stay, as it's the currently supported way to do things.

m417z:
Yes, the AI reviewer doesn't know that. I updated the wiki it refers to, it should be more informed from now on. And yes, please add `windhawk-mod.exe` which is the 2.0 equivalent of `windhawk.exe`, and makes sense since it's a valid fallback for the mod (e.g. for a non-elevated portable version).

PR #5535, opened 2026-09-16
feat: hide taskbar and system tray tooltips (1.0.1)

gilnett:
- Suppress all native taskbar hover tooltips on Windows 11

## Changelog

*(N/A - Initial new mod submission)*

## Mod authorship
This mod was created by:
- [ ] Claude
- [ ] ChatGPT
- [x] Gemini
- [ ] Other (please specify):


gilnett:
The restart prompt is kept because without restarting Explorer, pre-existing taskbar buttons keep showing tooltips.

gilnett:
step 3 : We cannot hook ToolTip::put_IsOpen in Wh_ModInit because XAML controls cannot be created outside the UI thread, so we hook it safely once on the first live instance

gilnett:
I removed the Explorer restart prompt to follow Windhawk recommendations and avoid potential UI hangs or issues with `taskkill`. A note has been added to the README instead.
I'm open to suggestions if you recommend a better way to handle this.

m417z:
Wow, that's some harsh review from Claude on code that [I originally wrote in 2023](https://github.com/ramensoftware/windhawk-mods/commit/dd94ba0d) :) it was improved to not be necessary in this mod in most cases, but generally I don't think it's as bad as the review describes it. 

I do agree with this:

> Better still: it's worth checking whether a restart is actually needed at all. The `put_IsOpen` / `IsToolTipEnabled` / `UpdateToolTipPlacementForHoverState` hooks all run at hover time, so they should take effect on already-created taskbar elements immediately; only `AddToolTipContent` is creation-time. If that's the case, drop the note too.

So if it's unnecessary, that's great and should be removed, and if it can be made unnecessary, that'd be nice. It's also better not to misuse it - if every other mod will pop up a restart prompt, it won't be pretty. On the other hand, if a restart is truly needed, it can be more convenient to have the prompt than asking the user to restart Explorer via Task Manager.

m417z:
Also, remove the mangled symbols, such as `LR"(?IsToolTipEnabled@RecommendedItemViewModel@implementation@Taskbar@winrt@@QEBA_NXZ)"`. You can use [Windhawk Symbol Helper](https://github.com/ramensoftware/windhawk-symbol-helper) to see the exact symbols in your dll/exe files.

gilnett:
I will remove all the mangled symbols (`?...`) and keep only the demangled ones from your link

Regarding the restart, during my tests on Windows 11 (though I might have made mistakes), hooking at hover time (like `put_IsOpen` or clearing content) only removed the text, leaving an empty tooltip frame popping up. Only a full Explorer restart cleanly suppressed the entire container for already existing elements.

That's why I kept the note in the README for now. But if you know of a clean way to suppress the container at runtime without a restart, I'm all ears because I can't find any better ways to do it

m417z:
I can't give an answer without researching it, but you might be able to achieve with via native hooks, hooking a function such as `CreateWindowExW`, `ShowWindow`, or `SetWindowPos` depending on how the tooltips are shown. Then you'll have to find a way to identify the window from a native window handle. One way is hooking a higher-level caller and using a thread_local variable to mark that you're inside the tooltip creation/showing function.

PR #5536, opened 2026-09-16
Add System Tray Folders mod

RisPNG:
This adds a mod that lets you group the icons in the Windows 11 "Show hidden icons" tray flyout (the "^" popup) into folders, so a long list of tray icons can be sorted into a few tidy groups.

Without the mod

Without the mod

With folders

With folders

Inside a folder

Inside the Hardware folder

## How it works

Folders are set up in the mod's settings. Each folder has a name, an optional icon, and a list of the tray icons that belong in it. Icons are picked by their app name or tooltip, and `*` / `?` wildcards can be used, so one entry like `Razer*` can cover several icons. The readme explains how to find the exact names.

- Each folder shows up as a tile in the flyout, with a small preview of the icons inside it, or the icon you chose for it.
- Clicking a folder shows its icons in place of the main grid, with a back button to return, and the flyout resizes to fit.
- Icons inside a folder are the real tray icons, so clicking, right click menus and tooltips work the same as anywhere else.
- An icon goes into the first folder that matches it, and icons that don't match any folder stay in the main grid as usual.
- Opening the flyout always starts at the main grid, and it works with the keyboard as well (arrow keys, Enter, Esc).
- Dragging icons between the taskbar and the flyout, and reordering them, works as usual. An icon moved back into the flyout returns to its folder. Which folder an icon is in is decided by the settings, not by dragging.
- Folders can be placed before or after the other icons, and empty folders are hidden unless you choose to show them.
- It follows the dark and light theme.
- Nothing changes in the flyout until at least one folder has icons set, and turning the mod off puts the flyout back to how it was.

## Mod authorship

This mod was created by:

- - [x] Claude (Opus 5 Ultracode)
- - [x] ChatGPT (GPT-6 Astra Light)
- - [ ] Gemini
- - [x] ZCode (GLM-5.3 Max)


RisPNG:
Pushed 4cf7cf4, which goes through everything from the second review.

> **1. The teardown of everything the mod plants in the shell's tree hinges on one best-effort round trip**

The thread that attaches to the overflow is recorded now, and `Wh_ModBeforeUninit` / `Wh_ModSettingsChanged` do the round trip through any window on that thread instead of looking up `TopLevelWindowForOverflowXamlIsland` or `Shell_TrayWnd` by class name. `Detach` also calls `itemsControl.UpdateLayout()` right after the shell's `ItemsSource` is put back, so the old collection and the tile containers are dropped there and then.

> **2. `Detach` can leave a stale `loadedToken`, after which the mod stops re-attaching**

Good catch, it's cleared unconditionally now.

## Optional improvements

Did all of them:

- The default config is inert now. When no folder has a pattern (and empty folders are hidden), the mod doesn't attach at all, and removing the last pattern while it's attached detaches it and gives the shell its list back.
- Names are logged in their original case and only lowercased for matching. The readme documents `folder=N` too.
- `Wh_ModSettingsChanged` applies the settings directly when there's no overflow thread yet, instead of dropping them.
- Blank folders are skipped the same way blank icon entries are. It only stops after two blanks in a row.
- The `NotificationAreaOverflow` constructor is hooked now (optional symbol), and the instance is paired with the manager from the `InitializeIfNeeded` hook that created it. I kept the field scan as a fallback though, because `InitializeIfNeeded` only runs once per Explorer process, so without it the mod would never find the flyout when it's enabled or updated while Explorer is already running. The scanned pointer still has to pass the vtable check and have a `XamlRoot`.

## Functionality notes

- Drag and drop: tested all of it and it works fine.
  - Dragging icons from the taskbar into the flyout and back
  - Reordering in the flyout
  - Dragging icons from inside a folder out to the taskbar and back again, where they go back into their folder based on the rules
  - Reordering inside a folder

  So I don't think a readme line is needed for that one.
- XAML isn't re-parsed on rebuild anymore. Each folder tile builds its preview (with 4 `Image`s) once when it's created, and rebuilds only swap the `Image.Source`s.
- Added an "App name matching unavailable" log line for when those symbols don't resolve.

Also tested the keyboard since I was at it: Esc inside a folder closes the flyout and reopening starts at the root again, arrows work inside folders, and Enter works on the back tile and on icons inside and outside folders. Both dark and light theme look right too.

/ai-review

RisPNG:
Addressed the second review in https://github.com/ramensoftware/windhawk-mods/pull/5536#issuecomment-5707008025.

/ai-review

RisPNG:
Addressed the second review in https://github.com/ramensoftware/windhawk-mods/pull/5536#issuecomment-5707008025.

/ai-review

RisPNG:
Addressed the second review in https://github.com/ramensoftware/windhawk-mods/pull/5536#issuecomment-5707008025.

/ai-review

RisPNG:
Addressed the second review in https://github.com/ramensoftware/windhawk-mods/pull/5536#issuecomment-5707008025.

/ai-review

RisPNG:
Addressed the second review in https://github.com/ramensoftware/windhawk-mods/pull/5536#issuecomment-5707008025.

/ai-review

RisPNG:
Pushed 79e6ac4, which goes through the third review.

> **1. Return `FALSE` from `Wh_ModInit` when no folder has icons configured.**

Done. I checked the reload part first with a test mod: it returned `FALSE` with empty settings, and after the setting was filled in, Windhawk loaded it again into the already running Explorer, so no restart is needed. The readme mentions it too.

> **2. `g_createdOverflowManager` / `g_createdOverflowInterface` are raw pointers that are never invalidated.**

Both are cleared right after the `InitializeIfNeeded` attach, `g_initializingManager` is reset with `ScopeExit`, and the constructor scan checks `IsReadable` before each read.

> **3. Harden the teardown path in `Wh_ModBeforeUninit`.**

`GetOverflowThreadWindow` falls back to the process's `Shell_TrayWnd` now. On my machine that window is on the same thread as the overflow island anyway.

## Optional improvements

- All the view-model symbols are optional now, so a missing one degrades the mod instead of stopping it, and the tooltip/name check is reachable.
- XAML metadata: I checked this in Explorer. Only `SystemTray.IconViewModel.Configuration` and `ContentVM` are registered with the metadata provider. `ToolTipText`, `AutomationPropertyName`, `IconImageSource`, `IconConfiguration.DataModel` and `NotificationAreaIconsDataModel.UDKObject` all come back missing, so the vtable slots have to stay.
- `Wh_ModSettingsChanged` applies the settings directly when the marshalled call fails.
- `FindDescendantByName` has no depth limit anymore, and the `Loaded` handler logs when the items control can't be found.
- Added `<initializer_list>`, and the `INotificationAreaIcon6` comment now says the IID and member order come from `windowsudk.winmd`.

## Functionality notes

- Rebuilds are skipped while the flyout is closed, since `Show` rebuilds anyway. `XamlRoot.IsHostVisible` stays true after the flyout closes, so it checks the island window's visibility instead, which I confirmed flips on open and close.
- The preview, its icons and the glyphs scale with the grid's item size now.
- Tiles are only recreated when the folder count or a glyph changes. Renaming a folder keeps the same tiles.
- Esc as back: I tried it with a `PreviewKeyDown` handler on the overflow control, but it never fires for Escape. The shell closes the flyout before XAML gets the key, so it would need subclassing the flyout window, which I don't think is worth it. Esc closing the flyout and reopening at the main grid works fine as it is.

/ai-review

RisPNG:
Pushed 9632c73, which goes through the fourth review.

> **1. Finding the overflow control by scanning raw object memory is a crash risk, and a much simpler path already exists in the repo.**

The early return is in: `EnsureAttached` now returns the existing state when it's attached with a live overflow and items control, before any lookup or tree walk. So the scan only runs once, and only when the mod is loaded into an Explorer that's already running, since a fresh start is covered by the constructor hook.

I didn't switch to field `[5]` though, because it isn't the overflow root grid on this build (SystemTray.dll 2607.28000). A read-only dump of the manager's fields shows `[2]` is the flyout's island window, `[5]` and `[6]` point to objects owned by Windows.UI.Xaml.dll, and the `NotificationAreaOverflow` interface is at `[7]`. During development, the exact `((IUnknown**)pThis)[5]->QueryInterface(...)` call from taskbar-notification-icon-spacing crashed Explorer here twice, which is why this mod looks for the verified vtable instead of trusting an offset. The scan never calls into anything that doesn't have the exact `NotificationAreaOverflow` vtable, and the field it matches belongs to the manager, which keeps the flyout alive.

> **2. `state.islandWindow` can be associated with the wrong overflow island.**

Two things I checked here. With a secondary taskbar on the same thread there's still only one overflow island, since secondary taskbars don't have the "^" flyout. And `XamlRoot().IsHostVisible()` stays true after the flyout closes, so using it would never skip anything. So instead, the island is recorded per state right after that manager's `Show`, and only when exactly one island window is visible at that point. Otherwise it's left unset and every change rebuilds as before.

> **3. Cleanup is best-effort, and on failure mod code stays registered in the live XAML tree.**

`Wh_ModBeforeUninit` and `Wh_ModSettingsChanged` now try every window on the tray thread, then the taskbar window, until one round trip actually runs. The cleanup callback is safe to run more than once.

## Optional improvements

- Settings are only ever applied on the tray thread now. `Wh_ModSettingsChanged` stores them as pending and dispatches, and if no round trip gets through, they're applied on the next `InitializeIfNeeded` or `Show`, so there's no write from another thread and nothing is dropped. `LoadSettings` runs once.
- Tiles remember what they're showing and swap only their content when the glyph changes, so the `as<Controls::Panel>()` can't hit a `TextBlock`, and a settings change no longer clears all tiles.
- `Detach` resets the state before rethrowing, and states are removed from the list before they're detached.
- The redundant resize in `EnsureFolderTile` is gone.

## Functionality notes

- Drag and drop: tested with three folders in front, dragging icons between the taskbar and the flyout, and reordering both in the main grid and inside an open folder. It works because the shell's drag handling goes by the icon views, not positions in `OverflowIcons`.
- Icon info is cached by view model and read again every time the flyout opens, so updates while it's open reuse what's already read.
- A folder with nothing to preview (no image icons, or empty with "Show empty folders" on) shows a folder glyph instead of an empty plate. Tiles are also rebuilt when the grid's item size changes, so they follow the icon spacing mod.

/ai-review

RisPNG:
Pushed f056b99, which goes through the fifth review.

> **1. `Detach` can abort partway and leave mod-owned XAML handlers registered**

These removals can't throw. In the C++/WinRT headers this compiles against, the removers used here are all declared `noexcept` and call `remove_...` without checking the result:

- `consume_Windows_UI_Xaml_IFrameworkElement<D>::Loaded(winrt::event_token const& token) const noexcept`
- `consume_Windows_Foundation_Collections_IObservableVector<D, T>::VectorChanged(winrt::event_token const& token) const noexcept`
- `consume_Windows_UI_Xaml_Controls_Primitives_IButtonBase<D>::Click(winrt::event_token const& token) const noexcept`

So nothing throws there and `ClearTiles` always runs, which means wrapping them in `step` wouldn't change anything. I updated the comment to say this directly. The wrong-thread case doesn't apply here either: I checked that `Shell_TrayWnd` is on the same thread as the overflow island (`TopLevelWindowForOverflowXamlIsland`), so the fallback window runs the cleanup on the right thread.

> **2. The field scans read past the end of the target objects**

On this build the scan doesn't get far, because it stops at the first match: a read-only dump of the manager's fields shows the `NotificationAreaOverflow` interface at `[7]` (offset 0x38), with the island window at `[2]` and objects owned by Windows.UI.Xaml.dll at `[5]` and `[6]`. The manager also has object pointers out to `[21]`, so I lowered the limit from 0x100 to 0x80. That keeps the fallback inside the object while still covering `[7]`.

I didn't add `[5]` as the known field, because it isn't the overflow control here. The `((IUnknown**)pThis)[5]->QueryInterface(...)` call from taskbar-notification-icon-spacing crashed Explorer twice during development. With the vtable check, `[5]` would just be skipped anyway, since its vtable belongs to Windows.UI.Xaml.dll. The constructor scan reads within the `NotificationAreaOverflow` object itself, which is 0xB0 bytes.

## Optional improvements

- The `LoadLibraryExW` lookup and hook are checked now, and `Wh_ModInit` logs and returns `FALSE` if either fails.
- I kept the blank-entry tolerance in `LoadSettings`. The default settings ship `icons: [""]`, so a new user's first entry is blank, and stopping at the first blank would drop every pattern they add after it. The second review also asked for blank folders to be tolerated.
- Lowercase names are stored in `IconInfo` when the icon is first read, and `FindFolderForIcon` is passed those stored strings instead of temporaries.
- The readme says "Set **Debug logging** to **Mod logs**" now.
- I didn't add a count of stranded states to the `Wh_ModBeforeUninit` failure log, because it would mean reading `g_states` off the tray thread. The states are left alone on purpose in that case, since their XAML objects can't be released from another thread.

## Functionality notes

- Tile size on the first open: after the layout pass in the `Show` hook, the tiles are sized again if the grid's item size changed. Also, taskbar-notification-icon-spacing sets `WrapGrid.ItemWidth`/`ItemHeight`, which is exactly what the tiles are sized from, so they follow its overflow icon width. The "Unsupported class name" log for the tiles is expected and harmless.
- Drag and drop: with three folders in front, dragged icons land where they're dropped, both in the main grid and inside an open folder, and icons dragged back from the taskbar go back into their folder.
- `islandWindow` is cleared now unless exactly one visible island is found, so live updates never depend on another island's visibility.

/ai-review

RisPNG:
Pushed 3586219, which goes through the sixth review.

> **1. On unload, the mod can leave its own COM objects registered inside Explorer**

The restore fallback is changed. `Detach` now tries the original `ItemsSource` value first, then the shell's current `OverflowIcons` list, then clears the value, so the mod's list is only left installed if all three fail. It's still filled with the original icons in that case, since that's all that's left to do. `Wh_ModBeforeUninit` also retries the round trip three times, and if all of them fail it logs an explicit error saying the flyout still uses the mod's list. The states and XAML member handles are left alone in that case on purpose, because releasing XAML objects from another thread isn't safe.

> **2. `g_createdOverflowManager` / `g_createdOverflowInterface` are unowned raw pointers that survive an exception**

`g_createdOverflowInterface` is a `winrt::com_ptr<IUnknown>` now (with `[[clang::no_destroy]]`), so the recorded object is kept alive while it's recorded, and both are cleared by a `ScopeExit` at the start of the `InitializeIfNeeded` hook, so they're released even if the original throws. I tested this after an Explorer restart: the constructor records the overflow at `this+0x18`, the lookup uses that record, and the mod attaches on the first open.

## Optional improvements

- `CreateGlyph` uses just `L"Segoe Fluent Icons"` now.
- `FindVtableSlot`: the slot is only used on an object whose vtable was checked to be the one that was scanned, and the only thing that can match is the getter address resolved from symbols. So even a slot found past the end of the vtable calls exactly that getter on the right object, and an arbitrary call isn't possible.
- The readme now says that two empty rows in a row end a list. I kept the blank-row tolerance because the default settings ship `icons: [""]`.
- The icon cache holds `std::shared_ptr<IconInfo>`, so a rebuild shares the entries instead of copying them.
- The readme mentions **Show empty folders** in the "unloads itself" sentence now.

## Functionality notes

- Field scans: after the first attach, `EnsureAttached` returns the existing state through its weak refs without any lookup, so the scan only runs for the first attach when the mod is enabled in an Explorer that's already running. The constructor scan stops at `this+0x18`, which is well within the 0xB0-byte `NotificationAreaOverflow` object.
- `AppFriendlyName` at slot 6: `windowsudk.winmd` on this build has `WindowsUdk.UI.Shell.INotificationAreaIcon6 {D76CAD80-3103-54F4-9ABE-5F047CBFE85C}` with a single method, `[6] get_AppFriendlyName`, which matches the IID and slot used here. It also returns the right app names for all the icons at runtime.
- Drag reorder: tested with the folders in front and inside an open folder, and icons land where they're dropped. Since that works with the tiles shifting everything by three, the shell goes by the dragged item rather than its index, which also covers `folderPosition: end`.
- The readme notes that a changed tooltip is picked up the next time the flyout opens.

/ai-review

RisPNG:

Pushed 721fb42 for the optional items from the seventh review.

- `Wh_ModBeforeUninit` returns right away when `g_overflowThreadId` was never set, so an Explorer process that only hosts folder windows doesn't wait or log the error anymore. `Wh_ModSettingsChanged` skips the round trip in that case too; the settings stay pending and are applied on the next `InitializeIfNeeded` or `Show` if a tray ever shows up in that process. I tested it with a folder window running in its own `explorer.exe /factory,... -Embedding` process: disabling the mod there logs nothing beyond the entry line.
- First open in a fresh Explorer session: this is the order on this build, since `Show` is what calls `InitializeIfNeeded`. Icon names are now logged whenever the icon cache was empty at the start of a rebuild, and after the original `Show` returns, the hook looks the state up again, so the tile size check and the island lookup also run on that first open. After an Explorer restart, the first open now logs the `Icon: app=... folder=N` lines and records the island window.
- Removed `<cstdlib>`.

The removed-icon trade-off while the flyout is closed is deliberate, as you noted, and the build-specific lookups fail safe by never attaching.

/ai-review

m417z:
Thanks for the submission. From a quick glance, one thing I noticed is the usage of `IsReadable` which is quite fragile. Please consider replacing it with more reliable ways of getting the offsets or data members. Let me know if you need help with that and I'll try to help.

RisPNG:

> From a quick glance, one thing I noticed is the usage of `IsReadable` which is quite fragile. Please consider replacing it with more reliable ways of getting the offsets or data members.

Thanks for the heads up. `IsReadable` is completely gone now, together with the guessing that needed it.

Before, the mod had to go looking for the tray flyout inside the shell's memory, checking spot by spot until it found something that looked right. Now it doesn't look for it at all, it just waits for the shell to hand it over. Every time the flyout opens, the shell measures it once before showing it, and the mod listens in on that exact moment (`consume_Windows_UI_Xaml_IUIElement<NotificationAreaOverflow>::Measure`), so it gets the flyout straight from the shell. Since this happens on every open, it works the same whether Explorer was just started or the mod was turned on while Explorer was already running. That made the constructor and `InitializeIfNeeded` hooks unnecessary, so both are removed too.

For reading the icon names, the mod used to search through each icon's function table to find where the getters are. It now calls the getters directly through the addresses Windhawk already resolves from the symbols, after the same check that the icon is the type it expects, so `FindVtableSlot` is gone as well.

Overall the file got about 120 lines shorter. Let me know if this is close to what you had in mind.

/ai-review

PR #5540, opened 2026-09-16
Tourne'Table [Audio Visualizer] 1.2.0

USER-TOURNE:
A real-time desktop audio visualizer that draws behind the desktop icons. 8 shapes, 9 color modes,
configurable FFT size with log / linear / mel frequency scaling, a peak-frequency readout, Now Playing
text with album-art coloring from the native Windows media session, a clickable media control strip,
keyboard and drag repositioning, and idle behavior that stops rendering entirely when audio stops.

Runs as a tool mod in its own process rather than inside `explorer.exe`, so a fault in the visualizer
cannot take the shell down with it, and it survives an `explorer.exe` restart.

## On the catalog-structure question from the previous review

The AI review on #5480 raised this as the main blocker and asked me to state my position explicitly so
the maintainer can rule on it. Doing that here rather than leaving it open.

**I am asking for this to be its own entry.** My reasoning, and I accept the decision either way:

- It is a **tool mod** (`@include windhawk.exe`), running in its own process. `desktop-audio-visualizer`
  injects into `explorer.exe`. That is not a settings-level difference, it changes the failure model,
  the teardown requirements and the lifetime of every thread in the mod. Folding this back into the
  original would mean rewriting its process model, not adding options to it.
- The feature surface is substantially larger: 8 shapes against the original's bar renderer, 9 color
  modes including album-art and accent-derived, configurable FFT size, three frequency scalings, a
  peak-frequency readout, Now Playing text, a media control strip, keyboard and drag repositioning,
  independent panels and borders for each overlay, per-side padding, and settings validation.
- The rendering path was rebuilt rather than tuned: frame pacing off a high-resolution waitable timer,
  a cached wallpaper blur instead of a per-frame convolution, a widget-sized render surface instead of
  a desktop-spanning one, cached geometry, and a capped frame latency.

If the ruling is that this belongs as a version bump to `desktop-audio-visualizer` instead, say so and
I will take that route.

## On the comparative benchmarking

The previous review noted that the mod page read as a critique of a neighbouring mod rather than as
documentation, and that the comparative benchmarking belonged in my own repo.

That has been removed. The page now documents what the mod does, and the only performance figures on it
are measurements of this mod against an idle baseline on the same machine, with no other mod referenced
or named. The full methodology lives in my own repository.

## Licensing

The previous review suggested that because `desktop-audio-visualizer` declares no `@license`,
re-licensing derived code as MIT would need the original author's explicit agreement.

This repository's own submission policy already settles that, in the README under *Submitting a New Mod*:

> Mods which don't specify a license are submitted under the MIT license. It's the author's
> responsibility to specify the appropriate license for third-party code.

A mod with no `@license` was therefore submitted under MIT, so MIT is the correct declaration for work
derived from it, and no separate grant is required. Flagging it here only so the point does not need
raising a second time.

## What changed since the 1.0.0 submission

Frame pacing measured elapsed time with `GetTickCount64()`, which only advances on the ~15.625 ms system
timer tick, so any Target FPS above about 64 was silently clamped to 64 (`1 / 15.625 ms = 64.0`, measured
at 63.97 FPS against a 144 target). Elapsed time is now measured with `QueryPerformanceCounter` and the
interval computed in floating point, which also closes the integer-division quantization the previous
review noted separately. Measured at 141.71 FPS against the same 144 target.

The three per-frame `Present` calls moved from a sync interval of 1 to 0. On a DirectComposition swap
chain DWM owns presentation timing and the waitable timer already paces the loop, so the extra wait only
parked the render thread. Across 219 five-second windows, time in `Present` fell by a third and the share
of windows where the thread blocked for more than 8% of wall time fell from 18% to 4%. The two one-off
`Present` calls are unchanged, and the swap chain stays at 2 buffers, which measured as making no
difference.

## Changelog

New mod, so there is no existing changelog to add to. Changes since the 1.0.0 submission are described above.

## Mod authorship

This mod was created by:

- - [x] Claude
- - [ ] ChatGPT
- - [ ] Gemini


USER-TOURNE (commented):
/ai-review

USER-TOURNE:
/ready-for-reviewer

I've investigated the scale of changes, some have slight merit. 

1.2 will include fixes I can reliably make with testing attached. All previous claims made by the ai-reviewer have extensive testing attached as to why they are not implemented within my own repo page, in the full readme, and within the changelog.

m417z:
Thanks for the submission.

Fix item 7 in the latest review. As it is now, it's not compatible with the portable Windhawk version.

Fix this:

> `[[clang::no_destroy]] static std::thread g_gsmtcThread` (line 1201) works because it's always joined first, but the documented form for a global thread is `[[clang::no_destroy]] std::optional<std::thread>` with `join()` + `reset()` - see [#1-worker-thread-stdthread](https://github.com/ramensoftware/windhawk/wiki/Global-objects-and-process-shutdown#1-worker-thread-stdthread).

Remove `// @architecture        x86-64`, windhawk.exe is 32-bit in 1.x versions. I also suggest testing the mod with v1.7.3, which the vast majority of the users have right now.

USER-TOURNE:
> Thanks for the submission.
> 
> Fix item 7 in the latest review. As it is now, it's not compatible with the portable Windhawk version.
> 
> Fix this:
> 
> > `[[clang::no_destroy]] static std::thread g_gsmtcThread` (line 1201) works because it's always joined first, but the documented form for a global thread is `[[clang::no_destroy]] std::optional<std::thread>` with `join()` + `reset()` - see [#1-worker-thread-stdthread](https://github.com/ramensoftware/windhawk/wiki/Global-objects-and-process-shutdown#1-worker-thread-stdthread).
> 
> Remove `// @architecture x86-64`, windhawk.exe is 32-bit in 1.x versions. I also suggest testing the mod with v1.7.3, which the vast majority of the users have right now.

/ready-for-reviewer

Thanks. All three are done in 1.2.0.

**Item 7.** Runtime placements now go through `Wh_SetStringValue` / `Wh_GetStringValue` instead of a file of the mod's own. `GetPositionOverridePath()` is gone.

Rather than only changing where it writes, first run migrates: it reads the old `%LOCALAPPDATA%\TourneTable\position_override.txt`, carries the contents into the value store, then deletes the file and the directory. Anyone upgrading keeps their layout, and the file the old versions left behind is cleaned up rather than orphaned. That path runs once from init, never from an input hook, and can be dropped in a later version.

One side effect worth having: saving a nudge previously meant `CreateDirectory` plus `CreateFile` plus `WriteFile` from inside a `WH_KEYBOARD_LL` callback. It is now a single value write.

**The thread.** `g_gsmtcThread` is now `[[clang::no_destroy]] std::optional<std::thread>`, created with `emplace()` and torn down with `join()` followed by `reset()`.

**`@architecture`.** Removed. Flagging what that exposed, since removing the line on its own would have made things worse rather than better.

With the declaration gone I built for `i686` and it failed. Five Win32 enumeration callbacks were being passed capture-less lambdas:

```
EnumDisplayMonitors    EnumFontFamiliesEx    EnumWindows (x3)
```

Those parameters are `__stdcall`. On x86-64 there is only one calling convention, so the compiler accepts a lambda without comment. On 32-bit `__stdcall` is distinct, a capture-less lambda converts to a `__cdecl` pointer, and none of the five bind. The mod would not have compiled for a 32-bit host at all, which is to say for most installs.

Each of the five now declares `__stdcall` explicitly. It builds clean for both `i686` and `x86_64`, with zero warnings on either.


Tested on Windhawk 1.7.3 (portable) as suggested, and it installs and runs correctly there.
The remaining AI review items are unchanged, and the note I left on those still stands.

m417z:
This is a new mod, why is a migration needed?

USER-TOURNE:
> This is a new mod, why is a migration needed?

Ah this was because I had a few users (brother, partner etc.) setup a build on their systems when I was testing still, and because I have a page up in the discord for anyone who wanted to test previously while the PR was up.

In this case (and in my own) I wanted a way for those users to have settings migration with no hassle due to the nature of my keyboard move's position saving

That said, if you want it stripped I can do that!

m417z:
OK, we can keep it for now and remove it in a future update.

USER-TOURNE:
> OK, we can keep it for now and remove it in a future update.

Sweet, ill make sure its gone in future updates! Appreciate the help letting me know it didn't build in portable.

PR #5549, opened 2026-09-17
SnapSentry 0.21.1

mario0318:
adds an option to clean up identical screenshots after it copies the image. off by default, and it only looks at recent captures from this run, not the older files already sitting in the folder. a duplicate goes to the recycle bin even when normal deletion is set to permanent, so it's always recoverable.

it rechecks the earlier copy before recycling anything. if that one's gone or changed, or the two don't still match byte for byte, the new shot stays. changing a setting cancels a pending cleanup. each capture gets a log line for what happened to it.

tested the file handling on windows 11, recycle bin included.


mario0318:
> The AI review bot is under constructions, so no review was posted this time. Please try again in a few hours.


/ai-review

mario0318:
Handing this over for human review.

One item from the round 3 review is left as-is: if the earlier copy is removed (by me or a sync client) within the ten-minute window, an identical later capture can still be recycled as a duplicate, which the description frames as keeping the older one. It always goes to the Recycle Bin so it stays recoverable, and it needs that specific timing to happen at all, so I'd rather not add more churn over it. Glad to check the keeper still exists before recycling if you'd prefer that.

m417z:
> I'd rather not add more churn over it. Glad to check the keeper still exists before recycling if you'd prefer that.

I think it'd be better to add a check for that, it'd be less surprising, and a user might not think of checking the recycle bin. I believe the implementation should be straightforward, the last review shows the main parts (`std::wstring keeper`, `GetFileAttributesW`).

I'm also OK with merging it as is and leaving it for future updates. Let me know what you prefer.

mario0318:
Decided to just fix the keeper check rather than leave it. A later identical capture is now recycled only when the earlier copy still exists; if that copy is already gone, the new one is kept. Folded in the small cleanups from the review too (drive-root dedup, delete sharing on the capture handle, the toast handler symmetry). Re-running the review before handing back.

PR #5550, opened 2026-09-17
Add Explorer Tags mod

buedgik:
Adds `mods/explorer-tags.wh.cpp` — colored tags for File Explorer.

Windows' own tags (`System.Keywords`) only work for file types with a property handler, so `.txt`, `.zip`, `.rar` and folders can't be tagged. This mod tags anything.

**What it does**

- A **Tags** panel at the bottom of the navigation pane, below This PC and Network. Drag files or folders onto a tag to tag them; click a tag to open its folder in the current tab; click the title to collapse the panel.
- A **Tags ▸** submenu in the classic file context menu. A checked entry means every selected item has that tag, and picking it removes the tag; unchecked adds it. "Remove all tags" shows up when at least one selected item is tagged.
- Tags (name and color) are defined in the mod settings.

**How it stores things**

- One shortcut folder per tag under `%USERPROFILE%\Tags`, so a tag's contents are an ordinary Explorer view. Deleting a shortcut removes the tag; the original file is never touched.
- The record of what is tagged lives outside that folder, in `%LOCALAPPDATA%\WindhawkExplorerTags\tags.tsv`, so deleting or moving the tags folder rebuilds the shortcuts instead of losing tags. Each tag folder carries a hidden `.tag` marker that distinguishes "the user deleted this shortcut" from "this folder is new or was rebuilt".
- Tagged files are tracked by NTFS file ID, so renaming or moving one within a drive keeps the tag and fixes the shortcut. A file replaced at the same path by an editor is recognized by path.
- All disk work runs on a worker thread; the Explorer UI thread never waits on disk or on the cross-process mutex. Shortcut changes are announced with `SHChangeNotify`, so open views update.

**Compatibility**

The context menu part needs the classic context menu, e.g. with the Classic context menu mod or by holding Shift; the Windows 11 menu is not modified. The submenu is only added to the file context menu of a `SHELLDLL_DefView` inside a `ShellTabWindowClass`, and only when the menu has a default item and no ID-0 "Cancel" entry, which keeps it out of the right-drag menu and the folder background menu.

**Testing**

Tested on Windows 11 24H2 (26200) with File Explorer: the panel in several windows including ones opened later, tag click navigation, collapse, drag-and-drop tagging, context menu tag and untag with check marks, a tagged file renamed (tag follows, shortcut retargeted), a shortcut deleted (tag removed, original untouched), and open views refreshing on both tag and untag.

Source repository: https://github.com/buedgik/explorer-tags

## Changelog

* New mod, no changelog.

## Mod authorship

This mod was created by:

- - [x] Claude
- - [ ] ChatGPT
- - [ ] Gemini

The design decisions and the testing plan were the submitter's; the C++ was written by Claude (Claude Code), reviewed adversarially in several passes, and then tested on a real Windows 11 machine before this submission.


buedgik:
Closing and reopening to re-run the validation after restoring the Mod authorship section.

buedgik:
﻿Thanks for the review — all three findings are addressed in 0.3.0.

**1. No pinning.** `PinForLateCalls` and both 5 s fallbacks are gone. `Wh_ModBeforeUninit` now waits unconditionally for `g_busy` and `g_drags`, and `Wh_ModUninit` waits for `g_hookCalls`, however long a menu stays open; the `g_msgCancelMenu`/`EndMenu` and `WM_CANCELMODE` paths stay as the way to end an open menu promptly. `Wh_ModInit` resets `g_unloading`, `g_workerThread` and the counters anyway, as suggested.

**2. No shell link loads on the UI thread.** The worker now records each shortcut's target alongside the counts (`g_linkTargets`, filled in `UpdateCounts` from the rows it already has), and the menu path resolves a selection with a hash lookup. No `CoCreateInstance` and no `IPersistFile::Load` on the menu path, whatever the selection size.

**3. Screenshot** added to the README, hosted on `raw.githubusercontent.com`, showing the panel and the submenu with its check marks.

Optional items: typed `WindhawkUtils::SetFunctionHook` with the failures logged (a failed `CreateWindowExW` hook now fails init; failing menu hooks only lose the submenu), `g_collapsed` is `std::atomic<bool>`, `-loleaut32` dropped, the current folder is read on navigation and refresh instead of on every paint, settings are a `shared_ptr<const Settings>` so the hot paths don't copy the tag list, the dead `inset`, the Portuguese comment and the ignored `SetWindowSubclassFromAnyThread` result are fixed.

Functionality notes: the tags folder is no longer created until something is actually tagged (`WatchRoot` doesn't create it, `SyncTag` skips a tag with no rows, and the folder is born on the first tag); the wheel honors `SPI_GETWHEELSCROLLLINES` and accumulates partial notches; and the README and the `folder` setting's description now warn about synced folders, about the old folder being left behind when the setting changes, and about the data outliving the mod being disabled.

Left alone, with reasons: the keyboard-accessible panel and the `SyncAll` cost per tagged file are real but are a bigger change than I want in a first version. On the column-header menu: I couldn't reach it to measure — injected input doesn't work in my test environment, and `WM_CONTEXTMENU` sent to the DefView or its DirectUIHWND produces the folder background menu. That background menu is rejected by the heuristic (logged: `View menu: 19 items, default -1, not the files menu`), and the right-drag menu is rejected by the ID-0 Cancel entry, measured in the shell32 menu resources.

Testing status, to be precise: 0.2.2 (the panel, tag navigation, collapsing, drag-and-drop tagging, the context menu tag/untag with check marks, a renamed tagged file, an untag by deleting the shortcut, and open views refreshing) was tested end to end on the Windows 11 machine this was written on. The 0.3.0 changes above compile clean and are being re-tested on that machine now; I'll report back here if anything turns up.

/ai-review

buedgik:
0.3.1 pushed.

**The unload wait that could never end** — you're right, and it was a deadlock, not a stall: `g_drags` is held up by the mod's own reference on the target, which `Detach` is what releases. The detach loop now runs before the wait, exactly as suggested, so a drag whose source died ends as soon as OLE releases its reference. `g_lastDragTick` is gone with the rule it documented.

Optional items, all done: the three handles are closed on every failed `Wh_ModInit` (via a small `InitFailed()` that both `Wh_ModUninit` and the failure paths share), `WindhawkUtils::StringSetting` replaces the manual free in `LoadSettings`, the `RegisterClassW` retry is removed, the 64-tag read limit and the menu limit are now one `MAX_TAGS`, a failed `CreateThread` is logged, `WriteShortcut` sends `SHCNE_CREATE` or `SHCNE_UPDATEITEM` depending on whether the shortcut existed, and `<utility>`, `<cwchar>` and `<cstring>` are included directly.

Functionality notes: `%LOCALAPPDATA%\WindhawkExplorerTags` is now created only by the two functions that write, so the README line is true again; duplicate targets in a selection are collapsed after the cached resolution, which fixes the check mark and the double queueing; `Wh_ModBeforeUninit` posts `WM_CANCELMODE` to this process's top-level windows so a menu the mod never touched doesn't hold the unload; and the README and the `folder` description now say that deleting a shortcut untags while renaming or moving one isn't tracked, and that the whole tree under the tags folder is watched.

Left as is: the panel is still mouse-only and `SyncAll` still re-resolves every row on a folder change — both are real, both are a bigger change than I want in a first version. The persistent data under `%USERPROFILE%\Tags` and `%LOCALAPPDATA%` is intentional and documented, and I agree it's for the human reviewer to weigh in on.

Testing status: 0.3.0 was re-tested end to end on the Windows 11 machine here (panel, tag navigation, collapse, context menu tag and untag with check marks, a renamed tagged file, an untag by deleting the shortcut, and an open tag folder view refreshing on both). 0.3.1 compiles clean with Windhawk's own compiler and flags; the same pass is being run on it now.

/ai-review

buedgik:
Heads-up: 0.3.1 doesn't build for x86 or arm64, and 0.3.2 (just pushed) fixes it.

The `WM_CANCELMODE` sweep I added in 0.3.1 passed a capture-less lambda to `EnumWindows`. That converts to a `__cdecl` function pointer, which is the same as `CALLBACK` on x86-64 but not on x86 or arm64, so it compiled in my check and failed in Windhawk. It's a named `BOOL CALLBACK` function now, like the other enumeration callbacks in the file.

My local check script only built `x86_64-w64-mingw32`, which is how this got past me; it now builds all three targets with Windhawk's compiler and flags, and I confirmed the 0.3.1 file fails on x86 under it while 0.3.2 passes on all three.

Sorry for the noise mid-review — if the review in flight was taken against 3e0faae/ba408c9, the only difference in 0.3.2 is that one callback.

buedgik:
0.4.0 pushed. The first finding was a real bug and a good catch — it could only ever have been found by reading, since every machine that had run an earlier build already had the folder.

**1. Nothing was ever stored on a clean machine.** Confirmed by measuring rather than reasoning: opening a file whose parent folder doesn't exist fails with `ERROR_PATH_NOT_FOUND` (3), while a missing file in an existing folder gives `ERROR_FILE_NOT_FOUND` (2). `LoadRows` accepted only the second, so every caller bailed out and `EnsureParentFolder` — reachable only after a successful load — never ran. `LoadRows` now accepts both, with the measurement in the comment. I'll re-test with `%LOCALAPPDATA%\WindhawkExplorerTags` removed on the machine here.

**2. The unload path.** The process-wide `WM_CANCELMODE` sweep is gone — you're right that cancelling other windows' drags and selections on every toggle is worse than the problem. The menu-driven drains (`g_busy`, `g_hookCalls`) now give up after 10 s with a log line; the `g_drags` drain stays unbounded, since OLE holds a pointer into the image until it releases the target and detaching first means it ends as soon as OLE lets go. `g_msgCancelMenu` to the panels and `WM_CANCELMODE` to `g_menuOwners` stay. Noted on `SendMessageW(tree, g_msgDetach)` blocking on a wedged UI thread: the detach has to happen, so I've left it, aware that it's another unbounded point.

**3. A newline in a tag name** split the record and lost that tag's rows on the next load. Tab, CR and LF are all replaced now.

Optional items done: a menu command is only treated as ours when `GetMenuItemInfoW` finds it in our submenu, the separators are inserted with `InsertMenuItemW` so the ID survives and the cleanup finds them, `UnregisterClassW` failure is logged, the row height is cached per DPI and invalidated on theme and DPI changes instead of a `FindWindowEx` + `SendMessage` per mouse move, `GlobalLock` is checked, and the recursion guard in `CollectMenuIds` is an early return. The `Wh_ModInit` resets stay, with the comment rewritten to say what they actually are — belt and braces, not a scenario that can happen.

Functionality notes: the README now says the desktop and file dialogs are deliberately left alone. `ReadDirectoryChangesW` to scope the sync to the tag that changed, and the duplicated work when folder windows run in separate processes, are both real and both bigger than I want to take on in a first version.

One more thing worth flagging: 0.3.1 (the sha this review covers) didn't compile for x86 or arm64 — a capture-less lambda passed to `EnumWindows` — and 0.3.2 fixed it. My check script only built x86-64; it now builds all three targets, and everything above was verified on all three.

Testing status: 0.3.2 was tested end to end on the machine here (panel, collapse, tag click navigation, context menu tag and untag, a renamed tagged file, an untag by deleting the shortcut, open views refreshing, and a disable/enable cycle in which the DLL left both explorer.exe processes cleanly and the panels came back on re-enable). 0.4.0 compiles clean on all three targets and is being tested now, including the clean-machine case above.

/ai-review

SoftwareExplorer6600:
@buedgik I got a few suggestions for this tagging mod, could we have it run an AI that would go and automatically search through windows to apply tags and assign tag names based on context and content, even able to create new ones, and maybe instead of tag names being in the mod with a color assigned, there should be a 'create new tag' button that when pressed opens a WinUI 3 dialog that has a title 'Create New Tag?' which then under it has a dropdown of predefined colors and a space to implement the name.

buedgik:
0.6.0 pushed. Both required items are done, all four optional ones too, and one of them I'd push back on.

**1. The record moved into the mod's storage.** `DbFolder()` is now `Wh_GetModStoragePath` plus a folder named after the user's SID (`GetTokenInformation(TokenUser)` + `ConvertSidToStringSidW`), and `tags.tsv`, `tags.tsv.tmp` and `last-root.txt` live in there. You were right that the per-user concern doesn't justify a path outside Windhawk's storage.

Two things I added around it:

- **The old record is brought over once.** Anyone running 0.5.x has a record in `%LOCALAPPDATA%\WindhawkExplorerTags`, and an upgrade that silently ignored it would look exactly like every file losing its tags. The worker copies both files on its first pass, under the database lock, then renames the originals to `.old` so a later fresh install doesn't import them again. Measured here: four rows, same 587 bytes, in the new location; the old pair renamed.
- **The mutex takes the global namespace when it can.** `Global\WindhawkExplorerTagsDb-<SID>`, falling back to `Local\` when creating a global object isn't permitted, which is the ordinary case for a non-elevated account. Same-user-twice then shares one lock where the privilege allows it, and nothing is worse than before where it doesn't, since both processes are the same account and get the same answer.

The README and the mod's own readme now say where the record lives and that uninstalling (not disabling) takes it along.

**2. Segoe Fluent Icons.** Asked for rather than assumed: `EnumFontFamiliesExW` for the family, falling back to `Segoe MDL2 Assets`. I went with the enumeration instead of the build number because the build number says nothing about a font that was added or removed by hand, and the callback is a named `int CALLBACK` — a capture-less lambda there compiles for x64 and breaks x86 and arm64, which this mod has already been caught by once.

**Optional items:**

- **`CountTagHits`** is now a lookup per selected file in a `path → tags` map the worker builds with the paths already lowered, instead of a walk with an allocation over every row.
- **The stale comment** about 5 s is gone.
- **`GetCurrentFolder`** uses `SHGetPathFromIDListEx` with a 4096-character buffer.
- **The Recycle Bin** keeps tags: a recycled file resolves as `Unknown`, so the row and the shortcut stay, pointing where the file was. Restoring it brings the tags back; emptying the bin makes the ID stop resolving and the existing `Gone` path cleans up. Tested end to end here (tag, recycle, restore, purge). The README has the caveat about a restored *shortcut*, which is a different thing and stays an orphan.

**The one I'd leave as it is:** the `.tag` marker is already documented in the README, in "How files are tracked" — "Each tag folder carries a hidden `.tag` marker, which is how the mod tells 'you deleted this shortcut, so untag it' from 'this folder is new or was rebuilt'." I've added the panel's middle-click and right-click "Open in new window", which genuinely weren't there.

**Something else, found by testing rather than by the review:** in 0.5.3 a single middle-click on a tag opened *two* Explorer windows, repeatably. The deferred open armed the pending click before attempting the immediate one, and `ShellExecuteW` runs messages: the worker's answer was dispatched inside the attempt, found a click still waiting, and opened a second window. The wait is now armed only after the attempt, a per-panel click counter stops a late arrival from overwriting a newer click, and the failure path asks the worker for the folder again instead of posting itself a refresh that would be answered with the same stale readiness that had just failed. `ShellExecuteExW` with `SEE_MASK_FLAG_NO_UI` also keeps the shell's "cannot find" box off the window thread, since that case is one the caller handles itself.

Verified on this build, on the machine: the migration, one middle-click one window three times over, a click with the tag folder deleted underneath (recreated and opened), tagging and untagging from the context menu with its check marks, the panel collapsing, and the Recycle Bin round trip. Compiles clean for x86_64, i686 and aarch64.

buedgik:
0.6.1 and 0.7.0 pushed. 0.6.1 is the three optional items from your last round; 0.7.0 is a change I'd like read carefully, because it moves where the truth about a tag lives.

**0.6.1 — the optional items.**

- `MigrateOldRecord` is gone, for the reason you gave: `%LOCALAPPDATA%\WindhawkExplorerTags` only ever existed in this pull request's builds, so a merged mod would carry a probe of a folder no released version used, and leave `.old` files behind if it ever fired. I moved my own machine by hand before deleting it.
- The re-ask in the deferred-open handler stops after twice. You're right that it couldn't spin forever — the 10 s deadline ended it — but a folder that is there and still won't open had the worker and the window shouting at each other until then.
- The first `QueueEnsureFolder` in `OpenTag` went: after the rework the success path has nothing to make and the failure path does its own asking.
- And the comment about `SeCreateGlobalPrivilege` was simply wrong — that check is for file-mapping objects, so an ordinary account does get the global mutex. The fallback stays, for the name being held by another account, and the comment now says that.

**0.7.0 — the record can be rebuilt from the shortcuts.**

The rule is now: what is in a tag's folder is what is tagged. At the end of each `SyncTag`, a shortcut in the folder that no row accounts for is taken in — the folder names the tag, the shortcut names the file, and the file's identity is read back from it. Only in a folder the mod marked as its own, never for a shortcut pointing at nothing or at something inside the tags folder itself, and never one the same pass is about to delete.

Three things follow from it:

- **The record is recoverable.** `Wh_GetModStoragePath`'s directory goes when the mod is removed, which is documented and fine — but it means the record is the one part of this mod that can vanish while the user's tags are still visibly sitting there in the folders. Now they come back on the next pass. (What prompted this: on my machine that directory was emptied at the exact second one of these builds was compiled — the folder's creation time equals the DLL's. I could not reproduce it on the next compile, so I'm not claiming compiling does that; I'm saying the record turned out to be losable in a way the shortcuts weren't.)
- **Dropping a shortcut into a tag's folder tags the file**, which is the obvious thing to expect from a folder full of shortcuts.
- **Three caveats in the README became behaviour that works**: renaming a shortcut keeps the tag, moving one into another tag's folder moves the tag, and restoring one from the Recycle Bin brings the tag back.

I break-tested the change before pushing it and it came back with six defects, four serious, all fixed in what's pushed:

1. **The rebuild could never run in the case it was written for.** Trust in a folder's marker was dropped whenever the root couldn't be confirmed, and the root is remembered in a file next to the record — so whatever takes one takes the other. The root was then only written back once rows existed, and rows could only come from the rebuild: the two conditions held each other shut for good. "The root changed" and "I have no memory of the root" are now different answers, and only the first distrusts the markers.
2. **Removing a tag undid itself when the shortcut couldn't be deleted** (open elsewhere, read-only). The row went, the shortcut stayed, and the next pass took it back in — silently. The shortcut goes first now and the row only follows if it is really gone; `DeleteLinkIn` reports failure, clears a read-only attribute and retries once.
3. **A second shortcut to the same file made a tag unremovable.** Removing a tag deleted only the shortcut the record named; any copy left behind was adopted the moment the row disappeared. Removal now sweeps the folder for every shortcut pointing at that file.
4. **The `.tag` marker claimed folders the mod had never written into** — `!trusted && complete` is true for a folder it merely found. With adoption that meant a tags folder pointed at a directory that already had a subfolder of that name would hand every shortcut in it to the tag. Only a folder this mod created, or keeps a tag's rows in, is claimed now.
5. Adoption skipped the "the tag folders themselves don't get tagged" guard that `AddPaths` has.
6. It compared paths as text where the rest of the mod compares (volume, file id) first, so one file reached by two spellings became two rows.

Verified on the machine, on this build: the record deleted and rebuilt from the shortcuts in under a second with the file identities found again; a shortcut dropped in by hand adopted; one pointing inside the tags folder left alone and not deleted; a shortcut renamed and the tag following it; a tag removed from a file that had two shortcuts, both deleted and the tag staying gone; plus the panel, the counts and the middle click as before. Compiles clean for x86_64, i686 and aarch64.

The README and the mod's readme document the rule, including the one limit: a tag whose name is no longer in the settings can't be rebuilt, since nothing maps its folder back to it — put the name back and its files come back with it.

buedgik:
0.7.1 pushed: both functionality notes and both polish items.

**The panel-click folder could never take anything in.** You're right, and it's the most natural way to use the feature: click a tag, the folder opens, drag a file in with "create shortcut here" — and nothing was tagged. `EnsureTagFolder` makes the folder without a marker, and an unmarked folder with no rows never became `ours`, so it was never marked and `AdoptStrays` never ran in it. I took your second suggestion: a folder holding no `*.lnk` at all counts as ours, since there is nothing in it to hand to the tag, which is exactly what the `ours` flag was guarding against. That covers the panel-click folder, a folder the user made by hand, and one left empty after everything in it was untagged — without claiming a folder that came with contents. (`FindFirstFileEx` failing for any reason other than `ERROR_FILE_NOT_FOUND` leaves it unclaimed.)

**The root is now remembered whenever the record file exists**, not only when it has rows. Same trap as the one my own break-test found, one step over: a record emptied of rows plus a changed folder setting stayed `Changed` on every pass, so no marker was ever trusted again. The "enabling the mod shouldn't create the folder" promise still holds, because the file existing is what the check now asks about.

**The untag sweep only opens what the record doesn't know.** A `name → row` map per folder, built from the rows whose tag folder matches; a known name takes its target and identity from the row, and `ReadShortcutTarget` is left for the names that aren't in the record — which is the only place an unrecorded duplicate can be. Untagging one file in a folder of a thousand shortcuts is now one COM load per stray instead of a thousand.

Break-testing that one found a hole in it, which is worth passing on: the record is good enough to *rule a shortcut out*, but not to delete it. A .lnk replaced behind the mod's back keeps its name while pointing somewhere else, and its row keeps the old target until a sync corrects it — so an untag would have deleted it on the record's word and taken another file's tag with it. The shortcuts that are about to be deleted are now read from the disk first and must still match, so only the ones actually going pay for the read.

**And the two leftovers**: the comment about deferred deletes is back above `Pending` where it belongs, and `rootPrefix` is hoisted out of the `AdoptStrays` loop.

Compiles clean for x86_64, i686 and aarch64.

One correction to something I wrote in my last comment: I said the mod's storage directory was emptied on my machine at the exact second one of these builds was compiled. That measurement stands, but a later compile did not reproduce it, so I can't tell you what caused it and I'm not asking you to act on it. The rebuild earns its place on the documented behaviour alone — the directory goes when the mod is removed, and the tags in the folders outlive it.

Ready for `/ready-for-reviewer` whenever you'd like; say the word if you'd rather I run one more `/ai-review` on 0.7.1 first.

Generated by AI. Model: Claude Opus 5. (So were my earlier replies on this pull request, before the disclosure line was asked for.)

buedgik:
A correction to my own correction, since I now have a third data point, and it goes the other way.

Compiling 0.7.1 emptied the mod's storage directory again: the directory's creation time, the new DLL's creation time and `SettingsChangeTime` in the registry all read `07:24:07`, to the second, and `tags.tsv` and `last-root.txt` were gone. That is two out of three compiles of this mod on this machine (0.6.1 yes, 0.7.0 no, 0.7.1 yes). I still can't tell you what makes the difference between them — I don't know which UI path each compile took — so I'm not filing this as a bug against Windhawk, and nothing in the mod depends on an answer.

I'm passing it on because it's the same event the rebuild exists for, and because if it can happen to a mod being compiled locally, the "the directory is removed when the mod is removed" guarantee is worth reading as "and possibly sooner". Verified on this build right after it happened: the record came back from the shortcuts one second after the first change in the tags folder, with the file identities read back from the files themselves.

The rest of 0.7.1 tested on the machine as well, all of it after the wipe, on a record that started empty:

- a file tagged from the context menu, and the record created for it;
- the record deleted with the shortcuts left in place, and rebuilt in a second, `last-root.txt` written with it;
- a shortcut replaced behind the mod's back (same name, pointing at another file) and then the recorded file untagged: the shortcut was left alone rather than deleted, the row went, and the next pass took the shortcut in for the file it actually points at — the self-healing your note said the optimisation had destroyed;
- a tag folder deleted, recreated by clicking the tag in the panel, claimed while empty, and a shortcut dropped into it by hand taken in a second later — the case that tagged nothing before;
- a shortcut inside a tag folder right-clicked, and the tag shown checked for the file it points at;
- a second shortcut to the same file not counted twice, and "Remove all tags" deleting both of them with the tag staying gone;
- the panel, its count, and a middle-click opening exactly one window.

Generated by AI. Model: Claude Opus 5.

buedgik:
One note to go with the handover, on the finding I left as it is.

The dangling shortcut after "Remove all tags" on a file renamed since the last sync: I traced it and it holds, and your fix is the right one — a `SyncTag` pass over the affected tags at the top of `RemovePaths`, the way `AddPaths` already does it, so the shortcut is retargeted and renamed before the sweep looks for it.

I'm not making the change now because it is narrow, it is not a regression (the pre-0.7.1 sweep missed the same way), and the worst it leaves behind is a `.lnk` pointing at a name that no longer exists — which the mod refuses to adopt and refuses to delete, so nothing is lost and nothing ends up wrongly tagged. It's first on the list for the update after this lands.

Generated by AI. Model: Claude Opus 5.

PR #5556, opened 2026-09-18
Create d3d9-flip-model-upgrade.wh.cpp

triatomic:
New mod: **D3D9 Flip Model Upgrade**.

Direct3D 9 games running windowed/borderless present with the legacy blt model, and Windows 11's "Optimizations for windowed games" only covers DX10/11. This mod transparently creates the game's device as a Direct3D 9Ex device with a `D3DSWAPEFFECT_FLIPEX` swap chain, so a borderless D3D9 game can be promoted to Independent Flip by DWM.

What it does:
* Hooks `Direct3DCreate9`/`Direct3DCreate9Ex` of the system `d3d9.dll` (never loads it itself, so proxy DLLs in the game folder keep working), then `CreateDevice`/`CreateDeviceEx`/`Reset`/`ResetEx` to adjust the present parameters for FLIPEX, falling back to the game's own parameters on failure.
* Forces the NULL rects/window override FLIPEX requires in `Present`/`PresentEx`/`IDirect3DSwapChain9::Present`.
* Emulates `D3DPOOL_MANAGED`, which doesn't exist in D3D9Ex: the game gets a system memory texture with a default pool twin attached as private data, uploaded with `UpdateTexture` at draw time. Managed buffers go to the default pool.
* Hides `IDirect3DDevice9Ex` from `d3dx9_*.dll`, which otherwise refuses managed pool loads.
* Exclusive fullscreen is left alone unless "Force borderless fullscreen" is enabled.

The mod has no `@include`; users add the game's executable to the custom process inclusion list. Tested with Command & Conquer: Red Alert 3 (32-bit), presentation mode confirmed as "Hardware Composed: Independent Flip".

## Changelog

N/A - new mod.

## Mod authorship

This mod was created by:

- - [x] Claude
- - [ ] ChatGPT
- - [ ] Gemini


triatomic:
Pushed fixes for findings 1-5:

1. The flip state now reflects the swap effect that was actually applied, so `Present` isn't rewritten with `flipModel` off (and `forceBorderless` works in that mode).
2. `EnsureD3D9Hooks` runs for every object; the throwaway Ex object and `Wh_ApplyHookOperations` are one-shot / only when something new was hooked.
3. Added `Wh_ModUninit`, and a warning near the top of the readme about disabling the mod mid-game.
4. Bound texture tracking is per device and SRW-locked. It also no longer holds references to bound textures (that would have kept the device alive); pointers are only compared, and only textures pending upload are referenced until the next draw.
5. Uploads happen at bind time and when the game modified a bound texture, detected via unlock/add-dirty hooks on 2D/cube/volume textures and surface `UnlockRect`/`ReleaseDC`. Draw calls cost one atomic check otherwise.

Also took: type-safe `SetFunctionHook`, `StringSetting`, extended frame styles, no `SWP_SHOWWINDOW`, `GetProcAddress` null check, removed the unreachable clamp, frame latency on the fallback device and after Reset, keeping the tracking when a twin from `GetTexture` is re-bound, and the two suggested limitation notes.

Not addressed: state block tracking, cube face `UpdateSurface` redirection, caching the Ex check on resource creation, pruning `g_flipDevices`, and a readme image - these are edge cases I'd rather handle if a real game needs them.

/ai-review

triatomic:
Pushed fixes for the second review:

1. The unload warning is now the first thing in the readme and also covers background mod updates. As noted, there is no code fix: the D3D9Ex device and the emulated textures cannot be reverted under a running game. Flagging it for the maintainer to sign off.
2. MakeBorderless uses SWP_ASYNCWINDOWPOS when the window belongs to another thread.
3. S_PRESENT_OCCLUDED / S_PRESENT_MODE_CHANGED are mapped to D3D_OK in Present and IDirect3DSwapChain9::Present, only for devices the game created as plain Direct3D 9 (tracked separately from the flip state), so native 9Ex games still get them.

Optional items taken: d3d9.dll is referenced instead of pinned and freed in Wh_ModUninit; Wh_ModUninit no longer releases textures off the render thread; the upgrade is refused for a device whose key functions are not the hooked ones; per-device state is cleared when a device is created; the D3DX caller lookup is cached per module; __uuidof instead of copied IIDs; pp = {}; SetTexture returns early until the managed pool emulation is in use; readme notes about managed buffers, state blocks, DPI scaling and aspect ratio distortion.

Not addressed: caching the Ex check on resource creation (load time only, and a cached answer could go stale on device address reuse) and a readme image.

Tested with Command & Conquer: Red Alert 3 (32-bit): menu video, a match, Alt+Tab.

/ready-for-reviewer

triatomic:
Pushed fixes for findings 1 and 2, plus a real bug found while testing with a second game.

**1. MakeBorderless blocking on another thread.** You were right that the style change is the blocking part. The styles are now only written when they would actually change, so the steady state makes no blocking call at all, and the first conversion is skipped (and logged) when the window belongs to another thread.

**2. Software vertex processing.** Managed vertex/index buffers moved to the default pool now get `D3DUSAGE_SOFTWAREPROCESSING` when the device was created with `D3DCREATE_SOFTWARE_VERTEXPROCESSING`. Devices with `D3DCREATE_MIXED_VERTEXPROCESSING` are no longer upgraded at all, since no default pool buffer works with both paths - logged, and noted in the readme.

**3. Reversibility.** Left as-is for the maintainer to sign off, as you suggested.

**Separately: the flip model conflicted with in-game overlays.** Testing a second game (Warhammer 40,000, 64-bit, Steam) showed the game rendering incorrectly with the Steam overlay enabled. The cause was mine: `CreateDeviceEx_Hook` upgraded *every* device it saw, so the overlay's own device targeting the game window also got FLIPEX, which the runtime refused with `E_ACCESSDENIED`. Two fixes:

* Only the first device of a window is upgraded; later devices on the same window keep the swap effect they asked for.
* Devices created without `hDeviceWindow` and without a focus window are not upgraded, since it's impossible to tell which window they will target. The Steam overlay creates one of these.

The game now renders correctly with the overlay enabled. Independent Flip still isn't reached while an overlay draws on top - that's the flip model requiring the window for itself, not something the mod can work around - so the readme now says to turn the overlay off for the full effect.

Not addressed: the optional items from this round (the `MarkTextureDirty` fast path, gating `UpdateSurface_Hook`, clearing the tracking after a successful reset rather than before, erasing `g_deviceTextures` entries, and the `Wh_ModInit` ordering), and the functionality notes. Happy to pick any of them up if you'd like them in before merge.

Tested with Command & Conquer: Red Alert 3 (32-bit, no overlay) and Warhammer 40,000 (64-bit, Steam overlay on and off).

PR #5562, opened 2026-09-18
Add Teams Multi-Instance mod

TSA3000:
New mod: **Teams Multi-Instance** (`teams-multi-instance`)

Allows running multiple independent instances of the new Microsoft Teams desktop app (`ms-teams.exe`), e.g. your own tenant and a client tenant side by side.

**How it works**
Teams enforces single-instancing with two named mutexes, `Teams-Tfw-instance` and `Teams-Tfw-server`. The mod hooks `NtCreateMutant` / `NtOpenMutant` (and the semaphore equivalents) in ntdll and appends `_<pid>` to names matching a configurable pattern. Only the top-level `ms-teams.exe` touches those objects (verified with Log only); child `ms-teams.exe` processes never do, so per-PID salts are safe. Sections and events are deliberately not hooked — they fire on the loader path and hooking them crashed the process during development.

The salting decision is made once per process on the initial thread (TEB `SameTebFlags`) and kept for the process lifetime via a PID-stamped environment marker, so a mod reload or Windhawk restart never un-salts a running instance, and an instance that was already running when the mod was enabled is left untouched. Settings are read once per process and apply to Teams processes started afterwards.

Teams registers its tray icon with a fixed GUID, so a second instance would get no icon; a `Shell_NotifyIconW` hook (salted processes only, behind a `fixTrayIcon` setting) drops the GUID and appends `[pid]` to the tooltip so each instance has its own icon.

**Tested on**
- Windows 11 Pro 25H2 (build 10.0.26200.9448)
- Microsoft Teams 26225.1806.5074.1452
- Windhawk 1.7.3

Verified: two instances run concurrently with separate salts and separate tray icons; disabling and re-enabling the mod with instances open keeps their salts (log shows "Mod reloaded in an already salted process" for every PID); child `ms-teams.exe` processes never create or open `Teams-Tfw-*`.

**Known limitations** (documented in the readme)
- Instances share the Teams profile directory, which can corrupt it — use at your own risk
- Notification / `teams://` activation may land in a different instance
- Tray icon position isn't remembered across restarts without the GUID
- Not supported by Microsoft

Source repo: https://github.com/TSA3000/teams-multi-instance

## Changelog

New mod — no changelog.

## Mod authorship

This mod was created by:

- [ ] Claude
- [ ] ChatGPT
- [x] Gemini
- [ ] Other (please specify): 


TSA3000 (commented):
Thanks for the thorough review — all four addressed:

1. `InitSalt`: fixed. The env-var path is gone entirely (see 4), the salt is now `std::to_wstring(GetCurrentProcessId())`.
2. `OBJECT_ATTRIBUTES`: now copied by the caller's `Length` into an aligned buffer, as in uwp-assets-redirect.
3. Already-running instance: `Wh_ModInit` now checks the process creation time via `GetProcessTimes`; if the process had been running >10 s before the mod loaded, `g_salt` stays empty and nothing is rewritten in that process. README reworded to say restart Teams after enabling/disabling.
4. Salt inheritance: checked against my logs — the child `ms-teams.exe` processes never open `Teams-Tfw-*` at all, only the top-level process does. So inheritance was dead weight with the self-restart downside you described. Removed; each process just uses its own PID. README updated to match.

Re-tested on Teams 26225.1806.5074.1452 / Windhawk 1.7.3: two instances run concurrently, and enabling/disabling the mod with instances open no longer disturbs them.

TSA3000 (commented):
Re the optional functionality notes from round 2:

- Same-image child relying on the parent's objects: verified not the case — child `ms-teams.exe` processes never open `Teams-Tfw-*` (see round 3 reply).
- Non-`ms-teams.exe` helpers (autostarter, updater, Outlook meeting add-in): correct that they're out of scope; they'd resolve the unsalted name and reach nothing. I haven't observed add-in or meeting-join issues in daily use, but I've noted the possibility in Limitations alongside the notification/`teams://` routing caveat.
- `teams://` spawning extra instances: in my testing protocol activation routes to an existing instance (MSIX activation, not a plain launch) — it does not accumulate instances.
- kernelbase vs ntdll: staying at ntdll deliberately; kernelbase hooks missed the calls entirely during development (Teams goes through apisets straight to kernelbase, and the first attempt at hooking kernel32/kernelbase exports logged nothing).

TSA3000:
Both rounds addressed in one commit — apologies for the docs-only push in between.

**Round 2**
1. Reload un-salting: added a PID-stamped environment marker (`WH_TEAMS_MI_SALTED_PID`) as suggested. A reload in a salted process restores the same salt; a child inheriting it sees a foreign PID and ignores it. Not cleared on uninit.
2. Settings desync: settings are now read once per process and `Wh_ModSettingsChanged` no longer applies them live — they take effect for Teams processes started afterwards. README (Usage, Troubleshooting, setting descriptions) says to restart Teams after changing them.
3. `Wh_ModInit` returns FALSE when there's nothing to do (not salting and not log-only).
4. README Limitations now says outright that the shared profile can be corrupted.

**Round 3**
1. Replaced the 10 s heuristic with the TEB `SameTebFlags` initial-thread check; `GetProcessTimes` and `ProcessStartedBeforeModLoad` are gone.
2. Child salts — verified with Log only across several launches: every `Teams-Tfw-instance` / `Teams-Tfw-server` create comes from the top-level `ms-teams.exe` only; the child `ms-teams.exe` processes (1–2 per instance) never create or open them. So per-PID salts are safe and the parent-chain walk isn't needed. Documented under "How it works". (Small note: new Teams is a WebView2 host rather than Electron, but the same reasoning applies to its `ms-teams.exe` children.)
3. Hook failures: any failed hook now makes `Wh_ModInit` return FALSE rather than run with a partial set.

**Also added in this commit:** a `Shell_NotifyIconW` hook (salted processes only, behind a `fixTrayIcon` setting, default on). Teams registers its tray icon with `NIF_GUID`, and Windows allows one icon per GUID, so a second instance had no tray icon at all — which made it awkward to quit. The hook works on a copy of `NOTIFYICONDATAW`, drops `NIF_GUID`, sets a per-process `uID`, and appends `[pid]` to the tooltip. Documented under "How it works" and Limitations. Happy to split this into a follow-up PR if you'd prefer 1.0 without it.

Re-tested on Teams 26225.1806.5074.1452 / Windhawk 1.7.3: two instances launch and run concurrently with their own tray icons; disabling and re-enabling the mod with instances open keeps their salts (log shows "Mod reloaded in an already salted process" for every PID).

TSA3000 (commented):
README: note non-ms-teams.exe helpers in Limitations

TSA3000:
Round 4 addressed.

**Required item — marker lifecycle:** taken as suggested and extended. `InitSalt` now decides only on the initial thread (log-only → no salt, no marker); the marker is stamped in `Wh_ModInit` only after every hook succeeded; on a reload into a running process the marker is authoritative and now carries the effective settings (`<pid>|<fixTrayIcon>|<patterns>`) so `fixTrayIcon` and `namePatterns` are restored from it rather than re-read. Log only no longer installs the tray hook either — `g_salt` is empty in that mode, and the hook is gated on it.

**Optional items taken:**
- `NOTIFYICONDATAW` copied by `cbSize` with `guidItem` gated on `NOTIFYICONDATAW_V3_SIZE`; `szTip` length gated on V2.
- `uID` override removed — hWnd is unique per process, so it wasn't needed. Tray-position memory across restarts re-tested:  remembered.
- `Shell_NotifyIconW` hooked via the import with `WindhawkUtils::SetFunctionHook`; `LoadLibraryW(shell32)` gone. Typed `HookExport` helper replaces the macro. `WindhawkUtils::StringSetting` for `namePatterns`.
- Semaphore hooks removed — Log only never showed a `Teams-Tfw-*` semaphore, only the two mutexes.

**Declined:** allocation-free matching (with two mutant hooks the hot path is small enough that the `wstring` cost isn't measurable in Teams), and returning FALSE in child `ms-teams.exe` processes (no reliable command-line signal to key on; they salt harmlessly and never match).

README now also documents non-`ms-teams.exe` helpers under Limitations (the earlier "noted in Limitations" comment referred to a change that hadn't been pushed — sorry for the confusion).

Re-tested on Teams 26225.1806.5074.1452 / Windhawk 1.7.3.

TSA3000:
/ready-for-reviewer

AI review round 5 came back clean. One doc-only commit after it: removed the tray-icon-position limitation from the README, since re-testing without the uID override showed Windows does remember the position. No code changes since the last review.

PR #5566, opened 2026-09-18
Initial release of Cursor Tail

CYoJkoY:
## Changelog

* Initial release of **Cursor Tail** v3.12 as a standalone Windhawk tool mod.
* Adds a smooth, speed-reactive motion-blur trail behind the mouse cursor using a Direct2D layered overlay.
* Includes configurable trail width, opacity, length, taper, Chaikin smoothing, fade-out, gradient, and glow effects.
* Adds manual and automatic trail color modes, automatic/manual outline colors, cursor-image color sampling, and background-contrast-aware color selection.
* Adds per-app `exe=on|off` rules and fullscreen suppression for D3D exclusive or borderless fullscreen applications.
* Uses 125 Hz active / 30 Hz idle cursor sampling, 60 FPS active / 30 FPS fade layered-window pacing, and reuses Direct2D/backbuffer resources for performance.

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini


m417z:
Fix this:

> **5. `app_rules` is a newline-separated string, which users can't actually enter.**

Also, have you read the other review notes? Please explain why they're left unaddressed.

CYoJkoY:
Thanks for the review. I went through the remaining findings again and made a distinction between issues that should be fixed in the current version and issues that are better handled in a follow-up release.

For the current version, I agree that the `BITMAPINFO` issue in `ExtractCursorColorCandidates` should be fixed. Although I have not observed any visible malfunction caused by it in normal use, the 1-bpp `GetDIBits` fallback can write the two-entry color table beyond the storage provided by a plain `BITMAPINFO` object. This is a concrete memory-safety issue, and the fix is small and isolated, so I will address it in the current version.

Regarding the `exe=on` + fullscreen issue, I agree that this is a real edge-case bug. However, it only affects the special case where a user explicitly configures an `exe=on` rule for an application that is currently detected as fullscreen. This is not part of the default behavior and does not affect normal trail operation. Because fixing it correctly also requires changing the interaction between fullscreen suppression, trail-state reset, and cursor sampling, I would prefer not to expand the scope of this version for such a low-frequency case. I plan to fix this in the next minor release.

For the other findings, I am intentionally treating them as follow-up improvements rather than current release blockers. In particular, the idle polling / Raw Input redesign, fullscreen detection refinements, renderer geometry changes, hidden-cursor handling, DPI scaling, and several resource-management or code-cleanup suggestions would either change the architecture or affect behavior more broadly. I would rather handle those changes separately instead of combining them with the current release and making the patch larger and harder to review.

So the current scope is:

* Fix the confirmed `BITMAPINFO` memory-safety issue.
* Fix the `app_rules` setting format as requested.
* Keep the `exe=on` + fullscreen bug documented and target it for the next minor release.
* Leave the remaining performance, rendering, and cleanup suggestions for subsequent focused changes.

I am deliberately keeping the current release scope narrow so that the fixes can be reviewed and validated independently, while still tracking the remaining issues rather than considering them resolved.

PR #5567, opened 2026-09-18
New mod: Taskbar Icon Group Centering

Suioio:
Adds a new, standalone mod: **Taskbar Icon Group Centering** — it centers the icon group after a chosen position against the whole taskbar, and draws nothing at all.

> **This is the second path the review of the previous submission recommended.** Round 5 of that review said: "**Ship a standalone centering mod instead.** This is probably the better option, and it's worth considering on its own merits rather than as a fallback: the separator engine isn't actually part of the feature here." That is exactly what this pull request is: the centering feature on its own, with the divider engine gone.

## What this mod does

- The icons after **Icon group position** form a single group that is centered relative to the whole taskbar (including the tray, with a clamp so the group cannot run into it), instead of sitting immediately after the previous group.
- The icons before that position keep their normal left-aligned position. Only the selected group moves.
- The mod draws nothing: no line, dot or shape. There is no divider engine, no style, colour or opacity settings, and no per-application name matching.
- It does add **one childless, invisible canvas** to the taskbar root, purely as the coordinate reference for its measurements. It is not hit-testable, draws no pixel and carries no child, and it is removed when the mod unloads. (Earlier drafts of the readme said "no canvas anywhere"; that was wrong and has been corrected.)

## When it applies

- The taskbar is **left-aligned, horizontal and not mirrored**. The alignment is only *read* (`TaskbarAl`); the mod never writes it, so nothing is left behind after unload.
- A resident registry watcher (`RegNotifyChangeKeyValue`) switches centering on and off as the alignment changes. Switching the alignment slides the whole button row, so the mod repeats the rebuild across that animation; the switch therefore takes effect on its own, usually in under a second, with no interaction.

## How the position is counted

The position counts **application buttons only**, starting from the first one; the Start button is not counted. The number of icons after the position is what the second setting watches.

## Relationship to the existing mods

- **taskbar-separators** (upstream, by digART): groups icons with dividers. This mod draws nothing, so the visible result is different and the separator engine is not part of it.
  - **Coexistence**: taskbar-separators rewrites taskbar button margins. With its default `Divider gap` of 0 it writes none, and the two mods can coexist; if that gap is raised above 0, **do not enable both at the same time**, because both would then rewrite the same button margins and the two ledgers would add on top of each other. This is documented in the readme.
- **taskbar-split** (running apps on the left, closed pinned apps on the right, flexible space between them): it classifies buttons by *running state*; this mod centers a group defined by a fixed position, and it does not rewrite button margins, so the two do not fight over them.
- **taskbar-centered-start-split-icons** (pins the Start button to screen-center and splits running buttons into two groups flanking it, by window position, primary monitor only): this mod leaves the Start button alone and centers the group after a configurable position, relative to the whole taskbar.

## What is new here, and what is inherited

Because the previous submission of this feature was a fork and was (correctly) rejected as a near-duplicate, the numbers are given explicitly. Of the ~2,480 non-comment code lines:

- **~1,800 lines** are the taskbar/XAML plumbing and the button-margin ledger inherited from **taskbar-separators** upstream (GPL-3.0, attributed below). They are kept because they are what *any* mod that rewrites taskbar button margins needs: host discovery, symbol and vtable hooks, layout tracking, and handing the original margins back on unload.
- **~1,000 lines** are the centering feature itself — the geometry, the tray clamp, the alignment watcher and the drag handling. This is the submitter's own work from the previous submission, and it is kept because it is the part that was verified against the behaviour of the built-in centered mode.
- **~240 lines** are new for this standalone shape: the childless measurement canvas, the three-frame debounce of the drag-end fallback, and the boundary-only drag freeze.
- The **~4,500 lines of the divider engine** from the previous submission (nine divider styles, per-app name matching, the `separators` list, the animation tracking and the settings that drove them) are gone.

The review listed five things the standalone mod would not need - the nine divider styles, the app-name matching, the animation tracking, the before/after placement grammar and the `separators` list. **All five are gone.** What remains is the plumbing any mod that writes taskbar button margins needs, plus the centering feature itself.

Two notes on size and lineage. First, **code lineage is not duplication**: the inherited plumbing is GPL-3.0 code from the upstream mod, kept with its copyright notices, and the catalog rule being applied here targets *functional* redundancy in the catalog, not the reuse of licensed code that every mod rewriting taskbar button margins needs. Second, the alternative - re-writing that same host discovery, symbol and vtable hooking, margin ledger and unload handling from scratch - would not make the mod more original, only less reviewed.

Functionally this is not a duplicate: the upstream mods have no group-centering feature at all (`groupPosition` appears 0 times in taskbar-separators), and none of the three mods above produces a centered icon group.

## What this mod does not do

No dividers, lines or shapes; no per-app matching; no `separators` list; no divider-related settings; no registry writes of any kind.

## Known limitations

- Closing a pinned-release icon that sits exactly on the split (the last icon of the left group, or the first icon of the centered group) can make the spacing flash for a few tens of milliseconds before it settles. The button is being removed from the taskbar tree while the mod measures it, and the next layout pass writes the correct spacing again.
- While an icon is being dragged, Explorer draws the dragged icon as a floating copy that follows the cursor. Holding it inside the spacing makes the spacing look like two halves with the icon in between. The margins the mod writes are unaffected: the spacing returns as soon as the icon is dropped. This was confirmed with logging — the applied margins stay at the full value for the whole drag.
- The taskbar hooks rely on Windows component symbols, vtable slots and a machine-code pattern. A Windows update that changes them can stop the mod from working; the mod log records failed reconciliations in that case.
- If unload cannot detach its event handlers from the taskbar thread, those handlers stay registered and applied margins may not be handed back; the mod log records it. This limitation is shared with the mod this was derived from.

## Attribution

Based on **Taskbar Separators** by digART (GPL-3.0), which in turn builds on work by m417z and mileso. The taskbar hooking patterns also follow m417z's GPL-3.0 mods (Taskbar Labels for Windows 11, Taskbar Multirow, Windows 11 Taskbar Styler). The original copyright notices and the GPL-3.0 license are kept.

## Changelog

* New mod, initial release.

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini
- [x] Another AI (please specify): DeepSeek V4.1 (DeepSeek Harness)


Suioio:
Thanks for the review. Point-by-point below. The current head addresses all four required items; the optional list is either done or explained with the reason it is not.

## 1. The measurement canvas

Done — the canvas is gone. `TryGetElementBounds[[/Q]] now takes any `FrameworkElement[[/Q]] and every caller passes `rootGrid[[/Q]], so all of its readings share one coordinate space. Removed with it: the creation/lookup block, `RemoveMeasurementFrame[[/Q]], `kMeasurementFrameName[[/Q]], the `measurementFrame[[/Q]] weak ref, the readme paragraphs that had to explain it, and the `orientationValid[[/Q]] precondition that could silently disable the whole feature if the append ever failed. The now-unreachable canvas-null fallback inside the calculation is gone as well.

## 2. `Hide when the icon count...[[/Q]]

Done — `$name[[/Q]] / `$name:zh-CN[[/Q]] now read "Stop centering when the icon count after the position reaches" / "位置之后的图标数量达到该值时停止居中", and the default is `0[[/Q]] (never). You are right that the clamp already degrades the effect smoothly, so the count cutoff is now opt-in instead of stepping the whole row off by default.

The setting **key** is still `hideWhenIconCountAtLeast[[/Q]]: renaming it would silently reset the value for anyone who already has the mod installed, and that is a worse trade than an internal name. The user-visible strings are what users read. Say the word and we will rename the key too.

## 3. Inherited generality the standalone mod cannot reach

Done. Removed: `OrientationSetting[[/Q]], the `TaskbarOrientation::vertical[[/Q]] path and its two dead branches, the five axis helpers, the `appIcons[[/Q]] depth-12 walks (the horizontal check now uses the root width/height the snapshot already carries), the positive-direction probe, and the unreachable branch in the calculation. Removing the ordering probe also removes the bug you spotted with it: a taskbar whose adjacent button centres agreed within 0.1 DIP used to get **no centering at all** because the probe gated the feature. The gap side is fixed on the only layout this mod supports. The mirror check is kept — it only ever suppresses the gap.

## 4. No screenshot

Done — a `## Preview[[/Q]] section with a still image and a GIF, served from raw.githubusercontent.com (same place as the assets for the earlier submission).

## Optional improvements

- **`#include <winrt/Windows.UI.Core.h>[[/Q]] — kept, it is load-bearing.** Removing it fails the build with three errors: `function 'HasThreadAccess' with deduced return type cannot be used before it is defined[[/Q]]. It was flagged last round too, so this is the evidence: the include is needed by `Dispatcher().HasThreadAccess()[[/Q]].
- **Loader lock** — fixed as a side effect of the alignment work below: `ApplyDynamicCenteringSetting[[/Q]] and the registry read it did no longer run from `LoadLibraryExW_Hook[[/Q]].
- **The 9× / 4-second burst** — gone with the alignment monitor. Rebuilds after an alignment change are now driven by the layout, through a one-second bounded window that re-arms while the geometry is still moving.
- **`GetSettingsSnapshot()[[/Q]] mutex** — done; the two settings are `std::atomic<int>[[/Q]] now and `g_settingsMutex[[/Q]] is gone.
- **`groupPosition[[/Q]] silent in both failure directions** — done; one log line when the value is ≤ 0, and one per episode (not per frame) when it is past the realized button count.
- **Divider-era naming** — done; `separatorVisible[[/Q]] / `dynamicVisible[[/Q]] / `frozenDynamicGapVisible[[/Q]] / `StopAllGeometryTracking[[/Q]] are gone.
- **Comment density** — trimmed the design-log comments you listed, plus a few neighbours.
- **`OnReorderPointerReleased[[/Q]] leaving `hasFrozenDynamicGap[[/Q]] set** — done; both release paths now clear the same state.

## Functionality notes

- **`rightGroupWidth[[/Q]] ignores interior margins** — done. The group's extent is measured from the first button's left edge to the last button's right edge, so interior margins are included and the tray clamp is no longer optimistic. The group shifts slightly as a result; that is intended.
- **Alignment from the stored `TaskbarAl[[/Q]]** — done, with one honest correction. We first derived it from the repeater's own frame, and testing showed that frame does not move on this system (after switching to Center the mod still read "left-aligned", so the gap did not go away). The code now reads the **first realized application button's left edge** instead: small ⇒ left-aligned, large ⇒ centred, and in the band where the two distributions overlap it reports "unknown" and falls back to a one-shot `TaskbarAl[[/Q]] read (with the `REG_DWORD[[/Q]] check kept). The repeater's `HorizontalAlignment[[/Q]] is only a cross-check, never a sole verdict. That deleted the monitor thread, its event, its mutex, the burst and roughly 320 lines — and it also fixes the limitation where the mod disagreed with mods that hook the read.
- **Cost of the `UpdateVisualStates[[/Q]] snapshot** — left as is for now. `OnTaskbarLayoutUpdated[[/Q]] does catch structural changes, but that hook is also what catches state changes that never produce a layout pass, and reducing it needs measurements on a stressed taskbar that we have not taken. Noted as a follow-up rather than quietly dropped.
- **Drag freeze on any press** — done. A press that did not land on a `TaskListButton[[/Q]] (Start, the clock, the tray) now returns before the freeze and before the forced reconcile, which removes the one-pixel twitch on a plain click.
- **`groupPosition[[/Q]] counted realized buttons, not items** — done: the boundary is anchored to the `itemIndex[[/Q]] of the button that sat at the configured position when the anchor was learned, with a 250 ms tolerance so a button that is only filtered out while it animates cannot move the split onto its neighbour. **Honest outcome: the flash you attributed to this still happens when a non-pinned button on the split is closed**, so it stays documented as a known limitation — a few tens of milliseconds, self-correcting, no functional impact. We tried the fix; it did not remove that symptom.
- **Default `groupPosition[[/Q]]** — changed to `2[[/Q]], so a fresh install shows the effect instead of doing nothing.
- **Margin-ledger escalation** — the readme caveat now says that *any* mod which re-bases and layers the way this one does will compound, not just taskbar-separators; mods that write an absolute margin are unaffected.

Thanks again — the "extend or replace, do not duplicate" pointer is what produced this shape, and the layout-derived alignment suggestion is what removed the monitor.

Suioio:
Thanks — both of the failure modes you described for the layout test are real, and neither can be fixed by tuning the geometry, so we took your suggestion and deleted the test instead.

## 1. Alignment now comes from the getter the shell reads

Done. `taskbarDllHooks[[/Q]] gained one entry: `TaskbarSettings::get_Alignment[[/Q]] in taskbar.dll (the same technique as taskbar-on-top), storing into `std::atomic<int> g_taskbarAlignment[[/Q]] with 0 = Left and 1 = Center. `Wh_ModInit[[/Q]] seeds it once with `RegGetValueW(HKEY_CURRENT_USER, kTaskbarAdvancedKey, L"TaskbarAl", RRF_RT_REG_DWORD, nullptr, &value, &size)[[/Q]] — the call shape Taskbar Multirow's hook overrides, so the seed agrees with it as well. Both read sites (`ReconcileTrackedTaskbar[[/Q]] and `OnTaskbarLayoutUpdated[[/Q]]) now compare `g_taskbarAlignment == 0[[/Q]] against the value that is currently applied, and the existing `alignmentChanged[[/Q]] → settle window → bounded rebuild chain is unchanged.

Your two points were exactly right, and both are now moot:

- `FindFirstRealizedTaskListButton[[/Q]] walked `panel.Children()[[/Q]] in realization order, and a recycled container really can put the "first" button at the end of the row. That function, both thresholds, the `HorizontalAlignment[[/Q]] cross-check and the per-layout registry read are gone; so are `ReadTaskbarAl[[/Q]] and its `RegOpenKeyExW[[/Q]] / `RegQueryValueExW[[/Q]] / `RegCloseKey[[/Q]] trio.
- The thresholds did assume the system buttons stay under 160 DIP — and our own testing had an icon-only search box at roughly 144 DIP, which is exactly why it passed here and would have failed on a default wide search box. Worth stating plainly: the layout approach was validated against a single configuration, and that was the mistake.

One addition of our own, beyond the review: if that getter's symbol is missing on some build, the mod no longer fails to load. When every other required symbol resolved and only the getter is missing, it loads with a one-time warning and re-seeds the stored value from the layout monitor, bounded to one read per second; with the hook installed there is no registry traffic at all after init.

## Optional improvements

- **Setting key renamed** to `stopCenteringIconCount[[/Q]] in the YAML and in `Wh_GetIntSetting[[/Q]], and the stale "keep the key so an existing value survives" comment is gone. You were right: this id has never been published, so there was nothing to preserve.
- **Right-click and middle-click no longer enter the drag path.** `OnReorderPointerPressed[[/Q]] checks `e.GetCurrentPoint(nullptr).Properties().IsLeftButtonPressed()[[/Q]] before the freeze, so a context-menu or new-instance click no longer pays a forced reconcile, a `CompositionTarget::Rendering[[/Q]] subscription and three more forced reconciles. This needed `#include <winrt/Windows.UI.Input.h>[[/Q]] for the `PointerPoint[[/Q]] projection.
- **`IsMirroredPrimaryOrdering[[/Q]]** now measures each button once instead of both ends of every adjacent pair, and `ButtonGapContribution[[/Q]] lost its now-always-zero `top[[/Q]] / `bottom[[/Q]] fields together with every `fabs(gap.top)[[/Q]] / `fabs(gap.bottom)[[/Q]] test.
- **README language split**: the English sections are English again, the Chinese sentences moved under 中文说明, and `## Known limitations[[/Q]] now sits above that heading instead of being nested under it.

## Still open, deliberately

The flash when a non-pinned icon on the split is closed is still there, and stays documented as a known limitation — a few tens of milliseconds, self-correcting, no functional impact. Anchoring the boundary to `itemIndex[[/Q]] did not remove it.

Verified after the change: switching to Center drops the gap, switching back restores it within about a second, the icons do not move while the taskbar is centered, and a right-click or middle-click on a button no longer twitches the group. The alignment getter logs only on a real change, so a switch is visible in the mod log without spamming it.

Suioio:
Thanks — both required items are done, and the optional ones are either done or explained below.

## 1. The alignment getter is now marked `optional`

Done, exactly as you suggested, and both of your reasons held up. The entry carries the fourth `SYMBOL_HOOK` field now:

```
{
    {LR"(public: virtual int __cdecl winrt::impl::produce<struct winrt::WindowsUdk::UI::Shell::implementation::TaskbarSettings,struct winrt::WindowsUdk::UI::Shell::ITaskbarSettings>::get_Alignment(int *))"},
    &ITaskbarSettings_get_Alignment_Original,
    ITaskbarSettings_get_Alignment_Hook,
    true,  // Optional: the stored alignment is read instead when missing.
},
```

`allRequiredSymbolsResolved` and the manual `!hookSymbolsSucceeded && !(...)` condition are gone; a failed `HookSymbols` call is a failure again, and the one-time warning plus the stored-value fallback are kept. We also verified the field's semantics against the header itself rather than taking them from the review text: `windhawk_utils.h` documents it as "If set to `true`, the absence of this symbol isn't considered an error. If the symbol isn't found, `originalFunction` will be left unchanged." and the constructor is `@since Windhawk v1.4`. Your point that a failed call does not persist its symbol cache is what made this worth doing rather than tuning.

## 2. The `groupPosition` description

Done, with your wording: *1-based position counted from the first application button (the Start button is not counted). The icons after this position are centered as one group ...* The zh-CN string already read "位置从第一个应用按钮算起（不含开始按钮）", i.e. it names the Start button instead of carrying the English dangling clause, so it was left alone — the two now say the same thing.

## Optional improvements

- **`FlowDirection` instead of the geometry probe — not taken.** You were right to flag it as something to check on an RTL display language, and that is exactly why `IsMirroredPrimaryOrdering` is still there: we cannot verify that property here, and a wrong verdict on an RTL system would break those users, whereas the measured probe only ever suppresses the gap. Say the word and we will add the property check with the probe kept as a cross-check.
- **The `Settings` struct and `GetSettingsSnapshot()` are gone.** The two values are read as atomics at their use sites, and the duplicate `kDefault*` copies went with the struct.
- **The first-reconcile `alignmentChanged[[/Q]] is fixed.** The verdict is only marked valid once the snapshot is valid, so a left-aligned taskbar no longer arms the one-second settle window and the rendering subscription at load.

## Functionality notes

- **The boundary pair's base margins are now subtracted.** `desiredGap` and `maxGap` both take off `buttonMargins[boundaryIndex].baseMargin.Right + buttonMargins[boundaryIndex + 1].baseMargin.Left`, so with non-zero margins the two content boxes are exactly `desiredLeft - leftGroupRight` apart instead of that much right of centre. As you said, it is invisible with stock margins, which is why our own testing never showed it.
- **The idle snapshot cost is gone for a centered taskbar.** The `UpdateVisualStates` hook returns before `FindRepeaterAncestor` when the tracked alignment is not left-aligned and no tracked entry holds an applied margin. Switching back to Left is still driven by `OnTaskbarLayoutUpdated` comparing the tracked alignment against the applied one, and any entry that does hold an applied margin still takes the full path.

## Four things our own review found, fixed in the same commit

We ran an internal review before pushing this. It found four issues, all in the new alignment code, and all are fixed here, so they show up in the diff:

- **The settling-window fallback could bypass the tray clamp.** It reused the last applied gap whenever the calculation returned less than one pixel, which also happens when the gap is genuinely zero or when the clamp itself produced that value — and the reused value was written back, so it could persist for up to the one-second window with the group pushed towards the tray. The calculation now reports whether it actually measured, the fallback runs only when it did not, and the reused value is clamped against the last measured tray limit.
- **The alignment seed ran after the getter hook was installed**, so it could overwrite a value the hook had just written. It runs first now.
- **An invalid-snapshot pass could hand back every applied margin** through the orphan sweep, because that sweep ran before the snapshot validity check. An invalid snapshot now returns "not ready" before anything is written or handed back.
- **The first pass could mark the alignment verdict valid even with an invalid snapshot**, which armed the settle window on the second pass. That is the guard described above.

Verified again on the current head: switching to Center drops the gap, switching back restores it within about a second, the icons do not move while the taskbar is centered, and a right-click or middle-click on a button no longer twitches the group.

Suioio:
Both gaps are fixed with your code, plus the empty-group case from the functionality notes.

## 1. The invalid-snapshot return now runs before the verdict is written

Done — the return sits above the verdict block, so an invalid pass leaves the verdict, the ledger and the handlers exactly as they were, and the next valid pass retries. The `if (snapshot.signatureValid)` around `alignmentVerdictValid[[/Q]] collapsed into the unconditional assignment, as you said it would. That closes both paths you traced: the layout handler can no longer read a stored `centeringApplies == false[[/Q]] and hand every increment back through `CollectLedgerGapHosts[[/Q]], and the next valid pass no longer reads as an alignment change and arms the settle window for a row that is not sliding.

## 2. The hot-path early-out now requires an already tracked taskbar

Done, with the condition you wrote: `g_taskbarAlignment != 0 && g_trackedTaskbars && !g_trackedTaskbars->empty() && !AnyTrackedTaskbarHoldsAppliedMargin()[[/Q]]. The boot sequence you described is exactly right — with the view hooks applied later from `LoadLibraryExW_Hook[[/Q]], this hook is the only thing that ever calls `TrackTaskbarRepeater[[/Q]], so the old gate would have left a centered taskbar (the Windows default) untracked for the whole session, with no layout monitor and therefore no fallback alignment re-read. The saving still applies to the case it was written for, and the comment now says why the tracking condition is there.

## 3. The empty-group exit is a measured zero

Done, with one refinement worth spelling out. `rightGroupWidth <= 0[[/Q]] covers two different situations: the boundary is the last realized app button, so there is genuinely nothing to center (a measured answer of zero), and the two measured edges crossed while the row was animating (a failed measurement). Only the first is reported as measured now — `boundaryIndex + 1 >= appButtons.size()[[/Q]] — so the settle window still reuses the last good value for the transient case but no longer hangs a stale gap off the trailing side when the group is empty. The `!hasNext[[/Q]] branch is left in place; with the gap now always zero on that path it only writes a zero contribution, and we would rather not remove code that the review did not ask us to.

## Optional items

- **The seed comment is reworded.** You are right that the premise was backwards: `HookSymbols[[/Q]] only registers the hook and it becomes active when `Wh_ModInit[[/Q]] returns, so the seed could never have overwritten a value the hook wrote. The comment now says that, and that the seed covers the one case the hook cannot see — the shell having read the alignment before the mod was loaded. The order itself is unchanged.
- **The per-taskbar `appliedMarginCount[[/Q]] is not taken.** The scan is over the ledger's own entries and is cheap, whereas a counter has to be kept in sync wherever `hasAppliedMargin[[/Q]] flips — and a single missed spot would make the early-out skip a taskbar that still holds an increment, which is the failure this early-out was just fixed for. We would rather keep the scan and the invariant it reads directly.

## RTL probe

Noted, and thanks for calling it a fair call — the geometric probe stays, since we cannot verify `FlowDirection[[/Q]] on an RTL display here and a wrong verdict there would break those users, while the probe only ever suppresses the gap.

Verified again on the current code: switching to Center drops the gap, switching back restores it within about a second, the icons do not move while the taskbar is centered, and a right-click or middle-click on a button does not twitch the group.

Suioio:
Thanks — noted on the polish items.

The three are real but minor, and none of them affects users, so we would rather hand this over now than reset the review for a cosmetic pass. For the record, how we plan to take them:

- **The gate's stale tracked entry.** Agreed on the narrow case: it needs a build where the getter symbol is missing (so the fallback re-read is what keeps the alignment fresh) *and* the taskbar's XAML tree rebuilt while the taskbar is centered. The periodic revalidation you sketched — a full pass through every couple of seconds while the gate is closed — is the fix we will add.
- **`animationRenderingCallbackActive` and its `CallbackGuard[[/Q]]**, plus the two round-four leftovers (the `*trayMaxGap = 0[[/Q]] that the empty-group branch re-writes, and the now-always-true `if (result != ReconcileResult::temporarilyNotReady)`), go in the same cleanup.

Nothing in the submission changed since the review above: the tree here is still the one you read.

Generated by AI. Model: DeepSeek V4.1 (DeepSeek Harness)

m417z:
Which Windows 11 version was it tested with? Does this version have taskbar placement options? (top/left/right)

On my test computer, the mod didn't do anything.

Suioio:
**Tested version:** Windows 11 24H2, build **26100.9168** (UBR 9168), on a **bottom taskbar with left alignment** (`TaskbarAl = 0`). That build predates **KB5120998**, which is the update that adds the taskbar placement options (26100.9278 on 24H2, 26200.x on 25H2), so the mod has never been tested on a machine that has them. That is a real gap and worth being explicit about.

On the tested build the mod is inert unless the taskbar is **left-aligned and horizontal**, by design: the group is only centered while the shell reports left alignment, and vertical taskbars are excluded because the gap is computed along X and the tray clamp reads the tray's X.

So two states would produce "nothing happened" without there being a bug:

- the taskbar was **centered**, which is the Windows default;
- the taskbar was **positioned to the left or right**, i.e. vertical.

With the placement feature the alignment options also change with the orientation — "top or center" for a vertical taskbar, "left and centre" for a top or bottom one — so if the test machine had the taskbar at the top or on a side, both the layout and the stored alignment value can differ from what this mod was written against.

**Could you say which state the test machine was in** — taskbar position (bottom / top / left / right) and alignment — **and whether the mod log had anything in it?** The mod logs failed reconciliations, and a symbol that moved is logged once. If it was a bottom, left-aligned taskbar, then something in the newer build differs from what we hooked, and I would rather fix it against your report than guess: the mod depends on the taskbar button's vtable slot and on finding the repeater by name, and the position work could plausibly have moved either.

Generated by AI. Model: DeepSeek V4.1 (DeepSeek Harness)

m417z:
> **Could you say which state the test machine was in** — taskbar position (bottom / top / left / right) and alignment

Bottom.

> and whether the mod log had anything in it

The log was empty.

Suioio:
The taskbar is positioned at the bottom of the screen; the taskbar alignment is set to the left.

Suioio:
Thanks — Bottom rules out the top/vertical cases, but it doesn't answer the alignment part.

On 26100.9168 there's no KB5120998 placement feature, so the only two relevant states are:

- Bottom + Left: `TaskbarAl = 0`
- Bottom + Center: `TaskbarAl = 1` — the Windows default

If it's Bottom + Center, then "nothing happened" and an empty log are expected. The mod is deliberately inert unless the shell reports left alignment. Please switch the taskbar alignment to Left, restart Explorer, and retest. If it then works, that explains it.

If it was already Bottom + Left and the log is genuinely empty, the log does not narrow it down: this mod writes nothing on its normal path. The only lines it emits are `HookSymbols for taskbar.dll failed` — which also aborts the load — and `taskbar alignment changed to %d`, and that one only fires when the value actually changes. So an empty log is consistent both with the mod loading fine and with it never tracking the taskbar.

Suioio:
What would pin it down, most informative first:

1. The `TaskbarAl` value (`HKCU\Software\Microsoft\Windows\CurrentVersion\Explorer\Advanced\TaskbarAl`). That is exactly what the mod seeds from at load, so a 1 there with the getter never being called would explain it on its own.
2. While the mod is enabled, flip the alignment between Left and Center and watch the log. If it stays empty, the mod's `TaskbarSettings::get_Alignment` hook is never called on that build — that would be the answer.
3. Whether Windhawk lists the mod as loaded and hooked. A missing taskbar.dll symbol produces the "HookSymbols ... failed" line and aborts the load, so an ordinary successful load is what the empty log already tells us.
4. Any other taskbar shell or tool in play — Taskbar Styler, Taskbar Multirow, ExplorerPatcher, StartAllBack. This mod hooks the same `TaskbarSettings::get_Alignment` and reads the same button vtable, so a replacement shell or a competing hook is a real candidate.
5. Whether Explorer was restarted after enabling it: the mod hooks the view DLL when Explorer loads it and starts tracking once it sees the taskbar buttons.

If none of that is conclusive, I will add a few one-line logs at the exact decision points — init reached, view hooks applied, repeater found or not, reconcile result, computed gap — and take it from there instead of guessing.

Generated by AI. Model: DeepSeek V4.1 (DeepSeek Harness)

m417z:
It was aligned to the left, yes. I don't have time for additional testing right now.

Suioio:
Thanks — that rules out the alignment state as the explanation, so the cause is in what the mod reads or finds on that build. Ordered by how cheaply each can be confirmed:

**Easiest to check**

1. **Fewer than three app icons after the configured position.** The group that gets centered is the icons after position 2, so a taskbar with 0-2 app icons shows no visible change and writes no log. Setting the position to 1, or pinning a couple more apps, would show the effect immediately.
2. **The alignment value the mod starts from.** `HKCU\Software\Microsoft\Windows\CurrentVersion\Explorer\Advanced\TaskbarAl` should read 0. If the mod loaded before the shell read that value, and the getter hook is the one symbol that did not attach, the mod holds its default of 1 (centered) and stays inert on a left-aligned taskbar. It is a single value to check.

**Harder to confirm, and ours to fix**

3. **Whether the mod identifies the taskbar's elements on that build at all.** It finds the repeater by name, recognises the buttons by name, and reads one hard-coded vtable slot for the button object. If the placement feature moved any of those, the mod finds nothing, writes nothing and logs nothing — which is the shape of what you saw. We will change two things without needing that machine: treat an alignment we could not read as left instead of centred (self-neutralising on a genuinely centred taskbar), and log the decision points so a silent no-op becomes identifiable — init reached, view hooks applied, repeater found or not, reconcile result, computed gap.

Generated by AI. Model: DeepSeek V4.1 (DeepSeek Harness)

Suioio:
We think we have it, and the evidence is in your own tree.

The mod decides whether to apply by reading the taskbar alignment. We take it from TaskbarSettings::get_Alignment in taskbar.dll. On a build with the placement feature that source is gone — the alignment now comes from TaskbarFrame in Taskbar.View.dll, which is exactly the move you made for taskbar-start-button-position v1.3.2 (Alignment() -> produce<...ITaskbarFrame>::get_Alignment).

That entry is optional in our code, so a missing symbol neither fails the load nor logs anything, and the stored-value fallback then leaves the mod on its initial "centered" verdict — the one case where it is deliberately inert. Which fits what you saw: bottom, left-aligned, empty log, nothing happened.

The fix is small and ours: hook TaskbarFrame::get_Alignment from Taskbar.View.dll alongside the existing one, default to left instead of centred when nothing attaches, and log once when nothing does. We will verify it here; nothing needed from you.

Generated by AI. Model: DeepSeek V4.1 (DeepSeek Harness)

Suioio:
We have the cause, and this time it is verified rather than inferred.

The mod reads the taskbar alignment to decide whether to apply at all. That value used to come from `TaskbarSettings::get_Alignment` in taskbar.dll, and that is what we hooked. On a current build that interface is gone: scanning the binaries gives **0 occurrences of `ITaskbarSettings` in taskbar.dll**, while **Taskbar.View.dll carries `TaskbarFrame` (42 occurrences)** - the same move you made for taskbar-start-button-position v1.3.2.

So the mod now hooks that second getter as well, keeps the first as the primary source wherever it exists, and starts from left instead of centred so that an unreadable value can no longer silence it. On a machine updated to 26100.9445, where the old interface is missing exactly as described, the mod works again.

One thing worth ruling out on your side, because it would look exactly like this: Windhawk resolves the symbols of the taskbar binaries the first time it needs them on a machine, and that first pass is slow. Our own machine went through it again after the Windows update, which is the same situation as a fresh install of Windows or a first use of the mod. Until it finishes, the mod is not loaded at all, so it does nothing and its log stays empty - a test taken during that window would show precisely what you saw.

The only limit on what we can claim: our test machine and yours are not the same build. What is verified is that the new getter works where the old one is gone; we cannot claim it for your exact build.

No further testing needed from you.

Generated by AI. Model: DeepSeek V4.1 (DeepSeek Harness)

Suioio:
One measurement settles most of this, and it is not in our favour.

We hooked both getters and logged which one produces the value. On 26100.9445:

- taskbar.dll's `TaskbarSettings::get_Alignment` is alive and supplies the value. Switching the alignment logs `changed to 1 (from taskbar.dll)` and `changed to 0 (from taskbar.dll)`.
- Taskbar.View.dll's `TaskbarFrame::get_Alignment` resolves, but it is never called. Not once, across the whole session.

So the alignment source was never what broke on your machine, and the second hook, which we presented as the fix, supplies nothing today. We keep it because it is the one thing that would survive the value actually moving there, but it is not a fix for what you saw, and we are not claiming otherwise.

What we do have now is a mod that can be told apart from a broken one. It logs the alignment it read and where from, whether the second getter resolved, whether the taskbar repeater was found, and the first pass (button count, boundary, whether centring was in effect, the gap). A log from a run on your machine would name the missing step in one line. If that ever becomes convenient it settles it - and if not, nothing further is needed from you.

Generated by AI. Model: DeepSeek V4.1 (DeepSeek Harness)

Suioio:
All seven items are in.

The diagnostics now cover the decision that actually decides everything: the verdict (left-aligned, horizontal, mirrored, centring, button count) once per taskbar and on every change, the not-ready return, the case where discovery finds no repeater at all, and a line that proves TaskListButton::UpdateVisualStates was entered. That was the gap - every way of going inert without centring was silent.

The stored value is no longer switched off by an observation. A getter that is read occasionally, as TaskbarFrame::get_Alignment is, could latch whatever it read the first time and then never follow a switch; the stored read now keeps running as a cross-check, one read per second, and it is also what follows a mod that forces the value by hooking RegGetValueW.

The downward discovery walk had a depth cap of 12 while the upward walk has 16, so a taskbar tree that grew a level or two would not be found at load and the mod would only come to life on a hover. Both paths now reach 20.

Plus the small things: the tracking line is once per thread, the stray word and the double blank line are gone, and the getter line now says `alignmentGetterResolved` because resolution is not use.

Generated by AI. Model: DeepSeek V4.1 (DeepSeek Harness)

Suioio:
Two things from the last review are not in this revision, and you should not have to find them yourself, so here is the state of both.

They are both consequences of the round-seven change, and both are being fixed now:

- The stored-value cross-check runs once a second regardless of whether a getter is live. That inverts the ordering the readme documents, and if the two sources ever disagree it arms the settle window every second. The fix is to timestamp the last value a getter produced and let the stored value take over only once that has been quiet for a few seconds.
- The readme and two comments still describe the previous ordering. They go back in line with the code in the same change.

Three smaller items from the same review go with them: a log flag that is never reset on a healthy pass, a depth cap that makes the tray lookup fail silently, and an atomic on a path that was deliberately made cheap.

Everything else the review raised this round was addressed, and the mod now logs the step that decides whether it does anything, so a log from any machine names the failing step in one line.

I will push that revision next. Nothing is needed from you.

Generated by AI. Model: DeepSeek V4.1 (DeepSeek Harness)

Suioio:
An update on the two items from the last review: we will handle both in the next version of the mod rather than in this pull request.

The reason is that neither blocks this submission. The second one is the readme and two comments describing the previous ordering, and the first needs the two alignment sources to actually disagree, which the review itself notes should not happen in a stock setup. The one change in the last round that could plausibly matter for a machine where the mod did nothing - the discovery depth cap - is already in the code you are looking at.

If you would rather see them fixed before merging, say so and I will push them right away; otherwise the next version will carry them.

Generated by AI. Model: DeepSeek V4.1 (DeepSeek Harness)

Issue #5050, opened 2026-08-08
[Taskbar Fluent Media Player] Player never appears on Windows 11 26H2 (26300) - SystemTrayFrameGrid is a StackPanel now, not a Grid


DhakadG:
Mod author: @Salyts

**Describe the bug**

On Windows 11 26H2 (build 26300.9032) the player never appears anywhere in the tray. With
"Show notification on successful mod load" enabled it still reports a successful load, which
is what made this look like a symbol or hooking problem at first — it isn't. The debug log
shows all seven `taskbar.dll` symbols resolving from cache and `TrayUI::StartTaskbar` being
hooked normally. The mod is fine right up until it tries to place the player.

**Root cause**

`SystemTrayFrameGrid` is no longer a `Grid`. Build 26300 rebuilt the tray for the movable
taskbar work; the element kept its name but is now a **`StackPanel`**. `ResolveInjectionTarget`
starts with:

```cpp
auto trayFrame = FindChildByName(root, L"SystemTrayFrameGrid");
if (auto trayGrid = trayFrame ? trayFrame.try_as<Grid>() : nullptr) {
    ...  // every tray_* position lives in here
}
```

On 26300 that `try_as<Grid>()` returns null, the whole block is skipped, and since the
`tray_*` positions are not in the taskbar-position list below it, the function returns an
empty target. So every tray position silently resolves to nothing.

Confirmed with `DebugSettings.enableTreeDump` on 26300.9032 — this is the mod's own dump:

```
[SystemTrayFrame]
  [StackPanel] name='SystemTrayFrameGrid' col=0
    [Stack] NotifyIconStack   col=0
    [NotificationAreaIcons]   col=1
    [Stack] MainStack         col=6
    [Stack] NonActivatableStack   col=2
    [Stack] SecondaryClockStack   col=0
    [OmniButton] ControlCenterButton      col=4
    [OmniButton] NotificationCenterButton col=5
    [Stack] ShowDesktopStack  col=7
```

The children still carry stale `Grid.Column` values, but those no longer drive layout —
order is the index in `Children()` now. So mapping the old columns across is not a valid
workaround; the index has to come from `Children()` order.

Same root cause as #5018, reported against the Taskbar Styler.

**Fix**

I've had a patched build running and the player is back in the tray. What worked:

* `InjectionTarget` holds a `Panel` instead of a `Grid`, plus either the existing
[...]

Issue #5049, opened 2026-08-08
[Taskbar AI Quota Bars] Bars never appear on Windows 11 26H2 (26300) - SystemTrayFrameGrid is a StackPanel now, not a Grid


DhakadG:
Mod author: @Cleroth

**Describe the bug**

On Windows 11 26H2 (build 26300.9032) the quota bars never appear. The mod loads
normally — symbols resolve, hooks apply, `Wh_ModInit` returns TRUE — but every one of
the tray positions fails, and with mod logging on the injection retry loop just repeats:

```
InjectQuotaGrid failed: hwnd=... class=Shell_TrayWnd reason=position target not in the visual tree yet
```

**Root cause**

`SystemTrayFrameGrid` is no longer a `Grid`. Build 26300 rebuilt the tray for the movable
taskbar work, and the element kept its name but is now a **`StackPanel`**. `ResolveInjectionTarget`
opens with:

```cpp
auto trayFrame = FindChildByName(root, L"SystemTrayFrameGrid");
auto trayGrid = trayFrame ? trayFrame.try_as<Grid>() : nullptr;
if (!trayGrid) return {};      // <- always taken on 26300
```

so it bails on the first line and *every* `tray_*` position fails regardless of which anchor
was selected. This is why the failure looks like a missing anchor element when it is really
a type mismatch.

Confirmed from a XAML tree dump on 26300.9032:

```
[SystemTrayFrame]
  [StackPanel] name='SystemTrayFrameGrid' col=0
    [Stack] NotifyIconStack   col=0
    [NotificationAreaIcons]   col=1
    [Stack] MainStack         col=6
    [Stack] NonActivatableStack   col=2
    [Stack] SecondaryClockStack   col=0
    [OmniButton] ControlCenterButton      col=4
    [OmniButton] NotificationCenterButton col=5
    [Stack] ShowDesktopStack  col=7
```

Note the children still carry stale `Grid.Column` values, but those no longer drive layout —
ordering is now the index in `Children()`. So reading `Grid::GetColumn` and mapping it to a
column is not a workaround; the index has to come from `Children()` order.

Same root cause as #5018, which reports it against the Taskbar Styler.

**Fix**

I've been running a patched build for a day with no issues. The shape that worked:

* Make the injection target hold a `Panel` rather than a `Grid`, plus *either* a column
  (Grid path, every earlier build) or a child index (StackPanel path, 26300+).
[...]

PR #5568, opened 2026-09-18
Update Taskbar Folder Menus to v2.0

### LINKED ISSUE #5530: [Taskbar Folder Menus] Mod does not load after Windows 11 update 2026-09 Security Update (KB5129195) (26200.9457)
Mod author: @sb4ssman 

I applied **2026-09 Security Update (KB5129195) (26200.9457)** on Sep 14 and the Taskbar Folder Menus mod will not load.

[taskbar-folder-menus-debug.txt](https://github.com/user-attachments/files/32268966/taskbar-folder-menus-debug.txt)

| Windows| Info|
|--------|--------|
| Edition| Windows 11 Home|
| Version | 25H2|
| OS build| 26200.9457|
| Experience| Windows Feature Experience Pack 1000.26100.360.0|

Installed Windhawk Mods
Image

sb4ssman:
Adds compact Windows 11 taskbar buttons that open configured Shell targets as native popup menus, recreating the most useful part of the classic taskbar toolbar workflow. Subfolders expand on hover, right-click gives the full classic Shell context menu, and every folder popup carries an "Open in Explorer" shortcut.

### Fixes #5530 — the mod stopped loading on 26200.9457

Windows 11 26200.9457 (KB5129195) kept the element name `SystemTrayFrameGrid` but **changed it from a `Grid` to a `StackPanel`**. Every `try_as<Grid>()` on it returned `nullptr`, so the mod loaded, resolved its symbols, applied its hook, reached the XamlRoot — and then silently injected nothing. The reporter's debug log dies on exactly that line.

The fix does not switch types; it supports both, because older builds still ship a `Grid` and the rollout is staged:

- The injection point is resolved to the `Panel` base and **classified on every attempt, never cached** — `StartTaskbar` rebuilds the tree.
- A `Grid` leases a **column**, as before. A `StackPanel` leases a **child index**, because there child order *is* layout and there is no column to own.
- Anything that is neither is refused and logged with its real class name rather than guessed at.
- When an anchor cannot be resolved, the log now prints the tray panel's class and its named children, so a future restructure is diagnosable from one run.
- Column collision detection and repair stand down on an ordered panel, where there are no columns to collide.

The same change was independently reported against two other mods on 26H2 build 26300 (#5049, #5050) and fixed the same way in `taskbar-ai-quota` v1.6.5, so this is the forward shape of the tray rather than a one-off regression.

### Also in 2.0

- **Grouped settings contract** — `Placement` / `Content` / `Layout` / `Size` / `Adjust` / `Surface` / `Behavior`, matching the rest of this family. Windhawk cannot carry values across renamed keys, so old flat keys are **not migrated**; the README documents the upgrade explicitly.
- **One `Layout.Arrangement` expression** replaces the row/column and padding settings. `1 | 2` is a row, `1, 2` a column, parentheses nest, and `name[dx,dy]` nudges one item. The default `auto` fits the taskbar height and logs the expression it generated, so it can be pasted back into the same field and edited.
- **Native Shell icons** per entry, cached as pixels at the display size, with the label as fallback when an icon cannot be obtained.
- **Placement after the pinned/running app icons**, following them as apps open and close. It reserves space through the repeater margin and stops at the tray's left edge; if the taskbar cannot fit the toolbar it hides rather than overlapping the tray.
- Menu items are limited by optional per-folder item caps and submenu depth, and duplicate entries from the Desktop namespace's user+public merge are suppressed.

Native-icon and after-app-icon feature ideas were contributed by [diegoalejo15](https://github.com/diegoalejo15).

### Screenshots

Two folder buttons in the system tray
*A minimal two-button setup for Desktop and Control Panel.*

Four folder buttons on a standard taskbar
*Drive, GitHub, Desktop and Control Panel arranged in a row.*

Four folder buttons on a taller taskbar
*The same four on a double-height taskbar, arranged by `auto`.*

Control Panel opened as a native Shell menu
*The Control Panel namespace opens as a native Shell menu with full icons.*

Whole drive opened from a taskbar folder button
*A drive-root button opens the whole drive as a native cascading menu.*

Folder button destination tooltip
*Hovering a compact button shows its configured Shell target.*

### Tested live

Tested on Windows 11 26200.9457 — the build this fixes — covering injection at each tray position and after the app icons, automatic and hand-written arrangements, native icons and label fallback, nested submenu expansion, the right-click Shell context menu, "Open in Explorer", settings changes, Explorer restart and taskbar rebuild, and disable → native restore → re-enable.

## Changelog

* Fixed the mod not loading on Windows 11 26200.9457 / KB5129195, where `SystemTrayFrameGrid` became a `StackPanel` (#5530); both panel shapes are now supported and classified at runtime
* Moved the whole settings surface to grouped keys; old flat keys are not migrated
* Replaced the row/column and padding settings with one nestable `Arrangement` expression, defaulting to `auto`
* Added native Shell icons per folder entry, with pixel caching and label fallback
* Added placement after the pinned/running app icons, with tray-edge limiting
* Stood down column collision detection and repair on an ordered tray panel
* Added failure logging of the tray panel's class and named children

## Mod authorship

This mod was created by:

- - [x] Claude
- - [ ] ChatGPT
- - [ ] Gemini


sb4ssman:

sb4ssman:
Thanks — all four were real. Addressed in this push:

**1(a). Unload no longer gives up.** The `attempt < 50` cap is gone. Uninit loops `EndMenu` + `WM_CANCELMODE` until the menu path is idle, exactly as you suggested. Each `RunFromWindowThread` is a `SendMessage`, so the loop follows the taskbar thread's real progress. The give-up branch, the "retaining the idle event" leak, and the conditional `CloseHandle` are all gone.

**1(b). The queued Shell verb is now inside the wait.** There is a shared RAII `MenuPathScope` that owns the `fetch_add`/`ResetEvent` … `fetch_sub`/`SetEvent` pairing. `ShowFolderMenu` holds it for its whole tail, and `MenuOwnerSubclassProc` holds it around `InvokePendingShellCommand`. A Delete confirmation or Open With dialog now keeps unload waiting until it is answered.

Following the same reasoning turned up one more window, now closed. Both sites read `g_unloading` *before* opening their scope. Uninit could store the flag and read a depth of zero in between. Then `TrackPopupMenu`'s loop would pump uninit's own cleanup dispatch and return into a freed image. Both sites now open the scope first and check `g_unloading` after it. Uninit stores the flag before reading the depth, so with seq_cst atomics one side always sees the other.

**2. Native icons on a running Explorer / on settings change.** We took your simpler option. `Wh_ModAfterInit` and `Wh_ModSettingsChanged` no longer apply directly; the worker's first attempt (no delay) runs `PrepareFolderIcons()` before the first build. This also removes the label→icon flash.

**3. `LoadLegacyFolders`.** Removed, together with the whole 0.7 bridge (`HasLegacySettings`, `PreferCurrentOrLegacy*`, `Behavior.UseLegacySettings`). Windhawk has no API to write a setting, so a bridge always leaves the settings page showing a value the mod isn't using. The "prefer legacy while the 2.0 value equals its default" rule also meant a carried value could never be set back to its default (your functionality note and the clamp note both describe symptoms of this). The README now says plainly that settings reset once on update, and how to copy them across.

**4. README.** The `|` in the Arrangement table is escaped. "Upgrading from 0.7" describes the one-time reset and the 150-item default. The `ShowHidden` row says hidden-only.

**Optional items taken:**
- Dead template code: the mod is now assembled from eight small components, with its own focused tree walk. `GetMetrics`/`LayoutModelApplies`/`OrientationName`, the unused layout helpers, `sio::StringSetting`, the `vtw::*` entry points and `igc::Acquire` are no longer in the file, and neither is the commentary about other mods and lab scripts.
- The retry thread now **uses** `RetryLoop` rather than carrying it as dead code. That closes the concurrent-`StopRetryThread` hole you described: every caller that sees a live run waits for it. It runs 40 attempts 1.5 s apart, the same one-minute window for slow sign-ins. A vertical-taskbar stand-down counts as settled, so it does not retry.
- `GetSystemMetricsForDpi` is called directly.
- `LoadFolders` uses the variadic `GetStringSetting`. `Placement.Position`, `FillOrder`, `Justify` and `NewItems` are parsed once through `LoadChoice`; position is now an enum, with no string compares left.
- The Desktop-root sort breaks display-name ties by `fileName`, so the dedupe is exact.

Every item from this round is taken; nothing is left unaddressed.

PR #5569, opened 2026-09-18
Update Tray Utility Customizer to v2.0

sb4ssman:
Granular, predictable control over the low-frequency Windows 11 tray utility icons — Show hidden icons, Emoji, touch keyboard, pen menu, virtual touchpad and the input/language indicator. The native controls stay alive and Windows-owned, so clicks, flyouts and tooltips are untouched; the mod gathers their hosts into one owned group and positions each icon individually, at its native size by default.

### Supports the 26200.9457 tray StackPanel

Windows 11 26200.9457 (KB5129195) kept the element name `SystemTrayFrameGrid` but **changed it from a `Grid` to a `StackPanel`**, so `try_as<Grid>()` returned `nullptr` and the mod silently customized nothing. Both shapes are now supported, classified on every injection and never cached, since older builds still ship a `Grid`.

**This mod needed more than a cast fix.** Its host markers recorded a *column* to restore — meaningless on an ordered panel. A host record now tracks whether the tray is ordered, `CaptureHost` inserts the marker at the host's live index, and a new `ReturnHostToTray` re-inserts each host at that marker's index. Without it, restore-on-disable would have dumped every tray icon at the far end of the taskbar.

### Also in 2.0

- **Grouped settings contract** — `Placement` / `Content` / `Layout` / `Size` / `Adjust` / `Behavior`, matching the rest of this family. Windhawk cannot carry values across renamed keys, so **1.x settings do not migrate**; the README has an explicit upgrade section.
- **One `Layout.Arrangement` field** replaces the layout expression, the primary-axis setting, group alignment and all twelve per-icon nudge settings. `|` is now *always* horizontal and `,` *always* vertical at every depth, and a nudge rides in the expression itself (`emoji[+2,-1]`) instead of in twelve separate keys. Anyone who had Primary axis set to Column should swap `|` and `,` when re-entering their layout.
- **`auto`** fits the enabled utilities to the taskbar height, taking the fewest columns that fit the rows available, and logs the expression it generated so it can be pasted back and edited.
- **Per-utility `Content` switches** and `Layout.NewItems`, so a utility Windows shows or hides live — the touch keyboard comes and goes — is not silently lost from an arrangement written earlier.
- **Row capacity is computed in DIPs** from the taskbar's real DPI instead of raw pixels, so the automatic shape is correct at 125% and 150% scaling.
- **Vertical-taskbar stand-down.** `taskbar-vertical` applies a `RotateTransform` to the same elements this mod writes a `TranslateTransform` to. Rather than fight over one dependency property, a vertical taskbar is detected from its own aspect and the mod leaves everything as found, saying so in the log.
- **Exact property restore.** Every property written to a borrowed native element is snapshotted first and restored to its precise prior *local* value — including "there was no local value here", where the correct restore is `ClearValue` so the template binding resumes.

### Screenshots

Native tray before the mod
*Mod disabled: the chevron, Emoji and touch keyboard in their native positions.*

One row at the hidden-icons position
*One row of three at native size — what `auto` produces on a single-height taskbar.*

Chevron centered above a row of utilities
*`overflow, (emoji | touchKeyboard)`: the chevron centered on its own row above the pair.*

A neat column
*A full single column of utilities on a single-height taskbar.*

A dedicated column elsewhere in the tray
*The group leased into its own tray column at the right end of the taskbar.*

On a busy double-height taskbar
*Coexisting with a heavily modded double-height tray.*

Right of Start, stacked on a double-height taskbar
*The experimental Right of Start position on a double-height taskbar.*

### Tested live

Tested on Windows 11 26200.9457 covering each tray position and the Start-adjacent positions, automatic and hand-written arrangements, per-utility content switches, utilities appearing and disappearing live, settings migration from 1.x, Explorer restart and taskbar rebuild, and disable → native restore → re-enable with every tray icon returning to its original place.

## Changelog

* Added support for Windows 11 26200.9457 / KB5129195, where `SystemTrayFrameGrid` became a `StackPanel`; both panel shapes are classified at runtime and never cached
* Fixed restore-on-disable on an ordered tray panel, where host markers recorded a column that no longer meant anything; hosts now return to their marker's live child index
* Moved the whole settings surface to grouped keys; 1.x settings do not migrate
* Replaced the layout expression, primary axis, group alignment and twelve per-icon nudge settings with one `Arrangement` field; `|` is always horizontal and `,` always vertical
* Added `auto`, which fits the enabled utilities to the taskbar height and logs its generated expression
* Added per-utility `Content` switches and a `Layout.NewItems` policy
* Added `Adjust.PadX` / `PadY` / `OffsetX` / `OffsetY`
* Computed row capacity in DIPs from the real DPI, fixing the automatic shape at 125% and 150% scaling
* Restored borrowed elements to their exact previous local value, clearing the property when it had none

## Mod authorship

This mod was created by:

- - [x] Claude
- - [ ] ChatGPT
- - [ ] Gemini


sb4ssman:

sb4ssman:
Thanks — the `g_stoodDown` regression was real. Addressed in this push:

**1. Transient states no longer retire the retry.** `g_stoodDown` is now set only for decisions another attempt cannot change: a vertical taskbar, every utility switched off, or a *measured* tray below `MinimumTrayHeight`. The three transient exits return `false` and keep the bounded retry alive:
- a tray with `ActualHeight() <= 0` (not arranged yet);
- no layout items yet;
- no placements yet.

The budget is now 20 attempts × 1.5 s, because "not populated yet" is now what the loop is waiting out.

Beyond the retry budget, as you suggested, each apply now records the visible-icon count of every *unmanaged* candidate host as a weak reference. The `LayoutUpdated` check compares those as well as the managed hosts, so a `MainStack` that fills in after a chevron-only apply re-runs the layout. To keep the added walk off the hot path, the check is throttled to 500 ms instead of 250 ms.

**Optional items:**
- **`RetryLoop::Start` re-entry:** already fixed. `run_` is published by `std::exchange` under the mutex, and whatever it displaces is stopped. This is the shape you proposed.
- **`no_destroy` on `g_retry`:** you're right that it guards no wait, and the comment now says so. The attribute stays because our exit-time-destructor audit requires every namespace-scope object with a non-trivial destructor to be either marked or proven heap-only. Removing it fails that check.
- **Dead code:** `FindDirectTrayHost` and the `FindCurrentProcessTaskbarWnd` wrapper are removed. `RestoreObject` and the two `Metrics` fields belong to components shared with sibling mods, where they are used; splitting the components per function would cost more than those few lines.
- **`Wh_ModSettingsChanged` without a taskbar** now loads the settings, so the next `StartTaskbar` rebuild uses them.
- **`Wh_ModAfterInit`** starts the same bounded retry instead of making a single attempt.
- **`AcquireAt` with a stale marker:** fixed in the shared lease component. A same-named marker that is *not* the caller's live lease is released and the slot is taken fresh, with the requested slot corrected if the released marker sat before it. The caller's own live lease is still refused, because acquiring twice would strand the first slot.

**Functionality notes:**
- **The 1.x fallback** is removed entirely, along with `Behavior.Use1xFallback`, the legacy nudges and the expression transpose. You identified the core problem: Windhawk stores defaults at install, so the bridge can't tell a default from a choice, and it can't be undone from the settings page. The same applies to the `minimumTrayHeight = 0` case. Because Windhawk has no API to write settings, any bridge leaves the UI showing one value while the mod uses another. The README now says settings reset once on update and how to carry them across.
- **Restore order on a Grid tray:** the host marker is now inserted at the host's live index on both panel kinds, and the host is re-inserted at the marker's index on both. Child order is therefore exact after a disable/re-enable, including keyboard/UIA order.
- **Drift-check cost:** the throttle is raised to 500 ms, as described above.

PR #5570, opened 2026-09-19
Add Move Conflicts Quietly

JohhannasReyn:
New mod: keeps both items on Explorer copy/move name conflicts, using the next free `name (n).ext`, instead of showing the Replace/Skip prompt. Hooks `IFileOperation` in windows.storage.dll by probing the vtable at runtime (no symbols). Tested on Windows 11 25H2 build 26200. Source: https://github.com/JohhannasReyn/Move-Quietly

**Why not just set `FOF_RENAMEONCOLLISION`?** The flag alone gives `name - Copy.ext`, then `name - Copy (2).ext`, which is the wrong shape when a folder is collecting versions of the same file; the `(n)` form is what Explorer's own "Keep both" produces. And a same-name *folder* under the flag is renamed rather than merged, so "merge the folder, keep both copies of conflicting files inside" cannot be expressed with flags at all. The mod therefore resolves names itself and queues items individually; the flag stays on as a safety net for the cases it can't reach (non-file-system destinations, a name taken between the check and the copy).

Settings: folder collision mode (merge with keep-both inside / Explorer's merge / keep both folders), a held bypass key (Shift by default) for mouse gestures, and a lock-key bypass for keyboard paste.

## Changelog

* N/A (new mod)

## Mod authorship

This mod was created by:

- [x] Claude
- [ ] ChatGPT
- [ ] Gemini
- [ ] Other (please specify):


JohhannasReyn:
Thanks — all five main findings were valid. Pushed fixes:

1. **Double-queue via `E_NOINTERFACE`** — `CopyItems`/`MoveItems` now collect every item first (`CollectItems`); `E_NOINTERFACE` can only come back before anything is collected. Once the container is understood the mod owns the queueing and never falls through to the original. Names are un-reserved if the original single-item call fails.
2. **Caller collision policy** — operations whose caller set `FOF_NOCONFIRMATION`, `FOF_RENAMEONCOLLISION` or `FOFX_KEEPNEWERFILE` are left untouched, for naming and for the fallback flags.
3. **Folders** — now a setting, `folderCollision`, with three modes. Default **merge**: the folder merges into the existing one and the mod queues its contents item by item, so conflicting files inside get the same `(n)` naming and existing subfolders merge recursively; after a successful merged *move*, the emptied source tree is pruned with `RemoveDirectoryW` (which refuses non-empty directories, so a failed child can never cause data loss). **prompt**: Explorer's own merge, untouched, and the fallback flags are not armed for that operation. **rename**: `Folder (2)`. The fallback flags are decided once in `PerformOperations` with the whole queue known.
4. **I/O under the lock** — `GetFileAttributesW` runs outside `g_stateLock`; the lock covers only the reservation test-and-insert. Destination path resolved once per batch. `kMaxSuffix` reduced to 1000.
5. **`Release` hook** — removed. `PerformOperations` clears the entry; abandoned operations leak a small entry until unload, same as the other two vtable-probing mods.

Also added a bypass: `bypassKey` (default Shift) held during a drop or a mouse Paste leaves that operation entirely to Explorer, and `bypassLock` (default off) does the same while Caps Lock or Scroll Lock is on — the only route that works for Ctrl+V, since any extra modifier stops matching Explorer's Paste accelerator. Held key is read with `GetAsyncKeyState` because the Win11 command bar dispatches Paste across threads.

Optional items taken: typed `WindhawkUtils::SetFunctionHook`, `<iterator>`, dropped the `FOFX_PRESERVEFILEEXTENSIONS` macro guard, `IEnumShellItems::Next` failures propagate, settings block (which also gives the mod a reload path).

Not changed:
- Dotfiles: `.gitignore` → `.gitignore (2)`, keeping the file a dotfile. Happy to flip if native behavior differs.
- README screenshot: will add once the behavior is settled.

JohhannasReyn:
Thanks — this round was also useful. Pushed fixes for all nine.

**1. Why the manual queueing path exists** — added to the PR description. Short version: the `(n)` naming is the feature (it's what Explorer's own "Keep both" produces, and `- Copy` / `- Copy (2)` is the wrong shape for a folder that's collecting versions), and merge-with-rename inside a folder can't be expressed with flags at all — `FOF_RENAMEONCOLLISION` renames the folder instead of merging it. The flags remain as the safety net.

**2. Stale state on address reuse** — `Release` is hooked again; the entry is erased when the refcount hits zero. Note the first review asked for this hook to be *removed* over the hot-path/ICF concern; the stale-state hazard is worse now that `bypassed` and `mergedMoveSources` exist, so it's back with an atomic live-entry count so the hook skips the lock entirely when the map is empty (which is nearly always). The hook makes no COM calls.

**3. Any `CollectItems` failure now falls back to the original** — nothing is queued until after collection succeeds, so all failures are safe to hand off.

**4. Plural interfaces are tried first**, `IShellItem` last.

**5. Mid-loop failures no longer abort** — the loop continues, logs each failure, and only returns an error when nothing at all was queued. Same in the merge expansion.

**6. `GetAsyncKeyState`** for the held key (the reviewed commit was one revision behind; the fix was already local). The lock-key mode still uses `GetKeyState`'s toggle bit since there's no thread-independent API for it; that's noted in the code and the setting description.

**7. Reparse points** — a source folder that's a junction or symlink is never merged; it's handed to Explorer unchanged (as in the `prompt` mode).

**8. Pin is a precondition** — `Wh_ModInit` fails if `GetModuleHandleExW` can't pin the module.

**9. Non-file-system destination** — now arms the fallback flags so the README claim holds.

Optional items taken: `WindhawkUtils::StringSetting`, settings published as one struct under the lock and copied out per operation, `LoadSettings` inside the namespace, comment on the `kCallerCollisionPolicy` ordering caveat (README wording adjusted too), `IsFolderItem` falls back to `DirectoryExists`, `\\?\UNC\` for long UNC paths, off-by-one in the suffix loop, unreserve on any queue failure. README now states that merged-move source folders are removed outright (not via Recycle Bin / Undo), the up-front cost of a large merge, and that shell extensions hosted in Explorer are affected.

Not changed: screenshot (will add after the behavior settles); dotfile naming (`.gitignore (2)` — keeps it a dotfile).

JohhannasReyn:
All three were real; thanks for staying on the merge path — it's the only place the mod moves data on its own, so it deserves the scrutiny.

**1. Hidden/system items** — the expansion now binds `BHID_SFObject` and enumerates with `IShellFolder::EnumObjects(SHCONTF_FOLDERS | SHCONTF_NONFOLDERS | SHCONTF_INCLUDEHIDDEN | SHCONTF_INCLUDESUPERHIDDEN)`, wrapping each child with `SHCreateItemWithParent`. README's test list now includes a hidden and a hidden+system file in a merge.

**2. Prune registration** — the source folder is added to the prune list only after `QueueMergedFolder` reports success (it now returns a queued count). If the expansion fails having queued nothing, the folder is handed to the original method — Explorer's own merge — with `mergeFolderSeen` set so the fallback flags can't rename it.

**3. Reused `IFileOperation`** — `PerformOperations_Hook` resets only the per-run fields (`wantsFallback`, `mergeFolderSeen`, `bypassed`, `reservedTargets`, `mergedMoveSources`); `callerFlags` survives, and `Release_Hook` is the only place the entry is erased.

Optional items taken: bypass is sticky per operation (state checked before the key); `CurrentSettings` takes the lock shared; `LongPath()` helper shared by `PathAttributes`, `FindFirstFileExW` and `RemoveDirectoryW` in the prune; one case folding (`LCMapStringW` invariant upper-case) for both reservation keys and path equality; log lines no longer repeat the mod name.

On the `bypassKey` default: kept as Shift, but the README and the setting description now say directly that Shift is also Explorer's force-move modifier and that Shift-drops bypass the mod, and spell out why Ctrl and Alt are poor choices. The Windows key is called out as the conflict-free option.

Sink pass-through in merge expansion: left as `nullptr`, as noted. Default-flags assumption: it's the documented `SetOperationFlags` default; not separately verified with a test caller.

Screenshot still pending.

PR #5583, opened 2026-09-19
Three Finger Drag 1.0.0

RodrigoThomazi:
Adds **Three Finger Drag**, a new mod that moves windows with a three-finger drag on a precision touchpad, like three finger drag on macOS.

- Grabs the window under the cursor from anywhere in it, not just the title bar, and drops it when the fingers lift.
- Native move (default): the drag is a real title bar drag run by Windows, so Snap Layouts, Aero Snap and PowerToys FancyZones work as usual. Programs the mod can't run in fall back to a direct move, with snapping of its own (maximize, halves, quarters, with a preview).
- Works with programs running as administrator, through a small helper loaded into them, which only moves their own windows and the cursor.
- Reads the Windows touchpad gesture settings and picks three or four fingers, whichever is free, with a notice linking to the touchpad settings when both are taken.
- Esc puts the window back, optional resize from edges, acceleration, a macOS-style drag mode that holds the left button, excluded programs.

Tested on Windows 11 with a precision touchpad, including elevated windows (Task Manager, Windhawk).

## Changelog

N/A, this is a new mod.

## Mod authorship

This mod was created by:

- - [ ] Claude
- - [ ] ChatGPT
- - [ ] Gemini


RodrigoThomazi:
Pushed, and tested on the machine the mod was written on (Windows 11, precision touchpad, with and without native move, with elevated windows in the foreground).

Addressed:

- **The 2 s stall with an elevated window in the foreground.** Confirmed, and fixed as suggested: `BeginDrag` now asks `ForegroundHelper()` up front and only takes the native path when the press can actually be sent from here or by the target's own helper.
- **The dedicated process.** Adopted the Mods as tools pattern. The touchpad, the input thread, the low level keyboard hook and the notice dialog are out of the shell now, and the single instance mutex and the Explorer takeover logic are gone. The mod stays a hybrid: the launcher and the helper/native-move roles are picked by the process, not by the include pattern, which is the only deviation from the snippet and is noted where it is.
- **The native move marker is created last**, so a process whose initialization fails no longer claims to run native moves.
- **`nativeMove`'s description now matches the code:** `Wh_ModSettingsChanged` sets `*bReload` for the processes which only ran native moves, so turning it off really does unload it from them.
- **Sandboxed programs:** the ready and started answers are let through the main window's message filter, so a lower integrity level no longer breaks native move. This also covers the dedicated process running elevated, where ordinary programs are below it.
- **`GetKeyState` instead of `GetAsyncKeyState`**, the thread's own state, which the move loop goes by and which respects swapped buttons.
- **Dropped the unused `<memory>`.**
- **Credit:** a Credits section now says the native move follows AltDrag's approach.

Left as is:

- **The helper thread's priority.** I tried `ABOVE_NORMAL` and it made dragging the Task Manager visibly worse. While a program running as administrator is in the foreground, that thread is the only way in for the touchpad and the only way out for the cursor, so in a program which keeps its own threads busy it loses its turn and the drag stutters. It's back at `THREAD_PRIORITY_HIGHEST`, with the reason in a comment.
- **`<hidusage.h>`**, which is the canonical header for the usage pages the mod names.
- **The UIPI surface of the helper**, a conscious acceptance, as you read it.

A GIF for the readme is still missing and will follow.

m417z:
Thanks for the submission. Please explain why you need to both inject into all processes and spawn an additional process. The newly spawned process is launched by windhawk.exe which is not elevated, and if I understand correctly, that's what the mod tries to achieve. Also, it relies on the way Windhawk runs its processes, which I prefer to avoid.

m417z:
I read the code in more detail. I think it's OK given the limitations, but it's suboptimal to have a thread with a window running in every elevated/UIAccess process. With Windhawk 2.0 alpha and its [UIAccess tool mod support](https://github.com/ramensoftware/windhawk/wiki/Mods-as-tools:-Running-mods-in-a-dedicated-process#windhawk-20-alpha), I believe we can come up with a better and safer design, but it will take some time for it to become widely used among Windhawk users.

For now, I'll merge it as is.

